PR #24, opened 2018-04-16
Remove mesh "ordering"

chrisrichardson:
Removed all the mesh vertex ordering code (which doesn't work, anyway for quad/hex).
I had to skip some tests to make this branch pass.


python/test/unit/mesh/test_mesh.py
```diff
@@ -443,7 +443,7 @@ def test_mesh_topology_against_fiat(mesh_factory, ghost_mode):
                 assert all(vertex_global_indices[numpy.array(entity_topology)]
                            == entity_vertices)
 
-
+@pytest.mark.skip
```
garth-wells: Could you add a comment here so we remember why it's been disabled?

chrisrichardson:
Depends what you mean by ordering. They will still have the same relation to the cells as before - it is just that the numerical sorting of the indices will not be done. Looking at the test (test_mesh_ufc_ordering), it seems to be fairly broken - I might just delete it entirely.

garth-wells:
I mean will the indices for a facet have some sensible ordering, e.g. consistently clockwise or consistently ant-clockwise, etc?

PR #50, opened 2018-04-30
Remove TimingClear enum

chrisrichardson:
Removes the TimingClear enum which was always needed every time you asked for a timing.
Should just be a function to clear the timers (if needed).


blechta:
Please update also `demo_timing.py`.

PR #51, opened 2018-04-30
Permute cell points in constructor and for I/O

chrisrichardson:
Permutes the cell points from VTK ordering to UFC ordering internally, and reverses the permutation when outputting with XDMF.


michalhabera:
Permutations are becoming a confusion now. Do not forget to remove `CellType::vtk_mapping()`. 
Why is topology computation for higher order meshes in body of `XDMFFile::add_topology_data` but for linear meshes in `XDMFFile::compute_topology_data`?  
Is the mesh constructor the best place where this permutation could be stored?

From the point of view of my work - as far as ordering in mesh after whatever permutation is compliant to VTK, write/read checkpoint won't be affected.

chrisrichardson:
@michalhabera - I know there is still `vtk_mapping()`... and there are still some things that need improving... it might be better in another branch?

Issue #20, opened 2018-04-10
Sort out naming in XDMFFile method names


garth-wells:
From @blechta 

It has been agreed in a discussion that following renaming scheme is favourable

```
class XDMFFile:
    read_checkpoint() -> read()
    write_checkpoint() -> write()
    write() -> write_vertex_values()
```

Arguably `write_vertex_values()` could be removed once support for `write_checkpoint()` format in Paraview is packaged and distributed well. User can instead do more explicit operation involving some kind of interpolation or projection.

See https://bitbucket.org/fenics-project/dolfinx/issues/27/sort-out-naming-in-xdmffile for comments.

michalhabera:
Addressed in https://github.com/FEniCS/dolfinx/tree/michal/sort-naming-xdmf
Instead of `read_checkpoint -> read` I applied `read_checkpoint -> read_function` to stay consistent with the rest of XDMF, where write is implicit and read is explicit (argument type in name).

But I don't like the wrapping in `python/dolfin/io.py` ... Did you guys (@garth-wells @chrisrichardson ) found a better way for this `dolfin.function.function.Function` casting?

garth-wells:
@michalhabera I'm tending toward nothing in the `cpp` namespace being 'public'. All Python functionality would be implemented in Python with calls the backend. In this case we would not have to do anything casting in the pybind11 code.

This would also help with documentation. The Python and C++ interface would have their own docs.

garth-wells:
Some will just be `import foo as bar`.

I don't expect it will be more work that what we already have, and likely less . In cases it will be less because we won't have to manage something one on of two things, e.g. a `dolfin.function.Function` or a `dolfin.cpp.function.Function`.

michalhabera:
The naming above is too complicated. We need to keep things easy.
With the proper and fast interpolation (https://github.com/FEniCS/dolfinx/issues/161) most of the under-the-hood interpolation should happen explicitly by user action. 

And in order to avoid new function memory allocation just to IO, we can have some local, cell-wise interpolation methods. These can work as iterators and cell by cell flush the values.

jorgensd:
@garth-wells  this is not resolved. Right now we have removed write_checkpoint, and we have to reintroduce it at some point.

PR #62, opened 2018-05-05
Sort out naming in XDMFFile methods

michalhabera:
Solves issue #20 . 

But there is still a thing to be discussed. New proposed name `write_vertex_values` is used also for `DG0` functions (and in future could be used also for VTKish XDMF for `P2` functions). Might be a bit misleading what vertex values for DG0 are...

### DIFF
--- a/python/src/io.cpp
+++ b/python/src/io.cpp
@@ -391,7 +391,7 @@ void io(py::module& m)
       .def("read_mvc_double", &dolfin::io::XDMFFile::read_mvc_double,
            py::arg("mesh"), py::arg("name") = "")
       // Checkpointing
-      .def("read", &dolfin::io::XDMFFile::read,
+      .def("_read_function", &dolfin::io::XDMFFile::read_function,
            py::arg("V"), py::arg("name"), py::arg("counter") = -1);
 }
 } // namespace dolfin_wrappers
--- a/python/src/io.cpp
+++ b/python/src/io.cpp
@@ -285,9 +285,9 @@ void io(py::module& m)
            py::arg("encoding") = dolfin::io::XDMFFile::Encoding::HDF5)
       // MeshValueCollection size_t
       .def("write",
-           py::overload_cast<const dolfin::mesh::MeshValueCollection<std::size_t>&,
-                             dolfin::io::XDMFFile::Encoding>(
-               &dolfin::io::XDMFFile::write),
+           py::overload_cast<
+               const dolfin::mesh::MeshValueCollection<std::size_t>&,
+               dolfin::io::XDMFFile::Encoding>(&dolfin::io::XDMFFile::write),
            py::arg("mvc"),
            py::arg("encoding") = dolfin::io::XDMFFile::Encoding::HDF5)
       // MeshValueCollection int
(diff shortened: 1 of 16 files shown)


python/src/io.cpp
```diff
@@ -229,17 +229,17 @@ void io(py::module& m)
   // dolfin::io::XDMFFile::write
   xdmf_file
       // Function
-      .def("write",
+      .def("write_vertex_values",
```
garth-wells: Use the C++14 style syntax from pybind11 for overloaded functions.

python/test/unit/io/test_XDMF_P2.py
```diff
@@ -34,4 +34,4 @@ def test_read_write_p2_function(tempdir):
```
garth-wells: Do we need another file? Can it all go in `test_XDMF.py`?

michalhabera: I don't know why is it separated. Guess that @chrisrichardson wanted to have P2 test alone to run them faster(?). He'll probably remove/merge tests after higher order geometry is settled.

But anyway XDMF tests file is getting a bit bloated. From clarity reasons I'd suggest to split it.

garth-wells: XDMF test file is bloated, but the fix is to clean it up rather than create a new file.

michalhabera:
@blechta I completely agree, the behaviour for "DG0" is the point I mentioned in PR description. Let me comment more on this:
1.) First, we want to distinguish two XDMF ways of writing a function to a file. The standard VTKish one (celldata for DG0 or pointdata for P1, P2) vs. the new FiniteElementFunction format.
2.) We want to let user know explicitly what is going on, since VTKish P2 is not yet implemented in dolfin. When you write in VTKish way P2 (or any other unsupported element) it saves just vertex values.

The solution I suggest:
Let us distinguish VTKish and FiniteElementFunction XDMF outputs on the method's name level, i.e. 
have e.g. `write_nodal_values` for VTKish and `write` for the new way. The fact that unsupported elements are interpolated to vertex values will be strongly emphasized in docs and in a warning message to user.

Agree, `XDMFFile::has_cell_centred_data()` is strange, because it is how XDMF thinks - there is this data being "centered" to cells, nodes, edges, etc. It is the XDMF specification. But let this be another issue, not related to this PR.

blechta:
Ok, `XDMFFile::has_cell_centred_data()` can be another issue. In fact it might not be needed at all if we follow the suggestion I will state now.

Let's have `write_vertex_values()`, `write_cell_values()`, `write_p2_values()` with the following behaviour:

* `write_vertex_values()` interpolates vertex values if needed and writes vertex values,
* `write_cell_values()` interpolates cell values if needed and writes cell values,
* `write_p2_values()` interpolates P2 values if needed and writes P2 values.

michalhabera:
You are addressing the problem 2.) I described. That problem is actually also present in VTKFile class = same cell centered data check and consequent interpolation to P1. In order to stay consistent we would have to apply `write_vertex_values, write_cell_values, write_p2_values` also in VTKFile.

@blechta The problem is more complex. There is still this sub-, iso-, and super-paramateric "elements" problem. Think of "write_vertex_values" for P1 function on P2 mesh. VTK has *only* isoparametric elements. We must save our P1 function as P2 function with interpolated values at midpoints for VTK (not implemented yet). Then user might be confused why write_vertex_values when it writes also midpoint values...

Maybe follow VTK terminology and use: 
1.) `write_cell_data` interpolates to (not necessarily isoparametric) P0 if needed and writes cell data,
2.) `write_point_data` interpolates to isoparametric P1 (or isoparametric P2 if implemented) if needed and writes. With proper warning to user if nontrivial interpolation (data loss) is happening.

We can use these names for both, VTKFile and XDMFFile with the same implementation meaning.

blechta:
Yes, that makes sense. XDMF/VTK provides two or three modes: cell data, vertex data, and maybe P2 data. All of them isoparametric. All of them should be invoked by a separate function in DOLFIN explicitly.

michalhabera:
Ok. I am closing this PR and updating the issue #20 .

PR #71, opened 2018-05-09
Move tabulate_tensor to Form

chrisrichardson:
Move tabulate_tensor to `Form` and relieves the assembler for the responsibility of updating
coefficients, or choosing which integral to use based on cell index.

* add a constructor to `Form` without any UFC integrals
* allow user-defined cell integrals (e.g. with numba) to be attached to `Form`
* add test for user-defined numba kernel

Obviously there is more to do, and much to clean up when we get rid of `SystemAssembler`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
             apt-get -y install python3-tk
       - run:
           name: Install/update Python dependencies
-          command: pip3 install decorator flake8 matplotlib pytest pytest-xdist sphinx sphinx_rtd_theme --upgrade
+          command: pip3 install decorator flake8 numba matplotlib pytest pytest-xdist sphinx sphinx_rtd_theme --upgrade
       - run:
           name: Install FEniCS Python components
           command: |
--- a/cpp/dolfin/fem/Form.cpp
+++ b/cpp/dolfin/fem/Form.cpp
@@ -250,7 +250,6 @@ void Form::tabulate_tensor(
     double* A, mesh::Cell cell,
     Eigen::Ref<const EigenRowArrayXXd> coordinate_dofs) const
 {
-
   // Switch integral based on domain from dx MeshFunction
   std::uint32_t idx = 0;
   if (dx)
--- a/cpp/dolfin/fem/Form.cpp
+++ b/cpp/dolfin/fem/Form.cpp
@@ -242,14 +242,14 @@ void Form::tabulate_tensor(
   // Switch integral based on domain from dx MeshFunction
   std::shared_ptr<const ufc_cell_integral> integral;
 
-  std::uint32_t i = 0;
+  std::uint32_t idx = 0;
   if (!dx)
     integral = _integrals.cell_integral();
   else
   {
-    // FIXME: range checks on cell and i
-    i = (*dx)[cell];
-    integral = _integrals.cell_integral(i);
+    // FIXME: range checks on cell and idx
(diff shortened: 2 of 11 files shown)


.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ jobs:
             apt-get -y install python3-tk
       - run:
           name: Install/update Python dependencies
-          command: pip3 install decorator flake8 matplotlib pytest pytest-xdist sphinx sphinx_rtd_theme --upgrade
+          command: pip3 install decorator flake8 numba matplotlib pytest pytest-xdist sphinx sphinx_rtd_theme --upgrade
```
garth-wells: Should numba go into the `Dockerfile`?

cpp/dolfin/fem/Form.cpp
```diff
@@ -61,6 +64,17 @@ Form::Form(std::shared_ptr<const ufc_form> ufc_form,
...
 //-----------------------------------------------------------------------------
+Form::Form(const std::vector<std::shared_ptr<const function::FunctionSpace>>
+               function_spaces)
+    : _coefficients({}), _function_spaces(function_spaces)
+{
+  // Set _mesh from function::FunctionSpace and check they are the same
+  if (!function_spaces.empty())
+    _mesh = function_spaces[0]->mesh();
+  for (auto& f : function_spaces)
+    assert(_mesh == f->mesh());
```
garth-wells: This should throw an exception rather than an assert.

cpp/dolfin/fem/Form.cpp
```diff
@@ -232,3 +246,46 @@ void Form::set_vertex_domains(
...
+  {
+    // FIXME: range checks on cell and idx
+    idx = (*dx)[cell];
+  }
+
+  const bool* enabled_coefficients = _integrals.cell_enabled_coefficients(idx);
+  // Restrict coefficients to cell
+  for (std::size_t i = 0; i < _coefficients.size(); ++i)
+  {
+    if (!enabled_coefficients[i])
```
garth-wells: Could this be
```
if (enabled_coefficients[i])
{
   // Do something
}
```

(no `continue`)

cpp/dolfin/fem/Form.cpp
```diff
@@ -232,3 +246,46 @@ void Form::set_vertex_domains(
...
+    const auto& element = _coefficients.element(i);
+    coefficient->restrict(_wpointer[i], element, cell, coordinate_dofs);
+  }
+
+  // Compute cell matrix
+  auto tab_fn = _integrals.cell_tabulate_tensor(idx);
+  tab_fn(A, _wpointer.data(), coordinate_dofs.data(), 1);
+}
+//-----------------------------------------------------------------------------
+void Form::initialise_w()
```
garth-wells: Could this have a better/more descriptive name?

cpp/dolfin/fem/Form.cpp
```diff
@@ -232,3 +246,46 @@ void Form::set_vertex_domains(
...
+{
+  const std::size_t num_coeffs = _coefficients.size();
+  std::vector<std::uint32_t> n = {0};
+  for (std::uint32_t i = 0; i < num_coeffs; ++i)
+  {
+    const auto& element = _coefficients.element(i);
```
garth-wells: This all needs some comments. Could you not use `auto` here to make the type clear.

PR #72, opened 2018-05-09
Make la::PETScMatrix and la::PETScVector compatible with PETSc-complex

IgorBaratta:
The purpose of this PR is to make `la::PETScMatrix `and `la::PETScVector` compatible with PETSc compiled with complex number support.

 - Change the scalar type that represents data (in vectors and matrices) from `double` to `PetscScalar`.
       - In  principle, there is no impact if petsc is built in normal mode.
- Add `PETScMatrix::is_hermitian` to test whether a matrix is Hermitian.


cpp/dolfin/la/PETScVector.h
```diff
@@ -126,21 +126,23 @@ class PETScVector
...
-  void axpy(double a, const PETScVector& x);
+  void axpy(PetscScalar a, const PETScVector& x);
 
   /// Replace all entries in the vector by their absolute values
   void abs();
 
   /// Return dot product with given vector
-  double dot(const PETScVector& v) const;
+  /// For complex vectors, the argument v gets
+  /// complex conjugate
```
michalhabera: Due to http://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Vec/VecDot.html the second argument gets the complex conjugate - which should be vector itself, not argument `v`. Double-check please.

michalhabera:
I also noted that `VecMin` and `VecMax` in PETScVector return `PetscReal` (double). Due to petsc source code it seems to take the minimum over real part of elements. This might create a lot of confusion, would be good if you can document it carefully.

cpp/dolfin/la/PETScVector.h
IgorBaratta: You're right. The way it is the vector itself gets conjugated.
However if by a.dot(b) we mean (a,b) the order of the arguments in VecDot should be changed.
For example, in petsc4py it's define that way.

michalhabera: Changing the order of arguments makes sense to me. Real case wont be affected.

PR #81, opened 2018-05-15
Add some complex compatibility changes

IgorBaratta:
- Add MPI definition for complex double  (`MPI_DOUBLE_COMPLEX`)
- Allow complex dot products in ` la::VectorSpaceBasis`


cpp/dolfin/la/PETScMatrix.cpp
```diff
@@ -452,6 +452,35 @@ void PETScMatrix::transpmult(const PETScVector& x, PETScVector& y) const
     petsc_error(ierr, __FILE__, "MatMultTranspose");
 }
 //-----------------------------------------------------------------------------
+void PETScMatrix::conjtranspmult(const PETScVector& x, PETScVector& y) const
+{
+  assert(_matA);
+
+  if (size(0) != x.size())
+  {
+    log::dolfin_error(
```
garth-wells: We're not using `dolfin_error` anymore. Switch to `throw std::foo` exceptions.

cpp/dolfin/la/PETScMatrix.cpp
```diff
@@ -452,6 +452,35 @@ void PETScMatrix::transpmult(const PETScVector& x, PETScVector& y) const
     petsc_error(ierr, __FILE__, "MatMultTranspose");
 }
 //-----------------------------------------------------------------------------
+void PETScMatrix::conjtranspmult(const PETScVector& x, PETScVector& y) const
```
garth-wells: Could be make the function name closer to the name of the PETSc function?

cpp/dolfin/la/VectorSpaceBasis.cpp
```diff
@@ -29,7 +29,7 @@ void VectorSpaceBasis::orthonormalize(double tol)
     // orthonormalized vectors
     for (std::size_t j = 0; j < i; ++j)
     {
-      const double dot_ij = _basis[i]->dot(*_basis[j]);
+      const auto dot_ij = _basis[i]->dot(*_basis[j]);
```
garth-wells: I'd like to avoid `auto` to make the type clear.

IgorBaratta:
> I'm wondering what degree we want to wrap PETSc calls vs just using PETSc. We used to support multiple backends, but now that we're PETSc only maybe we should just use PETSc and petsc4py?

I added the function to have a consistency between the complex and real interfaces.  But, as PETSc and petsc4py are now both required dependencies, it may be  a question of clean up rather than adding new functions for the complex case.

garth-wells:
Maybe we should remove `PETScMatrix::transpmult` - I don't think we want to wrap the whole PETSc interface. @chrisrichardson @blechta ?

chrisrichardson:
My question would be: what is it for? Do we need it internally? If not, then it is part of the user interface, which can be done with PETSc/petsc4py directly.

garth-wells:
Run `clang-format` on the code (the config file for `clang-format` is in the root of the repo).

It's possible to configure editors to apply it for you.

PR #83, opened 2018-05-16
Re-enable ghost shared vertex

chrisrichardson:
This re-enables the 'ghost via shared vertex' ghost method.

*  fixes a bug where the global number of cells was incorrect for ghosted meshes.

* also improves ghost mesh tests

### DIFF
--- a/cpp/dolfin/mesh/MeshPartitioning.cpp
+++ b/cpp/dolfin/mesh/MeshPartitioning.cpp
@@ -182,8 +182,9 @@ mesh::Mesh MeshPartitioning::build(
   timer.stop();
 
   // Build mesh from points and distributed cells
+  const std::int32_t num_ghosts = new_cell_vertices.rows() - num_regular_cells;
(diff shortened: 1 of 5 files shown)


cpp/dolfin/mesh/MeshPartitioning.cpp
```diff
@@ -287,209 +285,201 @@ MeshPartitioning::reorder_cells_gps(
...
+      sendv.push_back(vc_it.first);
+      for (Eigen::Index v = 0; v < cell_vertices.cols(); ++v)
+      {
+        if (cell_vertices(q, v) != vc_it.first)
+          sendv.push_back(cell_vertices(q, v));
+      }
+    }
+  }
+
+  MPI::all_to_all(mpi_comm, send_vertcells, recv_vertcells);
```
garth-wells: Is avoiding all-to-all for later?

chrisrichardson: Yes, it's another issue.

PR #87, opened 2018-05-18
Add complex suppport to la::SLEPcEigenSolver

IgorBaratta:
Although SLEPc separates the real and imaginary part of eigenvalues and eigenvectors, the scalar type is complex (with zero imaginary part) when PETSc is compiled with complex support.   

Changes:
* Set eigenvalues scalars to PetscScalar
* Convert spectral_shift to PetscScalar before calling EPSSetTarget

Note: The parameter "spectral_shift" is still real. For complex spectral shift: use complex parameters or a parameter for real and imaginary parts?

### DIFF
--- a/cpp/dolfin/la/SLEPcEigenSolver.h
+++ b/cpp/dolfin/la/SLEPcEigenSolver.h
@@ -126,7 +126,7 @@ public:
   void solve(std::int64_t n);
 
   /// Get ith eigenvalue
-  void get_eigenvalue(PetscScalar& lr, PetscScalar& lc, std::size_t i) const;
+  std::complex<PetscReal> get_eigenvalue(std::size_t i) const;
 
   /// Get ith eigenpair
   void get_eigenpair(PetscScalar& lr, PetscScalar& lc, PETScVector& r,
--- a/cpp/dolfin/la/SLEPcEigenSolver.h
+++ b/cpp/dolfin/la/SLEPcEigenSolver.h
@@ -126,11 +126,11 @@ public:
   void solve(std::int64_t n);
 
   /// Get ith eigenvalue
-  void get_eigenvalue(double& lr, double& lc, std::size_t i) const;
+  void get_eigenvalue(PetscScalar& lr, PetscScalar& lc, std::size_t i) const;
 
   /// Get ith eigenpair
-  void get_eigenpair(double& lr, double& lc, PETScVector& r, PETScVector& c,
-                     std::size_t i) const;
+  void get_eigenpair(PetscScalar& lr, PetscScalar& lc, PETScVector& r,
+                     PETScVector& c, std::size_t i) const;
 
   /// Get the number of iterations used by the solver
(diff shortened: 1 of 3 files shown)


cpp/dolfin/la/SLEPcEigenSolver.h
```diff
@@ -126,11 +126,11 @@ class SLEPcEigenSolver : public common::Variable
...
 
   /// Get ith eigenvalue
-  void get_eigenvalue(double& lr, double& lc, std::size_t i) const;
+  void get_eigenvalue(PetscScalar& lr, PetscScalar& lc, std::size_t i) const;
 
   /// Get ith eigenpair
-  void get_eigenpair(double& lr, double& lc, PETScVector& r, PETScVector& c,
-                     std::size_t i) const;
+  void get_eigenpair(PetscScalar& lr, PetscScalar& lc, PETScVector& r,
+                     PETScVector& c, std::size_t i) const;
```
garth-wells: Could we change this to:
```
std::tuple<std::complex, PETScVector, PETScVector> get_eigenpair(std::size_t i) const;
```
Does SLEPc create the Vec for the eigenvector (which would avoid copies)? Maybe not?

This change needs some thought.

garth-wells: Forget this suggestion. Doesn't map well onto SLEPc.

IgorBaratta: It makes sense to change only `get_eigenvalue`?

In PETSc-complex the eigenvalue is stored directly in lr and the eigenvector in Vr.  The imaginary part is set to zero.
Since PETSc doesn't support pointers to real or imaginary part of a complex vector, maybe we could overlad these functions:
```
void get_eigenpair(double& lr, double& lc, PETScVector& r, PETScVector& c, std::size_t i) const;
void get_eigenpair(std::complex<double>& l, PETScVector& V, std::size_t i) const;

```
```
void get_eigenvalue(std::complex<double> l, std::size_t i) const;
void get_eigenvalue(double& lr, double& lc, std::size_t i) const;
```

IgorBaratta: Ok..Updated also python wrapper.

python/src/la.cpp
```diff
@@ -814,9 +814,7 @@ void la(py::module& m)
                         & dolfin::la::SLEPcEigenSolver::solve)
       .def("get_eigenvalue",
            [](dolfin::la::SLEPcEigenSolver& self, std::size_t i) {
-             double lr, lc;
-             self.get_eigenvalue(lr, lc, i);
-             return py::make_tuple(lr, lc);
+             return self.get_eigenvalue(i);
```
garth-wells: `.def("get_eigenvalue", &dolfin::la::SLEPcEigenSolver::get_eigenpair)` should be enough with the new C++ interface.

PR #93, opened 2018-05-24
Update python la to handle complex numbers

IgorBaratta:
- Update the  `la` python wrapper to handle complex numbers, (when PetscScalar type is complex).
- Fix norm calculation for complex vectors in `la::VectorSpaceBasis`


python/src/la.cpp
```diff
@@ -185,13 +185,14 @@ void la(py::module& m)
       .def("set", py::overload_cast<PetscScalar>(&dolfin::la::PETScVector::set))
       .def("get_local",
            [](const dolfin::la::PETScVector& self) {
-             std::vector<double> values;
+             std::vector<PetscScalar> values;
              self.get_local(values);
-             return py::array_t<double>(values.size(), values.data());
+             return py::array_t<PetscScalar>(values.size(), values.data());
```
garth-wells: Will this work when `PetscScalar` is complex? Will pybind11/NumPy know what to do with it?

michalhabera:
Isn't `<pybind11/complex.h>` header required for complex bindings?

python/src/la.cpp
IgorBaratta: Yes. It will understand PetscScalar as `std::complex<double>` and in that case the numpy array will have `dtype=numpy.complex128`.

IgorBaratta: I've tested this changes on the branch [igor/petsc-complex](https://github.com/FEniCS/dolfinx/tree/igor/complex-petsc), compiled with petsc and petsc4py with complex scalar. 
E.g., the following test passed: 
```
def test_vector_complex():
   assert PETSc.ScalarType is np.complex128
   n = 10
   arr = np.random.rand(n) + 1j*np.random.rand(n)
   vec = PETSc.Vec().createSeq(n)
   vec.setArray(arr)
   b = PETScVector(vec)
   assert np.allclose(b.get_local(), arr)
```

Issue #31, opened 2018-04-20
Fix assembly of elements that need orientation information


chrisrichardson:
Removing `mesh.order()` means that the alignment of dofs between cells is no longer given by the global vertex indices.
- [x] adjust code in `DofMapBuilder` for higher order (P3 and higher) elements
- [x] figure out how to support elements which need a global direction on facets (Hdiv, Hcurl)
* possibly by using global cell/vertex indices across boundary

garth-wells:
See also https://github.com/FEniCS/ffcx/issues/56

mscroggs:
This is done (for simplex meshes) in FEniCS/ffcx#224

PR #96, opened 2018-06-01
Define ufc_scalar for a complex UFC interface

IgorBaratta:
Define `ufc_scalar` for use in _ufc.h_.
This PR should be considered together with FFC [PR #31](https://github.com/FEniCS/ffcx/pull/31).

### DIFF
--- a/cpp/dolfin/common/types.h
+++ b/cpp/dolfin/common/types.h
@@ -13,9 +13,9 @@
 
 // Typedefs for ufc_scalar - to be used in UFC.h
 #ifdef PETSC_USE_SCALAR
-using ufc_scalar = std::complex<double>;
+using ufc_scalar_t = std::complex<double>;
 #else
-using ufc_scalar = double;
+using ufc_scalar_t = double;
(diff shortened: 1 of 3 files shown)


cpp/dolfin/common/types.h
```diff
@@ -7,9 +7,17 @@
 #pragma once
 
 #include <Eigen/Dense>
+#include <complex>
 #include <cstdint>
 #include <petscsys.h>
 
+// Typedefs for ufc_scalar - to be used in UFC.h
+#ifdef PETSC_USE_SCALAR
```
blechta: Is this a typo? Should be `PETSC_USE_SCALAR_COMPLEX`?

cpp/dolfin/common/types.h
```diff
@@ -7,9 +7,17 @@
 #pragma once
 
 #include <Eigen/Dense>
+#include <complex>
```
blechta: Shouldn't be moved to the ifdef below?

PR #102, opened 2018-06-16
Add dolfin version to JIT hash

blechta:
This is a backport of DOLFIN old commit [1211d17cbf1d35727a57b01c61127bc904becaf1](https://bitbucket.org/fenics-project/dolfin/commits/1211d17cbf1d35727a57b01c61127bc904becaf1)
where it fixed [problems on Bamboo](http://magpie.bpi.cam.ac.uk:8085/browse/DOL-DODO-PUTPY3-802).

### DIFF
--- a/python/dolfin/jit/jit.py
+++ b/python/dolfin/jit/jit.py
@@ -167,7 +167,7 @@ def compile_class(cpp_data):
         if hasattr(v, '_cpp_object') and isinstance(v._cpp_object, cpp.function.GenericFunction):
             property_str += '*'
 
-    hash_str = str(statements) + str(property_str)
+    hash_str = str(statements) + str(property_str) + cpp.__version__
     module_hash = hashlib.md5(hash_str.encode('utf-8')).hexdigest()
     module_name = "dolfin_" + name + "_" + module_hash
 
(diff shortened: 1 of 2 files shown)


blechta:
This is expression JIT and cpp code JIT, not FFC jit. So it depends on DOLFIN.

I'm not saying this is the best solution. Possibly one should hash all the inputs to `jit_generate`, including parameters which are telling how to link to DOLFIN. This was just a quick fix how to make a module rebuild when `libdolfin.so.2017.2` changed into `libdolfin.so.2018.1`, see http://magpie.bpi.cam.ac.uk:8085/browse/DOL-DODO-PUTPY3-802.

chrisrichardson:
Looks fine to me. Anyway, harmless.

garth-wells:
I don't agree about harmless. I don't believe it is necessary, but will leads to the future 'this must be needed. Let's keep it simple.

garth-wells:
Ok, this is an example of where the version is needed (at the moment).

The plans is to remove support for code like `Expression('x[0]', degree=1)`, since this can be more elegantly done in dolfinx using NumPy-style vectorisation.

On a wider note, it would be ideal if the auto-generated code (i.e., code generated transparently to the user)  has no DOLFIN dependencies. We should aim towards this.

chrisrichardson:
So, shall we merge or delete this PR?

blechta:
This is not clear to me. I don't consider the solution good. The issue is: 1. jitted modules have dependecies, hence 2. dependencies have to be hashed. But hashing dolfin version is not taking care of 2. completely.

(I pretty much opened the PR as a bug report, with backport of dolfin-old fix-hack.)

PR #112, opened 2018-06-28
Reorder complex values

IgorBaratta:
-  Move implementation of `DistributedMeshTools::reorder_values_by_global_indices` to private a template function `DistributedMeshTools::reorder_values` on header file. 
- Create public wrappers for real and complex `DistributedMeshTools::reorder_values_by_global_indices`

The purpose of these changes is to allow the reordering of complex values in XDMFFile.


michalhabera:
As we discussed. Do not put function's implementation into header files (reduces user code compilation time, hides function implementation details...), use explicit instantiation. 
I guess you also do not need to create new dummy private function template `reorder_values`. Just make `reorder_values_by_global_indices` a template function defined in .cpp file and just below its definition put its explicit instantiation for double and for complex.

cpp/dolfin/mesh/DistributedMeshTools.cpp
```diff
@@ -1101,13 +1102,17 @@ void DistributedMeshTools::init_facet_cell_connections(Mesh& mesh)
 EigenRowArrayXXd
```
garth-wells: Let's remove `DistributedMeshTools::reorder_points_by_global_indices(const Mesh& mesh)`. No longer needed.

cpp/dolfin/mesh/DistributedMeshTools.cpp
```diff
@@ -1101,13 +1102,17 @@ void DistributedMeshTools::init_facet_cell_connections(Mesh& mesh)
...
-  return reorder_values_by_global_indices(mesh.mpi_comm(),
+  return reorder_values_by_global_indices<double>(mesh.mpi_comm(),
                                           mesh.geometry().points(),
                                           mesh.geometry().global_indices());
 }
 //-----------------------------------------------------------------------------
-EigenRowArrayXXd DistributedMeshTools::reorder_values_by_global_indices(
-    MPI_Comm mpi_comm, const Eigen::Ref<const EigenRowArrayXXd>& values,
+template <typename Scalar>
+Eigen::Array<Scalar, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
```
garth-wells: `Scalar` is confusing to me. My immediate reaction is that it's a concrete type, as `PetscScalar` is. 

I would suggest using `T` or `X`.

cpp/dolfin/mesh/DistributedMeshTools.cpp
```diff
@@ -1172,3 +1181,15 @@ EigenRowArrayXXd DistributedMeshTools::reorder_values_by_global_indices(
...
+    MPI_Comm mpi_comm,
+    const Eigen::Ref<const EigenRowArrayXXd>& values,
+    const std::vector<std::int64_t>& global_indices);
+//-----------------------------------------------------------------------------
+template Eigen::Array<std::complex<double>, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
+DistributedMeshTools::reorder_values_by_global_indices<std::complex<double>>(
+    MPI_Comm mpi_comm,
+    const Eigen::Ref<const Eigen::Array<std::complex<double>, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>& values,
+    const std::vector<std::int64_t>& global_indices);
+//-----------------------------------------------------------------------------
```
garth-wells: This explicit instantiation in the `cpp` file looks strange to me and is confusing. The declaration is a template, but it looks like it's relying on some dodgy link-time checks to make things work.

As a public function, it should be generic (in which case the implementation needs to go into the header file), or have concrete types.

It's also a bit confusing see the typedef for `EigenRowArrayXXd`, and then a full declaration for the complex vase.

cpp/dolfin/mesh/DistributedMeshTools.h
```diff
@@ -75,17 +74,23 @@ class DistributedMeshTools
...
   ///    MPI Communicator
   /// @param values
   ///    Values to reorder
   /// @param global_indices
   ///    Global index for each row of values
-  static EigenRowArrayXXd reorder_values_by_global_indices(
-      MPI_Comm mpi_comm, const Eigen::Ref<const EigenRowArrayXXd>& values,
+  template <typename Scalar>
+  static Eigen::Array<Scalar, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
+  reorder_values_by_global_indices(
```
garth-wells: Remove 'values' from the name - it doesn't really add anything.

cpp/dolfin/mesh/DistributedMeshTools.cpp
michalhabera: @garth-wells Explicit instantiation is indeed not needed in this case, since compiler knows what functions to compile from their usage across .cpp files. However, as you say, when you try to link your own code with the method templated by type not used (e.g. eigen array of int) it'll fail the linking. 
There are at least 3 ways to handle this: 
1.) Expose template definition in header.
2.) Make some function a concrete implementation of `reorder_values_by_global_indices` and make `reorder_values_by_global_indices` private - as done in XDMFFile::write(meshfunction).
3.) Just template declaration in header, template definition in source and compiler deduction of type from its usage.

I like the most 3., since there are just 2 types to use with and method is not meant to be heavily used by users, but its your decision.

Ad `EigenRowArrayXXd`: I think @IgorBaratta suggested to typedef something like `EigenRowArrayXXc` for complex and make it look more symmetric, but you didn't want to add another typedef.

michalhabera:
Ok. To sum up for @IgorBaratta : Make it as XDMFFile::write(MeshFunction) is. Similar to what you had as first variant, but do not define private template in header file. Or put it in anonymous namespace.
Sorry for the confusion.

IgorBaratta:
Following the recommendations in the comments, the following changes have been made:
- Move  template function `reorder_values_by_global_indices` to anonymous namespace.
- Use `T` as the templated parameter instead of `Scalar`.
- Remove `DistributedMeshTools::reorder_points_by_global_indices(const Mesh& mesh)`.
- Remove explicit instantiations.

chrisrichardson:
Ready to merge, unless anyone has any remaining issues?

Issue #73, opened 2018-05-10
Make PETSc and petsc4py required dependencies


chrisrichardson:
Make PETSc and petsc4py both required dependencies in CMakeLists.txt

jhale:
Fixed in #111

chrisrichardson:
Actually, I think this was slightly wrong. petsc4py should not be a required dependency of the C++ build, only the Python build.

jhale:
Looking at `master`, that is how it is.

garth-wells:
The C++ build system should be fixed so it does not depend on petsc4py. All the necessary info is available via the PETSc headers.

chrisrichardson:
```

garth-wells:
We should get `generate-form-files.py` to take command lines arguments that are passed through to ffc.

PR #120, opened 2018-07-06
Fix some tests and demos to work in both real and complex modes

IgorBaratta:
Key changes:
 - Replace `*` by `inner`
 - Check position of Test and Trial functions in inner products


blechta:
Should we distinguish `skip_if_complex` and `xfail_if_complex` as a case which is not expected to work and a case which is broken (e.g, IO?)?

python/src/common.cpp
```diff
@@ -57,7 +57,7 @@ void common(py::module& m)
         "Return `True` if DOLFIN is configured with mpi4py");
   m.def("has_parmetis", &dolfin::has_parmetis);
   m.def("has_scotch", &dolfin::has_scotch);
-  m.def("has_petsc_complex", &dolfin::has_petsc_complex,
+  m.def("complex_mode", &dolfin::has_petsc_complex,
```
blechta: Why is `has_petsc_complex` wrapped as `complex_mode`? This is just confusing. It should be wrapped as `has_petsc_complex`.

python/src/common.cpp
```diff
@@ -57,7 +57,7 @@ void common(py::module& m)
...
         "Return `True` if DOLFIN is configured with PETSc compiled with "
         "scalars represented as complex numbers ");
```
blechta: `Return True if PETSc scalar is complex.`

python/test/unit/fem/test_numba_assembly.py
```diff
@@ -6,17 +6,20 @@
...
 
 from dolfin.jit.jit import ffc_jit
 
 # Define scalar_type here (change to np.complex128 for complex)
-# e.g. if has_petsc_complex(): scalar_type = np.complex128
-scalar_type = np.float64
+if complex_mode():
+    scalar_type = np.complex128
+else:
+    scalar_type = np.float64
```
blechta: Is petsc4py required dependency? If yes, use `petsc4py.PETSc.ScalarType`.

python/test/unit/fem/test_numba_assembly.py
```diff
@@ -6,21 +6,18 @@
...
 from numba import cfunc, types, carray, typeof
-import numpy as np
-
 from dolfin.jit.jit import ffc_jit
-
-# Define scalar_type here (change to np.complex128 for complex)
-# e.g. if has_petsc_complex(): scalar_type = np.complex128
-scalar_type = np.float64
+import numpy as np
+from petsc4py.PETSc import ScalarType
```
blechta: Can anybody confirm that `petsc4py` is a required dependency? @chrisrichardson @garth-wells

IgorBaratta: I assumed that petsc4py is a required dependency due the closed issue [73](https://github.com/FEniCS/dolfinx/issues/73), and it's in the list of required dependencies in [installation.rst](https://github.com/FEniCS/dolfinx/blob/ed9064cb3151f1804e71128f3c46e5cd0f94bd49/python/doc/source/installation.rst).
If it is not, I'll change back to the old code.

blechta: I see, I was just confused by a mess which I just reported https://github.com/FEniCS/dolfinx/issues/125

Issue #8, opened 2018-04-10
Add support for h5py


garth-wells:
Allow use if h5py (http://www.h5py.org/) for advanced manipulation of HDF5 files.

Maybe it can be supported in a similar way to petsc4py. It would make the DOLFIN-side wrapping of HDF5 simpler.

garth-wells:
Not really required now - mesh constructors, etc can now work with plain data arrays so users can freely use `h5py` and then just pass the data to the initialisers.

PR #122, opened 2018-07-06
Add file modes to XDMFFile, put encoding to constructor

michalhabera:
`XDMFFile`'s encoding should really be a state-depend variable bond to XDMFFile at its construction time.
This PR removes all XDMFFile's methods encoding argument.

I can possibly make encoding non-const and allow lifetime changes.

UPDATE:
**HDF5Interface.cpp**:
- File mode "a" now in C-style, i.e. when addressed file is not existing create a new empty one.
Reason: Unify file-modes for XDMFFile and HDF5File to C lang meaning.

**XDMFFile.cpp**:
Generally, make XDMFFile more state-dependent.
- Add file modes {`a`,`r`, `w`, `ab`, `wb`} to XDMFFile. With the meaning defined here https://en.cppreference.com/w/cpp/io/c/fopen . Second letter "b" maps to HDF5 encoding. Internal encoding variable not strictly needed, possibly remove.
- Remove internal `_counter` variable. 
Reason: This variable was a pain, it was incremented to keep unique numbers of multiple entities written to XDMF, but restored to 0 when new XMDFFile is constructed. Thus re-opening and appending file with meshfunction or VTK-style timeseries was buggy. In addition there was only one counter for all entities written to a file...
-  Use `_hdf5_file` more. For `wb` and `ab` this pointer is set to opened/created HDF file at XDMFFile construction time - in contrast to doing this at write time. 
Reason: Make XDMFFile more state dependent. No need to open HDF file and check existence and validity etc. in every write/read method. 
- Make new `_h5_id` internal variable. 
Reason: Lifetime of XDMFFile is now more bound to HDF5File (if binary encoding specified).
- XDMF XML file is also prepared to desired state (empty <Xdmf> and <Domain> nodes) or read at XDMFFile construction time. 
Reason: Make XDMFFile more state dependent. Simplifies handling of XML file. No need to check validity of XML structure in every write/read method.
- Make `add_points` append `<Domain>` node instead of `<Xdmf>` node. 
Reason: Every other internal appending methods does this, Xdmf node is too high level.
- Add check for same function name in parallel to `write` with time
Reason: Since internal counter was removed this is essential for correct HDF path. Unfortunately, some methods accept name as string and some take `Variable::name`. This should be unified.


garth-wells:
I thought about this a while back, and my reaction was that it might be nice to be able to mix ASCII and HDF5.

michalhabera:
What do you mean? Make encoding non-const or discard the idea of whole PR?

garth-wells:
I mean decide when writing how the output should be encoded, rather than when opening the file. It would be nice to have a discussion of the pros and cons.

michalhabera:
This can be accomplished making `encoding` private member variable and having getter and setter. Setter would during lifetime check if the new encoding is supported (i.e. you have HDF, ...)

Code could be
```python
with XDMFFile(MPI.comm_world, "outfile.xdmf", XDMFFile.Encoding.ASCII) as outfile:
    outfile.write_checkpoint(u, "u_ascii")
    outfile.encoding = XDMFFile.Encoding.HDF5
    outfile.write_checkpoint(u, "u_hdf")
```
Not possible with `XDMFFile::write` which always overwrites (so physical .xdmf would be discarded).

In principle, you either have or don't have HDF library.
1.) If you don't have HDF then no point to change encoding - would throw error anyway.
2.) If you have HDF you use ASCII output for some debugging purposes, you could be interested only in case that you have huge file with many functions stored and want to see just DOFs of one particular function for debugging. But for debugging it's probably better to create another ASCII XDMFFile.

chrisrichardson:
The real separation should be between read and write modes. Opening a file for "read" should not need to specify an encoding. It would be better if XDMFFile specified "read" or "write" when opening, then the encoding could be an option (for write mode only).

garth-wells:
@chrisrichardson good point - we should really discuss this and try to get it right.

michalhabera:
@chrisrichardson @garth-wells Well, I agree with read/write modes separation. That would make it very clear for users - now they don't know what is overwriting, appending, etc. 

But then I'd suggest to follow C-style IO flags and have modes `r[b, +], w[b, +], a[b, +]`. The "encoding" is there specified together with IO mode. Or if you don't want to have it all merged in one string make encoding another argument - but that I did in this PR.

chrisrichardson:
I agree with the idea of C-style flags, except that `rb` and `r+` do not make much sense here. Modes could be `w`, `wb`, `a`, `ab` and `r`.

michalhabera:
Good. There are modes `a, r, w` in `HDF5File`, so the idea maps nicely. 

You are right that `+` modes that allow both reading and writing doesn't make much sense in typical fenics code-flow.

michalhabera:
Work in progress.

michalhabera:
Updated PR description. Work still in progress. There are many linked problems, so PR is bloating...

cpp/dolfin/io/HDF5Interface.cpp
```diff
@@ -34,37 +33,45 @@ hid_t HDF5Interface::open_file(MPI_Comm mpi_comm, const std::string filename,
...
+    if (file_exists)
     {
-      // Open file existing file for append
       file_id = H5Fopen(filename.c_str(), H5F_ACC_RDWR, plist_id);
     }
-    else if (mode == "r")
-      file_id = H5Fopen(filename.c_str(), H5F_ACC_RDONLY, plist_id);
     else
     {
       log::dolfin_error("HDF5Interface.cpp", "open HDF5 file",
```
garth-wells: Could you change this to `throw std::runtime_error("...")`? (Shame we can't use `std::filesystem` yet . . . )

cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -40,10 +40,47 @@ using namespace dolfin;
...
+  if (!(_file_mode == "a" or _file_mode == "r" or _file_mode == "w"
+        or _file_mode == "wb" or _file_mode == "ab"))
+  {
+    throw std::runtime_error("Unknown file mode used in XDMFFile.");
+  }
+
+  // FIXME: Encoding serves just as proxy here, not really needed to keep it
+  // For binary modes set HDF5 encoding
+  if (_file_mode == "wb" or _file_mode == "ab")
+  {
```
garth-wells: Remove braces - we tend to use one line for short control statements and bodies.

cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -56,6 +93,73 @@ XDMFFile::XDMFFile(MPI_Comm comm, const std::string filename)
...
+      // Prepare new XML structure
+      pugi::xml_node xdmf_node = _xml_doc->append_child("Xdmf");
+      assert(xdmf_node);
+      xdmf_node.append_attribute("Version") = "3.0";
+
+      pugi::xml_node domain_node = xdmf_node.append_child("Domain");
+      assert(domain_node);
+    }
+  }
+  // XML file doesn't exist we need to prepare some
```
garth-wells: Can you put this inside the braces?

michalhabera:
Tests are green. @garth-wells @chrisrichardson @blechta Please comment if you want see some changes. I am ok with it, but there are more changes to come that I'll propose in other PRs.

.circleci/config.yml
```diff
@@ -84,7 +84,7 @@ jobs:
           name: Run Python unit tests (MPI)
           command: |
             cd python/test/unit
-            mpirun --allow-run-as-root -n 3 python3 -m pytest .
+            mpirun --allow-run-as-root -n 3 python3 -m pytest . -vsx
```
garth-wells: What does `-vsx` do?

cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -56,6 +85,73 @@ XDMFFile::XDMFFile(MPI_Comm comm, const std::string filename)
   // Flush datasets to disk at each timestep. Allows inspection of the
   // HDF5 file whilst running, at some performance cost.
   parameters.add("flush_output", false);
+
+  // Synchronize before filesystem check
+  _mpi_comm.barrier();
```
garth-wells: Why a barrier?

cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -56,6 +85,73 @@ XDMFFile::XDMFFile(MPI_Comm comm, const std::string filename)
...
+  bool exists = boost::filesystem::exists(_filename);
+  // If XDMF XML file exists physically
+  if (exists)
+  {
+    // Load its XML structure into internal XML file
+    pugi::xml_parse_result result = _xml_doc->load_file(_filename.c_str());
```
garth-wells: Should we read the XML on one process only? I can't imagine filesystems liking this with 10k+ processes.

cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -68,76 +164,34 @@ void XDMFFile::close()
...
-
-  // Add XDMF node and version attribute
-  _xml_doc->append_child(pugi::node_doctype)
-      .set_value("Xdmf SYSTEM \"Xdmf.dtd\" []");
-  pugi::xml_node xdmf_node = _xml_doc->append_child("Xdmf");
-  assert(xdmf_node);
-  xdmf_node.append_attribute("Version") = "3.0";
-  xdmf_node.append_attribute("xmlns:xi") = "http://www.w3.org/2001/XInclude";
+  if (_file_mode[0] == 'a')
+    // TODO: Implement mesh appending
```
garth-wells: Remove comment.

cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -256,7 +238,7 @@ void XDMFFile::write_checkpoint(const function::Function& u,
       = function_name + "_" + std::to_string(counter);
 
   const mesh::Mesh& mesh = *u.function_space()->mesh();
-  add_mesh(_mpi_comm.comm(), func_temporal_grid_node, h5_id, mesh,
+  add_mesh(_mpi_comm.comm(), func_temporal_grid_node, _h5_id, mesh,
```
garth-wells: This function name is unclear to me. A better name would be good.

cpp/dolfin/io/XDMFFile.h
```diff
@@ -97,18 +90,32 @@ class XDMFFile : public common::Variable
...
-#endif
-
-  /// Constructor
-  XDMFFile(const std::string filename) : XDMFFile(MPI_COMM_WORLD, filename) {}
+  /// Overloaded XDMFFile constructor
+  ///
+  /// Constructs XDMFFile with MPI_COMM_WORLD
+  XDMFFile(const std::string filename, const std::string file_mode)
+      : XDMFFile(MPI_COMM_WORLD, filename, file_mode)
+  {
```
garth-wells: Move braces up, e.g. `, file_mode) {}`

cpp/dolfin/io/XDMFFile.h
```diff
@@ -576,6 +540,9 @@ class XDMFFile : public common::Variable
   // MPI communicator
   dolfin::MPI::Comm _mpi_comm;
 
+  // Integer HDF file identificator
```
garth-wells: `// HDF5 file handle`?

python/demo/undocumented/poisson-disc/demo_poisson-disc.py
```diff
@@ -70,8 +70,8 @@ def compute_rates():
     for coordinate_degree in (1, 2):
         for element_degree in (1, 2):
             print("\nUsing coordinate degree %d, element degree %d" % (coordinate_degree, element_degree))
-            ufile = XDMFFile(MPI.comm_world, "poisson-disc-degree-x%d-e%d.xdmf" % (coordinate_degree, element_degree))
-            encoding = XDMFFile.Encoding.HDF5 if has_hdf5() else XDMFFile.Encoding.ASCII
+            encoding = "b" if has_hdf5() else ""
+            ufile = XDMFFile(MPI.comm_world, "poisson-disc-degree-x%d-e%d.xdmf" % (coordinate_degree, element_degree), "w" + encoding)
```
garth-wells: Using the Python 3 `format` syntax is nicer and preferred in dolfinx.

python/src/io.cpp
```diff
@@ -212,12 +212,17 @@ void io(py::module& m)
...
-                                                           filename);
+      .def(py::init([](const MPICommWrapper comm, std::string filename,
+                       const std::string file_mode) {
+             return std::make_unique<dolfin::io::XDMFFile>(comm.get(), filename,
+                                                           file_mode);
            }),
-           py::arg("comm"), py::arg("filename"))
-      .def(py::init<std::string>())
+           py::arg("comm"), py::arg("filename"), py::arg("file_mode"))
+      .def(
```
garth-wells: Remove this constructor. Can add a helper in the Python layer later if we want.

cpp/dolfin/io/XDMFFile.cpp
michalhabera: Since only root process saves XML file, it could happen, that root is slow and didn't save the XML file of previous write method, while other faster process continues and possibly constructs file in "r" mode but apparently can't find the XML file so prepares new empty structure. This process then has no information on where in HDF file (hdf path) the object to read could be.

cpp/dolfin/io/XDMFFile.cpp
michalhabera: That makes more sense, I agree. But this is the past design. 
I agree, that the only process working with XML file on filesystem level should be root process, while information needed to read from HDF file like hdf filepath of dataset must be communicated through MPI. I'd rather make this change in separate PR.

cpp/dolfin/io/XDMFFile.h
michalhabera: True handle is the `_hd5_file` pointer. This one is -1 for ASCII encoding or no HDF support. But I'd like to remove this `_h5_id` variable. Its redundant once we keep the `_hdf5_file` also. It introduces now unmanaged depencency - when one changes `_hdf5_file` no change to `_h5_id` is made...

Having just `_hdf5_file` is nice, since it gives access to HDF5File from XDMFFile, but I need to add some interface for it. This will allow together with https://github.com/FEniCS/dolfinx/issues/8 more custom control of XDMFFile.

garth-wells: My comment is on the comment. `// HDF5 file handle` would be a better comment.

michalhabera:
@garth-wells I wrapped IO classes in commit https://github.com/FEniCS/dolfinx/pull/122/commits/f4aef3d86c0caa8568bfda3eb7b411cbb1988e59
That should also fix https://github.com/FEniCS/dolfinx/issues/124. 

I hope I got it right. Is this the way we're aiming towards?

python/dolfin/io/xdmffile.py
```diff
@@ -0,0 +1,43 @@
...
+        Note
+        ----
+        Parameter `V: Function space` must be the same as saved function space
+        except for ordering of mesh entities.
+
+        Returns
+        -------
+        dolfin.Function
+        """
+        if not isinstance(V, FunctionSpace):
```
garth-wells: Could we allow `cpp.function.FunctionSpace` objects?

python/src/io.cpp
```diff
@@ -232,140 +232,99 @@ void io(py::module& m)
...
            },
-           py::arg("u"), py::arg("function_name"), py::arg("time_step") = 0.0,
-           py::arg("encoding") = dolfin::io::XDMFFile::Encoding::HDF5)
+           py::arg("u"), py::arg("function_name"), py::arg("time_step") = 0.0)
       .def("write_checkpoint",
            [](dolfin::io::XDMFFile& instance, const py::object u,
-              std::string function_name, double time_step,
-              dolfin::io::XDMFFile::Encoding encoding) {
+              std::string function_name, double time_step) {
              auto _u
```
garth-wells: It would be good to remove all the unwrapping in the pybind11 layer and handle it from the Python interface.

python/dolfin/io/xdmffile.py
michalhabera: Is there a strong need for that once rule "no `cpp.*` object being public" is applied?
I am afraid of mixing `cpp.*` (backend) with other objects. This will cause (and already caused in 2018.1) many confusion within users.

garth-wells: We don't want any `cpp` objects being available without an explicit `dolfin.cpp` import. This doesn't mean a function can't accept a `cpp` object if, especially if handling it at the Python layer is tidier than handling it at the pybind11 layer.

PR #131, opened 2018-07-18
Rename dolfin.has_foo() to dolfin.config.has_foo (property)

blechta:


python/dolfin/__init__.py
```diff
@@ -31,13 +31,12 @@
...
 
-from .cpp.common import (Variable, has_debug, has_hdf5, has_scotch,
-                         has_hdf5_parallel, has_mpi4py, has_petsc_complex,
-                         has_petsc4py, has_parmetis, has_slepc, has_slepc4py,
-                         git_commit_hash, DOLFIN_EPS, DOLFIN_PI, TimingType,
-                         timing, timings, list_timings)
+from .cpp.common import Variable, TimingType, timing, timings, list_timings
+from .cpp.common import config
+DOLFIN_EPS = config.DOLFIN_EPS  # FIXME: Remove me
+DOLFIN_PI = config.DOLFIN_PI  # FIXME: Remove me
```
blechta: These variables need removing (because `dolfin.config.DOLFIN_FOO` is read-only but `dolfin.DOLFIN_FOO` is not any more). Will appear in another PR.

python/demo/undocumented/curl-curl/demo_curl-curl.py
```diff
@@ -37,7 +37,7 @@
     print("This demo requires PETSc to be configured with HYPRE.")
     exit()
 
-if not has_petsc4py():
+if not config.has_petsc4py:
```
IgorBaratta: Isn't petsc4py a required dependency too?

blechta: I don't think so. See `python/CMakeLists.txt`.

python/dolfin/__init__.py
blechta: Solved in #133

blechta:
pybind11 does not allow to have read-only properties in a module (because of Python architecture I believe). So `dolfin.config` is not a module but it is a class.

It would be possible to have writable (!) variable in `dolfin` namespace.

python/demo/undocumented/curl-curl/demo_curl-curl.py
garth-wells: petsc4py isn't a required dependency. I would be reluctant to make it requires because it's not always in sync with PETSc master.

IgorBaratta: I asked this because [`python/doc/source/installation.rst`](https://github.com/FEniCS/dolfinx/blob/896e36c46ac12d91b696d8f336dbddb17fc84b29/python/doc/source/installation.rst) states that petsc4py is a required dependency for the python interface.

garth-wells: @jhale might have added that to the docs.

blechta:
@garth-wells We have to decide whether to use `dolfin.config` namespace (it is a non-constructible class in fact). Without that, we can't move on. Read-only module attributes do not exist as far as I know, so read-only `dolfin.complex_mode` is not  possible; standard way is a function `dolfin.mode_is_complex()`.

blechta:
Consensus has not been reached.

PR #132, opened 2018-07-18
Rename config.has_petsc_complex to config.complex_mode 

blechta:
Based on #131 

Rename `config.has_petsc_complex` to `config.complex_mode`

Also rename `dolfin::has_petsc_complex()` to `dolfin::has_complex_mode()`.
Keeping `has_` because there is not an appropriate namespace.

### DIFF
--- a/python/test/unit/fem/test_complex_assembler.py
+++ b/python/test/unit/fem/test_complex_assembler.py
@@ -11,10 +11,6 @@ import numpy as np
 import pytest
 from ufl import dx, grad, inner
 
-# FIXME: Move to dolfin_utils.test.skips
-pytestmark = pytest.mark.skipif(not dolfin.config.has_petsc_complex,
-                                reason="Only works in complex mode.")
-
 
 def test_complex_assembly():
     """Test assembly of complex matrices and vectors"""
--- a/python/test/unit/fem/test_complex_assembler.py
+++ b/python/test/unit/fem/test_complex_assembler.py
@@ -9,8 +9,10 @@ import ufl
 import dolfin
 import numpy as np
 from ufl import dx, grad, inner
+from dolfin_utils.test.skips import skip_if_not_complex
 
 
+@skip_if_not_complex
 def test_complex_assembly():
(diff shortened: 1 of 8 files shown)


blechta:
Yes, `dolfin.config.scalar_[d]type` can be added in another PR. Let's decide about this first.

python/test/unit/fem/test_complex_assembler.py
```diff
@@ -8,14 +8,11 @@
 import ufl
 import dolfin
 import numpy as np
-import pytest
 from ufl import dx, grad, inner
-
-# FIXME: Move to dolfin_utils.test.skips
-pytestmark = pytest.mark.skipif(not dolfin.config.has_petsc_complex,
-                                reason="Only works in complex mode.")
+from dolfin_utils.test.skips import skip_if_not_complex
```
IgorBaratta: What about  `pytestmark = skip_if_not_complex` at the beginning after the imports instead a decorator for each function.
I think it would be clearer that all tests should be skipped if we are not in complex mode.

IgorBaratta: Ok, in the current case it does not make much difference because we have only two tests.

blechta: Honestly I originally removed it (because it was a variable not used and broke the tests) because I didn't realize that the import has a side effect. I find it weird, but I am fine with doing it your way.

IgorBaratta:
> has_complex_mode does sound right semantically. mode_is_complex would make more sense, but happy to hear other suggestions.

I would stick with `dolfin::has_petsc_complex()` (and `dolfin.config.complex_mode`) to maintain the `has_` prefix.  
If it's not needed, `mode_is_complex()` looks fine to me. Or `is_complex_mode()` to resemble a question.

blechta:
Oh yeah, `has_complex_mode` is obviously incorrect name. Sorry :laughing:

blechta:
Depends on closed #131.

PR #133, opened 2018-07-18
Remove DOLFIN_EPS and DOLFIN_PI from Python dolfin namespace

blechta:
Based on #131 

Move `dolfin.DOLFIN_EPS` and `dolfin.DOLFIN_PI` to `dolfin.config.DOLFIN_EPS` and `dolfin.config.DOLFIN_PI` respectively.


blechta:
@garth-wells I agree.

garth-wells:
@blechta removing is an easy fix then!

PR #134, opened 2018-07-18
Add Complex support to XDMFFile

IgorBaratta:
Add support to:
 - Write complex functions 
        - `XDMFFile::write(const function::Function& u, const Encoding encoding)`
        - `XDMFFile::write(const function::Function& u, double time_step, const Encoding encoding)`

Different `Attribute` (`AttributeType = {Scalar, Vector, Tensor}`) are created for each component (real and imaginary). The prefixes `"real"` and `"imag"` are attached to the function name to identify the component. 

Excerpt from an output file :
```        
<Attribute Name="real_field" AttributeType="Scalar" Center="Node">
      <DataItem Dimensions="36 1" Format="HDF">output.h5:/VisualisationVector/real/0</DataItem>
</Attribute>
<Attribute Name="imag_field" AttributeType="Scalar" Center="Node">
      <DataItem Dimensions="36 1" Format="HDF">output.h5:/VisualisationVector/imag/0</DataItem>
</Attribute>
```

 - complex checkpointing (write and read): 
Three options have been tested for complex checkpointing:
1 - ~Add a `DataItem` for each component  (avoids additional copies but it is not compatible with Paraview)~
2 - Add a `FiniteElementFunction` for each component, similar to applying `add_function` twice
3 - ~Add a `Temporal` collection for each component, similar to calling `write_checkpoint` twice (Compatible with Paraview visualization; however data storage is doubled).~

**Edited:**
After some discussions,  option 2 was considered more convenient, 
All test are now green on both complex and real modes.

### DIFF
--- a/cpp/dolfin/io/XDMFFile.cpp
+++ b/cpp/dolfin/io/XDMFFile.cpp
@@ -1629,9 +1629,9 @@ XDMFFile::read_checkpoint(std::shared_ptr<const function::FunctionSpace> V,
                           + func_name + "']/Grid[" + selector + "]")
                              .c_str())
             .node();
-
   assert(grid_node);
 
+#ifdef PETSC_USE_COMPLEX
   pugi::xml_node fe_attribute_node
       = grid_node
             .select_node(("Attribute[@ItemType=\"FiniteElementFunction\" and"
@@ -1639,6 +1639,12 @@ XDMFFile::read_checkpoint(std::shared_ptr<const function::FunctionSpace> V,
                           + func_name + "']")
                              .c_str())
             .node();
+#else
+  pugi::xml_node fe_attribute_node
+      = grid_node.select_node("Attribute[@ItemType=\"FiniteElementFunction\"]")
+            .node();
+#endif
+
   assert(fe_attribute_node);
 
   // Get cells dofs indices = dofmap
--- a/cpp/dolfin/io/XDMFFile.cpp
+++ b/cpp/dolfin/io/XDMFFile.cpp
@@ -1639,24 +1639,13 @@ XDMFFile::read_checkpoint(std::shared_ptr<const function::FunctionSpace> V,
                           + func_name + "']")
                              .c_str())
             .node();
-  assert(fe_attribute_node);
-
-  // Find FE attribute node of the imaginary component
-  pugi::xml_node imag_fe_attribute_node
-      = grid_node
-            .select_node(("Attribute[@ItemType=\"FiniteElementFunction\" and"
-                          "@Name='imag_"
-                          + func_name + "']")
-                             .c_str())
-            .node();
-  assert(imag_fe_attribute_node);
 #else
   pugi::xml_node fe_attribute_node
       = grid_node.select_node("Attribute[@ItemType=\"FiniteElementFunction\"]")
             .node();
-  assert(fe_attribute_node);
 #endif
 
+  assert(fe_attribute_node);
   // Get cells dofs indices = dofmap
   pugi::xml_node cell_dofs_dataitem
       = fe_attribute_node.select_node("DataItem[position()=1]").node();
@@ -1720,6 +1709,16 @@ XDMFFile::read_checkpoint(std::shared_ptr<const function::FunctionSpace> V,
   // Read real component of function vector
   std::vector<double> real_vector = get_dataset<double>(
       _mpi_comm.comm(), vector_dataitem, parent_path, input_vector_range);
+
+  // Find FE attribute node of the imaginary component
+  pugi::xml_node imag_fe_attribute_node
+      = grid_node
+            .select_node(("Attribute[@ItemType=\"FiniteElementFunction\" and"
+                          "@Name='imag_"
+                          + func_name + "']")
+                             .c_str())
+            .node();
+  assert(imag_fe_attribute_node);
 
   // Get extra FE attribute of the imaginary component
   pugi::xml_node imag_vector_dataitem
--- a/cpp/dolfin/io/XDMFFile.cpp
+++ b/cpp/dolfin/io/XDMFFile.cpp
@@ -396,15 +396,15 @@ void XDMFFile::write(const function::Function& u, const Encoding encoding)
       = cell_centred ? "Cell" : "Node";
 
   // Add imaginary attribute node
-  pugi::xml_node imaginary_attribute_node = grid_node.append_child("Attribute");
-  assert(imaginary_attribute_node);
-  imaginary_attribute_node.append_attribute("Name") = imag_attr_name.c_str();
-  imaginary_attribute_node.append_attribute("AttributeType")
+  pugi::xml_node imag_attribute_node = grid_node.append_child("Attribute");
+  assert(imag_attribute_node);
+  imag_attribute_node.append_attribute("Name") = imag_attr_name.c_str();
+  imag_attribute_node.append_attribute("AttributeType")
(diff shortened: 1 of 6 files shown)


cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -531,14 +563,53 @@ void XDMFFile::write(const function::Function& u, double time_step,
...
+    const std::string dataset_name
+        = "/VisualisationVector/" + component + "/" + std::to_string(_counter);
+
+    std::vector<double> component_data_values(data_values.size());
+    for (unsigned int i = 0; i < data_values.size(); i++)
+    {
+      if (component == components[0])
+        component_data_values[i] = data_values[i].real();
+      else if (component == components[1])
+        component_data_values[i] = data_values[i].imag();
```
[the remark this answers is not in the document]
IgorBaratta: I was not aware of that. I'm gonna check it out.

cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -13,6 +13,7 @@
 #include <boost/filesystem.hpp>
 #include <boost/format.hpp>
 #include <boost/lexical_cast.hpp>
+#include <complex>
```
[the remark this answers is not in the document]
IgorBaratta: Removed! 
It was not really necessary.

cpp/dolfin/io/XDMFFile.cpp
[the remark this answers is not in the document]
IgorBaratta: I'm working on it. 
If it is OK, I would rather include this in a new PR,  since most of modifications are not in XDMFFile.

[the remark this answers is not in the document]
blechta: Yes, please do that in a separate PR.

cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -386,8 +391,47 @@ void XDMFFile::write(const function::Function& u)
...
+  }
+
+#else
+  // Add attribute node
+  pugi::xml_node attribute_node = grid_node.append_child("Attribute");
+  assert(attribute_node);
+  attribute_node.append_attribute("Name") = u.name().c_str();
+  attribute_node.append_attribute("AttributeType")
+      = rank_to_string(u.value_rank()).c_str();
+  attribute_node.append_attribute("Center") = cell_centred ? "Cell" : "Node";
```
blechta: This code block seems the same as above in the component-for-loop. Would it makes sense to factor it out, maybe into a small private function?

blechta: Or maybe it could work like this:
```
#ifdef PETSC_USE_COMPLEX
  std::vector<std::string> components = {"_real", "_imag"};
#else
  std::vector<std::string> components = {""};
#endif
for (const std::string component : components)
{
  // Common code
}
```

cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -529,14 +573,53 @@ void XDMFFile::write(const function::Function& u, double time_step)
...
+      else if (component == components[1])
+        component_data_values[i] = data_values[i].imag();
+    }
+
+    // Add data item of component
+    add_data_item(_mpi_comm.comm(), attribute_node, h5_id, dataset_name,
+                  component_data_values, {num_values, width});
+  }
+
+#else
```
blechta: The same consideration as above - is sharing more code with the non-complex case possible? It is probably beneficial to leave this out for the `H5Sselect_hyperslab` issue though.

cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -1341,11 +1427,28 @@ void XDMFFile::add_function(MPI_Comm mpi_comm, pugi::xml_node& xml_node,
 
   // Get all local data
   const la::PETScVector& u_vector = *u.vector();
-  std::vector<double> local_data;
+  std::vector<PetscScalar> local_data;
   u_vector.get_local(local_data);
 
+#ifdef PETSC_USE_COMPLEX
+  // FIXME: Avoid copies by writing directly a compound data
```
blechta: Note: With fixing this, this can probably also can be done with
```
#ifdef PETSC_USE_COMPLEX
  std::vector<std::string> components { ... }
#else
  std::vector<std::string> components { ... }
#endif.
```

cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -1529,13 +1632,34 @@ XDMFFile::read_checkpoint(std::shared_ptr<const function::FunctionSpace> V,
...
+  // Find FE attribute node of the imaginary component
+  pugi::xml_node imag_fe_attribute_node
+      = grid_node
+            .select_node(("Attribute[@ItemType=\"FiniteElementFunction\" and"
+                          "@Name='imag_"
+                          + func_name + "']")
+                             .c_str())
+            .node();
+  assert(imag_fe_attribute_node);
+#else
```
blechta: AKA
```
#ifdef PETSC_USE_COMPLEX
  std::vector<std::string> components { "_real", "_imag" };
#else
  std::vector<std::string> components { "" };
#endif

std::vector<pugi::xml_node> nodes(components.size());
for (std::size_t i = 0; i < components.size(); ++i)
{
  // Common code
}
```

cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -1596,9 +1720,33 @@ XDMFFile::read_checkpoint(std::shared_ptr<const function::FunctionSpace> V,
...
+      _mpi_comm.comm(), imag_vector_dataitem, parent_path, input_vector_range);
+
+  assert(real_vector.size() == imag_vector.size());
+
+  // Compose complex function vector
+  std::vector<PetscScalar> vector;
+  vector.reserve(real_vector.size());
+  std::transform(begin(real_vector), end(real_vector), begin(imag_vector),
+                 std::back_inserter(vector),
+                 [](double r, double i) { return r + i * PETSC_i; });
```
blechta: The memory footprint of this is rather large. It seems we can again, as in the Hyperslab issue above, tell HDF5 to write the data directly into strided memory. See https://support.hdfgroup.org/HDF5/Tutor/selectsimple.html#def. Can be done in the next PR.

cpp/dolfin/io/XDMFFile.cpp
IgorBaratta: This whole `#ifdef PETSC_USE_COMPLEX ...  #endif` can go away with the correct use of H5Sselect_hyperslab. We have to specify the `stride=2`, and `offset = 0 or 1` depending on the component (real and imaginary). 
Just have figure out how this information should reach `HDF5Interface::write_dataset`.

cpp/dolfin/io/XDMFFile.cpp
IgorBaratta: Yes, reusing makes code look cleaner.
I selected the second option for all cases that it can apply.

IgorBaratta:
This blocks with unnecessary copies will be removed  in the next PR.
```
std::vector<double> component_data_values(data_values.size());
for (unsigned int i = 0; i < data_values.size(); i++)
{
    if (component == components[0])
        component_data_values[i] = data_values[i].real();
    else if (component == components[1])
       component_data_values[i] = data_values[i].imag();
}
```

The idea is to make `HDF5Interface::write_dataset` aware that the dataset is compound and set a stride to be used in `H5Sselect_hyperslab`.

PR #135, opened 2018-07-18
Make HDF append mode C-style

michalhabera:
When addressed file is not existing create new empty one.

### DIFF
--- a/cpp/dolfin/io/HDF5Interface.cpp
+++ b/cpp/dolfin/io/HDF5Interface.cpp
@@ -28,7 +28,7 @@ hid_t HDF5Interface::open_file(MPI_Comm mpi_comm, const std::string filename,
     MPI_Info info;
     MPI_Info_create(&info);
     herr_t status = H5Pset_fapl_mpio(plist_id, mpi_comm, info);
-    assert(status > 0);
+    assert(status >= 0);
     MPI_Info_free(&info);
   }
 #endif
--- a/cpp/dolfin/io/HDF5Interface.cpp
+++ b/cpp/dolfin/io/HDF5Interface.cpp
@@ -10,7 +10,6 @@
 #include <dolfin/common/MPI.h>
 #include <dolfin/log/log.h>
 
-
 #define HDF5_FAIL -1
 #define HDF5_MAXSTRLEN 80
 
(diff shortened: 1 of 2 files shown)


cpp/dolfin/io/HDF5Interface.cpp
```diff
@@ -35,34 +34,44 @@ hid_t HDF5Interface::open_file(MPI_Comm mpi_comm, const std::string filename,
...
-  else
+  else if (mode == "a")
   {
-    // Check that file exists
-    if (!boost::filesystem::is_regular_file(filename))
+    if (file_exists)
     {
-      log::dolfin_error("HDF5Interface.cpp", "open HDF5 file",
-                   "File \"%s\" does not exist", filename.c_str());
+      file_id = H5Fopen(filename.c_str(), H5F_ACC_RDWR, plist_id);
```
garth-wells: Add below `H5Fopen`:
```cpp
if (file_id < 0)
  throw std::runtime_error("......");

File: cpp/dolfin/io/HDF5Interface.cpp [hunk 2]
```diff
@@ -35,34 +34,44 @@ hid_t HDF5Interface::open_file(MPI_Comm mpi_comm, const std::string filename,
...
     }
-
-    if (mode == "a")
+    else
+    {
+      // If file doesn't exists but we want to append it
+      // create new
+      file_id
+          = H5Fcreate(filename.c_str(), H5F_ACC_TRUNC, H5P_DEFAULT, plist_id);
```
garth-wells: Add below `H5Fopen`:
```cpp
if (file_id < 0)
  throw std::runtime_error("......");

File: cpp/dolfin/io/HDF5Interface.cpp [hunk 3]
```diff
@@ -35,34 +34,44 @@ hid_t HDF5Interface::open_file(MPI_Comm mpi_comm, const std::string filename,
...
+    }
+  }
+  else if (mode == "r")
+  {
+    if (file_exists)
     {
-      // Open file existing file for append
       file_id = H5Fopen(filename.c_str(), H5F_ACC_RDWR, plist_id);
```
garth-wells: Add below `H5Fopen`:
```cpp
if (file_id < 0)
  throw std::runtime_error("......");

File: cpp/dolfin/io/HDF5Interface.cpp [hunk 4]
```diff
@@ -35,34 +34,44 @@ hid_t HDF5Interface::open_file(MPI_Comm mpi_comm, const std::string filename,
   }
 #endif
 
+  // Query file existence only once
+  bool file_exists = boost::filesystem::exists(filename);
```
garth-wells: Is this needed? Is it in the right place? It's not relevant for wrote mode, and in other cases will the HDF5 error code pick up the issue?

cpp/dolfin/io/HDF5Interface.cpp
```diff
@@ -421,7 +430,7 @@ void HDF5Interface::set_mpi_atomicity(const hid_t hdf5_file_handle,
   herr_t status = H5Fset_mpi_atomicity(hdf5_file_handle, atomic);
   if (status == HDF5_FAIL)
```
garth-wells: `if (status < 0)`

michalhabera: I am confused with this request. `HDF5_FAIL` is defined in header, there are many uses throughout this file, why change just this one?

garth-wells: The HDF5 docs simply say that the return code is negative in the case of any error, e.g. https://support.hdfgroup.org/HDF5/doc/RM/RM_H5F.html#File-SetMpiAtomicity, the so the check should respect this.

cpp/dolfin/io/HDF5Interface.cpp
michalhabera: @garth-wells It is needed if we do not want to change HDF error handling in this PR. I tried to move responsibility for filesystem checks to HDF, but it prints error to stack, even if we just probe to open, otherwise create.

I can 
1.) Turn off error stack printing globally. Then handle error stack for every failure check.
2.) Temporarily turn stack off, then turn on after this probe. (But I am having difficulties finding default stack handler).

garth-wells: We can have the check, but it's not needed in write mode (or append mode, if the fallback it to create a new file, which may or may not be a good thing).

PR #136, opened 2018-07-19
Wrap cpp.io in python

michalhabera:
Idea of this PR is to separate `io` namespace and try new "nothing in cpp namespace be public" approach.
- Removed import of IO to global namespace in `__init__.py`.
- Wrapped `VTKFile`, `XDMFFile`, `HDF5File` in `io.py`

### DIFF
--- a/python/dolfin/io.py
+++ b/python/dolfin/io.py
@@ -1,32 +1,70 @@
 # -*- coding: utf-8 -*-
-# Copyright (C) 2017 Chris N. Richardson and Garth N. Wells
+# Copyright (C) 2017 Chris N. Richardson, Garth N. Wells, Michal Habera
 #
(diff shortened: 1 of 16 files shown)


python/dolfin/io.py
```diff
@@ -1,32 +1,70 @@
...
+        Note
+        ----
+        Parameter `V: Function space` must be the same as saved function space
+        except for ordering of mesh entities.
+        Returns
+        -------
+        dolfin.Function
+        """
 
+        # Try to extract cpp object
```
garth-wells: Comment can probably be removed.

python/dolfin/io.py
```diff
@@ -1,32 +1,70 @@
...
+        V_cpp = getattr(V, "_cpp_object", V)
```
garth-wells: Close up space if comment is removed.

PR #138, opened 2018-07-20
Start with hiding of dolfin.cpp (for IO) and improved doc generation

garth-wells:
This is a start on designing a way to make the `dolfin.cpp` interface private and to generate much improved documentation for the Python interface. Only `dolfin.io` is addressed at this stage.

- Making the `dolfin.cpp` interface private (still accessible, but not part of the 'public' interface) gives a nice separation between the C++ and Python interfaces, and means that a change in the C++ interface does not need to lead to a consequent change in the Python interface (which will be a very good thing).
- Declaring a 'public' Python interface gives us something we can aim to keep stable and manage changes/deprecations in an orderly fashion, and document well. Advanced users can access the `dolfin.cpp` interfaces, with the awareness that they can change.
- The unwrapping of objects, e.g. getting the `cpp.function.Function` from `dolfin.Function` to pass into the C++ code can be done easily in the Python layer, rather than the more complex present implementation in the pybind11 layer (which involves accessing Python attributes from C++).
- We don't  need special Python functions in the pybind11 layer, e.g. `__exit__` and `__exit__`, since these are implemented in the Python interface.

This isn't the final product for `dolfin.io`, but a basis for further design experimentation and improvement.

### DIFF
--- a/python/dolfin/io.py
+++ b/python/dolfin/io.py
@@ -38,15 +38,15 @@ class HDF5File:
         """Get atomicity of the HDF5 file"""
         return self._cpp_object.get_mpi_atomicity()
 
-    def set_mpi_atomicity(self, atomicity: bool):
+    def set_mpi_atomicity(self, atomicity: bool) -> None:
         """Set atomicity of the HDF5 file"""
         self._cpp_object.set_mpi_atomicity(atomicity)
 
-    def close(self):
+    def close(self) -> None:
         """Close file"""
         self._cpp_object.close()
 
-    def write(self, o, name, t=None):
+    def write(self, o, name, t=None) -> None:
         """Write object to file"""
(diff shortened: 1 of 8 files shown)


python/dolfin/io.py
```diff
@@ -1,19 +1,92 @@
...
+        """Write object to file"""
+        o_cpp = getattr(o, "_cpp_object", o)
+        if t is None:
+            self._cpp_object.write(o_cpp, name)
+        else:
+            self._cpp_object.write(o_cpp, name, t)
+
+    def read_mvc(self, mesh, name: str = ""):
+        # FIXME: figure type out from file (or pass string)  and return?
+        raise NotImplementedError("General MVC read function not implemented.")
```
chrisrichardson: Should be easy to determine the type of a dataset in HDF5 with a small utility function.

PR #139, opened 2018-07-20
Put encoding to XDMFFile constructor

michalhabera:
Adds private variable `_encoding` to XDMFFile. This variable is set by constructor and allows more state dependent calls to io methods.
Reason:
- simplify interface for io methods, no need to spoil encoding everywhere

### DIFF
--- a/cpp/dolfin/io/XDMFFile.h
+++ b/cpp/dolfin/io/XDMFFile.h
@@ -217,8 +217,7 @@ public:
   /// @param    encoding (_Encoding_)
   ///         Encoding to use: HDF5 or ASCII
   ///
-  void write(const mesh::MeshFunction<int>& meshfunction,
-             Encoding encoding = default_encoding);
+  void write(const mesh::MeshFunction<int>& meshfunction);
 
   /// Save mesh::MeshFunction to file using an associated HDF5 file, or
   /// storing the data inline as XML.
--- a/cpp/dolfin/io/XDMFFile.h
+++ b/cpp/dolfin/io/XDMFFile.h
@@ -105,10 +105,11 @@ public:
 #endif
 
   /// Constructor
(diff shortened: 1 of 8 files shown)


cpp/dolfin/io/XDMFFile.h
```diff
@@ -105,10 +105,11 @@ class XDMFFile : public common::Variable
 #endif
 
   /// Constructor
-  XDMFFile(const std::string filename) : XDMFFile(MPI_COMM_WORLD, filename) {}
+  XDMFFile(const std::string filename)
```
garth-wells: Let's remove this constructor from C++. Nice to have  just one way. We can always add a convenience version to Python

cpp/dolfin/io/XDMFFile.h
```diff
@@ -590,6 +577,8 @@ class XDMFFile : public common::Variable
   // The XML document currently representing the XDMF which needs to be
   // kept open for time series etc.
   std::unique_ptr<pugi::xml_document> _xml_doc;
+
+  Encoding _encoding;
```
garth-wells: Can make this `const`.

python/dolfin/io.py
```diff
@@ -118,7 +118,7 @@ class XDMFFile:
     # Import encoding (find better way?)
     Encoding = cpp.io.XDMFFile.Encoding
 
-    def __init__(self, mpi_comm, filename: str):
+    def __init__(self, mpi_comm, filename: str, encoding=Encoding.HDF5):
```
garth-wells: Having a default doesn't make sense when reading.

michalhabera: Agree. Later with file modes we can make overloaded constructor that takes file mode and encoding and throws on "r" mode. 
In python layer we can default encoding=None and dispatch inside __init__.py. Makes sense?

PR #141, opened 2018-07-27
Uptade XDMF tests to cover complex mode

IgorBaratta:
This PR aims to increase the coverage of the current xdmf tests to treat the complex case.
The tests remain the same in real mode. 
In the complex mode, a nonzero imaginary part is added to the functions.

### DIFF
--- a/python/test/unit/io/test_XDMF.py
+++ b/python/test/unit/io/test_XDMF.py
@@ -68,7 +68,7 @@ def worker_id(request):
 
 @pytest.fixture
 def imaginary_part():
-    """Return the imaginary unit in complex mode (complex scalar), 
+    """Return the imaginary unit in complex mode (complex scalar),
     in real mode return zero (float)
 
     """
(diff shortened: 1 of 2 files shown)


python/test/unit/io/test_XDMF.py
```diff
@@ -65,6 +66,18 @@ def worker_id(request):
         return 'master'
 
 
+@pytest.fixture
```
michalhabera: Why is this fixture? It is not used as fixture, just as a method.

python/test/unit/io/test_XDMF.py
```diff
@@ -165,7 +178,8 @@ def test_save_1d_scalar(tempdir, encoding):
     # FIXME: This randomly hangs in parallel
     V = FunctionSpace(mesh, "Lagrange", 2)
     u = Function(V)
-    u.vector()[:] = 1.0
+    A = imaginary_part()
```
michalhabera: Seems too long and cryptic to me. What about `u.vector()[:] = 1.0 + 1j * has_petsc_complex()` ?

IgorBaratta: `1.0 + 1j * has_petsc_complex()` returns a complex number with zero imaginary part `(1+0j)`.
However, in real mode a float is expected.

python/test/unit/io/test_XDMF.py
IgorBaratta: Indeed.. I forgot to remove the decorator.

python/test/unit/io/test_XDMF.py
IgorBaratta: It could also be replaced by:
```
A = 1j if has_petsc_complex() else 0.
u.vector()[:] = 1.0 + A
 ```
or something like.

Comment (michalhabera) [reply python/test/unit/io/test_XDMF.py]: Ok, I would replace `A` with `i`. Its not too important anyway, these are just tests.

Comment (blechta) [reply python/test/unit/io/test_XDMF.py]: Let's make it:
```
u.vector()[:] = 1.0 + (1j if has_petsc_complex() else 0)
```
This is perfectly readable on non-cryptic.

PR #143, opened 2018-07-28
Control Eigen alignment for compatibility with AVX user code

w1th0utnam3:
In order to use the underlying memory of Eigen matrices in functions that rely on AVX instructions, a Preprocessor directive has to be set (https://eigen.tuxfamily.org/dox/TopicPreprocessorDirectives.html).

The change in `cpp/CMakeLists.txt` ensures that the cpp library is compiled with the right Eigen configuration, the change in `cpp/cmake/templates/UseDOLFIN.cmake.in` ensures that any projects depending on Dolfin (including tests, demos and the python module) get compiled with the right settings.

**Update - Open points:**
 - Consider adding some kind of "test compilation" that obtains the real alignment used by Eigen as `-march` compiler flags from environment variables may override the preprocessor directive
 - The actual value should be written to a new pkg-config variable which can be used in the Python module to issue warnings/errors if the user tries to use incompatible alignments/vectorization modes

### DIFF
--- a/cpp/dolfin/CMakeLists.txt
+++ b/cpp/dolfin/CMakeLists.txt
@@ -86,10 +86,15 @@ endif()
 # UFC
 target_include_directories(dolfin SYSTEM PUBLIC ${UFC_INCLUDE_DIRS})
 
+# Currently only used for Eigen
+set(DOLFIN_ALIGNMENT_BYTES "32" CACHE STRING "\
+Minimum alignment in bytes used for numeric arrays. \
+For compatibility with AVX2, AVX-512... vectorized user code. \
+")
+
 # Eigen3
 target_include_directories(dolfin SYSTEM PUBLIC ${EIGEN3_INCLUDE_DIR})
-# Force Eigen to use 32-byte alignment for compatibility with AVX code
-target_compile_definitions(dolfin PUBLIC EIGEN_MAX_ALIGN_BYTES=32)
+target_compile_definitions(dolfin PUBLIC "EIGEN_MAX_ALIGN_BYTES=${DOLFIN_ALIGNMENT_BYTES}")
 
 # Boost
 target_link_libraries(dolfin PUBLIC Boost::boost)
--- a/cpp/dolfin/CMakeLists.txt
+++ b/cpp/dolfin/CMakeLists.txt
@@ -102,7 +102,7 @@ target_include_directories(dolfin SYSTEM PUBLIC ${UFC_INCLUDE_DIRS})
 set(DOLFIN_EIGEN_MAX_ALIGN_BYTES "32" CACHE STRING "\
 Minimum alignment in bytes used for Eigen data structures. Set to 32 for \
 compatibility with AVX user-compiled code and 64 for AVX-512 user-compiled \
-code. Set to 0 for ideal alignment on your CPU. Note that if an architecture \
+code. Set to 0 for ideal alignment according to -march. Note that if an architecture \
 flag (e.g. \"-march=skylake-avx512\") is set for DOLFIN, Eigen will use the \
 appropriate ideal alignment instead if it is stricter. Otherwise, the value \
 of this variable will be used by Eigen for the alignment of all data structures.\\ 
--- a/cpp/dolfin/CMakeLists.txt
+++ b/cpp/dolfin/CMakeLists.txt
@@ -88,6 +88,8 @@ target_include_directories(dolfin SYSTEM PUBLIC ${UFC_INCLUDE_DIRS})
 
 # Eigen3
 target_include_directories(dolfin SYSTEM PUBLIC ${EIGEN3_INCLUDE_DIR})
+# Force Eigen to use 32-byte alignment for compatibility with AVX code
+target_compile_definitions(dolfin PUBLIC EIGEN_MAX_ALIGN_BYTES=32)
 
 # Boost
 target_link_libraries(dolfin PUBLIC Boost::boost)
--- a/cpp/dolfin/CMakeLists.txt
+++ b/cpp/dolfin/CMakeLists.txt
@@ -86,15 +86,31 @@ endif()
 # UFC
 target_include_directories(dolfin SYSTEM PUBLIC ${UFC_INCLUDE_DIRS})
 
-# Currently only used for Eigen
-set(DOLFIN_ALIGNMENT_BYTES "32" CACHE STRING "\
-Minimum alignment in bytes used for numeric arrays. \
-For compatibility with AVX2, AVX-512... vectorized user code. \
+# DOLFIN uses Eigen data structures for dense linear algebra operations. Eigen
+# performs 'ideal' memory alignment based around the -march flag passed to the
+# compiler.  However, because Python DOLFIN JIT compiles code at runtime, it is
+# possible for the user to build shared objects with incompatible alignment
+# (ABI) if they use a different -march flag than that used to originally build
+# DOLFIN. DOLFIN_EIGEN_MAX_ALIGN_BYTES can be used to force alignment.
+# See: https://eigen.tuxfamily.org/dox/TopicPreprocessorDirectives.html
+# See: https://github.com/FEniCS/dolfinx/pull/143
+
+# Note: The name EIGEN_MAX_ALIGN_BYTES is confusing. In practice, Eigen
+# computes the ideal alignment based around -march.  If the ideal alignment is
+# greater than EIGEN_MAX_ALIGN_BYTES, the ideal alignment is used. If the ideal
+# alignment is less, then EIGEN_MAX_ALIGN_BYTES is used for alignment.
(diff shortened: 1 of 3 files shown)


w1th0utnam3:
Before merging: How to deal with AVX-512 case which actually requires 64-byte alignment?

pefarrell:
In case you figure out how to deal with the AVX-512 case, I've run into exactly this problem in firedrake: https://github.com/firedrakeproject/firedrake/issues/1238

w1th0utnam3:
@pefarrell I don't have an AVX-512 CPU for testing but I would assume it would suffice to just use the preprocessor directive from this PR and replace 32 with 64?

jhale:
Hi Patrick,

We're discussing this in #simd on slack.

Jack

blechta:
Does the definition get propagated to the DOLFIN pkgconfig file?

w1th0utnam3:
@blechta I'm a little bit confused by pkgconfig. The file `/usr/local/lib/pkgconfig/dolfin.pc` is generated by the `cpp/cmake/templates/dolfin.pc.in` file. For my build, I have a line:
```
definitions= -DDEBUG -DHAS_HDF5 -D_FORTIFY_SOURCE=2 -DHAS_SCOTCH -DDOLFIN_VERSION="2018.2.0.dev0"
```
in the generated file. The corresponding line in the template is `definitions=@PKG_DEFINITIONS@`. Where does the content of `@PKG_DEFINITIONS@` come from? Neither searching for `PKG_DEFINITIONS` nor `DEFINITIONS` leads to useful results in the repository.

blechta:
@w1th0utnam3 I don't know. Maybe a CMake feature?

cpp/dolfin/CMakeLists.txt
```diff
@@ -88,6 +88,8 @@ target_include_directories(dolfin SYSTEM PUBLIC ${UFC_INCLUDE_DIRS})
 
 # Eigen3
 target_include_directories(dolfin SYSTEM PUBLIC ${EIGEN3_INCLUDE_DIR})
+# Force Eigen to use 32-byte alignment for compatibility with AVX code
+target_compile_definitions(dolfin PUBLIC EIGEN_MAX_ALIGN_BYTES=32)
```
garth-wells: This should be a CMake option (so the user can set it), with a default value.

w1th0utnam3: Fixed. Not sure about naming though. Using `EIGEN_MAX_ALIGN_BYTES` suggest IMO that it only affects Eigen itself, whereas it actually also affects coupling with user code. Therefore, called it `DOLFIN_ALIGNMENT_BYTES`

w1th0utnam3:
@garth-wells Do you have an opinion about the default value for the alignment? Currently I'm using 32 bytes. 64 bytes however would allow working with AVX-512 and setting it as the default value allows anybody to use it who might rely on a pre-build package. It shouldn't waste too much memory/performance because we don't have many small matrices? Or are you aware of a use case where a lot of small matrices are stored consecutively?

garth-wells:
We could experiment with 64 byte alignment as the default. Important that user can ultimately control the build.

Assume or anticipate use cases at one's peril! Some past FEniCS design mistakes were motivated by assumption on usage. I know of cases where lots of small matrices might  be used.

cpp/dolfin/CMakeLists.txt
```diff
@@ -86,8 +86,15 @@ endif()
 # UFC
 target_include_directories(dolfin SYSTEM PUBLIC ${UFC_INCLUDE_DIRS})
 
+# Currently only used for Eigen
+set(DOLFIN_ALIGNMENT_BYTES "32" CACHE STRING "\
```
garth-wells: Maybe `DOLFIN_EIGEN_MAX_ALIGNMENT_BYTES` is a better name?

cpp/dolfin/CMakeLists.txt
```diff
@@ -86,8 +86,15 @@ endif()
...
+Minimum alignment in bytes used for numeric arrays. \
```
garth-wells: Min of max? Eigen name has `MAX` in it.

cpp/dolfin/CMakeLists.txt
jhale: Sounds good.

jhale:
@garth-wells Placing the compiler flags into pkgconfig would be one idea and I think it would be nice to have that.

However, it doesn't cover the following use case:

The user has installed a binary version of DOLFIN (Ubuntu, Docker, Anaconda) compiled with a rather conservative set of compiler flags (maybe just `-mtune=intel` or `-march=x86-64`). Despite using a generic binary install, this user still potentially has a lot to gain by setting up DOLFIN to JIT modules using `-march=native` and using @w1th0utnam3 SIMD assembler. Setting `EIGEN_MAX_ALIGN_BYTES=32` would allow this to work seamlessly. Without it, the user will get DOLFIN that segfaults if they set `-march=native`.

I see two options for implementation:

1) We set `EIGEN_MAX_ALIGN_BYTES=32` at the source level, i.e. we developers enforce it. As yo say, this may have unintended side effects.
2) We allow/instruct binary maintainers to set `EIGEN_MAX_ALIGN_BYTES=32` or `64` based around the perceived needs of their users.

@w1th0utnam3 As already discussed, we will not be force setting `-march=native` anywhere!

w1th0utnam3:
> @jhale: As already discussed, we will not be force setting -march=native anywhere!

I don't want to force any compiler flags either.


> @garth-wells: "I'm concerned that this could get confusing. If I choose to use march=native consistently I guess I would have to reverse engineer the flags to figure out the alignment."

This has to be documented but I don't think that it is too confusing. If `-march=...` requires stricter alignment, Eigen will use it automatically and override `EIGEN_MAX_ALIGN_BYTES`. If the user specifies a stricter alignment with `EIGEN_MAX_ALIGN_BYTES` than the (default) `-march`, Eigen will use that. That's all.

cpp/dolfin/CMakeLists.txt
w1th0utnam3: Ok

cpp/dolfin/CMakeLists.txt
w1th0utnam3: Yes, this is weird. Eigen docs say "EIGEN_MAX_ALIGN_BYTES - Must be a power of two, or 0. Defines an upper bound on the memory boundary in bytes on which dynamically and statically allocated data may be aligned by Eigen."
Why would max make sense in this context? As stated in my latest comment below, Eigen uses an alignment in bytes which is **at least** this big. If `-march` requires a stricter (i.e. larger) alignment, this is used instead. Therefore I wrote "minimum" in the explaining comment.

jhale: I found the naming `EIGEN_MAX_ALIGN_BYTES` confusing too when I read the Eigen documentation but I thought I was being stupid :octocat:

w1th0utnam3: It may make sense from their internal use:
```
#if EIGEN_IDEAL_MAX_ALIGN_BYTES > EIGEN_MAX_ALIGN_BYTES
    #define EIGEN_DEFAULT_ALIGN_BYTES EIGEN_IDEAL_MAX_ALIGN_BYTES
#else
    #define EIGEN_DEFAULT_ALIGN_BYTES EIGEN_MAX_ALIGN_BYTES
#endif
```
So `EIGEN_MAX_ALIGN_BYTES` can be interpreted something like `EIGEN_MAX_USER_REQUESTED_ALIGN_BYTES` but still... it's weird...

garth-wells: @w1th0utnam3 Could you put some text to this effect in the comment? Otherwise we'll all forget this and be rehashing it again in 12 months.

w1th0utnam3: @garth-wells Yes, that's reasonable. Regarding the comment: unfortunately, the CMake console gui uses only a single line to display the comment. The comment for the explanation will obviously be longer than most terminal widths. Do you think this is ok?
image
Full comment:
```
set(DOLFIN_EIGEN_MAX_ALIGN_BYTES "32" CACHE STRING "\
Sets memory alignment of Eigen matrices. Can be used for compatibility with \
user code that is compiled with AVX2 or AVX-512 vector instructions. \
Note that if an architecture flag (e.g. \"-march=skylake-avx512\") is \
set for DOLFIN, Eigen will use the appropriate ideal alignment instead if \
it is stricter. Otherwise, the value of this variable will be used by Eigen \
for the alignment of all arrays/matrices.\
")
```
At least it's documented if you look into CMakeLists.txt or the cache...

garth-wells: CMake help string looks good, but I had in mind a developer 'code' comment on max vs min so in 12 months when we all forget we can be reminded of your reasoning.

PR #146, opened 2018-08-02
[WIP] Set nightly builds for complex mode

IgorBaratta:
Proposal:
 - nightly builds that run every day at 00:00 am UTC,  to test complex mode.
 - reuse code (steps) for the real and complex workflows.

Real mode (default) is triggered when pushing to any branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -206,7 +206,7 @@ workflows:
   nightly:
     triggers:
       - schedule:
-          cron: "0 2 * * *"
+          cron: "0 5 * * *"
           filters:
             branches:
               only:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -202,11 +202,11 @@ workflows:
           filters:
             branches:
               only:
-                - igor/nightly-build
+                - master
   nightly:
     triggers:
       - schedule:
-          cron: "0 0 * * *"
+          cron: "0 2 * * *"
           filters:
             branches:
               only:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,3 +1,120 @@
+install-dependencies: &install-dependencies
+  name: Install/update Python (test/doc) dependencies
+  command: pip3 install decorator flake8 matplotlib numba pygmsh pytest pytest-xdist sphinx sphinx_rtd_theme --upgrade
+
+install-python-components: &install-python-components
+  name: Install FEniCS Python components
+  command: |
+            pip3 uninstall --yes fenics-fiat fenics-ufl fenics-dijitso fenics-ffc
+            pip3 install git+https://bitbucket.org/fenics-project/fiat.git --upgrade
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -148,4 +202,14 @@ workflows:
           filters:
             branches:
               only:
-                - master
+                - igor/nightly-build
```
blechta: Why is this changed?

.circleci/config.yml
```diff
@@ -148,4 +202,14 @@ workflows:
...
+  nightly:
+    triggers:
+      - schedule:
+          cron: "0 0 * * *"
+          filters:
+            branches:
+              only:
+                - igor/nightly-build
```
blechta: Should be changed to `master` when merged? Or could we run all branches?

.circleci/config.yml
```diff
@@ -148,4 +202,14 @@ workflows:
           filters:
             branches:
               only:
-                - master
+                - igor/nightly-build
+  nightly:
+    triggers:
+      - schedule:
+          cron: "0 0 * * *"
```
blechta: What timezone is it? UTC? If so maybe we should use even less busy hours.

.circleci/config.yml
IgorBaratta: I was getting an error to persist a temporary file to the `pushdoc` job.
Changed back to master.

.circleci/config.yml
IgorBaratta: Yes, I was thinking in changing it to `master`.
If we program all branches to run at the same time it can generate a large queue.
I haven't found a filter to run only active branches.

.circleci/config.yml
IgorBaratta: UTC.
02:00 am would be better?

blechta: UTC 2.00 is still quite early for those working in CEST. What about UTC 5.00? Seems early enough before busy hours in Cambridge at winter.

.circleci/config.yml
blechta: https://circleci.com/docs/2.0/configuration-reference/#branches-1 seems to imply that only pushed commits schedule to a build. Could you verify that experimentally?

IgorBaratta: Yeah, the default behavior is to trigger the workflow when pushing to a branch. With the keyword `schedule`  it behaves a bit different.
I will schedule two consecutive builds (to avoid waiting two days) without committing between them to verify if in that case a commit is needed.

IgorBaratta: It runs the workflow on schedule regardless of having no new commit.
See for example (hourly builds): https://circleci.com/gh/IgorBaratta/dolfinx

blechta: Too bad, good digging. Ok, do it the way you think is the best.

IgorBaratta:
The complex build will not be green until [#144](https://github.com/FEniCS/dolfinx/pull/144) is merged, and the `scalar_type` form compiler parameter is set.

PR #148, opened 2018-08-02
[WIP] Implementing cross-cell integral assembly

w1th0utnam3:
This PR implements basic batched assembly for cell integrals. Intended to work together with cross cell vectorization in FFC. 

This PR **depends on** https://github.com/FEniCS/ffcx/pull/52 for changes in [ufc.h](https://github.com/FEniCS/ffcx/blob/fabian/gcc-vec-ext/ffc/backends/ufc/ufc.h), however these changes could be extracted from the the large vectorization PR.

**Open points:**
 - Add convenience function to Python package that can set correct FFC and dijitso parameters for vectorization, might also check alignments supported by DOLFIN
 - Check if it makes sense to make `coordinate_dofs` evaluation and coefficient restriction somehow batched? Currently they write into separate arrays for each cell followed by copying them into cell strided/interleaved memory. Maybe it makes sense to leave the inner workings of these methods as-is and just allow working with `Eigen::Map` classes to allow strides which would remove one copy for each cell?
 - Are the implemented test cases sufficient for now? Are there corner cases that have to be checked?
 - Consider forms with subdomains. Currently it is assumed that all cells have the same size of dofmaps and use the same `tabulate_tensor` function. Implementing support for multiple `tabulate_tensor` functions is would also require corresponding tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -19,7 +19,7 @@ jobs:
             pip3 install git+https://bitbucket.org/fenics-project/fiat.git --upgrade
             pip3 install git+https://bitbucket.org/fenics-project/ufl.git --upgrade
             pip3 install git+https://bitbucket.org/fenics-project/dijitso.git --upgrade
-            pip3 install git+https://github.com/fenics/ffcx.git --upgrade
+            pip3 install git+https://github.com/fenics/ffcx.git@fabian/gcc-vec-ext --upgrade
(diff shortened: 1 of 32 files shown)


blechta:
@w1th0utnam3 could you resolve the conflicts?

w1th0utnam3:
@blechta Fixed

garth-wells:
Closing for now - a lot has changed in the meantime.

Issue #158, opened 2018-08-28
MeshEntity.entities(dim) fails when dim is same dimension as entity


chrisrichardson:
For example in `Subdomain::mark()` there is:
```
const std::int32_t* v = facet.entities(0);
```
which fails in 1D, because a facet is of dimension 0.

PR #162, opened 2018-09-06
Update code for entities when same dim

chrisrichardson:
Fixes issue #158

### DIFF
--- a/cpp/dolfin/mesh/MeshEntity.h
+++ b/cpp/dolfin/mesh/MeshEntity.h
@@ -125,6 +125,9 @@ public:
   /// dimension.
   inline std::size_t num_entities(std::size_t dim) const
   {
+    if (dim == _dim)
+      return 1;
+
     return _mesh->topology().connectivity(_dim, dim).size(_local_index);
   }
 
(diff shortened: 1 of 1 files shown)


cpp/dolfin/mesh/MeshEntity.h
```diff
@@ -125,6 +125,9 @@ class MeshEntity
   /// dimension.
   inline std::size_t num_entities(std::size_t dim) const
   {
+    if (dim == _dim)
+      return 1;
+
```
garth-wells: could you add an `else` in?

cpp/dolfin/mesh/MeshEntity.h
```diff
@@ -139,6 +142,9 @@ class MeshEntity
   ///         dimension.
   std::size_t num_global_entities(std::size_t dim) const
   {
+    if (dim == _dim)
+      return 1;
+
```
garth-wells: could you add an `else` in?

cpp/dolfin/mesh/MeshEntity.h
```diff
@@ -152,6 +158,9 @@ class MeshEntity
   ///         The index for incident mesh entities of given dimension.
   const std::int32_t* entities(std::size_t dim) const
   {
+    if (dim == _dim)
+      return &_local_index;
+
```
garth-wells: could you add an `else` in?

cpp/dolfin/mesh/MeshEntity.h
chrisrichardson: I could do - does it make it better? Some compilers will complain that "control reaches end of function returning void" or something...

PR #174, opened 2018-09-15
Work on new assembler with simpler interface.

garth-wells:


cpp/dolfin/fem/Assembler.cpp
```diff
@@ -52,7 +52,31 @@ Assembler::~Assembler()
   // }
 }
 //-----------------------------------------------------------------------------
-void Assembler::assemble(la::PETScMatrix& A, BlockType block_type)
+la::PETScMatrix Assembler::assemble_matrix(BlockType block_type)
+{
+  // Check if matrix should be nested
+  assert(!_a.empty());
```
michalhabera: I saw in LLVM they use assert in the following fashion,
`assert(!_a.empty() && "List of bilinear forms should not be empty!")`

Maybe we can consider that as new asserts style, I really like it.

PR #180, opened 2018-09-19
Update XDMF topolgy types with full name

nate-sime:
Change the XDMF topolgy types to match the schema

https://github.com/Kitware/VTK/blob/master/ThirdParty/xdmf3/vtkxdmf3/Xdmf.dtd#L34


michalhabera:
Adding test for XDMF XML is good idea. Its more objective test than just output/input/compare. It assures us, that we are not doing IO of our own format. I like it.
Lets merge this one, I am making issue for schema validation.

PR #182, opened 2018-09-22
Refactor dolfin python's Form

michalhabera:
No python layer wrapper should subclass from its `cpp.foo` counterpart.

I removed this inheritance and added inheritance from `ufl.Form` instead. (the inheritance is not used anywhere, but we should be more systematic)

I also propose to strictly follow the rule: 
> 1. Any dolfin's python object that attaches data (function spaces, coefficients, ...) to UFL object should subclass it.

In the virtue of this rule, `dolfin.fem.Form` is indeed attaching function spaces to `cpp.fem.Form`.

### DIFF
--- a/python/dolfin/fem/__init__.py
+++ b/python/dolfin/fem/__init__.py
@@ -11,6 +11,8 @@ import dolfin.cpp
 
 from .form import Form
 
+__all__ = ["Form"]
+
 
 def create_coordinate_map(o):
     """Return a compiled UFC coordinate_mapping object"""
--- a/python/dolfin/fem/__init__.py
+++ b/python/dolfin/fem/__init__.py
@@ -9,6 +9,8 @@ import ufl
 from dolfin.jit.jit import ffc_jit
 import dolfin.cpp
 
+from .form import Form
+
 
 def create_coordinate_map(o):
     """Return a compiled UFC coordinate_mapping object"""
--- a/python/dolfin/fem/__init__.py
+++ b/python/dolfin/fem/__init__.py
@@ -8,8 +8,7 @@
 import ufl
 from dolfin.jit.jit import ffc_jit
 import dolfin.cpp
-
-from .form import Form
+from dolfin.fem.form import Form
 
 __all__ = ["Form"]
 
--- a/python/dolfin/fem/assembling.py
+++ b/python/dolfin/fem/assembling.py
@@ -18,7 +18,7 @@ rely on the dolfin::Form class which is not used on the Python side.
 
 import ufl
 import dolfin.cpp as cpp
-from dolfin.fem.form import Form
+import dolfin.fem as fem
 
 __all__ = ["assemble_local", "SystemAssembler"]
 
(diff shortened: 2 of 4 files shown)


python/dolfin/fem/assembling.py
```diff
@@ -39,10 +38,10 @@ def _create_cpp_form(form, form_compiler_parameters=None,
                 "Ignoring form_compiler_parameters when passed a dolfin Form!")
         return form
     elif isinstance(form, ufl.Form):
-        return Form(
+        form = Form(
```
garth-wells: Change this to `form = fem.Form` (and modify the above imports to reflect the change).

This will help emphasise distinction between `ufl.Form`, `dolfin.fem.(form).Form` and `dolfin.cpp.fem.Form`.

python/dolfin/fem/form.py
```diff
@@ -1,85 +1,86 @@
...
-        # Check form argument
-        if not isinstance(form, ufl.Form):
-            raise RuntimeError("Expected a ufl.Form.")
-
-        sd = form.subdomain_data()
-        self.subdomains, = list(sd.values())  # Assuming single domain
-        domain, = list(sd.keys())  # Assuming single domain
-        mesh = domain.ufl_cargo()
+class Form(ufl.Form):
+    def __init__(self, form: ufl.Form, form_compiler_parameters: list=[]):
```
garth-wells: I thought form compiler options were usually a `dict` (rather than a `list`).

python/dolfin/fem/form.py
```diff
@@ -1,85 +1,86 @@
...
+        C++ Form.
+        """
+        self.form_compiler_parameters = form_compiler_parameters
 
         # Add DOLFIN include paths (just the Boost path for special
         # math functions is really required)
         # FIXME: move getting include paths to elsewhere
-        if form_compiler_parameters is None:
-            form_compiler_parameters = {"external_include_dirs": dolfin_pc["include_dirs"]}
+        if self.form_compiler_parameters is None:
```
garth-wells: But default is now `[]` rather than `None`.

python/dolfin/fem/form.py
```diff
@@ -1,85 +1,86 @@
...
-            self.set_mesh(mesh)
+            self._cpp_object.set_mesh(mesh)
 
         # Attach subdomains to C++ Form if we have them
-        subdomains = self.subdomains.get("cell")
+        subdomains = self._subdomains.get("cell")
         if subdomains is not None:
-            self.set_cell_domains(subdomains)
-        subdomains = self.subdomains.get("exterior_facet")
+            self._cpp_object.set_cell_domains(subdomains)
```
garth-wells: Should be no need to do `if subdomains is not None:`, etc. Just doing
```
self._cpp_object.set_cell_domains(subdomains)
``` 
should work fine in all cases. When `subdomains` is `None`, it will turn into `null` going into the C++ layer.

python/dolfin/fem/form.py
michalhabera: Made default back to `None`

michalhabera:
Fixed, green now @garth-wells 

But we have fix the `import` scheme ASAP. Its hitting back again and again...

python/dolfin/fem/__init__.py
```diff
@@ -9,6 +9,10 @@
 from dolfin.jit.jit import ffc_jit
 import dolfin.cpp
 
+from .form import Form
```
garth-wells: Not sure about relative imports. I don't mind, but I think we reached an earlier consensus to avoid.

Issue #183, opened 2018-09-23
Remove GenericFunction


michalhabera:
`GenericFunction` is too generic, covers too wide range of scenario.
This generality leads to implicit behaviour. 

Example 1: Values of BCS are GenericFunction. Inside DirichletBC call to `GenericFunction.restrict` is made, which "interpolates" values in every cell. It works for `Constant`, since constant inherits from `GenericFunction`, but leads to hidden suboptimal behaviour.

Example 2: Coefficients for tabulation kernels are GenericFunction, e.g. in Form.cpp. By call to `restrict` they are again cell-wise interpolated to target FE space. It works for Expression, since expression inherits from `GenericFunction`, but hides the interpolation to FE element. Very common pitfall for users.

I will remove GenericFunction in following steps (each one as a PR):

- [x] Make `Function` not inherit from GenericFunction.
- [x] Make `Expression` not inherit from GenericFunction.
- [x] Replace `GenericFunction`  with `Function`.

The arguments supporting this issue are also:
- We must strictly separate concepts of FE function (Function.cpp) and anything else.
- We should specialise evaluation of constant valued expr and optimise for performance.
- We must make user more aware of relation between objects in his UFL form and heavy dolfin data.
- Explicit is better than implicit.
- Simple is better than complex.

michalhabera:
Let me motivate the proposal again in more detail. I will here talk about why the implicit behaviour
is bad in our case.

### The issue of `Expression` implicit behaviour

In the current code, when user writes (lets stay on triangles)

```python
a = Expression("sin(x[0]*x[1])", degree=2) * TestFunction(P1) * dx
```

this translates into a tabulation kernel which accepts one coefficient `w[6]`.
It is the vector of coefficients with respect to P2 basis functions.
For every cell, there is an underlying local interpolation of `sin(x[0]*x[1])`
to P2 space, which results in cell-local set of these coefficients.

Instead, if one does
```python
x = SpatialCoordinate(mesh)
a = sin(x[0] * x[1]) * TestFunction(P1) * dx
```

function `sin(x[0] * x[1])` is evaluated at quadrature points and resulting tabulation
kernel doesn't have any coefficients as arguments. (no additional memory fetching
during assembly!)

This shows unnecesary implicitness and complexity that comes with acceptation
of generic object in form. It is not very "natural" to hide the true behaviour from user.

Making `degree` in expression more explicit we pushed a bit users to think about
[...]

PR #184, opened 2018-09-23
Make `Function` not inherit `GenericFunction`

michalhabera:
Supports #183 .

This PR makes `Function` again fully fledged independent class.
It also disables usage of Function as GenericFunction.

I redefined what could be attached as coefficient to a form. Now only `Function` is allowed, i.e. the true finite element function. This temporarily disables use of `Constant` and `Expression` in forms, however I replaced constant with scalar values for time being.

Since `Constant` and `Expression` are children of `GenericFunction`, they can still be used in interpolation and boundary conditions.

Ad `Expression`: each use of expression in tests could have been (and I did so) replaced with its true UFL equivalent. This is **the exact** representation at quadrature points. I had to redefine reference norms in tests.

I changed `Poisson.ufl` and its `main.cpp` such that no `Expression` is used. Again, Source term and flux term could be **exactly** replaced in .ufl file usign UFL primitives.


michalhabera:
Tests are green finally.

garth-wells:
This has some major interface changes and some loss of functionality that need to be discussed in detail. 

I think the existing high level interface for forms is pretty good, so changes should be made with caution. It would be best to sketch out a design proposal as a 'Project' (https://github.com/FEniCS/dolfinx/projects)

michalhabera:
Sure I will do so.

This is not really changing user high level interface. It only makes underlying concepts clearer. But dont be afraid to totally discard it, I am ok with it ;)

garth-wells:
As agreed in discussion, this change will be broken down into small steps with separate pull request.

PR #188, opened 2018-09-25
Wrap cpp.common in python layer

michalhabera:
Hides `cpp.common.*` namespace from public python use. Didn't add docstring yet.

Please note very nice segregation, i.e. how `cpp.common.*` is used exclusively in `dolfin/common.py` as a backend. No elsewhere. And that we should follow as a guideline.


michalhabera:
There are some duplicate comments showing as diff, because I forked other branch. But it should be merged fine.

python/dolfin/__init__.py
```diff
@@ -31,13 +31,15 @@
...
 
-from .cpp.common import (Variable, has_debug, has_hdf5, has_scotch,
-                         has_hdf5_parallel, has_mpi4py, has_petsc_complex,
-                         has_petsc4py, has_parmetis, has_slepc, has_slepc4py,
-                         git_commit_hash, DOLFIN_EPS, DOLFIN_PI, TimingType,
-                         timing, timings, list_timings)
 
-from .cpp import MPI
+from dolfin.common import (
+    Variable, has_debug, has_hdf5, has_scotch, has_hdf5_parallel,
```
garth-wells: Do we want `Variable` exposed?

python/dolfin/__init__.py
```diff
@@ -31,13 +31,15 @@
...
+    has_mpi4py, has_petsc_complex, has_petsc4py, has_parmetis,
+    has_slepc, has_slepc4py, git_commit_hash, DOLFIN_EPS, DOLFIN_PI,
```
garth-wells: I don't think we should expose `DOLFIN_EPS` and `DOLFIN_PI`. `math.pi`, etc are available.

python/dolfin/common.py
```diff
@@ -0,0 +1,249 @@
...
+        self._cpp_object.rename(new_name)
+
+    def name(self):
+        return self._cpp_object.name()
+
+    def id(self):
+        return self._cpp_object.id()
+
+
+def has_debug():
```
garth-wells: Make these properties.

python/dolfin/common.py
```diff
@@ -0,0 +1,249 @@
...
+
+def has_slepc4py():
+    return cpp.common.has_slepc4py()
+
+
+def git_commit_hash():
+    return cpp.common.git_commit_hash()
+
+
+DOLFIN_EPS = cpp.common.DOLFIN_EPS
```
garth-wells: Remove

python/dolfin/common.py
```diff
@@ -0,0 +1,249 @@
...
+DOLFIN_PI = cpp.common.DOLFIN_PI
```
garth-wells: Remove

python/dolfin/common.py
```diff
@@ -0,0 +1,249 @@
...
+    def start(self):
+        self._cpp_object.start()
+
+    def stop(self):
+        return self._cpp_object.stop()
+
+    def resume(self):
+        self._cpp_object.resume()
+
+    def elapsed(self):
```
garth-wells: Make property.

python/dolfin/common.py
```diff
@@ -0,0 +1,249 @@
...
+
+    @staticmethod
+    def init_level(args: list, required_thread_level: int):
+        """Initialise MPI with command-line args and required level
+        of thread support.
+        """
+        cpp.MPI.init(args, required_thread_level)
+
+    @staticmethod
+    def responsible():
```
garth-wells: Make properties where appropriate.

python/dolfin/__init__.py
michalhabera: DOLFIN_EPS is scattered all over demos and tests. What should I replace it with?

garth-wells: Eventually https://docs.scipy.org/doc/numpy/reference/generated/numpy.finfo.html.

How about just adding a comment for now that it might be removed?

python/dolfin/common.py
michalhabera: I don't see what you mean. Would really help If you can sketch one :) . Note these are theoretically runtime changing.

garth-wells: It would be nice to do:
```
foo = MPI.responsible
```
rather than
```
foo = MPI.responsible()
```

michalhabera:
Free methods in `MPI.py` can't be made properties. Only non-static member methods can be.

PR #192, opened 2018-09-27
Wrap cpp.fem.dofmap in python

michalhabera:
Wrapped `cpp.fem.dofmap` as `fem.dofmap`.

"Overloaded" constructor is solved via `@classmethod`.

### DIFF
--- a/python/dolfin/__init__.py
+++ b/python/dolfin/__init__.py
@@ -42,9 +42,6 @@ import dolfin.MPI
 
 from dolfin.fem import DofMap
 
-from .cpp.function import (Expression, Constant)
-from .cpp.fem import FiniteElement
-
 from .cpp.geometry import BoundingBoxTree, Point
 from .cpp.generation import IntervalMesh, BoxMesh, RectangleMesh
 
--- a/python/dolfin/__init__.py
+++ b/python/dolfin/__init__.py
@@ -40,8 +40,10 @@ from dolfin.common import (
 
 import dolfin.MPI
 
+from dolfin.fem import DofMap
+
 from .cpp.function import (Expression, Constant)
-from .cpp.fem import (FiniteElement, DofMap)
+from .cpp.fem import FiniteElement
 
 from .cpp.geometry import BoundingBoxTree, Point
 from .cpp.generation import IntervalMesh, BoxMesh, RectangleMesh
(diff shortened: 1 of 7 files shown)


python/dolfin/__init__.py
```diff
@@ -40,8 +40,7 @@
 
 import dolfin.MPI
 
-from .cpp.function import (Expression, Constant)
-from .cpp.fem import (FiniteElement, DofMap)
+from dolfin.fem import DofMap
```
garth-wells: Do we want this at the top level?

python/dolfin/fem/__init__.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-# Copyright (C) 2017-2018 Garth N. Wells
+# Copyright (C) 2017-2018 Garth N. Wells, Michal Habera
```
garth-wells: Use 'and'

python/dolfin/fem/__init__.py
```diff
@@ -12,6 +12,9 @@
                                    SystemAssembler)
 from dolfin.fem.coordinatemapping import create_coordinate_map
 from dolfin.fem.dirichletbc import DirichletBC, AutoSubDomain
+from dolfin.fem.dofmap import (DofMap, make_ufc_dofmap,
```
garth-wells: Do we need to import `make_ufc_foo`?

python/dolfin/fem/dofmap.py
```diff
@@ -0,0 +1,101 @@
...
+#
+# SPDX-License-Identifier:    LGPL-3.0-or-later
+
+from dolfin import cpp
+from dolfin import fem
+
+
+def make_ufc_finite_element(ufc_finite_element):
+    """Returns ufc finite element from a pointer
+    """
```
garth-wells: Should this be moved up a line for one-line docstrings?

python/dolfin/__init__.py
michalhabera: Now I am just working within `python/dolfin/` keeping global imports untouched (i.e. user space). 
After everything is wrapped I'll remove global imports and this needs a lot of work in tests/demos. Otherwise this PR will loose its 'locality'.

garth-wells: Ok, leave it for now. I think we'll remove it pretty soon.

python/dolfin/fem/__init__.py
michalhabera: I other `python/dolfin/` files I do `from dolfin import fem` and work directly with `fem.make_ufc_...`. I want to keep everything 2 level, even for internal wrapper calls.

michalhabera: But feel free to discuss.

michalhabera: Ok, let's export to 2nd level only public (be interfaced by users) classes. Done for dofmap.py

PR #195, opened 2018-09-29
Wrap geometry in python

michalhabera:
The same idea as #192 .

Most of the `Point._cpp_object` in public code will be removed once we make all wrappers accepting `dolfin.geometry.Point`.


michalhabera:
Yes, It is indeed redundant... https://github.com/FEniCS/dolfinx/issues/196
I am anyway happy for this to be merged and Point will be removed in later stages.

PR #206, opened 2018-10-12
Simplify Python FunctionSpace creation

garth-wells:
- Simplifies FunctionSpace initialiser
- Remove inelegant re-construction of UFL FiniteElements from string signatures from cpp objects
- Interface change to require 'bundling' of element family string and degree into into a list/tuple (leads to simplification of initialiser)

Also adds `legacy' module (name can be changed) to support old syntax.


python/dolfin/function/functionspace.py
```diff
@@ -5,43 +5,50 @@
...
 
 
+class ElementDef(typing.NamedTuple):
+    """Data for representing a finite element"""
+    family: str
+    degree: int
+    form_degree: typing.Optional[int] = None  # noqa
+
+
 class FunctionSpace(ufl.FunctionSpace):
```
michalhabera: I am not sure why `dolfin.function.FunctionSpace` should inherit from `ufl.FunctionSpace`. 
Its not directly related to this issue, but better think about it. We should probably eliminate inheritance wherever possible.

garth-wells: Agree it should be considered (as a separate issue).

python/dolfin/function/functionspace.py
```diff
@@ -5,43 +5,50 @@
...
-        # FIXME: This includes some hacks to support construction of
-        # function spaces with existing cpp FunctionSpace, which is
-        # typical when extracting subspaces.
-        #
-        # It would element if an element argument was a single object,
-        # e.g. a ufl.FiniteElement or a tuple(family degree)
-
+        # Create function space from a UFL element and existing cpp
+        # FunctionSpace
         if cppV:
```
jhale: `if cppV is not None:` ?

python/dolfin/function/functionspace.py
```diff
@@ -5,43 +5,50 @@
...
 # SPDX-License-Identifier:    LGPL-3.0-or-later
 
+import typing
+
 import ufl
 from dolfin import cpp, jit
 from dolfin.fem import dofmap
 
 
+class ElementDef(typing.NamedTuple):
```
jhale: `ElementDefinition`?

python/dolfin/function/functionspace.py
```diff
@@ -5,43 +5,50 @@
...
+    """Data for representing a finite element"""
+    family: str
+    degree: int
+    form_degree: typing.Optional[int] = None  # noqa
```
jhale: Doesn't seem complete without a cell type? I the cell_type was `None`, FunctionSpace could infer it from the mesh?

garth-wells: It's the necessary meta-data.  If the element is fully defined, one case just pass an element.

PR #207, opened 2018-10-12
Rewrite expressions to UFL in python tests

michalhabera:
Rewrite iff C++ Expression is used in a form to assemble.

C++ Expression still present in interpolation and DirichletBC.

### DIFF
--- a/python/test/unit/parallel-assembly-solve/test_solve_result_against_reference.py
+++ b/python/test/unit/parallel-assembly-solve/test_solve_result_against_reference.py
@@ -12,9 +12,8 @@ parallel assembly/solve."""
 import ufl
 
 from dolfin import (MPI, TestFunction, TrialFunction, FunctionSpace,
-                    UnitSquareMesh, UnitCubeMesh, Expression,
-                    inner, grad, dx, ds, Function, solve, cpp,
-                    SpatialCoordinate)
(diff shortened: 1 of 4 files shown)


python/test/unit/parallel-assembly-solve/test_solve_result_against_reference.py
```diff
@@ -97,22 +100,10 @@ def print_errors(errors):
...
-                 ("4x4x4 unit tet cube", 4): 74.49938266409099,
-                 ("16x16 unit quad square", 1): 9.550848071820747,
-                 ("16x16 unit quad square", 2): 18.423668706176354,
-                 ("16x16 unit quad square", 3): 27.295831017251672,
-                 ("16x16 unit quad square", 4): 36.168671281610855,
-                 ("4x4x4 unit hex cube", 1): 12.151954087339782,
-                 ("4x4x4 unit hex cube", 2): 28.965646690046885,
-                 ("4x4x4 unit hex cube", 3): 49.97349423895635,
-                 ("4x4x4 unit hex cube", 4): 74.49938136593539}
+    reference = {("16x16 unit tri square", 1): 9.544967937540488,
```
garth-wells: Why fewer reference values than before?

michalhabera: Were not used. So I just deleted. And with UFL expression it gives slightly different numbers anyway.

PR #212, opened 2018-10-14
More __call__ simplify

garth-wells:


python/test/unit/fem/test_permutation.py
```diff
@@ -42,6 +42,7 @@ def test_p4_scalar_vector():
 
         pts = numpy.array([[0.4, 0.4, 0.1], [0.4, 0.1, 0.4], [0.1, 0.4, 0.4]])
 
+        print("f type", type(F0))
```
michalhabera: Maybe a leftover?

PR #227, opened 2018-11-01
Wrap SLEPc solver in python layer

michalhabera:


michalhabera:
Good points. Tests must be added.

I'd like to use petsc4py/slepc4py whenever possible, maybe we don't need this wrapper at all.

jhale:
Couple of comments:

1) If we are going all the way with a wrapped python interface, it should be as pythonic as possible. `get_` and `set_` methods are not very pythonic. An alternative option would be to use properties e.g. https://stackoverflow.com/questions/6618002/using-property-versus-getters-and-setters

2) I want to be able to pull out the underlying SLEPc4py EPS and use it with the raw interface.

Jack

PR #229, opened 2018-11-04
Simplify expression

michalhabera:
Simplifies Expression rapidly.

New `Expression`:
1. is made independent on `GenericFunction` - final step before removing `GenericFunction`,
2. cannot be initialized from C++ string. Thus no JITed expression,
3. is super simple - just compare old `dolfin/function/expression.py` and the new one...,
4. is finite element-less, it is just eval method evaluated at points, no need to attach any element to it.
5. Is C-style oriented, eval method is not overridden by inheritance, but attached as member variable function,
6. Is vectorise-able, eval method's signature is ready for evaluation at arbitrary number of points.

Basically, the new expression could be evaluated at point by calling the eval method or interpolated. For this reason I had to duplicate current interpolate call to accept Function and Expression. This could look redundant, but interpolation will be ideally handled by different codepath in later stages.

In C++:
~You have to subclass `Expression` and override virtual methods, the same as before.~
You have to attach eval method without inheritance. Similar to how tabulation kernel method is attached.

In Python:
No need to subclass. It has member variable that holds python method.


michalhabera:
Ad `eval_cell`:
The cell is just one cell, the one in which the expression is being evaluated. This is done for every cell during interpolation. User has ability to prepare a cell dependent expression as 
```
def expr(values, x, cell):
    values[;, 0] = cell.(...)
```
or cell independent expression
```
def expr(values, x, cell=None):
    values[;, 0] = 1.0
```

The reason for always defining callable method with 3 arguments is, that in such case one can do both `Expression.eval(values, x)` and `interpolate(Expression)` (which in fact calls `Expression.eval_cell(values, x, cell)`).

garth-wells:
The function is not vectorise-able if only one cell is passed. The number of cells needs to be equal to x.shape[0].

jhale:
I'll leave more feedback once Garth's comments have been addressed.

michalhabera:
Cross-cell vectorisation could be added in future if we decide (by making everything accept Eigen::Array of cells...). Cross-point vectorisation within the cell should work.

Don't want to obstruct the big picture behind this change...

michalhabera:
Just for completeness:

Interpolation of pure python expression is significantly slower than JITed expression. Apparently due to repeated python call overhead. On 200x200 mesh JITed expression is 50x faster. 
Python call overhead will be handled with 1. vectorised call and 2. fast interpolation of UFL expressions.

Pure python expression should be used for more complex computations where callback overhead is negligible. Once fast interpolation of UFL expression is in place trivial Expression will be replaced.

python/dolfin/fem/projection.py
```diff
@@ -19,7 +19,7 @@ def project(v,
             V=None,
             bcs=None,
             mesh=None,
-            function=None,
+            funct=None,
```
jhale: Seems unnecessary?

PR #230, opened 2018-11-04
Remove not tested demos

michalhabera:
These demos were not tested and probably won't work anyway. They showcase the same functionality repeatedly.

We should structure demos due to techniques used and take this as inspiration: https://www.dealii.org/9.0.0/doxygen/deal.II/Tutorial.html

I don't see a problem with removing them now, demos are stored in git history anyway if we decide to add something back.


jhale:
Fine for me.

PR #237, opened 2018-11-14
Allow user to pass options to Numba JIT, Improved Documentation

jhale:
@garth-wells It turns out that `numba.cfunc` and `numba.jit` take different kwargs, so they are explicitly typed into the decorator. So in the end there are no `**kwargs`!

Passing options to decorators requires the decomaker pattern, hence the increase in complexity over the old implementation. Brackets `()` are now required on the decorator regardless of whether arguments are passed.

The current documentation is ambiguous with respect to how data is laid out in the case of multiple cells. Hopefully this will be sorted out down the line when we have a more concrete design.


python/dolfin/function/expression.py
```diff
@@ -50,32 +78,38 @@ def __init__(self,
...
+            ``(values_p, x_p, cells_p, num_points, value_size, gdim, num_cells)``
+            1. ``values_p`` is a pointer to a row-major array of
+               ``PetscScalar`` of shape ``(num_points, value_size)``.
+               The function itself is responsible for filling ``values_p``
+               with the desired Expression evaluations. ``values_p`` is not
+               zeroed before being passed to the function.
+            2. ``x_p`` is a pointer to a row-major array of ``double`` of shape
+               ``(num_points, gdim)``. The array contains the coordinates
+               of the points at which the expression function should be evaluated.
+            3. ``cells_p`` is a pointer to an array of ``int`` of shape
```
chrisrichardson: What is the difference between double reverse ticks and single reverse ticks?

jhale: Good question. We are following the numpy style?

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

chrisrichardson: Yes. The difference between one and two ticks is not very clear.

chrisrichardson:
Why does it require the brackets on the decorator?
`numba` itself manages without. e.g. I can do:

```
@numba.jit
def a(x):
    return x

@numba.jit(nopython=True)
def b(x):
    return x
```

jhale:
I guess there is some Python magic going on there! I'll have a go...

python/dolfin/function/expression.py
garth-wells: It's rst - http://docutils.sourceforge.net/docs/user/rst/quickstart.html#text-styles.

Double back ticks for a fixed-width font.

jhale:
Fixed @chrisrichardson, now decoration works without `()` as well. There is test coverage on both code paths.

jhale:
I'll do a squash and merge on this as the commit history is a bit messy!

python/dolfin/function/expression.py
```diff
@@ -1,47 +1,79 @@
...
 
 import numba
 from petsc4py import PETSc
 
 from dolfin import cpp
 
 
-def numba_eval(func):
-    """Decorator for JITable eval method
+def numba_eval(*args, numba_jit_options: dict = {"nopython": True, "cache": True},
```
garth-wells: I don't think it will work if the function can't be compiled ('nopython' not true), so maybe cache is the only 'real' option?

michalhabera: Not sure here. More important is the `cfunc` decorator. 
I think numba JIT could also work without nopython, but is super slow.

jhale: I tweaked a test at 02c202b that shows that everything works happily with numba's (very slow) cfunc and jit object compilation mode.

https://numba.pydata.org/numba-doc/latest/glossary.html#term-object-mode

jhale: Ignore the above, talking total rubbish. I won't protect the (advanced) user from overriding nopython and/or setting forceobj. Numba throws an exception, which is the pythonic behaviour. Test in place with xfail on exception `numba.errors.TypingError`.

PR #241, opened 2018-11-14
Allow dynamic attributes in pybind11 wrapped dolfin::la::PETScMatrix and dolfin::la::PETScVector

francesco-ballarin:
This is backport of the last item in https://bitbucket.org/fenics-project/dolfin/pull-requests/424/add-a-few-pybind11-wrappers/diff , that was merged to DOLFIN master after a brief discussion.

A couple of libraries that I develop rely on this for DOLFIN. pyadjoint will need this as well (see e.g. https://bitbucket.org/dolfin-adjoint/pyadjoint/src/master/fenics_adjoint/types/genericmatrix.py )

### DIFF
--- a/python/src/la.cpp
+++ b/python/src/la.cpp
@@ -145,7 +145,7 @@ void la(py::module& m)
 
   // dolfin::la::PETScVector
   py::class_<dolfin::la::PETScVector, std::shared_ptr<dolfin::la::PETScVector>>(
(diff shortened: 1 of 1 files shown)


garth-wells:
This is fine, but longer term probably won't be needed because we will likely wrap `cpp.la.PETScVector`, etc, at the Python level in which case attached attributes will 'just work'.

PR #244, opened 2018-11-16
Less raw addresses in Numba Expression

michalhabera:


python/dolfin/function/expression.py
```diff
@@ -75,23 +76,23 @@ def eval(values, x, cell_idx, num_points, value_size, gdim, num_cells):
         return decorator
 
 
-class Expression(cpp.function.Expression):
+class Expression:
     def __init__(self,
-                 eval_func: int,
+                 eval_func: numba.ccallback.CFunc = None,
```
garth-wells: Why `None` by default?

python/dolfin/function/expression.py
```diff
@@ -112,7 +113,21 @@ def __init__(self,
...
+        if eval_func is not None:
+            self._eval_address = eval_func.address
+            self._cpp_object.set_eval(self._eval_address)
+
+    @property
+    def eval_address(self): # noqa
+        return self._eval_address
+
+    @eval_address.setter
+    def eval_address(self, value: int):
```
garth-wells: Can we do everything via the initialiser and drop this?

python/dolfin/function/expression.py
```diff
@@ -7,6 +7,7 @@
 from typing import Callable
 
 import numba
+import numba.ccallback
```
garth-wells: Is this needed? Fully qualified namespace is used below.

python/dolfin/function/expression.py
michalhabera: I wanted to avoid isinstance check in __init__

garth-wells: Use `try`?

python/dolfin/function/expression.py
michalhabera: Seems needed, numba.ccallback doesn't get imported with numba.

python/dolfin/function/expression.py
```diff
@@ -112,4 +114,8 @@ def __init__(self,
         shape: tuple
             Value shape.
         """
-        super().__init__(eval_func, shape)
+        self.shape = shape
+        self._eval_func = eval_func
+
+        eval_address = getattr(eval_func, "address", eval_func)
```
garth-wells: Is it necessary to use `getattr`?

python/dolfin/function/expression.py
```diff
@@ -114,20 +114,8 @@ def __init__(self,
         shape: tuple
             Value shape.
         """
-        self._cpp_object = cpp.function.Expression(shape)
-
         self.shape = shape
```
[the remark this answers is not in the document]
michalhabera: I think not needed at all

python/dolfin/function/expression.py
michalhabera: If supplied `eval_func` has `eval_func.address` then it comes from JIT, otherwise take it as address itself. Its just short alternative to try catch.

garth-wells: It's not very clear to me. Better longer and clearer than short and hard to follow.

michalhabera: Ok, thats true.

python/dolfin/function/expression.py
```diff
@@ -112,4 +114,11 @@ def __init__(self,
         shape: tuple
             Value shape.
         """
-        super().__init__(eval_func, shape)
+        self._eval_func = eval_func
```
garth-wells: Add comment 'Hold reference to eval function to avoid premature garbage collection'.

python/dolfin/function/expression.py
```diff
@@ -112,4 +114,11 @@ def __init__(self,
...
+
+        try:
```
garth-wells:
```python
try:
    self._cpp_object = cpp.function.Expression(eval_func.address, shape)
except: FoorError:
    self._cpp_object = cpp.function.Expression(eval_func, shape)
```

garth-wells:
Merging now. Can polish more later is needed.

PR #245, opened 2018-11-16
Revert back to inheritance pattern for C++ Expression.

garth-wells:


cpp/dolfin/function/Expression.h
```diff
@@ -94,35 +103,50 @@ class Expression
...
+  ///         The values at the point.
+  /// @param    x (Eigen::Ref<const Eigen::VectorXd>)
+  ///         The coordinates of the point.
+  /// @param    cell (mesh::Cell)
+  ///         The cell which contains the given point.
+  virtual void eval(Eigen::Ref<Eigen::Array<PetscScalar, Eigen::Dynamic,
+                                            Eigen::Dynamic, Eigen::RowMajor>>
+                        values,
+                    const Eigen::Ref<const EigenRowArrayXXd> x,
+                    const dolfin::mesh::Cell& cell) const;
```
michalhabera: Here should be vector of `Cell` to keep all arguments vectorise-able. Or vector of cell indices as in C signature.

garth-wells: That will come later (and will be an index rather than cell).

garth-wells: Actually, the whole cell business is broken in the context of vectorisation because we would need to make `x` a three-dimensional array to know which points belong to which cell.

michalhabera: Could we for non-mixed meshes just put them together and say `x[0:3,:]` is for first cell, `x[3:6, :]` for second etc. First cell would have index `cell_idx[0]`, second `cell_idx[1]`, etc

garth-wells: If we keep the cell id, it should have the same number of rows as x, and for each x give the cell. This is general and clear - no assumption in fixed number of points per cell.

PR #248, opened 2018-11-18
CFFI Expression test

michalhabera:
Shows how create Expression from custom C code.

### DIFF
--- a/python/test/unit/function/test_function.py
+++ b/python/test/unit/function/test_function.py
@@ -368,6 +368,7 @@ def test_numba_expression_address(V):
     f.interpolate(f1)
     assert (f.vector().get_local() == 1.0).all()
 
+
 @skip_if_complex
 def test_cffi_expression(V):
 
--- a/python/test/unit/function/test_function.py
+++ b/python/test/unit/function/test_function.py
@@ -7,6 +7,8 @@
 
 import math
 
+import cffi
+import importlib
 import numba
 import numpy
 import pytest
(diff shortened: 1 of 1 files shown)


michalhabera:
It is apparent here, how iteration "for all points" is not optimal for row major layout of `values` and `x` as we have it now.

Just to keep in mind.

python/test/unit/function/test_function.py
```diff
@@ -365,3 +367,52 @@ def expr_eval(values, x, cell_idx):
...
+
+    # Import the compiled kernel
+    kernel_mod = importlib.import_module(module)
+    ffi, lib = kernel_mod.ffi, kernel_mod.lib
+
+    # Get pointer to the compiled function
+    eval_ptr = ffi.cast("uintptr_t", ffi.addressof(lib, "eval"))
+
+    # Handle C func address by hand
+    ex1 = Expression(int(eval_ptr))
```
garth-wells: Is `int` required?

michalhabera: Oddly enough it is. `ffi.cast("uintptr_t")` cast to cffi type `<class '_cffi_backend.CData'>`. With your latest changes on how int is casted to std::function it just doesn't work.

The old way that we have in `set_cell_tabulate` (and we had in expression) accepts also these cffi uintptr_t types.

michalhabera: I see now, pybind tries to be clever and binds cffi uintptr to constructor with `std::function` directly. I'll update PR asap.

michalhabera: OK, accepting `std::function` in pybind11 argument is nice but dangerous.
It accepts `cffi uintptr_t` objects and then tries to call it as a method inside evaluation. On the other hand it allows to construct Expression directly on pure python method.

I propose to keep `int()` as is in this PR. Without the `int()` there is an exception, so no ill-defined behaviour could happen.

PR #249, opened 2018-11-19
Various improvements and adjustments to Dockerfile

jhale:
- Tried to split out 'mandatory' build/run dependencies from nice-to-have and documentation related ones.
- Build PETSc/SLEPc in one `RUN` layer and remove python build deps afterwards.
- Allowed override of PETSc/SLEPc build flags and debug mode through `--build-args`.
- Install tini init system in notebooks for proper clean up of processes (required by Jupyter).
- Script in `.circleci/` to manually trigger `build-and-push-docker-images` job on CircleCI.


Dockerfile
```diff
@@ -256,18 +282,30 @@ RUN git clone https://github.com/fenics/dolfinx.git && \
...
+
+ENTRYPOINT ["/tini", "--", "jupyter", "notebook", "--ip", "0.0.0.0", "--no-browser", "--allow-root"]
 
 ########################################
 
 FROM complex as notebook-complex
 LABEL description="DOLFIN-X (complex mode) Jupyter Notebook"
+
+ARG TINI_VERSION=v0.18.0 
+ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
```
garth-wells: Do we need this? It would be nice to keep everything as vanilla as possible.

jhale: Yes, see e.g. all of the official Jupyter notebook stacks:

https://github.com/jupyter/docker-stacks/blob/master/base-notebook/Dockerfile

jhale: It is also recommended by Docker themselves:

https://github.com/docker-library/official-images#init

Dockerfile
```diff
@@ -20,18 +21,20 @@
...
 #
 
-ARG BUILD_THREADS=1
 ARG PYBIND11_VERSION=2.2.4
 ARG PETSC_VERSION=3.10.2
 ARG SLEPC_VERSION=3.10.1
 ARG PETSC4PY_VERSION=3.10.0
 ARG SLEPC4PY_VERSION=3.10.0
 
+ARG BUILD_THREADS=1
```
garth-wells: Should this be `PETSC_BUILD_THREADS` since we only use it for PETSc/SLEPc.

DOLFIN build with ninja, which has some smarts for parallel builds.

jhale: Well, ninja allocates cpus + 2 threads. However, there is quite some criticism of that approach here: https://github.com/ninja-build/ninja/issues/1441 and https://github.com/ninja-build/ninja/pull/1399 Perhaps it is not so smart!

For optimal control it might be good to add BUILD_THREADS to ninja as well.

jhale: I think the correct solution is to let both ninja and PETSc do their default thing (no flags) and let BUILD_THREADS override -j if set. It shouldn't be 1 by default.

garth-wells: I would rather let the build systems do there own things, unless we're seeing a problem that needs fixing.

PR #254, opened 2018-11-23
Remove check on _compiled_form attribute in _create_cpp_form

francesco-ballarin:
Hi,
I believe that the condition showed in the diff is a leftover from previous implementations, because:
(i) I cannot find any other _compiled_form in the entire fenics/fenicsx codebase
(ii) if there was, it definitely could not be attached to cpp.fem.Form object, because cpp.fem.Form wrapper was declared without py::dynamic_attr()

Feel free to close in case I overlooked something else and the check is actually needed.

Francesco

### DIFF
--- a/python/dolfin/fem/assembling.py
+++ b/python/dolfin/fem/assembling.py
@@ -24,12 +24,6 @@ from dolfin.fem.form import Form
 def _create_cpp_form(form, form_compiler_parameters=None):
     # First check if we got a cpp.Form
     if isinstance(form, cpp.fem.Form):
-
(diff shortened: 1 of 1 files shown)


francesco-ballarin:
I do not think that the build-real failure is related to this. Is it possible to run again circleci?

garth-wells:
Could you merge the `master` branch into your branch? The test on your branch is failing due to an issue that has been fixed in `master`.

francesco-ballarin:
Rebased on master

PR #267, opened 2018-12-19
Fix wrong use of SparsityPattern::insert_global in SparsityPatternBuilder

### LINKED ISSUE #265: performance regression in sparsity pattern builder for the case diagonal=True
Hi,
the following DOLFIN code 
```
import time
from dolfin import *

mesh = UnitSquareMesh(MPI.comm_world, 200, 200)
W = FunctionSpace(mesh, "CG", 1)
u = TrialFunction(W)
v = TestFunction(W)
t = time.time()
A = assemble(inner(u, v)*dx), keep_diagonal=True)
print(time.time() - t)
```
prints 0.03378748893737793 on my workstation.

Running the following DOLFIN-X code
```
import time
from dolfin import *
from dolfin.fem import assemble

mesh = UnitSquareMesh(MPI.comm_world, 200, 200)
W = FunctionSpace(mesh, ("CG", 1))
u = TrialFunction(W)
v = TestFunction(W)
t = time.time()
A = assemble(inner(u, v)*dx)
print(time.time() - t)
```
prints 0.033252716064453125, but this is not equivalent to the DOLFIN code above because of the hardcoded keep_diagonal=false on https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfin/fem/utils.cpp#L228

I manually changed that variable to true, recompiled DOLFIN-X and run again the code. After 15 minutes I gave up and interrupted the process.
If I understand the code correctly is it true that
https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfin/fem/SparsityPatternBuilder.cpp#L209
is inserting a nonzero on the cartesian product rows \times rows, rather than on the diagonal only? 

If that is indeed the case, and this obviously an undesired regression, I can create a PR to fix it.

francesco-ballarin:
closes #265

### DIFF
--- a/cpp/dolfin/fem/SparsityPatternBuilder.cpp
+++ b/cpp/dolfin/fem/SparsityPatternBuilder.cpp
@@ -200,13 +200,11 @@ la::SparsityPattern SparsityPatternBuilder::build(
         bs * (diagonal_range - primary_range[0]));
     std::iota(indices.begin(), indices.end(), bs * primary_range[0]);
 
-    // const std::array<common::ArrayView<const PetscInt>, 2> diags
-    //     = {{common::ArrayView<const PetscInt>(indices.size(),
-    //                                                     indices.data()),
-    //         common::ArrayView<const PetscInt>(indices.size(),
-    //                                                     indices.data())}};
-    Eigen::Map<const EigenArrayXpetscint> rows(indices.data(), indices.size());
-    pattern.insert_global(rows, rows);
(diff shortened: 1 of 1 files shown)


francesco-ballarin:
> I think it would be best to deal with 'real' spaces at the end of this.

Let me also note that, as mentioned in issue #265, this affects also spaces which are not necessarily composed by 'real' elements

francesco-ballarin:
Closing as the corresponding issue has been closed

PR #311, opened 2019-01-29
Switch to spdlog

jhale:
There is a lot of tedious work to do, seems like a good point to get feedback.


cpp/dolfin/common/MPI.cpp
```diff
@@ -17,7 +20,8 @@ dolfin::MPI::Comm::Comm(MPI_Comm comm)
   {
     int err = MPI_Comm_dup(comm, &_comm);
     if (err != MPI_SUCCESS)
-      log::error("Duplication of MPI communicator failed (MPI_Comm_dup");
+      spdlog::error("Duplication of MPI communicator failed (MPI_Comm_dup)");
+      throw std::runtime_error("Duplication of MPI communicator failed (MPI_Comm_dup)");
```
jhale: The old design logged and threw. The new design requires separate steps. Thoughts?

cpp/dolfin/mesh/IntervalCell.cpp
```diff
@@ -26,9 +27,10 @@ std::size_t IntervalCell::num_entities(std::size_t dim) const
   case 1:
     return 1; // cells
   default:
-    log::dolfin_error("IntervalCell.cpp",
-                      "access number of entities of interval cell",
-                      "Illegal topological dimension (%d)", dim);
+    spdlog::error("IntervalCell.cpp: "
+                  "access number of entities of interval cell. "
+                  "Illegal topological dimension ({}).", dim);
+    throw std::invalid_argument();
```
jhale: Again, duplicate error text?

cpp/dolfin/common/MPI.cpp
garth-wells: I think it looks ok - bare in mind that the logger will (or should) be used for much more than just error logging.

Do we have any indication of what is 'good' practice?

chrisrichardson:
OK, I went through and changed everything to use spdlog. Seems OK, but I'm sure there are some rough edges to be tidied up.

PR #315, opened 2019-01-31
Fixes to FindPETSc.cmake

jamtrott:
Minor changes to FindPETSc.cmake to enable debugging and fix an issue where testing for a working PETSc installation would fail due to undefined MPI symbols.


cpp/cmake/modules/FindPETSc.cmake
```diff
@@ -140,7 +148,7 @@ int main()
     CMAKE_FLAGS
     "-DINCLUDE_DIRECTORIES:STRING=${CMAKE_REQUIRED_INCLUDES}"
     "-DLINK_LIBRARIES:STRING=${CMAKE_REQUIRED_LIBRARIES}"
-    LINK_LIBRARIES PETSC::petsc
+    LINK_LIBRARIES PETSC::petsc "${MPI_LINK_LIBRARIES}"
```
garth-wells: I'm wondering why PETSc isn't providing this information in the pkg-config file? Is there a risk that this could mix up MPI versions?

jamtrott: I'm afraid that I don't know enough about how PETSc and pkg-config are supposed to work, so I don't have a good answer to your question. But maybe the installation info below might help?

For what it’s worth, I built PETSc from the tag v3.10.1 with the following configure flags:

`./configure --prefix=/nix/store/1h02gy3cw9dgq7ln4qj9mmd2s159f37v-petsc --with-fc=0 --with-mpi-dir=/nix/store/2nmwhv0fn32hgq3ldw5mkmyc55c9z701-openmpi-3.1.0 --with-hdf5-dir=/nix/store/2srjk8105vb79lnp27166ip356vw48nh-hdf5-1.10.3`

(Those long, slightly awkward paths stem from the nix package manager, which I am using to install packages.)

This produces the following pkg-config file, PETSc.pc:
```
$ cat /nix/store/1h02gy3cw9dgq7ln4qj9mmd2s159f37v-petsc/lib/pkgconfig/PETSc.pc
prefix=/nix/store/1h02gy3cw9dgq7ln4qj9mmd2s159f37v-petsc
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=${prefix}/lib
ccompiler=/nix/store/2nmwhv0fn32hgq3ldw5mkmyc55c9z701-openmpi-3.1.0/bin/mpicc
cflags_extra=-Wall -Wwrite-strings -Wno-strict-aliasing -Wno-unknown-pragmas -fstack-protector -Qunused-arguments -fvisibility=hidden -g3
cflags_dep=-MMD -MP
ldflag_rpath=-Wl,-rpath,

Name: PETSc
Description: Library to solve ODEs and algebraic equations
Version: 3.10.1
Cflags:  -I${includedir}
Libs: -L${libdir} -lpetsc
Libs.private: -L/nix/store/2srjk8105vb79lnp27166ip356vw48nh-hdf5-1.10.3/lib -llapack -lblas -lgfortran -lhdf5_hl -lhdf5 -lstdc++ -ldl
```


In this case, MPI is linked with -lmpi, but the path to the library is not specified.

The pkg-config file for openmpi is:
```
$ cat /nix/store/2nmwhv0fn32hgq3ldw5mkmyc55c9z701-openmpi-3.1.0/lib/pkgconfig/ompi.pc
# Copyright (c) 2010 Cisco Systems, Inc.  All rights reserved.
# Copyright (c) 2016      IBM Corporation.  All rights reserved.
#
Name: Open MPI
Description: Portable high-performance MPI implementation
Version: 3.1.0
URL: http://www.open-mpi.org/
#
prefix=/nix/store/2nmwhv0fn32hgq3ldw5mkmyc55c9z701-openmpi-3.1.0
exec_prefix=${prefix}
includedir=${prefix}/include
libdir=${exec_prefix}/lib
pkgincludedir=${includedir}/openmpi
#
# Note that -lopen-pal and all the EXTRA_LIBS are only necessary when
# static linking (they're pulled in by libopen-rte.so's implicit
# dependencies), so only list these in Libs.private.
#
Libs: -L${libdir}    -lmpi
Libs.private: -lm -lz
#
Cflags: -I${includedir}
```

In any case, there is still the problem with specifying both `-DLINK_lIBRARIES` and `LINK_LIBRARIES`.  According to the CMake manual, the former is ignored. (See the second paragraph in the section on LINK_LIBRARIES at https://cmake.org/cmake/help/latest/command/try_run.html.)

PR #326, opened 2019-02-03
Interface to ffc cffi jit

chrisrichardson:
Fairly minor changes to call new cffi based JIT in ffc.
* need to revert CI config before merge.


garth-wells:
Looking at the run times on CircleCI, is this a bit slower than with dijitso?

PR #331, opened 2019-02-07
Add back an overloaded SparsityPatternBuilder::build which takes a SparsityPattern as first input argument

francesco-ballarin:
Signature of SparsityPatternBuilder::build was changed a few PRs ago. With this PR, I add back the old signature. I'll happily provide a link to a concrete example in which I need this.

### DIFF
--- a/cpp/dolfin/fem/SparsityPatternBuilder.cpp
+++ b/cpp/dolfin/fem/SparsityPatternBuilder.cpp
@@ -32,13 +32,23 @@ la::SparsityPattern SparsityPatternBuilder::build(
   std::array<std::shared_ptr<const common::IndexMap>, 2> index_maps
       = {{dofmaps[0]->index_map(), dofmaps[1]->index_map()}};
 
-
-  const std::size_t D = mesh.topology().dim();
-
   // FIXME: Should check that index maps are matching
 
   // Create empty sparsity pattern
   la::SparsityPattern pattern(comm, index_maps);
+  
+  // Build sparsity pattern and return
+  SparsityPatternBuilder::build(pattern, mesh, dofmaps,
+                                cells, interior_facets, exterior_facets);
+  return pattern;
+}
+//-----------------------------------------------------------------------------
+void SparsityPatternBuilder::build(
+    la::SparsityPattern& pattern, const mesh::Mesh& mesh,
+    const std::array<const fem::GenericDofMap*, 2> dofmaps, bool cells,
+    bool interior_facets, bool exterior_facets)
+{
+  const std::size_t D = mesh.topology().dim();
 
   // Array to store macro-dofs, if required (for interior facets)
   std::array<Eigen::Array<PetscInt, Eigen::Dynamic, 1>, 2> macro_dofs;
@@ -121,7 +131,5 @@ la::SparsityPattern SparsityPatternBuilder::build(
       }
     }
   }
-
-  return pattern;
 }
 //-----------------------------------------------------------------------------
--- a/cpp/dolfin/fem/SparsityPatternBuilder.h
+++ b/cpp/dolfin/fem/SparsityPatternBuilder.h
@@ -27,10 +27,16 @@ class GenericDofMap;
 class SparsityPatternBuilder
 {
 public:
-  // FIXME: Simplify
   /// Build sparsity pattern for assembly of given bilinear form
   static la::SparsityPattern
   build(MPI_Comm comm, const mesh::Mesh& mesh,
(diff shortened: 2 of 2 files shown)


garth-wells:
This doesn't follow the design that DOLFIN-X is using. Can you explain why it's needed?

francesco-ballarin:
Sure!

I am working on updating one of my libraries to DOLFIN-X, see
https://gitlab.com/multiphenics/multiphenics/tree/fenicsx

Back in 2017, the main goal of the library used to be (i) allow users to assemble block matrices/vectors, (ii) allow subdomain/boundary/interface restricted variables, as long as they are defined on the same mesh.
Current DOLFIN-X development effort has fully covered (i) now, but IMHO the combination (i)+(ii) is still a peculiar feature of multiphenics, at least for my understanding of the current DOLFIN-X. We have a users interested in (ii), and for future releases I am trying to port the library to DOLFIN-X.

For block matrix sparsity pattern generation, it is my understanding that DOLFIN-X generates sparsity patterns for each block (based on the dofmap of each block), and then combines them through
```
SparsityPattern::SparsityPattern(
    MPI_Comm comm,
    const std::vector<std::vector<const SparsityPattern*>> patterns)
```
In my case, instead, due to subdomain/interface restrictions, combining dofmaps is much more involved than the offsetting of DOLFIN-X, and I'd rather have a custom class, which inherits from GenericDofMap, that does all the hard work of discarding DOFs and offsetting the undiscarded ones.
(see https://gitlab.com/multiphenics/multiphenics/blob/fenicsx/multiphenics/fem/BlockDofMap.h)
Given an integer i, in my custom dofmap class I also have the capability to take a view that results in a child of GenericDofMap containing only the DOFs associated to the i-th block. This is not the same as the DofMap of the i-th function space, because (a) DOF numbering is the one of the resulting block matrix, (b) some DOFs have been discarded by the restriction process.

Therefore, rather than allocating several sub-sparsity patterns and combining them afterwards, I allocate directly the final sparsity pattern (on the index map of the whole block matrix), and build in it the sub-sparsity pattern for each (i, j) view pair. This used to be possible in DOLFIN, and up until a few weeks ago in DOLFIN-X as well. An excerpt of the resulting code would be
```
  // Get dof maps
  std::array<const GenericDofMap*, 2> dofmaps
      = {{a.block_function_spaces()[0]->block_dofmap().get(),
          a.block_function_spaces()[1]->block_dofmap().get()}};


  Timer t0("Build sparsity");

  // Get IndexMaps for each dimension
  std::array<std::shared_ptr<const IndexMap>, 2> index_maps
      {{dofmaps[0]->index_map(), dofmaps[1]->index_map()}};
      
  // Create sparsity pattern
  SparsityPattern pattern(mesh.mpi_comm(), index_maps);
  
  // Build sparsity pattern for each block
  for (std::size_t i = 0; i < a.block_size(0); i++)
  {
    for (std::size_t j = 0; j < a.block_size(1); j++)
    {
      const Form & a_ij(a(i, j));
      if (
        a_ij.integrals().num_integrals(FormIntegrals::Type::cell) > 0
          ||
        a_ij.integrals().num_integrals(FormIntegrals::Type::interior_facet) > 0
          ||
        a_ij.integrals().num_integrals(FormIntegrals::Type::exterior_facet) > 0
      )
      {
        std::array<const GenericDofMap*, 2> dofmaps_ij{{
          &a.block_function_spaces()[0]->block_dofmap()->view(i),
          &a.block_function_spaces()[1]->block_dofmap()->view(j)
        }};
        SparsityPatternBuilder::build(pattern,
                                      mesh, dofmaps_ij,
                                      (a_ij.integrals().num_integrals(FormIntegrals::Type::cell) > 0),
                                      (a_ij.integrals().num_integrals(FormIntegrals::Type::interior_facet) > 0),
                                      (a_ij.integrals().num_integrals(FormIntegrals::Type::exterior_facet) > 0)
                                     );
      }
    }
  }
  
  // Finalize sparsity pattern
  pattern.assemble();
  t0.stop();
```
(see https://gitlab.com/multiphenics/multiphenics/blob/fenicsx/multiphenics/fem/block_assemble.cpp)

I hope this shows my issue. I can also send you the link of the classes responsible for the assembly, if it helps or it is of interest.

Thanks!

garth-wells:
@francesco-ballarin Can you see if #333 would work for you?

francesco-ballarin:
Yes, it does, thanks. I will close this as soon as #333 is merged.

garth-wells:
Great. I did it a bit differently because we're aiming to avoid having more than one way of doing something unless absolutely essential.

PR #335, opened 2019-02-09
Fix projection, add test

michalhabera:
Fixes parallel bug in `assemble_system` (used in projection).

Added small test to integrate projected `x^2` over unit cube. This would pick the bug.


python/test/unit/fem/test_assembler.py
```diff
@@ -487,3 +487,17 @@ def monitor(ksp, its, rnorm):
...
+
+def test_projection():
+    mesh = dolfin.UnitCubeMesh(dolfin.MPI.comm_world, 4, 4, 4)
+    V = dolfin.function.FunctionSpace(mesh, ("CG", 1))
+
+    x = ufl.SpatialCoordinate(mesh)
+    expr = x[0] ** 2
+
+    f = dolfin.project(expr, V)
+    integral = dolfin.fem.assemble(f * dolfin.dx)
```
garth-wells: Can we make this `ufl.dx`?

python/test/unit/fem/test_assembler.py
```diff
@@ -487,3 +487,17 @@ def monitor(ksp, its, rnorm):
...
+
+    integral_analytic = 1.0 / 3
+    assert numpy.isclose(integral_analytic, integral, 1.0e-6)
```
garth-wells: Use `assert integral == pytest.approx(1.0/3.0, , rel=1e-6, abs=1e-12)`, making `rel` and `abs` appropriately tight.

python/dolfin/fem/assembling.py
```diff
@@ -114,7 +114,7 @@ def assemble_system(A_form,
 
     fem.assemble(b_tensor, b_form)
     fem.apply_lifting(b_tensor, [A_form], [bcs])
-    b_tensor.ghostUpdate(addv=PETSc.InsertMode.ADD, mode=PETSc.ScatterMode.REVERSE)
+    b_tensor.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
```
garth-wells: I think this is wrong. We want to accumulate ghost contributions onto the owning processes.

michalhabera: Hmm, if no BCS are specified (as is usually the case for projection) then ghost update with `ADD` will change the `b_tensor`. It will gather and accumulate values for ghost positions. This test will fail for ADD.

But you're right `INSERT` is also wrong.

garth-wells: Changing the `b_tensor` is required also for no bcs, and it should be `ADD`. 

What's wrong here is `fem.assemble(b_tensor, b_form)`. It should be `assemble_vector`.

michalhabera: Thanks, fixed now.

garth-wells:
@michalhabera could you merge `master` into this branch?

python/test/unit/fem/test_assembler.py
```diff
@@ -499,3 +499,17 @@ def monitor(ksp, its, rnorm):
...
+    V = dolfin.function.FunctionSpace(mesh, ("CG", 1))
+
+    x = ufl.SpatialCoordinate(mesh)
+    expr = x[0] ** 2
+
+    f = dolfin.project(expr, V)
+    integral = dolfin.fem.assemble(f * ufl.dx)
+
+    integral_analytic = 1.0 / 3
+    assert numpy.isclose(integral_analytic, integral, rtol=1.e-6, atol=1.e-12)
```
garth-wells: Is there a reason why `numpy.isclose` is required, rather than  `pytest.approx`?

michalhabera:
Bug fixed in https://github.com/FEniCS/dolfinx/pull/337. Good job @garth-wells 
This just adds the test.

PR #336, opened 2019-02-11
Use fem.assemble_vector rather than fem.assemble in solve.

francesco-ballarin:
As commented in #335. Fixes wrong results in parallel in demo_poisson for me.


garth-wells:
Taken care of in #337.

PR #348, opened 2019-03-11
Wrap dolfin.generation module

michalhabera:
Another bunch of python wrappers. Calling `dolfin.cpp.generation` should be avoided in user code from now on.

### DIFF
--- a/python/demo/elasticity/demo_elasticity.py
+++ b/python/demo/elasticity/demo_elasticity.py
@@ -55,7 +55,7 @@ def build_nullspace(V):
 # XDMFFile(MPI.comm_world, "../pulley.xdmf").read(mesh)
 
 # mesh = UnitCubeMesh(2, 2, 2)
-mesh = BoxMesh.create(
+mesh = BoxMesh(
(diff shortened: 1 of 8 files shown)


garth-wells:
Looks good.

I'm wondering if `BoxMesh`, etc, look like a class rather than a function that return a `Mesh`?

michalhabera:
On cpp side its a class with static-only methods. Looks like a namespace to me. I tried to keep it simple and not create python class for such trivial state-less object.

garth-wells:
@michalhabera we definitely don't want a class - just wondering if we need 'create' in the name, and whether we should use lower-case.

michalhabera:
I don't see "create" in the name. 
Lower-case names: `interval_mesh`, `unit_square_mesh` etc would be more appropriate, but `UnitSquareMesh` is pretty rooted in users minds.

If you want just lower-cased namespaced methods (similar to what is in assembler) I can change both c++ and python sides. I would do it in separate PR if you agree.

michalhabera:
@garth-wells I think this is OK for public interface. Happy to have it merged.

PR #351, opened 2019-03-20
Remove some redundant members from mesh::Mesh

garth-wells:
Remove some methods that were just shortcuts for others in `mesh::Mesh`.


chrisrichardson:
I am not sure this helps readability of code.

garth-wells:
It removes duplication in low level interfaces (C++), which is important.

Issue #339, opened 2019-02-11
Get Helmholtz demo working in real mode


garth-wells:
The Helmholtz demo is run only in complex mode. It would be good if it ran in both real and complex mode to pick any/most issues when running and testing in real mode.

IgorBaratta:
The complex-valued solution comes from the incoming wave and the absorbing boundary condition. We could make a small change on them in real mode. 

If no one is already working on this issue, I'm willing to start right away.

garth-wells:
@IgorBaratta would be great if you could take a look.

garth-wells:
@IgorBaratta have you had a chance to look at this?

michalhabera:
Seems resolved with https://github.com/FEniCS/dolfinx/pull/364

PR #364, opened 2019-03-25
Real mode - helmholtz demo

IgorBaratta:
Related to[ issue 339](https://github.com/FEniCS/dolfinx/issues/339 ) .

Demo created with manufactured solution.  The changes were only made on the source expression and the boundary part of the variational problem is omitted.

- Proposal 1: New file with the real-valued demo. The current demo has no physical meaning in real mode.
- Proposal 2: One file with real and complex cases.  The downside is that `@function.expression.numba_eval` complains about the type.


garth-wells:
Can the two cases share more code? It is good is they can be a close as possible so that running the real case picks up 95% of and issues that would appear in the complex case.

IgorBaratta:
> Can the two cases share more code? It is good is they can be a close as possible so that running the real case picks up 95% of and issues that would appear in the complex case.

I updated the demo to share more code between the two modes. Only the definition of source amplitude is different. Also removed the boundary source term and the complex robin BC (in this case the solution is not a plane wave anymore).  We can go back to the plane wave solution (in complex mode), but I can not see how to share more code in this case.

PR #367, opened 2019-03-28
Add ElementDofLayout class to handle layout of dofs on a cell.

garth-wells:
This change adds a class (`ElementDofMap`) that defined the layout of dofs on an element.

This will allow user to easily provide their own input to the `DofMapBuilder`. It also isolates calls to the UFC generated dofmap code in one place. An effect is dramatic simplification of the dofmap construction code.

What is still missing is a definition of the reference cell topology. At present is is implicit (which is not great). This change will come later (this PR is already big) and will lead to further simplifications.


cpp/dolfin/fem/DofMap.cpp
```diff
@@ -22,15 +23,13 @@ using namespace dolfin::fem;
 //-----------------------------------------------------------------------------
 DofMap::DofMap(std::shared_ptr<const ufc_dofmap> ufc_dofmap,
                const mesh::Mesh& mesh)
-    : _cell_dimension(-1), _ufc_dofmap(ufc_dofmap), _global_dimension(0),
-      _ufc_offset(-1)
+    : _cell_dimension(-1), _global_dimension(0), _ufc_offset(-1)
```
chrisrichardson: Do we still need _ufc_offset?

.circleci/config.yml
```diff
@@ -70,7 +70,7 @@ unit-tests-python-mpi: &unit-tests-python-mpi
   name: Run unit tests (Python, MPI)
   command: |
     cd python/test/unit
-    mpirun -n 3 python3 -m pytest .
+    mpirun -n 3 python3 -m pytest -s -v .
```
chrisrichardson: Revert

michalhabera:
This is basically intermediate wrapper of `ufc_dofmap`, so its not passed directly to `DofMapBuilder` anymore?

Short description of some motivation in background would help.

PR #360, opened 2019-03-24
Integration on subdomains

chrisrichardson:

PR #369, opened 2019-03-30
More integration on subdomains

michalhabera:
Easy continuation of https://github.com/FEniCS/dolfinx/pull/360

After this PR all combinations of integrals on subdomains of type {cell, exterior_facet} with {matrix, vector, scalar} integrals should work.

### DIFF
--- a/python/test/unit/fem/test_assemble_domains.py
+++ b/python/test/unit/fem/test_assemble_domains.py
@@ -11,6 +11,7 @@ import pytest
 import dolfin
 import ufl
 
+
 @pytest.fixture
 def mesh():
     return dolfin.generation.UnitSquareMesh(dolfin.MPI.comm_world, 10, 10)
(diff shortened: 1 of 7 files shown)


michalhabera:
Ok, wait with merge then, I'll add to all tests some coefficient function.

michalhabera:
Added "constant" P1 coeff to all tests @chrisrichardson . Green. I hope that picks few cases. I will work more on interior facet integrals in coming PR.

python/test/unit/fem/test_assemble_domains.py
```diff
@@ -6,35 +6,163 @@
...
+    # Their union is the whole domain
+    values[0] = 111
+    values[1] = 222
+    values[2:] = 333
+
+    dx = dolfin.Measure('dx', subdomain_data=marker, domain=mesh)
+
+    w = dolfin.Function(V)
+    w.vector().set(0.5)
+    w.vector().ghostUpdate(addv=PETSc.InsertMode.INSERT,
```
garth-wells: This should be avoided. Set the 'local form' of the ghosted vector to avoid communication step.

PR #370, opened 2019-04-03
Add reference cell topology class

garth-wells:
This adds reference cell topology information. It is likely to change, but for now it allows the 'closure dof' functions in the FFC generated code to be removed.

The cell topology class will soon get entity ordering functions for constructing dof maps for different cell types.

### DIFF
--- a/cpp/dolfin/fem/ElementDofLayout.cpp
+++ b/cpp/dolfin/fem/ElementDofLayout.cpp
@@ -116,25 +116,25 @@ ElementDofLayout::ElementDofLayout(
 
       if (dim == 2)
       {
-        std::cout << "--------------------------- " << std::endl;
+        // std::cout << "--------------------------- " << std::endl;
         entity_closure[{dim, entity}][2].insert(entity);
         CellType face_type = ReferenceCellTopology::entity_type(_cell, 2);
         const int num_edges = ReferenceCellTopology::num_edges(face_type);
-        if (entity == 3)
-          std::cout << "Num edges: " << num_edges << std::endl;
+        // if (entity == 3)
+        //   std::cout << "Num edges: " << num_edges << std::endl;
         for (int e = 0; e < num_edges; ++e)
         {
           const int edge_index = face_edges[entity][e];
-          if (entity == 3)
-            std::cout << "   edge index: " << edge_index << std::endl;
+          // if (entity == 3)
+          //   std::cout << "   edge index: " << edge_index << std::endl;
           entity_closure[{dim, entity}][1].insert(face_edges[edge_index][e]);
           for (int v = 0; v < 2; ++v)
           {
-            if (entity == 3)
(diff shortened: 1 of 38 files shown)


cpp/dolfin/fem/ReferenceCellTopology.h
```diff
@@ -0,0 +1,61 @@
...
+
+namespace fem
+{
+
+class ReferenceCellTopology
+{
+public:
+  static int dim(CellType cell_type);
+  static const int* num_entities(CellType cell_type);
+  static int num_vertices(CellType cell_type);
```
chrisrichardson: We don't need `num_vertices`, `num_edges` and `num_faces`, can just use `num_entities()[0]` etc.

cpp/dolfin/fem/ElementDofLayout.cpp
```diff
@@ -5,46 +5,82 @@
...
+    _cell = dolfin::CellType::hexahedron;
+  else
+    throw std::runtime_error("Ooops");
+
+  const int* num_entities = ReferenceCellTopology::num_entities(_cell);
+  assert(num_entities);
+
+  // Compute closure entities
+  // [dim, entity] -> closure{sub_dim, (sub_entities)}
+  std::map<std::array<int, 2>, std::map<int, std::set<int>>> entity_closure
```
chrisrichardson: Not sure the second `map` is needed, as sub_dim will be 0, 1, 2 etc. i.e. could just be a vector?

garth-wells: Sure - feel free to push a change.

cpp/dolfin/fem/GenericDofMap.cpp
```diff
@@ -217,13 +216,14 @@ void GenericDofMap::ufc_tabulate_dofs(
       const int num_entity_dofs = entity_dofs[d][i].size();
```
chrisrichardson: Maybe use a const reference for `entity_dofs[d][i]` here, as it comes up quite a few times below.

cpp/dolfin/fem/ReferenceCellTopology.cpp
```diff
@@ -0,0 +1,371 @@
...
+  case CellType::hexahedron:
+    return hexahedron;
+  default:
+    throw std::runtime_error("Unknown cell type.");
+  }
+
+  return nullptr;
+}
+//-----------------------------------------------------------------------------
+// const int* ReferenceCellTopology::get_entities(CellType cell_type, int d0,
```
chrisrichardson: delete commented out code

michalhabera:
This introduces lot of code and logic duplication with `CellType.h/cpp, IntervalCell.h/cpp, ...`
I think old classes should be either removed - or this logic be moved there.

PR #373, opened 2019-04-04
Use reference to pass ufc_form

chrisrichardson:
Stop using `shared_ptr` where it is no longer needed for ufc objects passed into dolfin.

### DIFF
--- a/cpp/demo/hyperelasticity/main.cpp
+++ b/cpp/demo/hyperelasticity/main.cpp
@@ -170,15 +170,15 @@ int main(int argc, char* argv[])
       MPI_COMM_WORLD, pt, {{8, 8, 8}}, mesh::CellType::Type::tetrahedron,
       mesh::GhostMode::none));
 
-  auto space = std::unique_ptr<ufc_function_space, decltype(free)*>(
-      hyperelasticity_functionspace_create(), free);
-
-  auto ufc_map = std::shared_ptr<ufc_dofmap>(space->dofmap(), free);
+  ufc_function_space* space = hyperelasticity_functionspace_create();
+  ufc_dofmap* ufc_map = space->dofmap();
   auto V = std::make_shared<function::FunctionSpace>(
       mesh,
       std::make_shared<fem::FiniteElement>(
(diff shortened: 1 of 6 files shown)


michalhabera:
So this makes cpp user hands dirty with raw pointer. What is the argument for that?
You could have passed by reference and still use smart pointers.

chrisrichardson:
There is always a raw pointer from UFC. The alternative is `std::unique_ptr<ufc_foo, decltype(std::free)*>(ufc_foo_create(), std::free);`. So, you still need to tell the "smart" pointer how to destroy the malloc-created object. I think it is better to just use `free`. We can wrap all this up in a utility function, perhaps, later.

cpp/demo/hyperelasticity/main.cpp
```diff
@@ -170,15 +170,15 @@ int main(int argc, char* argv[])
...
-
-  auto ufc_map = std::shared_ptr<ufc_dofmap>(space->dofmap(), free);
+  ufc_function_space* space = hyperelasticity_functionspace_create();
+  ufc_dofmap* ufc_map = space->dofmap();
   auto V = std::make_shared<function::FunctionSpace>(
       mesh,
       std::make_shared<fem::FiniteElement>(
           std::shared_ptr<ufc_finite_element>(space->element(), free)),
       std::make_shared<fem::DofMap>(*ufc_map, *mesh));
+  std::free(ufc_map);
```
garth-wells: Is this correct? Does `space->dofmap()` create a dofmap, or just point to the dofmap owned by `space`?

chrisrichardson: It creates it. I guess it should have create in the name, but that's an ffc/ufc issue.

michalhabera:
Ok, just some ufc objects are handled with smart (ufc_finite_element) and some with raw pointers, which is not consistent now. Would it make sense to keep just one approach?

chrisrichardson:
@michalhabera - it still makes sense to use `shared_ptr` on `ufc_finite_element`, and `ufc_coordinate_mapping` because they are stored in the appropriate dolfin classes (a situation I would also like to change, but that will take longer).

PR #383, opened 2019-04-12
Simplify FiniteElement

chrisrichardson:
Creates all sub-elements at instantiation, and stores `value_dimension` as a vector.


cpp/dolfin/fem/FiniteElement.cpp
```diff
@@ -19,17 +19,13 @@ FiniteElement::FiniteElement(const ufc_finite_element& element)
       _tdim(element.topological_dimension), _space_dim(element.space_dimension),
       _value_size(element.value_size),
       _reference_value_size(element.reference_value_size),
-      _value_rank(element.value_rank), _degree(element.degree),
-      _num_sub_elements(element.num_sub_elements),
-      _hash(common::hash_local(signature())),
-      _value_dimension(element.value_dimension),
+      _degree(element.degree), _hash(common::hash_local(signature())),
```
garth-wells: This could probably be `std::hash`.

chrisrichardson: Probably. There is more to do here - feel free to do (or just merge).

PR #387, opened 2019-04-13
Turn off clang brace check

chrisrichardson:
* turn off flag which is only set in clang, but not gcc with -Wall.


garth-wells:
I would rather just keep the compiler happy. Don't see the need for this change.

PR #397, opened 2019-04-17
Simplify dofmap collapsing and remove inconsistencies

garth-wells:
Avoids constructing new dofmap with new ordering when possible.


cpp/test/unit/common/IndexMap.cpp
```diff
@@ -0,0 +1,66 @@
...
+  //   else
+  //     throw std::runtime_error("Received unexpected data (2).");
+  // }
+
+} // namespace
+} // namespace
+
+TEST_CASE("Scatter using IndexMap", "[index_map_scatter]")
+{
+  CHECK_NOTHROW(test_scatter());
```
chrisrichardson: This file seems to be commented out?

garth-wells: I changed the interface, and it was too difficult to figure out what the test was doing.

chrisrichardson: So delete it?

garth-wells: Or you could add it back :)

garth-wells: Added back.

PR #412, opened 2019-04-25
Import conj, real and imag from UFL

IgorBaratta:
+ alphabetically sort imports


IgorBaratta:
> We're working away from importing loads of functions. The plan is to allow the user to manage imports.

Good to know.
I'll keep using ufl.conj in my forms for now.

PR #414, opened 2019-04-28
Add time argument to Expression

garth-wells:
When using Numba expressions, it was not possible to get a 'time' parameter. This adds a time parameter to Expression that a user can update outside of an eval function.

Also remove the `cell` argument. The design was inconsistent w.r.t evaluation at an arbitrary number of points. E.g., for multiple cells it was not defined which cell a point belonged to. Moreover, it was not used in any tests or demos. Can be re-designed later if needed.


garth-wells:
@michalhabera There is actually a very wide use case. For many problems - nonlinear and dependent - boundary conditions and/or source terms are parameterised in in terms of (pseudo) time.

PR #416, opened 2019-04-30
Add IPython to images

jhale:
I find the standard Python interpreter unusable after getting used to IPython.


Dockerfile
```diff
@@ -105,7 +105,7 @@ ENV PATH=/usr/local/gmsh-${GMSH_VERSION}-Linux64/bin:$PATH
 # Second set of packages are recommended and/or required to build documentation or run tests.
 RUN pip3 install --no-cache-dir mpi4py numba && \
     export HDF5_MPI="ON" && \
-    pip3 install --no-cache-dir --no-binary=h5py cffi decorator flake8 h5py pygmsh pytest pytest-xdist sphinx sphinx_rtd_theme
+    pip3 install --no-cache-dir --no-binary=h5py cffi decorator flake8 h5py ipython pygmsh pytest pytest-xdist sphinx sphinx_rtd_theme
```
garth-wells: Is ipython needed to run the tests are per the docstring?

jhale: Good point, I'll split these lines out.

garth-wells: It would be good to put ipython, etc, further down in the 'user; images to keep the dev environment images as clean as possible.

jhale: Makes sense.

What about gmsh and friends? Perhaps the potential GSoC student would add integration tests on meshio/pygmsh/DOLFINX compatibility. Then it may be desirable for those deps to be in the lower images.

PR #417, opened 2019-04-30
Add "loguru" logging system

chrisrichardson:
* Remove many `LOG(ERROR)` statements and throw exceptions instead.
* Add loguru logging, which is compatible with google glog (and many other clones)
* Add python bindings
* Compile time unaffected

More can be done, but this is a good start.


garth-wells:
Handy that different libraries use the same syntax.

garth-wells:
@chrisrichardson Could you turn on some log output for a demo, e.g `demo_cahn-hilliard.py`?

chrisrichardson:
Yes, I'll turn it on for some demos. Just tidying up some more right now.

PR #426, opened 2019-06-04
[WIP] Non-UFC Poisson demo

chrisrichardson:
* Create a demo for the 2D Poisson example that uses no UFC generated code
* Main change in dolfin: `FiniteElement` uses `std::function` instead of calling ufc directly


cpp/demo/poisson_noufc/main.cpp
```diff
@@ -0,0 +1,591 @@
...
+//
+// The implementation is split in two files: a form file containing the
+// definition of the variational forms expressed in UFL and a C++ file
+// containing the actual solver.
+//
+// Running this demo requires the files: :download:`main.cpp`,
+// :download:`Poisson.ufl` and :download:`CMakeLists.txt`.
+//
+//
+// UFL form file
```
garth-wells: Best to delete doc that is no longer relevant.

cpp/demo/poisson_noufc/main.cpp
```diff
@@ -0,0 +1,591 @@
...
+
+using namespace dolfin;
+
+// Then follows the definition of the coefficient functions (for
+// :math:`f` and :math:`g`), which are derived from the
+// :cpp:class:`Expression` class in DOLFIN
+//
+// .. code-block:: cpp
+
+class P1Element
```
garth-wells: I don't see the need for a class. Could use a namespace if you want to group functions.

cpp/demo/poisson_noufc/main.cpp
chrisrichardson: :+1:

chrisrichardson:
I'm now slightly holding off merging, as I am working on interface changes to `tabulate_tensor`...

garth-wells:
@chrisrichardson Is this ready now? If not, maybe you could change it to a work-in-progress PR? (There is an option for that now.)

chrisrichardson:
How do I change to WIP?

PR #427, opened 2019-06-04
Prevent crash during HDF5File::read_mesh with quad/hex elements in parallel

david-kamensky:
This fixes the problem identified in Issue 1000 from the Bitbucket tracker and prevents `read_mesh` from crashing in parallel with quad/hex meshes (for certain combinations of meshes and numbers of processors).  

However, unlike in the current stable version of DOLFIN on Bitbucket, this fix **does not** lead to correct behavior of `read_mesh` with with the parameter `use_partition_from_file=true`, due to the orthogonal issue that, in DOLFIN-X, `build_distributed_mesh` no longer checks whether an existing partition exists, and always re-partitions the mesh.  (I will open an issue for discussion of this separate problem.)


michalhabera:
Thanks @david-kamensky !

PR #428, opened 2019-06-04
Pad input point to 3d

michalhabera:
Function evaluation takes a point with true geometrical dimensions,
e.g. for mesh with gdim=2  we call `f([0.5, 0.5])`, thus the point needs padding for use in bounding box tree cell computation.


IgorBaratta:
Wouldn't it be necessary to initialize point with some value?
e.g.:
`Eigen::Vector3d point = Eigen::Vector3d::Zero();`

The macro `EIGEN_INITIALIZE_MATRICES_BY_ZERO` is not defined by default.

chrisrichardson:
Also, initialise outside loop. `x.row(i).size()` is better as `x.cols()` ? Which is also `gdim`?

michalhabera:
Fixed @IgorBaratta @chrisrichardson

Issue #425, opened 2019-06-03
Incorrect number of assumed integrals


michalhabera:
The following
```
import numpy
import matplotlib.pyplot as plt
import scipy.sparse

import dolfin
import ufl

mesh = dolfin.generation.UnitIntervalMesh(dolfin.MPI.comm_world, 5)

V = dolfin.function.FunctionSpace(mesh, ("CG", 1))
u = dolfin.TestFunction(V)
g = dolfin.Function(V)
v = dolfin.TrialFunction(V)

f = u * ufl.ds + u * g * ufl.dx
a = ufl.derivative(f, g, v)

A = dolfin.fem.assemble_matrix(a)
```

fails with `RuntimeError: No exterior facet integrals`.

FFC optimised `ds` integral away, while in dolfin we try to set a domain to it, since dolfin is seeing `0 * ds` - has incorrect knowledge about number of exterior integrals.

michalhabera:
A quick fix is to remove checks on number of integrals == 0 in FormIntegrals.cpp. Do you see any side effect, @chrisrichardson ?

michalhabera:
I think because subdomain data is acquired from from which was didn't go through analysis stage, https://github.com/FEniCS/dolfinx/blob/master/python/dolfin/fem/form.py#L34

michalhabera:
Fixed in https://github.com/FEniCS/dolfinx/pull/430

PR #430, opened 2019-06-04
Fix issue #425

michalhabera:
Fixes issue https://github.com/FEniCS/dolfinx/issues/425

### DIFF
--- a/cpp/dolfin/fem/Form.h
+++ b/cpp/dolfin/fem/Form.h
@@ -173,29 +173,6 @@ public:
                                                        const PetscScalar*,
                                                        const double*, int));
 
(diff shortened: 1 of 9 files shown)


michalhabera:
But how without calling FFC from dolfin? 
Let say FFC does its magic and identifies there are only 2 effective tabulate exterior facet integrals for which it generates code. Dolfin only knows about original number, e.g. 5 apparent integrals and assignes domains based on this knowledge, https://github.com/FEniCS/dolfinx/blob/master/python/dolfin/fem/form.py#L34

michalhabera:
Removed some unused header code.
For number of integrals == 0 the code now skips iteration over meshfunction to identify marker entities.

PR #431, opened 2019-06-08
Fixes for mesh re-ordering and geometry

garth-wells:
Geometry 'dofs' were not being updated when re-ordering a mesh, leading to incorrect results.

This fixes the case of affine (P1) geometry. High-order geometry needs carefully consideration (refactoring, really) and does not work at present. Re-ordering a non-affine simplex mesh will throw an error.

### DIFF
--- a/cpp/dolfin/fem/DofMapBuilder.cpp
+++ b/cpp/dolfin/fem/DofMapBuilder.cpp
@@ -339,8 +339,6 @@ DofMapStructure build_basic_dofmap(const mesh::Mesh& mesh,
   // Build dofmaps from ElementDofmap
   for (auto& cell : mesh::MeshRange<mesh::Cell>(mesh, mesh::MeshRangeType::ALL))
   {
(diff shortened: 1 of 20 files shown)


PedroJJimenez:
Could this issue explain crappy results working with N1curl elements (combined space with Lagrange for Maxwell's equations in 2D) in a mesh imported from gmsh and converted with meshio? I have tried other xdmf meshes like the one in the stokes demo and everything else seems ok. Also the out-of-plane component that is projected in nodal elements and decoupled in vacuum looks well even with the imported mesh.

garth-wells:
@PedroJJimenez Yes, you need to call `mesh.Ordering.order_simplex(mesh)` before build a dofmap on the mesh.

We'd like to remove the need for this, but it is something that will require some thought and effort.

PedroJJimenez:
It works now! Many thanks and congrats for the great work here.

PR #439, opened 2019-06-13
Add function to read Information tag data in XDMF

iitrabhi:
This PR extends the functionality of `XDMFFile` interface with methods `read_information()` and `write_information()`. The methods are designed to read and write `<Information>` tag data of XDMF file format. The syntax is inline with XDMF standard and works well with PARAVIEW.

`read_information` method supports the following data types as element values in the XML tree structure:

- string
- double
- int
- size_t

It returns a map in C++ and a dictionary in python containing XML element-tag as `key` and XML element-value as `value`. 

`write_information` method is defined based on generic data type and thus supports all the above data types. As per the [XDMF schema](http://www.xdmf.org/index.php/XDMF_Model_and_Format#Information), the information is stored in CDATA inside the information element. Also, the XDMF reader of PARAVIEW enforces the use of “Name” and “Value” attributes. Thus, the information element is given the name “Information” and value is the number of children in the root node of CDATA or in other words the size of the map in C++ and dictionary in python. The method takes in a dictionary and appends the information to the XDMF file by writing it to the information tag.

Unit tests are defined to check the functionality of new methods.


francesco-ballarin:
I am guessing that this can be closed now, since XDMFFile has been overhauled completely?

michalhabera:
Yes, you can close this. We can think about Information tag in the scope of new XDMFFile interface a bit differently.

PR #441, opened 2019-06-14
Incorporate dijitso

chrisrichardson:


garth-wells:
I suggest just removing the dijitso dependency. See #443.

garth-wells:
Superseded by #443.

PR #451, opened 2019-06-21
Add KaHIP interface for mesh partitioning 

IgorBaratta:
- Create FIndKaHIP file to link KaHIP.
   - test the interface in serial
- Define KaHIP as an optional package
- Create the dolfin KaHIP interface 
- Add option to use the new interface  in `Partitioning.h`


garth-wells:
@IgorBaratta Could you generate some timing data for partitioning with KaHIP when the code is ready, but before merging? We can help with access to machines to test at large process counts if you need it.

IgorBaratta:
> @IgorBaratta Could you generate some timing data for partitioning with KaHIP when the code is ready, but before merging? We can help with access to machines to test at large process counts if you need it.

@garth-wells, sure, I can generate the timing. And I could definitely use some help to access some machines for testing.
I believe the machines I currently have access to are not particularly helpful for this task.

chrisrichardson:
@garth-wells - I think the timing results: 
 https://gist.github.com/IgorBaratta/038747f9aff44eb844c0913614bf0c03
are good enough that it is worth merging.

IgorBaratta:
I'm working on some tests on the branch [partition-py](https://github.com/FEniCS/dolfinx/tree/igor/partition-py) which will make easier to test this PR. And then i'll change the status to  "Ready for Review".

IgorBaratta:
No. It was installed directly on circleci (config.yml) for testing.
Should I add it to the Dockerfile?

garth-wells:
I think adding it to the Dockerfile would be good. It will get plenty of local testing with KaHIP installed, so best to have some testing of it on CircleCI.

IgorBaratta:
PR [#42](https://github.com/schulzchristian/KaHIP/pull/42) needs to be merged first.

jhale:
Could you pin the kahip pull against a particular version or even commit hash in the Dockerfile? Use the same strategy as an environment variable for the other packages.

IgorBaratta:
> Could you pin the kahip pull against a particular version or even commit hash in the Dockerfile? Use the same strategy as an environment variable for the other packages.

Good point. Probably better against a commit hash. But the other PR has to be landed first.

IgorBaratta:
KaHIP PR has been merged.
The kahip pull points to the commit  14be06c,  (make OpenMP an optional dependency for KaHIP).

I think it can be merged? @chrisrichardson @garth-wells

PR #458, opened 2019-07-02
Remove fem::GenericDofMap

garth-wells:
Remove pure base class `fem::GenericDofMap`.

Only one subclass, and with move to separate dofmap data from algorithms it not really required.


cpp/dolfin/fem/DofMap.h
```diff
@@ -230,6 +218,17 @@ class DofMap : public GenericDofMap
   /// Get dofmap array
   Eigen::Map<const Eigen::Array<PetscInt, Eigen::Dynamic, 1>> dof_array() const;
 
+  // FIXME: can this be removed?
```
IgorBaratta: Any reason why `tabulate_local_to_global_dofs` should be eventually removed?
I would like to state my interest in its maintenance since I use this information to construct my preconditioner.

garth-wells: It can be easily got from the `IndexMap`.

garth-wells:
@chrisrichardson `GenericDofMap::entity_dofs()` wasn't used anywhere.

PR #464, opened 2019-07-05
Simplify MeshFunction

michalhabera:
- removed MeshFunction on `bool`,
- removed many ununsed methods,
- replaced boost data container with `Eigen::Array`.

~Naming of underlying data array = `.values()` or other Python API details subject to discussion.~

Python API has `meshfunction.values` as a readonly (the reference is readonly) property now.


cpp/dolfin/io/HDF5File.cpp
```diff
@@ -469,25 +469,11 @@ void HDF5File::write(const mesh::MeshFunction<bool>& meshfunction,
   // so copy to int with values 1 and 0
   mesh::MeshFunction<int> mf(mesh, cell_dim, -1);
   for (auto& cell : mesh::MeshRange<mesh::MeshEntity>(*mesh, cell_dim))
-    mf[cell.index()] = (meshfunction[cell.index()] ? 1 : 0);
+    mf.values()[cell.index()] = (meshfunction.values()[cell.index()] ? 1 : 0);
```
garth-wells: Avoid calling `mf.values()` inside loops (quite a few occurrences of this in the PR).

In this particular case, `mf.values() = meshfunction.values()` might even work, or some other Eigen syntax that doesn't need a loop.

cpp/dolfin/io/VTKFile.cpp
```diff
@@ -192,7 +192,7 @@ void mesh_function_write(T& meshfunction, const std::string filename,
 
   // Write data
   for (auto& cell : mesh::MeshRange<mesh::MeshEntity>(mesh, cell_dim))
-    fp << meshfunction[cell.index()] << " ";
+    fp << meshfunction.values()[cell.index()] << " ";
```
garth-wells:
```
fp << meshfunction.values();
```
might work here.

cpp/dolfin/mesh/MeshValueCollection.h
```diff
@@ -197,18 +197,18 @@ MeshValueCollection<T>::MeshValueCollection(
   // Handle cells as a special case
   if ((int)D == _dim)
   {
-    for (std::size_t cell_index = 0; cell_index < mesh_function.size();
+    for (std::size_t cell_index = 0; cell_index < mesh_function.values().size();
```
garth-wells: Should this be `Eigen::Index` rather than `std::size_t`?

python/src/mesh.cpp
```diff
@@ -354,43 +354,16 @@ void mesh(py::module& m)
...
-      .def("set_all", [](dolfin::mesh::MeshFunction<SCALAR>& self,             \
-                         const SCALAR& value) { self = value; })               \
-      .def("where_equal", &dolfin::mesh::MeshFunction<SCALAR>::where_equal)    \
-      .def("array",                                                            \
-           [](dolfin::mesh::MeshFunction<SCALAR>& self) {                      \
-             return Eigen::Map<Eigen::Array<SCALAR, Eigen::Dynamic, 1>>(       \
-                 self.values(), self.size());                                  \
-           },                                                                  \
-           py::return_value_policy::reference_internal)
+      .def("values", py::overload_cast<>(                                      \
```
garth-wells: Should we make this a property?

cpp/dolfin/io/HDF5File.cpp
michalhabera: I moved `.values()` call outside of hot loop now.

PR #465, opened 2019-07-05
Remove unused comm parameter from read_mesh

IgorBaratta:
`HDF5File` and `XDMFFile` already have their communicator (passed for the constructor).
In `XXXXX::read_mesh` the communicator parameter was not used.

Removing the comm parameter avoids ambiguity if one uses more than 1 communicator for other tasks.


garth-wells:
What if I want to read on one communicator and create the `Mesh` in another?

chrisrichardson:
Igor has a branch (issue9) which breaks up the process. Xdmf should be about i/o not mesh creation.

PR #491, opened 2019-07-25
Remove mesh entity subclasses

garth-wells:
Work with dimension rather than shortcut names. Simplifies class hierarchy, removes need for templated iterators and ranges, and provides 'one' way of doing something.


garth-wells:
@michalhabera yes, that's the idea. *If* we want shortcuts, they're more easily done at the Python level.

PR #495, opened 2019-08-07
Remove std::move, not needed here

chrisrichardson:


cpp/dolfin/refinement/ParallelRefinement.cpp
```diff
@@ -287,7 +287,7 @@ mesh::Mesh ParallelRefinement::partition(bool redistribute) const
 
   mesh::DistributedMeshTools::init_facet_cell_connections(mesh);
```
garth-wells: This feels a bit awkward - is it just for distinguishing between interior and exterior facets?

chrisrichardson: It's in several places wherever a mesh is constructed in parallel. It is needed for the shared facets between processes for dofmap construction, as well as finding exterior facets.

PR #500, opened 2019-08-22
Zero tensor before calling tabulate, when needed.

chrisrichardson:
* For use with ffc version without `tensor_init_mode`. 
* All tensors "accumulate". 
* Tensors must be cleared in dolfin before calling `tabulate_tensor`.


chrisrichardson:
Need to reset CI config before merge

Issue #228, opened 2018-11-02
Reimplement Constant


michalhabera:
Constant was removed and must be reimplemented.

Underlying idea must be discussed in more detail before reimplementation.

michalhabera:
Resolved with #501

PR #501, opened 2019-08-25
Add new Constant support

michalhabera:
Based on FFCx https://github.com/FEniCS/ffcx/tree/michal/add-constant-terminal and UFL https://bitbucket.org/fenics-project/ufl/branch/michal/ufl-constant
Addresses https://github.com/FEniCS/dolfinx/issues/228

Constant with no finite element
----

New `ufl.Constant` is a true UFL's Terminal object without finite element. FFCx now generates tabulation kernel with signature with additional argument `PetscScalar* c`
```
tabulate_tensor(PetscScalar* A, const PetscScalar* w, const PetscScalar* c, ...)
```

The array `c[]` is prepared **outside** main assembly loop by merging values for all constants present in the Form. Please note, that all constants for the Form are prepared, even though only constants effectively not eliminated in integral (or equivalently, used inside tabulate_tensor body) could be used. However, the extra little cost (doesn't depend on mesh size) allows much simpler constants handling.

`dolfin::function::Constant` serves as a memory managing object, it stored a row-major flattened array of values.

As a bonus, the additional argument to tabulation kernel could have more general use in future - it could handle any non-finite element data that has to be passed.


python/dolfin/fem/form.py
```diff
@@ -62,6 +62,13 @@ def __init__(self, form: ufl.Form, form_compiler_parameters: dict = None):
             self._cpp_object.set_coefficient(
                 i, original_coefficients[j]._cpp_object)
 
+        # Constants are set based on their position in original form
+        original_constants = []
+        for c in form.constants():
+            original_constants.append(c._cpp_object)
```
chrisrichardson: This could be a list comprehension?

cpp/demo/poisson/main.cpp
```diff
@@ -189,6 +190,10 @@ int main(int argc, char* argv[])
       [](auto values, auto x) { values = Eigen::sin(5 * x.col(0)); });
   L->set_coefficients({{"f", f}, {"g", g}});
 
+  // Prepare and set Constants for the bilinear form
+  std::shared_ptr<function::Constant> kappa(new function::Constant({2.0}));
```
garth-wells: Can we use `auto kappa = std::make_shared<function::Constant>(2.0);`?

cpp/dolfin/fem/Form.cpp
```diff
@@ -74,6 +81,40 @@ int Form::original_coefficient_position(int i) const
   return _coefficients.original_position(i);
 }
 //-----------------------------------------------------------------------------
+void Form::set_constants(
+    std::map<std::string, std::shared_ptr<function::Constant>> constants)
```
garth-wells: `constants` should be `const`, at least for the shared pointer.

cpp/dolfin/fem/Form.cpp
```diff
@@ -74,6 +81,40 @@ int Form::original_coefficient_position(int i) const
...
+{
+  for (auto const& constant_in : constants)
```
garth-wells: `constant_in` -> `constant`.

`const` should be not be necessary following `auto` if function argument is made `const`.

cpp/dolfin/fem/Form.cpp
```diff
@@ -74,6 +81,40 @@ int Form::original_coefficient_position(int i) const
   return _coefficients.original_position(i);
 }
 //-----------------------------------------------------------------------------
+void Form::set_constants(
```
garth-wells: Is there a case where we could pass `Constant`s to the constructor,  and then call `set_constants`? If no, the check in this function would be much simpler.

cpp/dolfin/fem/Form.h
```diff
@@ -72,6 +73,9 @@ class Form
   Form(const std::vector<std::shared_ptr<const function::FunctionSpace>>&
```
garth-wells: Argument list is getting quite long - some documentation explaining the arguments would be helpful.

cpp/dolfin/fem/Form.h
```diff
@@ -123,6 +127,26 @@ class Form
   ///         coefficients.
   int original_coefficient_position(int i) const;
 
+  /// Set constants based on their names
+  ///
+  /// This method is used in command-line workflow, when users set
+  /// constants to the form in cpp file.
+  ///
+  /// Names of the constants must agree with their names in UFL file.
+  void set_constants(
```
garth-wells: These set functions are confusing given that the Constructor takes a constants argument.

cpp/dolfin/fem/Form.h
```diff
@@ -187,6 +210,15 @@ class Form
   /// Access form integrals (const)
   const FormIntegrals& integrals() const;
 
+  // Access constants (non-const)
```
garth-wells: Explain what the vector index gives, i.e. explain why this is a `std::vector` rather than a `std::map`.

python/src/function.cpp
```diff
@@ -140,5 +141,13 @@ void function(py::module& m)
...
       .def("tabulate_dof_coordinates",
            &dolfin::function::FunctionSpace::tabulate_dof_coordinates);
+
+  // dolfin::function::Constant
+  py::class_<dolfin::function::Constant,
+             std::shared_ptr<dolfin::function::Constant>>(
+      m, "Constant", "A value constant wrt. integration domain")
+      .def(py::init<std::vector<PetscScalar>>(),
+           "Create a constant from a scalar value array")
+      .def_readwrite("value", &dolfin::function::Constant::value);
```
garth-wells: Return a NumPy array instead?

python/dolfin/function/constant.py
```diff
@@ -0,0 +1,33 @@
...
+
+class Constant(ufl.Constant):
+    def __init__(self, domain, value):
+        np_value = numpy.asarray(value)
+        super().__init__(domain, np_value.shape)
+        self._cpp_object = dolfin.cpp.function.Constant(np_value.flatten())
+
+    @property
+    def value(self):
+        val = self._cpp_object.value
```
garth-wells: Would be cleaner to handle type in the pybind11 layer.

python/dolfin/function/constant.py
```diff
@@ -0,0 +1,33 @@
...
+        if (len(val) == 1):
+            return val[0]
+        else:
+            return numpy.asarray(val).reshape(self.ufl_shape)
+
+    @value.setter
+    def value(self, val):
+        np_value = numpy.asarray(val)
+        if (np_value.shape != self.ufl_shape):
```
garth-wells: I prefer to avoid these checks and let Python fail in the 'standard' way.

cpp/dolfin/function/Constant.h
```diff
@@ -0,0 +1,29 @@
...
+namespace function
+{
+
+class Constant
+{
+public:
+  /// Initialise as a vector
+  ///
+  /// The vector is a row-major (C style) flattened value of the constant
+  explicit Constant(std::vector<PetscScalar> value) : value(value){};
```
garth-wells: Should be include the rank/shape? Would be more expressive.

Would be nice to support the common case of rank 0 (`PETScScalar`), 1 (`std::vector` or `Eigen::Vector`) and 2 (`Eigen::Matrix`).

cpp/demo/poisson/main.cpp
chrisrichardson: Yes, but only by adding another constructor to `Constant`. `kappa` is in general, vector valued, so makes sense to have it as `auto kappa = std::make_shared<function::Constant>({2.0});` but this doesn't work due to "initialiser_list" issues in C++.

python/dolfin/function/constant.py
chrisrichardson: I put it here so it can return different types, e.g. a float, or a n-dimensional numpy array. Not easy to do in C++ with `variant` etc. I'm not sure if it is even possible for n>2 (for example Cijkl tensor) - does pybind work with `Eigen::Tensor`?

cpp/dolfin/fem/Form.cpp
michalhabera: Indeed. In command-line workflow, a Form is constructed with vector of constants, but only their names are known at that point, pointers to the actual constants are `nullptr`. Later on, in user's code, `form.set_constants` is called to match the names to actual `Constant`s that user prepared.
This is exactly the same as what happens for `Coefficient`.

cpp/dolfin/fem/Form.h
michalhabera: As said above, construction of Form takes places without full knowledge about Constants (and Functions). Just later, user sets the constants and functions. What can I do to avoid the confusion?

cpp/demo/poisson/main.cpp
garth-wells: @chrisrichardson what do you mean by "is in general, vector valued,"?

cpp/dolfin/fem/Form.cpp
garth-wells: Perhaps it would be better to set all coefficients and constants post-construction for consistency.

python/dolfin/function/constant.py
garth-wells: No need to involve Eigen. Suggestion is to wrap the C++ data as a NumPy array in the pybdind11 layer.

chrisrichardson: It's nice to be able to return a `float` when there is just one value.

cpp/dolfin/fem/Form.h
```diff
@@ -210,13 +218,21 @@ class Form
...
-  // Access constants (non-const)
-  std::vector<std::pair<std::string, std::shared_ptr<function::Constant>>>&
+  /// Access constants (non-const)
+  ///
+  /// @return Vector of attached constants with their names.
+  ///         Names are used to set constants in user's c++ code.
+  ///         Index in the vector is the position of the constant in the
+  ///         original (nonsimplified) form.
+  std::vector<
+      std::pair<std::string, std::shared_ptr<const function::Constant>>>&
```
garth-wells: Do we need the non-const version?

cpp/dolfin/function/Constant.h
```diff
@@ -20,10 +20,36 @@ class Constant
...
   /// The vector is a row-major (C style) flattened value of the constant
-  explicit Constant(std::vector<PetscScalar> value) : value(value){};
+  Constant(std::vector<PetscScalar> value, std::vector<int> shape)
+      : value(value), shape(shape){};
+
+  /// Initialise with a scalar value
+  Constant(PetscScalar value) : value({value}), shape({1}){};
+
+  /// Initialise with a 1D or 2D Array
+  Constant(Eigen::Ref<Eigen::Array<PetscScalar, Eigen::Dynamic, Eigen::Dynamic,
```
garth-wells: Move implementation to `Constant.cpp`.

cpp/dolfin/function/Constant.h
```diff
@@ -20,10 +20,36 @@ class Constant
   /// Initialise as a vector
   ///
   /// The vector is a row-major (C style) flattened value of the constant
-  explicit Constant(std::vector<PetscScalar> value) : value(value){};
+  Constant(std::vector<PetscScalar> value, std::vector<int> shape)
+      : value(value), shape(shape){};
+
+  /// Initialise with a scalar value
+  Constant(PetscScalar value) : value({value}), shape({1}){};
```
garth-wells: Move implementation to `Constant.cpp`.

cpp/dolfin/function/Constant.h
```diff
@@ -20,10 +20,36 @@ class Constant
   /// Initialise as a vector
   ///
   /// The vector is a row-major (C style) flattened value of the constant
-  explicit Constant(std::vector<PetscScalar> value) : value(value){};
+  Constant(std::vector<PetscScalar> value, std::vector<int> shape)
```
garth-wells: Move implementation to `Constant.cpp`.

python/dolfin/function/constant.py
```diff
@@ -5,29 +5,43 @@
...
+
 import numpy
 
 import ufl
 import dolfin.cpp
 
 
 class Constant(ufl.Constant):
-    def __init__(self, domain, value):
+    def __init__(self, domain, value: typing.Union[numpy.ndarray, list, int, float]):
```
garth-wells: Is `int` misleading? Can we generalise `list` to a sequence?

PR #504, opened 2019-09-05
Misc Constant tidy-up

garth-wells:
Make Python behaviour consistent for all rank.

### DIFF
--- a/cpp/dolfin/function/Constant.cpp
+++ b/cpp/dolfin/function/Constant.cpp
@@ -12,7 +12,7 @@ using namespace dolfin;
 using namespace dolfin::function;
 
 //-----------------------------------------------------------------------------
-Constant::Constant(PetscScalar value) : value({value}), shape({1})
+Constant::Constant(PetscScalar value) : value({value})
 {
   // Do nothing
 }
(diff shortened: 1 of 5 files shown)


cpp/dolfin/function/Constant.cpp
```diff
@@ -1,39 +1,37 @@
...
-    shape.pop_back();
-
   // Copy data from Eigen::Array to flattened vector
-  Eigen::Map<Eigen::Array<PetscScalar, Eigen::Dynamic, Eigen::Dynamic,
-                          Eigen::RowMajor>>
-      v(value.data(), array.rows(), array.cols());
-  v = array;
+  for (int i = 0; i < c.rows(); ++i)
+    for (int j = 0; j < c.cols(); ++j)
+      value[i * c.cols() + j] = c(i, j);
```
chrisrichardson: This may miss some vectorisation optimisations in Eigen.

garth-wells: The costs of this is trivial.

python/src/function.cpp
```diff
@@ -147,7 +147,7 @@ void function(py::module& m)
              std::shared_ptr<dolfin::function::Constant>>(
       m, "Constant", "A value constant wrt. integration domain",
       py::buffer_protocol())
```
chrisrichardson: The `py::buffer_protocol` should have been removed earlier. We are not using it.

PR #510, opened 2019-09-10
Add a partitioning to python interface and some tests 

IgorBaratta:
- Python interface for reading mesh data: `XDMFfile.read_mesh_data(comm)`
 - add test: `test_read_mesh_data`
 - add `partition_cells` and `build_from_partition` to cpp.mesh
 - add new test for partitioning with a subset of processors

### DIFF
--- a/python/dolfin/io.py
+++ b/python/dolfin/io.py
@@ -206,6 +206,13 @@ class XDMFFile:
         mesh.geometry.coord_mapping = fem.create_coordinate_map(mesh)
         return mesh
 
+    def read_mesh_data(self, mpi_comm):
+        """
+        Read in mesh data:
+        Cell type, geometric points, topological cells and global cell indices.
+        """
+        return self._cpp_object.read_mesh_data(mpi_comm)
+
     def read_checkpoint(self, V, name: str,
                         counter: int = -1) -> function.Function:
         """Read finite element Function from checkpointing format
--- a/python/src/mesh.cpp
+++ b/python/src/mesh.cpp
@@ -19,6 +19,7 @@
 #include <dolfin/mesh/MeshQuality.h>
 #include <dolfin/mesh/MeshValueCollection.h>
 #include <dolfin/mesh/Ordering.h>
+#include <dolfin/mesh/PartitionData.h>
 #include <dolfin/mesh/Partitioning.h>
 #include <dolfin/mesh/Topology.h>
 #include <dolfin/mesh/cell_types.h>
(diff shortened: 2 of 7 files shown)


python/dolfin/io.py
```diff
@@ -206,6 +206,13 @@ def read_mesh(self, ghost_mode):
         mesh.geometry.coord_mapping = fem.create_coordinate_map(mesh)
         return mesh
 
+    def read_mesh_data(self, mpi_comm):
```
garth-wells: Not clear from the doc string what the return type is.

python/src/mesh.cpp
```diff
@@ -339,5 +340,33 @@ void mesh(py::module& m)
...
+             std::shared_ptr<dolfin::mesh::PartitionData>>(
+      m, "PartitionData", "PartitionData object")
+      .def("num_procs", &dolfin::mesh::PartitionData::num_procs);
+
+  // dolfin::mesh::Partitioning::partition_cells
+  m.def("partition_cells",
+        [](const MPICommWrapper comm, int nparts,
+           dolfin::mesh::CellType cell_type,
+           const Eigen::Ref<const dolfin::EigenRowArrayXXi64> cells,
+           std::string partitioner) {
```
garth-wells: Could this be an enum rather than a string?

python/src/mesh.cpp
```diff
@@ -339,5 +340,33 @@ void mesh(py::module& m)
...
+              comm.get(), nparts, cell_type, cells, partitioner);
+        });
+
+  m.def("build_from_partition",
+        [](const MPICommWrapper comm, dolfin::mesh::CellType cell_type,
+           const Eigen::Ref<const dolfin::EigenRowArrayXXi64> cells,
+           const Eigen::Ref<const dolfin::EigenRowArrayXXd> points,
+           const std::vector<std::int64_t>& global_cell_indices,
+           dolfin::mesh::GhostMode ghost_mode,
+           dolfin::mesh::PartitionData& cell_partition) {
```
garth-wells: Can this be `const`? We're trying to avoid 'out' arguments.

python/test/unit/io/test_XDMF.py
```diff
@@ -132,6 +132,29 @@ def test_save_and_load_3d_mesh(tempdir, encoding):
     assert mesh.num_entities_global(dim) == mesh2.num_entities_global(dim)
 
 
+@pytest.mark.parametrize("mesh_tdim", mesh_tdims)
```
garth-wells: Could the names be made clearer than `mesh_tdim` and `mesh_n`?

python/dolfin/io.py
IgorBaratta: Changed the docstring.
Should type hints for return also be added? 
eg:
`->typing.Tuple[cpp.mesh.CellType, numpy.ndarray, numpy.ndarray,                                         typing.List[int]]:`

garth-wells: Yes, that would be good.

If we use type hints, we should remove the types from the docstring to avoid duplication.

PR #516, opened 2019-09-13
Miscellaneous demo cleanup

chrisrichardson:
Cleans up some comments and old bits of code in a few demos.

### DIFF
--- a/python/demo/elasticity/demo_elasticity.py
+++ b/python/demo/elasticity/demo_elasticity.py
@@ -87,7 +87,6 @@ rho = 10.0
 # Loading due to centripetal acceleration (rho*omega^2*x_i)
 x = SpatialCoordinate(mesh)
 f = as_vector((rho * omega**2 * x[0], rho * omega**2 * x[1], 0.0))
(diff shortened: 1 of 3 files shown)


python/demo/elasticity/demo_elasticity.py
```diff
@@ -85,9 +85,9 @@ def boundary(x, only_boundary):
 rho = 10.0
 
 # Loading due to centripetal acceleration (rho*omega^2*x_i)
-# f = Expression(("rho*omega*omega*x[0]", "rho*omega*omega*x[1]", "0.0"),
-
-f = as_vector((0.0, 1.0E+10, 0.0))
+x = SpatialCoordinate(mesh)
+f = as_vector((rho * omega**2 * x[0], rho * omega**2 * x[1], 0.0))
+# f = as_vector((0.0, 1.0E+10, 0.0))
```
garth-wells: Delete

chrisrichardson: I was also wondering about bringing back the `pulley.xdmf` file. It is about 1.8M as ASCII xdmf...

PR #517, opened 2019-09-13
Topology surface entities

chrisrichardson:
Create a list of `bool` for entities which are on the surface of a mesh.

### DIFF
--- a/cpp/dolfin/mesh/Topology.cpp
+++ b/cpp/dolfin/mesh/Topology.cpp
@@ -150,10 +150,11 @@ std::vector<bool> Topology::surface_entity_marker(int dim) const
 
   std::vector<bool> marker(size(dim), false);
 
+  const int num_facets = size(tdim - 1);
   // Special case for facets
   if (dim == tdim - 1)
   {
-    for (int i = 0; i < size(tdim - 1); ++i)
+    for (int i = 0; i < num_facets; ++i)
     {
       if (connectivity_facet_cell->size_global(i) == 1)
         marker[i] = true;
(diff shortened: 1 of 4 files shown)


cpp/dolfin/mesh/Topology.cpp
```diff
@@ -135,6 +135,54 @@ const std::vector<std::int32_t>& Topology::cell_owner() const
...
+{
+  const int tdim = this->dim();
+
+  if (dim >= tdim or dim < 0)
+  {
+    throw std::runtime_error("Invalid entity dimension: "
+                             + std::to_string(dim));
+  }
+
+  std::shared_ptr<const Connectivity> connectivity_facet_cell
```
garth-wells: There should be a check that `connectivity_facet_cell` is not null. At least an assertion, but probably an exception should be thrown.

cpp/dolfin/mesh/Topology.cpp
```diff
@@ -135,6 +135,54 @@ const std::vector<std::int32_t>& Topology::cell_owner() const
...
+    for (int i = 0; i < num_facets; ++i)
+    {
+      if (connectivity_facet_cell->size_global(i) == 1)
+        marker[i] = true;
+    }
+    return marker;
+  }
+
+  // Get connectivity from facet to entities of interest (vertices or edges)
+  std::shared_ptr<const Connectivity> connectivity_facet_entity
```
garth-wells: This connectivity has already been fetched.

chrisrichardson: It hasn't, but it looks very similar. Previously it was `(tdim - 1, tdim)` and nw it is `(tdim - 1, dim)`.

cpp/dolfin/mesh/Topology.cpp
chrisrichardson: :+1:

PR #519, opened 2019-09-13
C++ doc string cleanup

garth-wells:


cpp/dolfin/mesh/PartitionData.cpp
```diff
@@ -16,7 +16,7 @@ PartitionData::PartitionData(
     : _offset(1)
 
 {
-  for (std::size_t i = 0; i != cell_partition.size(); ++i)
```
chrisrichardson: This line should be reverted

Issue #527, opened 2019-09-17
Valgrind detects uninitialised values


massimiliano-leoni:
With the code I am running, `valgrind` outputs about 3000 lines of complaints for uninitialized values. The first one reads
```
==29144== Conditional jump or move depends on uninitialised value(s)
==29144==    at 0x7050CE2: VecNorm_MPI (in /home/mleoni/opt/petsc/install/lib/libpetsc.so.3.11.3)
==29144==    by 0x705A0C5: VecNorm (in /home/mleoni/opt/petsc/install/lib/libpetsc.so.3.11.3)
==29144==    by 0x705AEEB: VecNormalize (in /home/mleoni/opt/petsc/install/lib/libpetsc.so.3.11.3)
==29144==    by 0x7908AD1: KSPGMRESCycle (in /home/mleoni/opt/petsc/install/lib/libpetsc.so.3.11.3)
==29144==    by 0x7909C85: KSPSolve_GMRES (in /home/mleoni/opt/petsc/install/lib/libpetsc.so.3.11.3)
==29144==    by 0x7942D7D: KSPSolve (in /home/mleoni/opt/petsc/install/lib/libpetsc.so.3.11.3)
==29144==    by 0x4FE0606: dolfin::la::PETScKrylovSolver::solve(_p_Vec*, _p_Vec*, bool) (dolfin/la/PETScKrylovSolver.cpp:95)
==29144==    by 0x4145B9: main (main.cpp:204)
==29144==  Uninitialised value was created by a heap allocation
==29144==    at 0x4C2FB0F: malloc (in /usr/lib/valgrind/vgpreload_memcheck-amd64-linux.so)
==29144==    by 0x4F1411A: handmade_aligned_malloc (Memory.h:88)
==29144==    by 0x4F1411A: aligned_malloc (Memory.h:164)
==29144==    by 0x4F1411A: conditional_aligned_malloc<true> (Memory.h:214)
==29144==    by 0x4F1411A: conditional_aligned_new_auto<double, true> (Memory.h:374)
==29144==    by 0x4F1411A: resize (DenseStorage.h:408)
==29144==    by 0x4F1411A: Eigen::PlainObjectBase<Eigen::Array<double, -1, -1, 1, -1, -1> >::resize(long, long) (PlainObjectBase.h:293)
==29144==    by 0x4F1E843: _init2<long, int> (PlainObjectBase.h:739)
==29144==    by 0x4F1E843: Array<long, int> (Array.h:181)
[...]

PR #528, opened 2019-09-17
Fix uninitialized values (addresses issue 527)

massimiliano-leoni:
Zeroing this vector solves issue https://github.com/FEniCS/dolfinx/issues/527 completely.

### DIFF
--- a/cpp/dolfin/function/FunctionSpace.cpp
+++ b/cpp/dolfin/function/FunctionSpace.cpp
@@ -317,7 +317,7 @@ EigenRowArrayXXd FunctionSpace::tabulate_dof_coordinates() const
   // Dof coordinate on reference element
   const EigenRowArrayXXd& X = element->dof_reference_coordinates();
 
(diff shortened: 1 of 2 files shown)


cpp/dolfin/function/FunctionSpace.cpp
```diff
@@ -186,6 +186,7 @@ void FunctionSpace::interpolate(
                                          1, std::multiplies<>());
   Eigen::Array<PetscScalar, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
       values(x.rows(), value_size);
+  values = 0;
```
garth-wells: This shouldn't be needed.

massimiliano-leoni: I just realized that the error was showing up because my `interpolate` function was not setting all the values of `values`. Now that I fixed that, the error no longer shows up. Feel free to decline this PR if this case is not relevant.

PR #537, opened 2019-09-24
Simplify C++ creation of FunctionSpace from UFC

chrisrichardson:


cpp/dolfin/fem/utils.h
```diff
@@ -106,5 +107,12 @@ get_constants_from_ufc_form(const ufc_form& ufc_form);
 std::shared_ptr<const fem::CoordinateMapping>
 get_cmap_from_ufc_cmap(const ufc_coordinate_mapping& ufc_cmap);
 
+/// Convenience typedef for a ufc_function_space_create function
+using ufc_fs_ptr = ufc_function_space* (*)(void);
```
garth-wells: I'd rather not use a typedef. I think it obfuscates what's going on and it isn't re-used.

PR #539, opened 2019-09-25
MPI cleanup

chrisrichardson:
Removes a lot of unused wrappers in `MPI.h` and simplifies some of the stuff which is there.
* Makes the `MPI::Comm` just a wrapped around the `dup`'d comm, and nothing more.
* Removes the `Table` reductions, and puts them in `Table.cpp`


python/src/common.cpp
```diff
@@ -215,22 +215,7 @@ void mpi(py::module& m)
...
-                  [](const MPICommWrapper comm, dolfin::Table value) {
-                    return dolfin::MPI::sum(comm.get(), value);
-                  })
-      .def_static("avg", [](const MPICommWrapper comm, dolfin::Table value) {
-        return dolfin::MPI::avg(comm.get(), value);
-      });
-}
+                  });
+
+} // namespace dolfin_wrappers
```
IgorBaratta: Misplaced namespace comment? (Also in line 133)

chrisrichardson: I think it was auto-generated by clang-format. I'll check.

PR #544, opened 2019-09-27
Simplify Function::eval interface

garth-wells:
This changes the `Function` `eval` interface to take (i) a list of points and (ii) a list of cells indices that contain each point. Points with negative cell indices are ignored (the bounding box search now returns a negative index when the cell containing a point is not found). This makes point evaluations in parallel much simpler.

The change makes what is happening behind `eval` explicit and removes a ton of opaque magic behind the scenes, e.g. cell searches, assumptions on what to do when a point is 'close' to a cell but not within it, etc. Now allowing an arbitrary number of points to be processed permits a number of performance optimisations and minimises dynamics memory allocations.

Example usage:
```
x = numpy.array(......)
tree = geometry.BoundingBoxTree(mesh, mesh.geometry.dim)
cells = geometry.compute_first_entity_collision(tree, mesh, x)
f = u.eval(x, cells)
```

The is scope for a further optimisation is the user passes in points that are groups by the cell that contains them.

### DIFF
--- a/cpp/dolfin/function/Function.cpp
+++ b/cpp/dolfin/function/Function.cpp
@@ -202,7 +202,7 @@ void Function::eval(
         "fem::CoordinateMapping has not been attached to mesh.");
   }
 
-  // Get elemet
+  // Get element
   assert(_function_space->element());
   const fem::FiniteElement& element = *_function_space->element();
   const int reference_value_size = element.reference_value_size();
@@ -259,7 +259,7 @@ void Function::eval(
     {
       for (int j = 0; j < value_size; ++j)
       {
-        // TODO: Find an Eigen shortcut fot this operation
+        // TODO: Find an Eigen shortcut for this operation
         u.row(p)[j] += coefficients[i] * basis_values(0, i, j);
       }
     }
--- a/cpp/dolfin/function/Function.cpp
+++ b/cpp/dolfin/function/Function.cpp
@@ -166,20 +166,19 @@ void Function::eval(
     throw std::runtime_error(
         "Number of points and number of cells must be equal.");
   }
+  if (x.rows() != u.rows())
+  {
+    throw std::runtime_error("Length of array for Function values must be the "
+                             "same as the number of points.");
+  }
 
   assert(_function_space);
   assert(_function_space->mesh());
   const mesh::Mesh& mesh = *_function_space->mesh();
   const int gdim = mesh.geometry().dim();
   const int tdim = mesh.topology().dim();
-
-  assert(x.rows() == u.rows());
   assert(_function_space->element());
   const fem::FiniteElement& element = *_function_space->element();
-
-  // Create work vector for expansion coefficients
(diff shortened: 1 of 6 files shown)


cpp/dolfin/function/Function.cpp
```diff
@@ -266,42 +202,65 @@ void Function::eval(
         "fem::CoordinateMapping has not been attached to mesh.");
   }
 
-  std::size_t num_points = x.rows();
-  std::size_t reference_value_size = element.reference_value_size();
-  std::size_t value_size = element.value_size();
-  std::size_t space_dimension = element.space_dimension();
+  // Get elemet
```
chrisrichardson: typo

cpp/dolfin/function/Function.cpp
```diff
@@ -266,42 +202,65 @@ void Function::eval(
...
+    const mesh::MeshEntity cell(mesh, tdim, cell_index);
+    restrict(cell, coordinate_dofs, coefficients.data());
+
+    // Compute expansion
+    for (int i = 0; i < space_dimension; ++i)
     {
-      for (std::size_t j = 0; j < value_size; ++j)
+      for (int j = 0; j < value_size; ++j)
       {
         // TODO: Find an Eigen shortcut fot this operation
```
chrisrichardson: typo

cpp/dolfin/function/Function.h
```diff
@@ -110,28 +110,18 @@ class Function
...
 
-  /// Evaluate at given point in given cell
+  /// Evaluate the Function at points
   /// @param[in] x The coordinates of the points
-  /// @param[in] cell The cell which contains the given point
-  /// @param[in,out] u The values at the points
+  /// @param[in] cells The index of the cell which contains each point.
+  ///            Negative indices are ignored.
+  /// @param[in,out] u The values at the points. Values are not computed
+  ///                  for points with a negative cel index. This
```
chrisrichardson: typos cel, corrext

python/dolfin/function.py
```diff
@@ -82,11 +81,7 @@ def ufl_evaluate(self, x, component, derivatives):
             # Scalar evaluation
             return self(*x)
 
-    def eval_cell(self, x: np.ndarray, cell, u):
-        u = self._cpp_object.eval_cell(x, cell, u)
-        return u
-
-    def eval(self, x: np.ndarray, bb_tree: cpp.geometry.BoundingBoxTree, u=None) -> np.ndarray:
+    def eval(self, x: np.ndarray, cells: np.ndarray, u=None) -> np.ndarray:
```
chrisrichardson: Add description of cells to docstring.

michalhabera:
Now you call `element.transform_reference_basis` etc per each point, why is this call not vectorised -- as it way called for all points just once in old code.

garth-wells:
The old code could only process points for one cell per call, which was a performance problem. There are some optimisations that haven't been implemented yet in the case that the points are passed in order by cell.

PR #549, opened 2019-09-28
Remove Function::restriction

garth-wells:
Remove `Function::restriction`, which obfuscated a simple operation and involved  a number of operations that should be taken outside of tight loops.

This PR improves the code in `Function` and `FunctionSpace`, and adds placeholder functions in the assembly code that can be removed later with more re-factoring of coefficients in the assembly code.

### DIFF
--- a/cpp/dolfin/fem/assemble_matrix_impl.cpp
+++ b/cpp/dolfin/fem/assemble_matrix_impl.cpp
@@ -20,6 +20,25 @@
 
 using namespace dolfin;
 
+namespace
+{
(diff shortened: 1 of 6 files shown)


michalhabera:
Is there are performance difference? As far as `restrict` was not specific to Function, just a restriction of vector this change is ok.

garth-wells:
@michalhabera It will be give a performance boost as it avoids function calls inside tight loops. It also has some unused function arguments that in cases will no longer need to be computed.

Issue #545, opened 2019-09-27
dS integrals cause coefficients to be calculated incorrectly, when combined with other integrals.


chrisrichardson:
The coefficients for a `Form` are stored in `FormCoefficients`, but they need to be restricted on cells during the assembly process. For integrals over `dS` they need to be restricted over two neighbouring cells.

Currently, this is done by obtaining all coefficients twice (once for each cell). 
e.g. for an integral `f*v*dx + g*v*dS` the coefficients are fetched `[f(cell0), g(cell0), f(cell1), g(cell1)]`. However, the generated code expects `[xxxx, g(cell0), g(cell1)]`. It is not clear how best to resolve this in DOLFIN, as it really should be fixed in `ffc` by tying coefficients more strongly to integrals, rather than to forms.

francesco-ballarin:
(moved from #536) Tests for this issue are available at https://github.com/francesco-ballarin/dolfinx/commit/2bb0ca88f7fa10c693c56eaa74a52dddd1232d9f

michalhabera:
Should be fixed via https://github.com/FEniCS/dolfinx/pull/552, please reopen if not.

PR #552, opened 2019-09-30
Fix coefficient offsets

michalhabera:
Fixes issue #545 , see also FFCx https://github.com/FEniCS/ffcx/pull/154

However, current approach is still suboptimal. We fetch all coefficients for the whole form for each integral - even though if the integral is using only one of them. Improvement will come as additional PR, this is mainly to fix the bugs discovered by @francesco-ballarin .

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ install-python-components: &install-python-components
   command: |
             pip3 install git+https://github.com/FEniCS/fiat.git --upgrade
             pip3 install git+https://github.com/FEniCS/ufl.git --upgrade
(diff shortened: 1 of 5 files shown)


michalhabera:
@francesco-ballarin Can you try now with proposed dolfinx and ffcx branches?

francesco-ballarin:
Hi, I am out of office this week and it might take me longer then usual to test, but I will let you know as soon as I am able to do that.
Can you cherrypick my commit from #545 and run through CI? They fully reproduce the errors that I get in my code, and should be enough to test for the correctness of the bugfix (and accept this pull request).
Thanks!

michalhabera:
@francesco-ballarin Your test are now in master, and passing. Thanks a lot for reporting.

PR #554, opened 2019-09-30
Test custom partition array

IgorBaratta:


python/src/mesh.cpp
```diff
@@ -374,5 +381,14 @@ void mesh(py::module& m)
               ghost_mode, cell_partition);
         });
 
-} // namespace dolfin_wrappers
+  m.def("compute_halo_cells",
```
garth-wells: Note sure what this means. We don't tend to use the "halo" terminology.

IgorBaratta: Changed the name to `ghost_cell_mapping`, since it computes the map between the index of shared (ghost) cells and the sharing process, for a given partition.
However, I'm not sure about the name because the ghost cells may change later, depending on the ghost mode (except if the mode is `shared_facet`).

garth-wells: @chrisrichardson is working on getting consistency into the ghost concepts, so this may be refined later.

Issue #555, opened 2019-10-01
Make interface of FormIntegrals and assemblers consistent


mrambausek:
Currently, FormIntegrals has a (kind of) getter/setter pair for tabulate_tensor functions, i.e.
`register_tabulate_tensor(, ...fptr_t ptr)` and `get_tabulate_tensor_function(...) -> const std::function<fptr_t> &`. From an API perspective I found it more consistent if all of `register_tabulate_tensor`, `get_tabulate_tensor_function` and also the private struct `Integral` consistently worked with either raw function pointers (which can be stored in vectors) or `std::function<fptr_t>`. The difference is that raw functions pointers do only allow for free functions and caption-less lambdas but not for any stateful lambdas of callable objects with state. Thus, they offer less flexibility. The implication is that "custom integrals" cannot rely on run-time information except for such provided by dolfin coefficients and constants. 

The assembler implementations only accept `std::function<fptr_t>`. Hence, although there is
no way to feed `std::function<fptr_t>` implementations of the `tabulate_tensor` signature into `Form`s only such wrappers are used internally. I propose to unify things and go for std::function<fptr_t> everywhere. However, there might be some performance implications when passing around function pointers between C, python and C++ which need to be considered/measured.

I am happy to work on this and also do some testing of possible performance issues. But since this not only about performance but also about API design and "user interface", I do not want to rush ahead in an uncoordinated manner without any consultations. So, if you agree, I'll go ahead and come up with a pull request.

chrisrichardson:
I agree there is some discrepancy in the interface at the moment. If you want to experiment, I guess you could change the interface for `register_tabulate_tensor` to take a `std::function`. 

Then, it would be good to see some actual examples, e.g. with capture, and see where we go from there.

mrambausek:
Good, then I'll go in that direction and try to come up with some example.

chrisrichardson:
:+1: - try to make something simple... it's not certain that this is the way to go. We may just use raw pointers throughout instead. But fine to experiment.

garth-wells:
Can this issue be closed? Is it resolved by #556?

garth-wells:
No activity for a long time, so closing.

PR #556, opened 2019-10-01
Take tabulate signature from ufc.h

michalhabera:
Based on https://github.com/FEniCS/ffcx/pull/155

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ install-python-components: &install-python-components
   command: |
             pip3 install git+https://github.com/FEniCS/fiat.git --upgrade
             pip3 install git+https://github.com/FEniCS/ufl.git --upgrade
-            pip3 install git+https://github.com/FEniCS/ffcx.git --upgrade
+            pip3 install git+https://github.com/FEniCS/ffcx.git@michal/typedef-tabulate-signature --upgrade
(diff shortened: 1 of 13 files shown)


chrisrichardson:
While we are here, can we standardise on either `kernel` or `fn` in the assemblers? It is confusing to have a mixture.

michalhabera:
I agree, it became visible just now, when obscure type is shortcut :) I don't have any opinion which is better.

chrisrichardson:
Maybe we could also consider Issue #555 - and change the interface to `register_tabulate_tensor` to use a `std::function`?

chrisrichardson:
> I agree, it became visible just now, when obscure type is shortcut :) I don't have any opinion which is better.

I guess `kernel` says more. `fn` is pretty meaningless.

michalhabera:
@chrisrichardson Renamed and addressed #555 . Only `std::function` is in our interfaces.

mrambausek:
Thanks for considering my suggestion in #555 ! Btw. is there a reason why only cell integrals can be set but no facet integrals?

michalhabera:
@mrambausek Just a missing setter. But we will most likely reorganise these setters a bit, there are currently 2 ways of setting tabulation kernel -- via Form and via FormIntegrals, it should be preferably just in one place.

michalhabera:
I guess I will cherry-pick set/get kernel methods name unification as a separate PR.

PR #557, opened 2019-10-04
Allow vector assembly to set and not just add

garth-wells:


cpp/dolfin/fem/assemble_vector_impl.cpp
```diff
@@ -400,6 +485,12 @@ void fem::impl::assemble_vector(
 
   for (int i = 0; i < integrals.num_integrals(type::interior_facet); ++i)
   {
+    if (mode != fem::InsertMode::sum)
+    {
+      throw std::runtime_error("Insert mode \"sum\" only is supported for "
```
chrisrichardson: Improve text? Maybe "is the only mode supported..."

PR #560, opened 2019-10-06
[WIP] Add interface to evaluate Function at points on reference element

garth-wells:
Given coordinates on the reference element, `Function::eval_reference` evaluates the function at the reference coordinates in all elements.


cpp/dolfin/function/Function.cpp
```diff
@@ -275,6 +275,72 @@ void Function::eval(
...
+  assert(_function_space->element());
+  const fem::FiniteElement& element = *_function_space->element();
+  const int reference_value_size = element.reference_value_size();
+  const int value_size = element.value_size();
+  const int space_dimension = element.space_dimension();
+
+  // Compute basis on reference element
+
+  // NOTE: This step needs to be generalised for elements other than
+  //       non-manifold Lagrange elements
```
michalhabera: You mean "non-affine"?

cpp/dolfin/function/Function.cpp
```diff
@@ -275,6 +275,72 @@ void Function::eval(
...
 }
 //-----------------------------------------------------------------------------
+void Function::eval_reference(
+    const Eigen::Ref<
+        const Eigen::Array<double, Eigen::Dynamic, 3, Eigen::RowMajor>>& X,
+    Eigen::Ref<Eigen::Array<PetscScalar, Eigen::Dynamic, Eigen::Dynamic,
+                            Eigen::RowMajor>>
+        u) const
+{
+  // if (X.rows() != u.rows())
```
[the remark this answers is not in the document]
garth-wells: Fixed.

cpp/dolfin/function/Function.cpp
garth-wells: No, I meant manifolds where `tdim != `gdim` are not supported. I've added a check that throws an exception if `tdim != `gdim`.

garth-wells:
Overtaken by new `basix` library.

PR #562, opened 2019-10-06
Make expression functions used in interpolation pure

garth-wells:
This changes expression interpolation to:
```
def f(x):
    return x[:, 0]
u.interpolate(f)
```
or
```
u.interpolate(lambda x: x[:, 0])
```
There is no new issue with the allocation of memory for the return argument as this was being allocated internally at each call anyway.

This change makes the user more aware of the shape of the return argument (the shape of which which gets checked), reducing the likelihood of an error that leaves the expression values only partially filled. It also allows the use of lambda functions from Python.


cpp/dolfin/function/FunctionSpace.h
```diff
@@ -93,9 +93,24 @@ class FunctionSpace
...
+          const Eigen::Ref<const Eigen::Array<
+              double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>&)>& f)
+      const;
+  /// @endcond
+
+  /// Interpolate expression into function space, returning the vector
+  /// of expansion coefficients
+  /// @param[in] expansion_coefficients The expansion coefficients
+  /// @param[in] f The function to be interpolated
+  void interpolate_c(Eigen::Ref<Eigen::Matrix<PetscScalar, Eigen::Dynamic, 1>>
```
michalhabera: What does _c stands for? Not very clear.

garth-wells: The idea is that it's not really in the public interface. The `_c` was added to indicate this, alluding to its use for interfacing with compiled C functions coming from Numba/CFFI.

cpp/doc/Doxyfile
```diff
@@ -892,7 +892,7 @@ EXCLUDE_SYMLINKS       = NO
 # Note that the wildcards are matched against the file with absolute path, so to
 # exclude all test directories for example use the pattern */test/*
 
-EXCLUDE_PATTERNS       =
+EXCLUDE_PATTERNS       = *.cpp
```
chrisrichardson: This should not be needed. If there is a specific region of one cpp file that is causing a problem, it can be wrapped in `\\@cond` - `\\@endcond`.

garth-wells: I'm not so keen on changing the DOLFIN source code to work around bugs in old versions of doxygen. I'd rather localise work-arounds in the doxygen files.

chrisrichardson: But this is not local: `*.cpp`

PR #563, opened 2019-10-07
Order vertices using global sharing information

chrisrichardson:
Constructs the local-to-global map for vertices *after* distributing points, as this allows them to be ordered into three groups: local (unshared+owned), shared (owned) and ghost (shared+unowned).

This will form the first part of some changes that will see edges and facets also renumbered so that these ghost entities are also defined in the same way.


cpp/dolfin/mesh/Mesh.cpp
```diff
@@ -48,79 +48,157 @@ Eigen::ArrayXd cell_r(const mesh::Mesh& mesh)
...
-
-        // Insert (global_vertex_index, local_vertex_index) into map. If
-        // global index seen for first time, add to local-to-global map.
-        auto map_it = global_to_local.insert({q, local_to_global.size()});
-        if (map_it.second)
-          local_to_global.push_back(q);
-
-        // Set local node index in cell node list (applying permutation)
-        cell_nodes_local(c, cell_permutation[v]) = map_it.first->second;
+        // If lowest ranked sharing process is greather than this process,
```
garth-wells: I would like to isolate/encapsulate the ownership decision. Making the lowest rank the owner can lead to unbalanced partitions, and will make overlapping computation and communication possibly problematic.

chrisrichardson: Sure. The advantage of using "lowest rank is owner" is that it requires no communication. In any case, ownership is still quite a late decision in the logic of this PR, so this is something which can be easily revisited in the future.

francesco-ballarin:
Hi,
is it possible that this change requires an update to MeshFunction as well, with syncronization of nonlocal values?

I am able to loop over local and nonlocal vertices using `for (const auto& e : MeshRange(mesh, 0, MeshRangeType::ALL))` (while formerly I did not need to provide the third MeshRange argument), but it seems that accessing `vertex_mesh_function.values()[e.index()]' over nonlocal vertices always returns value 0, where `vertex_mesh_function` is integer MeshFunction on all vertices, which I mark from python assigning constant value > 0.

I can try and cook up a test (opening a corresponding issue) in the next few days if it helps.

chrisrichardson:
Yes, it is quite possible. Please go ahead.

PR #565, opened 2019-10-07
Support for abitrary order quadrilateral elements using VTK input format

jorgensd:
Arbitrary order quadrilateral mesh support assuming VTK input format as described at https://blog.kitware.com/modeling-arbitrary-order-lagrange-finite-elements-in-the-visualization-toolkit/

Minor change in FFCx needed for this to work (will revert circleci once that is merged).

We have kept the assumption that first order quadrilaterals follow a lexiographic ordering, needing special treatment in `Mesh.cpp`

Tests for CG2,3 and 4 meshes added in python layer.


cpp/dolfin/mesh/cell_types.cpp
```diff
@@ -511,17 +511,46 @@ std::vector<std::uint8_t> mesh::vtk_mapping(mesh::CellType type, int num_nodes)
...
+          for (int l = 2; l < i; ++l)
+            permutation[j++] = l * i + k;
+        assert(j == i * i);
+        return permutation;
+      }
     }
-    default:
-      throw std::runtime_error("Higher order quadrilateral not supported");
-    }
+    // case 9: (i = 3)
```
chrisrichardson: Delete commented lines

cpp/dolfin/mesh/cell_types.cpp
```diff
@@ -511,17 +511,46 @@ std::vector<std::uint8_t> mesh::vtk_mapping(mesh::CellType type, int num_nodes)
...
-    {
-    case 4:
-    {
+    if (num_nodes == 4)
       // Assumes mapping from lexiographic ordering, not counter-clockwise
       // which is the VTK format
       return {0, 1, 3, 2};
+    {
+      int i = sqrt(num_nodes);
+      if (i * i == num_nodes)
```
chrisrichardson: Catch the case where `i*i != num_nodes`

cpp/dolfin/mesh/cell_types.cpp
```diff
@@ -558,10 +587,13 @@ int mesh::cell_degree(mesh::CellType type, int num_nodes)
     else
       throw std::runtime_error("Higher order tetrahedron not supported");
   case mesh::CellType::quadrilateral:
-    if (num_nodes == 4)
-      return 1;
-    else
-      throw std::runtime_error("Higher order quadrilateral not supported");
+  {
+    int i = sqrt(num_nodes);
```
chrisrichardson: Maybe should be `std::sqrt` and `#include<cmath>`?

cpp/dolfin/mesh/cell_types.cpp
chrisrichardson: I guess it is caught below, but it looks a bit unclear.

cpp/dolfin/mesh/cell_types.cpp
```diff
@@ -558,10 +587,13 @@ int mesh::cell_degree(mesh::CellType type, int num_nodes)
...
+    if (num_nodes == i * i)
+      return i - 1;
+    throw std::runtime_error("Quadrilateral of order "
+                             + std::to_string(num_nodes) + " not supported");
```
chrisrichardson: Message could be better. If `num_nodes != i*i`, then it is not just the wrong order, it is a different element type, probably.

PR #567, opened 2019-10-07
Fix for hex meshes

jorgensd:
Yet another error introduced when I tried to remove "duplicate code". 
As hexes follow the same structure as quads, a lexiographic ordering, they need to use a separate map, which is not the same as the VTK map in the cell_permutations. 

Added additional test to test this for future rewrites.

### DIFF
--- a/python/test/unit/mesh/test_mesh.py
+++ b/python/test/unit/mesh/test_mesh.py
@@ -13,12 +13,14 @@ import pytest
 
 import dolfin
 import FIAT
-from dolfin import (MPI, BoxMesh, MeshEntity, MeshFunction, RectangleMesh,
+from dolfin import (MPI, BoxMesh, Constant, MeshEntity, MeshFunction, RectangleMesh,
                     UnitCubeMesh, UnitIntervalMesh, UnitSquareMesh, cpp)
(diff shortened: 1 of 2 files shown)


python/test/unit/mesh/test_mesh.py
```diff
@@ -536,3 +538,10 @@ def test_coords():
     d = mesh.coordinate_dofs().entity_points()
     d += 2
     assert numpy.array_equal(d, mesh.coordinate_dofs().entity_points())
+
+
+def test_UnitHexMesh_assemble():
+    mesh = UnitCubeMesh(MPI.comm_world, 6, 7, 5, CellType.hexahedron)
+    vol = assemble_scalar(Constant(mesh, 1) * dx)
```
michalhabera: Do not complicate here with Constant, `1.0 * ufl.dx(mesh)` should be enough...

chrisrichardson: I am also wondering if this is really a useful test. Does integrating a constant actually care about the ordering, for example.

jorgensd: This test is useful. If the ordering is wrong, the integral is not 1. For instance, without this fix the integral is 0.57.

PR #568, opened 2019-10-08
Rewrite of Mesh constructor and other mesh functionality

jorgensd:
- Additional constructor argument for mesh initialization is now the cell_permutation. Tested with sending in VTK mapping for first order quads, where we internally assume lexiographic ordering.
- Separated cell_permutation constructor from mesh constructor, calls function in cell_types.cpp which we can now modify at will.
- Fixed a bug with hex-meshes introduced a few PRs ago by merging vtk_mapping and cell_permutation.
- Added CG-4 triangle support
- Arbitrary order quadrilateral support
- Added python-layer mesh constructor so that we can pass through kwargs


cpp/dolfin/mesh/Mesh.cpp
```diff
@@ -227,25 +228,18 @@ Mesh::Mesh(
...
-  else
-    cell_permutation = mesh::vtk_mapping(type, cells.cols());
-
   // Find degree of mesh
   // FIXME: degree should probably be in MeshGeometry
   _degree = mesh::cell_degree(type, cells.cols());
 
+  // Permutation from Mesh input format to DOLFIN order for cell geometric nodes
+  std::vector<std::uint8_t> cell_permutation;
+  if (custom_permutation.size() == 0)
```
garth-wells: Use `if (custom_permutation.empty())`.

cpp/dolfin/mesh/Mesh.cpp
```diff
@@ -227,25 +228,18 @@ Mesh::Mesh(
...
+    cell_permutation = mesh::default_cell_permutation(type, _degree);
```
garth-wells: Is this needed?

cpp/dolfin/mesh/Mesh.h
```diff
@@ -90,14 +90,17 @@ class Mesh
   /// @param[in] ghost_mode The ghost mode
   /// @param[in] num_ghost_cells Number of ghost cells on this process
   ///                            (must be at end of list of cells)
+  /// @param[in] custom_permutation Mapping of DOLFIN/UFC vertex
```
garth-wells: I don't follow what the map is. Is it order of input to DOLFIN ordering? And should 'vertex' be 'node'?

cpp/dolfin/mesh/cell_types.cpp
```diff
@@ -511,17 +513,43 @@ std::vector<std::uint8_t> mesh::vtk_mapping(mesh::CellType type, int num_nodes)
...
-    {
-    case 4:
-    {
+  {
+    if (num_nodes == 4)
       // Assumes mapping from lexiographic ordering, not counter-clockwise
       // which is the VTK format
       return {0, 1, 3, 2};
+    {
+      int i = sqrt(num_nodes);
```
garth-wells: This looks weird and isn't clear. Add a comment, and see if it can be done more clearly.

cpp/dolfin/mesh/cell_types.h
```diff
@@ -84,5 +84,10 @@ int cell_degree(CellType type, int num_nodes);
 /// Mapping of DOLFIN/UFC vertex ordering to VTK/XDMF ordering
 std::vector<std::uint8_t> vtk_mapping(CellType type, int num_nodes);
 
+/// Returns the default mesh cell permutation assumed by dolfin
```
garth-wells: I don't follow the docstring - what does it map from, and what does it map to?

python/dolfin/mesh.py
```diff
@@ -43,6 +43,15 @@ def __new__(cls, value_type, mesh, dim=None):
             return mvc(mesh)
 
 
+class Mesh(cpp.mesh.Mesh):
+    def __init__(self, mpi_comm, cell_type, points, connectivity,
+                 global_cell_indices, ghost_mode, num_ghost_cells=0,
+                 custom_permutation=[]):
```
garth-wells: Can this be made `custom_permutation=[]`?

cpp/dolfin/mesh/Mesh.cpp
jorgensd: We will assume that the user does not supply the permutation, and then we need to deduce it from the element type and degree.

garth-wells: If not permutation is required it would be better that we apply no permutation. We could be explicit and pass in the cell type (or maybe just tdim and gdim?).

The less manipulation of data behind-the-scenes the better.

cpp/dolfin/mesh/Mesh.h
jorgensd: This is a mistake by me. i followed the old `vtk_mapping` documentation, which was not correct.

It should be Mapping of DOLFIN/UFC node ordering to VTK/XDMF ordering.
For a CG-1 triangle, the VTK ordering is counter clockwise, where it counts vertices, then counter clockwise on edges. The UFC mapping  counts vertices in the same way, but the edges from non-intersection lexiographically. This means that the permutation_map={0,1,2,5,3,4}, which means that the 4th dof in VTK ordering, `permutation_map[4]` is the `5`th dof in the UFC ordering

python/dolfin/mesh.py
jorgensd: What do you mean? isn't that what it says?

cpp/dolfin/mesh/Mesh.cpp
jorgensd: No permutation means that the input mesh has to be ordered as UFC would like it. This is rarely the case.This `if-test` only checks if the user has supplied a permutation, if not, it assumes that it follows our default permutations, which right now is VTK for triangles, tetrahedrons, higher order quads, and lexiographic for first order quadrilaterals and hexahedrons.

cpp/dolfin/mesh/cell_types.cpp
jorgensd: Better now?

cpp/dolfin/mesh/Mesh.h
garth-wells: Let's spin this out into a free function.

python/dolfin/mesh.py
garth-wells: Sorry, I meant to write `custom_permutation=None`.

cpp/dolfin/mesh/cell_types.cpp
garth-wells: Let's make it an assertion:
```
assert((sqrt(x) - floor(sqrt(x))) == 0); 
```

cpp/dolfin/mesh/Mesh.cpp
jorgensd: 1. I've now added documentation of the ordering (In as brief manner as possible.
2. There is a free function for this, `vtk_mapping`.

python/dolfin/mesh.py
jorgensd: No, it can't causes type-error in the wrapper.

cpp/dolfin/mesh/cell_types.cpp
jorgensd: Adding this assertion breaks the pytests, not sure why

cpp/dolfin/mesh/Mesh.h
jorgensd: Isn't `vtk_mapping` already a free function?

garth-wells: Yes, but it doesn't perform a permutation. We could have something like:
```
/// Convert input cell topology to DOLFIN cell ordering
Eigen::Array<std::int64_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> 
  gmsh_to_dolfin_topology(const Eigen::Ref<....> topology, 
                          std::vector<int> permutation);
```

jorgensd: Yes, I can have a look at that tomorrow.

jorgensd: Suggested layout:
```
Eigen::Array<std::int64_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
mesh::gmsh_to_dolfin_topology(
    Eigen::Array<std::int64_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
        topology,
    std::vector<int> permutation)
{
  Eigen::Array<std::int64_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
      cells_dolfin(topology.rows(), topology.cols());
  for (int c = 0; c < cells_dolfin.rows(); ++c)
  {
    for (int v = 0; v < cells_dolfin.cols(); ++v)
    {
      cells_dolfin(c, v) = topology(c, permutation[v]);
    }
  }
  return cells_dolfin;
}

```
Thus if input is a VTK conforming CG2 triangle , with ordering:
```
    #  *-----*-----*   3----6-----2
    #  |  \        |   | \        |
    #  |  \        |   |   \      |
    #  *     *     *   7     8    5
    #  |       \   |   |       \  |
    #  |         \ |   |        \ |
    #  *-----*-----*   0----4-----1
    cells = np.array([[0, 1, 3, 4, 8, 7], [1, 2, 3, 5, 6, 8]])
```
The call: 
 ` mesh::gmsh_to_dolfin_topology(cells, {0, 1, 2, 4, 5, 3});`
yields:
`new_cells = [[0,1,3,8,7,4],[1,2,3,6,8,5]]`
However, this means that we have to implement a inverse of `vtk_mapping` (with the name `dolfin_mapping`?)

cpp/dolfin/mesh/Mesh.cpp
michalhabera: @jorgensd @garth-wells I think we agreed to rename UFC interface as FEniCS interface (working on a PR), so I'd rather say it is a FEniCS ordering.

cpp/dolfin/mesh/cell_types.cpp
```diff
@@ -544,31 +575,88 @@ int mesh::cell_degree(mesh::CellType type, int num_nodes)
...
+    switch (degree)
+    {
+    case 1:
+      // First order quadrilateral cells does not follow counter
+      // clockwise order (cc), but lexiographic order (LG). This breaks
+      // the assumptions that the cell permutation is the same as the
+      // VTK-map.
+      return {0, 1, 2, 3};
+    default:
+      // Higher order assumes VTK
```
michalhabera: This is weird. `Q` elemnets coming from FIAT/FFC should have all lexicographic ordering, why do you take LG for linear but VTK for higher?

jorgensd: Because of the software I have available. Gmsh supplies me with higher order meshes, like quads. This will probably be rewritten as I’m working on a permutation function to transfer all VTK input to FEniCS ordering. See one of the other comments.:)

michalhabera: I do not follow. What software you have available that generates higher order quad meshes with lexicographic ordering? Moreover, the implementation in dolfin should not depends on what spftware you have available...

jorgensd: It’s the opposite way around. The built in meshes in dolfin are ordered lexiographically, while meshes from external sources are ordered VTK style. (Thus also first order quads, hexes). With introducing custom permutation, dolfin supports any consistent ordering. However, we need to assume some defaults for each element type (until I’ve added an actual perturbation function)

michalhabera: Therefore I ask, why do you assume default VTK for linear, but LG for other? Why not VTK for all?
The problem is, that you call gmsh for linear quads, but your manual meshes for higher order quads are LG.

Moreover, specification of a permutation could be confusing. Is it a permutation that will be applied (stored) after reading, or is it a permutation with respect to what we in dolfin expect as a local ordering (wrt. FEniCS ordering) (they are in inverse relation)?

jorgensd: To not break the built in meshes. I will hopefully have a better solution to this later today. The new idea is to apply the perturbation to the connectivity, changing the layout of the cell to FENICS format.

jorgensd: We could assume VTK ordering for them as well, and set a custom_permutation internally.

jorgensd: We could assume VTK ordering for them as well, and set a custom_permutation internally.

Issue #576, opened 2019-10-15
C++ unit testing might be broken in parallel on circleci


francesco-ballarin:
While testing PR #573 , it seemed that the section ``Build and run C++ unit tests (serial and MPI)`` on CI was broken in parallel. In particular, a test that was supposed to fail in parallel did not.
More precisely, I could see the stack trace with `FATL| CHECK FAILED` in the output, but CI went on as if no failure occured. I believe that this happens only if failures only occur in a strict subset of the three parallel processes.

I will open a PR shortly with a minimal example to (hopefully) reproduce again.

PR #573, opened 2019-10-11
Incompatibility of MeshFunction.mark and PR #563

francesco-ballarin:
This PR introduces a new test case for the issue I highlighted in the discussion of #563.
I believe that, after #563 and the forthcoming changes to the remaining mesh entities, MeshFunction's values for shared entities are currently out of sync between sharing processes after MeshFunction::mark.

I believe that a fix would be to add a third argument `MeshRangeType::ALL` to `mesh::MeshRange` on line 233 of MeshFunction.h. Please advise if that is indeed the correct fix, and I will apply in a commit to this branch.

PR #577, opened 2019-10-15
[WIP] C++ unit testing might be broken in parallel on circleci

francesco-ballarin:
Associated to issue #576 .

The silly test should be failing in parallel, as not all processes have rank 0.

I marked this as WIP to prevent merge, as this PR is only supposed to try and reproduce the issue that occured while testing #573, but should be discarded without merge after fixing.


francesco-ballarin:
Current failure is unrelated.

francesco-ballarin:
Latest (successful) CI run confirms that CI is affected by #576

IgorBaratta:
The test is now failing as expected.

IgorBaratta:
The test is now failing as expected.

PR #579, opened 2019-10-16
Cleanup in Mesh-class and additional support for triangle meshes < 10

jorgensd:
Moved all mesh permutation algorithms out of mesh class to separate file, `cell_conversion.cpp`. This file also contains a function that sets up the default choice of input_permutation for different orders.
 This PR mainly focus on triangles, and adds several tests that covers the permutations in a better 
way. 

For quadrilaterals, several arbitrary order transfer functions are added.
Note that there will be changes to the default of quadrilaterals of higher order in a future PR (to tensor product, and we should probably rewrite the built-in generators to this).

A tiny fix to hexahedrons is added, as there were some wrong assumptions on the input vs. vtk map.


jorgensd:
> We need to speed up the tests. The total runtime for the DOLFIN CI has roughly doubled. Python test time is up by a factor of about 3.

Im open for suggestions. I find it quite hard to find tests for this, as I dont have a mesh generator for higher order meshes, and writing them by hand is tedious. How about interpolation of the function, and point evaluation at the "dofs" of the mesh?

cpp/dolfin/io/cells.cpp
```diff
@@ -0,0 +1,323 @@
...
+                    18, 15, 11, 6, 7, 9, 16, 8,  13, 12};
+      break;
+    case 9:
+      remainders = {0,  6,  27, 1, 2, 3,  4,  5, 12, 17, 21, 24, 26, 25,
+                    22, 18, 13, 7, 8, 11, 23, 9, 10, 16, 20, 19, 14, 15};
+      break;
+    default:
+      return permutation;
+    }
+    for (int k = 0; unsigned(k) < remainders.size(); ++k)
```
chrisrichardson: use `std::size_t k = 0`

cpp/dolfin/io/cells.cpp
```diff
@@ -0,0 +1,323 @@
...
+
+    std::vector<std::uint8_t> permutation(num_nodes);
+    int j = 0;
+    std::vector<std::uint8_t> rows(n);
+    std::iota(std::next(rows.begin()), std::prev(rows.end()), 2);
+    rows.front() = 0;
+    rows.back() = 1;
+
+    std::vector<std::uint8_t>::iterator row;
+    for (row = rows.begin(); row != rows.end(); ++row)
```
chrisrichardson: `for (auto row: rows)`

cpp/dolfin/io/cells.cpp
```diff
@@ -0,0 +1,323 @@
...
+      return io::cells::vtk_to_tp(type, num_nodes);
+    default:
+      throw std::runtime_error("Unknown cell type.");
+    }
+  }
+}
+//-----------------------------------------------------------------------------
+Eigen::Array<std::int64_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
+io::cells::gmsh_to_dolfin_ordering(
+    Eigen::Array<std::int64_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
```
chrisrichardson: Should this be a const reference? e.g. `const Eigen::Ref<const Eigen::Array<...>>&` to avoid a copy?

cpp/dolfin/mesh/cell_types.cpp
```diff
@@ -482,91 +483,58 @@ mesh::cell_entity_closure(mesh::CellType cell_type)
...
+      return 7;
+    case 45:
+      LOG(WARNING) << "8th order mesh is untested";
+      return 8;
+    case 55:
+      LOG(WARNING) << "9th order mesh is untested";
+      return 9;
     default:
-      throw std::runtime_error("Unknown cell type.");
+      throw std::runtime_error("Triangle order > 9 is not supported.");
```
chrisrichardson: What if the triangle has 17 nodes? This error message is too specific.

jorgensd:
@garth-wells , @chrisrichardson are there anything more that should be changed, or are we happy with this ?

PR #586, opened 2019-10-18
Permute the dofs on simplices

mscroggs:
I've added a DofMapPermuter that creates permutations to rotate and reflect dofs on edges, on triangular faces and inside tetrahedra, then uses these to permute the global dof numbers assigned to each local dof point.

Currently one test fails: the Taylor Hood example (I think because it uses vector elements, and reversing the direction of vector elements is not yet implemented).


cpp/dolfin/fem/DofMapPermuter.cpp
```diff
@@ -0,0 +1,457 @@
...
+//-----------------------------------------------------------------------------
+void DofMapPermuter::set_cell_count(const int cells)
+{
+  _cell_orders.resize(cells, {});
+}
+//-----------------------------------------------------------------------------
+std::vector<int> DofMapPermuter::permute(std::vector<int> vec, std::vector<int> perm) const
+{
+  std::vector<int> output(perm.size());
+  for (int i = 0; i < perm.size(); ++i)
```
chrisrichardson: Use `std::size_t` with any `vector.size()`

chrisrichardson:
I think it would be better to redesign slightly. Having a `std:vector` for every cell (storing the permutation order) could be very costly. Better to just have an `int` lookup index, into a list of orders (or precomputed permutations), which could be stored as a rectangular array.

mscroggs:
> I think it would be better to redesign slightly. Having a `std:vector` for every cell (storing the permutation order) could be very costly. Better to just have an `int` lookup index, into a list of orders (or precomputed permutations), which could be stored as a rectangular array.

I have done this: orders are now stored in a rectangular eigen array

cpp/dolfin/fem/DofMapPermuter.h
```diff
@@ -0,0 +1,97 @@
...
+  /// @return The rotation and reflection orders for the triangle
+  std::pair<int, int> calculate_triangle_orders(int v1, int v2, int v3);
+
+  /// Calculates the permutation orders for a tetrahedron
+  /// @param[in] v1 The global vertex number of the tetrahedron's first vertex
+  /// @param[in] v2 The global vertex number of the tetrahedron's second vertex
+  /// @param[in] v3 The global vertex number of the tetrahedron's third vertex
+  /// @param[in] v4 The global vertex number of the tetrahedron's fourth vertex
+  /// @return The rotation and reflection orders for the tetrahedron
+  std::tuple<int, int, int, int> calculate_tetrahedron_orders(int v1, int v2,
```
garth-wells: Use `std::array<int, 4>`.

cpp/dolfin/fem/DofMapPermuter.cpp
```diff
@@ -76,47 +76,42 @@ std::array<int, 4> _calculate_tetrahedron_orders(int v1, int v2, int v3, int v4)
...
+  float root = std::sqrt(8 * blocks + 1);
   assert(root == std::floor(root) && int(root) % 2 == 1);
   int side_length = (root - 1) / 2; // side length of the triangle of face dofs
 
   std::vector<int> rotation(face_dofs);
   {
     int j = 0;
     int i = 1;
-    for (int st = face_dofs - 1; st >= 0; st -= (i++))
+    for (int st = blocks - 1; st >= 0; st -= (i++))
```
chrisrichardson: I find this kind of update confusing. Can we just put `++i;` somewhere in the loop instead?

cpp/dolfin/fem/DofMapPermuter.cpp
```diff
@@ -0,0 +1,459 @@
...
+  int side_length = (root - 1) / 2; // side length of the triangle of face dofs
+
+  std::vector<int> rotation(face_dofs);
+  {
+    int j = 0;
+    int i = 1;
+    for (int st = blocks - 1; st >= 0; st -= (i++))
+    {
+      int dof = st;
+      for (int sub = i + 1; sub <= side_length + 1; dof -= (sub++))
```
chrisrichardson: Similarly, maybe do the operation on `dof` outside the `for` statement.

mscroggs: I've changed all three loops like this

cpp/dolfin/fem/DofMapPermuter.cpp
```diff
@@ -0,0 +1,524 @@
...
+    cell_orders(cell_n, 3) = (vertices[0] > vertices[3]);
+    cell_orders(cell_n, 4) = (vertices[0] > vertices[2]);
+    cell_orders(cell_n, 5) = (vertices[0] > vertices[1]);
+
+    // Set the orders for the face rotations and reflections
+    const std::array<int, 2> tri_orders0
+        = calculate_triangle_orders(vertices[1], vertices[2], vertices[3]);
+    cell_orders(cell_n, 6) = tri_orders0[0];
+    cell_orders(cell_n, 7) = tri_orders0[1];
+    const std::array<int, 2> tri_orders1
```
chrisrichardson: Do these need to be a new array every time `tri_orders0`, `tri_orders1` etc. Maybe better to just reuse?

garth-wells: I made that change. I felt that a static allocation for each was clearer than re-assigning.

garth-wells:
Before this is merged we should have a plan for getting H(div) and H(curl) elements working soon. Or, if we (temporarily) re-order a simplex mesh will they still work?

python/test/unit/fem/test_dofmap.py
```diff
@@ -469,60 +469,122 @@ def test_readonly_view_local_to_global_unwoned(mesh):
...
-        edge_dofs_local = [dofmap.dof_layout.entity_dofs(1, e) for e in edges]
-        for edofs in edge_dofs_local:
-            assert len(edofs) == 2
-        edge_dofs = [dofs[0][edge_dofs_local[0]], dofs[1][edge_dofs_local[1]]]
-        assert set(edge_dofs[0]) == set(edge_dofs[1])
+@skip_in_parallel
+@pytest.mark.parametrize('space_type', [
+    ("P", 1), ("P", 2), ("P", 3), ("P", 4),
+    ("N1curl", 1),
+    pytest.param(("N1curl", 2), marks=pytest.mark.xfail),
```
michalhabera: Why are these expected to fail?

mscroggs: There's a bug in ffc that is causing these to fail

mscroggs: This bug (just added): https://github.com/FEniCS/ffcx/issues/175

mscroggs:
> Before this is merged we should have a plan for getting H(div) and H(curl) elements working soon. Or, if we (temporarily) re-order a simplex mesh will they still work?

If a mesh is ordered, then the DofMapPermuter will permute nothing and so vector elements will be fine. This could be a good temporary work around until flipping the directions of vectors is working?

garth-wells:
@mscroggs do we have some tests that show that things work with the old mesh ordering?

mscroggs:
@garth-wells just adding them to the repo

python/test/unit/function/test_vector_function.py
```diff
@@ -51,3 +51,42 @@ def perform_test(points, cells):
...
+        return output
+
+    # Direction is incorrect if mesh is not ordered
+    points = np.array([[0, 0, 0], [1, 0, 0], [0, 1, 0], [0, 0, 1], [1, 1, 1]])
+    cells = np.array([[0, 1, 2, 3], [1, 3, 2, 4]])
+    result = perform_test(points, cells)
+    for i, j in result:
+        assert np.allclose(i, -j)
+
+    # Direction is correct if mesh is ordered
```
garth-wells: Could we also use `Ordering::order_simplex` in this test?

mscroggs: Replaced in latest push

francesco-ballarin:
Hi,
I get

>> 2019-11-08 13:30:13.523 (   1.199s) [main thread     ]   dofs_permutation.cpp:28    WARN| Dof permutations are not defined for this cell type. High order elements may be incorrect.

when testing on a 1D mesh. Is this expected or shall I open an issue?

mscroggs:
@francesco-ballarin This is expected. (it's temporary behaviour for cell types that we've not implemented dof permutations for; although no permuting is needed in 1D so it won't cause issues there.)

francesco-ballarin:
@mscroggs ok thanks!

PR #591, opened 2019-10-23
Update MeshFunction ghost  values

IgorBaratta:
This is a proposal for updating MeshFunction ghost values. The ghost values are updated using non-blocking point-to-point communication.

Since the global numbering of mesh entities does not follow the same rules as dof numbering, `IndexMaps` could not be used directly .  Alternatively, we could work on global numbering and substitute `Topology::_shared_entities` and `Topology::_entity_owner` by an `IndexMap`.

### DIFF
--- a/cpp/dolfin/mesh/Mesh.h
+++ b/cpp/dolfin/mesh/Mesh.h
@@ -202,8 +202,7 @@ public:
   /// @return The communicator on which the mesh is distributed
   MPI_Comm mpi_comm() const;
 
-  /// Ghost mode used for partitioning. Possible values are same as
-  /// `parameters["ghost_mode"]`.
+  /// Ghost mode used for partitioning.
   /// @warning The interface may change in future without deprecation;
   ///          the method is now intended for internal library use.
   mesh::GhostMode get_ghost_mode() const;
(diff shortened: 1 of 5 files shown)


francesco-ballarin:
Thanks @IgorBaratta !

The C++ test from 6f049437 does indeed pass with additional update_ghosts() call.

However, I still have got tests in my library that are failing without the change in d7a040d, which I expected was not required anymore with this PR. I still need to investigate the cause (and propose a minimal test), but it seems related to lower dimensional subdomains. I will report back on this.
Can you spot any reason why looping over MeshRange(ALL) should be any different than looping over MeshRange(local) + updating ghosts?

IgorBaratta:
> However, I still have got tests in my library that are failing without the change in [d7a040d](https://github.com/FEniCS/dolfinx/commit/d7a040d50b0ad05cf2b179e4776d296915b6fad2), which I expected was not required anymore with this PR. I still need to investigate the cause (and propose a minimal test), but it seems related to lower dimensional subdomains. I will report back on this.

Actually this PR is incomplete since it only works for cells and vertices. 
I'll close it for now and create a new PR later, after sorting out how to fix it for edges and facets.

> Can you spot any reason why looping over MeshRange(ALL) should be any different than looping over MeshRange(local) + updating ghosts?

I guess that looping over MeshRange(ALL), means that you're accessing both regular and ghost values locally (within each process).  Updating the ghosts, on the other hand, requires gathering the values from the owning process.

francesco-ballarin:
OK thanks. I will test the new PR with my (non-minimal, for now) tests.

> I guess that looping over MeshRange(ALL), means that you're accessing both regular and ghost values locally (within each process).
Yes, this is what my library requires.

Issue #572, opened 2019-10-10
Update the Stokes-Taylor-Hood demo to offer monolithic, block and nest assembly


nate-sime:
Add monolithic, block and nest assembly options for the Stokes-Taylor-Hood demo. 

Inevitable polishing of the interface will be required as the updated demo takes shape.

nate-sime:
Now that the interface for blocked system assembly is improved, I was hoping I could get a little advice on what should be shown in the improved demo.

Currently the Stokes Taylor-Hood demo solves a Couette-type flow problem with a dolphin shape in the centre of the domain.

The following modifications would be simple:
- Nest/Block-monolithic assembly
- Use pressure mass matrix preconditioner
- Use PETSc KSP & fieldsplit and appropriate solvers in the solve

Would anyone propose any further additions or modifications to demonstrate the new features of dolfin-x?

garth-wells:
Sounds good. How about keeping it linear and doing 3D geometry with P2 boundary?

A nonlinear example with a block structure would make a good separate demo.

garth-wells:
Fixed in #780.

PR #594, opened 2019-10-25
Fix nonlinear monolithic-block assembly

nate-sime:
A prelude to #572.

Adds functionality / fixes nonlinear assembly of monolithic-blocked systems. PETSc Nest systems will follow in another pull request.

Also adds:
* Python decorator to compile `ufl.Form` arguments to `dolfin.cpp.fem.Form._cpp_object` objects.
* Re implementation of `_create_cpp_form` in the python layer.
* Create monolithic/block/nest vector/matrix utility functions in python.
* Nonlinear assembly tests


garth-wells:
See updated version in #600.

PR #596, opened 2019-10-28
Some interface changes for setting/getting tabulate function

chrisrichardson:


chrisrichardson:
No reason - I can change them to be the same.

PR #601, opened 2019-11-05
Add few wrappers to DirichletBC

michalhabera:


python/src/fem.cpp
```diff
@@ -212,7 +212,9 @@ void fem(py::module& m)
                     const std::vector<std::int32_t>&,
                     dolfin::fem::DirichletBC::Method>(),
            py::arg("V"), py::arg("g"), py::arg("facets"), py::arg("method"))
-      .def("function_space", &dolfin::fem::DirichletBC::function_space);
+      .def("dof_indices", &dolfin::fem::DirichletBC::dof_indices)
```
garth-wells: Perhaps each of these should be a read-only property?

PR #604, opened 2019-11-06
Changing mesh cell connectivity to DOLFIN structure.

jorgensd:
Meshes now have to be structured in the following way:
- UFLACS ordering for simplicies
- TensorProduct ordering for non-simplices.
This is the ordering used in FIAT.

As vertices of non-simplicies no longer are the X first nodes in the connectivity array, additional functions has been added to cope with this.

Improved tests for quadrilaterals.

Re-added VTK tests and exposing all of VTKFile to python.


cpp/dolfin/generation/RectangleMesh.cpp
```diff
@@ -249,8 +250,12 @@ mesh::Mesh build_quad(MPI_Comm comm, const std::array<Eigen::Vector3d, 2>& p,
       ++cell;
     }
 
+  Eigen::Array<std::int64_t, Eigen::Dynamic, 4, Eigen::RowMajor> topo_reordered
```
chrisrichardson: Can't we just create them in the correct order anyway?

cpp/dolfin/generation/UnitDiscMesh.cpp
```diff
@@ -92,8 +93,11 @@ mesh::Mesh UnitDiscMesh::create(MPI_Comm comm, int n,
     }
   }
 
+  Eigen::Array<std::int64_t, Eigen::Dynamic, 6, Eigen::RowMajor> cells_reordered
```
chrisrichardson: Same. Let's just change the code above to provide the correct ordering.

python/dolfin/mesh.py
```diff
@@ -10,6 +10,7 @@
 # __all__ = ["MeshFunction", "MeshValueCollection"]
 
 _meshfunction_types = {
+    "bool": cpp.mesh.MeshFunctionbool,
```
chrisrichardson: I thought we had remove MeshFunction<bool> deliberately?

michalhabera: MeshValueCollection and MeshFunction stayed. But I think specialization for bools shouldnt exist, it is a leftover probably.

cpp/dolfin/io/cells.h
```diff
@@ -46,24 +46,26 @@ std::vector<std::uint8_t> lex_to_tp(mesh::CellType type, int num_nodes);
 /// @return The map
 std::vector<std::uint8_t> vtk_to_dolfin(mesh::CellType type, int num_nodes);
 
-/// Convert gmsh cell ordering to FEniCS cell ordering
+/// Convert VTK cell ordering to FEniCS cell ordering
```
garth-wells: Where is the FEniCS cell ordering defined? It would be good point to where it's defined.

cpp/dolfin/io/cells.h
```diff
@@ -46,24 +46,26 @@ std::vector<std::uint8_t> lex_to_tp(mesh::CellType type, int num_nodes);
...
 /// @param[in] type Celltype to the permuter
 /// @return Permuted cell connectivities
 Eigen::Array<std::int64_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
-gmsh_to_dolfin_ordering(
+vtk_to_dolfin_ordering(
+    const Eigen::Ref<const Eigen::Array<
+        std::int64_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>& cells,
+    mesh::CellType type);
+
+/// Convert Lexicographic cell ordering to FEniCS cell ordering
```
garth-wells: Can the lexicographic be defined, maybe with a diagram? In the past it has sometimes increased fastest in the x-directions, and sometimes in the y-direction.

cpp/dolfin/mesh/Mesh.cpp
```diff
@@ -23,6 +23,7 @@
 
 using namespace dolfin;
 using namespace dolfin::mesh;
+using namespace dolfin::io;
```
garth-wells: Remove this. Don't use 'inside' other namespaces (in the case the `mesh` namespace).

cpp/dolfin/mesh/Mesh.cpp
```diff
@@ -87,12 +89,19 @@ compute_local_to_global_point_map(
 
   const std::int32_t num_cells = cell_nodes.rows();
   const std::int32_t num_nodes_per_cell = cell_nodes.cols();
+  const std::vector<int> vertex_indices = mesh::cell_vertex_index(
+      type, num_nodes_per_cell, num_vertices_per_cell);
+
+  std::vector<int> non_vertex_indices;
+  for (int i = 0; i < num_nodes_per_cell; ++i)
```
garth-wells: Add curly braces.

cpp/dolfin/mesh/Mesh.cpp
```diff
@@ -87,12 +89,19 @@ compute_local_to_global_point_map(
...
+    if (std::find(vertex_indices.begin(), vertex_indices.end(), i)
```
garth-wells: Add curly braces.

cpp/dolfin/mesh/cell_types.cpp
```diff
@@ -539,4 +538,20 @@ int mesh::cell_degree(mesh::CellType type, int num_nodes)
...
   }
 }
+
+std::vector<int> mesh::cell_vertex_index(mesh::CellType type, int num_nodes,
+                                         int num_vertices_per_cell)
+{
+  int degree = mesh::cell_degree(type, num_nodes);
+  switch (type)
+  {
+  case mesh::CellType::quadrilateral:
```
garth-wells: What about hexes?

cpp/dolfin/mesh/cell_types.h
```diff
@@ -81,5 +81,12 @@ cell_entity_closure(mesh::CellType cell_type);
 /// Find the order of a cell given the number of nodes in the element
 int cell_degree(CellType type, int num_nodes);
 
+/// Find indices for vertices for a cell
+/// @param[in] type The cell shape
+/// @param[in] num_nodes The number of cell 'nodes'
```
garth-wells: Explain node in this context?

cpp/dolfin/mesh/cell_types.h
```diff
@@ -81,5 +81,12 @@ cell_entity_closure(mesh::CellType cell_type);
...
+/// @param[in] num_vertices_per_cell Number of vertices per cell
```
garth-wells: Isn't the number of vertices determined by the `CellType`?

cpp/dolfin/mesh/cell_types.h
```diff
@@ -81,5 +81,12 @@ cell_entity_closure(mesh::CellType cell_type);
...
+/// @return List of vertex-indices
+ std::vector<int> cell_vertex_index(mesh::CellType type, int num_nodes, int num_vertices_per_cell);
```
garth-wells: This function name is not clear.

cpp/dolfin/mesh/cell_types.h
```diff
@@ -81,5 +81,12 @@ cell_entity_closure(mesh::CellType cell_type);
 /// Find the order of a cell given the number of nodes in the element
 int cell_degree(CellType type, int num_nodes);
 
+/// Find indices for vertices for a cell
```
garth-wells: Comment isn't clear to me.

python/test/unit/fem/test_dofmap.py
```diff
@@ -526,3 +526,54 @@ def check(mesh, edges):
     assert e0 == e1
     x0, x1 = check(mesh, [1, 2])
     assert np.allclose(x0, x1)
+
+
+@skip_in_parallel
```
garth-wells: Add some explanation of what the tests is doing/testing.

python/test/unit/io/test_vtk.py
```diff
@@ -10,24 +10,27 @@
...
+from dolfin.io import VTKFile
 
 assert (tempdir)
 
 
 # VTK file options
-@fixture
+@pytest.fixture
 def file_options():
     return ["ascii", "base64", "compressed"]
```
garth-wells: I thought compressed and base64 had been removed?

cpp/dolfin/mesh/cell_types.cpp
jorgensd: I wanted to add the Hex stuff separately. Its in https://github.com/FEniCS/dolfinx/commits/dokken/mesh-input-3D
I can of course add everything to this PR if that's whats easier for everyone.

garth-wells: Presumably the hex case should be added here and an exception thrown?

python/dolfin/mesh.py
jorgensd: I can remove the MeshFunctionbool for the python exposure and tests

cpp/dolfin/generation/RectangleMesh.cpp
jorgensd: Its easy to change this to be TensorProduct, and I will do this.

cpp/dolfin/generation/UnitDiscMesh.cpp
jorgensd: This one is slightly more complicated to change, as the connectivity is made manually. I do not really want change these.

cpp/dolfin/generation/RectangleMesh.cpp
jorgensd: Thinking about this, it should be needed to permute the connectivity. Since TP and Lexicographical ordering is just a rotation of the other for first  order meshes, and the connectivity for a cell doesnt care about what is the primal axis. however, to keep things clean I will still change this.

cpp/dolfin/mesh/cell_types.cpp
jorgensd: I'll just add the 3D stuff to this PR

python/test/unit/io/test_vtk.py
jorgensd: Right now we do not use `file_options` in the tests, as the initializer in CPP do not support filetype. I just updated the syntax from the previous dolfin in case we decide to re-add this. You can see that there are a lot of tests in this file that are marked with xfail.

python/src/io.cpp
```diff
@@ -49,16 +53,17 @@ void io(py::module& m)
...
-              const std::string data_path, bool use_partition_from_file) {
-             auto x = self.read_vector(comm.get(), data_path,
-                                       use_partition_from_file);
-             Vec _x = x.vec();
-             PetscObjectReference((PetscObject)_x);
-             return _x;
-           },
-           py::return_value_policy::take_ownership)
+      .def(
+          "read_vector",
```
michalhabera: Why do you put a space here (and everywhere below)?

python/test/unit/io/test_vtk.py
```diff
@@ -43,12 +46,12 @@ def test_save_1d_meshfunctions(tempfile, mesh_function_types, file_options,
...
-            VTKFile(tempfile + "mf.pvd", "ascii").write(mf)
-            f = VTKFile(tempfile + "mf.pvd", "ascii")
+            VTKFile(tempfile + "mf.pvd").write(mf)
+            f = VTKFile(tempfile + "mf.pvd")
             f.write(mf, 0.)
             f.write(mf, 1.)
-            for file_option in file_options:
-                VTKFile(tempfile + "mf.pvd", file_option).write(mf)
+            # for file_option in file_options:
+            #     VTKFile(tempfile + "mf.pvd", file_option).write(mf)
```
michalhabera: Leftover commented code?

python/src/io.cpp
jorgensd: Its clang-format  (Im using clang-format8), that does this.

python/test/unit/io/test_vtk.py
```diff
@@ -71,74 +74,80 @@ def test_save_3d_meshfunctions(tempfile, mesh_function_types, file_options,
...
 
 
+@pytest.mark.xfail(reason="file_option not added to VTK initializer")
```
michalhabera: But this turns off VTKFile testing. I am against xfailing whole test just because this file_options is not really supported.
Also, why is this done in this PR?

python/test/unit/io/test_vtk.py
jorgensd: Can remove it, its basically due to the fact that the VTK signature has changed (see all the xfail tests further down)

python/test/unit/io/test_vtk.py
michalhabera: Ok, I see VTK was not really tested anyway. But are these changes related to the PR?

jorgensd: I can Remove the file format options from the tests, and thus activate them again. I Added these back in here because VTK is the only reliable format for higher order problems, and i figured that we should the support all output formats in VTK

cpp/dolfin/generation/UnitDiscMesh.cpp
garth-wells: Can we remove `UnitDiscMesh` now that we have straightforward support for (py)gmsh?

cpp/dolfin/io/cells.h
michalhabera: Actually, that shouldn't matter. *Local* orientations are equivalence classes, i.e. all local orientations which are equal up to rotations are reflections should be considered the same. Or in other words, local orientation is a topological property.

python/test/unit/io/test_vtk.py
garth-wells: Removed

python/src/io.cpp
garth-wells: Might need more recent `clang-format`.

python/test/unit/io/test_vtk.py
garth-wells: Let's register and issue to clean up the VTK interface from Python.

python/test/unit/io/test_vtk.py
garth-wells: Removed.

python/dolfin/mesh.py
garth-wells: Removed.

cpp/dolfin/generation/UnitDiscMesh.cpp
chrisrichardson: Yes. We should probably also remove UnitTriangleMesh and UnitTetrahedronMesh

PR #606, opened 2019-11-06
Added test of all celltypes with a manufactured solution

jorgensd:
Testing CG 2 - CG 4 functionspaces for intervals, triangles, quads, tetrahedrons and hexes
finding the solution to the Poisson-problem with `u_ex=\Pi_{0}^gdim (1-x)*x**(gdim-1)`.

Numerical approximations are compared to the exact values by measuring absolute error.


python/test/unit/fem/test_fem_pipeline.py
```diff
@@ -0,0 +1,134 @@
...
+import sympy as sp
+
+from dolfin import (MPI, DirichletBC, Function, FunctionSpace, TestFunction,
+                    TrialFunction, UnitCubeMesh, UnitIntervalMesh,
+                    UnitSquareMesh, solve)
+from dolfin.cpp.mesh import CellType
+from ufl import SpatialCoordinate, dx, grad, inner
+
+
+class ManufacturedSolutionPoisson():
```
garth-wells: Seems a bit heavy-weight to have a class for this. Could a single function suffice?

michalhabera:
I think best would be to measure/test the convergence rates.

python/test/unit/fem/test_fem_pipeline.py
```diff
@@ -0,0 +1,134 @@
...
+    if gdim == 2:
+        mesh = UnitSquareMesh(MPI.comm_world, 15, 15, cell_type)
+    elif gdim == 3:
+        mesh = UnitCubeMesh(MPI.comm_world, 10, 10, 10, cell_type)
+    V = FunctionSpace(mesh, ("CG", degree))
+    u, v = TrialFunction(V), TestFunction(V)
+    x = SpatialCoordinate(mesh)
+
+    man_sol = ManufacturedSolutionPoisson(degree, gdim)
+    f = -eval(str(man_sol.laplace()))
```
garth-wells: How accurately is the `f` evaluated? I think (i) the quadrature degree needs to be specified to make sure the polynomial is integrated exactly, or `f` needs to be interpolated in a polynomial spaces such that it's represented exactly.

python/test/unit/fem/test_fem_pipeline.py
```diff
@@ -0,0 +1,134 @@
...
+    x_ = np.zeros((x.shape[0], 3))
+    # Fill with coordinates to work with eval
+    if gdim == 1:
+        x_[:, :1] = x
+    elif gdim == 2:
+        x_[:, :2] = x
+    else:
+        x_ = x
+
+    for i in range(V.mesh.num_cells()):
```
garth-wells: Checking at every point seems a bit excessive. How abut and just a few sample points?

python/test/unit/fem/test_fem_pipeline.py
```diff
@@ -0,0 +1,134 @@
...
+        if mesh.geometric_dimension() == 1:
+            exact_i = man_sol.eval(x_i[:, 0], 0, 0)
+        elif mesh.geometric_dimension() == 2:
+            exact_i = man_sol.eval(x_i[:, 0], x_i[:, 1], 0)
+        elif mesh.geometric_dimension() == 3:
+            exact_i = man_sol.eval(x_i[:, 0], x_i[:, 1], x_i[:, 2])
+        # Reshape
+        result_i = result_i.reshape(exact_i.shape)
+        # Measure absolute error as values close to zero breaks the rtol.
+        assert max(exact_i - result_i) < 1e-3
```
garth-wells: This tolerance is very lax. Probably because the the LHS isn't being evaluated very accurately.

Also, use `pytest.approx`.

garth-wells:
@michalhabera The objective with this test is to check against a manufactured polynomial solution so we know that the FE solution should be exact. This is a precise test that works on on small meshes.

python/test/unit/fem/test_fem_pipeline.py
jorgensd: Which points? We need to know which cell the points are in to do the evaluation. Computing entity collisions does not work for non-simplicies as far as I've understood. This is also related to the lax tolerance below. In certain cells, we hit the issue of small numbers, i.e.
Exact
```
[1.49250680e-71 4.05622938e-37 2.10285187e-04 2.46427953e-04
 2.26226946e-04 2.38517920e-04 2.75302632e-05 1.11055963e-04
 2.55902608e-05 9.79105632e-05 3.35024367e-06 5.17085965e-05
 3.11415872e-06 4.55879871e-05 1.40078372e-40 5.35952031e-40
 1.05333181e-04 4.90440208e-05 1.28183236e-05 1.19150421e-05]
```
Approximated:
```
[ 2.58843355e-20 -7.19762383e-21  3.91022441e-04  4.72741436e-04
  3.31029333e-04  3.42966176e-04  8.44471629e-05  1.92304646e-04
  1.09539362e-04  1.39022866e-04  1.89449749e-05  1.12650811e-04
  3.17445495e-05  6.74901978e-05  3.19225306e-21  2.03138808e-20
  1.70234919e-04  8.19984581e-05  4.04526864e-05  4.29223670e-05]
```
This fails with a smaller absolute tolerance.

python/test/unit/fem/test_fem_pipeline.py
michalhabera: Do not use python std library `max()` on larger arrays. It is too generic and very slow. Prefer `numpy.max(a,b)`.

jorgensd:
@garth-wells @michalhabera  I changed the tests, such that it measures the L2 error of the solution. However, this highlights that something is weird with the tetrahedrons, as the error is significantly bigger than for the other mesh types.

python/test/unit/fem/test_fem_pipeline.py
jorgensd: Test changed, so closing this request

garth-wells:
Updated version in #607.

PR #612, opened 2019-11-10
Re-factor block matrix assembly

garth-wells:
Unpicks stages, and re-used more of the non-block assembly.


garth-wells:
See #613.

PR #619, opened 2019-11-11
Improve types for rectangular arrays

garth-wells:
Use Eigen for rectangular arrays of pointers.


python/test/unit/fem/test_assembler.py
```diff
@@ -200,6 +200,7 @@ def boundary(x):
 
     # Monolithic blocked
     A0 = dolfin.fem.assemble_matrix_block(a_block, [bc])
+    return
```
chrisrichardson: Is this intentional?

Issue #616, opened 2019-11-11
Interpolation of finite element function causing core dumped


jorgensd:
Minimal example: 
```
from dolfin import UnitSquareMesh, MPI, Function, FunctionSpace

mesh = UnitSquareMesh(MPI.comm_world, 10, 10)
V = FunctionSpace(mesh, ("CG", 1))
u = Function(V)
uh = Function(V)
u.vector.set(1)
uh.interpolate(u)
```
Error caused in [interpolation call](https://github.com/FEniCS/dolfinx/blob/master/python/dolfin/function.py#L121-L124), as it no longer returns a TypeError

PR #620, opened 2019-11-11
Fixing python interface of interpolation to work with finite element functions

jorgensd:
Fixing issue #616

### DIFF
--- a/python/dolfin/function.py
+++ b/python/dolfin/function.py
@@ -115,13 +115,14 @@ class Function(ufl.Coefficient):
 
     def interpolate(self, u) -> None:
         """Interpolate an expression"""
-
         @singledispatch
         def _interpolate(u):
-            try:
-                self._cpp_object.interpolate(u)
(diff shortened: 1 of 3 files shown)


python/dolfin/function.py
```diff
@@ -115,13 +115,14 @@ def eval(self, x: np.ndarray, cells: np.ndarray, u=None) -> np.ndarray:
 
     def interpolate(self, u) -> None:
         """Interpolate an expression"""
-
         @singledispatch
         def _interpolate(u):
-            try:
-                self._cpp_object.interpolate(u)
-            except TypeError:
+            if isinstance(u, self.__class__):
```
garth-wells: We're trying to avoid `isinstance`. Was the `try`/`except` broken?

jorgensd: Yes, it threw: `Fatal Python error: Cannot recover from stack overflow.`

PR #623, opened 2019-11-11
Simplify Python/NumPy syntax for transposing layout

garth-wells:
The allows natural NumPy 'vectorised' expressions.


jorgensd:
> The allows natural NumPy 'vectorised' expressions.

This seems reasonable. I initially thought that you were going to transpose the storage of the mesh-nodes, which would have lead to loads of changes in the mesh constructor. However, as it only transposes the tablulated dof coordinates, it has few consequences. Additionally, it unifies the python interface of numpy expressions with the UFL-way of writing and expressions (using SpatialCoordinate), which is great.

garth-wells:
The inconsistency between functions that interpolate functions of `x` and functions that work with `ufl.SpatialCoordinate` were one of the motivations for trying this change.

PR #629, opened 2019-11-12
Unify assembler routines to 2-level imports

michalhabera:
Some recent changes didnt import all assembler routines to 2nd level.


michalhabera:
@garth-wells Any comments?

garth-wells:
Fine with me, but I'm not sure that all the assembler function names/interfaces have settled down quite yet.

PR #630, opened 2019-11-12
remove Mesh.geometric_dimension()

nate-sime:
Simple cleanup. Remove `Mesh.geometric_dimension()` in the python layer in favour of `Mesh.geometry.dim`

### DIFF
--- a/python/dolfin/mesh.py
+++ b/python/dolfin/mesh.py
@@ -68,13 +68,6 @@ def ufl_domain(self):
     return self._ufl_domain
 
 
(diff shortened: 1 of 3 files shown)


nate-sime:
> Not sure, but is this used in ufl, e.g. with `SpatialDimension(mesh)`?

The `ufl` necessities are glued onto `dolfin.cpp.mesh.Mesh` [here](https://github.com/FEniCS/dolfinx/blob/9a2f823ec2046ec48c149e964ceae06dcc0e6686/python/dolfin/mesh.py#L84-L86).

PR #631, opened 2019-11-12
Mscroggs/build distributed

mscroggs:
We've exposed the build_distributed_mesh function to python and added a test that uses it as an example. The new test tests the dof permuter in parallel.

### DIFF
--- a/python/test/unit/fem/test_dofmap.py
+++ b/python/test/unit/fem/test_dofmap.py
@@ -546,6 +546,12 @@ def test_triangle_dof_ordering_parallel(space_type):
     for x in range(N - 1):
         for y in range(N - 1):
             a = N * y + x
+            # Adds two triangle cells:
+            # a+N -- a+N+1
+            #  |   / |
+            #  |  /  |
+            #  | /   |
+            #  a --- a+1
             for cell in [[a, a + 1, a + N + 1], [a, a + N + 1, a + N]]:
                 cells.append([order[i] for i in cell])
 
--- a/python/test/unit/fem/test_dofmap.py
+++ b/python/test/unit/fem/test_dofmap.py
@@ -16,6 +16,7 @@ from dolfin import (MPI, FunctionSpace, Mesh, MeshEntity, UnitCubeMesh,
                     fem)
 from dolfin.cpp.mesh import CellType, GhostMode
 from ufl import FiniteElement, MixedElement, VectorElement
+from random import shuffle
 
 xfail = pytest.mark.xfail(strict=True)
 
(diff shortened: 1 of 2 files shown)


garth-wells:
More a comment on the current design than this change, but `build_distributed_mesh` is a bit weird with the current `Mesh` constructor design. It would be nicer if it just returned partitioned cells and vertices, which are then passed to the `Mesh` constructor.

The issue preventing this at the moment is the ghost handing, see https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfin/mesh/Partitioning.cpp#L683.

mscroggs:
> More a comment on the current design than this change, but `build_distributed_mesh` is a bit weird with the current `Mesh` constructor design. It would be nicer if it just returned partitioned cells and vertices, which are then passed to the `Mesh` constructor.
> 
> The issue preventing this at the moment is the ghost handing, see https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfin/mesh/Partitioning.cpp#L683.

This might be a useful temporary feature, as we need this to test the dof permuter and other mesh stuff (eg gmsh) in parallel

python/test/unit/fem/test_dofmap.py
```diff
@@ -524,6 +523,69 @@ def test_triangle_dof_ordering(space_type):
...
+    ("P", 1), ("P", 2), ("P", 3), ("P", 4),
+    ("N1curl", 1),
+    ("RT", 1), ("RT", 2), ("RT", 3), ("RT", 4),
+    ("BDM", 1),
+    ("N2curl", 1),
+])
+def test_triangle_dof_ordering_parallel(space_type):
+    """Checks that dofs on shared triangle edges match up"""
+    # Create a triangle mesh
+    from random import shuffle
```
garth-wells: Move to the top of the file?

python/test/unit/fem/test_dofmap.py
```diff
@@ -524,6 +523,69 @@ def test_triangle_dof_ordering(space_type):
...
+    N = 10
+    points = np.array([[x / 2, y / 2] for x in range(N) for y in range(N)])
+    cells = []
+    for x in range(N - 1):
```
garth-wells: Add explanation of what this does.

python/test/unit/fem/test_dofmap.py
```diff
@@ -524,6 +523,69 @@ def test_triangle_dof_ordering(space_type):
...
+        for y in range(N - 1):
+            a = N * y + x
+            for cell in [[a, a + 1, a + N + 1], [a, a + N + 1, a + N]]:
+                # Put points in cell in a random order to test dof permuting
+                shuffle(cell)
```
garth-wells: Do we ultimately want to support arbitrary ordering, or insist on anti-clockwise ordering?

python/test/unit/fem/test_dofmap.py
mscroggs: done

python/test/unit/fem/test_dofmap.py
mscroggs: done

python/test/unit/fem/test_dofmap.py
mscroggs: I've changed this to have the cells in anticlockwise order, but the points in the mesh give a random order

PR #636, opened 2019-11-13
Rename dolfin python package dolfinx

nate-sime:
Rename the `dolfin` python package to `dolfinx`.
Also rename `dolfin_utils` to `dolfinx_utils`.


francesco-ballarin:
If this is merged, should we expect to see soon the dolfin namespace in C++ files replaced by a dolfinx one as well?

PR #639, opened 2019-11-14
Fix #638

### LINKED ISSUE #638: MeshFunction.mark is not up to date with #623
After updating to #623 I went through my code and transposed the point storage. However, is seems that points passed to MeshFunction.mark still use the old layout, i.e. requiring to use `x[:, 1]` rather than `x[1]` in the python layer.

Example to reproduce
```
from numpy import array, finfo, isclose
from ufl import Measure
from dolfin import *
from dolfin.fem import assemble_scalar

mesh = RectangleMesh(MPI.comm_world, [array([-1.0, -2.0, 0.0]), array([1.0, 2.0, 0.0])], [50, 100])

subdomains = MeshFunction("size_t", mesh, 2, 0)
def subdomain(x):
    return x[1] < finfo(float).eps
subdomains.mark(subdomain, 1)

dx = Measure("dx", domain=mesh, subdomain_data=subdomains)

area1 = MPI.sum(mesh.mpi_comm(), assemble_scalar(1.*dx(1)))
print("area1 = ", area1)
assert isclose(area1, 4.)
```

michalhabera:


cpp/test/unit/mesh/MeshFunction.cpp
```diff
@@ -14,9 +14,9 @@ using namespace dolfin;
 namespace
 {
 Eigen::Array<bool, Eigen::Dynamic, 1>
-marking_function(const Eigen::Ref<const Eigen::Array<double, Eigen::Dynamic, 3, Eigen::RowMajor>>& x)
+marking_function(const Eigen::Ref<const Eigen::Array<double, 3, Eigen::Dynamic, Eigen::RowMajor>>& x)
 {
-  Eigen::Array<bool, Eigen::Dynamic, 1> inside(x.rows());
+  Eigen::Array<bool, 1, Eigen::Dynamic> inside(x.cols());
```
garth-wells: Do we need to flip this one?

michalhabera: I just did it to match the second dimension of incoming data, i.e. x[dim, num_points] --> bool[1, num_points]

francesco-ballarin:
Thanks!

PR #643, opened 2019-11-15
Improving and fixing XDMF interface for first order non-simplices (meshes and mesh functions)

jorgensd:
After an issue with the visualization of quadrilateral FacetFunction saved with XDMF reported by @michalhabera , I've gone through the XDMF read/write for non-simplices.

- Added possibility to read in hexahedral meshes with XDMF
- Fixing assumption on number of vertices for FacetFunctions of non-simplices
- Proper lowering of meshes with dim <= tdim to use the appropriate VTK permutation
- Parameterized most XDMF tests to consider both simplices and non-simplices

The last fix is only implemented for first order meshes, as saving higher order meshes is not a priority for XDMF (Using VTK or HDF5 is most probable right now). Also, a fix for higher order tetrahedrons would require a new set of vtk permutation functions, as the layout of element_dofs on a subtriangle of a higher order tetrahedron is no the same as for a higher order triangle in VTK.


cpp/dolfin/io/xdmf_write.cpp
```diff
@@ -202,8 +202,8 @@ std::vector<std::int64_t> compute_topology_data(const mesh::Mesh& mesh,
                                                 int cell_dim)
 {
   // Create vector to store topology data
-  const int num_vertices_per_cell = mesh::num_cell_vertices(
-      mesh::cell_entity_type(mesh.cell_type(), cell_dim));
+  mesh::CellType cell_type = mesh::cell_entity_type(mesh.cell_type(), cell_dim);
```
michalhabera: Cell type is also `const`.

cpp/dolfin/io/xdmf_write.cpp
```diff
@@ -212,8 +212,13 @@ std::vector<std::int64_t> compute_topology_data(const mesh::Mesh& mesh,
   MPI_Comm comm = mesh.mpi_comm();
 
   int num_nodes = mesh.coordinate_dofs().cell_permutation().size();
-  const std::vector<std::uint8_t> perm
-      = io::cells::dolfin_to_vtk(mesh.cell_type(), num_nodes);
+  std::vector<std::uint8_t> perm;
+  if (cell_dim == mesh.topology().dim())
+    perm = io::cells::dolfin_to_vtk(mesh.cell_type(), num_nodes);
```
michalhabera: Should here be `cell_type`?

cpp/dolfin/io/xdmf_write.cpp
michalhabera: So it this rather an `enitity_cell_type`?

jorgensd: Sure, i can rename

michalhabera:
This almost doubled the python tests time, which is not good. IO tests are already a bit blown, so I wouldn't add more tests - in addition, they do not really test the functionality.

jorgensd:
> This almost doubled the python tests time, which is not good. IO tests are already a bit bloated, so I wouldn't add more tests - in addition, they do not really test the functionality.

I can remove the parameterized tests.

michalhabera:
@jorgensd Yes, I would really recommend to do so (my opinion).

michalhabera:
@jorgensd Can you please fix the conflicts with master?

jorgensd:
> @jorgensd Can you please fix the conflicts with master?

Done. Waiting for tests to pass (Even if they dont really get touched by this)

PR #647, opened 2019-11-15
Update the stokes demo 

nate-sime:
Development of issue #572.

Initial phase solves the same 2D problem as before with the following enhancements

- Block (nest) assembly
- Pressure mass matrix (block) preconditioning matrix
- Symmetric block Jacobi preconditioner composition
- Near nullspace of rigid body modes attached to the velocity block with PETSc GAMG

Drawbacks and advice requested:

- Slower than direct solver for this 2D problem
- With Neumann BCs, there is no opportunity to demonstrate attaching the pressure nullspace
- Old mesh composed of piecewise linear boundary representation, not high order


garth-wells:
This should be a diagonal Schur preconditioner, and use MINRES for the outer Krylov iterations.

I'd suggest using LU as the preconitioner on each block to first check that the iteration count comes down to around 20-30.

garth-wells:
@nate-sime see a few tweaks in https://github.com/FEniCS/dolfinx/tree/garth/tweak-fieldsplit-options.

jpdean:
If you decide to change the problem to one that allows demonstration of how to attach the pressure nullspace, I have coded up a problem with an analytical solution from "Finite Element Methods for Flow Problems" by Donea and Huerta, which may be of use since it has a Dirichlet boundary condition applied over the entire boundary.

nate-sime:
Work continued in #780

PR #649, opened 2019-11-17
Use MPI neighbourhood collectives when building sparsity pattern

garth-wells:


cpp/dolfin/common/IndexMap.cpp
```diff
@@ -1,113 +1,137 @@
...
 
   //  May have repeated shared indices with different processes
-  MPI_Neighbor_alltoallv(reverse_indices.data(), reverse_sizes.data(),
-                         displs_reverse.data(), MPI_INT,
-                         _forward_indices.data(), _forward_sizes.data(),
-                         displs_forward.data(), MPI_INT, _neighbour_comm);
+  std::vector<int> indices_in(disp_in.back());
+  MPI_Neighbor_alltoallv(
+      out_indices.data(), out_edges_num.data(), disp_out.data(), MPI_INT,
+      indices_in.data(), // out
```
IgorBaratta: What does the comment "out" mean in this position?

garth-wells: It follows the MPI and graph jargon for edges, defining 'out' edges as ghost to owner (chosen as this is what we have to start), with 'in' edges being owner -> ghost. I think it's clearer thinking of it as a graph in terms of forward and reverse.

I think eventually we should store two neighbourhood communicators for communication in different directions.

PR #651, opened 2019-11-18
Removed meshio from higher order mesh tests. Reintroduced datadir

jorgensd:
Meshio removed from tests, as it adds the possibility of extra failures (particularly in combination with h5).


python/test/data/mesh.xdmf
```diff
@@ -0,0 +1,4109 @@
+<Xdmf Version="3.0">
```
garth-wells: Can this file be made smaller?

jorgensd: Not without changing the relative tolerance of the corresponding test.

jorgensd: Im using the following script to generate the geometry (for reference). Since we are generating a spherical shell, we have to require quite little curvature of each element.
```
order = 2 
R = 1 
res = R / 7
import meshio 
import pygmsh 
geo = pygmsh.opencascade.Geometry() 
geo.add_raw_code("Mesh.ElementOrder={0:d};".format(order)) 
geo.add_ball([0, 0, 0], R, char_length=res) 
element = "triangle{0:d}".format(int((order + 1) * (order + 2) / 2)) 
if order == 1: 
    element = element[:-1] 
msh = pygmsh.generate_mesh(geo, verbose=True, dim=2) 
meshio.write("mesh.xdmf", meshio.Mesh(points=msh.points, cells={element: msh.cells[element]}), data_format="XML")               
 ```

jorgensd: Mesh reduced and relative tolerance is less strict

PR #655, opened 2019-11-20
Expose shared_indices of IndexMap

chrisrichardson:
* Exposes useful information of `IndexMap`
* Needed for work on using `IndexMap` with `Topology`


cpp/dolfin/common/IndexMap.cpp
```diff
@@ -151,6 +149,34 @@ const Eigen::Array<PetscInt, Eigen::Dynamic, 1>& IndexMap::ghosts() const
   return _ghosts;
 }
 //-----------------------------------------------------------------------------
+std::map<std::int32_t, std::set<int>> IndexMap::shared_indices() const
+{
+  int indegree(-1), outdegree(-2), weighted(-1);
```
garth-wells: Maybe add a few comments for each step in this function.

cpp/dolfin/common/IndexMap.h
```diff
@@ -100,6 +102,9 @@ class IndexMap
   /// Owner rank (on global communicator) of each ghost entry
   Eigen::Array<std::int32_t, Eigen::Dynamic, 1> ghost_owners() const;
 
+  /// Map from shared indices to sharing processes
+  std::map<std::int32_t, std::set<int>> shared_indices() const;
```
garth-wells: I suggest renaming to `compute_shared_indices` (and update docstring) to make clear that it does some work behind the scenes.

Also, update the docstring to make clear what is being returned. E.g. is it just for owned indices?

python/src/common.cpp
```diff
@@ -65,6 +65,11 @@ void common(py::module& m)
       .def_property_readonly("ghosts", &dolfin::common::IndexMap::ghosts,
                              py::return_value_policy::reference_internal,
                              "Return list of ghost indices")
+      .def_property_readonly("shared_indices",
+                             &dolfin::common::IndexMap::shared_indices,
+                             py::return_value_policy::reference_internal,
```
garth-wells: This doesn't like right since the function returns a value rather than reference or pointer.

cpp/dolfin/common/IndexMap.h
chrisrichardson: Yes, that is a question I was wondering about. It is only for owned indices, but it would be useful to have the sharing of ghosts too (although that requires further communication). Should I add another function to do that? Or an optional argument? Or just let it be done externally?

chrisrichardson:
The complete functionality is not essential. 
However, we do need access to the forward_indices, somehow, and this is a convenient wrapping.

cpp/dolfin/common/IndexMap.h
```diff
@@ -100,6 +102,10 @@ class IndexMap
   /// Owner rank (on global communicator) of each ghost entry
   Eigen::Array<std::int32_t, Eigen::Dynamic, 1> ghost_owners() const;
 
+  /// Create a map from shared indices to sharing processes
```
garth-wells: Does this include ghosts? It's ambiguous if ghosts are 'shared'.

IgorBaratta: Good point.
Since this information is relevant as a map only for owned indices (each index is owned by only one processes and may be shared with another two or so), I think we just need to explain this on documentation and use it consistently.

IgorBaratta:
Maybe not directly related to this PR, but since it's a prelude to a Topology PR, I think it would be good to think of reusing some nomenclature in Topology and maybe other classes.
It's a little bit confusing to have `num_ghosts`, `size_local` on one side and `size` and `ghost_offset` on the other.

cpp/dolfin/common/IndexMap.h
chrisrichardson: I will rethink some of the naming

chrisrichardson:
I think we can use `topology.index_map(0).num_ghosts()`, `topology.index_map(0).size_local()` and `topology.index_map(0).size_global` and remove `topology.ghost_offset` etc.

PR #659, opened 2019-11-21
Implement the interface changes to SparsityPatternBuilder proposed in #657

### LINKED ISSUE #657: Proposal of interface change to SparsityPatternBuilder
Dear all,
in the context of the multiphenics library, I have a custom `BlockDofMap`, for which the `cell_dofs()` method may only return a subset of dolfinx `DofMap::cell_dofs()`. In particular, I then use my custom `BlockDofMap` during the initialization of the sparsity pattern.

In the old dolfin, my `BlockDofMap` inherited from `GenericDofMap`, and since `GenericDofMap::cell_dofs()` was virtual, 
I could simply reuse the code in `SparsityPatternBuilder`.

In the current dolfinx, even if my `BlockDofMap` were to inherit from `DofMap` (it does not, at the moment), I cannot reuse the code in `cpp/dolfin/fem/SparsityPatternBuilder.cpp` because `DofMap::cell_dofs()` is not virtual anymore.

I understand that effort was put into removing all virtual methods. In order to allow the same code reuse that I had in the legacy version, would it be possible to change the last argument of
```
void SparsityPatternBuilder::*(
    la::SparsityPattern& pattern, const mesh::Mesh& mesh,
    const std::array<const fem::DofMap*, 2> dofmaps)
```
e.g. to a function pointer, where such a function is taking as input an index i (either 0 or 1) and an cell index c, and it is wrapping all the calls to `dofmaps[i]->cell_dofs(c)`?
The interface change will not hinder the overall functionality of dolfinx, and I would be able to to use my custom implementation of `cell_dofs()` without having to duplicate all the `SparsityPatternBuilder` code.

Thanks,
Francesco

(discussed this week with @michalhabera and @jhale)

francesco-ballarin:
Closes #657

### DIFF
--- a/cpp/dolfin/fem/utils.cpp
+++ b/cpp/dolfin/fem/utils.cpp
@@ -121,9 +121,9 @@ la::PETScMatrix dolfin::fem::create_matrix(const Form& a)
   }
 
   // Get dof maps
-  std::array<const DofMap*, 2> dofmaps
-      = {{a.function_space(0)->dofmap().get(),
-          a.function_space(1)->dofmap().get()}};
+  std::array<std::shared_ptr<const DofMap>, 2> dofmaps
+      = {{a.function_space(0)->dofmap(),
+          a.function_space(1)->dofmap()}};
 
   // Get mesh
   assert(a.mesh());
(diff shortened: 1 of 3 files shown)


cpp/dolfin/fem/utils.cpp
```diff
@@ -121,9 +121,9 @@ la::PETScMatrix dolfin::fem::create_matrix(const Form& a)
   }
 
   // Get dof maps
-  std::array<const DofMap*, 2> dofmaps
-      = {{a.function_space(0)->dofmap().get(),
-          a.function_space(1)->dofmap().get()}};
+  std::array<std::shared_ptr<const DofMap>, 2> dofmaps
```
francesco-ballarin: I changed from pointer to shared_ptr as there was no need to getting raw pointers anymore.

cpp/dolfin/fem/utils.cpp
```diff
@@ -137,14 +137,14 @@ la::PETScMatrix dolfin::fem::create_matrix(const Form& a)
 
   // Create and build sparsity pattern
   la::SparsityPattern pattern(mesh.mpi_comm(), index_maps);
+  SparsityPatternBuilder::get_cell_dofs_function get_cell_dofs = 
+    [&dofmaps](unsigned int i, int cell_index) { return dofmaps[i]->cell_dofs(cell_index); };
```
francesco-ballarin: Will any resonable compiler be able to avoid creating a copy of the temporary returned by `dofmaps[i]->cell_dofs(cell_index)`?

garth-wells: May be use https://en.cppreference.com/w/cpp/utility/functional/mem_fn?

I would like ultimately to get just pass a single array of dofs (maybe with on offset array to handle a variable numbers) and pass the single array.

francesco-ballarin: Not sure how to use mem_fn. Since it seems to be designed to only work with class members, I guess I would put it around the inner call to `dofmaps[i]->cell_dofs(cell_index)`, but the (possible) copy is happening in the outer layer.

francesco-ballarin: The links in the first answer at https://stackoverflow.com/questions/41118520/when-is-return-by-value-ok might be relevant to my original question.

garth-wells:
This feels a bit complicated to me, especially in view of the code in `SparsityPaternBuilder` being rather simple (the more complicated steps are in ``SparsityPattern`). 

@francesco-ballarin Could you easily call the insertions into the `SparsityPattern` yourself?

francesco-ballarin:
> @francesco-ballarin Could you easily call the insertions into the SparsityPattern yourself?

Yes, and/but this is actually what I am doing right now with the duplication of the SparsityPaternBuilder code, which led me to open #657

garth-wells:
I suggest we close this. The SparsityPattern(Builder) has been simplified quite a bit and one can work directly with a `SparsityPattern` object for more complicated maps.

PR #662, opened 2019-11-22
Simplify interface for building a SparsityPattern from dofmaps

garth-wells:
Simplify sparsity pattern construction from dofmap data.

This change is consistent with the move towards working with data rather than class methods.

@francesco-ballarin would this work for you?

### DIFF
--- a/cpp/dolfin/fem/DiscreteOperators.h
+++ b/cpp/dolfin/fem/DiscreteOperators.h
@@ -40,9 +40,9 @@ class DiscreteOperators
 {
 public:
   /// Build the discrete gradient operator A that takes a
-  /// \f$w \in H^1\f$ (P1, nodal Lagrange) to \f$v \in H(curl)\f$
-  /// (lowest order Nedelec), i.e. v = Aw. V0 is the H(curl) space,
-  /// and V1 is the P1 Lagrange space.
+  /// \f$w \in H^1\f$ (P1, nodal Lagrange) to \f$v \in H(curl)\f$ (lowest order
+  /// Nedelec), i.e. v = Aw. V0 is the H(curl) space, and V1 is the P1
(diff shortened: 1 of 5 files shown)


francesco-ballarin:
Hi, I am afraid it would not work for my cases, because the values of the inputs dim0 and dim1 would need to be cell dependent.

garth-wells:
@francesco-ballarin that was a bit it a shortcut on my part. The DofMap code should really store the offset for each cell.

What if an offset array for each cell is passed rather than `dim`?

francesco-ballarin:
To avoid managing the offsets I actually store what will be the output of `BlockDofMap::cell_dofs(cell_index)` in a `std::map<PetscInt, std::vector<PetscInt>>`, where the map key is the cell number and the value will be wrapped in a `Eigen::Map` container on return. I should probably change the value type to be an Eigen vector, as you did in several parts of the code. I am using a map rather than a vector because there are cases in which I expect most of cell indices to return an empty cell_dofs, but I guess I can change map to vector as well.

In the standard dolfinx case, my approach would sort of be storing your private member `DofMap::_dofmap` as a matrix, where the first index is the cell number, and have `DofMap::cell_dofs(cell_index)` return the `cell_index`-th row than a segment. I cannot do that because not all rows are necessarily of the same length.

I can try and change my design to store `_dofmap` as a vector, and handle offsets. Apart from the map key lookup, what would you say disadvantages of my storing the dofmap in such a "matrix" format are?

Would the change in the PR mean that the cell_dofs method will be removed from the interface?

francesco-ballarin:
> Apart from the map key lookup...

please also note that before this PR I never need an equivalent of `DofMap::dof_array()`, otherwise linearizing my matrix storage would obviously be a huge disadvantage of my approach.

Issue #101, opened 2018-06-12
Performance regression in serial mesh library


blechta:
DOLFIN:

```
time python3 -c"from dolfin import *; UnitCubeMesh(MPI.comm_world, 100, 100, 100)"

real    0m0.867s
user    0m0.761s
sys    0m0.106s
```

DOLFIN-X:

```
time python3 -c"from dolfin import *; UnitCubeMesh(MPI.comm_world, 100, 100, 100)"

real    0m15.661s
user    0m14.440s
sys    0m1.223s
```

chrisrichardson:
This is due to the serial implementation using the same code as in parallel. For example, it calculates the dual graph for partitioning the mesh (not needed in serial).

jorgensd:
Confirming that this is still an issue:
DOLFINx
```
python3 -c"import dolfinx; from mpi4py import MPI; dolfinx.UnitCubeMesh(MPI.COMM_WORLD, 100, 100, 100)"

real	0m11.179s
user	0m9.546s
sys	0m1.611s
```
DOLFIN:
```
fenics@f596c6c6e934:/root/shared$ time python3 -c"from dolfin import *; UnitCubeMesh(MPI.comm_world, 100, 100, 100)"

real	0m1.046s
user	0m0.907s
sys	0m0.140s
```

IgorBaratta:
To avoid computing the dual graph in serial we can call a custom partitioner, which sets the destination of all cells to process 0 (in serial): 
```python
from mpi4py import MPI
import dolfinx
import numpy


def partitioner(mpi_comm, nparts, tdim, cells, ghost_mode):
    dest = numpy.zeros(cells.num_nodes, dtype=numpy.int32)
    return dolfinx.cpp.graph.AdjacencyList_int32(dest)


mesh = dolfinx.UnitCubeMesh(
    MPI.COMM_WORLD, 100, 100, 100, partitioner=partitioner)
```

**with custom partitioner:**

> real    0m6.091s
> user    0m4.249s
> sys     0m2.458s
```
[MPI_AVG] Summary of timings                                   |  reps  wall avg  wall tot
------------------------------------------------------------------------------------------
Build BoxMesh                                                  |     1  4.254135  4.254135
Build dofmap data                                              |     1  1.060930  1.060930
Compute SCOTCH graph re-ordering                               |     1  0.141512  0.141512
Compute dof reordering map                                     |     1  0.666798  0.666798
[...]

PR #666, opened 2019-11-25
Avoid some MPI calls for mesh in serial

IgorBaratta:
Proposal to fix Issue #101 .

```
from dolfin.common import list_timings, TimingType
from dolfin import *

mesh = UnitCubeMesh(MPI.comm_world, 100, 100, 100)
list_timings([TimingType.wall])
```
Time to build UnitCubeMesh in serial:
- Dolfin-x (computing facet-cell connections):  25.31 s
- This branch (computing facet-cell connections): 7.6271 s
- This branch  (without computing facet-cell connections): 0.54866 s
- Dolfin (without computing facet-cell connections):  0.55464 s

Should we initialize facet-cell connections by default?

### DIFF
--- a/cpp/dolfin/mesh/Mesh.cpp
+++ b/cpp/dolfin/mesh/Mesh.cpp
@@ -163,6 +163,23 @@ compute_point_distribution(
         points,
     mesh::CellType type)
 {
+  int mpi_size = dolfin::MPI::size(mpi_comm);
+
+  if (mpi_size == 1)
+  {
+    std::map<std::int32_t, std::set<int>> shared_points;
+    std::vector<std::int64_t> local_to_global(points.rows());
+    std::iota(local_to_global.begin(), local_to_global.end(), 0);
+    std::array<int, 4> num_vertices_local;
+    num_vertices_local.fill(points.rows());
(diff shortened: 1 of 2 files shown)


chrisrichardson:
I'm planning some major changes to `compute_point_distribution` to use the IndexMap for vertices, so some of this will become redundant.
I agree there is no point calling the partitioner in serial, but on the whole, I don't really feel the need to have a lot of special cases for serial. It is probably cleaner just to use the same code path for all cases, where possible... maybe we can hold on to these ideas a bit longer, and apply after merging `chris/topology-use-indexmap`.

IgorBaratta:
> I'm planning some major changes to `compute_point_distribution` to use the IndexMap for vertices, so some of this will become redundant.
> I agree there is no point calling the partitioner in serial, but on the whole, I don't really feel the need to have a lot of special cases for serial. It is probably cleaner just to use the same code path for all cases, where possible... maybe we can hold on to these ideas a bit longer, and apply after merging `chris/topology-use-indexmap`.

I agree. I'll close the PR, and update the issue with some data. 
Since part of the timing difference is not only due to the execution of parallel code for serial mesh, but also the computation of facet-cell connections.

PR #667, opened 2019-11-26
Fixing issue 350 and a typo

jorgensd:
Removing `has_cell_centred` check, as it is not working as it shoud (Issue 350).
Changing `XDMF.write` signature to have a keywordargument `cell_centred`, default is False.
```
f.write(v, cell_centred=True)
f.write(v)
f.write(v, 0, cell_centred=True)
f.write(v, 0)
```


cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -69,15 +69,11 @@ std::string rank_to_string(std::size_t value_rank)
...
-  int cell_based_dim = 1;
-  for (int i = 0; i < u.value_rank(); i++)
-    cell_based_dim *= u.function_space()->mesh()->topology().dim();
-
   assert(u.function_space());
-  assert(u.function_space()->dofmap());
-  assert(u.function_space()->dofmap()->element_dof_layout);
-  return (u.function_space()->dofmap()->element_dof_layout->num_dofs()
-          == cell_based_dim);
+  return (u.function_space()->element()->family().compare(
```
garth-wells: Signatures are for user convenience and shouldn't be be relied upon internally (this caused a lot of pain in the past). `degree` should also be avoided.

Comparing `ElementDofLayout::num_entity_dofs(tdim)` and `ElementDofLayout::num_dofs()` should work.

cpp/dolfin/io/XDMFFile.cpp
```diff
@@ -69,15 +68,13 @@ std::string rank_to_string(std::size_t value_rank)
...
   assert(u.function_space());
   assert(u.function_space()->dofmap());
-  assert(u.function_space()->dofmap()->element_dof_layout);
-  return (u.function_space()->dofmap()->element_dof_layout->num_dofs()
-          == cell_based_dim);
+  std::shared_ptr<const fem::ElementDofLayout> dof_layout
+      = u.function_space()->dofmap()->element_dof_layout;
+  assert(u.function_space()->mesh());
+  const int tdim = u.function_space()->mesh()->topology().dim();
+  return dof_layout->num_entity_dofs(tdim) == dof_layout->num_dofs();
```
garth-wells: @jorgensd This will return true for any DG element, not just DG0.

jorgensd:
Should be reconsidered after #916

PR #669, opened 2019-11-27
Lower-level DirichletBC

michalhabera:
This is a proposal to lower the level of strong BC application.

**UPDATE:**
This PR proposes
`DirichletBC(g, dofs, V=optional)`
- `g` a boundary value function
- `dofs` are local indices of degrees of freedom bcs are applied
- `V`, optional,  a function space of the problem to which the boundary condition is applied. If `V` is passed then `dofs` must be an array of shape (num_dofs, 2), first column representing dof indices in the space of `g`, second column being dofs in the space `V`.

Helper method `locate_dof_topological` has an optional bool argument `remote` to specify if it should locate also remote dofs. Remote dofs are those which current processor owns, but are not marked by mesh entities.

---
**Motivation**
In the past, `DirichletBC` was responsible for transforming some geometrical (or physical) information into indices of degrees of freedom, their storage and application in the assembler. However, this logic for extraction of dof indices could be used in much broader context - not related to DirichletBC (e.g. extracting a function on a marked subdomains, custom assembler matrix changes, ...)
In addition, the flexibility of DirichletBC was limited to `geometric/topological/pointwise` methods.

**New API proposal**
This PR proposes
`DirichletBC(V, g, dofs_V, dofs_g = dofs_V)`
with `V, g` the same as in the old interface:
- `V` a function space where the BC is applied,
- `g` a boundary value function which must have the same mesh and finite element as the space `V`,
- `dofs_V` are local indices of degrees of freedom in the space `V` where bcs are applied,
- `dofs_g` are local indices of degrees of freedom in the space of `g` which is used as a value for the boundary condiiton.

There are 2 helper methods provided to extract dof indices from a function space:
- `locate_dofs_topological(V, mesh_entity_dim, entities_idcs)`
    which finds degrees of freedom which topologically belong to mesh entities of specified dimension and indices. This is equivalent to the current dolfinx/master DirichletBC implementation.
- `locate_dofs_geometrical(V, marking_function)` which tabulates positions of dofs of the `V` space and returns those matched by the marking function. This is equivalent to the old pointwise method of DirichletBC.

Note: Currently, `dofs_V` and `dofs_g` must have the same shape as an arrays, but they do not need to have identical values - this is the case for `g` being a sub-function of the `V` space, see block assembly tests. For brevity reasons `dofs_g` is set to `dofs_V` if not provided - which is the most common use case in non-block assembly.

Note 2: For topological approach, we have to produce a dummy `MeshFunction` and use its `mark()` method to identify facets/points given a marking function. This can be in future simplified to avoid additional MeshFunction as an intermediate - but in general, a way how indices of facets/... are obtained is totally up to user.


cpp/dolfin/fem/DirichletBC.cpp
```diff
@@ -495,125 +234,84 @@ void DirichletBC::dof_values(
...
+{
+
+  const Eigen::Array<double, 3, Eigen::Dynamic, Eigen::RowMajor> dof_coordinates
+      = V.tabulate_dof_coordinates().transpose();
+  const Eigen::Array<bool, Eigen::Dynamic, 1> marked_dofs
+      = marker(dof_coordinates);
+
+  std::vector<PetscInt> dofs;
+
+  for (PetscInt i = 0; i < marked_dofs.size(); ++i)
```
chrisrichardson: What's this got to do with Petsc?

michalhabera: What type do you suggest? `std::int32_t` since this are just local dof indices?

garth-wells:
Considering the interface `DirichletBC(V, g, dofs_V, dofs_g = dofs_V)`, are `V` and `g` needed in their entirety , or is just some small subset of their functionality? It feels a bit over-elaborated given the the dof indices are being passed.

michalhabera:
@garth-wells The function space `V` is used to get the index map to fetch get remote bcs. The function `g` is used for 1. its function space for the same purpose as `V` and also as a vector of BCs values which basically attaches values to the passed dof indices.

Personally, I think that passing index maps is too low-level.

cpp/dolfin/fem/DirichletBC.cpp
garth-wells: It should be whatever type ` marked_dofs.size()` is.

garth-wells:
@michalhabera The 'low-levelness' of the arguments seems unbalanced to me. The function space and Function are rich and high-level and only a small part of them is required, whereas the last two arguments are very low level. We should try to balance the arguments, and if helpful a hierarchy of interfaces.

PR #671, opened 2019-11-28
Simplify form creation

chrisrichardson:
* Keeps the malloc/free out of user code, by passing a function pointer to the create function.


cpp/dolfin/fem/utils.cpp
```diff
@@ -512,6 +512,17 @@ fem::get_constants_from_ufc_form(const ufc_form& ufc_form)
   return constants;
 }
 //-----------------------------------------------------------------------------
+std::shared_ptr<fem::Form> fem::create_form(
+    ufc_form* (*fptr)(void),
```
garth-wells: What is `fptr`?

chrisrichardson: `/// @param[in] fptr pointer to a function returning a pointer to ufc_form.`

garth-wells: It's a bit cryptic.

I could also mention that the function it points to takes no arguments.

chrisrichardson: That should be clear from the signature `(void)` - but I admit I don't really like C-style function pointer signatures, as they are... cryptic.

mrambausek: Use std::function<ufc_form*(void)> as argument type and rely on implicit conversion when a C-style function is provided (no difference for caller)?

PR #673, opened 2019-11-28
Check constants and coefficients are set before assembly

chrisrichardson:
Simple check for `nullptr` on all constants and coefficients as they are fetched before assembly.


chrisrichardson:
It's not a costly check, and the assembler will definitely fail if they are not set.

chrisrichardson:
How about we have a utility function `pack_constants` (similar to `pack_coefficients`) which moves this repeated code into `utils.cpp`.

garth-wells:
@chrisrichardson I think there should be a user function to allow a user to also check that coefficients are set.

massimiliano-leoni:
A similar functionality is implemented in `FormCoefficients::offsets()`. How does this relate to that?

cpp/dolfin/fem/utils.cpp
```diff
@@ -643,21 +643,45 @@ fem::create_functionspace(ufc_function_space* (*fptr)(void),
   return V;
 }
 //-----------------------------------------------------------------------------
+void fem::check_coefficients(const fem::Form& form)
```
garth-wells: This function name doesn't say anything much. It should be something like:
```
std::set<foo> unset_constants
```
or 
```
bool coefficients_all_set();
```
(and likely a member of `FormCoefficients`).

cpp/dolfin/fem/utils.cpp
```diff
@@ -643,21 +643,45 @@ fem::create_functionspace(ufc_function_space* (*fptr)(void),
...
+{
+  const fem::FormCoefficients& coefficients = form.coefficients();
+  for (int i = 0; i < coefficients.size(); ++i)
+  {
+    auto coeff = coefficients.get(i);
```
garth-wells: Don't use `auto` here.

cpp/dolfin/fem/utils.cpp
```diff
@@ -643,21 +643,45 @@ fem::create_functionspace(ufc_function_space* (*fptr)(void),
...
+    if (!coeff)
+    {
+      throw std::runtime_error("Coefficient " + std::to_string(i)
```
garth-wells: This function shouldn't throw. Let the caller decide how to handle.

cpp/dolfin/fem/assemble_matrix_impl.cpp
```diff
@@ -41,6 +41,9 @@ void fem::impl::assemble_matrix(Mat A, const Form& a,
   const int num_dofs_per_cell1 = dofmap1.element_dof_layout->num_dofs();
 
   // Prepare constants
+  if (a.list_unset_constants().size() > 0)
```
garth-wells: `if (!a.list_unset_constants().empty())`

chrisrichardson: I considered it. "Not unset empty" is a triple negative, and harder to make sense of.

garth-wells: Not sure I agree about the triple - parses easy enough for me. `empty` is generally preferred because it is defined (efficiently) for more STL containers.

PR #654, opened 2019-11-19
Permute the dofs on quads and hexes

mscroggs:
This pull request adds the necessary permutations to `dof_permutation.cpp` for quads and hexes.

Currently DG (https://github.com/FEniCS/dolfinx/issues/645) and vector function spaces (https://github.com/FEniCS/dolfinx/issues/31) do not work unless the mesh is ordered, `Ordering::order_simplex` should be called in these cases.

PR #674, opened 2019-11-28
Improving higher order quad tests

jorgensd:
With #654 , the higher order quad tests can be extended to multiple cells.
Also fixing a bug in VTKWriter that would cause celltype 70 to be written as F.


cpp/dolfin/io/VTKWriter.cpp
```diff
@@ -246,7 +246,7 @@ void write_ascii_mesh(const mesh::Mesh& mesh, int cell_dim,
        << "ascii"
        << "\">";
   for (int types = 0; types < num_cells; types++)
-    file << vtk_cell_type << " ";
+    file << int(vtk_cell_type) << " ";
```
garth-wells: Is this cast needed?

jorgensd: Without the cast, this yields:
```
<DataArray  type="UInt8"  Name="types"  format="ascii">F F </DataArray>
```
For vtk_cell_type=70

garth-wells: Declare it as `int` in the first place, or try `std::to_string`.

The `int` cast is something that someone could easily remove later because it's not obvious why it's needed.

jhale: It is best not to write non-string data to `std::cout` the results can be... unexpected. There is `boost::lexical_cast` and the slightly more basic but still fine `std::to_string`.

PR #665, opened 2019-11-25
Dockerfile adjustments

jhale:
Changes:

1) dev-env images are built twice a week. => In circleci trigger `biweekly`.
2) actual builds of dolfinx are built nightly => In new circleci trigger `daily`.
3) User can easily build current dolfinx from `master` using `ONBUILD` feature => New `real-onbuild` and `complex-onbuild` targets, usage: `FROM quay.io/fenicsproject/dolfinx:complex-onbuild` at start of personal `Dockerfile`.
4) Split `dev-env` and finalised images (e.g. `real`) into two `Dockerfiles`.

PR #680, opened 2019-12-03
Install pybind11 casters

### LINKED ISSUE #656: Should pybind11 casters be installed?
Dear all,
while working on an external library (which relies on pybind11 for wrapping C++ code on the fly), I need to use MPI and PETSc casters present in `python/src`. Unfortunately, those casters do not get installed with the rest of dolfinx.
While I can easily workaround this by copying those casters in my library (as I am doing right now), I am opening this issue to inquire whether you think that they should get installed as part of the `python/setup.py`, so that external projects need not to duplicate them.

Thanks,
Francesco

(discussed today on this with @jhale)

francesco-ballarin:
Closes #656


python/setup.py
```diff
@@ -77,6 +77,7 @@ def build_extension(self, ext):
             os.makedirs(self.build_temp)
         subprocess.check_call(['cmake', ext.sourcedir] + cmake_args, cwd=self.build_temp, env=env)
         subprocess.check_call(['cmake', '--build', '.'] + build_args, cwd=self.build_temp, env=env)
+        subprocess.check_call(['cmake', '--build', '.', '--target', 'install'] + build_args, cwd=self.build_temp, env=env)
```
francesco-ballarin: In very recent cmake versions (>= 3.15) this can be replaced by `cmake --install .`. I probably would not want to force using such a recent cmake version just for this line.

python/test/unit/common/test_mpi.py
```diff
@@ -21,3 +27,55 @@ def test_mpi_comm_wrapper():
...
     assert isinstance(w1, MPI.Comm)
     assert isinstance(w2, MPI.Comm)
+
+
+def test_mpi_comm_wrapper_cppimport(tempdir):  # noqa: F811
+    """
+    Test MPICommWrapper <-> mpi4py.MPI.Comm conversion for code compiled with cppimport
+    """
+
+    cppimport = pytest.importorskip("cppimport")
```
francesco-ballarin: I am not sure how to change the circle CI configuration to make sure that cppimport is installed. Should I change the .yml file in .circleci, or the docker file?

I have tested them locally, but current CI run might skip the two new tests.

python/test/unit/common/test_mpi.py
```diff
@@ -21,3 +27,55 @@ def test_mpi_comm_wrapper():
...
+
+    @mpi_jit_decorator
+    def compile_module():
+        cpp_code_header = f"""
+        <%
+        setup_pybind11(cfg)
+        cfg['include_dirs'] += {dolfin_pc["include_dirs"] + [mpi4py.get_include()]}
+        cfg['compiler_args'] += {["-D" + dm for dm in dolfin_pc["define_macros"]]}
+        cfg['libraries'] += {dolfin_pc["libraries"]}
```
francesco-ballarin: This is probably overkill, as neither dolfin nor its dependencies are not needed here. Let me know if you want me to remove this.

python/test/unit/la/test_casters.py
```diff
@@ -0,0 +1,75 @@
...
+    cppimport = pytest.importorskip("cppimport")
+
+    @mpi_jit_decorator
+    def compile_module():
+        cpp_code_header = f"""
+        <%
+        setup_pybind11(cfg)
+        cfg['include_dirs'] += {dolfin_pc["include_dirs"] + [petsc4py.get_include()]}
+        cfg['compiler_args'] += {["-D" + dm for dm in dolfin_pc["define_macros"]]}
+        cfg['libraries'] += {dolfin_pc["libraries"]}
```
francesco-ballarin: This is probably overkill as well, as only the petsc libraries are needed, but not dolfin itself or the other dependencies. However, rather than fiddling with PETSc's pc, I'd leave it as is for future users that may need to use DOLFIN classes as well.

python/test/unit/common/test_mpi.py
chrisrichardson: It should be fine to install in the .circleci config for now.

jhale: Best to put it here: https://github.com/FEniCS/dolfinx/blob/master/Dockerfile#L54

francesco-ballarin: Yes great, will put in Dockerfile. Do you want me to wait after #665 is merged to avoid conflicts? Even, if I were to change that line now, I would still need to wait till one of the two days of bi-weekly target to see any effect, right?

jhale: I can manually trigger a build of the docker images, so go ahead.

jhale: Merge conflict is trivial, don't worry about it.

python/test/unit/common/test_mpi.py
garth-wells: If it can be removed that would be good.

garth-wells:
@francesco-ballarin Could you run `isort` (https://pypi.org/project/isort/) on the Python files?

garth-wells:
We going to need to wind this back. It installs DOLFIN pybind11 code into the DOLFIN CPP include path, which is a problem because: (a) it mixes the Python and C++ installs; (b) the person installing the Python bindings might not have write permission to the DOLFIN C++ installed include paths; and (c) getting the C++ include path relies on old CMake commands to we should remove.

The headers should be installed somewhere inside the DOLFIN Python module tree, like mpi4py, etc.

PR #681, opened 2019-12-03
Fix behaviour of timings, list_timings

chrisrichardson:
After recent changes, these were ignoring their input arguments.
Allow selection of subset of timing types again.

### DIFF
--- a/cpp/dolfin/common/TimeLogger.cpp
+++ b/cpp/dolfin/common/TimeLogger.cpp
@@ -59,17 +59,31 @@ Table TimeLogger::timings(std::set<TimingType> type)
 {
   // Generate log::timing table
   Table table("Summary of timings");
+
+  bool time_wall = (type.find(TimingType::wall) != type.end());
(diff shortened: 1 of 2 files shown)


cpp/dolfin/common/TimeLogger.cpp
```diff
@@ -59,17 +59,31 @@ Table TimeLogger::timings(std::set<TimingType> type)
 {
   // Generate log::timing table
   Table table("Summary of timings");
+
+  bool time_wall = (type.find(TimingType::wall) != type.end());
```
garth-wells: Are the outer braces necessary?

chrisrichardson: I don't know, but I think it looks better. `a = b != c;` looks a bit odd.

garth-wells: Looks ok to me. We do `return a ==b` rather than `return (a ==b)`.

PR #685, opened 2019-12-11
Refactoring dof_ordering

mscroggs:
I have restructured the code that calculates the number of times each permutation should be applied to each cell.

In the old version, similar code was copied and pasted for each cell type, with the entities hardcoded for each type. In the new version, this information in inferred from the `CellType` object leading to more general (and less) code.


cpp/dolfin/fem/dofs_permutation.cpp
```diff
@@ -429,233 +429,152 @@ hexahedron_rotations_and_reflection(const int volume_dofs, const int blocksize)
...
+  entity_count += 1;
+
+  // Get lists of vertices on each entity from the cell type
+  std::vector<std::pair<mesh::CellType, Eigen::Array<int, 1, Eigen::Dynamic>>>
+      entities(entity_count);
+  { // scope for j
+    int j = 0;
+    for (int dim = 1; dim < t_dim; ++dim)
+    {
+      auto vertices = mesh::get_entity_vertices(type, dim);
```
garth-wells: Best to replace most of the `auto` to improve readability.

cpp/dolfin/fem/dofs_permutation.cpp
```diff
@@ -429,233 +429,152 @@ hexahedron_rotations_and_reflection(const int volume_dofs, const int blocksize)
...
+      auto e_type = mesh::cell_entity_type(type, dim);
+      for (int i = 0; i < mesh::cell_num_entities(type, dim); ++i)
+      {
+        // Store the cell type and vertices associated with the ith
+        // entity of dimension dim
+        entities[j].first = e_type;
```
garth-wells: Try
```
entities[j] = {e_type,  vertices.row(i)};
```
If that doesn't work,
```
entities[j] = std::pair(e_type,  vertices.row(i));
```

cpp/dolfin/fem/dofs_permutation.cpp
```diff
@@ -429,233 +429,152 @@ hexahedron_rotations_and_reflection(const int volume_dofs, const int blocksize)
...
+  int entity_count = 0;
+  for (int dim = 1; dim < t_dim; ++dim)
+    entity_count += mesh::cell_num_entities(type, dim);
+  entity_count += 1;
+
+  // Get lists of vertices on each entity from the cell type
+  std::vector<std::pair<mesh::CellType, Eigen::Array<int, 1, Eigen::Dynamic>>>
+      entities(entity_count);
+  { // scope for j
+    int j = 0;
```
garth-wells: Can the need for `j` be removed by using `push_back`?

cpp/dolfin/fem/dofs_permutation.cpp
```diff
@@ -429,233 +429,152 @@ hexahedron_rotations_and_reflection(const int volume_dofs, const int blocksize)
...
-  }
+    int j = 0;
+    // iterate over the cell's entities
+    for (std::size_t e_n = 0; e_n < entities.size(); ++e_n)
+    {
+      // Look up the cell type of the entity, then calculate the orders
+      // for that entity type
+      auto cell_type = entities[e_n].first;
+      auto v = entities[e_n].second;
+      switch (cell_type)
```
garth-wells: Not ideal to have a switch in the inner most loop. Is there another way?

mscroggs: In 48ad08512e7171016f69ee6ad6931358b993c99b, I've moved the switch outside the loop and made it store the function needed rather than the cell type

cpp/dolfin/fem/dofs_permutation.cpp
mscroggs: I've replaced most, but left a couple in where the type is very long so I think would hurt readability (eg `std::function<Eigen::Array<std::int8_t, 1, Eigen::Dynamic>Eigen::Array<std::size32_t, 1, Eigen::Dynamic>)>`)

cpp/dolfin/fem/dofs_permutation.cpp
mscroggs: The first worked

cpp/dolfin/fem/dofs_permutation.cpp
```diff
@@ -35,155 +35,208 @@ int get_num_permutations(mesh::CellType cell_type)
...
+calculate_point_orders(const Eigen::Array<std::int64_t, 1, 1>& vs)
+{
+  return {};
+}
+//-----------------------------------------------------------------------------
+/// Calculates whether or not an interval should be flipped
+/// @param[in] vs The global vertex numbers of the interval's vertices
+/// @return The reflection order for the interval
+Eigen::Array<std::int8_t, 1, Eigen::Dynamic>
+calculate_interval_orders(const Eigen::Array<std::int64_t, 1, 2>& vs)
```
garth-wells: These function will need to take a `Eigen::Array<std::int8_t, 1, Eigen::Dynamic>` as an input argument for performance reasons. See comment below.

cpp/dolfin/fem/dofs_permutation.cpp
```diff
@@ -429,233 +482,110 @@ hexahedron_rotations_and_reflection(const int volume_dofs, const int blocksize)
...
+    {
+      // Use the functions for each entities and that entity's global
+      // vertex numbers to calculate the number of times each
+      // permutation should be applied on this cell
+      auto f = entities[e_n].first;
+      Eigen::Array<int, 1, Eigen::Dynamic> v = entities[e_n].second;
+      Eigen::Array<std::int64_t, 1, Eigen::Dynamic> global_v(v.size());
+      for (int i = 0; i < v.size(); ++i)
+        global_v[i] = cell_vs[v[i]];
+      Eigen::Array<std::int8_t, 1, Eigen::Dynamic> orders = f(global_v);
```
garth-wells: This is a very likely performance problem because of the dynamic memory allocation to create `orders`. Change `f` to take `orders` as an in/out argument.

cpp/dolfin/fem/dofs_permutation.cpp
```diff
@@ -429,233 +482,110 @@ hexahedron_rotations_and_reflection(const int volume_dofs, const int blocksize)
...
+    int j = 0;
+    // iterate over the cell's entities
+    for (std::size_t e = 0; e < entities.size(); ++e)
+    {
+      // Use the functions for each entities and that entity's global
+      // vertex numbers to calculate the number of times each
+      // permutation should be applied on this cell
+      auto f = entities[e].first;
+      Eigen::Array<int, 1, Eigen::Dynamic> v = entities[e].second;
+      Eigen::Array<std::int64_t, 1, Eigen::Dynamic> global_v(v.size());
```
garth-wells: Need to move this memory allocation outside loop.

garth-wells: Might be possible to keep this by telling Eigen the max size, see https://eigen.tuxfamily.org/dox/group__TutorialMatrixClass.html#title9.

mscroggs:
Closing this as is I've added an improved version of it to https://github.com/FEniCS/dolfinx/tree/mscroggs/arrangement_types

PR #690, opened 2020-01-04
Remove python build thread detection

jhale:
On e.g. a 28-core node in a cluster, this sets 28 build threads, even if the scheduler has only reserved a single core. This often results in failed builds due to out-of-memory errors. Propose to remove, or replace with something more sensible.


jhale:
I did a bit more investigating. It seems like the correct way to set this in cmake >= 3.12:

https://cmake.org/cmake/help/v3.12/manual/cmake.1.html#build-tool-mode

1) Using the `--parallel` option to `cmake --build`. In the case is is not passed, the default for the tool is used (e.g. 1).
1) Via the `CMAKE_BUILD_PARALLEL_LEVEL` environment variable.

So I would propose make the default as 3 in all cases, and let the user set `CMAKE_BUILD_PARALLEL_LEVEL` environment variable if needed.

michalhabera:
I agree, this failed my builds on UNI LU HPC cluster and took me a day to figure out. Default to 3, let user set otherwise via ENV makes sense.

jhale:
Will have to revisit this when we have newer cmake. Just sticking with three seems like a sensible default for 99% of situations.

PR #634, opened 2019-11-13
Added DG test

jpdean:
The test is based on test_fem_pipeline.py.

PR #696, opened 2020-01-10
Pass a permutation into FFC so that it can permute the quadrature points.

mscroggs:
Pass an integer into ffc to tell it how to make two sides of a facet match.
Added the DG test (https://github.com/FEniCS/dolfinx/pull/634) as this now passes.

For facet integrals, two numbers (a permutation for each cell attached to a facet) will be passed into `tabulate_tensor`'s `quadrature_permutation` argument.

For 2D cells, the permutation number will be 0 if no permutation is needed, 1 is a reflection is needed. For 3D cells, N%2 gives the number of reflections needed and floor(N/2) gives the number of rotations needed. `MeshEntity::facet_permutation` computes this number for a facet.

This pull request is coupled to https://github.com/FEniCS/ffcx/pull/188.


mscroggs:
Closing this as the changes here are a subset of those in https://github.com/FEniCS/dolfinx/pull/702

PR #698, opened 2020-01-17
Fix embedded manifold BC lifting

michalhabera:
This bug caused meshes with `tdim < gdim` fail in DirichletBC lifting.


garth-wells:
Could you add a test that would pick this issue up?

michalhabera:
@garth-wells Added test with handwritten 1-D mesh in 2-D space.

PR #701, opened 2020-01-23
Fix reading interval mesh 2nd order

michalhabera:
There was a permutation missing for the general order case.


michalhabera:
@jorgensd Ok, I see the point. However, lets not put that assumption there... The cost of the inverse computation is negligible, and user has just one place where to change the code if dolfin-to-vtk ordering changes.

PR #702, opened 2020-01-24
Quadrature point permuting for facets and vector dof directions

mscroggs:
This PR containts all https://github.com/FEniCS/dolfinx/pull/696 plus stuff to make non-scalar dofs work on non ordered meshes.

>Pass an integer into ffc to tell it how to make two sides of a facet match.
>Added the DG test (https://github.com/FEniCS/dolfinx/pull/634) as this now passes.
>
>For facet integrals, two numbers (a permutation for each cell attached to a facet) will be passed into `tabulate_tensor`'s `quadrature_permutation` argument.
>
>For 2D cells, the permutation number will be 0 if no permutation is needed, 1 is a reflection is needed. For 3D cells, N%2 gives the number of reflections needed and floor(N/2) gives the number of rotations needed. `MeshEntity::facet_permutation` computes this number for a facet.
> (from https://github.com/FEniCS/dolfinx/pull/696)

Additionally, this passes bools into ffc when transforming basis functions from the reference element to tell FFC which entities need to be reflected to match their low->high orientations. FFC can then multiply and dofs that represent vectors by -1 on these entities so that the direction of vector dofs is consistent.

This pull request is coupled with https://github.com/FEniCS/ffcx/pull/190

Current status (checked items are working):
- Triangles
- [x] Quadrature permuting (for DG)
- [x] RT
- [x] BDM
- [x] N1curl
- [x] N2curl
- Tetrahedra
- [x] Quadrature permuting (for DG)
- [x] RT
- [x] BDM
- [ ] N1curl
- [x] N2curl (lowest order)
- [ ] N2curl (higher order)
- Quadrilaterals
- [x] Quadrature permuting (for DG)
- Hexahedra
- [x] Quadrature permuting (for DG)


cpp/dolfin/fem/assemble_matrix_impl.cpp
```diff
@@ -225,6 +228,7 @@ void fem::impl::assemble_exterior_facets(
     // Get local index of facet with respect to the cell
     const int local_facet = cell.index(facet);
     const int orient = 0;
+    const int perm = cell.facet_permutation(facet);
```
michalhabera: `facet_permutation()` is quite heavy in logic and contains loads of ifs and branches. I think you should rather precompute `const int perm` for possible connectivities and store somewhere. Assemblers should rather just access the contiguous data, not do such computations. I am sure @garth-wells would agree.

mscroggs: Agreed, I'll work on getting this and `Mesh::get_entity_reflections` (https://github.com/FEniCS/dolfinx/blob/b6fee46adada0eaedae7c178871dc61fcb4771be/cpp/dolfinx/mesh/Mesh.cpp#L398-L434) precomputed

cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -394,6 +395,44 @@ const Geometry& Mesh::geometry() const
...
 //-----------------------------------------------------------------------------
+bool* Mesh::get_entity_reflections(int cell_n) const
+{
+  // FIXME: cache this somewhere to avoid computing it every time its needed
+  const int tdim = _topology->dim();
+  int entities_per_cell = 1;
+  for (int d = 0; d < tdim; ++d)
+    entities_per_cell += cell_num_entities(_cell_type, d);
+
+  bool* entity_reflections = new bool[entities_per_cell];
```
garth-wells: Avoid using `new`. It leads to memory leaks.

cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -394,6 +395,44 @@ const Geometry& Mesh::geometry() const
   return *_geometry;
 }
 //-----------------------------------------------------------------------------
+bool* Mesh::get_entity_reflections(int cell_n) const
```
garth-wells: Returning a raw pointer will cause a leak.

garth-wells: Is the a property of the `Mesh`, or just a single `Cell`? Or does it belong to `Topology`?

Should it be computed for all cells at once?

cpp/dolfinx/mesh/MeshEntity.h
```diff
@@ -70,10 +70,33 @@ class MeshEntity
...
+    for (int v = 0; v < num_entities(0); ++v)
+      if (vertices[v] == v_index)
+        return v;
+    throw std::runtime_error("Vertex was not found");
+  }
+
+  /// Returns the local indices of vertex on the given entity
+  /// @param[in] entity The entity
+  /// @return The local indices
+  int* get_local_vertex_indices(const MeshEntity& entity) const
```
garth-wells: This will cause a leak.

mscroggs: I've removed this function

mscroggs:
`entity_reflections` and `entity_permutations` are now precomputed and stored in `Topology`. This removes any need for the `new`s or returning of pointers

python/test/unit/fem/test_fem_pipeline.py
```diff
@@ -109,3 +111,111 @@ def test_manufactured_poisson(degree, filename, datadir):
...
     assert np.absolute(error) < 1.0e-14
+
+
+@skip_in_parallel
+@pytest.mark.parametrize("filename", ["UnitSquareMesh_triangle.xdmf",
+                                      "UnitCubeMesh_tetra.xdmf",
+                                      # "UnitSquareMesh_quad.xdmf",
+                                      # "UnitCubeMesh_hexahedron.xdmf"
+                                      ])
+@pytest.mark.parametrize("degree", [1])
```
IgorBaratta: The RT and BDM tests share a lot of code.
Why not to parameterize the element family?

garth-wells: I think they can all be combined eventually,  but it's still a bit hit-and-miss at the moment which cases are working so it useful to have tests that are green.

mscroggs:
I believe there are only two issues remaining to fix before this is all working:

- #736: This is causing problems for `N2curl` spaces.
- #737: This is causing problems for (order>2) `N1curl` spaces

mscroggs:
Fixing the Nedelec spaces is looking to be a large job, so I've readded mesh ordering so that this pull request can be considered as the Nedelec spaces still work on ordered meshes.

I've opened issues #736, #737 and #742 to document the still ongoing work on vector function spaces.

mscroggs:
Order 2 N2Curl now works in the branch https://github.com/FEniCS/dolfinx/tree/mscroggs/move_base_perms_to_ffc

I believe that FaceTangent type dofs are the only ones still causing problems, although they appear to be least easy to fix.

PR #703, opened 2020-01-24
XDMFFile IO for MVC on higher order meshes

michalhabera:
I am a bit careful when it comes to adding more tests on XDMF IO. Do you think we need to add some, @garth-wells , @jorgensd ?


garth-wells:
@michalhabera it would be nice to speed up the XDMF tests without losing coverage.

PR #705, opened 2020-01-24
Use neighbour comm for mesh refinement

chrisrichardson:
* add some MPI wrappers for neighbour all_to_all
* replace generic all_to_all


cpp/dolfin/refinement/ParallelRefinement.cpp
```diff
@@ -29,6 +58,8 @@ ParallelRefinement::ParallelRefinement(const mesh::Mesh& mesh)
   if (!_mesh.topology().connectivity(1, 0))
     throw std::runtime_error("Edges must be initialised");
 
+  _neighbour_comm = create_neighbor_comm_via_edge(mesh);
```
garth-wells: I can't see where this communicator get destroyed.

chrisrichardson: Good point. I tend to forget they are not C++ objects with destructors.

garth-wells: Memory management of the comms can be a bit tricky. It might be possible to attached a counter each time we taker hold of one, or we can make more use of the comm wrapper.

PR #706, opened 2020-01-26
Use modern CMake 

garth-wells:
Use modern CMake. Reduced amount of hand-coded CMake.


cpp/CMakeLists.txt
```diff
@@ -1,18 +1,11 @@
+#------------------------------------------------------------------------------
 # Top level CMakeLists.txt file for DOLFIN
 cmake_minimum_required(VERSION 3.10)
```
chrisrichardson: I think this needs to change? I can't now build with cmake 3.10.2

garth-wells: Strange, the CMake version in the Docker containers running on CircleCI is 3.10.2.

What's the error message?

chrisrichardson: Sorry, my problem. Now fixed.

PR #717, opened 2020-01-29
Add H-curl test

IgorBaratta:
Addition of the first test for Hdiv/**Hcurl** elements.

This test is a simplification of a 3d test, which would take much longer.


IgorBaratta:
The test is not supposed to work unless the mesh is "judiciously" ordered.
Closing the PR. It could be reopened later after the ongoing work on permutations.

PR #730, opened 2020-02-03
Skip test if a non-dependency module is not installed.

IgorBaratta:


python/test/unit/mesh/test_higher_order_mesh.py
```diff
@@ -22,6 +19,11 @@
 from ufl import dx
 
 
+scipy = pytest.importorskip("scipy")
+sympy = pytest.importorskip("sympy")
```
garth-wells: FIAT depends on SymPy, so I'm surprised that this is needed.

python/test/unit/mesh/test_higher_order_mesh.py
```diff
@@ -22,6 +19,11 @@
 from ufl import dx
 
 
+scipy = pytest.importorskip("scipy")
```
garth-wells: I think we could make SciPy a test dependency.

python/test/unit/mesh/test_higher_order_mesh.py
IgorBaratta: Indeed, it's not really needed for SymPy.

python/test/unit/mesh/test_higher_order_mesh.py
IgorBaratta: I'm not sure how to make SciPy a test dependency.
Using requirements files, maybe? eg: `test_requirements.txt`

IgorBaratta:
Out-of-date.

PR #467, opened 2019-07-09
Add method to construct MeshValueCollection from arrays

iitrabhi:
This PR extends the functionality of MeshValueCollection class with a constructor that supports its creation from primitive arrays. The syntax is as follows:

`mvc = MeshValueCollection("size_t", mesh, dim, cells, cell_data)`

When creating a MeshValueCollection an argument specifying the type of the MeshValueCollection must be given. Allowed types are ‘int’, ‘size_t’, ‘double’ and ‘bool’. The second argument is the mesh and third is the topological dimension of the MeshValueCollection. The fourth argument is an array specifying the vertices of the mesh entity to be marked and the fifth argument is an array containing the marker of tagged entities. The length of the arrays of argument four and five must be the same. 

A unit test is defined to check the functionality of the new constructor. A demo file is also added that demonstrates the new functionality.

PR #733, opened 2020-02-04
MeshValueCollection from arrays

michalhabera:
Slightly improved version of #467 . This enables completely file-free gmsh-to-dolfinx workflow, including boundaries and subdomains tagging.


michalhabera:
@francesco-ballarin @iitrabhi I am having the test hanging in parallel. Do you see where the bug is?

garth-wells:
I did recently find some bugs in `MeshValueCollection` were is was calling topology/connectivity creation functions (which are collective) from inside a `set` function. Make sure all collective calls are in the constructor.

francesco-ballarin:
@michalhabera I believe that line 225 should read
```
for (auto& m : mesh::MeshRange(*mesh, _dim, mesh::MeshRangeType::ALL))
```
as we did in MeshFunction.h. In this way, tests pass for me in parallel.

michalhabera:
Thanks @francesco-ballarin . Fixed now without mesh iterators. Pure loop is now more explicit in iteration over ghosts too.

PR #746, opened 2020-02-09
Use  IndexMap in mesh topology

garth-wells:
This is a major simplification around topology representation and computation, and dofmap construction.

All mesh entities have an 'owner', which is handled neatly by IndexMap. This makes dofmap construction, ghosting and other operations simpler and more efficient. An AdjacencyList class has been introduced and used to store connectivity.

The creation of ghosted meshes is broken at the moment, but should be fixed soon. Making the PR now as there have been many low-level changes which get out-of-sync with other changes quickly.


cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -246,73 +260,109 @@ Mesh::Mesh(
...
+    std::map<int, int> node_index_to_vertex;
+    for (std::size_t i = 0; i < vertices.size(); ++i)
+      node_index_to_vertex.insert({vertices[i], i});
+
+    for (int i = 0; i < vertex_cols.rows(); ++i)
+      for (std::int32_t j = 0; j < num_vertices_per_cell; ++j)
+        vertex_cols(i, j)
+            = node_index_to_vertex[coordinate_nodes(i, vertex_indices[j])];
+
+    std::stringstream s;
```
michalhabera: Debugging leftover

cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -405,21 +464,29 @@ std::size_t Mesh::create_entities(int dim) const
...
+  if (_topology->connectivity(dim, 0))
+    return -1;
 
   // Create local entities
-  TopologyComputation::compute_entities(*mesh, dim);
+  const auto [cell_entity, entity_vertex, num_new_entities]
+      = TopologyComputation::compute_entities(_mpi_comm.comm(), *_topology,
+                                              _cell_type, dim);
+
+  // Mesh* mesh = const_cast<Mesh*>(this);
```
michalhabera: Remove commented code

cpp/dolfinx/fem/DofMapBuilder.cpp
```diff
@@ -344,22 +110,51 @@ DofMapStructure build_basic_dofmap(const mesh::Mesh& mesh,
...
+
+  // Storage for local-to-global map
+  std::vector<std::int64_t> local_to_global(local_size);
+
+  // Dof (dim, entity index) marker
+  std::vector<std::pair<std::int8_t, std::int32_t>> dof_entity(local_size);
 
   // Build dofmaps from ElementDofmap
-  for (auto& cell : mesh::MeshRange(mesh, D, mesh::MeshRangeType::ALL))
+  for (int c = 0; c < connectivity[0]->num_nodes(); ++c)
```
michalhabera: Are you starting to remove mesh iterators, in general? In other places I do not see this change.

cpp/dolfinx/mesh/DistributedMeshTools.cpp
```diff
@@ -92,22 +80,23 @@ compute_entity_numbering(const Mesh& mesh, int d)
...
-    std::vector<bool> vertex_shared(mesh.num_entities(0), false);
+    auto map0 = topology.index_map(0);
+    assert(map0);
+    const int num_vertices = map0->size_local() + map0->num_ghosts();
+    std::vector<bool> vertex_shared(num_vertices, false);
     for (const auto& v : shared_vertices_local)
       vertex_shared[v.first] = true;
 
-    for (auto& e : mesh::MeshRange(mesh, d, mesh::MeshRangeType::ALL))
+    for (int e = 0; e < size; ++e)
```
michalhabera: `size` is defined as `int32_t`, no need for `e` be `int` (i.e. 64-bit on some machines).

cpp/dolfinx/fem/DofMapBuilder.cpp
garth-wells: In low-level places it is good to avoid the iterators.

Some problems with the iterators were:
1. They were used in places during mesh construction. This is messy because the Mesh isn't yet fully formed so it's something of a guess as to whether or not the iterators are ready (or possibly only half ready)
2. It was often unclear whether to iterate over all the entities or just the owned entities. Dropping to a lower level forces one to think about this
3. It's easy to make the iterators slow - any changes need careful testing

cpp/dolfinx/mesh/DistributedMeshTools.cpp
chrisrichardson: I am planning to delete this code soon, anyway, as it is largely duplicated in `TopologyComputation`.

michalhabera:
I agree with @chrisrichardson here, confused me for a while, I was looking also for num_facets :)

garth-wells:
@chrisrichardson I found very clear once I got in the groove of thinking about data connectivity as an adjacency graph.

What we shouldn't use is 'size' - that is ambiguous.

michalhabera:
... but the naming issue could be handled also just with documentation...

PR #750, opened 2020-02-10
Remove compute numbering

chrisrichardson:
Removes `DistributedMeshTools::compute_numbering` as many of the same calculations were being done in `TopologyComputation`.


cpp/dolfinx/mesh/TopologyComputation.cpp
```diff
@@ -121,7 +122,9 @@ get_shared_entities(MPI_Comm neighbour_comm,
 //-----------------------------------------------------------------------------
 // Communicate with sharing processes to find out which entities are ghost
 // and return a mapping vector to move them to the end of the local range.
-std::vector<int> get_ghost_mapping(
+std::tuple<std::vector<int>, std::shared_ptr<common::IndexMap>,
```
garth-wells: Use Doxygen-style doc string to document in arguments and return value

cpp/dolfinx/mesh/TopologyComputation.cpp
```diff
@@ -121,7 +122,9 @@ get_shared_entities(MPI_Comm neighbour_comm,
...
+           std::map<std::int32_t, std::set<std::int32_t>>>
+get_ghost_mapping(
```
garth-wells: Use a more descriptive name that includes the from/to.

cpp/dolfinx/mesh/TopologyComputation.cpp
```diff
@@ -259,34 +271,30 @@ std::vector<int> get_ghost_mapping(
...
-      ++c;
-    }
-  assert(c == entity_count);
+  std::shared_ptr<common::IndexMap> index_map
+      = std::make_shared<common::IndexMap>(comm, num_local, ghost_indices, 1);
+
+  // Remap shared entities to new order
+  std::map<std::int32_t, std::set<int>> remapped_shared_entities;
+  for (auto q : shared_entities)
+    remapped_shared_entities[mapping[q.first]] = q.second;
```
garth-wells: Use `insert`.

cpp/dolfinx/mesh/TopologyComputation.cpp
```diff
@@ -500,15 +509,17 @@ compute_from_map(const graph::AdjacencyList<std::int32_t>& c_d0_0,
 
 //-----------------------------------------------------------------------------
 std::tuple<std::shared_ptr<graph::AdjacencyList<std::int32_t>>,
-           std::shared_ptr<graph::AdjacencyList<std::int32_t>>, std::int32_t>
+           std::shared_ptr<graph::AdjacencyList<std::int32_t>>,
+           std::shared_ptr<common::IndexMap>,
+           std::map<std::int32_t, std::set<int>>>
```
garth-wells: Make the `set` `std::int32_t` or `std::int64_t` so it's clear if it hold local for global indices.

chrisrichardson: Neither. It is for process numbers - I guess this should be int32.

garth-wells: `std::int32_t` would be good then.

Seeing `int` in this part of the code makes we wondering if the int type is correct.

PR #751, opened 2020-02-10
Expose petsc-vector from C++

jorgensd:
Exposed with initializer from IndexMap, so that one does not have to go through a Function with an appropriate function space to get a vector (used in MPC).


jorgensd:
Already exposed through `dolfinx.cpp.fem.create_vector`

PR #752, opened 2020-02-10
Expose AdjacencyList::num_links()

IgorBaratta:
Helpful for custom assemblers.

### DIFF
--- a/python/dolfinx/wrappers/graph.cpp
+++ b/python/dolfinx/wrappers/graph.cpp
@@ -26,18 +26,19 @@ void graph(py::module& m)
   py::class_<dolfinx::graph::AdjacencyList<std::int32_t>,
              std::shared_ptr<dolfinx::graph::AdjacencyList<std::int32_t>>>(
       m, "AdjacencyList", "Adjacency list")
-      .def(
-          "links",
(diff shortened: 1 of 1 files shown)


python/dolfinx/wrappers/graph.cpp
```diff
@@ -26,18 +26,19 @@ void graph(py::module& m)
...
+             return self.links(i);
+           },
+           "Links (edges) of a node",
+           py::return_value_policy::reference_internal)
       .def("array", &dolfinx::graph::AdjacencyList<std::int32_t>::array,
            "All edges", py::return_value_policy::reference_internal)
       .def("offsets", &dolfinx::graph::AdjacencyList<std::int32_t>::offsets,
            "Index to each node in the links array",
            py::return_value_policy::reference_internal)
+      .def("num_links", &dolfinx::graph::AdjacencyList<std::int32_t>::num_links,
```
garth-wells: I left this out because I thought `c.links().shape` was enough. What do you think?

IgorBaratta: Indeed `c.links(0).shape[0]` is enough.

Closing the PR.

PR #760, opened 2020-02-12
Added parallel test for Function.eval

jorgensd:


jorgensd:
New Pr coming up covering all celltypes

PR #770, opened 2020-02-14
Quadrature permutations for DG and some negating of vector-valued spaces

### LINKED ISSUE #645: Permuting the dofs of DG spaces
The DG fem pipeline test (https://github.com/FEniCS/dolfinx/pull/634) is failing for tetrahedrons. I believe it only passes for other shapes because the `Unit*Mesh`ses are ordered nicely.

These tests pass if order_simplex is called on the mesh

For DG spaces, all the dofs are interior dofs, so the `dof_permutation` is moving all the dofs, including those associated with vertices. This may be the issue, or it may be related to https://github.com/FEniCS/dolfinx/issues/31 and how jumps etc are computed.

### LINKED ISSUE #728: Bug in dof permutations
The following triggers an assertion in the function `triangle_rotation_and_reflection` in `dofs_permutation.cpp`.

```python
from dolfinx import MPI, FunctionSpace, UnitSquareMesh
from dolfinx.cpp.mesh import CellType

V = FunctionSpace(UnitSquareMesh(MPI.comm_world, 6, 6,
                                 CellType.triangle),  ('N1curl', 2))
```

mscroggs:
- Pass facet rotation and reflection data into FFCx so that FFCx can use these to calculate quadrature permutations and vector direction swaps
 - Moved dof permutations into FFC, as these will likely eventually be calculated using data directly from FIAT.
 - The numbers of times each permutation should be applied to each cell is now stored in Topology.
 - High order vector-valued spaces are not yet supported on unordered meshes, and the meshes for these are currently ordered in the pipeline tests.

This pull request resolves #728 and resolves #645 and includes #634.

This is an updated version of #702, and is twinned with FEniCS/ffcx#201.


cpp/dolfinx/fem/assemble_matrix_impl.cpp
```diff
@@ -132,19 +134,29 @@ void fem::impl::assemble_cells(
...
   {
     // Get cell coordinates/geometry
     for (int i = 0; i < num_dofs_g; ++i)
       for (int j = 0; j < gdim; ++j)
         coordinate_dofs(i, j) = x_g(cell_g[pos_g[cell_index] + i], j);
 
+    const mesh::MeshEntity cell(mesh, gdim, cell_index);
+
+    Eigen::Array<bool, 1, Eigen::Dynamic> cell_edge_reflections
+        = mesh.topology().get_edge_reflections(cell_index);
```
michalhabera: This is performance critical code, we should be careful. I would suggest to set the storage to Eigen::RowMajor in the topology - you really want to extract the contiguous chunk of memory, default eigen's is ColumnMajor...

cpp/dolfinx/fem/assemble_matrix_impl.cpp
```diff
@@ -132,19 +134,29 @@ void fem::impl::assemble_cells(
...
+    Eigen::Array<bool, 1, Eigen::Dynamic> cell_face_reflections
```
michalhabera: I think you want these be `const Eigen::Ref<...>`, similar change needed in `get_face_reflections` etc. definitions.

mscroggs:
I believe the test currently failing will pass once #791 is merged

cpp/dolfinx/fem/assemble_matrix_impl.cpp
```diff
@@ -328,11 +360,17 @@ void fem::impl::assemble_interior_facets(
 
     // Get local index of facet with respect to the cell
     const int local_facet[2] = {cell0.index(facet), cell1.index(facet)};
-    const int orient[2] = {0, 0};
 
     // Get cell vertex coordinates
     const int cell_index0 = cell0.index();
     const int cell_index1 = cell1.index();
+
+    const std::uint8_t perm[2]
```
michalhabera: Do we use pure C arrays somewhere else? This could be `const std::array<std::uint8_t, 2>` instead

cpp/dolfinx/fem/utils.cpp
```diff
@@ -445,12 +445,14 @@ fem::create_element_dof_layout(const ufc_dofmap& dofmap,
             num_entity_dofs.data());
 
   // Fill entity dof indices
+  int dof_count = 0;
```
michalhabera: Where is this variable used?

cpp/dolfinx/function/Function.cpp
```diff
@@ -252,9 +254,18 @@ void Function::eval(
     // Compute basis on reference element
     element.evaluate_reference_basis(basis_reference_values, X);
 
+    Eigen::Array<bool, 1, Eigen::Dynamic> cell_edge_reflections
```
michalhabera: Also add `const Eigen::Ref<>` etc...

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -269,5 +271,65 @@ std::string Topology::str(bool verbose) const
   return s.str();
 }
 //-----------------------------------------------------------------------------
+Eigen::Ref<const Eigen::Array<bool, 1, Eigen::Dynamic>>
+Topology::get_edge_reflections(const int cell_n) const
```
michalhabera: It probably doesnt matter here, but cell indices are usually std::int32_t, right?

cpp/dolfinx/fem/utils.cpp
mscroggs: it must've been used in some code I've since deleted

cpp/dolfinx/fem/assemble_matrix_impl.cpp
mscroggs: perm is passed into ffc. Changing this to a `std::array` is causing errors (unless I'm doing something stupid?)

michalhabera: Sure, but you would need to pass `perm.data()` to get the underlying pointer. It would be more c++-ish. But I see now raw arrays at more places, I would prefer to see more std::arrays.

mscroggs: I've replaced perm[2] and local_facet[2] with std:arrays

PR #776, opened 2020-02-16
Add ability to pass locate_dofs_geometrical two function spaces

jpdean:
This PR adds the ability to pass two function spaces to [`locate_dofs_geometrical`](https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfinx/fem/DirichletBC.h#L79-L81) in a similar manner to [`locate_dofs_topological` ](https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfinx/fem/DirichletBC.h#L63-L67), which already has this functionality.

An example of usage can be found [here](https://github.com/jpdean/dolfinx/blob/jpdean/stokes_demo/python/demo/stokes-taylor-hood/demo_stokes-taylor-hood.py#L422).


python/test/unit/fem/test_bcs.py
```diff
@@ -0,0 +1,35 @@
+import dolfinx
```
garth-wells: Add a copyright/license text at the top - see other test files for an example.

PR #781, opened 2020-02-17
Bump Dockerfile to Ubuntu 20.04

garth-wells:
In preparation for Ubuntu 20.04. Updates some packages we use.


docker/Dockerfile
```diff
@@ -144,6 +135,17 @@ RUN cd /usr/local && \
 
 ENV KAHIP_ROOT=/usr/local/KaHIP
 
+# Download Install Gmsh SDK
```
michalhabera: Can you use ubuntu's `python3-gmsh` and `gmsh`? https://packages.ubuntu.com/focal/python3-gmsh

garth-wells: I don't think we want to - the Ubuntu gmsh package is already old, and after two years it will be really old.

PR #783, opened 2020-02-18
Add Hypre to complex build of PETSc

garth-wells:


michalhabera:
Hypre seems not supported in complex, https://www.mcs.anl.gov/petsc/documentation/linearsolvertable.html
Did something change?

garth-wells:
Seemed to build for me locally. Let's see what happens on CircleCI.

michalhabera:
I think it builds but one cannot use it anyway. It will fail to solve with Hypre anything runtime.

PR #784, opened 2020-02-18
Cell ordering documentation for meshes

jorgensd:


cpp/dolfinx/mesh/Mesh.h
```diff
@@ -80,7 +80,11 @@ class Mesh
   /// @param[in] points Array of geometric points, arranged in global
   ///                   index order
   /// @param[in] cells Array of cells (containing the global point
-  ///                  indices for each cell)
+  ///                  indices for each cell). The ordering of each cell depends on the cell type.
```
michalhabera: I would add "The **local** ordering of each ...".

PR #792, opened 2020-02-19
Expose issues with partitioning in parallel for higher order quadrilaterals

jorgensd:
Adding minimal unitsquare mesh for second order quads to the datadir to use it in the pipeline test.

Pipeline test passes in serial and crashes in parallel.


garth-wells:
Testing via gmsh is now supported.

PR #800, opened 2020-02-21
Remove cell number inputs from permutation information

mscroggs:
`get_edge_reflections` etc in Topology now return the full data arrays, not rows of these, and I've added these to the Python wrappers.

This data is needed to make custom assemblers (which @jorgensd) is currently using, and restructuring like this makes it easier to get hold of it from Python.


cpp/dolfinx/fem/assemble_matrix_impl.cpp
```diff
@@ -132,6 +132,14 @@ void fem::impl::assemble_cells(
   Eigen::Matrix<PetscScalar, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
       Ae;
 
+  const Eigen::Ref<const Eigen::Array<bool, Eigen::Dynamic, Eigen::Dynamic>>
```
michalhabera: I am not sure here, maybe @garth-wells knows, but I would add Eigen::RowMajor also here (and all below), to avoid any possible magic performed by Eigen when trying to proxy RowMajor as ColumnMajor. Remember, default Eigen storage is ColumnMajor.

michalhabera: Of course, don't do it for 1D arrays.

mscroggs: Good point, I've added them throughout

PR #802, opened 2020-02-22
Disable pygmsh

garth-wells:
Tests failing due to pygmsh/meshio problem. Re-enable when pygmsh is updated.


michalhabera:
Maybe add as an issue? We use those tests to test dolfinx functionality. However, we can avoid meshio/pygmsh with direct use of gmsh python API now.

francesco-ballarin:
I believe this can be reverted now, a new pygmsh version was released yesterday

PR #803, opened 2020-02-22
Remove Mesh::Entity::index(e)

garth-wells:
Not needed and hid some mildly expensive operations.


michalhabera:
Nice, did you measure if any speedup is visible?

garth-wells:
I didn't measure, but it certainly won't be slower.

PR #805, opened 2020-02-22
Create distributed adjacency list

garth-wells:
Progress towards new distributed graph building.

Fixes some memory issues with IndexMap.


python/test/unit/mesh/test_mesh_utils.py
```diff
@@ -104,4 +111,57 @@ def test_partition():
...
+    print(topology.connectivity(2, 0).array())
+    print(global_to_local_vertices)
+    # return
+    cells, vertex_map, l2g = cpp.mesh.create_distributed_adjacency_list(cpp.MPI.comm_world, topology,
+                                                                        global_to_local_vertices)
+
+    print("l2G", l2g)
+    print("cells: ", cells.array())
+    print("offset: ", cells.offsets())
+    return
```
michalhabera: Forgotten here? Also I am not sure if you wanted to leave all those prints here. Running `pytest . -s` is then too poluted.

PR #814, opened 2020-02-25
Strict FFCX JIT compiler in tests

michalhabera:
Enable `-Wall -Werror` for FFCX JIT compilation in all tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,7 @@ build-documentation-python: &build-documentation-python
 unit-tests-python: &unit-tests-python
   name: Run unit tests (Python, serial)
   environment:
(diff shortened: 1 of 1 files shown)


michalhabera:
Understand. But I am not sure how to address this. Passing fixture to each test is not good given the number of tests. Another option is to "hack" ENV variables in tests...

garth-wells:
What issues could these flags pick up that wouldn't be/should be picked up by FFCX tests?

michalhabera:
Currently, there are no tests for JIT compilation of more exotic elements (Hdiv/Hcurl).
Ok, but that could be a good idea - add more tests to FFCX and jit-test Hdiv/Hcurl also there.

PR #820, opened 2020-02-26
Exposed pack_coefficients and created pack_constants

jorgensd:
Removes duplication code making it easier for restructuring later on.
These pybindings are required to efficiently being able to do advanced forms with coefficients and constants.


python/dolfinx/wrappers/fem.cpp
```diff
@@ -148,6 +148,10 @@ void fem(py::module& m)
 
   m.def("create_sparsity_pattern", &dolfinx::fem::create_sparsity_pattern,
         "Create a sparsity pattern for bilinear form.");
+  m.def("pack_coefficients", &dolfinx::fem::pack_coefficients,
```
garth-wells: This won't be very efficient because the C++ side returns at `std::vector`, which can't be wrapped as a NumPy array.

michalhabera: Good point, `std::vector` is automatically casted to a `list` which is very slow in Python. This could cause a lot of hidden performance issues when passing `list` to a custom kernel.
You should probably make pack_foo return Eigen arrays - which is nicely returned as NumPy array in Python, or you can in pybind11 return a py::array_t with some hacks to avoid copying. I would prefer Eigen.

jorgensd: @michalhabera I've proposed a solution, where we use eigen arrays to pack constants. Is there any more efficient way of inserting the std::vectors into the Eigen arrays that I've missed?

michalhabera: @jorgensd The best way is to use Eigen::Map and map a continuous chunk of memory as Eigen::Array. You can find example in many places in dolfinx. (so just prepare std::vector as before, but in the very last line return an Eigen::Map)

michalhabera: I do not think pac_constants is performance critical in preparation of the eigen array, but personally I find Eigen:Mapping nice and readable.

jorgensd: latest commit uses Eigen::Map  as I agree it is very readable.

jorgensd:
@garth-wells Can we merge this when CI passes?

cpp/dolfinx/fem/utils.cpp
```diff
@@ -758,3 +760,21 @@ fem::pack_coefficients(const fem::Form& form)
   return c;
 }
 //-----------------------------------------------------------------------------
+Eigen::Array<PetscScalar, Eigen::Dynamic, 1>
+fem::pack_constants(const fem::Form& form)
+{
+  const std::vector<
+      std::pair<std::string, std::shared_ptr<const function::Constant>>>
+      constants = form.constants();
+  std::vector<PetscScalar> constant_values;
```
garth-wells: Is it necessary to build a vector? Can the data go straight into the `Eigen::Array<PetscScalar, Eigen::Dynamic, 1>`?

michalhabera: Not so easily, since the size of the array is not known before the loop. Std::vector here is beneficial as a resizeable storage with push_back. And the loop neither the array is will ever be big.

PR #823, opened 2020-02-27
Fix out-of bounds buffer access in DofMapBuilder

chrisrichardson:
* currently failing


garth-wells:
Rather than using assert for small loops we can use std::vector::at rather than std::vector::operator[]. std::vector::at does range checking.

chrisrichardson:
I don't know about the relative costs of `[]` vs `::at` in a tight loop. `assert` is removed on a Release build. Anyway, I now removed the assert and made it a required check. I am not 100% sure I have done this correctly, as I've not looked at `DofMapBuilder` recently, so a review would be helpful.

garth-wells:
The performance isn't an issue here because it's not a tight loop.

PR #824, opened 2020-02-27
Document base permutations

mscroggs:
- Added some documentation for base permutations input of `ElementDofLayout`.
 - Added asserts in the `ElementDofLayout` to check the shape of the base permutations, so that the assertion is hit rather than a segfault later.

### DIFF
--- a/cpp/dolfinx/fem/ElementDofLayout.cpp
+++ b/cpp/dolfinx/fem/ElementDofLayout.cpp
@@ -16,6 +16,34 @@
 using namespace dolfinx;
 using namespace dolfinx::fem;
 
+namespace
+{
+//-----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


cpp/dolfinx/fem/ElementDofLayout.cpp
```diff
@@ -75,6 +75,16 @@ ElementDofLayout::ElementDofLayout(
       _num_dofs += entity_dofs_[dim][entity_index].size();
     }
   }
+
+  // Assert that base_permutations is the correct shape
+#ifdef DEBUG
```
garth-wells: This is quite a cheap, so I would suggest doing it in all cases and throwing a `runtime_error` if the sizes are wrong.

PR #828, opened 2020-02-29
Fix direction of VTK -> DOLFIN map 

garth-wells:
Was the transpose (VTK <- DOLFIN), which was compensated for by another transpose in `permute_ordering`.

Also fixes permutation for HEX27


michalhabera:
There is plenty of commented code.

garth-wells:
Yes - we need @jorgensd to check that the direction of the map is consistent with the function name.

PR #830, opened 2020-03-01
Proper separation of mesh topology and geometry and new parallel distribution

garth-wells:
This change properly separates mesh topology and geometry, which considerably simplifies much of the code, especially for high-order geometries and in parallel. The geometry now has its own dofmap (`AdjacencyList`) which is constructed by `DofMapBuilder` which uses re-ordering for locality.

A new `Mesh` constructor takes a `Topology` and a `Geometry` object. There are examples/tests that now support P1 and P2 geometries from Gmsh which can be easily read-in in parallel with (py)gmsh. These can be output in XDMF (P1 and P2).

The new code sits alongside the old code as there are a *lot* of bugs that need to be fixed elsewhere that that change exposed. The main problem is old code conflating topology and geometry, which by virtue of some fortuitous ordering worked before. The new code is not used by default as it triggers too many existing bugs. These will be tackled later.

Quite a few bugs have been fixed in this PR that the proper separation of geometry and topology exposed.


cpp/dolfinx/generation/RectangleMesh.cpp
```diff
@@ -26,6 +27,12 @@ mesh::Mesh build_tri(MPI_Comm comm, const std::array<Eigen::Vector3d, 2>& p,
   {
     Eigen::Array<double, 0, 2, Eigen::RowMajor> geom(0, 2);
     Eigen::Array<std::int64_t, 0, 3, Eigen::RowMajor> topo(0, 3);
+
+    // New mesh builder
```
michalhabera: Maybe it could help with later search to add `// TODO` flag here (or some other keyword to identify sketched patterns in the code)?

cpp/dolfinx/io/cells.cpp
```diff
@@ -152,120 +156,63 @@ std::vector<std::uint8_t> io::cells::dolfin_to_vtk(mesh::CellType type,
...
-                                               int num_nodes)
-{
-  switch (type)
-  {
-  case mesh::CellType::quadrilateral:
-  {
-    assert((std::sqrt(num_nodes) - std::floor(std::sqrt(num_nodes))) == 0);
-    // Number of nodes in each direction
-    const int n = sqrt(num_nodes);
+// std::vector<std::uint8_t> io::cells::lex_to_tp(mesh::CellType type,
```
michalhabera: Do we need this commented code here, even for @jorgensd ? Git has a decent history.

cpp/dolfinx/mesh/Geometry.h
```diff
@@ -45,56 +65,94 @@ class Geometry
...
-  /// Return array of coordinates for all points (const version)
-  const Eigen::Array<double, Eigen::Dynamic, 3, Eigen::RowMajor>&
-  points() const;
+  /// Return coordinate array for point with local index n
+  Eigen::Ref<const Eigen::Vector3d> x(int n) const;
+
+  /// @todo Remove this once all Geometry objects have an IndexMap
+  ///
+  /// Return the number of global points in the geometry
+  std::size_t num_points_global() const;
```
michalhabera: This seems to be `std::int64_t`

python/test/unit/mesh/test_mesh_utils.py
```diff
@@ -190,18 +328,47 @@ def test_topology_partition():
...
+    # Build coordinate dof array
+    x_g = coords[l2l]
+
+    # Create Geometry
+    geometry = cpp.mesh.Geometry(dof_index_map, dofmap, x_g, l2g, order)
+
+    # Create mesh
+    mesh = cpp.mesh.Mesh(cpp.MPI.comm_world, topology, geometry)
+
+    # Write mesh to file
```
michalhabera: This was for debugging I guess.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -342,4 +346,97 @@ void Topology::set_entity_permutation(std::size_t cell_n, int entity_dim,
...
+                                           1);
+  topology_local.set_index_map(0, map);
+
+  // Create facets for local topology, and attach to the topology
+  // object. This will be used to find possibly shared cells
+  auto [cf, fv, map0]
+      = TopologyComputation::compute_entities(comm, topology_local, tdim - 1);
+  topology_local.set_connectivity(cf, tdim, tdim - 1);
+  topology_local.set_index_map(tdim - 1, map0);
+  if (fv)
```
chrisrichardson: Why wouldn't `fv` exist?

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -342,4 +346,97 @@ void Topology::set_entity_permutation(std::size_t cell_n, int entity_dim,
...
+      = std::make_shared<graph::AdjacencyList<std::int32_t>>(cells_local);
+  topology_local.set_connectivity(_cells_local, tdim, 0);
+
+  const int n = local_to_global_vertices.size();
+  map = std::make_shared<common::IndexMap>(comm, n, std::vector<std::int64_t>(),
+                                           1);
+  topology_local.set_index_map(0, map);
+
+  // Create facets for local topology, and attach to the topology
+  // object. This will be used to find possibly shared cells
```
chrisrichardson: Is this comment correct? Shared cells will come from the cell index map.

cpp/dolfinx/mesh/PartitioningNew.h
```diff
@@ -144,6 +146,31 @@ class PartitioningNew
...
+  /// @param[in] local Adjacency list with local, contiguous link
+  ///   indices
+  /// @return Map from local index to global index, which if applied to
+  /// the local adjacency list indices would yield the global adjacency
+  /// list
+  static std::vector<std::int64_t> compute_local_to_global_links(
+      const graph::AdjacencyList<std::int64_t>& global,
+      const graph::AdjacencyList<std::int32_t>& local);
+
+  /// Compute a local0-to-local1 map from two local-to-global maps with
```
chrisrichardson: Is there a better name than `local_to_local`? The global indices of `local0` must be a subset of `local1`... maybe that can be stated in a comment or included in the function name?

cpp/dolfinx/mesh/Topology.cpp
garth-wells: Pointers should always be checked.

cpp/dolfinx/mesh/Topology.cpp
garth-wells: It's the local topology, i.e. on a single rank, so the IndexMap is on `MPI_COMM_SELF` and is unaware of neighbours.

cpp/dolfinx/mesh/Topology.h
```diff
@@ -218,5 +250,10 @@ class Topology
   // are interior to the domain
   std::shared_ptr<const std::vector<bool>> _interior_facets;
 };
+
+/// Create distributed topology
```
chrisrichardson: document

cpp/dolfinx/mesh/PartitioningNew.cpp
```diff
@@ -446,9 +444,9 @@ PartitioningNew::create_local_adjacency_list(
   // FIXME: Update AdjacencyList to avoid this
   const Eigen::Array<std::int32_t, Eigen::Dynamic, 1>& offsets
       = cells.offsets();
-  std::vector<std::int32_t> _offsets(offsets.data(),
-                                     offsets.data() + offsets.rows());
-  return {graph::AdjacencyList<std::int32_t>(array_local, _offsets),
+  // std::vector<std::int32_t> _offsets(offsets.data(),
```
chrisrichardson: delete commented code

cpp/dolfinx/mesh/Topology.cpp
chrisrichardson: OK, I just don't understand "possibly shared cells" here

garth-wells: Create a 'local' Topology from the cells that I have. If a call is on the boundary of this local topology it might share a facet with another process.

python/test/unit/mesh/test_mesh_utils.py
garth-wells: Also picks up some issues sometimes, but mainly for visual inspection.

cpp/dolfinx/mesh/PartitioningNew.h
garth-wells: Let's revisit this one later.

PR #838, opened 2020-03-03
Fix ghosting in KaHIP and ParMetis

IgorBaratta:


cpp/dolfinx/graph/KaHIP.cpp
```diff
@@ -20,9 +20,8 @@ using namespace dolfinx;
 
 graph::AdjacencyList<std::int32_t> dolfinx::graph::KaHIP::partition(
     MPI_Comm mpi_comm, int nparts,
-    const graph::AdjacencyList<unsigned long long>& adj_graph)
+    const graph::AdjacencyList<unsigned long long>& adj_graph, bool ghosting)
```
garth-wells: Could we used fixed-width ints in place of `unsigned long long`?

cpp/dolfinx/graph/KaHIP.cpp
```diff
@@ -66,113 +65,111 @@ graph::AdjacencyList<std::int32_t> dolfinx::graph::KaHIP::partition(
...
+  // Create a map of local nodes to their additional destination processes,
+  // due to ghosting. If no ghosting, this will remain empty.
+  std::map<std::int32_t, std::set<std::int32_t>> local_node_to_dests;
+  if (ghosting)
   {
-    for (int j = 0; j < adj_graph.num_links(i); ++j)
+    common::Timer timer2("Compute graph halo data (KaHIP)");
+
+    // Work out halo cells for current division of dual graph
+    const std::int32_t* xadj = adj_graph.offsets().data();
```
garth-wells: Can this be moved closer to where it's used?

cpp/dolfinx/graph/KaHIP.cpp
```diff
@@ -66,113 +65,111 @@ graph::AdjacencyList<std::int32_t> dolfinx::graph::KaHIP::partition(
...
-        proc_other = part[other_cell - elm_begin];
+      cell_ownership[*p] = *(p + 1);
+    }
 
-      if (proc_this != proc_other)
+    // Generate mapping for where new boundary cells need to be sent
+    for (std::int32_t i = 0; i < ncells; i++)
+    {
+      const std::int32_t proc_this = part[i];
+      for (std::int32_t j = xadj[i]; j < xadj[i + 1]; ++j)
```
garth-wells: Is `xadj` a raw pointer? Can we use it via Eigen?

cpp/dolfinx/graph/KaHIP.cpp
chrisrichardson: It is written into KaHIP. Not sure what are the implications for maintaining compatibility on all systems.

cpp/dolfinx/graph/KaHIP.cpp
chrisrichardson: Yes, there is no need to use it raw here.

cpp/dolfinx/graph/KaHIP.cpp
IgorBaratta: Move down and changed to `Eigen::Map`

cpp/dolfinx/graph/KaHIP.cpp
IgorBaratta: Using `Eigen::Map` now.

chrisrichardson:
Does it need map? Isn't it already just an eigen array?

IgorBaratta:
> Does it need map? Isn't it already just an eigen array?

No need at all. Just changed this.

PR #849, opened 2020-03-07
Switch to new mesh topology/geometry separation for meshes

garth-wells:
More work in support of proper geometry/topology separation, and new mesh distribution.

Significant changes to XDMF in support of forthcoming changes.


francesco-ballarin:
Hi @garth-wells ,
after this PR, how is one supposed to specify the non-default ghost mode (e.g., shared_facet) when reading a mesh from XDMF file? Is it not necessary anymore, or should it be done in a different way?

Thanks.

francesco-ballarin:
Furthermore, was the `XDMFFile::write(const MeshFunction<int>&)` left out on purpose from the pybind11 wrapper? (although there is the corresponding read_mf_int)

Thanks

garth-wells:
The ghost interface needs to come back. 

`MeshFunction` IO will likely go in favour of a much simplified `MeshValueCollection` IO. This code was legacy from old DOLFIN and wasn't easily salvageable,  which is why it was removed.

francesco-ballarin:
OK thanks. So in the meantime I should switch to `HDF5File.read_mvc_size_t` for reading in the `MeshValueCollection`, and then transform it into a `MeshFunction` with 
`mf = MeshFunction("size_t", mesh, mvc_from_file, 0)` whenever I needed the mesh function that I used to read in from file?

PR #859, opened 2020-03-10
Turn on more vector dof tests and remove Ordering

mscroggs:
The ordering of simplex meshes has been removed. More tests for vector dofs have been turned on. Tests now pass without ordering thanks to FEniCS/ffcx#224.

(We may want to turn a few of the slowest tests off before merging into master.)


mscroggs:
Looks like the slowest test is currently failing in complex due to timing out

PR #872, opened 2020-03-12
Implement move assignment operator in MPI::Comm to make sure that previously held communicator is freed.

francesco-ballarin:
The following code

```
import gc
import numpy as np
from dolfinx import MPI, Mesh
from dolfinx.cpp.mesh import CellType, GhostMode

def unit_cell():
    points = np.array([[0.], [1.]])
    cells = np.array([list(range(len(points)))])
    mesh = Mesh(MPI.comm_world, CellType.interval, points, cells,
                [], GhostMode.none)
    return mesh

gc.disable()  # Manually control garbage collection

counter = 0
while True:
    print("Iteration", counter)
    unit_cell()
    counter += 1
    gc.collect()  # Force garbage collection
```

fails on master (on a dolfinx docker image) with
```
[...]
Iteration 2038
Iteration 2039
Iteration 2040
Iteration 2041
  File "tmp.py", line 18, in <module>
    unit_cell()
  File "tmp.py", line 10, in unit_cell
    [], GhostMode.none)
RuntimeError: Duplication of MPI communicator failed (MPI_Comm_dup)
```

This PR fixes the issue by making sure that the move assignment operator in MPI::Comm frees the previously held communicator.

### DIFF
--- a/cpp/dolfinx/common/MPI.cpp
+++ b/cpp/dolfinx/common/MPI.cpp
@@ -51,6 +51,26 @@ dolfinx::MPI::Comm::~Comm()
   }
 }
 //-----------------------------------------------------------------------------
+dolfinx::MPI::Comm&
+dolfinx::MPI::Comm::operator=(dolfinx::MPI::Comm&& comm)
+{
+  // Free the currently held comm
+  if (this->_comm != MPI_COMM_NULL)
+  {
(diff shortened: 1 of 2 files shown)


garth-wells:
Thanks @francesco-ballarin - I've hit this problem recently.

PR #873, opened 2020-03-12
Fix VTK Viusalization of facet functions

jorgensd:
After separation of Geometry and Topology, the topology vertex ordering is different from the geometry vertex ordering. Therefore we have to convert them to be able to visualize mesh functions properly.
Minimal example were visualization fails on master:
```
import dolfinx
import dolfinx.io
mesh = dolfinx.UnitSquareMesh(dolfinx.MPI.comm_world, 2, 2)
mf = dolfinx.MeshFunction("size_t", mesh, mesh.topology.dim-1, 0)
dolfinx.io.VTKFile("mf.pvd").write(mf)
```

### DIFF
--- a/cpp/dolfinx/io/VTKWriter.cpp
+++ b/cpp/dolfinx/io/VTKWriter.cpp
@@ -184,20 +184,18 @@ void write_ascii_mesh(const mesh::Mesh& mesh, int cell_dim,
   {
     // Special case where the cells are visualized (Supports higher order
     // elements)
-    const graph::AdjacencyList<std::int32_t>& connectivity_g
+    const graph::AdjacencyList<std::int32_t>& x_dofmap
         = mesh.geometry().dofmap();
-    const Eigen::Array<std::int32_t, Eigen::Dynamic, 1>& cell_connections
-        = connectivity_g.array();
-    const Eigen::Array<std::int32_t, Eigen::Dynamic, 1>& pos_g
-        = connectivity_g.offsets();
-    num_nodes = connectivity_g.num_links(0);
+    // FIXME: USe better way to get number of nods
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/io/VTKWriter.cpp
```diff
@@ -210,6 +210,22 @@ void write_ascii_mesh(const mesh::Mesh& mesh, int cell_dim,
       throw std::runtime_error("MeshFunction of lower degree than the "
                                "topological dimension is not implemented");
     }
+
+    // Build a map from topology to geometry
+    auto cell_connectivity = mesh.topology().connectivity(tdim, 0);
+    const std::int32_t num_mesh_vertices = mesh.num_entities(0);
+    std::vector<std::int32_t> topology_to_geometry(num_mesh_vertices);
+    auto x_dofs = mesh.geometry().dofmap();
+    for (int j = 0; j < mesh.num_entities(tdim); ++j)
```
garth-wells: Avoid `num_entities`. It will disappear soon. In the case you can use `cell_connectivity.num_nodes()`.

cpp/dolfinx/io/VTKWriter.cpp
```diff
@@ -210,6 +210,22 @@ void write_ascii_mesh(const mesh::Mesh& mesh, int cell_dim,
       throw std::runtime_error("MeshFunction of lower degree than the "
                                "topological dimension is not implemented");
     }
+
+    // Build a map from topology to geometry
+    auto cell_connectivity = mesh.topology().connectivity(tdim, 0);
+    const std::int32_t num_mesh_vertices = mesh.num_entities(0);
```
garth-wells: Avoid `num_entities`.

PR #874, opened 2020-03-12
Reintroduced parts of the XDMF IO of write_checkpoint to take over for write

jorgensd:
Replacing write(function, t), which computed vertex values of the function, with one recognizing finite elements, similar to what @michalhabera had in write_checkpoint. The main difference is that the last two DataItem's as they are not required for visualization.

An issue I can see with changing the IO to this, is that we can no longer visualize on lines and hexes, as they are not supported by paraview (but supported by XDMF). Could add an if statement which checks if it can be visualized or not, and does the old code path if it cant be visualized.

### DIFF
--- a/cpp/dolfinx/io/XDMFFile.cpp
+++ b/cpp/dolfinx/io/XDMFFile.cpp
@@ -44,8 +44,7 @@ std::string get_hdf5_filename(std::string filename)
 } // namespace
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


jorgensd:
> I would like to revise and simplify the IO code before we start putting things back.

That ok. I'll do some adjustments to this anyhow, just to have somewhere to start from.

jorgensd:
Should be reconsidered after #916

PR #877, opened 2020-03-12
Updates to parallel refinement

chrisrichardson:
Clean up code and remove old style

### DIFF
--- a/cpp/dolfinx/refinement/ParallelRefinement.cpp
+++ b/cpp/dolfinx/refinement/ParallelRefinement.cpp
@@ -144,6 +144,7 @@ void ParallelRefinement::update_logical_edgefunction()
   {
     data_to_send.insert(data_to_send.end(), _marked_for_update[i].begin(),
                         _marked_for_update[i].end());
+    _marked_for_update[i].clear();
     send_offsets.push_back(data_to_send.size());
   }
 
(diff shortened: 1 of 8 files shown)


cpp/dolfinx/refinement/PlazaRefinementND.cpp
```diff
@@ -52,7 +51,7 @@ void enforce_rules(ParallelRefinement& p_ref, const mesh::Mesh& mesh,
       bool any_marked = false;
       auto edges = f_to_e->links(f);
       for (int i = 0; i < edges.rows(); ++i)
-        any_marked |= p_ref.is_marked(edges[i]);
+        any_marked = any_marked or p_ref.is_marked(edges[i]);
```
garth-wells: `p_ref.is_marked` seems a bit heavy weight - is it just an array lookup?

cpp/dolfinx/refinement/ParallelRefinement.cpp
```diff
@@ -25,26 +25,43 @@ using namespace dolfinx::refinement;
 
 //-----------------------------------------------------------------------------
 ParallelRefinement::ParallelRefinement(const mesh::Mesh& mesh)
-    : _mesh(mesh), _marked_edges(mesh.num_entities(1), false),
-      _marked_for_update(MPI::size(mesh.mpi_comm()))
+    : _mesh(mesh), _marked_edges(mesh.num_entities(1), false)
```
garth-wells: Can we avoid `mesh.num_entities`? I'd like to remove it soon in favour of going via `topology` or `geometry` so that it's clear what's being requested.

PR #883, opened 2020-03-14
Change FormIntegrals domains marker type to MeshFunction<int> rather than MeshFunction<std::size_t>

francesco-ballarin:
Recent changes to `XDMFFile` have limited to int only the `MeshFunction` types for I/O. Thus, in order to be able to use markers imported from file when assembling forms, I propose in this PR to change the domain marker type to use `int`, rather than `size_t`.

I will close this if you want me to do it the other way around (allow `XDMFFile` to read in `MeshFunction<size_t>` as well, and do not change `FormIntegrals` marker type).

I realize that whichever of the two options is chosen, it will probably be a short term solution (until `MeshValueCollection` is used instead of `MeshFunction`).

### DIFF
--- a/cpp/dolfinx/fem/Form.cpp
+++ b/cpp/dolfinx/fem/Form.cpp
@@ -168,27 +168,27 @@ void Form::set_tabulate_tensor(
     _integrals.set_default_domains(*_mesh);
 }
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


garth-wells:
`MeshFunction` is about to be replaced in #889.

garth-wells:
Superseded by #889.

Issue #536, opened 2019-09-24
Two incompatibilities with ds and dS defined from ufl.Measure


francesco-ballarin:
Dear developers,
I am facing two issues when using ds and dS measures.

I have customized test_ghost_mesh_assembly.py in
https://github.com/francesco-ballarin/dolfinx/commit/63b57bac2
to add a test parametrization on dx as follows:
  1. `dx_from_ufl`: use ufl.dx, as in master
  2. `dx_from_measure`: use ufl.Measure("dx", subdomain_data=marker), where marker is the constant 1 on the whole mesh
   3. `dx_from_measure_and_subdomain`: take the output of `dx_from_measure` and apply a restriction to the subdomain 1, which effectively should be the whole domain

and similarly for ds and dS.

As soon as `ds_from_measure_and_subdomain` is used, tests are failing
```
test_ghost_mesh_assembly.py::test_ghost_mesh_assembly[ds_from_measure_and_subdomain-dx_from_ufl-mode0] FAILED
test_ghost_mesh_assembly.py::test_ghost_mesh_assembly[ds_from_measure_and_subdomain-dx_from_measure-mode0] FAILED
test_ghost_mesh_assembly.py::test_ghost_mesh_assembly[ds_from_measure_and_subdomain-dx_from_measure_and_subdomain-mode0] FAILED
```
and an incorrect norm is obtained
```
>       assert normA == pytest.approx(0.6713621455570528, rel=1.e-6, abs=1.e-12)
E       assert 2.7208495255521115 == 0.6713621455570528 ± 6.7e-07
E         -2.7208495255521115
E         +0.6713621455570528 ± 6.7e-07
```
I believe this to be an incompatibility between sums of integrals in ds and dx, because `test_assembly_ds_domains` (which only has integrals in ds) is working.

Furthermore, the test with dS(1)
```
test_ghost_mesh_assembly.py::test_ghost_mesh_dS_assembly[dS_from_measure_and_subdomain-mode0] FAILED
```
is failing with
```
    @functools.singledispatch
    def assemble_matrix(a,
                        bcs: typing.List[DirichletBC] = [],
                        diagonal: float = 1.0) -> PETSc.Mat:
        """Assemble bilinear form into a matrix. The returned matrix is not
        finalised, i.e. ghost values are not accumulated.
    
        """
        a_cpp = _create_cpp_form(a)
        A = cpp.fem.create_matrix(a_cpp)
        A.zeroEntries()
[...]

PR #889, opened 2020-03-16
Add MeshTags

michalhabera:
`MeshTags` is a new data storage class for tagging mesh entities with values. It is basically a sparse `MeshFunction`, because it stores local-to-process indices of mesh entities and their values.

`MeshTags` is used for example in attaching a subdomain_data for UFL measure.

In the upcoming PR new IO for `MeshTags` will be added, and `MeshFunction` and `MeshValueCollection`, including its IO, will be removed in favour of this class.

Typical workflow for user is:
1. find mesh entities indices using `locate_entities_geometrical(mesh, dim, marking_function)`,
2. prepare `values` data array,
3. `MeshTags(mesh, dim, indices, values)`


python/dolfinx/mesh.py
```diff
@@ -35,11 +37,15 @@ def compute_marked_boundary_entities(mesh: cpp.mesh.Mesh,
...
+    return cpp.mesh.locate_entities_geometrical(mesh, dim, marker, boundary_only)
 
 
-# __all__ = ["MeshFunction", "MeshValueCollection"]
-
+_meshtags_types = {
+    numpy.int8: cpp.mesh.MeshTags_int8,
+    numpy.intc: cpp.mesh.MeshTags_int,
+    numpy.int64: cpp.mesh.MeshTags_int64,
+    numpy.double: cpp.mesh.MeshTags_int64
```
francesco-ballarin: Is this a typo?

python/test/unit/fem/xtest_ghost_mesh_assembly.py
```diff
@@ -6,62 +6,27 @@
...
-def dx_from_measure_and_subdomain(mesh):
-    dx = dx_from_measure(mesh)
-    return dx(1)
-
-
 def ds_from_ufl(mesh):
     return ufl.ds
 
 
-def ds_from_measure(mesh):
```
francesco-ballarin: Is this removal temporary or permanent? (not that it matters in the short time period, as the file is currently skipped)

michalhabera: What was the point of `_from_measure` which was anyway marking all mesh entities?  I would like to add later tests that do the integration on dx, dS, ds but only on subset of mesh entities.

francesco-ballarin: Ok, so a (better) variation of d*_from_measure_and_subdomain on a subset of mesh entities. The point was #536.

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -0,0 +1,202 @@
...
+           const Eigen::Array<T, Eigen::Dynamic, 1>& values);
+
+  /// Destructor
+  ~MeshTags() = default;
+
+  /// Associated mesh
+  const std::shared_ptr<const Mesh> mesh;
+
+  /// Topological dimension of tagged mesh entities
+  const int dim;
```
garth-wells: Make this a method - direct access to `const` member data prevents move semantics.

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -0,0 +1,202 @@
...
+  ///    will be copied.
+  MeshTags(std::shared_ptr<const Mesh> mesh, int dim,
+           const Eigen::Array<std::int32_t, Eigen::Dynamic, 1>& indices,
+           const Eigen::Array<T, Eigen::Dynamic, 1>& values);
+
+  /// Destructor
+  ~MeshTags() = default;
+
+  /// Associated mesh
+  const std::shared_ptr<const Mesh> mesh;
```
garth-wells: Add method for access and remove the first `const`.

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -0,0 +1,222 @@
...
+  /// Values attached to mesh entities
+  Eigen::Array<T, Eigen::Dynamic, 1>& values();
+
+  /// Values attached to mesh entities (const.)
+  const Eigen::Array<T, Eigen::Dynamic, 1>& values() const;
+
+  /// Append new indices with their values
+  /// @param[in] indices
+  /// @param[in] values
+  void append(const Eigen::Array<std::int32_t, Eigen::Dynamic, 1>& indices,
```
garth-wells: Might be nicer to (also) have a `std::vector<std::pair<int, T>>` interface.

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -0,0 +1,151 @@
...
+  /// Create from entities of given dimension on a mesh
+  /// @param[in] mesh The mesh associated with the tags
+  /// @param[in] dim Topological dimension of mesh entities
+  ///    to tag.
+  /// @param[in] indices Array of indices, will be copied, sorted
+  ///    with duplicates removed. Local-to-process.
+  /// @param[in] values Array of values attached to indices,
+  ///    will be copied, sorted and duplicates removed according
+  ///    to indices array.
+  MeshTags(const std::shared_ptr<const Mesh>& mesh, int dim,
```
garth-wells: It would be helpful to add a flag that allows caller to say that the input data is unique and sorted.

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -0,0 +1,151 @@
...
+  int dim() const;
+
+  /// Return mesh
+  std::shared_ptr<const Mesh> mesh() const;
+
+  /// Name
+  std::string name = "mesh_tags";
+
+  /// Unique ID
+  std::size_t id = common::UniqueIdGenerator::id();
```
garth-wells: `id` should be private, with a const member function to access it.

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -0,0 +1,151 @@
...
+
+  //
+  // Sort indices and values according to indices and remove duplicates
+  //
+
+  // Prepare the sorting permutation
+  std::vector<int> perm(indices.size());
+  std::iota(perm.begin(), perm.end(), 0);
+
+  std::sort(perm.begin(), perm.end(), [&](const int& a, const int& b) {
```
garth-wells: Does the `&` capture everything? Would be good to restrict visibility and use `const` where possible.

PR #890, opened 2020-03-17
Lambda function assembler

chrisrichardson:
Replaces `MatSetValuesLocal` in the assembler with a lambda function (which should capture the matrix object).

This should allow us more flexibility with linear algebra backends and assembly in general.


chrisrichardson:
There aren't  any new `typedefs`, only template parameters.

chrisrichardson:
And there is explicit instantiation for `PetscScalar`. The problem with instantiating both `double` and `complex` is the call to `Form::Integrals::get_tabulate_tensor` which I haven't templated yet. I just wanted to get something that works, to build on further. We need to assess carefully, to avoid viral templating.

PR #892, opened 2020-03-18
Pack permutation info

mscroggs:
I have packed data that was stored in `face_rotations`, `face_reflections` and `edge_reflections` into a single 32 bit int for each cell. This pull request is coupled with FEniCS/ffcx#227.

For 2D cells, one bit is used for each edge, to represent whether or not the edge is reversed: the least significant bit is for edge 0, the next for edge 1, etc.

For example, if a quadrilateral has cell permutation info ....0111 then:
 - edge 0 is reflected (1)
 - edge 1 is reflected (1)
 - edge 2 is reflected (1)
 - edge 3 is not permuted (0)

For 3D cells, three bits are used for each face, and for each edge: the least significant bit says whether or not face 0 is reflected, the next 2 bits say how many times face 0 is rotated; the next three bits are for face 1, then three for face 2, etc; after all the faces, there is 1 bit for each edge to say whether or not they are reversed.

For example, if a tetrahedron has cell permutation info ....011010010101001000 then (from right to left):
 - face 0 is not permuted (000)
 - face 1 is reflected (001)
 - face 2 is rotated twice then reflected (101)
 - face 3 is rotated once (010)
 - edge 0 is not permuted (0)
 - edge 1 is reflected (1)
 - edge 2 is not permuted (0)
 - edge 3 is reflected (1)
 - edge 4 is reflected (1)
 - edge 5 is not permuted (0)


mscroggs:
I tried it (Removed in https://github.com/FEniCS/dolfinx/pull/892/commits/1ffde44da207a8bde8967a1e098a664c26ad5409) but converting them to the right type int to pass into ffc seemed more complicated overall than using bit operations

garth-wells:
I was thinking more just as a helper to then convert the data to `uint32_t` via https://en.cppreference.com/w/cpp/utility/bitset/to_ulong.

PR #895, opened 2020-03-20
Flake8 fix for #890

francesco-ballarin:
Not sure why CI did not catch these, but a manual run on my installation did.


francesco-ballarin:
I am not sure. I usually run with `pytest -m flake8 --flake8`, while on CI there is `python3 -m flake8`, but that is not supposed to make any difference

francesco-ballarin:
Got it, see #896

PR #907, opened 2020-03-25
Cast inside pybind

chrisrichardson:
Fixes a warning about an implicit cast to int in python, by casting inside pybind11 instead.

### DIFF
--- a/python/dolfinx/wrappers/fem.cpp
+++ b/python/dolfinx/wrappers/fem.cpp
@@ -480,7 +480,7 @@ void fem(py::module& m)
              auto tabulate_tensor_ptr = (void (*)(
                  PetscScalar*, const PetscScalar*, const PetscScalar*,
                  const double*, const int*, const std::uint8_t*,
-                 const std::uint32_t))addr.cast<std::intptr_t>();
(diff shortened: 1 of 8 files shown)


python/dolfinx/wrappers/fem.cpp
```diff
@@ -476,11 +476,11 @@ void fem(py::module& m)
...
-              int i, std::intptr_t addr) {
-             auto tabulate_tensor_ptr
-                 = (void (*)(PetscScalar*, const PetscScalar*,
-                             const PetscScalar*, const double*, const int*,
-                             const std::uint8_t*, const std::uint32_t))addr;
+              int i, py::object addr) {
+             auto tabulate_tensor_ptr = (void (*)(
+                 PetscScalar*, const PetscScalar*, const PetscScalar*,
+                 const double*, const int*, const std::uint8_t*,
+                 const std::uint32_t))addr.cast<std::intptr_t>();
```
garth-wells: Would it be better as. `std::uintptr_t`?

chrisrichardson: Could be

PR #897, opened 2020-03-21
Generate C++ demo and test files in build directory, rather than source directory

francesco-ballarin:
This PR changes CMakeLists.txt and form generation stages for C++ demos and tests so that the files are generated in the build directory and not in the source tree.

@RizzerOnGitHub I believe this will fix both #888 and #893. If possible start from a clean repo and a clean build dir.

The changes included here are not fully backward compatible. In particular, the demos and unittests target are not added anymore to the main makefile. Instead, demos and tests should be configured as standalone projects after the core library has been installed. CI has been updated to reflect this change. I believe that this separation is beneficial for CI testing, because these standalone projects (especially demos) were previously not covered, but could have been used by end users.

The only remaining build file to still be generated in the source dir is now dolfinx/common/version.h.

PR #910, opened 2020-03-29
Github Actions testing

garth-wells:
- Adds Github Actions testing
- Build a single testing container using Github Actions that include the matrix of PETSc/SLEPc builds (real, complex) x (32-bit indices, 64-bit indices), including `petsc4py` and `slepc4py`. Use `PETSC_ARCH` to select the build
- Fixes for the clang/llvm 10
- A bunch of clang-tidy static analyser and style fixes


.github/workflows/ccpp.yml
```diff
@@ -0,0 +1,106 @@
...
+      - name: Configure C++
+        run: cmake -G Ninja -DCMAKE_BUILD_TYPE=Developer -B build -S cpp/
+      - name: Build and install C++ library
+        run: |
+          cmake --build build
+          cmake --install build
+
+      - name: Build and run C++ unit tests (serial and MPI)
+        run: |
+          cmake --build ./build -t unittests
```
francesco-ballarin: This line (and the similar one in demos) would change in #897. Let me know if you are willing to merge this before #897, and I will rebase accordingly.

cpp/dolfinx/io/pugixml.cpp
```diff
@@ -11,24 +11,24 @@
  * Copyright (C) 2003, by Kristen Wegner (kristen@tima.net)
  */
 
-#ifndef SOURCE_PUGIXML_CPP
-#define SOURCE_PUGIXML_CPP
+#ifndef _HOME_GARTH_CODE_FENICS_DOLFINX_CPP_DOLFINX_IO_PUGIXML_CPP
```
francesco-ballarin: This feels like a sed typo?

garth-wells: That is a bit weird . . .

.github/workflows/ccpp.yml
garth-wells: I can merge this first.

francesco-ballarin: OK, I will update #897 as soon as this is merged.

PR #919, opened 2020-04-01
Restore in-place generation of CMakeLists.txt files for demos

francesco-ballarin:
This is a follow up to #897, that restores the capability to have automatically generated CMakeLists.txt in the demos source directory, as per @garth-wells request.

The changes introduced in #897 to accommodate the packaging pipeline by @RizzerOnGitHub are still in place (please @RizzerOnGitHub confirm by running your pipeline again, if possibile). This PR only fixes the non-backward compatible change that was introduced in #897: end users (and CI) are still supposed to configure and build demos in the build (or install directory), but CMakeLists.txt files are generated in the source tree as well (as it used to be before #897) to ensure better backward compatibility and easier development of new demos.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -304,8 +304,9 @@ endif()
 
 if (GENERATE_DEMO_TEST_DATA)
   message(STATUS "")
-  message(STATUS "Generating CMakeLists.txt files in demo and test directories")
+  message(STATUS "Generating CMakeLists.txt files in demo directory")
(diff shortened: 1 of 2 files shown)


garth-wells:
Generated CMake files are missing the name of the (to be) generated form file.

francesco-ballarin:
Sorry @garth-wells , I didn't realize this. It should be fixed now.

RizzerOnGitHub:
I'm waiting for a HDF5 library transition to push through in Debian, then I can test.

PR #924, opened 2020-04-02
Use 32-bit index for DofMap

chrisrichardson:
Remove PetscInt from DofMap and assembly, copying dofs to 64-bit on-the-fly if PETSc is configured with 64-bit-indices.

### DIFF
--- a/cpp/dolfinx/fem/assemble_matrix_impl.cpp
+++ b/cpp/dolfinx/fem/assemble_matrix_impl.cpp
@@ -87,16 +87,12 @@ void fem::impl::assemble_matrix(
   }
 }
 //-----------------------------------------------------------------------------
-// \cond doxygen should ignore
-
 // Explicit instantiation with PetscScalar
 template void fem::impl::assemble_matrix<PetscScalar>(
     const std::function<int(std::int32_t, const std::int32_t*, std::int32_t,
                             const std::int32_t*, const PetscScalar*)>&
         mat_set_values_local,
     const Form& a, const std::vector<bool>& bc0, const std::vector<bool>& bc1);
-
-// \endcond
 //-----------------------------------------------------------------------------
 template <typename ScalarType>
 void fem::impl::assemble_cells(
--- a/cpp/dolfinx/fem/assemble_matrix_impl.cpp
+++ b/cpp/dolfinx/fem/assemble_matrix_impl.cpp
@@ -87,12 +87,14 @@ void fem::impl::assemble_matrix(
   }
 }
 //-----------------------------------------------------------------------------
+// @cond - protect from Doxygen
 // Explicit instantiation with PetscScalar
 template void fem::impl::assemble_matrix<PetscScalar>(
     const std::function<int(std::int32_t, const std::int32_t*, std::int32_t,
                             const std::int32_t*, const PetscScalar*)>&
         mat_set_values_local,
     const Form& a, const std::vector<bool>& bc0, const std::vector<bool>& bc1);
+// @endcond
 //-----------------------------------------------------------------------------
 template <typename ScalarType>
 void fem::impl::assemble_cells(
--- a/cpp/dolfinx/fem/assemble_matrix_impl.cpp
+++ b/cpp/dolfinx/fem/assemble_matrix_impl.cpp
@@ -87,7 +87,8 @@ void fem::impl::assemble_matrix(
   }
 }
 //-----------------------------------------------------------------------------
-// @cond - protect from Doxygen
+// @cond
+// protect from Doxygen
 // Explicit instantiation with PetscScalar
 template void fem::impl::assemble_matrix<PetscScalar>(
     const std::function<int(std::int32_t, const std::int32_t*, std::int32_t,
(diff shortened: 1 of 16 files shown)


cpp/dolfinx/fem/assemble_matrix_impl.cpp
```diff
@@ -56,56 +56,56 @@ void fem::impl::assemble_matrix(
...
     const std::vector<std::int32_t>& active_facets
         = integrals.integral_domains(type::interior_facet, i);
-    fem::impl::assemble_interior_facets<IndexType, ScalarType>(
+    fem::impl::assemble_interior_facets<ScalarType>(
         mat_set_values_local, mesh, active_facets, dofmap0, dofmap1, bc0, bc1,
         fn, coeffs, c_offsets, constant_values);
   }
 }
 //-----------------------------------------------------------------------------
 // \cond doxygen should ignore
```
garth-wells, line 90: I think we can remove these doxygen statements now that the test image has a more recent doxygen.

cpp/dolfinx/fem/assembler.cpp
```diff
@@ -23,6 +23,57 @@
 using namespace dolfinx;
 using namespace dolfinx::fem;
 
+namespace
+{
+
+#ifdef PETSC_USE_64BIT_INDICES
+const std::function<int(std::int32_t, const std::int32_t*, std::int32_t,
+                        const std::int32_t*, const PetscScalar*)>
+make_petsc_lambda(Mat& A, std::vector<PetscInt>& storage_p)
```
garth-wells, line 32: I think it should be `Mat A`. `Mat` is already a pointer.

cpp/dolfinx/fem/assembler.cpp
```diff
@@ -23,6 +23,57 @@
...
+{
+
+  const std::function<int(std::int32_t, const std::int32_t*, std::int32_t,
+                          const std::int32_t*, const PetscScalar*)>
+      f = [&A, &storage_p](std::int32_t nrow, const std::int32_t* rows,
```
garth-wells, line 37: `A` rather than `&A`.

cpp/dolfinx/fem/assembler.cpp
```diff
@@ -23,6 +23,57 @@
...
+          la::petsc_error(ierr, __FILE__, "MatSetValuesLocal");
+#endif
+        return 0;
+      };
+  return f;
+}
+#else
+const std::function<int(std::int32_t, const std::int32_t*, std::int32_t,
+                        const std::int32_t*, const PetscScalar*)>
+make_petsc_lambda(Mat& A, std::vector<PetscInt>&)
```
garth-wells, line 57: See above comments on `A` vs `&A`.

cpp/dolfinx/fem/assembler.cpp
```diff
@@ -186,19 +229,10 @@ void fem::assemble_matrix(
...
-             const PetscInt* cols, const PetscScalar* y) {
-          PetscErrorCode ierr
-              = MatSetValuesLocal(A, nrow, rows, ncol, cols, y, ADD_VALUES);
-#ifdef DEBUG
-          if (ierr != 0)
-            la::petsc_error(ierr, __FILE__, "MatSetValuesLocal");
-#endif
-          return 0;
-        };
+  std::vector<PetscInt> sp;
```
garth-wells, line 232: `sp` is a bit cryptic.

python/test/unit/fem/test_nonlinear_assembler.py
```diff
@@ -301,7 +301,7 @@ def boundary(x):
 
     snes.getKSP().setType("preonly")
     snes.getKSP().getPC().setType("lu")
-    snes.getKSP().getPC().setFactorSolverType("mumps")
+    snes.getKSP().getPC().setFactorSolverType("superlu_dist")
```
garth-wells, line 304: PETSc 3.13 is out, and allow MUMPS to be configured for 64-bit indices.  Just waiting on petsc4py before updating the test system,

cpp/dolfinx/fem/assembler.cpp
chrisrichardson: I don't know why I called it that, in fact.

cpp/dolfinx/fem/assembler.cpp
chrisrichardson: I am a bit torn about `A` vs `&A`. Practically it makes no difference - a reference to a pointer, versus copying the pointer, right? But if `Mat` was an object, then this would show that it is not `const` - we are changing it in the lambda. I understand your point, but I think the reference shows the intent...

garth-wells: But we're not changing `A` - we're changing what `A` points to. We overcomplicated interfacing to PETSc in the past because we treated `Mat` as an object rather than as a pointer.

python/test/unit/fem/test_nonlinear_assembler.py
chrisrichardson: Do you want to switch back to mumps, or just leave it on `superlu_dist`?

garth-wells: Let's leave it as `superlu_dist` for now - gives use test coverage with 64-bit indices.

PR #937, opened 2020-04-06
Use boost::filesystem instead of std::filesystem for compatibility with gcc < 9

francesco-ballarin:


chrisrichardson:
I think this can be fixed by linking with `-lstdc++fs` rather than going back to boost.

michalhabera:
I do not have a strict opinion here, because I always work with new tools and compiler. Google suggests https://stackoverflow.com/questions/53201991/how-to-use-stdfilesystem-on-gcc-8

francesco-ballarin:
> I think this can be fixed by linking with `-lstdc++fs` rather than going back to boost.

This is true, but then I would argue that we should be consistent across the codebase. Right now we are using both boost and std.

garth-wells:
Might also be an issue with Intel compilers. Not very nice if users need to add a flag by hand.

garth-wells:
I suggest we merge this, and make the switch to `std::filesystem` once gcc 9 is more widespread.

PR #940, opened 2020-04-06
Simplify (and remove) MPI wrappers

garth-wells:
- These wrappers are largely left-over from when MPI was an optional dependency. Some were unused.
- Removes some home-grown wrappers from Python (not yet all). Use `mpi4py` instead. The remainder can be removed later.
-  Easy-to-use MPI wrappers that hide/obscure underlying complexity/cost get mis-used in the core code.
- Avoid potentially expensive `std::vector<std::vector<T>>` interface to all_to_all.


chrisrichardson:
Let's also get rid of the optional put/get implementation of all_to_all - it is not any improvement.

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -335,28 +333,33 @@ create_meshtags(MPI_Comm comm, const std::shared_ptr<const mesh::Mesh>& mesh,
...
-    for (std::int32_t e = 0; e < num_recv_ents; ++e)
+    auto recv_ents_p = recv_ents.links(i);
+    auto recv_vals_p = recv_vals.links(i);
+    for (std::int32_t e = 0; e < recv_ents_p.rows() / nnodes_per_entity; ++e)
     {
-      std::vector<std::int64_t> _entity(&recv_ents[i][nnodes_per_entity * e],
-                                        &recv_ents[i][nnodes_per_entity * e]
-                                            + nnodes_per_entity);
+      std::vector<std::int64_t> _entity(
+          recv_ents_p.data() + nnodes_per_entity * e,
```
michalhabera, line 355: I think `recv_ents_p[e].data()` and `recv_ents_p[e].data() + nnodes_per_entity` could work here.

garth-wells: Code block now substantially re-structured.

PR #944, opened 2020-04-07
Initialize all required connectivities in dolfinx::mesh::locate_entities_geometrical

francesco-ballarin:
Adds some missing connectivity initializations


michalhabera:
I think our new policy is to throw if a connectivity is missing, @garth-wells can specify. Motivation is to be more explicit and do not hide expensive operations inside other function calls. I already added such checks into IO.

francesco-ballarin:
> I think our new policy is to throw if a connectivity is missing, @garth-wells can specify. Motivation is to be more explicit and do not hide expensive operations inside other function calls. I already added such checks into IO.

I agree in general on this policy, but in this particular case how would the user be supposed to know (before seeing the error) that the internal implementation of such a function requires connectivity built for vertices (lines 739-740)?
As an average user, I might realize that if I am asking to locate entities of dimension `dim` it might be my responsibility to create connectivities associated to dimension `dim`, but I would certainly not be able to guess that internally this requires also initializing some connectivity for vertices. Or would I?

garth-wells:
I'm not entirely sure what to do. We shouldn't be violating const-ness of the Mesh deep in the code, but maybe we can live with it at the highest levels for usability.

Or, we strictly respect const-ness of the Mesh in the  C++ layer and do behind-the-scenes initialisation in Python?

michalhabera:
@garth-wells What benefits could we get from having strict const-corectness on Mesh in its deepest sense? Wouldn't it be better and more sustainable to keep only const corectness which compiler can catch?

I like being more explicit, but as @francesco-ballarin says, it could be not necessary here and force user to run the code first to collect throwing error and post-tweak his code. I think doing connectivity computations inside the code is just less explicit, but surely has not overall performance harm.

garth-wells:
@michalhabera Violating const-ness only at the highest level means that most functions are cons-correct. It makes it 'less bad'.

How about we dispense with all magic const-violating mesh initialisation in the C++ layer? There is some potential performance harm in terms of memory usage. If a user explicitly initialises mesh connectivity data they can choose to discard it later. If it's magically created the memory implications are easily overlooked.

francesco-ballarin:
@mrambausek @michalhabera @garth-wells since the changes are very minor, would anybody of you disagree with merging this now? These changes can obviously be reverted as part of the bigger refactor undergoing in #951 (or its follow up that we are discussing there)

PR #946, opened 2020-04-07
Make RTCF spaces work on quad meshes

mscroggs:
Restructured and simplified the fem pipeline tests, and added tests for all elements in the [Periodic Table of the Finite Elements](http://femtable.org/).

Currently on quads, only the tests for RTCF are active as these are the only elements that are implemented once FEniCS/ffcx#232 is merged


mscroggs:
Closing for now until bugs are fixed

PR #947, opened 2020-04-07
Avoid copies in Mesh construction

garth-wells:
Store actual topology/geometry objects (rather than pointers) and add move semantics to constructor.


cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -239,44 +201,45 @@ void Mesh::create_connectivity(int d0, int d1) const
   // connectivities are needed.
 
   // Attach connectivities
-  Mesh* mesh = const_cast<Mesh*>(this);
+  Topology& topology = const_cast<Topology&>(_topology);
```
michalhabera, line 204: Is this ok? I would say the whole method should be not marked const and then no need for this cast?

garth-wells: This is the issue we've been discussing - if the member function was not const we could then not create entities/topology for meshes the are const (which we do all over).

PR #949, opened 2020-04-08
Avoid const_cast for manipulating the cache-like members of a const object

mrambausek:
The rationale of this draft pull request is to find a cleaner solution to the modification or lazy initialization of members of a const object through non-members. 
Such usages of `const_cast` appear in Mesh.cpp and also in graph partitioning functions.
One drawback of const_casts is that the object does not know that someone messes with it's internals, which may lead to errors that are difficult to track down.
As an alternative approach I suggest to use "mutable" for such members that need to be modified on const objects from outside together with a dedicated accessor. From here, one could go further
and for example set some kind of lock/flag if such an accessor to a mutable member is called. 
This could further indicate a temporarily "undefined" state of an object and/or the need for some recomputation.

What is beyond my knowledge is, why the "partition" functions 
`dolfinx::graph::*::partition`
actually have a const adj_graph as argument, which obviously is to be modified.

I am willing to spend some time on this and learn about the internals dolfin-X. So any feedback in this draft, also whether it actually has a reasonable objective or not, is welcome.

### DIFF
--- a/cpp/dolfinx/mesh/Mesh.cpp
+++ b/cpp/dolfinx/mesh/Mesh.cpp
@@ -154,9 +154,8 @@ std::int32_t Mesh::create_entities(int dim) const
 {
   // This function is obviously not const since it may potentially
   // compute new connectivity. However, in a sense all connectivity of a
-  // mesh always exists, it just hasn't been computed yet. The
-  // const_cast is also needed to allow iterators over a const Mesh to
-  // create new connectivity.
+  // mesh always exists, it just hasn't been computed yet. This resembles the
+  // notion of a cache or lazy construction.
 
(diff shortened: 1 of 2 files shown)


francesco-ballarin:
Thanks @mrambausek for the PR.

We have indeed been discussing recently about const correctness for Mesh classes, see e.g. #947 and #944 .
I suggest to coordinate the efforts you will carry out in this PR especially with @garth-wells @michalhabera

mrambausek:
Thanks @francesco-ballarin, I definitely do not want to proceed here in an un-coordinated way. What's the preferred to discuss such API things -- just via pull-requests/comments or is there another regular discussion/messaging channel?

mrambausek:
Superseded by #951

Issue #948, opened 2020-04-07
Add DirichletBC interface with MeshTags


jorgensd:
It would very convenient in both Python and C++ to have the following interface
`DirichletBC(FunctionSpace V, Function u, MeshTag mt, index i)` as we had with `MeshFunction`. 
Right now we have to backwards engineer to get the dofs given the facets and an index.

francesco-ballarin:
@michalhabera if I remember correctly one motivation for the lower level DirichletBC was to handle more easily the case of BCs on a component of a mixed function space. I fail to see how having made the switch from MeshFunction to MeshTags would change the underlying motivation in this case, and allow to bring the old interface back.

michalhabera:
Yes, I am also against the change proposed here. The motivation for new DirichletBC interface was to simplify the polymorphic DirichletBC interface and provide full control to the user.
A bit more sense would make to have `locate_dofs_topological(MeshTags mt, value i)` but its doing something very trivial anyway...
@jorgensd It is not now backwards engineering to get dofs, dof indices are what you need to apply dirichlet BC on algebraic level.

jorgensd:
> Yes, I am also against the change proposed here. The motivation for new DirichletBC interface was to simplify the polymorphic DirichletBC interface and provide full control to the user.
> A bit more sense would make to have `locate_dofs_topological(MeshTags mt, value i)` but its doing something very trivial anyway...
> @jorgensd It is not now backwards engineering to get dofs, dof indices are what you need to apply dirichlet BC on algebraic level.

In Python it is possible to do the meshtag to DirichletBC with 2-3 lines of code, Which isn’t terrible. 
In C++, @chrisrichardson and I was working on a demo with ~25 BCs, Which ends up having alot of code to go from a MeshValueCollection to the dof values. 

I agree that this would be solved by having a `locate_dofs_topological(meshtag mt, value i)`would solve the issue

garth-wells:
@jorgensd Can you post some sample code? It would be helpful to start from the issue to see if there are refinements that could be made.

jorgensd:
Sample code: 
```
  // Set up boundary condition on inner surface
  const std::vector<std::int32_t>& facet_array = facet1.values();
  const std::vector<std::int32_t>& facet_indices = facet1.indices();
  int n2 = std::count(facet_array.begin(), facet_array.end(), 2);
  Eigen::Array<std::int32_t, Eigen::Dynamic, 1> facets2(n2);
  int i2 = 0;
[...]

PR #950, opened 2020-04-08
Added meshtag convenience function

jorgensd:
Following the discussion in #948 , Here is a convenience function for locating the indices in `MeshTags` with a given value. Only exposing this functionality in C++ as it is trivial in python


cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -75,6 +75,26 @@ class MeshTags
   /// Move assignment
   MeshTags& operator=(MeshTags&& tags) = default;
 
+  /// Find all MeshTag indices of a given value, local-to-process
+  /// @param[in] i The value
+  /// @return The indices
+  std::vector<std::int32_t> find_indices(T i) const
```
michalhabera, line 81: I would call argument `T value`, its more telling.

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -242,6 +242,15 @@ void mesh(py::module& m)
...
                                    py::none());                                \
                              })                                                \
+      .def(                                                                    \
+          "find_indices",                                                      \
+          [](dolfinx::mesh::MeshTags<SCALAR>& self, SCALAR i) {                \
+            std::vector<std::int32_t> indices_i = self.find_indices(i);        \
+            return py::array_t<std::int32_t>(indices_i.size(),                 \
+                                             indices_i.data());                \
+          },                                                                   \
+          "Return MeshTag indices having value i",                             \
```
francesco-ballarin, line 252: typo: MeshTags

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -242,6 +242,15 @@ void mesh(py::module& m)
                                    self.values().size(), self.values().data(), \
                                    py::none());                                \
                              })                                                \
+      .def(                                                                    \
+          "find_indices",                                                      \
+          [](dolfinx::mesh::MeshTags<SCALAR>& self, SCALAR i) {                \
+            std::vector<std::int32_t> indices_i = self.find_indices(i);        \
+            return py::array_t<std::int32_t>(indices_i.size(),                 \
```
francesco-ballarin, line 249: Why is pybind11 not automatically converting the output of
```
.def("find_indices", &dolfinx::mesh::MeshTags<SCALAR>::find_indices)
```
to a list?

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -75,6 +75,26 @@ class MeshTags
...
+  {
+    int n = std::count(_values.begin(), _values.end(), i);
```
michalhabera, line 83: I would do rather
```cpp
std::vector<std::int32_t> indices;
indices.reserve(_indices.size());
for (std::int32_t i = 0; i < _values.size(); ++i)
{
  if (_values[i] == value)
    indices.push_back(_indices[i]);
}
return indices;
```

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -242,6 +242,15 @@ void mesh(py::module& m)
...
+                                             indices_i.data());                \
+          },                                                                   \
+          "Return MeshTag indices having value i",                             \
+          py::return_value_policy::reference)                                  \
```
michalhabera, line 253: I think you should just have automatic (default) rvp here. The automatic should be `move` here which is what you want.

cpp/dolfinx/mesh/MeshTags.h
michalhabera, line 81: Could the name be just `std::vector<std::int32_t> indices(T value) const`?

python/dolfinx/wrappers/mesh.cpp
michalhabera: It is and that is what really hurts. We do not want a list of ints here but rather a numpy array (because of all issues why iterating over large python's list is terrible idea, data locality etc)

cpp/dolfinx/mesh/MeshTags.h
garth-wells: That could use a lot more memory than required. Perhaps:
```
int n = std::count(_values.begin(), _values.end(), i);
indices.reserve(n);
for (std::int32_t i = 0; i < _values.size(); ++i)
  if (_values[i] == value)
    indices.push_back(_indices[i]);

return indices;
```

cpp/dolfinx/mesh/MeshTags.h
garth-wells: The return type needs to be an Eigen type to avoid a copy when passing through to Python.

cpp/dolfinx/mesh/MeshTags.h
michalhabera: Maybe we should go without fancy `reserve`. I think `std::count()` has surely complexity O(n) in container size which is also not good. Just a pure push_back has sufficiently clever preallocation strategy.

garth-wells: Clever -> potentially large memory overhead, needing `shrink` afterwards. The search is already O(n), so preallocating doesn't affect complexity and I expect will be faster.

cpp/dolfinx/mesh/MeshTags.h
michalhabera: @garth-wells We talked about not complicating return types with Eigen where not needed? Would be good if we find a way how to return py::array_t and make sure the underlying std::vector is moved.

jorgensd: One thing we could do, is to not expose this function to python, as in python it is quite easy to get the tags. If we only expose this in C++, it can be kept as a list.

garth-wells: Good point - can just use NumPy.

chrisrichardson:
Looking at this, maybe it is worthwhile also considering the case where multiple tags are required? Would an interface like: `std::vector<std::vector<std::int32_t>> indices(std::vector<T> values)` make sense? It could be more efficient when pulling out multiple sets at the same time...

jorgensd:
> Looking at this, maybe it is worthwhile also considering the case where multiple tags are required? Would an interface like: `std::vector<std::vector<std::int32_t>> indices(std::vector<T> values)` make sense? It could be more efficient when pulling out multiple sets at the same time...

Sounds good to me. Summarizing, I will remove the binding to python, use reserve on the vector, and make a for-loop over the function Garth suggested above.

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -75,6 +75,26 @@ class MeshTags
...
+  /// @return The correspodning indices
+  std::vector<std::vector<std::int32_t>> indices(std::vector<T> values) const
+  {
+    std::vector<std::vector<std::int32_t>> indices;
+    for (std::int32_t i = 0; i < values.size(); ++i)
+    {
+      std::vector<std::int32_t> indices_i;
+      int n = std::count(_values.begin(), _values.end(), values[i]);
+      indices_i.reserve(n);
+      for (std::int32_t j = 0; j < _values.size(); ++j)
```
michalhabera, line 90: Body of this `for` is more than 1 line so you should enclose it in `{}`.

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -75,6 +75,28 @@ class MeshTags
   /// Move assignment
   MeshTags& operator=(MeshTags&& tags) = default;
 
+  /// Find all MeshTags for a list of values and return the corresponding
+  /// indices per value, local-to-process
+  /// @param[in] values A list values
+  /// @return The correspodning indices
+  std::vector<std::vector<std::int32_t>> indices(std::vector<T> values) const
```
garth-wells, line 82: Add `const` and `&`: `std::vector<std::vector<std::int32_t>> indices(const std::vector<T>& values) const`

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -75,6 +75,28 @@ class MeshTags
...
+  {
+    std::vector<std::vector<std::int32_t>> indices;
+    for (std::int32_t i = 0; i < values.size(); ++i)
```
garth-wells, line 85:
```cpp
for (std::int32_t value : values)
```

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -75,6 +75,28 @@ class MeshTags
...
+    {
+      std::vector<std::int32_t> indices_i;
+      int n = std::count(_values.begin(), _values.end(), values[i]);
+      indices_i.reserve(n);
+      for (std::int32_t j = 0; j < _values.size(); ++j)
+      {
+        if (_values[j] == values[i])
+          indices_i.push_back(_indices[j]);
+      }
+      indices.push_back(indices_i);
```
garth-wells, line 95: It would be good to restructure this `push_back` to ensure that `indices_i` is not copied.

chrisrichardson:
In the case of many indices, it might be more efficient to traverse the data only once. We could make e.g. a map before the main loop, and then add to each vector as the index occurs.

cpp/dolfinx/mesh/MeshTags.h
michalhabera: He needs the index also, see below.

jorgensd:
@chrisrichardson , @garth-wells, @michalhabera I've now added an hybrid of all the comments above. The big question is if we need to reserve space in the vector on before-hand, or just let push_back do its magic.

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -76,6 +76,32 @@ class MeshTags
   /// Move assignment
   MeshTags& operator=(MeshTags&& tags) = default;
 
+  /// Find all MeshTags for a list of values and return the corresponding
+  /// indices per value, local-to-process
+  /// @param[in] values A list values
+  /// @return The corresponding indices
+  const std::vector<std::vector<std::int32_t>>&
+  indices(std::vector<T> values) const
```
michalhabera, line 84: @garth-wells surely meant to put `const std::vector<T>& values` as an argument, not to return const vector by reference. What you have here is wrong.

michalhabera:
We should not overcomplicate, its amazing how many comments we made for about < 50 lines of code... This is just a helper, if users has some corner-case of many values being matched he is not forced to use this helper.

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -76,6 +76,32 @@ class MeshTags
   /// Move assignment
   MeshTags& operator=(MeshTags&& tags) = default;
 
+  /// Find all MeshTags for a list of values and return the corresponding
+  /// indices per value, local-to-process
+  /// @param[in] values A list values
+  /// @return The corresponding indices
+  std::vector<std::vector<std::int32_t>>
```
garth-wells, line 83: Is this a helpful return type given that `locate_dofs_topological` takes an Eigen array?

cpp/dolfinx/mesh/MeshTags.h
```diff
@@ -76,6 +76,32 @@ class MeshTags
...
+  indices(const std::vector<T>& values) const
```
garth-wells, line 84: The function names need to better indicate that this isn't a no-op call, making it very different to `MeshTags::indices()`.

jorgensd: The initial name was `find_indices`. Any other suggestion? `locate_value_inidices`?

PR #951, opened 2020-04-08
Move topology computation into topology

mrambausek:
This PR supersedes the draft PR #949 after some discussion on how to move on.

What happens here is that
 1. All members that modify the `Topology` associated to a mesh are now directly called on the topology object instead of having the mesh more or less forwarding these things.
 2. For this to work also for `const Mesh`, there is a new getter `topology_mutable` which returns
a non-const `Topology&` even for `const Mesh` [works because `Mesh::_topology` is mutable].
The choice to *not* just name the getter `topology` is to avoid breaking usual C++ API patterns/conventions. Moreover, for `const Mesh` it makes sense to have one `const Topology&` getter for just reading (maybe doing minimal computations internally) and one getter for "expert" or "expensive" tasks.

There is some need for cleanup in the calling logic, i.e. where the Topology object is manipulated
but this involves some more thinking about the internals of `Topology` and how we want to use it.

The main work in this PR so far is in the first two commits of which the first moved things to `Topology` keeping the old API intact, the second changed the `Mesh` API and updated the corresponding calls.
Hence, one could just stop after the first commit, if the API changes turn out to have undesired side-effects.

Later commits try to fix some oddities.


mrambausek:
Currently, checks fail because of `Topology` having now an `MPIComm` member which has its copy *assignment* deleted. Clang10 then forbids copy-assignment of `Topology`. Interestingly, it does not complain about identical code in `Mesh`. GCC does not complain.

mrambausek:
Following the complaints of clang: should Mesh and/or Topology support copy assignment?

mrambausek:
Python test currently fail because they have not yet been adapted to the API changes.

python/cmake/FindPETSc4py.cmake
```diff
@@ -45,6 +45,16 @@ if(PETSC4PY_INCLUDE_DIRS)
   set(PETSC4PY_FIND_QUIETLY TRUE)
 endif(PETSC4PY_INCLUDE_DIRS)
 
+message(STATUS ${PYTHON_EXECUTABLE})
```
michalhabera, line 48: Why is this change here?

python/cmake/FindPETSc4py.cmake
```diff
@@ -45,6 +45,16 @@ if(PETSC4PY_INCLUDE_DIRS)
...
+
+execute_process(
```
francesco-ballarin, line 50: ... and also this one: I cannot see any difference compared to lines 58-64

python/test/unit/mesh/test_mesh_utils.py
```diff
@@ -249,7 +249,7 @@ def test_topology_partition(tempdir, shape, order):
 
     # Create (i) local topology object and (ii) IndexMap for cells, and
     # set cell-vertex topology
-    topology_local = cpp.mesh.Topology(layout.cell_type)
+    topology_local = cpp.mesh.Topology(cpp.MPI.comm_self, layout.cell_type)
```
francesco-ballarin, line 252: in this test file, you will need to update calls that use `cpp.MPI` due to #956

python/cmake/FindPETSc4py.cmake
mrambausek: Ah, I believe I once had some build issues. Overlooked that...

python/cmake/FindPETSc4py.cmake
mrambausek: Same as above, I had to debug the cmake because for some reason my cmake (started from my IDE) did not choose the right python executable. I'll revert these changes.

python/test/unit/mesh/test_mesh_utils.py
mrambausek: Thanks for pointing out. I just rebased on master...

PR #955, opened 2020-04-09
Allow Github Actions to skip ci [temporary]

IgorBaratta:
Palliative solution while Github Actions does not support [skip ci].


.github/workflows/ccpp.yml
```diff
@@ -10,6 +10,7 @@ on:
 
 jobs:
   build:
+    if: "!(contains(github.event.head_commit.message, 'skip') && contains(github.event.head_commit.message, 'ci'))"
```
francesco-ballarin, line 13: Is it possible to make the check stricter?
A commit message like `Should not skip build while updating scientific publications` (note the not in front of skip, and the ci in scientific) would trigger an unexpected ci skip.

garth-wells: Agree - `[skip ci]` and `[ci skip]` seem to be the standard strings.

IgorBaratta: It is  now skipped when the commit message contains` [skip ci] `or `[ci skip]`,

PR #960, opened 2020-04-11
Simplify MeshTag construction by requiring sorted and unique constructor arguments

garth-wells:
- Makes implementation less implicit
- Move algorithms out of MeshTags
- Remove the bug-prone "double default arguments" of the same type in `MeshTags` constructor 
- Remove non-pure private functions

### DIFF
--- a/cpp/dolfinx/common/utils.h
+++ b/cpp/dolfinx/common/utils.h
@@ -23,7 +23,8 @@ namespace common
 {
 
 /// Sort two arrays based on the values in array @p indices. Any
-/// duplicate indices and the corresponding value are removed.
+/// duplicate indices and the corresponding value are removed. In the
+/// case of duplicates, the entry with the smallest value is retained.
 /// @param[in] indices Array of indices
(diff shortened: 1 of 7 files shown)


jorgensd:
> @jorgensd @chrisrichardson Please check your recent codes that create MeshTags to avoid bugs.

I'll test it on my [MPC library](https://github.com/jorgensd/dolfinx_mpc) this evening or tomorrow morning. Could you check the RR geometries @chrisrichardson ? I think that code needs a new round of updates anyhow after the many UI changes this week.

michalhabera:
@jorgensd All should be fine in IO since that now explicitly sorts etc,. Only need to check when MeshTags are created within dolfinx.

jorgensd:
> @jorgensd All should be fine in IO since that now explicitly sorts etc,. Only need to check when MeshTags are created within dolfinx.

I've got some tests doing that. Then I guess Chris does not need to do anything wit RR as those are input.

PR #964, opened 2020-04-12
Exception created but not thrown

IgorBaratta:


cpp/dolfinx/io/HDF5File.h
```diff
@@ -128,7 +128,7 @@ void HDF5File::write_data(
     dset_name = "/" + dataset_name;
 
   std::int64_t global_rows = 0;
-  MPI_Allreduce(&data.rows(), &global_rows, 1, MPI_INT64_T, MPI_SUM,
+  MPI_Allreduce(data.rows(), &global_rows, 1, MPI_INT64_T, MPI_SUM,
```
garth-wells, line 131: Let's change this to:
```cpp
std::int64_t local_rows =data.rows();
MPI_Allreduce(&local_rows, &global_rows, 1, MPI_INT64_T, MPI_SUM,
```
to avoid integer type issues.

PR #968, opened 2020-04-13
Fix UFC JIT objects memory management

### LINKED ISSUE #966: Fix memory management of ufc JIT objects
See related FFCX issue (https://github.com/FEniCS/ffcx/issues/234) and PR (https://github.com/FEniCS/ffcx/pull/235).

There are issues with the memory manage of FFCX JIT objects. Functions like
https://github.com/FEniCS/dolfinx/blob/e77360f65281cb70fba2a7d4f96c5324c2e3704a/python/dolfinx/wrappers/fem.cpp#L104 in the pybind11 layer are wrapping pointers to data that was created by malloc as `std::shared_ptr`s. This is a problem because (i) `free` rather than `delete` should be used to cleanup, and it would require a very specific sequence of calls to avoid a leak rather than the garbage collector just 'working'.

https://github.com/FEniCS/ffcx/pull/235 uses the Python garbage collector, so the DOLIFIN layer should not get in the way of this and attempt to free the memory twice.

A possible solution is to add an interface to DOLFINX objects that require a UFC object/function for construction that takes a `std::uintptr_t`.

michalhabera:
An attempt to fix #966

**Update 1:**
Helpers `make_ufc_foo` now removed, pybind11 layer takes `uintptr_t`, dereferences and calls C++ helpers directly with ufc objects.

--- 

This fix changes `make_ufc_foo` to return from `uintptr_t` an actual `ufc_foo` object.

This has the advantage in lifetime management, because anyone calling `make_ufc_foo` makes a copy of an actual ufc object pointed to. It defaults to pybind11 `return_value_policy::copy` so new returned object has lifetime now managed on Python side, while the "old" cdata structure returned directly by FFCx JIT is usually garbage collected (we do not store a reference to it, e.g. in form.py).

### DIFF
--- a/python/dolfinx/wrappers/fem.cpp
+++ b/python/dolfinx/wrappers/fem.cpp
@@ -74,50 +74,8 @@ namespace dolfinx_wrappers
 void fem(py::module& m)
 {
 
-  // UFC objects
-  py::class_<ufc_finite_element, std::shared_ptr<ufc_finite_element>>(
-      m, "ufc_finite_element", "UFC finite element object");
-  py::class_<ufc_dofmap, std::shared_ptr<ufc_dofmap>>(m, "ufc_dofmap",
-                                                      "UFC dofmap object");
-  py::class_<ufc_form, std::shared_ptr<ufc_form>>(m, "ufc_form",
-                                                  "UFC form object");
   py::class_<ufc_coordinate_mapping, std::shared_ptr<ufc_coordinate_mapping>>(
       m, "ufc_coordinate_mapping", "UFC coordinate_mapping object");
-
-  // Functions to convert pointers (from JIT usually) to UFC objects
-  m.def(
-      "make_ufc_finite_element",
-      [](std::uintptr_t e) {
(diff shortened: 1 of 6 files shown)


python/dolfinx/wrappers/fem.cpp
```diff
@@ -74,51 +74,9 @@ namespace dolfinx_wrappers
...
 {
 
-  // UFC objects
-  py::class_<ufc_finite_element, std::shared_ptr<ufc_finite_element>>(
-      m, "ufc_finite_element", "UFC finite element object");
-  py::class_<ufc_dofmap, std::shared_ptr<ufc_dofmap>>(m, "ufc_dofmap",
-                                                      "UFC dofmap object");
-  py::class_<ufc_form, std::shared_ptr<ufc_form>>(m, "ufc_form",
-                                                  "UFC form object");
   py::class_<ufc_coordinate_mapping, std::shared_ptr<ufc_coordinate_mapping>>(
```
garth-wells, line 77: Can we remove this?

michalhabera: Sure, leftover.

PR #973, opened 2020-04-14
Add tol to collision detection

jorgensd:
As discussed earlier, the dof-coordinates (from compute_physical coordiantes) and mesh.geometry.coordinates does not match to more than machine precision. 
This causes proper collision detection to fail.

This PR will introduce a machine precision tolerance when checking orient3D for different geometries.

I've only added tolerances for the test cases I've stumbled upon as of now, but if the idea here is accepted, I will add it to all collision predicates.


jorgensd:
As we are removing this code from dolfinx entirely in #1028 , this PR is not outdated

PR #976, opened 2020-04-15
Predicates cleanup

jorgensd:
This is an initial start of trying to make sense of collision detection in dolfinx.

I have removed all the code that I can see that is unused. 
Please let me know if things should be done differently (for instance not wrapping all function in the dolfinx::geometry namespace), or if there are any functions there are worth keeping as of now. To me, this code is just a massive mess, and it would be good to remove ~1300 lines.

We could also think about moving to a newer version of the code like: https://github.com/danshapero/predicates.
For reference, the orignal code can be found at: http://www.cs.cmu.edu/~quake/robust.html

### DIFF
--- a/cpp/dolfinx/geometry/predicates.cpp
+++ b/cpp/dolfinx/geometry/predicates.cpp
@@ -1,4 +1,9 @@
 #include "predicates.h"
+#include <cmath>
+#include <cstdio>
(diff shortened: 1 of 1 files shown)


jorgensd:
Not required due to #1028

PR #977, opened 2020-04-18
Fix numba cffi_support import

michalhabera:
Probably a bug in 0.49.0 Numba version, the behaviour doesn't match their documentation.

### DIFF
--- a/python/demo/mixed-elasticity-sc/static-condensation-elasticity.py
+++ b/python/demo/mixed-elasticity-sc/static-condensation-elasticity.py
@@ -11,20 +11,20 @@
 
 import os
 
-import cffi
 import numba
-import numba.cffi_support
+import numba.core.typing.cffi_utils as cffi_support
 import numpy
-from mpi4py import MPI
-from petsc4py import PETSc
 
+import cffi
 import dolfinx
 import dolfinx.cpp
 import dolfinx.io
 import dolfinx.la
-from dolfinx.mesh import locate_entities_geometrical
-from dolfinx.fem import locate_dofs_topological
 import ufl
+from dolfinx.fem import locate_dofs_topological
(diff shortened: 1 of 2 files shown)


garth-wells:
I've triggered a rebuild of the Github Actions test image to update Numba, which should allow the tests to pass.

RizzerOnGitHub:
numba 0.49 is only 4 days old. This patch breaks on installations with numba 0.48.  Is it worth allowing both?

    try:
        import numba.core.typing.cffi_utils as cffi_support
    except:
        from numba import cffi_support

michalhabera:
@RizzerOnGitHub Sounds good to me (but I'd check for ImportError, not general except). Could you make a PR?

jorgensd:
I guess it should be ModuleNotFoundError ?

garth-wells:
I wouldn't allow both - just adds clutter at this stage. And it leads to a need for a deprecation policy, a version support policy, more test coverage requirements.

michalhabera:
This is just in one a demo and test - no need for deprecation policy. The import statement is meant to be executed by user. Not critical for me anyway, current state "breaks installation" on 0.49.0 only in sense that user cannot execute correctly demos and tests.

RizzerOnGitHub:
It does add a modicum of clutter, but the current state already needs a version support policy since it currently only works with numba >= 0.49.  Not many systems will have numba 0.49 installed yet, and other software may be using the old numba API.  So all projects would need to be upgraded at the same time, causing more headaches for users.

This patch will save having to worry about those headaches and will cause no more trouble than carrying 3 lines of conditional code (unless there are other cases of numba API change elsewhere, in which case a version support policy with numba >= 0.49 would certainly be needed).

RizzerOnGitHub:
I've created PR #988 to enable the tests to work with new and old versions of numba.

PR #978, opened 2020-04-18
Avoid unintended copy

mrambausek:
Mesh, Topology and Geometry have copy construction but no copy assignment. I propose to also "delete" the copy constructors, which are probably anyway not supposed to be called. Deleting them revealed possible unintended copies. The first commit of this PR deals with the unintended copies, the second with the deletion of constructors. I chose that order because I do not know whether copy construction is totally unwanted. Note however, that I did not find any intended copy in the code.

### DIFF
--- a/cpp/dolfinx/mesh/Geometry.h
+++ b/cpp/dolfinx/mesh/Geometry.h
@@ -64,7 +64,7 @@ public:
   }
 
   /// Copy constructor
-  Geometry(const Geometry&) = default;
+  Geometry(const Geometry&) = delete;
 
   /// Move constructor
   Geometry(Geometry&&) = default;
--- a/cpp/dolfinx/mesh/Geometry.h
+++ b/cpp/dolfinx/mesh/Geometry.h
@@ -63,20 +63,29 @@ public:
     }
   }
 
-  /// Copy constructor
-  Geometry(const Geometry&) = delete;
-
   /// Move constructor
   Geometry(Geometry&&) = default;
 
   /// Destructor
   ~Geometry() = default;
 
-  /// Copy Assignment
-  Geometry& operator=(const Geometry&) = delete;
-
   /// Move Assignment
   Geometry& operator=(Geometry&&) = default;
+
+private:
+  /// Copy constructor: set private to avoid unintended copies. Use
+  /// Geometry::copy() instead.
+  Geometry(const Geometry&) = default;
+
+  /// Copy Assignment: set private to avoid unintended copies. Use
+  /// Geometry::copy_assign(const Geometry& other) instead.
+  Geometry& operator=(const Geometry&) = default;
+public:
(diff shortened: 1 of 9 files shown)


mrambausek:
Okay, the python tests reveal that copy construction is currently required because Mesh stores Geometry and Topology as object not as shared_ptr, but python handles all of them as shared_ptr. Shall these copies be avoided as well, ie. store shared_ptr<(const) Topology|Geometry>?

mrambausek:
If the copies at the python level should stay, second commit just might be dropped.

cpp/dolfinx/mesh/Geometry.h
```diff
@@ -64,7 +64,7 @@ class Geometry
   }
 
   /// Copy constructor
```
garth-wells, line 66: Probably want to change `///` -> `//` so the deleted constructor doesn't get documented. Not sure what doxygen does with deleted constructors.

cpp/dolfinx/mesh/Mesh.h
```diff
@@ -90,7 +90,7 @@ class Mesh
 
   /// Copy constructor
```
garth-wells, line 91: Probably want to change `///` -> `//`

cpp/dolfinx/mesh/Mesh.h
```diff
@@ -90,7 +90,7 @@ class Mesh
...
   /// @param[in] mesh Mesh to be copied
```
garth-wells, line 92: Deleted

cpp/dolfinx/mesh/Topology.h
```diff
@@ -67,7 +67,7 @@ class Topology
   }
 
   /// Copy constructor
```
garth-wells, line 69: Probably want to change `///` -> `//`

mrambausek:
Actually, I hesitate because the python version of Mesh's constructor using Topology and Geometry either requires copies or Mesh has to hold these members using shared_ptr<Topology|Geometry>
The copy in this case probably probably does not match the usual shared ownership of python.
What do you prefer? In any case I can just go ahead with adding a an explicit copy members.

garth-wells:
Yes, I see the issue now in the Python interface so let's keep the copy constructor and assignment operator.

mrambausek:
Just pushed a proposal with explicit copy and copy assignment. The Mesh constructor in python is probably the only rare case where this shows up. But feel free to drop this and I will just create a small PR with only the first commit.

garth-wells:
Having considered the use cases with the new `Mesh` design, I'm no longer so sure it's best to hide the copy constructor. There are some common cases where a copy is expected.

@mrambausek  Could you make a PR that avoids the unnecessary copy so we can discuss the copy semantics separately from the bug fix?

mrambausek:
See  #981

PR #986, opened 2020-04-23
Avoid copy in `create-vector`

mrambausek:
Here is a probably unintended copy of DofMap in create_vector (admittedly, that the getters returning shared_ptr are a tricky copy trap...). Also, I suppose that the "heavy" data member of DofMap, `_dofmap` could be turned into a shared_ptr<const>, which would make copies cheap.

### DIFF
--- a/cpp/dolfinx/function/Function.cpp
+++ b/cpp/dolfinx/function/Function.cpp
@@ -37,8 +37,8 @@ la::PETScVector create_vector(const function::FunctionSpace& V)
   common::Timer timer("Init dof vector");
 
   // Get dof map
-  assert(V.dofmap());
   std::shared_ptr<const fem::DofMap> dofmap = V.dofmap();
+  assert(dofmap);
 
   // Check that function space is not a subspace (view)
   assert(dofmap->element_dof_layout);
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/function/Function.cpp
```diff
@@ -38,20 +38,20 @@ la::PETScVector create_vector(const function::FunctionSpace& V)
 
   // Get dof map
   assert(V.dofmap());
```
garth-wells, line 40: Remove this line.

cpp/dolfinx/function/Function.cpp
```diff
@@ -38,20 +38,20 @@ la::PETScVector create_vector(const function::FunctionSpace& V)
...
-  const fem::DofMap& dofmap = *(V.dofmap());
+  std::shared_ptr<const fem::DofMap> dofmap = V.dofmap();
```
garth-wells, line 42: Add `assert(dofmap);`

mrambausek: doesn't the line above already do?

mrambausek: Sorry, started to read at the bottom for some reason...

PR #988, opened 2020-04-24
enable importing cffi_support from new or old numba versions

RizzerOnGitHub:
The latest numba 0.49 moved cffi_support to numba.core.typing.cffi_utils

Previously it was located at numba.cffi_support.

This patch enables tests and demos to work with both old and new
versions of numba.


RizzerOnGitHub:
This PR allows demos and tests to work with both the new version of numba, v0.49, as well as older versions, as discussed in #977.

garth-wells:
This may seem a small change, but it raises a range of issues, especially since we have not yet made a release:
- Which versions of dependencies to we support?
- If we support older versions, for how long should they be supported for?
- What combinations of dependency versions should the test system cover?

garth-wells:
Closing this now because Numba has made several releases in the meantime.

PR #989, opened 2020-04-24
Rework topology and associated data handling (draft/proposal)

mrambausek:
# Motivation

A good amount of topological data is stored as soon as one "consumer" requires it. However, 
while being careful to not create data data is not required, there are currently
no means to get rid of it, once there is not further use. Since the data creation
explicitly asks for a modification (physical, not logical) of an object, it is semantically
reasonable to allow create_XYZ only for non-const objects. This gets in the way quite often 
however. In addition, there is no simple way to compute topological data without storing anything.
As a proposal for some improvement, I suggest to introduce
 * A mechanism to drop created data. Just because I need it for one form or part of my problem,
   I might not need it for the whole simulation, Maybe I only need to inspect something...
 * Make the use of TopologyComputation and getters of the Topology more intuitive, ie.
   let them return data that is logically there (without storing by default).
 * Provide the option to cache data this is computed on demand.
 * Make caching scoped such that one can has better control of which data lands in the 
   cache and which not.

The draft code provided tries to achieve these goals and introduces several features and 
(possible) changes, of which the most important summarized below.

# Key features/changes
 * Topology and TopologyComputation should always return.
 * No data is stored by default through on-demand computations.
 * A layered cache is used to optionally cache on-demand results.
 * Cached data is discarded by releasing all locks associated to a storage layer. A release-check
   is also triggered when a lock goes out of scope (->RAII).
 * create_XYZ members work as ever.
 * Data created by create_XYZ can be discarded. This happens explicitly for the default 
   "remanent" storage and through locks if layers have been acquired (same layer mechanism as for cache, 
   but with more control when something is computed).
 * The fundamental driver is implemented quite generically in common/Memory.h
 * A simple C++ test case is provided in common/layered_memory.cpp.
 * Both mechanisms are exposed to python (see test_topology_storage in mesh/test_mesh.py)
 * All topological data is stored as shared_ptr<const ...>. This allows shallow 
   copies for all data. Whether const is okay? I would say 'yes' because changes
   of Topology invalidate a lot of things which have to be recreated.
 * New constructors: Either with a Storage object or directly with essential 
   IndexMaps and Connectivity. They replace setters which were only used for filling an 
   empty object. Their removal is connected to my expectation that changes to topological
   entail recreation instead of in-vivo-surgery of depending objects, like FunctionSpaces and  
   DofMaps.

# Open
 * Actual caching/computing policy: Should getters compute on-demand by default? Currently, 
   there are separate getters to really only read the data.
 * Actually apply the new design, avoid mesh.topology_mutable(). I deliberately did not make
   changes all over. There are enough already.
 * There is currently not way clear option to close the cache, i.e. to block subsequent writes.
   What is possible by now is to create (layers of) remanent data and do not ask for cache.
 * Currently, for better chance to remove data, layers from anywhere of the stack can be 
   removed by releasing their lock. Not sure whether this is really useful 
   (but doesn't cost much either).
 * I thought about other structures that layers potentially storing all data, for example
   a cache per property, which is quite (bot not totally) orthogonal to the current design.
 * Restrict the interface to what is actually useful, in my opinion there are too many options
   ATM. For example, the "discard_intermediate" flag indicating whether requirements for 
   the requested data should be stored as well is probably overkill.

# Other thoughts
 * The Mesh class is quite empty. Should it really only hold geometry and topology?
   [the functions hmin() ... are implemented via free functions and thus would not have to 
   be members. This might however be comfortable in the python layer.
 * Thread safety: The current implementation is not thread safe. Somewhere in the middle
   I did some draft implementation which I removed later on, since I did not see a reasonable
   way to test it (I guess multithreading is not a use case currently). Nevertheless, the
   implementation can of course be made thread safe if required.

I believe that with these options at hand, there is additional convenience for the user
without memory overhead (since cache has to be acquired explicitly) and there is also
the chance to avoid topology_mutable and replace this by a more fine-grained "mutability".
In the end , one might not need the create_XYZ any more within the library code. It might
be called by the user who is in possession of a non-const mesh.

Nevertheless, it is currently just a proposal with a draft -- thus open for an open discussion.
I am looking forward to your feedback.

--
Btw, I am aware that the commit history is quite messy -- I iterated back and forth at some places.
Thus, if you consider my draft worth being brought to a state where it can be merged, I can
try to reorganize the changes once there is agreement on the whole thing.

Since I was not explicitly ask for doing this work, I will of course not complain if it is not accepted at some point.

EDIT: according to some discussion, I gave up on a couple of rather low-level custom memory management types.


mrambausek:
Concerning checks: I have no clue why flake complains about things which are not there any more. I ran flake as well as all other tests I could find locally to check that they pass.

jorgensd:
@mrambausek As this is now very out of date with master, do you want to continue with this draft, or should we close the PR?

mrambausek:
@jorgensd this PR is indeed outdated. Due to the many changes in underlying data structures, I will not continue on this PR. Let me know, if some of these things are still of interest.

PR #993, opened 2020-04-27
Refactor Topology::create_topology

chrisrichardson:
* uses same code for ghosted and non-ghosted meshes

### DIFF
--- a/cpp/dolfinx/mesh/Mesh.cpp
+++ b/cpp/dolfinx/mesh/Mesh.cpp
@@ -118,7 +118,8 @@ Mesh mesh::create(MPI_Comm comm,
   }
 
   // FIXME: improve. Clip cell data, removing any unused ghost cells
-  // before sending to create_geometry
+  // before sending to create_geometry. Maybe AdjacencyList can have a "resize"
+  // method?
   int n_cells_local = topology.index_map(tdim)->size_local()
                       + topology.index_map(tdim)->num_ghosts();
   Eigen::Matrix<std::int32_t, Eigen::Dynamic, 1> off1
--- a/cpp/dolfinx/mesh/Mesh.cpp
+++ b/cpp/dolfinx/mesh/Mesh.cpp
@@ -90,6 +90,9 @@ Mesh mesh::create(MPI_Comm comm,
   const auto [cell_nodes, src, original_cell_index, ghost_owners]
       = graph::Partitioning::distribute(comm, cells, dest);
 
+  // Create cells and vertices with the ghosting requested. Input topology
+  // includes cells shared via facet, but output will remove these, if not
+  // required by ghost_mode.
   Topology topology = mesh::create_topology(
       comm,
       mesh::extract_topology(element.cell_shape(), element.dof_layout(),
@@ -114,8 +117,16 @@ Mesh mesh::create(MPI_Comm comm,
     }
   }
 
+  // FIXME: improve. Clip cell data, removing any unused ghost cells
+  // before sending to create_geometry
+  int n_cells_local = topology.index_map(tdim)->size_local()
+                      + topology.index_map(tdim)->num_ghosts();
+  auto off1 = cell_nodes.offsets().head(n_cells_local + 1);
+  auto data1 = cell_nodes.array().head(off1[n_cells_local]);
+  graph::AdjacencyList<std::int64_t> cell_nodes_2(data1, off1);
+
   Geometry geometry
-      = mesh::create_geometry(comm, topology, element, cell_nodes, x);
+      = mesh::create_geometry(comm, topology, element, cell_nodes_2, x);
 
   return Mesh(comm, std::move(topology), std::move(geometry));
 }
--- a/cpp/dolfinx/mesh/Mesh.cpp
+++ b/cpp/dolfinx/mesh/Mesh.cpp
@@ -121,8 +121,10 @@ Mesh mesh::create(MPI_Comm comm,
   // before sending to create_geometry
   int n_cells_local = topology.index_map(tdim)->size_local()
                       + topology.index_map(tdim)->num_ghosts();
-  auto off1 = cell_nodes.offsets().head(n_cells_local + 1);
-  auto data1 = cell_nodes.array().head(off1[n_cells_local]);
+  Eigen::Matrix<std::int32_t, Eigen::Dynamic, 1> off1
+      = cell_nodes.offsets().head(n_cells_local + 1);
+  Eigen::Matrix<std::int64_t, Eigen::Dynamic, 1> data1
+      = cell_nodes.array().head(off1[n_cells_local]);
   graph::AdjacencyList<std::int64_t> cell_nodes_2(data1, off1);
 
   Geometry geometry
--- a/cpp/dolfinx/mesh/Mesh.cpp
+++ b/cpp/dolfinx/mesh/Mesh.cpp
@@ -79,10 +79,12 @@ Mesh mesh::create(MPI_Comm comm,
                                cells);
 
   // Compute the destination rank for cells on this process via graph
-  // partitioning
+  // partitioning. Always get the ghost cells via facet, though these may be
+  // discarded later.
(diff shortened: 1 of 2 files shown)


cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -79,15 +79,20 @@ Mesh mesh::create(MPI_Comm comm,
                                cells);
 
   // Compute the destination rank for cells on this process via graph
-  // partitioning
+  // partitioning. Always get the ghost cells via facet, though these may be
```
garth-wells, line 82: What if the user requests ghosted by vertex?

cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -112,8 +117,19 @@ Mesh mesh::create(MPI_Comm comm,
...
+  // FIXME: improve. Clip cell data, removing any unused ghost cells
+  // before sending to create_geometry. Maybe AdjacencyList can have a "resize"
+  // method?
+  int n_cells_local = topology.index_map(tdim)->size_local()
+                      + topology.index_map(tdim)->num_ghosts();
+  Eigen::Matrix<std::int32_t, Eigen::Dynamic, 1> off1
+      = cell_nodes.offsets().head(n_cells_local + 1);
+  Eigen::Matrix<std::int64_t, Eigen::Dynamic, 1> data1
+      = cell_nodes.array().head(off1[n_cells_local]);
+  graph::AdjacencyList<std::int64_t> cell_nodes_2(data1, off1);
```
garth-wells, line 129: Use 
```cpp
graph::AdjacencyList<std::int64_t> cell_nodes_2(cell_nodes.offsets().head(n_cells_local + 1), 
                                                cell_nodes.array().head(off1[n_cells_local]));
```

cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -112,8 +117,19 @@ Mesh mesh::create(MPI_Comm comm,
...
+
   Geometry geometry
-      = mesh::create_geometry(comm, topology, element, cell_nodes, x);
+      = mesh::create_geometry(comm, topology, element, cell_nodes_2, x);
```
garth-wells, line 132: Or adapt this function such that it ignores ghost cells. Or maybe ghost cells are helpful?

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -24,15 +26,19 @@ namespace
...
 {
   const int mpi_size = dolfinx::MPI::size(comm);
+  // Create a global address space to use with all_to_all post-office
+  // algorithm and find the owner of each index within that space
   std::int64_t global_space = 0;
-  const std::int64_t max_index
-      = *std::max_element(unknown_indices.begin(), unknown_indices.end());
+  std::int64_t max_index = 0;
+  if (!unknown_indices.empty())
+    max_index
```
garth-wells, line 40: Add curly braces.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -24,15 +26,19 @@ namespace
 {
 //-----------------------------------------------------------------------------
 // Given a list of indices (unknown_indices) on each process, return a
-// map to sharing processes for each index, taking the owner as the
+// map to sharing processes for each index, with the (random) owner as the
```
garth-wells, line 29: Document the in argument.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -24,15 +26,19 @@ namespace
...
 // first in the list
 std::unordered_map<std::int64_t, std::vector<int>>
 compute_index_sharing(MPI_Comm comm, std::vector<std::int64_t>& unknown_indices)
```
garth-wells, line 32: Feels like this function is in the wrong file.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -427,143 +441,151 @@ mesh::create_topology(MPI_Comm comm,
...
+    index_map_c = std::make_shared<common::IndexMap>(comm, num_local_cells,
+                                                     cell_ghost_indices, 1);
+  }
 
-  // Cell IndexMap
-  const int num_local_cells = cells.num_nodes() - cell_ghost_indices.size();
-  auto index_map_c = std::make_shared<common::IndexMap>(comm, num_local_cells,
-                                                        cell_ghost_indices, 1);
+  // Map from existing global vertex index to local index
+  // putting ghost indices last
```
garth-wells, line 463: The below map empty, so can't be as described.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -427,143 +441,151 @@ mesh::create_topology(MPI_Comm comm,
...
+      = compute_index_sharing(comm, unknown_indices);
 
-    // Make a list of all vertex indices whose ownership needs determining
-    std::vector<std::int64_t> unknown_indices(ghost_boundary_vertices.begin(),
-                                              ghost_boundary_vertices.end());
-    std::unordered_map<std::int64_t, std::vector<int>> global_to_procs
-        = compute_index_sharing(comm, unknown_indices);
+  // Number all indices which this process now owns
+  std::int32_t c = 0;
+  for (std::int64_t gi : local_vertex_set)
```
garth-wells, line 503: Use some more expressive than `gi`.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -427,143 +441,151 @@ mesh::create_topology(MPI_Comm comm,
...
-      assert(insert);
+      // Should already be in map, but needs index
+      auto it_gi = global_to_local_index.find(gi);
+      assert(it_gi != global_to_local_index.end());
+      assert(it_gi->second == -1);
+      it_gi->second = c++;
     }
-    for (std::int64_t gi : ghost_boundary_vertices)
+  }
+  std::int32_t nlocal = c;
```
garth-wells, line 525: Add comment. Can it be `const`?

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -427,143 +441,151 @@ mesh::create_topology(MPI_Comm comm,
...
+    proc_to_neighbours.insert({neighbours[i], i});
+
+  // Communicate new global index to neighbours
+  MPI_Comm neighbour_comm;
+  MPI_Dist_graph_create_adjacent(comm, neighbours.size(), neighbours.data(),
+                                 MPI_UNWEIGHTED, neighbours.size(),
+                                 neighbours.data(), MPI_UNWEIGHTED,
+                                 MPI_INFO_NULL, false, &neighbour_comm);
+
+  std::vector<std::vector<std::int64_t>> send_pairs(neighbours.size());
```
garth-wells, line 548: Would be good if we could avoid these data structures.

cpp/dolfinx/mesh/Mesh.cpp
chrisrichardson: It's currently not supported. I will add some kind of message or error.

cpp/dolfinx/mesh/Mesh.cpp
chrisrichardson: I can't get that to work - some kind of type inference problems, but a const reference seems OK.

garth-wells: ok, we can look at a solution later.

cpp/dolfinx/mesh/Topology.cpp
chrisrichardson: I decided to keep this one. Precomputing the sizes and offsets requires traversing a map, which seems possibly just as costly. The size here is small, anyway, just neighbours.

garth-wells: Ok, it's fine if it's just neighbours.

PR #1001, opened 2020-05-06
Template MeshTags IO

michalhabera:
This exposes a template version of write/read_meshtags.
I've currently only specialized it in the python layer for `int32` and `double`.


michalhabera:
Will come after #1020

PR #1011, opened 2020-05-12
Re-implementation of writing to XDMF finite element spaces

jorgensd:
Currently, our IO only supports CG-1 interpolation of finite element data to XDMF.
This draft is an outline of a re-implementation of the `write_checkpoint` function we previously had
(Following the old code quite closely).

However, some things to note are:

- The new implementation replace CG-1 interpolation of all functions, instead of creating an alternative code branch.
- The error message is suggesting users to project/interpolate the function to a suitable visualization format.

A downside of the new code is that one cannot visualize hex element in XDMF, as paraview does not support visualization of `Q` elements (it seems to be a bug in XDMF/Paraview). There is a similar issue with Polylines, see: https://discourse.paraview.org/t/unable-to-visualise-xdmf-checkpoint-in-1d/2781/14 .

At least I think it is about time we start discussing this.


francesco-ballarin:
Great, thanks. Will the new code allow to read the stored function back in, as it used to be possible with `read_checkpoint`?

jorgensd:
> Great, thanks. Will the new code allow to read the stored function back in, as it used to be possible with `read_checkpoint`?

That is something that I think we should discuss. As of now, the draft only has the write functionality (still working out some kinks there). However, adding back the read functionality would be nice.

francesco-ballarin:
> That is something that I think we should discuss. As of now, the draft only has the write functionality (still working out some kinks there). However, adding back the read functionality would be nice.

I am definitely voting +1 on having in future the read functionality, not necessarily in this PR of course. I rely on it in one of my libraries (or, to be more precise, the mathematical method itself relies on the availability of this functionality); I can provide more details in future PR if needed.

jorgensd:
Just as a side note, seems like Paraview is not able to visualize DG-3 fields on triangles, even if they are written nicely to file. I guess one of the steps in this PR has to be to make a list of what we can visualize and what we cant.

cpp/dolfinx/io/xdmf_function.cpp
```diff
@@ -82,28 +54,48 @@ void xdmf_function::add_function(MPI_Comm comm, const function::Function& u,
...
+         {"DQ", "DQ"}};
+
+  const std::map<mesh::CellType, std::string> cell_shape_repr
+      = {{dolfinx::mesh::CellType::interval, "interval"},
+         {dolfinx::mesh::CellType::triangle, "triangle"},
+         {dolfinx::mesh::CellType::tetrahedron, "tetrahedron"},
+         {dolfinx::mesh::CellType::quadrilateral, "quadrilateral"},
+         {dolfinx::mesh::CellType::hexahedron, "hexahedron"}};
+  // Check that element is supported
+  auto const it = family_abbr.find(element_family);
```
garth-wells, line 82: `const` not required here.

cpp/dolfinx/io/xdmf_function.cpp
```diff
@@ -130,38 +122,102 @@ void xdmf_function::add_function(MPI_Comm comm, const function::Function& u,
...
+        = std::to_string(element_degree).c_str();
+    attribute_node.append_attribute("ElementCell") = element_cell.c_str();
     attribute_node.append_attribute("Name") = attr_name.c_str();
+    attribute_node.append_attribute("Center") = "Other";
     attribute_node.append_attribute("AttributeType")
         = rank_to_string(u.value_rank()).c_str();
-    attribute_node.append_attribute("Center") = cell_centred ? "Cell" : "Node";
 
-    const bool use_mpi_io = (dolfinx::MPI::size(comm) > 1);
+    // Prepare and save number of dofs per cell (x_cell_dofs) and cell
```
garth-wells, line 135: This comment isn't clear to me.

cpp/dolfinx/io/xdmf_function.cpp
```diff
@@ -130,38 +122,102 @@ void xdmf_function::add_function(MPI_Comm comm, const function::Function& u,
...
+    std::vector<std::size_t> x_cell_dofs;
+    const std::size_t n_cells = mesh->topology().index_map(tdim)->size_local();
+    x_cell_dofs.reserve(n_cells);
+
+    Eigen::Array<std::int64_t, Eigen::Dynamic, 1> local_to_global_map
+        = dofmap.index_map->indices(true);
+
+    // Add number of dofs for each cell
+    // Add cell dofmap
+    for (std::size_t i = 0; i != n_cells; ++i)
```
garth-wells, line 152: `i < n_cells`

cpp/dolfinx/io/xdmf_function.cpp
```diff
@@ -82,28 +54,48 @@ void xdmf_function::add_function(MPI_Comm comm, const function::Function& u,
...
-  // Get function::Function data values and shape
-  std::vector<PetscScalar> data_values;
-  const bool cell_centred = has_cell_centred_data(u);
-  if (cell_centred)
-    data_values = xdmf_utils::get_cell_data_values(u);
-  else
-    data_values = xdmf_utils::get_point_data_values(u);
+  std::string element_family = u.function_space()->element()->family();
+  const bool use_mpi_io = (dolfinx::MPI::size(mesh->mpi_comm()) > 1);
+  const std::size_t element_degree = u.function_space()->element()->degree();
```
garth-wells, line 59: Don't use `size_t`. Use `int`, `std::int32_t` or  `std::int64_t` as appropriate.

cpp/dolfinx/io/xdmf_function.cpp
```diff
@@ -130,38 +122,102 @@ void xdmf_function::add_function(MPI_Comm comm, const function::Function& u,
...
+        auto p = cell_dofs_i[j];
+        assert(p < (std::int32_t)local_to_global_map.size());
+        cell_dofs.push_back(local_to_global_map[p]);
+      }
+    }
+
+    // Add offset to CSR index to be seamless in parallel
+    const std::size_t offset
+        = MPI::global_offset(mesh->mpi_comm(), cell_dofs.size(), true);
+    for (auto& x : x_cell_dofs)
```
garth-wells, line 167: Don't use `auto` here.

cpp/dolfinx/io/xdmf_function.cpp
```diff
@@ -130,38 +122,102 @@ void xdmf_function::add_function(MPI_Comm comm, const function::Function& u,
...
+      x += offset;
+
+    std::int64_t num_cell_dofs_global = 0;
+    size_t local_size = cell_dofs.size();
+    MPI_Allreduce(&local_size, &num_cell_dofs_global, 1, MPI_INT64_T, MPI_SUM,
```
garth-wells, line 172: The in/out int types should be the same.

jorgensd:
This PR is very out of date. We should reconsider if we want to use XDMF for higher order function space visualization or not.

PR #1012, opened 2020-05-12
Use ghosted meshes by default

### LINKED ISSUE #1003: `GhostMode.shared_facet` not covered in most tests
`GhostMode.shared_facet` should be considered for all relevant unit tests. For example [here](https://github.com/FEniCS/dolfinx/blob/master/python/test/unit/fem/test_assembler.py#L173-L210).

This test will fail if the mesh is constructed with `mesh = UnitSquareMesh(MPI.COMM_WORLD, 12, 12, ghost_mode=dolfinx.cpp.mesh.GhostMode.shared_facet)` and `mpirun -np 3 pytest test_assembler.py::test_assembly_bcs`.

garth-wells:
Switch to ghosted meshes (meshes with ghost cells) by default.
- This case is more general, works for DG in parallel.
- Improves test coverage. Codes that work and tests that pass with ghosted mesh work (almost always?) with non-ghosted meshes. The reverse is not often true.
- PR contains some fixes for ghosted meshes.
- Ghosted meshes are faster for detecting facets that lie on the exterior of the domain when the mesh is distributed.

Drawback
- Slightly more memory used to store the mesh

Fixes #1003.


jhale:
This seems like a safe default. In the past it's been too easy to use a feature that relies on shared facet ghosting and have silent errors introduced.

PR #1014, opened 2020-05-14
IndexMap: Remove _all_ranges and MPI_Allgather calls

IgorBaratta:
- Remove `_all_ranges` completely and also some unnecessary `MPI_Allgather` calls (one per IndexMap construction). For that, I included a vector with the ghost owners to the IndexMap constructor;
- The function `get_global_ghost_owners` is only used for debugging.  I can remove it;
- Added global_size as a private member variable, can also be removed and calculated on demand. But the communication cost is hidden (nonblocking communication).

Next PR asymmetric neighborhood communication and optimization. I've split it into more parts to make it easier to review.

### DIFF
--- a/cpp/dolfinx/common/IndexMap.h
+++ b/cpp/dolfinx/common/IndexMap.h
@@ -161,8 +161,7 @@ public:
   std::int64_t local_to_global(std::int32_t local_index) const
   {
     assert(local_index >= 0);
-    const std::int32_t local_size
-        = static_cast<const int32_t>(_local_range[1] - _local_range[0]);
+    const std::int32_t local_size = (_local_range[1] - _local_range[0]);
     if (local_index < local_size)
     {
       const std::int64_t global_offset = _local_range[0];
--- a/cpp/dolfinx/common/IndexMap.h
+++ b/cpp/dolfinx/common/IndexMap.h
@@ -162,10 +162,10 @@ public:
   {
     assert(local_index >= 0);
     const std::int32_t local_size
-        = _all_ranges[_myrank + 1] - _all_ranges[_myrank];
+        = static_cast<const int32_t>(_local_range[1] - _local_range[0]);
     if (local_index < local_size)
     {
-      const std::int64_t global_offset = _all_ranges[_myrank];
+      const std::int64_t global_offset = _local_range[0];
       return global_offset + local_index;
     }
     else
@@ -187,9 +187,6 @@ public:
   /// Owner rank (on global communicator) of each ghost entry
   Eigen::Array<std::int32_t, Eigen::Dynamic, 1> ghost_owners() const;
 
-  /// Get MPI rank that owns index (global block index)
-  int owner(std::int64_t global_index) const;
-
   /// Return array of global indices for all indices on this process,
   /// including ghosts
   Eigen::Array<std::int64_t, Eigen::Dynamic, 1>
@@ -199,7 +196,7 @@ public:
   /// @return The communicator on which the IndexMap is defined
   MPI_Comm mpi_comm() const;
 
-  /// Neighbors for neigborhood communicator
+  /// Neighbors for neighborhood communicator
   const std::vector<std::int32_t>& neighbours() const;
 
   /// @todo Aim to remove this function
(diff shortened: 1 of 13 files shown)


cpp/dolfinx/la/SparsityPattern.cpp
```diff
@@ -67,11 +67,19 @@ SparsityPattern::SparsityPattern(
   for (auto& ghosts : ghosts_new1)
     ghosts1.insert(ghosts1.end(), ghosts.begin(), ghosts.end());
 
+  std::vector<int> ghost_owners0;
+  std::vector<int> ghost_owners1;
```
garth-wells, line 71: Move onto one line.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -323,16 +323,17 @@ mesh::create_topology(MPI_Comm comm,
   if (ghost_mode == mesh::GhostMode::none)
   {
     index_map_c = std::make_shared<common::IndexMap>(
-        comm, num_local_cells, std::vector<std::int64_t>(), 1);
+        comm, num_local_cells, std::vector<std::int64_t>(), std::vector<int>(),
```
garth-wells, line 326: Can `std::vector<std::int64_t>(), std::vector<int>(),` be `{}, {}`?

cpp/dolfinx/mesh/TopologyComputation.cpp
```diff
@@ -331,12 +332,12 @@ get_local_indexing(
...
     }
-
-    const Eigen::Array<std::int64_t, Eigen::Dynamic, 1> recv_global_index_data
+    const graph::AdjacencyList<std::int64_t> recv_data
         = dolfinx::MPI::neighbor_all_to_all(
-              neighbour_comm, send_global_index_offsets, send_global_index_data)
-              .array();
+            neighbour_comm, send_global_index_offsets, send_global_index_data);
 
+    auto& recv_global_index_data = recv_data.array();
```
garth-wells, line 339: Use the precise type here rather than `auto` - helps readability.

cpp/dolfinx/mesh/TopologyComputation.cpp
```diff
@@ -357,7 +362,8 @@ get_local_indexing(
   MPI_Comm_free(&neighbour_comm);
 
   std::shared_ptr<common::IndexMap> index_map
-      = std::make_shared<common::IndexMap>(comm, num_local, ghost_indices, 1);
+      = std::make_shared<common::IndexMap>(comm, num_local, ghost_indices,
```
garth-wells, line 365: Using `auto` here would be fine because the type is clear.

cpp/dolfinx/refinement/ParallelRefinement.cpp
```diff
@@ -385,16 +385,17 @@ ParallelRefinement::partition(const std::vector<std::int64_t>& cell_topology,
     // set cell-vertex topology
     mesh::Topology topology_local(comm, _mesh.geometry().cmap().cell_shape());
     const int tdim = topology_local.dim();
-    auto map = std::make_shared<common::IndexMap>(
-        comm, cells_local.num_nodes(), std::vector<std::int64_t>(), 1);
+    auto map = std::make_shared<common::IndexMap>(comm, cells_local.num_nodes(),
+                                                  std::vector<std::int64_t>(),
```
garth-wells, line 389: Use `{}` for empty vectors.

cpp/dolfinx/common/IndexMap.h
```diff
@@ -161,11 +165,10 @@ class IndexMap
   std::int64_t local_to_global(std::int32_t local_index) const
```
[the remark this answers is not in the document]
IgorBaratta: Yes, I think so. It would be fairly simple to use `_local_range` and `ghosts` in-line instead.
I'll create a new PR removing it, and a few more cleanups.

cpp/dolfinx/mesh/Topology.cpp
garth-wells: Ah, that's right - `std::make_shared` doesn't like initialiser lists.

PR #1019, opened 2020-05-17
Remove deprecated C++ mesh constructor

garth-wells:


francesco-ballarin:
Dear @garth-wells , @jorgensd and @michalhabera ,
the pygmsh demo introduced here (and extended in more recent PRs) is in my opinion a great addition for our end users. I have been able to greatly simplify the mesh generation scripts associated to my demos thanks to the work you have been carrying out on this, thank you.

There have been only a couple of things that I found a bit confusing in switching from my former pygmsh+meshio generation to the current way:
1) in 2D, the user has to be mindful to pass `prune_z_0=True` at each `pygmsh.generate_mesh` call (e.g. [on line 28](https://github.com/FEniCS/dolfinx/blob/master/python/demo/gmsh/demo_gmsh.py#L28)), otherwise the mesh will be embedded in a 3D space because `x.shape[1]` [on line 32](https://github.com/FEniCS/dolfinx/blob/master/python/demo/gmsh/demo_gmsh.py#L32) will silently be 3. Since an inexperienced user might have expected it to be 2, and this might be a common mistake for end users, do we want to add a case that shows this?
2) I found the notation `pygmsh_mesh.cells[-1]` and `pygmsh_mesh.cells[-2]` (e.g. [line 29](https://github.com/FEniCS/dolfinx/blob/master/python/demo/gmsh/demo_gmsh.py#L29) and [line 58](https://github.com/FEniCS/dolfinx/blob/master/python/demo/gmsh/demo_gmsh.py#L58)) a bit confusing, in particular for facets.
At first I though that `-1` meant all entities of codimension 0, `-2` all entities of codimension 1, etc, and that seemed like a great way to avoid hardcoding `"line"`, `"triangle"`, `"tetra"`, etc. However, in my demos I realized that my assumption is actually false. `-1` will indeed be associated to entities of codimension 0, but if two distinct boundary markers have been defined in pygmsh, they will correspond to two different indices (`-2` and `-3`, respectively). This might get confusing for users, especially if they copy and paste the code of the demo without realizing that adding a new boundary id in the gmsh part will affect also the MeshTags code. 
For now, in my demos I reverted to using `pygmsh_mesh.cells_dict`, for which all entities of the same codimension are collected under the same key. For instance, if the mesh is 2D and I have two boundary tags then `pygmsh_mesh.cells_dict["line"]` contains both `pygmsh_mesh.cells[-2]` and `pygmsh_mesh.cells[-3]`. Hopefully there is a way which allows us to avoid hardcoding `"line"` but still does not require users to recall how many boundaries they marked.
This issue is hidden in the current demo by the fact that there is only one boundary marker.

Cheers,
Francesco

jorgensd:
@francesco-ballarin 
1. The demo still needs polishing. I think a finialized demo should have a 2D mesh (with prune_z_0=True), a 2D manifold and a 3D mesh, to illustrate the different syntaxes.
2.  You are right that this is not correct. You can for instance see this is in MWE on [discourse](https://fenicsproject.discourse.group/t/transitioning-from-mesh-xml-to-mesh-xdmf-from-dolfin-convert-to-meshio/412/104?u=dokken). I am not sure what is the nicest fix to this, but maybe an wrapper that takes a meshio mesh, and returns a mesh for each entity.
3. Something I mentioned to Garth and Michal this morning is that fact that gmsh has a Python API, see [this dolfiny demo](https://github.com/michalhabera/dolfiny/blob/master/demo/bingham/bingham_block.py#L32-L39). I think we should discuss what demos should be in dolfinx, and what we should have in dolfiny, and which gmsh interfaces we would like to document.

francesco-ballarin:
I agree on item 3, feel free to keep me in the loop. I am not knowledgeable enough in either gmsh or pygmsh to have an opinion, but if the gmsh python interface predates the beginning of pygmsh development, I think that we can get some hints from upstream (what were the motivations for developing pygmsh over the standard gmsh API, and see if those motivations somehow overlap with our goals, or not).

PR #1020, opened 2020-05-20
Topological create_meshtags

michalhabera:
This PR splits `create_meshtags` into more elementary steps. New `create_meshtags` now takes entities defined using local-to-process vertex numbering. It doesn't require any parallel communication.

The part with parallel communication was moved to `extract_local_entities` which takes entities defined using input global (node) indices. This method then filters out indices which doesn't belong to vertices, so parallel communication is cheaper. Communication steps are the same as in the previous implementation.

I also fixed IO for saving mesh subentities of lower topological dimension for higher-order cells.

Demo `demo_gmsh.py` is still WIP due to missing gmsh-dolfinx permutations. I handcoded there one.


cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -270,6 +287,7 @@ void XDMFFile::write_meshtags(const mesh::MeshTags<std::int32_t>& meshtags,
 //-----------------------------------------------------------------------------
 mesh::MeshTags<std::int32_t>
 XDMFFile::read_meshtags(const std::shared_ptr<const mesh::Mesh>& mesh,
+                        const fem::CoordinateElement& entity_element,
```
garth-wells, line 290: Why is a `CoordinateElement` required here?

cpp/dolfinx/io/XDMFFile.h
```diff
@@ -136,16 +145,17 @@ class XDMFFile
...
                       const std::string xpath = "/Xdmf/Domain");
 
   /// Read MeshTags
   /// @param[in] mesh The Mesh that the data is defined on
+  /// @param[in] entity_element Coordinate element for the entities of MeshTags
   /// @param[in] name
   /// @param[in] xpath XPath where MeshTags Grid is stored in file
   mesh::MeshTags<std::int32_t>
   read_meshtags(const std::shared_ptr<const mesh::Mesh>& mesh,
+                const fem::CoordinateElement& entity_element,
```
garth-wells, line 158: Same question again on the `CoordinateElement`

cpp/dolfinx/io/xdmf_mesh.cpp
```diff
@@ -277,10 +259,26 @@ xdmf_mesh::read_mesh_data(MPI_Comm comm, const hid_t h5_id,
   // Read geometry data
   const std::vector<double> geometry_data
```
garth-wells, line 260: Would be nice to change the return type at some point.

cpp/dolfinx/io/XDMFFile.cpp
michalhabera: It is needed here because it is needed below in `extract_local_entities`. The reason is to be able to tell which nodes that define the entity are the ones associated with vertices. I am throwing away non-vertex nodes at the very beginning since these are not needed and parallel communication sends fewer data.

I agree I can avoid this argument if I am able to tell which nodes on a cell sub-entity are associated with vertices. One approach I had in mind was to first ask which nodes (geometry dofs) for the whole cell are associated with vertices. Then ask which dofs live on the entity of given subdimension and then find their intersection. What do you think @garth-wells ?

garth-wells: Can you use the `CoordinateElement` that is associated with the Mesh? Otherwise I'd suggest passing just an array that picks out the vertices so we avoid mixing up geometry and topology as far as possible.

cpp/dolfinx/io/xdmf_mesh.cpp
michalhabera: I will be templating MeshTags IO in later PRs.

garth-wells: I had in mind using Eigen to get multi-dimensional array.

PR #1021, opened 2020-05-20
Fix some problems with ghosting and refinement

chrisrichardson:
Some fixes for refinement of a ghosted mesh:
* redistributes and resets the ghost mode to none or shared_facet, depending on original mesh

More work is needed for the case where the mesh is not redistributed.


garth-wells:
Will it raise an exception for a mesh with ghosting that is refined but not re-distributed?

cpp/dolfinx/refinement/ParallelRefinement.cpp
```diff
@@ -425,19 +425,25 @@ ParallelRefinement::partition(const std::vector<std::int64_t>& cell_topology,
...
-
   // Build mesh
   if (redistribute)
   {
-    return mesh::create(_mesh.mpi_comm(),
-                        graph::AdjacencyList<std::int64_t>(cells),
-                        _mesh.geometry().cmap(), _new_vertex_coordinates,
-                        mesh::GhostMode::none);
+    mesh::GhostMode ghost_mode = mesh::GhostMode::none;
+    if (num_ghost_cells > 0)
```
garth-wells, line 432: I think this step requires communication - some processes might not have ghost cells.

chrisrichardson: I don't believe this is a problem, but we could do a reduce here to be sure.
Maybe instead the user should supply a ghost mode for the refined mesh, rather than trying to infer it from the original.

chrisrichardson:
Trying to refine a mesh which has ghosting, but not redistributing will not raise an exception: it will create a refined mesh with no ghosting.

cpp/dolfinx/refinement/ParallelRefinement.cpp
garth-wells: It might be unusual, but there are cases where it could happen, e.g. two disconnected bodies. These cases can arise is contact and other problems.

garth-wells:
In the absence of something more explicit, I think a user could reasonably expect that a ghosted mesh is ghosted after refinement. If a check is added an exception could be raised with a helpful message.

PR #1022, opened 2020-05-20
Add direct interface to msh with gmsh permutations

jorgensd:
Removes some duplicate code, as well as adding all gmsh permutations (and the inverse).
Added visualization of dolfin ordering for all celltypes (excluding lines).


cpp/dolfinx/io/VTKWriter.cpp
```diff
@@ -180,8 +140,9 @@ void write_ascii_mesh(const mesh::Mesh& mesh, int cell_dim,
                       std::string filename)
 {
   const int num_cells = mesh.topology().index_map(cell_dim)->size_local();
-  const int degree = get_degree(mesh.geometry().cmap().cell_shape(),
-                                mesh.geometry().cmap().dof_layout().num_dofs());
+  const int degree
+      = io::cells::cell_degree(mesh.geometry().cmap().cell_shape(),
```
garth-wells, line 144: This function should be private, i.e. in an anonymous namespace.

python/demo/gmsh/demo_msh.py
```diff
@@ -0,0 +1,96 @@
+# Copyright (C) 2020 Garth N. Wells and Jørgen S. Dokken
```
garth-wells, line 1: This file doesn't look ready.

python/dolfinx/io.py
```diff
@@ -56,3 +60,220 @@ def read_mesh(self, ghost_mode=cpp.mesh.GhostMode.shared_facet, name="mesh", xpa
         mesh._ufl_domain = domain
 
         return mesh
+
+
+def gmsh_to_dolfin(gmsh_cell):
```
garth-wells, line 65: I think this should be in C++.

jorgensd: I don't know. Do many people interface gmsh through C++ (Seems unnecessary to place it in C++ if it is not used there)?
And then there will be a lot of duplication of the mesh ordering if we would like to expose that to the user in both python and C++

python/demo/gmsh/demo_msh.py
jorgensd: I can add some documentation for each of the mesh generation function explaining what kind of meshes they create.

cpp/dolfinx/io/VTKWriter.cpp
jorgensd: Why? This function is also used in: https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfinx/io/cells.cpp#L116

python/dolfinx/io.py
garth-wells: Why would there be duplication? We should just be handling gmsh and VTK in the same way.

jorgensd: The duplication I am thinking of is the node ordering for each dolfin element. I thought it would be best to have this in the python layer, which is then again exposed in the python docs.
We can of course expose it just through the C++ docs instead.

jorgensd:
> It would be good to split this up. First PR could be adding `gmsh_to_dolfin` in the C++ interface.

Which parts in particular would you like to have in a second PR? I dont feel this PR is very big, and most of the changes in C++ is so that we can read/write the meshes we can now create with gmsh (after the `cell_degree` function was created, most of the higher order IO broke.).

garth-wells:
It would be best to have one PR that deals the `gmsh_to_dolfin` map, mirroring `vtk_to_dolfin`, and another that deals with output. Otherwise we get bogged down with details of one.

The new demo looks under cooked.

jorgensd:
> It would be best to have one PR that deals the `gmsh_to_dolfin` map, mirroring `vtk_to_dolfin`, and another that deals with output. Otherwise we get bogged down with details of one.
> 
> The new demo looks under cooked.

I do not think we should split the pull requests. The only way of verifying the permutations is by visual inspection, and therefore the fixes to output is needed. 

I think we could merge the demo you created (demo_gmsh) with my demo (demo_msh), in some way.
The key additions in the demo I've created is that it generates meshes for all celltypes from gmsh that we support.

cpp/dolfinx/io/VTKWriter.cpp
garth-wells: Because it's not well defined, and should not be exposed. We don't care about the degree - we just need a VTK element string.

The `CoordinateMap` is what defines the cell geometry map.

python/dolfinx/io.py
garth-wells: If it's in c++ it is (1) consistent with the the VTK, and (2) is available from C++ with basically no extra work.

PR #1025, opened 2020-05-21
Add <Information> to XDMFFile

michalhabera:
This simple addition allows us to store any string into XDMFFile under <Information> tag. Supports also reading it back.

Useful to store any external information as units, simulation material parameters, etc. Also useful to store map from numeric MeshTags values to readable string information.


cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -353,5 +353,42 @@ XDMFFile::read_cell_type(const std::string grid_name, const std::string xpath)
...
+    throw std::runtime_error("XML node '" + xpath + "' not found.");
+
+  pugi::xml_node info_node = node.append_child("Information");
+  assert(info_node);
+  info_node.append_attribute("Name") = name.c_str();
+  info_node.append_child(pugi::node_pcdata).set_value(value.c_str());
+
+  // Save XML file (on process 0 only)
+  if (MPI::rank(_mpi_comm.comm()) == 0)
+    _xml_doc->save_file(_filename.c_str(), "  ");
```
garth-wells, line 371: Should we be keeping the file open until the user closes it or it goes out of scope?

michalhabera: It sounds like a good idea. I will prepare a separate PR for it.

PR #1028, opened 2020-05-24
Add new proximity/collision algorithm, clean up some Bounding Box methods

chrisrichardson:
* Adds new GJK algorithm for proximity/collision detection
* Removes old collision predicates code based on Shewchuk.
* Removes several methods that mixed `Mesh` and `BoundingBoxTree` in favour of methods that first find "candidate" bounding box leaf nodes, which then need testing for collision in a separate step.


cpp/dolfinx/geometry/BoundingBoxTree.cpp
```diff
@@ -309,29 +298,18 @@ BoundingBoxTree::BoundingBoxTree(const mesh::Mesh& mesh, int tdim) : _tdim(tdim)
...
-    global_tree.reset(new BoundingBoxTree(recv_bbox, global_leaves.begin(),
-                                          global_leaves.end()));
+    const auto send_bbox = _bbox_coordinates.bottomRows(2);
+    Eigen::Array<double, Eigen::Dynamic, 3, Eigen::RowMajor> recv_bbox(
+        mpi_size * 2, 3);
+    MPI_Allgather(send_bbox.data(), 6, MPI_DOUBLE, recv_bbox.data(), 6,
+                  MPI_DOUBLE, comm);
+
+    auto [global_bboxes, global_coords] = build_from_leaf(recv_bbox);
+    global_tree.reset(new BoundingBoxTree(global_bboxes, global_coords));
```
garth-wells, line 308: Use `_global_free = std::make_unique<BoundingBoxTree>(global_bboxes, global_coords);`

cpp/dolfinx/geometry/GJK.h
```diff
@@ -0,0 +1,24 @@
...
+{
+namespace geometry
+{
+
+/// Calculate the distance between two convex bodies bd1 and bd2
+/// @param[in] p Body 1 list of vertices
+/// @param[in] q Body 2 list of vertices
+/// @return shortest vector between bodies
+Eigen::Vector3d
+gjk_vector(const Eigen::Matrix<double, Eigen::Dynamic, 3, Eigen::RowMajor>& p,
```
garth-wells, line 21: Function name is cryptic.

cpp/dolfinx/geometry/GJK.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include <Eigen/Dense>
+
+namespace dolfinx
+{
+namespace geometry
+{
+
+/// Calculate the distance between two convex bodies bd1 and bd2
+/// @param[in] p Body 1 list of vertices
```
garth-wells, line 17: Vertices of what?

cpp/dolfinx/geometry/GJK.cpp
```diff
@@ -0,0 +1,215 @@
...
+#include <array>
+#include <iomanip>
+#include <iostream>
+#include <stdexcept>
+
+using namespace dolfinx;
+
+namespace
+{
+std::pair<Eigen::Matrix<double, Eigen::Dynamic, 3, Eigen::RowMajor>,
```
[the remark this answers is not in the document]
chrisrichardson: I know. It could alternatively return a fixed size (we know the max number of rows), and an int, indicating the number of rows, but that seemed more messy.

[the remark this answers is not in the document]
garth-wells: Eigen has a template argument for the 'max' rows/cols to handle cases like this without dynamic memory allocation.

cpp/dolfinx/geometry/GJK.cpp
```diff
@@ -0,0 +1,215 @@
...
+          vmin = v;
+          smin = snew;
+        }
+      }
+    }
+    return {smin, vmin};
+  }
+
+  assert(s.rows() == 3);
+  const Eigen::Vector3d& a = s.row(0);
```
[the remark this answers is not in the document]
chrisrichardson: Maybe just use `const auto`?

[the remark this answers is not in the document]
garth-wells: I think it's nice to be explicit with the type in case where the return type is not obvious.

cpp/dolfinx/geometry/GJK.h
chrisrichardson: Should probably be points.

cpp/dolfinx/geometry/BoundingBoxTree.cpp
chrisrichardson: Actually, that won't work, because the constructor is private...

python/test/unit/geometry/test_gjk.py
```diff
@@ -0,0 +1,174 @@
+from dolfinx.cpp.geometry import gjk_vector
```
jorgensd, line 1: Header with authors and license should be added.

cpp/dolfinx/geometry/utils.h
```diff
@@ -94,41 +60,27 @@ compute_closest_entity(const BoundingBoxTree& tree,
...
-/// this function for computing the (squared) distance to a triangle.
-double squared_distance_interval(const Eigen::Vector3d& point,
-                                 const Eigen::Vector3d& a,
-                                 const Eigen::Vector3d& b);
-
+/// From the given Mesh, select up to n cells from the list of candidates which
+/// collide with point p. n may be zero (selects all valid cells). Less than n
+/// cells may be returned.
+std::vector<int>
+select_cells_from_candidates(const dolfinx::mesh::Mesh& mesh,
```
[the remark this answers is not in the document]
jorgensd: Is `check_for_actual_collision` a better name?

cpp/dolfinx/geometry/GJK.h
jorgensd: Is `compute_distance_between_bodies` better?

garth-wells: Or maybe `compute_distance_gjk` - a bit shorter

chrisrichardson: But is it a distance? Isn't that implying a scalar result?

garth-wells: True. But what is it? Docstring says distance, return type is a vector . . . Is a vector useful if we don't know the closest point on one of the bodies?

chrisrichardson: often one body is a point, so then it is useful, I think.

PR #1029, opened 2020-05-24
Change <Information> storage

michalhabera:
It seems `<Information>` cannot have value in its body, despite XDMF wiki says so. We need to store value in an attribute `<Information Value="...">`, otherwise ParaView crashes. Credit: @anzil


jhale:
Does an explicit CDATA section work?

michalhabera:
Even if you would put
```xml
<Information Name="whatever">1</Information>
```
or 
```xml
<Information Name="whatever"><![CDATA[1]]></Information>
```
PV fails with
```
'Value' not found in itemProperties in XdmfInformation::populateItem
libc++abi.dylib: terminating with uncaught exception of type XdmfError: 'Value' not found in itemProperties in XdmfInformation::populateItem
```

It is nothing but issue in ParaView's xdmf reader which requires Value attribute be specified.

PR #1030, opened 2020-05-24
Save XML file only on XDMFFile.close()

michalhabera:
Avoids filesystem touching on every write operation.


garth-wells:
I think we need a `flush` method that writes the pugi XML file without closing the `XDMFFile` - helpful if a computation get interrupted.

michalhabera:
Ok, I'll add `XDMFFile.flush` that would flush both HDF and XML. The default behavior would then be to not flush, which is the performant way, but prone to corrupted files in case of interruption.

michalhabera:
@garth-wells Flush added

garth-wells:
I think we need to add `flush` to the Python interface too.

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -118,6 +118,8 @@ XDMFFile::~XDMFFile() { close(); }
 //-----------------------------------------------------------------------------
 void XDMFFile::close()
 {
+  flush();
```
garth-wells, line 121: I don't think flush should be called here - no need to also flush the HDF5 file.

PR #1033, opened 2020-05-25
Update README.md

garth-wells:
Update Docker badge


README.md
```diff
@@ -2,7 +2,7 @@
 
 [![CircleCI](https://circleci.com/gh/FEniCS/dolfinx.svg?style=svg)](https://circleci.com/gh/FEniCS/dolfinx)
```
IgorBaratta, line 3: Wouldn't it be better to use the shield style here? `style=shield`

Issue #1032, opened 2020-05-25
Read XDMF XML file only on one process


garth-wells:
The XDMF XML file should be read only on one process, and the reader can then send the required data over MPI. This avoids a large number of processes accessing the filesystem.

garth-wells:
Closing since we will eventually deprecate XDMF as its maintenance future isn't clear.

PR #1036, opened 2020-05-26
Write XML on XDMFFile.close, add flush

michalhabera:


michalhabera:
This PR avoids touching filesystem with pugi XML file on every `write_foo` call (beneficial in timeloops). It does so only on explicit `.close()` or `.flush()`.

Barriers are needed only is some places in tests that unrealistically read right after writing. Processes must wait for rank0 to flush the XML file to disk. Barriers could be removed after #1032 is solved.

garth-wells:
Closing since we plan to deprecate XDMF anyway.

PR #1037, opened 2020-05-26
Add eval cell test for all celltypes, in serial and parallel.

jorgensd:
With PR #1028 we can now use eval on all celltypes.
This test checks this along a line inside the mesh.
The test runs in serial and parallel using allgather to compare results over different processors.

Note that issue #1038 restricts the evaluation on non-simplices to structured meshes.

Also added in deactivated test as #1028 added a less strict tolerance for finding cells.


jorgensd:
Testing this on unstructured simplicies has revealed a bug in evaluate basis function on unstructured quads and hexes. Will be back with a MWE.

python/test/unit/function/test_function.py
```diff
@@ -361,3 +367,65 @@ def test_interpolation_function(mesh):
     uh = Function(Vh)
     uh.interpolate(u)
     assert np.allclose(uh.vector.array, 1)
+
+
+@pytest.mark.parametrize("ct, tdim", [(cpp.mesh.CellType.triangle, 2),
```
garth-wells, line 372: Use expressive names -> `ct` -> `cell_type`

python/test/unit/function/test_function.py
```diff
@@ -361,3 +367,65 @@ def test_interpolation_function(mesh):
...
+                                      (cpp.mesh.CellType.tetrahedron, 3),
+                                      (cpp.mesh.CellType.hexahedron, 3)])
+def test_eval_parallel(ct, tdim):
+    if tdim == 2:
+        mesh = UnitSquareMesh(MPI.COMM_WORLD, 7, 7, cell_type=ct)
+    else:
+        mesh = UnitCubeMesh(MPI.COMM_WORLD, 7, 7, 7, cell_type=ct)
+    V = FunctionSpace(mesh, ("CG", 1))
+
+    def func(x):
```
garth-wells, line 383: Maybe use lambda function in `interpolate`?

python/test/unit/function/test_function.py
```diff
@@ -361,3 +367,65 @@ def test_interpolation_function(mesh):
...
+    y_line = np.linspace(0, 0.9, n_points)
+    if tdim == 2:
+        z_line = np.zeros(n_points)
+    else:
+        z_line = np.linspace(0, 0.7, n_points)
+    points = np.vstack((x_line, y_line, z_line))
+
+    # Create boundingboxtree
+    tree = geometry.BoundingBoxTree(mesh, mesh.geometry.dim)
+    actual_cells = np.zeros(n_points, dtype=np.int32)
```
garth-wells, line 404: This variable name is a bit odd, and does it need to be declared here?

python/test/unit/function/test_function.py
```diff
@@ -361,3 +367,65 @@ def test_interpolation_function(mesh):
...
+
+    # Find first colliding cell
+    for i, point in enumerate(points.T):
+        actual_cell = dolfinx.geometry.compute_colliding_cells(tree, mesh,
+                                                               point, 1)
+        # If cell not on process insert -1 such that eval returns 0
+        if len(actual_cell) == 0:
+            actual_cells[i] = -1
```
garth-wells, line 412: Initialise `actual_cells` -1 to avoid this `if` case.

python/test/unit/function/test_function.py
```diff
@@ -361,3 +367,65 @@ def test_interpolation_function(mesh):
...
+        else:
+            actual_cells[i] = actual_cell[0]
+
+    # Evaluate function at points, and create the exact solution
+    u_eval = u.eval(points.T, actual_cells)
+    exact = func(points)
+
+    # Gather all cell data. As eval does not distinguish between ghost and non
```
garth-wells, line 420: Just exclude ghosts using the cell ownership range.

python/test/unit/function/test_function.py
```diff
@@ -171,9 +172,11 @@ def e3(x):
     x0 = (mesh.geometry.x[0] + mesh.geometry.x[1]) / 2.0
     tree = geometry.BoundingBoxTree(mesh, mesh.geometry.dim)
     cell_candidates = geometry.compute_collisions_point(tree, x0)
-    cell = dolfinx.cpp.geometry.select_colliding_cells(mesh, cell_candidates, x0, 1)
+    cell = dolfinx.cpp.geometry.select_colliding_cells(mesh, cell_candidates,
+                                                       x0, 1)
```
garth-wells, line 176: No need to break lines - the pytest config sets our allowable code line length.

python/test/unit/function/test_function.py
jorgensd: Since the function is reused in to obtain the exact solution, I do not think we should use a lambda function.

garth-wells:
I think eventually it would be better if this test used the reference cell coordinates rather than pulling back the physical coordinates.

jorgensd:
> I think eventually it would be better if this test used the reference cell coordinates rather than pulling back the physical coordinates.

There will be some tests put in FFCX, for single cells (also higher order elements), that will be better for such use cases. Right now they are in a dolfinx [branch](https://github.com/FEniCS/dolfinx/compare/chris/quad-hex-use-newton). However, Chris and I plan to move them to FFCX once we have finalized some debugging of interpolation on tetrahedron elements.

jorgensd:
@chrisrichardson Should we still try to merge these tests, or close the PR?

jorgensd:
Close as it is very out of date, and eval should be tested somewhere else (FFC-X)

PR #1039, opened 2020-05-27
Add mapping for gmsh cell types line, line3, line4 and vertex

anzil:
This adds missing cell type mapping for `line`, `line3` and `line4` as well as `vertex` to dictionary `_gmsh_cells`.


garth-wells:
Do we have permutation arrays for each of these cell types?

jorgensd:
> Do we have permutation arrays for each of these cell types?

All 1D elements have the same node ordering in DOLFINX, VTK and GMSH
0---------1   0----2----1       0---2---3---1

michalhabera:
@garth-wells Yes, in other words, there is the identity map for all of these already, https://github.com/FEniCS/dolfinx/blob/michal/ffcx/cpp/dolfinx/io/cells.cpp#L324

PR #1042, opened 2020-05-30
Remove MeshEntity

garth-wells:
Also removes some unused and untested code that was almost certainly broken for higher-order cells.


python/test/unit/mesh/test_mesh.py
```diff
@@ -289,32 +286,32 @@ def test_GetCoordinates():
 
 @skip_in_parallel
 def xtest_cell_inradius(c0, c1, c5):
```
michalhabera, line 288: Is this still failing, so has to be "xtest"ed?

python/test/unit/mesh/test_mesh.py
```diff
@@ -289,32 +286,32 @@ def test_GetCoordinates():
...
 
 @skip_in_parallel
 def test_cell_h(c0, c1, c5):
     for c in [c0, c1, c5]:
-        assert cpp.mesh.h(c.mesh(), [c.index()], c.dim) == pytest.approx(math.sqrt(2.0))
+        assert cpp.mesh.h(c[0], [c[2]], c[1]) == pytest.approx(math.sqrt(2.0))
 
 
 @skip_in_parallel
 def xtest_cell_radius_ratio(c0, c1, c5):
```
michalhabera, line 311: The same for "xtest" here.

python/test/unit/mesh/test_mesh.py
garth-wells: Yes, it still fails.

PR #1044, opened 2020-06-02
Get reference coordinates from basis

chrisrichardson:
* Work in progress
* Uses ffcx branch see: https://github.com/FEniCS/ffcx/pull/253
* Rewrite Newton solver from generated code in Eigen

TODO:
 - ~Simplify for affine case~
 - ~Deal with manifold meshes~

### DIFF
--- a/cpp/dolfinx/fem/CoordinateElement.cpp
+++ b/cpp/dolfinx/fem/CoordinateElement.cpp
@@ -107,11 +107,11 @@ void CoordinateElement::compute_reference_geometry(
 
     // Compute Jacobian and inverse
     _evaluate_basis_derivatives(dphi.data(), 1, 1, X0.data());
-    Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> J0(
-        _gdim, _tdim);
+    Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor, 3, 3>
+        J0(_gdim, _tdim);
     J0 = cell_geometry.matrix().transpose() * dphi;
-    Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> K0(
-        _tdim, _gdim);
+    Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor, 3, 3>
+        K0(_tdim, _gdim);
     if (_gdim == _tdim)
     {
       K0 = J0.inverse();
@@ -141,8 +141,10 @@ void CoordinateElement::compute_reference_geometry(
   else
   {
     // Newton's method for non-affine geometry
-    Eigen::VectorXd xk(x.cols());
-    Eigen::RowVectorXd Xk(_tdim);
+    Eigen::Matrix<double, Eigen::Dynamic, 1, Eigen::ColMajor, 3, 1> xk(x.cols(),
+                                                                       1);
+    Eigen::Matrix<double, Eigen::Dynamic, 1, Eigen::ColMajor, 3, 1> Xk(_tdim,
+                                                                       1);
 
     for (int ip = 0; ip < num_points; ++ip)
     {
--- a/cpp/dolfinx/fem/CoordinateElement.cpp
+++ b/cpp/dolfinx/fem/CoordinateElement.cpp
@@ -98,9 +98,9 @@ void CoordinateElement::compute_reference_geometry(
 
   if (_is_affine)
   {
-    Eigen::RowVectorXd X0(_tdim);
+    Eigen::Matrix<double, Eigen::Dynamic, 1, Eigen::ColMajor, 3, 1> x0(_gdim);
+    Eigen::Matrix<double, Eigen::Dynamic, 1, Eigen::ColMajor, 3, 1> X0(_tdim);
     X0.setZero();
-    Eigen::VectorXd x0(x.cols());
     // Compute physical coordinates at X=0.
     _evaluate_basis_derivatives(phi.data(), 0, 1, X0.data());
     x0 = cell_geometry.matrix().transpose() * phi;
--- a/cpp/dolfinx/fem/CoordinateElement.cpp
+++ b/cpp/dolfinx/fem/CoordinateElement.cpp
@@ -96,6 +96,12 @@ void CoordinateElement::compute_reference_geometry(
   assert(K.dimension(1) == this->topological_dimension());
   assert(K.dimension(2) == this->geometric_dimension());
 
+  if (_gdim != _tdim)
+  {
+    throw std::runtime_error(
+        "Compute_reference_geometry not yet implemented for tdim != gdim");
+  }
+
   if (_is_affine)
   {
     const int d = cell_geometry.rows();
@@ -125,24 +131,25 @@ void CoordinateElement::compute_reference_geometry(
     Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> K0(
         _tdim, _gdim);
     K0 = J0.inverse();
+
+    // Fill result for J, K and detJ
+    Eigen::Map<
+        Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>
+        Jview(J.data(), _gdim * num_points, _tdim);
+    Jview = J0.replicate(num_points, 1);
+    Eigen::Map<
+        Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>
+        Kview(K.data(), _tdim * num_points, _gdim);
+    Kview = K0.replicate(num_points, 1);
     detJ.fill(J0.determinant());
 
+    // Calculate X for each point
     for (int ip = 0; ip < num_points; ++ip)
-    {
-      Eigen::Map<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic,
-                               Eigen::RowMajor>>
-          Jview(J.data() + ip * _gdim * _tdim, _gdim, _tdim);
-      Eigen::Map<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic,
-                               Eigen::RowMajor>>
-          Kview(K.data() + ip * _gdim * _tdim, _tdim, _gdim);
-      Jview = J0;
-      Kview = K0;
       X.row(ip) = K0 * (x.row(ip).matrix().transpose() - x0);
-    }
   }
   else
   {
-    // Newton's method
+    // Newton's method for non-affine geometry
     Eigen::VectorXd xk(x.cols());
     const int d = cell_geometry.rows();
 
--- a/cpp/dolfinx/fem/CoordinateElement.cpp
+++ b/cpp/dolfinx/fem/CoordinateElement.cpp
@@ -5,6 +5,7 @@
 // SPDX-License-Identifier:    LGPL-3.0-or-later
 
 #include "CoordinateElement.h"
+#include "FiniteElement.h"
 #include <iostream>
 #include <unsupported/Eigen/CXX11/Tensor>
 
@@ -63,14 +64,16 @@ void CoordinateElement::push_forward(
   assert(x.rows() == X.rows());
   assert(x.cols() == _gdim);
   assert(X.cols() == _tdim);
-  //  _compute_physical_coordinates(x.data(), X.rows(), X.data(),
-  //                                cell_geometry.data());
 
   // Compute physical coordinates
-  Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> phi(
-      X.rows(), cell_geometry.rows());
-  _evaluate_reference_basis(phi.data(), X.rows(), X.data());
-  x = phi * cell_geometry.matrix();
+  Eigen::Tensor<double, 3, Eigen::RowMajor> phi(1, X.rows(),
+                                                cell_geometry.rows());
+  Eigen::Map<
+      Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>
+      phi_mat(phi.data(), X.rows(), cell_geometry.rows());
+
+  _finite_element->evaluate_reference_basis(phi, X);
+  x = phi_mat * cell_geometry.matrix();
 }
 //-----------------------------------------------------------------------------
 void CoordinateElement::compute_reference_geometry(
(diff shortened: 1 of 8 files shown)


chrisrichardson:
I have converted this from a Draft PR to "ready to review" to get some feedback. I think this is a useful chunk of work (whilst there is more to be done), and I would like to merge before going further.

The question at this point, is how far to go. I have given `CoordinateElement` a `_finite_element` but clearly there is a lot of overlap. If we are not sure this is the right way, I can just make pointers to `evaluate_basis_(derivatives)` instead, which is what I had done on a previous version.

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -5,6 +5,8 @@
 // SPDX-License-Identifier:    LGPL-3.0-or-later
 
 #include "CoordinateElement.h"
+#include "FiniteElement.h"
+#include <iostream>
```
jorgensd, line 9: I guess this is a leftover include from debugging?

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -86,8 +95,110 @@ void CoordinateElement::compute_reference_geometry(
...
+        // Increment to new point in reference
+        Eigen::VectorXd dX = Kview * (x.row(ip).matrix().transpose() - xk);
+        if (dX.squaredNorm() < 1e-12)
+          break;
+        Xk += dX;
+      }
+      if (k == max_its)
+      {
+        throw std::runtime_error(
+            "Iterations exceeded in Newton iteration for cmap");
```
michalhabera, line 194: "cmap" is a bit cryptic here. Maybe something along "Newton method didn't converge for non-affine geometry"?

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -49,12 +47,20 @@ void CoordinateElement::push_forward(
...
-                                cell_geometry.data());
+
+  // Compute physical coordinates
+  Eigen::Tensor<double, 3, Eigen::RowMajor> phi(1, X.rows(),
+                                                cell_geometry.rows());
+  Eigen::Map<
+      Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>
+      phi_mat(phi.data(), X.rows(), cell_geometry.rows());
+
+  _finite_element->evaluate_reference_basis(phi, X);
```
jorgensd, line 62: Is it more natural to do this operation before declaring phi_mat, or am I missing something fundamental?

cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -29,19 +32,14 @@ class CoordinateElement
...
-  CoordinateElement(
-      mesh::CellType cell_type, int topological_dimension,
-      int geometric_dimension, const std::string& signature,
-      const ElementDofLayout& dof_layout,
-      std::function<void(double*, int, const double*, const double*)>
-          compute_physical_coordinates,
-      std::function<void(double*, double*, double*, double*, int, const double*,
-                         const double*)>
-          compute_reference_geometry);
+  /// @param[in] reference_midpoint
```
garth-wells, line 35: Expand all the docstrings.

cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -123,6 +123,22 @@ void FiniteElement::evaluate_reference_basis(
   }
 }
 //-----------------------------------------------------------------------------
+void FiniteElement::evaluate_reference_basis_derivatives(
```
garth-wells, line 126: Could be renamed to `evaluate_basis_derivatives`.

cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -123,6 +123,22 @@ void FiniteElement::evaluate_reference_basis(
...
+  int ret = _evaluate_reference_basis_derivatives(reference_values.data(),
+                                                  order, num_points, X.data());
+  if (ret == -1)
+  {
+    throw std::runtime_error("Generated code returned error "
+                             "in evaluate_reference_basis_derivatives");
+  }
+}
+//-----------------------------------------------------------------------------
 void FiniteElement::transform_reference_basis(
```
garth-wells, line 142: This is a bit muddled now - really belongs to the geometry map.

cpp/dolfinx/fem/CoordinateElement.cpp
chrisrichardson: Yes, I will remove

cpp/dolfinx/fem/CoordinateElement.cpp
chrisrichardson: Sure no problem to change

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -86,8 +95,110 @@ void CoordinateElement::compute_reference_geometry(
...
+  Eigen::Map<
+      Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>
+      dphi(dphi_tensor.data(), d, _tdim);
+
+  if (_is_affine)
+  {
+    Eigen::RowVectorXd X0(_tdim);
+    X0.setZero();
+    Eigen::VectorXd x0(x.cols());
+    // Compute physical coordinates at X=0.
```
jorgensd, line 114: Are all reference elements parameterized for the interval [-1,1]^d in FIAT? If so, we could add a note that this would be the midpoint of the element, (and maybe add the note from: https://github.com/FEniCS/ffcx/blob/master/ffcx/codegeneration/coordinate_mapping.py#L469)

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -49,12 +47,20 @@ void CoordinateElement::push_forward(
...
   assert(x.rows() == X.rows());
   assert(x.cols() == _gdim);
   assert(X.cols() == _tdim);
-  _compute_physical_coordinates(x.data(), X.rows(), X.data(),
-                                cell_geometry.data());
+
+  // Compute physical coordinates
+  Eigen::Tensor<double, 3, Eigen::RowMajor> phi(1, X.rows(),
+                                                cell_geometry.rows());
+  Eigen::Map<
```
garth-wells, line 58: Is there a neater  way to do this with Eigen?

cpp/dolfinx/fem/CoordinateElement.cpp
chrisrichardson: I like to keep the declaration and map next to each other. To be honest the `Eigen::Tensor` is a bit of a pain when extracting sub-matrices.

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -86,8 +95,110 @@ void CoordinateElement::compute_reference_geometry(
...
+
+    for (int ip = 0; ip < num_points; ++ip)
+    {
+      Eigen::Map<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic,
+                               Eigen::RowMajor>>
+          Jview(J.data() + ip * _gdim * _tdim, _gdim, _tdim);
+      Eigen::Map<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic,
+                               Eigen::RowMajor>>
+          Kview(K.data() + ip * _gdim * _tdim, _tdim, _gdim);
+      Eigen::RowVectorXd Xk(_tdim);
```
garth-wells, line 164: Avoid the dynamic memory allocation. Could use max size template argument,

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -86,8 +95,110 @@ void CoordinateElement::compute_reference_geometry(
...
+      // Xk = _reference_midpoint.head(_tdim);
```
jorgensd, line 165: I guess this is dependent of if  all reference elements parameterized for the interval [-1,1]^d in FIAT? If so, we could add a note that this would be the midpoint of the element, (and maybe add the note from: https://github.com/FEniCS/ffcx/blob/master/ffcx/codegeneration/coordinate_mapping.py#L469)

chrisrichardson:
I guess I am looking for a big-picture review rather than details at this point. Any thoughts about the attributes of `CoordinateMap` ? should it hold a `_finite_element`? If so, what else should be deleted - e.g. tdim. etc.

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -86,8 +95,110 @@ void CoordinateElement::compute_reference_geometry(
...
+        xk = cell_geometry.matrix().transpose() * phi;
+
+        // Compute Jacobian and inverse
+        _finite_element->evaluate_reference_basis_derivatives(dphi_tensor, 1,
+                                                              Xk);
+        Jview = cell_geometry.matrix().transpose() * dphi;
+        if (_gdim == _tdim)
+          Kview = Jview.inverse();
+        else
+          Kview = (Jview.transpose() * Jview).inverse() * Jview.transpose();
```
jorgensd, line 183: Maybe add that this is the Penrose-Moore pseudo-inverse as a comment (as you did above).

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -86,8 +95,110 @@ void CoordinateElement::compute_reference_geometry(
...
+
+        // Increment to new point in reference
+        Eigen::VectorXd dX = Kview * (x.row(ip).matrix().transpose() - xk);
+        if (dX.squaredNorm() < 1e-12)
```
jorgensd, line 187: In theory, we could make this tolerance an kwarg now?

cpp/dolfinx/fem/CoordinateElement.cpp
chrisrichardson: The midpoints for `X` are defined in `ufc_geometry.h` for each cell type. I'm not sure how important this really is. I am inclined to remove the midpoint stuff for now, and leave it as a TODO item.

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -76,6 +76,14 @@ class FiniteElement
       const Eigen::Ref<const Eigen::Array<
           double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>>& X) const;
 
+  /// Evaluate all basis function derivatives of order order at given point in
```
jorgensd, line 79: of order order sounds a bit strange, how about "Evaluate all basis function derivatives of given order at given point in the reference cell"

cpp/dolfinx/fem/CoordinateElement.cpp
chrisrichardson: Yes, I thought about this. I think it might be better to template on tdim and gdim to get the best performance out of Eigen, since it will then use fixed size (stack-based) arrays. TODO.

cpp/dolfinx/fem/FiniteElement.h
chrisrichardson: :+1:

cpp/dolfinx/fem/CoordinateElement.cpp
chrisrichardson: Yes, I suppose so.

cpp/dolfinx/fem/CoordinateElement.h
chrisrichardson: :+1:

cpp/dolfinx/fem/FiniteElement.cpp
chrisrichardson: And I guess the same goes for `evaluate_reference_basis`? I was wondering whether we need both interfaces, since `evaluate_reference_basis` is basically `evaluate_reference_basis_derivatives` of order=0.

cpp/dolfinx/fem/CoordinateElement.cpp
chrisrichardson: Maybe. There is `.slice` and `.chip` - I will try out.

cpp/dolfinx/fem/CoordinateElement.cpp
chrisrichardson: Anyway, I can move this one out of the loop at least

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -86,8 +88,106 @@ void CoordinateElement::compute_reference_geometry(
...
+    {
+      Eigen::Map<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic,
+                               Eigen::RowMajor>>
+          Jview(J.data() + ip * _gdim * _tdim, _gdim, _tdim);
+      Eigen::Map<Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic,
+                               Eigen::RowMajor>>
+          Kview(K.data() + ip * _gdim * _tdim, _tdim, _gdim);
+      // TODO: Xk - use cell midpoint instead?
+      Xk.setZero();
+      const int max_its = 10;
```
jorgensd, line 156: I think this could be a kwarg now, as we are throwing an error if we hit max iteration count. Then we could adjust the error message to:  "Newton method failed to converge for non-affine geometry. Try increasing the maximum number of iterations".

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -86,8 +88,106 @@ void CoordinateElement::compute_reference_geometry(
...
+        if (_gdim == _tdim)
+          Kview = Jview.inverse();
+        else
+          // Penrose-Moore pseudo-inverse
+          Kview = (Jview.transpose() * Jview).inverse() * Jview.transpose();
+
+        // Increment to new point in reference
+        Eigen::Matrix<double, Eigen::Dynamic, 1, Eigen::ColMajor, 3, 1> dX
+            = Kview * (x.row(ip).matrix().transpose() - xk);
+        if (dX.squaredNorm() < 1e-12)
```
jorgensd, line 176: Similar comment here. As we have chosen a given tolerance, should we give the user the ability to control it?

cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -24,24 +24,18 @@ class CoordinateElement
...
-      mesh::CellType cell_type, int topological_dimension,
-      int geometric_dimension, const std::string& signature,
-      const ElementDofLayout& dof_layout,
-      std::function<void(double*, int, const double*, const double*)>
-          compute_physical_coordinates,
-      std::function<void(double*, double*, double*, double*, int, const double*,
-                         const double*)>
-          compute_reference_geometry);
+  /// @param[in] is_affine Boolean flag indicating affine mapping
+  /// @param[in] evaluate_basis_derivatives
```
jorgensd, line 33: Document `evaluate_basis_derivative`.

Issue #970, opened 2020-04-13
clean up docker repositories


francesco-ballarin:
We currently have several docker repositories around:
* github actions pushes to https://hub.docker.com/u/fenicsproject
* circle CI pushes to https://hub.docker.com/u/dolfinx
* former tests in https://github.com/FEniCS/dolfinx/packages still shown in the packages sections on github.

This is fine for now I guess, but in future we might want to clean this up a bit (ideally only one source would suffice), in order not to confuse end users. I am available to help in this task, if granted 
 appropriate access.

jhale:
The current plan is this:

-  We keep 'non-user' images at https://hub.docker.com/u/fenicsproject. We've identified the need for some images, e.g. CI testing images that users don't need to see (confusing) but are necessary for developers.
- The 'user facing' images will be kept at https://hub.docker.com/u/dolfinx. Some of the images we have now will be combined (hopefully, real and complex!), and some will be removed (notebook).
- Unfortunately, we cannot delete the unwanted github packages (has already been raised with GitHub support). A workaround is to take the repository private, then delete the packages. This will be done in a quieter period when it affects less people as it has consequences (all followers will be removed, for example).

garth-wells:
Looks like https://github.com/FEniCS/dolfinx/packages is cleaned up. @jhale is there anything left to do or can this issue be closed?

jhale:
This is quite nicely implemented now, I think.

jhale:
Yep, I completely forgot, will do ;) Thanks for the reminder!

PR #1047, opened 2020-06-09
Simplify Docker image hierarchy

jhale:
I propose three concrete images in this new `Dockerfile`: `dolfinx/dev-env`, `dolfinx/dolfinx` and `dolfinx/lab`.

Still TODO:
1. Shell script to switch between dolfinx in real and complex modes.
2. Update CircleCI (and/or GH actions) nightly builds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -233,27 +233,11 @@ jobs:
           name: Login to Dockerhub
           command: echo ${DH_TOKEN} | sudo docker login -u ${DH_USERNAME} --password-stdin
       - build-and-push-target:
-          docker-target: "base"
+          docker-target: "dev-env"
       - build-and-push-target:
-          docker-target: "dev-env-real"
-      - build-and-push-target:
-          docker-target: "dev-env-complex"
-      - build-and-push-target:
-          docker-target: "real-onbuild"
-      - build-and-push-target:
-          docker-target: "complex-onbuild"
-      - build-and-push-target:
-          docker-target: "real"
-      - build-and-push-target:
-          docker-target: "complex"
-      - build-and-push-target:
-          docker-target: "notebook"
-      - build-and-push-target:
-          docker-target: "notebook-complex"
+          docker-target: "dolfinx"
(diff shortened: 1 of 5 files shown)


garth-wells:
Do we need `dolfinx/dev-env`? Doesn't look much different to `https://hub.docker.com/r/fenicsproject/test-env`. Also a bit weird having `dolfinx/foo` that doesn't have dolfin.

jhale:
Don't really agree. The Dockerfiles for test-env and dev-env are similar, so they could be merged. But the concrete versions of those images (particularly, the PETSc ARCH and the compiler options) are quite different. 

Two options:

1. Merge the Dockerfiles and use arguments at build time to produce the two concrete images (one for testing dolfinx, one for nightly builds aimed at end users, or at people wishing to build a performant version).

2. Build a full selection of PETSc/SLEPc ARCHs in one 'does it all' image.

garth-wells:
Let me expand my point - why do we need `dolfinx/dev-env` when we have `dolfinx/dolfinx`? The latter contains the former.

jhale:
Ah, I see. Because it's clean. The use case is I want to bring in my own branches of FEniCS Project components and build them against a performant and well setup set of dependencies.

garth-wells:
I would suggest installing the 'real' version into `/usr/local/dolfinx-real` to avoid files in the system location possibly being picked up when in complex mode.

On the point of a 'clean' image, if the DOLFIN builds are out of the system paths is the issue then the install of the FEniCS Python modules in a system location?

jhale:
Can do that. I will make the default `ENV` point at the real version which is likely to be by far the most commonly used, and the script will then override to complex mode.

Yes, the issue is it is 'non-trivial' to both install and run DOLFINX and it's Python components outside of the standard paths. You wouldn't want to have to remember the technicalities if you just wanted to bring up a reasonable environment and install some custom branches. It is nice just to be able to install into the default paths.

garth-wells:
Related issue: #970

PR #1048, opened 2020-06-10
Pass form compiler and JIT parameters

michalhabera:
Allows user to pass any valid FFCx and JIT parameters.

### DIFF
--- a/python/dolfinx/fem/form.py
+++ b/python/dolfinx/fem/form.py
@@ -31,8 +31,6 @@ class Form:
         and attaching coefficients and domains specific data to the underlying
         C++ Form.
         """
-        self.form_compiler_parameters = form_compiler_parameters
-        self.jit_parameters = jit_parameters
 
         # Extract subdomain data from UFL form
         sd = form.subdomain_data()
@@ -43,8 +41,8 @@ class Form:
         # Compile UFL form with JIT
         ufc_form = jit.ffcx_jit(
             form,
-            form_compiler_parameters=self.form_compiler_parameters,
-            jit_parameters=self.jit_parameters,
+            form_compiler_parameters=form_compiler_parameters,
+            jit_parameters=jit_parameters,
             mpi_comm=mesh.mpi_comm())
 
         # For every argument in form extract its function space
--- a/python/dolfinx/fem/form.py
+++ b/python/dolfinx/fem/form.py
@@ -12,7 +12,7 @@ from dolfinx import cpp, jit
 
 
 class Form:
-    def __init__(self, form: ufl.Form, form_compiler_parameters: dict = None):
+    def __init__(self, form: ufl.Form, form_compiler_parameters: dict = {}, jit_parameters: dict = {}):
         """Create dolfinx Form
 
         Parameters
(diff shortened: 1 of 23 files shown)


garth-wells:
What happens if I explicitly pre-compile a form, and subsequently call `assemble` (maybe within a loop) without passing JIT options?

michalhabera:
Then I guess you are passing `dolfinx.fem.Form` to assemble directly and no parameters are taken into account.

python/dolfinx/fem/form.py
```diff
@@ -29,6 +32,7 @@ def __init__(self, form: ufl.Form, form_compiler_parameters: dict = None):
         C++ Form.
         """
         self.form_compiler_parameters = form_compiler_parameters
+        self.jit_parameters = jit_parameters
```
garth-wells, line 35: Do we need to store the form compiler and JIT options?

python/dolfinx/function.py
```diff
@@ -243,7 +243,7 @@ def __init__(self,
 
         # Compile dofmap and element and create DOLFIN objects
         ufc_element, ufc_dofmap_ptr = jit.ffcx_jit(
-            self.ufl_element(), form_compiler_parameters=None, mpi_comm=mesh.mpi_comm())
+            self.ufl_element(), mpi_comm=mesh.mpi_comm())
```
garth-wells, line 246: Should it be possible to pass compile options here?

python/dolfinx/jit.py
```diff
@@ -90,35 +90,40 @@ def mpi_jit(*args, **kwargs):
...
+    p.update(form_compiler_parameters)
 
     # CFFI compiler options/flags
-    extra_compile_args = ['-g0', '-O3', '-march=native']
-    user_cflags = os.getenv('DOLFINX_JIT_CFLAGS')
-    if user_cflags is not None:
-        extra_compile_args = user_cflags.split(" ")
-    cffi_options = dict(cffi_extra_compile_args=extra_compile_args, cffi_verbose=False,
-                        cffi_debug=False)
+    jit_params = {"cffi_extra_compile_args": ["-O3", "-g0", "-march=native"],
```
garth-wells, line 100: I would prefer less aggressive defaults  - aggressive optimisation can add a lot to the compile time.

python/dolfinx/jit.py
```diff
@@ -90,35 +90,40 @@ def mpi_jit(*args, **kwargs):
...
+                  "cffi_debug": False, "cffi_verbose": False,
+                  "cffi_libraries": None, "cache_dir": "~/.cache/fenics", "timeout": 10}
```
garth-wells, line 102: I know that was thee before, but we should really find a way to get the platform-dependent cache location.

python/dolfinx/jit.py
chrisrichardson: I thought we agreed to get rid of `-march=native`

michalhabera: I just copied existing default args in this PR. Didn't want to touch defaults, but I can do more at once...

python/dolfinx/jit.py
chrisrichardson: I'm not sure there is a way to get a platform independent cache directory. However, we could at least use `pathlib.Path.home()` to get the home directory, and then add `.cache/fenics` to that. It will then work on Windows :smiley:

PR #1059, opened 2020-06-26
Remove extra loop brackets

mscroggs:
Due to brackets in the wrong place, this block is being run for each cell rather than just once.


chrisrichardson:
Normally our style would be to have brackets around the j-loop.

PR #1062, opened 2020-07-02
Add Vector object that is templated over the type

garth-wells:
This is a start towards supporting multiple types.

It is very likely to evolve further, but merging now will allow other simplifications and it doesn't affect the user interface.


cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -203,9 +204,8 @@ class DirichletBC
   void dof_values(Eigen::Ref<Eigen::Matrix<T, Eigen::Dynamic, 1>> values) const
   {
     assert(_g);
-    la::VecReadWrapper g(_g->vector().vec());
     for (Eigen::Index i = 0; i < _dofs.rows(); ++i)
-      values[_dofs(i, 0)] = g.x[_dofs(i, 1)];
+      values[_dofs(i, 0)] = _g->x()->array()[_dofs(i, 1)];
```
chrisrichardson, line 208: Maybe these should be pulled out as const references before the loop?

garth-wells: Yes - change was made quickly.

cpp/dolfinx/function/Function.h
```diff
@@ -74,13 +73,15 @@ class Function
   /// @return The function space
   std::shared_ptr<const FunctionSpace> function_space() const;
 
-  /// Return vector of expansion coefficients (non-const version)
+  /// Return vector of expansion coefficients at as PETSc Vec
```
chrisrichardson, line 76: typo here

cpp/dolfinx/function/Function.cpp
```diff
@@ -101,49 +99,57 @@ Function Function::collapse() const
...
+  // Check that data type is the same as the PETSc build
+  // if constexpr (std::is_same<T, PetscScalar>::value)
+  // {
+  if (!_vector)
+  {
+    _vector = create_ghosted_vector(*_function_space->dofmap()->index_map,
+                                    _x->array());
+  }
   return _vector;
+  // }
```
michalhabera, line 145: Leftover commented code.

cpp/dolfinx/function/Function.h
```diff
@@ -147,7 +148,9 @@ class Function
   std::shared_ptr<const FunctionSpace> _function_space;
 
   // The vector of expansion coefficients (local)
-  la::PETScVector _vector;
+  std::shared_ptr<la::Vector<PetscScalar>> _x;
+
+  // The vector of expansion coefficients
+  mutable Vec _vector = nullptr;
```
michalhabera, line 154: Vector here being stored twice is weird. Is `_vector` only for caching? Is this duplicating the storage?

python/dolfinx/wrappers/la.cpp
```diff
@@ -81,6 +82,12 @@ void la(py::module& m)
         return self[i]->vec();
       });
 
+  // dolfinx::la::VectorSpaceBasis
```
michalhabera, line 85: VectorSpaceBasis?

cpp/dolfinx/function/Function.cpp
garth-wells: It's deliberate - will be activated immediately after OR.

cpp/dolfinx/function/Function.h
garth-wells: `Vec` is a PETSc vector that wraps the data in `_x`. There is no duplication of storage. `_vector` will probably move at some point once the design becomes clearer, but for now this avoids user-level changes.

PR #1065, opened 2020-07-06
Change int32 to intc in test that calls UFC interface.

jhale:
C type `int` is not guaranteed to be an `np.int32`, better to use `np.intc`.

Reflects change made in FFCX unit tests (https://github.com/FEniCS/ffcx/pull/271).

### DIFF
--- a/python/test/unit/fem/test_custom_assembler.py
+++ b/python/test/unit/fem/test_custom_assembler.py
@@ -186,7 +186,7 @@ def assemble_vector(b, mesh, dofmap):
 def assemble_vector_ufc(b, kernel, mesh, dofmap):
     """Assemble provided FFCX/UFC kernel over a mesh into the array b"""
     pos, x_dofmap, x = mesh
-    entity_local_index = np.array([0], dtype=np.int32)
+    entity_local_index = np.array([0], dtype=np.intc)
     perm = np.array([0], dtype=np.uint8)
     geometry = np.zeros((3, 2))
     coeffs = np.zeros(1, dtype=PETSc.ScalarType)


garth-wells:
Maybe we should change the interface to use fixed-width integers?

jhale:
Well, a `uint8` would certainly be enough and this has a direct correspondence `np.uint8`.

I had a look through StackOverflow and found two schools of thought:

1) Those who say that overusing fixed-width types can be detrimental for performance, i.e. this is a case of trying to be 'cleverer than the compiler'. `int` is at least 16-bits but the compiler can choose a bigger underlying type if it thinks it is a good idea. Furthermore, integer promotion adds extra casting operations (https://www.geeksforgeeks.org/integer-promotions-in-c/) when performing operations on smaller types.
2) Those that say that using fixed widths is better for both portability and specifying the intent of the programmer.

garth-wells:
I think `int` is basically legacy. On (1.), DOLFIN would seriously break if `int` was 16 bit. I'm not aware of any 'normal' system that uses 16-bit `int` and I don't know of any compilers that mess with the size of `int`. I don't see how integer promotion can work for arrays - it would break interfaces with external libraries.

Fixed width ints (introduced in C99) are nice for portability and expressing intent. We use the latter extensively to distinguish between local (`int32_t`) and global (`int64_t`) indices.

michalhabera:
I also support more use of fixed-width integers, especially in places where user is expected to interface the code. Modern compilers are anyway doing plenty of padding with zeros if the variable needs to fit into bigger register for vectorised instr.

There are few more places where fixed width should be used more, e.g. `MeshTags<int>` should really be `MeshTags<int32>`, ... Another confusion which fixed-width "solves" is the fact that `numpy.int` is nowadays 64-bit integer.

jhale:
There are embedded systems that use 16-bit for `int` but I agree these are not 'normal' systems for DOLFINX users. Nevertheless, what I said is correct, `int` can only be assumed to be 16-bit. We could change all instances of `int` into a fixed width in DOLFIN but it's a lot of work.

A slight aside, an interesting class of type that is currently unused on DOLFINX is the `uintX_fast_t` which specifies a minimum size but optionally, the compiler can optionally choose a larger integer size if it wishes to do so. It doesn't get exported to numpy dtypes but in the C++ layer it could be an interesting alternative.

@michalhabera second point is actually the most pertinent one, it is far too easy to see a C `int` in an interface and then use `numpy.int`, leading to disastrous consequences! It would be hard to get this wrong with explicit fixed-width types.

**So, in summary, I would propose changing instances of `int` in the UFC headers to either fixed 16-bit or 32-bit signed/unsigned integers.**

jhale:
I merged this because it's still the correct fix, happy to continue the discussion on changing UFC to fixed-width here.

PR #1079, opened 2020-07-12
Remove macro and replace with template

chrisrichardson:


python/dolfinx/wrappers/mesh.cpp
```diff
@@ -31,6 +31,48 @@ namespace py = pybind11;
...
+      .def_readwrite("name", &dolfinx::mesh::MeshTags<T>::name)
+      .def_property_readonly("dim", &dolfinx::mesh::MeshTags<T>::dim)
+      .def_property_readonly("mesh", &dolfinx::mesh::MeshTags<T>::mesh)
+      .def("ufl_id", &dolfinx::mesh::MeshTags<T>::id)
+      .def_property_readonly("values",
+                             [](dolfinx::mesh::MeshTags<T>& self) {
+                               return py::array_t<T>(self.values().size(),
+                                                     self.values().data(),
+                                                     py::none());
+                             })
```
garth-wells, line 59: This need an extra argument to tie the lifetime of the `MeshTags` object to the return array.

chrisrichardson: OK, I just copied the existing code. Will add reference_internal.

PR #1089, opened 2020-07-19
Add Numba parallel (threaded) assembly example

garth-wells:
- Add 'transpose' dofmap (maps global index to local cell contributions). Used the assemble/accumulate unassembled vectors in a thread-safe fashion.
- Add test that uses `numba.prange` to parallelise an assembly loop
- Some performance optimisations in the assemblers

### DIFF
--- a/cpp/dolfinx/fem/DofMap.cpp
+++ b/cpp/dolfinx/fem/DofMap.cpp
@@ -195,7 +195,7 @@ fem::transpose_dofmap(graph::AdjacencyList<std::int32_t>& dofmap)
       data[pos[dofs[i]]++] = cell_offsets[c] + i;
   }
 
-  Sort the source indices for each global index
+  // Sort the source indices for each global index
   for (int index = 0; index < max_index; ++index)
   {
     std::sort(data.begin() + index_offsets[index],
--- a/cpp/dolfinx/fem/DofMap.cpp
+++ b/cpp/dolfinx/fem/DofMap.cpp
@@ -165,7 +165,6 @@ fem::transpose_dofmap(graph::AdjacencyList<std::int32_t>& dofmap)
   const std::int32_t max_index = dofmap.array().maxCoeff();
   std::vector<int> dofs_per_cell(dofmap.num_nodes());
   std::vector<int> num_local_contributions(max_index + 1);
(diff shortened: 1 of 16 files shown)


cpp/dolfinx/fem/DofMap.cpp
```diff
@@ -157,6 +157,54 @@ fem::DofMap build_collapsed_dofmap(MPI_Comm comm, const DofMap& dofmap_view,
...
+
+  // Compute offset for each global index
+  std::vector<int> index_offsets(num_local_contributions.size() + 1, 0);
+  std::partial_sum(num_local_contributions.begin(),
+                   num_local_contributions.end(), index_offsets.begin() + 1);
+
+  // Compute offset for each cell block
+  std::vector<int> cell_offsets(dofs_per_cell.size() + 1, 0);
+  std::partial_sum(dofs_per_cell.begin(), dofs_per_cell.end(),
+                   cell_offsets.begin() + 1);
```
chrisrichardson, line 188: This can be simplified by calculating the cell offsets on the fly in the next loop, eliminating both `cell_offsets` and `dofs_per_cell`.

garth-wells: @chrisrichardson can you make the change in this branch?

cpp/dolfinx/fem/DofMap.cpp
```diff
@@ -157,6 +157,54 @@ fem::DofMap build_collapsed_dofmap(MPI_Comm comm, const DofMap& dofmap_view,
...
+  std::vector<std::int32_t> data(index_offsets.back());
+  std::vector<int> pos = index_offsets;
+  for (int c = 0; c < num_cells; ++c)
+  {
+    auto dofs = dofmap.links(c);
+    for (int i = 0; i < dofs.rows(); ++i)
+      data[pos[dofs[i]]++] = cell_offsets[c] + i;
+  }
+
+  // Sort the source indices for each global index
```
chrisrichardson, line 199: Is this needed? If so, maybe add a comment for the reason.

garth-wells: It should help with linear data access.

PR #1091, opened 2020-07-20
MeshTags dtype

chrisrichardson:
Use Python layer to wrap templated MeshTags types, using `dtype` to distinguish them.


python/dolfinx/mesh.py
```diff
@@ -121,19 +121,53 @@ def Mesh(comm, cell_type, x, cells, ghosts, degree=1, ghost_mode=cpp.mesh.GhostM
...
+                values = numpy.full(indices.shape, values, dtype=numpy.double)
+
+            dtype = values.dtype.type
+            if dtype not in _meshtags_types.keys():
+                raise KeyError("Datatype {} of values array not recognised".format(dtype))
+
+            fn = _meshtags_types[dtype]
+            self._cpp_object = fn(mesh, dim, indices, values)
+            self.values = self._cpp_object.values
+            self.indices = self._cpp_object.indices
```
jhale, line 142: else: raise?

chrisrichardson: No. else pass. The constructor is called with no arguments in the classmethod.

chrisrichardson: I'm not very happy with this design, but I am not sure how to do it better.

python/dolfinx/mesh.py
```diff
@@ -111,19 +104,85 @@ def create_mesh(comm, cells, x, domain, ghost_mode=cpp.mesh.GhostMode.shared_fac
...
+
+    @property
+    def values(self):
+        return self._cpp_object.values
+
+    def ufl_id(self):
+        return self._cpp_object.id
+
+
+def create_meshtags(mesh: cpp.mesh.Mesh, dim: int, indices: typing.Union[numpy.array, cpp.graph.AdjacencyList_int32],
```
michalhabera, line 161: I find it confusing to merge 2 very different code paths into seemingly one method.
1. A code path where I pass directly indices of mesh entities to pass is a very fast simple assignment in C++.
2. If I pass AdjacencyList, i.e. local entity-vertex connectivities this has to do much more work to find actual entity indices.

Moreover C++'s create_meshtags() is reserved for the option 2. We discussed that simple object instantiation which is almost no-op (option 1.) should be done in constructor.

chrisrichardson:
Closing as now very out-of-date. This is still an issue though.

PR #1095, opened 2020-07-27
Remove singleton class (SubSystemsManager)

garth-wells:
This removes the rather ugly singleton for initialising and finalising MPI, PETSc, etc from C++. It makes the actions explicit and removes one of the problem areas for multi-threaded applications.

This change is consistent with the C++ interface being explicit. It does not affect the Python interface.

### DIFF
--- a/cpp/demo/hyperelasticity/main.cpp
+++ b/cpp/demo/hyperelasticity/main.cpp
@@ -113,12 +113,11 @@ int main(int argc, char* argv[])
     // .. code-block:: cpp
 
     // Create mesh and define function space
-    std::array pt{Eigen::Vector3d(0, 0, 0), Eigen::Vector3d(1, 1, 1)};
-
     auto cmap
         = fem::create_coordinate_map(create_coordinate_map_hyperelasticity);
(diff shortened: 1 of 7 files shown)


garth-wells:
We've had this for a long time - the default has been to initialise MPI with threads enabled.

chrisrichardson:
I'd say it's better just to use regular mpi_init, as this is confusing, since we don't use the features.

garth-wells:
We have tests that used threaded assembly (parallel Numba), although in this case `mpi4py` should do the initialisation.

chrisrichardson:
>  MPI_THREAD_MULTIPLE informs MPI that the application is multithreaded, that any thread may issue MPI calls and that different threads may issue MPI calls at the same time.

This is totally beyond anything we need. Even if using threading, we don't issue MPI calls within it. If using threads with no MPI calls, it doesn't need to know about it.

garth-wells:
@chrisrichardson that's not quite true. Using threads requires at least `MPI_THREAD_FUNNELED`.

PR #1096, opened 2020-07-29
pass PYTHON_EXECUTABLE as well as Python3_EXECUTALBE from setup.py to make

js947:
The pybind11 CMake module reads PYTHON_EXECUTABLE and the PETSc4py CMake module reads Python3_EXECUTABLE. We should specify both to avoid potential problems of having two different pythons in the same build.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -49,6 +49,7 @@ class CMakeBuild(build_ext):
     def build_extension(self, ext):
         extdir = os.path.abspath(os.path.dirname(self.get_ext_fullpath(ext.name)))
         cmake_args = ['-DCMAKE_LIBRARY_OUTPUT_DIRECTORY=' + extdir,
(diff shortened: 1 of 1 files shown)


jhale:
Could you add a comment? We'll never remember this in two years!

js947:
CMake will probably change everything in 2 years ;-) but I take your point, comment added.

garth-wells:
What is the recommend variable name - `PYTHON_EXECUTABLE` or `Python3_EXECUTABLE`? DOLFIN provides `FindMPI4PY.cmake` and `FindPETSc4py.cmake` so these we can easily change.

js947:
I think Python3_EXECUTABLE is the more normal one. It is what you get from FindPython3. 

PYTHON_EXECUTABLE is required by the pybind11 modules, which come with the pybind11 install. Specifically the FindPythonLibsNew.cmake file which looks like it might be a shim from quite some time ago.

PR #1097, opened 2020-07-30
add missing python dependencies scipy and pybind11

js947:
scipy is used in `python/dolfinx/fem/assemble.py` and pybind11 is used in lots of places to build the python interface. 

Inspired by the comment here https://github.com/spack/spack/pull/17743#discussion_r462688105

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -17,6 +17,7 @@ RESTRICT_REQUIREMENTS = ">=2019.2.0.dev0,<2019.3"
 
 REQUIREMENTS = [
     "numpy",
+    "scipy",
     "mpi4py",
     "petsc4py",
     "fenics-ffcx",
--- a/python/setup.py
+++ b/python/setup.py
@@ -18,6 +18,7 @@ RESTRICT_REQUIREMENTS = ">=2019.2.0.dev0,<2019.3"
 REQUIREMENTS = [
     "numpy",
     "scipy",
+    "pybind11"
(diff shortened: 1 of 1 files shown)


python/setup.py
```diff
@@ -17,6 +17,8 @@
 
 REQUIREMENTS = [
     "numpy",
+    "scipy",
```
garth-wells, line 20: The SciPy dependency should be removed in DOLFIN - it's really just a bit of testing code. I can be a test dependency.

python/setup.py
```diff
@@ -17,6 +17,8 @@
...
+    "pybind11",
```
garth-wells, line 21: This is tricky - `pip install pybind11` isn't enough because it misses the `pybind11` CMake files.

python/setup.py
js947: I see the only one reference to scipy (outside of testing code) which is here https://github.com/FEniCS/dolfinx/blob/0c0fbc15367ae3ba1faddc867177bf38a8ef56e2/python/dolfinx/fem/assemble.py#L196

chrisrichardson: I think it is also in pybind implicitly (pybind understands scipy sparse <-> Eigen sparse translation).

garth-wells:
PR #1103 remove the scipy dependency.

pybind11 is tricky because we need the pybind11 CMake files, but the pip package does not provide these.

PR #1111, opened 2020-08-18
Fix github actions if syntax

jhale:


francesco-ballarin:
Apologies for messing it up in the latest PR.

I am puzzled, why is it that https://github.com/FEniCS/dolfinx/blob/cb8047cc68778a13113fd5f3467d49881637d92b/.github/workflows/ccpp.yml#L13 does not need the `${{ ... }}`? Is it because of the quotes?

jhale:
No, I made the same 'error', but it's not 100% clear to me why either the quotes `""` or brackets `${{ }}` are needed:

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions

jhale:
Infact, it's still not fixed: https://github.com/FEniCS/dolfinx/runs/997438705?check_suite_focus=true

PR #1115, opened 2020-08-21
Use package dir

chrisrichardson:
* Add package_dir to setup.py for dolfinx (similar to setup for ffcx)

This seems to help when installing to non-system paths (i.e. not to /usr/lib/python3.7/site-packages but another user location).
Otherwise, if setup.py generates an easy-install.pth file, it just gets ignored.

Perhaps this is not an issue when using `pip install`  - but with `setup.py` on our spack build it seems to be a problem.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -86,4 +86,5 @@ setup(name='fenics-dolfinx',
       ext_modules=[CMakeExtension('dolfinx.cpp')],
       cmdclass=dict(build_ext=CMakeBuild),
       install_requires=REQUIREMENTS,
+      package_dir={"dolfinx": "dolfinx"},
       zip_safe=False)


garth-wells:
I can't see why this is needed. Which version of Python are you using?

chrisrichardson:
3.8 - if there is a better solution, that's fine. Currently what happens is that it installs to a folder, adds a `.pth` file and sets the PYTHONPATH to that folder. If it is a "system" python folder, then the pth file is used, otherwise it is not, and hence the package is not found... unless a `site.py` file is also installed. Adding this line seems to force the creation of the `site.py` file.

garth-wells:
Can this be closed now that the spack install has been figured out?

PR #1124, opened 2020-09-09
Re-enable subset partitioning test in c ++

IgorBaratta:
Additionally, I'm working on the interface to parmetis and kahip and performance tests (subset partitioning and different partitioners in performance-test), that will come in future PR.


cpp/test/unit/mesh/cmap.ufl
```diff
@@ -0,0 +1,10 @@
+coord_element = VectorElement("Lagrange", triangle, 1)
+mesh = Mesh(coord_element)
+
+element = FiniteElement("Lagrange", triangle, 1)
+V = FunctionSpace(mesh, element)
+
+u = TrialFunction(V)
```
chrisrichardson, line 7: Do we need these lines?

IgorBaratta: Not really, but if I leave only the first two lines:
```
coord_element = VectorElement("Lagrange", triangle, 1)
mesh = Mesh(coord_element)
```
I get:
```
line 82, in main
    code_h, code_c = compiler.compile_ufl_objects(
.....
if isinstance(ufl_objects[0], ufl.form.Form):
IndexError: list index out of range
```
Would that be a ffcx issue?

michalhabera: Probably no-one generates a code just for cmap. If we'd like to allow this then some minor changes are needed in FFCx, register an issue.

IgorBaratta: Simple/palliative solution on the ffcx side:

https://github.com/FEniCS/ffcx/compare/igor/compile-meshes

garth-wells (approved):

PR #1126, opened 2020-09-09
Use namespaces instead of classes with only static members

IgorBaratta:
Some classes do not define a new data type/object and have only static member functions. I think that it's a better pattern to use namespaces for free/helper functions.

The idea of this PR is to discuss whether these changes are desirable, and if agreed, I can apply the same pattern where it is reasonable.


chrisrichardson:
I agree.

PR #1133, opened 2020-09-17
Adding back python interface for writing VTKFile for time dependent functions

jorgensd:
Also fixing a bug in pvtu files if original pvd file is in a subfolder.
I.e. if the "pvd" files would be written to results/name.pvd, the pvtu file would be written to results/results/name_p0_00000.pvtu


cpp/dolfinx/io/VTKFile.cpp
```diff
@@ -520,11 +520,14 @@ void pvtu_write_function(std::size_t dim, std::size_t rank,
   data_array_node.append_attribute("NumberOfComponents")
       = (unsigned int)num_components;
 
+  std::string delimiter = "/";
+  std::string local_filename = filename.substr(filename.find(delimiter) + 1);
```
garth-wells, line 524: This looks strange and platform-dependent. Any reason to not use `boost::filesystem`?

python/dolfinx/wrappers/io.cpp
```diff
@@ -112,6 +112,11 @@ void io(py::module& m)
            py::overload_cast<const dolfinx::function::Function<PetscScalar>&>(
                &dolfinx::io::VTKFile::write),
            py::arg("u"))
+      .def("write",
```
garth-wells, line 115: Do we need a need function, or can the existing write function be extended with `t=None`?

garth-wells (changes requested):

python/dolfinx/wrappers/io.cpp
jorgensd: We need two separate functions, as they point to two separate functions in the C++ layer. Note that we should change this in a rewrite of VTKFile.

garth-wells (approved):

PR #1139, opened 2020-09-29
Surface facet calculation

chrisrichardson:
Add two utility functions to `mesh::` 
* `entities_to_geometry` - computes the indices in `mesh::Geometry` of the vertices of the given entities
 - this is needed in several places in the library, and will be useful for future refactoring
* `exterior_facet_indices` - computes the facet indices of facets which lie on the external surface.
 - the logic is similar to that in `FormIntegrals`. It does not return exterior facets of ghost cells.

By using `facet_indices = exterior_facet_indices(mesh);` we can obtain a list of topology indices for surface facets (unoriented),
Further by using `geom_indices = entities_to_geometry(mesh, tdim -1, facet_indices, true);` we get the oriented geometry.

### DIFF
--- a/cpp/dolfinx/mesh/utils.cpp
+++ b/cpp/dolfinx/mesh/utils.cpp
@@ -833,11 +833,12 @@ mesh::entities_to_geometry(
       // Compute vector triple product of two edges and vector to midpoint
       Eigen::Vector3d p0 = mesh.geometry().node(entity_geometry(i, 0));
       Eigen::Matrix3d a;
-      a.row(0) = p0 - midpoint;
+      a.row(0) = midpoint - p0;
       a.row(1) = mesh.geometry().node(entity_geometry(i, 1)) - p0;
       a.row(2) = mesh.geometry().node(entity_geometry(i, 2)) - p0;
-      // Switch order if sign is negative
-      if (a.determinant() < 0.0)
+      // Midpoint direction should be opposite to normal, hence this should be
+      // negative. Switch points if not.
+      if (a.determinant() > 0.0)
         std::swap(entity_geometry(i, 1), entity_geometry(i, 2));
     }
   }
--- a/cpp/dolfinx/mesh/utils.cpp
+++ b/cpp/dolfinx/mesh/utils.cpp
@@ -797,7 +797,8 @@ mesh::entities_to_geometry(
   Eigen::Array<std::int32_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
       entity_geometry(entity_list.size(), num_entity_vertices);
 
-  if (orient and cell_type != dolfinx::mesh::CellType::tetrahedron)
+  if (orient
+      and (cell_type != dolfinx::mesh::CellType::tetrahedron or dim != 2))
     throw std::runtime_error("Can only orient facets of a tetrahedral mesh");
 
   const auto xdofs = mesh.geometry().dofmap();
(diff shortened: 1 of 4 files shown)


cpp/dolfinx/mesh/utils.cpp
```diff
@@ -785,3 +785,105 @@ Eigen::Array<std::int32_t, Eigen::Dynamic, 1> mesh::locate_entities_boundary(
...
+  // also not shared forward (ghost on another process)
+  const int num_facets = topology.index_map(tdim - 1)->size_local();
+  for (int f = 0; f < num_facets; ++f)
+  {
+    if (f_to_c->num_links(f) == 1
+        and fwd_shared_facets.find(f) == fwd_shared_facets.end())
+      surface_facets.push_back(f);
+  }
+
+  // Copy over to Eigen::Array
```
garth-wells, line 881: Can we avoid the copy here? Could change return type or avoid the duplication.

If copy is required, can just create an `Eigen::Map` and let the data copy happen in the return.

cpp/dolfinx/mesh/utils.h
```diff
@@ -108,5 +108,24 @@ Eigen::Array<std::int32_t, Eigen::Dynamic, 1> locate_entities_boundary(
         const Eigen::Ref<const Eigen::Array<double, 3, Eigen::Dynamic,
                                             Eigen::RowMajor>>&)>& marker);
 
+/// Compute the geometry indices of given entities from the mesh geometry
+/// @param[in] mesh Mesh
+/// @param[in] dim Topological dimension of the entities of interest
+/// @param[in] entity_list List of entities to obtain vertex geometry
```
garth-wells, line 114: `@param[in] entity_list List of entity indices (local)`

cpp/dolfinx/mesh/utils.h
```diff
@@ -108,5 +108,24 @@ Eigen::Array<std::int32_t, Eigen::Dynamic, 1> locate_entities_boundary(
         const Eigen::Ref<const Eigen::Array<double, 3, Eigen::Dynamic,
                                             Eigen::RowMajor>>&)>& marker);
 
+/// Compute the geometry indices of given entities from the mesh geometry
```
garth-wells, line 111: The doc description isn't clear. Does it get geometry indices for the vertices of the mesh entities?

cpp/dolfinx/mesh/utils.h
```diff
@@ -108,5 +108,24 @@ Eigen::Array<std::int32_t, Eigen::Dynamic, 1> locate_entities_boundary(
...
+/// @param[in] mesh Mesh
+/// @param[in] dim Topological dimension of the entities of interest
+/// @param[in] entity_list List of entities to obtain vertex geometry
+/// @param[in] orient If true, in 3D, reorients facets to have consistent normal
+/// direction
+/// @return Vertex indices in the geometry dofmap of the mesh for each entity
```
garth-wells, line 117: `/// @return Indices in the geometry array for the mesh entity vertices, e.g.  indices(i, j) is the position in the geometry array of the j-th vertex of the entity entity_list[i].`

cpp/dolfinx/mesh/utils.h
```diff
@@ -108,5 +108,24 @@ Eigen::Array<std::int32_t, Eigen::Dynamic, 1> locate_entities_boundary(
...
+Eigen::Array<std::int32_t, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
+entities_to_geometry(
+    const mesh::Mesh& mesh, const int dim,
+    const Eigen::Array<std::int32_t, Eigen::Dynamic, 1>& entity_list,
+    bool orient);
+
+/// Compute exterior facet indices only
```
garth-wells, line 124: `/// Compute the indices (local) of all exterior facets. An exterior facet (co-dimension 1) is one that is connected globally to only one cell of co-dimension 0).`

garth-wells (changes requested):
Not all the docstrings were clear to me. Suggestions for re-wording inline.

PR #1140, opened 2020-09-30
Share source into Docker build context

jhale:
And some other tweaks to the Docker CI infrastructure.


.github/workflows/docker-end-user.yml
```diff
@@ -11,6 +11,7 @@ on:
   schedule:
     # '*' is a special character in YAML, so string must be quoted
     - cron: "0 4 * * *"
+  workflow_dispatch: ~
```
francesco-ballarin, line 14: FYI, I believe that the ~ is actually unnecessary (I have workflow with empty `workflow_dispatch:` and that seems to work all the same).

PR #1141, opened 2020-10-01
Slim down final docker images

jhale:
Needs to be tested using manual execution of actions.


docker/dolfinx-complex-mode
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
-export LD_LIBRARY_PATH=/usr/local/dolfinx-complex/lib:$LD_LIBRARY_PATH
 export PATH=/usr/local/dolfinx-complex/bin:$PATH
```
garth-wells, line 2: What does DOLFIN install in `bin/`?

docker/dolfinx-complex-mode
```diff
@@ -1,6 +1,6 @@
...
 export PKG_CONFIG_PATH=/usr/local/dolfinx-complex/lib/pkgconfig:$PKG_CONFIG_PATH
 export PETSC_ARCH=linux-gnu-complex-32
 export PYTHONPATH=/usr/local/dolfinx-complex/lib/python3.8/dist-packages:$PYTHONPATH
+export LD_LIBRARY_PATH=/usr/local/petsc/$PETSC_ARCH/lib:/usr/local/dolfinx-complex/lib:$LD_LIBRARY_PATH
```
garth-wells, line 6: Is this needed? Is the issue runtime linking issues? I would have expected CMake to take care of linking.

docker/dolfinx-complex-mode
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
-export LD_LIBRARY_PATH=/usr/local/dolfinx-complex/lib:$LD_LIBRARY_PATH
 export PATH=/usr/local/dolfinx-complex/bin:$PATH
 export PKG_CONFIG_PATH=/usr/local/dolfinx-complex/lib/pkgconfig:$PKG_CONFIG_PATH
 export PETSC_ARCH=linux-gnu-complex-32
```
garth-wells, line 4: I would expect a corresponding `PETSC_DIR` to be consistent with how the PETSc install system is intended to be used.

garth-wells (commented):
Looks like it's getting quite complicated . . . 

Does each image have only one type of PETSc build?

docker/dolfinx-complex-mode
jhale: Nothing, hadn't noticed :) Will remove.

docker/dolfinx-complex-mode
jhale: I managed to fix this before but forgot to remove the hacky changes to LD_LIBRARY_PATH, will remove and test.

jhale:
> Looks like it's getting quite complicated . . .
> 
> Does each image have only one type of PETSc build?

For simplicity we agreed to build all types of PETSc into each image.

The alternative is to have a proliferation of tags (I think we would need 12 in total to cover all current combinations).

garth-wells:
I like all PETSc builds in the image, so that's good.

Does petsc4py get installed into `/usr(/local)/lib/python.....`?, with the correct 'version' determined by `PETSC_ARCH`?

jhale:
Yes, petsc4py picking up PETSC_ARCH and dynamically loading the correct PETSc shared object works perfectly.

I have removed the LD_LIBRARY_PATH hacks for PETSc/petsc4py and it works great, I'll push that now.

PR #1143, opened 2020-10-02
Compatibility with pygmsh 7.0.0

francesco-ballarin:
See https://github.com/nschloe/pygmsh/commit/62516e7186af53ed5c8b02deb9a9876d79c6f0f9

PR #1145, opened 2020-10-02
Switch from pygmsh to the gmsh Python API

jorgensd:
As pygmsh has started using the gmsh python API from version 7.0.0, there is no point adding an extra dependency and compatibility layer between gmsh and dolfinx.

Meshio has also been deprecated from the docker images. It is not required to run demos or tests.
A direct interface from msh->dolfinx has been added, using the GMSH API to read in the mesh on one processor and distribute the data using dolfin-x.

Also worthwhile discussing whether the functions introduced in `demo_gmsh.py` should be part of `dolfinx.io`.

Demo rewritten using only GMSH python API. Related to #1143


docker/Dockerfile
```diff
@@ -145,12 +145,12 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
...
 #   documentation or run tests.
 # - Third set of packages are optional but required for
-#   pygmsh/meshio/DOLFINX mesh pipeline.
+#   gmsh/meshio/DOLFINX mesh pipeline.
 RUN pip3 install --no-cache-dir mpi4py numba && \
     pip3 install --no-cache-dir cffi cppimport flake8 pytest pytest-xdist sphinx sphinx_rtd_theme && \
     export HDF5_MPI="ON" && \
     export CC=mpicc && \
-    pip3 install --no-cache-dir --no-binary=h5py h5py meshio pygmsh
+    pip3 install --no-cache-dir --no-binary=h5py h5py meshio
```
garth-wells, line 153: Do we still need `meshio`?

python/demo/gmsh/demo_gmsh.py
```diff
@@ -5,33 +5,101 @@
 # SPDX-License-Identifier:    LGPL-3.0-or-later
 #
 # =====================================
-# Mesh generation using Gmsh and pygmsh
+# Mesh generation using the GMSH python API
 # =====================================
```
garth-wells, line 9: Fix the rst `=` markers.

python/demo/gmsh/demo_gmsh.py
```diff
@@ -5,33 +5,101 @@
...
+from dolfinx.mesh import create_mesh, create_meshtags
+from dolfinx.io import XDMFFile, ufl_mesh_from_gmsh
+from dolfinx.cpp.io import perm_gmsh, extract_local_entities
+from dolfinx import cpp
 from mpi4py import MPI
+import gmsh
+import numpy as np
+
+
+def gather_gmsh_physical_entities():
```
garth-wells, line 20: This function loops really complicated. Can it be simplified? 

At the lest, removing some line breaks and adding come comments would improve readability.

python/demo/gmsh/demo_gmsh.py
```diff
@@ -5,33 +5,101 @@
...
+    Extracts geometrical nodes from all entities in gmsh
+    and sorts them starting from index 0.
+    """
+    idx, points, _ = gmsh.model.mesh.getNodes()
+    points = points.reshape(-1, 3)
+    idx -= 1
+    srt = np.argsort(idx)
+    assert np.all(idx[srt] == np.arange(len(idx)))
+    x = points[srt]
+    return x
```
garth-wells, line 67: `return points[srt]`

python/demo/gmsh/demo_gmsh.py
```diff
@@ -5,33 +5,101 @@
...
+
+def sort_gmsh_points():
+    """
+    Extracts geometrical nodes from all entities in gmsh
+    and sorts them starting from index 0.
+    """
+    idx, points, _ = gmsh.model.mesh.getNodes()
+    points = points.reshape(-1, 3)
+    idx -= 1
+    srt = np.argsort(idx)
```
garth-wells, line 64: `srt` is cryptoc.

python/demo/gmsh/demo_gmsh.py
```diff
@@ -5,33 +5,101 @@
...
+                                        np.full(len(e_tags), tag)), axis=0)
+                else:
+                    meshes[e_type] = {"cells":
+                                      e_cells,
+                                      "cell_data": np.full(len(e_tags), tag)}
+
+    return meshes
+
+
+def sort_gmsh_points():
```
garth-wells, line 56: The is confusing without an in argument. Or the docstring is unclear.

python/demo/gmsh/demo_gmsh.py
```diff
@@ -5,33 +5,101 @@
...
+    """
+    Extracts geometrical nodes from all entities in gmsh
```
garth-wells, line 58: I don't understand this docstring.

python/dolfinx/io.py
```diff
@@ -62,17 +62,20 @@ def read_mesh(self, ghost_mode=cpp.mesh.GhostMode.shared_facet, name="mesh", xpa
...
-                   line=("interval", 1), line3=("interval", 2), line4=("interval", 3),
-                   vertex=("point", 1))
+# Map from Gmsh float to DOLFIN cell type and degree
+# http://gmsh.info//doc/texinfo/gmsh.html#MSH-file-format
+_gmsh_to_cells = {1: ("interval", 1), 2: ("triangle", 1),
+                  3: ("quadrilateral", 1), 4: ("tetrahedron", 1),
+                  5: ("hexahedron", 1), 8: ("interval", 2),
+                  9: ("triangle", 2), 10: ("quadrilateral", 2),
+                  11: ("tetrahedron", 2), 12: ("hexahedron", 2),
+                  15: ("point", 1), 21: ("triangle", 3),
```
garth-wells, line 72: I would have expected "point" to have degree 0.

garth-wells (changes requested):
Nice to see a change to working through the API in place of the filesystem.

docker/Dockerfile
jorgensd: It is not a strict dependency any longer. However, to interface with msh-files, we would require it. Maybe we should create a separate PR with a msh->dolfin demo.

garth-wells: Not sure I see the point. I would remove it if it isn't a test dependency. The few dependencies, the fewer things can can break a build.

python/demo/gmsh/demo_gmsh.py
jorgensd: We need this function to gather physical tags into meshes. For instance when using `BooleanFragements` with gmsh,  you get multiple volume entities that has to be loaded.
This splitting is very common with entities of lower dimensions, such as facets.
I can add in some comments to explain what is going on.

python/demo/gmsh/demo_gmsh.py
jorgensd: There are no in-arguments as everything is stored in the `gmsh.model`.
I will try to improve the doc string.

python/dolfinx/io.py
jorgensd: I can change it. For this, I was just following the previous `_gmsh_cells`.

docker/Dockerfile
jorgensd: We need to have an interface to `msh` files at some point, as it is quite strict to require everyone to go through the `gmsh` interface.
In many cases, people might have msh files from other software.
The main problem is that meshio `msh`->`xdmf` is not directly compatible with dolfinx, see: https://fenicsproject.discourse.group/t/transitioning-from-mesh-xml-to-mesh-xdmf-from-dolfin-convert-to-meshio/412/79?u=dokken

python/demo/gmsh/demo_gmsh.py
```diff
@@ -5,33 +5,101 @@
...
+    Returns a dictionary where the key is the gmsh cell type
+    and its content is a dictionary of cells and cell data.
+
+    """
+    phys_grps = gmsh.model.getPhysicalGroups()
+    meshes = {}
+    for dim, tag in phys_grps:
+        entities = gmsh.model.getEntitiesForPhysicalGroup(dim, tag)
+        for entity in entities:
+            element_types, elem_tags, node_tags =\
```
francesco-ballarin, line 34: As a followup to Garth comment. I believe that these two lines can fit in a single one (we have the limit at 120 chars, right?)

python/demo/gmsh/demo_gmsh.py
```diff
@@ -5,33 +5,101 @@
 # SPDX-License-Identifier:    LGPL-3.0-or-later
```
francesco-ballarin, line 5: Are the functions in this file general enough to be included as part of the library, and not only in this demo?

jorgensd: I think they eventually should be part of the library. I put them here, so that we can discuss how we would like to interact with the gmsh API.

python/demo/gmsh/demo_gmsh.py
jorgensd: Function has been slightly simplified and is now well documented. 
Should we make it part of the dolfinx.io?

python/demo/gmsh/demo_gmsh.py
francesco-ballarin: Great. Do you and @garth-wells think that these two functions can have `gmsh.model` or `gmsh.model.mesh` as input? This would allow avoiding `import gmsh` in `dolfinx.io`.
(BTW, how would one create two meshes at the same time with such gmsh api?)

jorgensd: We could put the model name as input, (then using Gmsh.model.setCurrent) such that it sets the current model, exports the geometry and topology, and returns the mesh.

jorgensd: > Great. Do you and @garth-wells think that these two functions can have `gmsh.model` or `gmsh.model.mesh` as input? This would allow avoiding `import gmsh` in `dolfinx.io`.
> 
> (BTW, how would one create two meshes at the same time with such gmsh api?)

However, i am not sure how one would take a Gmsh model as input.

jorgensd:
@garth-wells , @francesco-ballarin  I have now rewritten the IO such that one can either do:
1) Create the mesh in the python script and export the topology and geometry for mesh creation without any IO
2) Added IO from msh files, reading in the msh file with GMSH on processor 0, and distributing the data over the other processors. This function can also extract facet data and cell data, and create corresponding mesh tags.

garth-wells (changes requested):
Looking good, just a few more suggestions for changes.

python/demo/gmsh/demo_gmsh.py
garth-wells: Let's see how it shakes down and then decide if it can go somewhere in the library. Maybe rather than going deep into the core code we could have a `dolfinx.utilities` module or something similar.

cpp/dolfinx/io/cells.cpp
```diff
@@ -237,11 +237,7 @@ std::vector<std::uint8_t> gmsh_tetrahedron(int num_nodes)
   case 4:
     return {0, 1, 2, 3};
   case 10:
-    // NOTE: GMSH DOCUMENTATION IS WRONG, it would have the following
-    // permutation:
-    // return std::vector<std::uint8_t>{0, 1, 2, 3, 9, 6, 8, 7, 4, 5};
-    // This is the one returned by pygmsh
-    return {0, 1, 2, 3, 9, 6, 8, 7, 5, 4};
+    return std::vector<std::uint8_t>{0, 1, 2, 3, 9, 6, 8, 7, 4, 5};
```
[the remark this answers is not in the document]
jorgensd: Your right, this was just a remainder from an older iteration of this code.

python/dolfinx/io.py
```diff
@@ -62,17 +65,207 @@ def read_mesh(self, ghost_mode=cpp.mesh.GhostMode.shared_facet, name="mesh", xpa
...
+                    numpy.concatenate((topologies[element_type]["cell_data"],
+                                       numpy.full(num_el, tag)), axis=0)
+            else:
+                topologies[element_type] = {"topology": element_topology,
+                                            "cell_data": numpy.full(num_el, tag)}
+
+    return topologies
+
+
+def extract_geometry(gmsh_model, model_name=None):
```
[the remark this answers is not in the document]
francesco-ballarin: Would not that require importing gmsh, thus making gmsh a required dependency?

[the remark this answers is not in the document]
jorgensd: Not sure what the hint should be. To make a proper hint we would have to import gmsh at the top of the file, instead of inside `read_from_msh`.

docker/Dockerfile
```diff
@@ -134,23 +133,16 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
     apt-get -y install \
     libglu1 \
     libxcursor-dev \
-    libxinerama1 && \
-    apt-get -y install \
-    python3-lxml && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
+    libxinerama1
```
francesco-ballarin, line 136: Why were the last two lines removed? I gather that they were there to reduce the docker image size.

docker/Dockerfile
```diff
@@ -134,23 +133,16 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
...
 # - Second set of packages are recommended and/or required to build
 #   documentation or run tests.
-# - Third set of packages are optional but required for
-#   pygmsh/meshio/DOLFINX mesh pipeline.
 RUN pip3 install --no-cache-dir mpi4py numba && \
     pip3 install --no-cache-dir cffi cppimport flake8 pytest pytest-xdist sphinx sphinx_rtd_theme && \
     export HDF5_MPI="ON" && \
-    export CC=mpicc && \
-    pip3 install --no-cache-dir --no-binary=h5py h5py meshio pygmsh
+    export CC=mpicc
```
francesco-ballarin, line 145: Was exporting CC only needed by h5py?

python/demo/gmsh/demo_gmsh.py
jorgensd: > Let's see how it shakes down and then decide if it can go somewhere in the library. Maybe rather than going deep into the core code we could have a `dolfinx.utilities` module or something similar.

If we do not make a type hint for the gmsh model, the only import of gmsh now is when reading in msh files.

francesco-ballarin: Both `dolfinx.io` and `dolfinx.utilities` sound good to me.

python/dolfinx/io.py
```diff
@@ -62,17 +65,210 @@ def read_mesh(self, ghost_mode=cpp.mesh.GhostMode.shared_facet, name="mesh", xpa
...
+                   (Default: False)
+        facet_data: Boolean, True if a mesh tag for facet data should be
+                    returned (Default: False)
+        gdim: Geometrical dimension of problem (Default: 3)
+    """
+    if gdim is None:
+        gdim = 3
+    if MPI.COMM_WORLD.rank == 0:
+        import gmsh
+        gmsh.initialize()
```
francesco-ballarin, line 168: What would happen if a user had called already `gmsh.initialize()` before calling this function?

python/dolfinx/io.py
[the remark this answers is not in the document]
garth-wells: Indeed, normally it would require an import but to avoid that I think the type can be given as a string, e.g. `gmsh_model: "gmsh.foo".

docker/Dockerfile
jorgensd: Good question, I guess so. I'll remove them

python/dolfinx/io.py
jorgensd: Fixed in latest commit

python/dolfinx/io.py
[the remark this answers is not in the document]
jorgensd: This doesn't pass flake8 (See failing CI).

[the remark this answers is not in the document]
jorgensd: I will revert it if you do not have any other suggestions.

python/dolfinx/io.py
```diff
@@ -165,8 +165,14 @@ def read_from_msh(filename: str, cell_data=False, facet_data=False, gdim=None):
         gdim = 3
     if MPI.COMM_WORLD.rank == 0:
         import gmsh
-        gmsh.initialize()
-
+        # Check if gmsh is allready initialized
```
francesco-ballarin, line 168: typo: already

python/demo/gmsh/demo_gmsh.py
```diff
@@ -4,32 +4,82 @@
...
+    model.addPhysicalGroup(2, [rect_tag], tag=7)
+    # Tag the line [0,0]->[0,1] with tag 11
+    facet_entities = model.occ.getEntities(dim=1)
+    for entity in facet_entities:
+        midpoint = model.occ.getCenterOfMass(dim=1, tag=entity[1])
+    if np.allclose(np.array(midpoint), [0, 0.5, 0]):
+        model.addPhysicalGroup(1, [entity[1]], tag=11)
+    model.mesh.generate(2)
+    gmsh.write("mesh.msh")
+
```
francesco-ballarin, line 45: Demos seems to be hanging in parallel. Is it possible that a barrier is needed here?

francesco-ballarin (changes requested):

python/demo/gmsh/demo_gmsh.py
garth-wells: Why are we writing to file?

jorgensd: To show that we can read from msh.
The issue here was that I only initialized gmsh on one processor, then using it on every processor later on.

garth-wells: I think it should be removed, or encapsulated in the demo. DOLFIN supports reading from XDMF. Support for other formats needs to be discussed since it's not clear where one should stop.

jorgensd: I disagree, as there is no clear description of how the dolfinx-XDMF format is.
There are plenty of XDMF-files (for instance those created by meshio-convert from msh) that does not work with dolfinx. This makes it possible for any msh mesh to be loaded, and I think that should be highlighted

jorgensd: I've removed them from the demo and dolfinx.io, as they would make the demo quite complex if encapsulated.

francesco-ballarin (approved):

jorgensd:
> Looking good, just a few more suggestions for changes.

Is this ready to merge now?

garth-wells (approved):

Issue #1152, opened 2020-10-06
Move gmsh Python utility functions to separate module


garth-wells:
`dolfin.io` (`io.py`) contains a number of utility function with working with the gmsh via its Python API. It might be best to isolate the helper functions for a non-core 3rd party tools in a separate module, e.g. `dolfinx.mesh.gmsh`.

PR #1153, opened 2020-10-06
Restructured GMSH IO as separate sub module.

jorgensd:
Addresses issue #1152


jorgensd:
This PR is heavily outdated, so closing it. However, we should think about moving gmsh to a separate module

Issue #1156, opened 2020-10-08
Read mixed topologies from XDMF


jorgensd:
Currently we cannot read mixed topologies. In the current state, this is relevant for topologies of different dimensions (Triangle, Line) etc. However, in the future this would be relevant for "quad-triangle" meshes.

The problem is that when the topology data is of the following form if the element type is "Mixed":
```
[cell_type_1 node_{1,1}, .... node_{1,N_1}, cell_type_2 node_{2,1}, ...., node_{2, N_2},... ]
```
For each cell in the mesh, not for each individual type, i.e. if triangle has id 4, the topology for two triangles will be
```
4, 0, 1, 2, 4, 1, 2, 3
```
To further obscure matters, if the topology type is a line, there will be two input arguments, the cell_id (2) and the number of nodes describing the line (In all cases it should be 2, as the have a special edge3 element as well). I.e. two first order lines will be:
```
2, 2, 0, 1, 2, 2, 1, 2
```
Due to this I suggest that all data is read in on processor 0, then distributed in the mesh partitioning, as we cannot aprior
know how to split the data between elements.
MWE:
```
import dolfinx.io
import meshio
import gmsh

from mpi4py import MPI

# Create 2D mesh with 1D faces
gmsh.initialize()
gmsh.option.setNumber("Mesh.CharacteristicLengthMin", 0.5)
gmsh.option.setNumber("Mesh.CharacteristicLengthMin", 0.5)
rect = gmsh.model.occ.addRectangle(0, 0, 0, 1, 1)
gmsh.model.occ.synchronize()

facets = gmsh.model.getEntities(dim=1)
for i, facet in enumerate(facets):
    gmsh.model.addPhysicalGroup(facet[0], [facet[1]], i)
gmsh.model.addPhysicalGroup(2, [rect], 1)

gmsh.model.mesh.generate(2)
gmsh.write("mesh.msh")
gmsh.finalize()

# Convert MSH to XDMF
meshio_mesh = meshio.read("mesh.msh")
meshio.write("test_mesh.xdmf", meshio_mesh)


# Read into dolfin-X
with dolfinx.io.XDMFFile(MPI.COMM_WORLD, "test_mesh.xdmf", "r") as xdmf:
    mesh = xdmf.read_mesh(name="Grid")
```

PR #1157, opened 2020-10-09
Add UI for XDMF mixed topologies

jorgensd:
See details about the input data in #1156 

This PR adds the possibility of selecting with topological dimension of a XDMFTopology of typed "mixed" one should extract. Default behavior will be to extract the topology with highest topological dimension.

Similar features added for reading in meshtags.

Note that all data is read on the first processor, then distributed later on, as there is no reasonable way of partitioning the cell_topology or cell_values a priori.

### DIFF
--- a/cpp/dolfinx/io/XDMFFile.cpp
+++ b/cpp/dolfinx/io/XDMFFile.cpp
@@ -165,9 +165,10 @@ mesh::Mesh XDMFFile::read_mesh(const fem::CoordinateElement& element,
                                const std::string xpath) const
 {
   // Read mesh data
-  const Eigen::Array<std::int64_t, Eigen::Dynamic, Eigen::Dynamic,
-                     Eigen::RowMajor>
-      cells = XDMFFile::read_topology_data(name, xpath);
+  const std::pair < dolfinx::mesh::CellType,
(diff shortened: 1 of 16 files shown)


michalhabera:
If this PR makes reading the entire topology on the rank0 processor, that is surely wrong.

What do you mean, "no reasonable way of partitioning cell_topology"? Do you mean partitioning the HDF file to select a chunk that should be read? This could be achieved by reading a chunk of size which is a multiple of the least common multiple of topology sizes of all cell types.

jorgensd:
> If this PR makes reading the entire topology on the rank0 processor, that is surely wrong.
> 
> What do you mean, "no reasonable way of partitioning cell_topology"? Do you mean partitioning the HDF file to select a chunk that should be read? This could be achieved by reading a chunk of size which is a multiple of the least common multiple of topology sizes of all cell types.

As PolyLine has an unbounded number of nodes in the topology (defined by the second argument), there is no common multiple. If we assume PolyLine always has 2 nodes (being first order), we can find the least common multiple. We can have an asset for such types.

garth-wells:
This is a non-starter if all data is read on one rank.

jorgensd:
Closed due to out of data and soon to be deprecated (xdmf) format

PR #1158, opened 2020-10-10
Fix 3D plotting with matplotlib

jorgensd:
3D plotting with dolfinx and matplotlib has not been working for a long time due to the lack of boundary mesh, along with an update in matplotlib, yielding the following error:
```
NotImplementedError: It is not currently possible to manually set the aspect on 3D axes
```
when running:
```
import matplotlib.pyplot as plt
from dolfinx.plotting import plot
from dolfinx import UnitCubeMesh
from mpi4py import MPI

mesh = UnitCubeMesh(MPI.COMM_WORLD, 3, 3, 3)
plot(mesh, color="w", shade=True, edgecolors="b")

plt.savefig("mesh.png")
```
This fix utilizes #1139 to restore plotting of 3D meshes.


jmv2009:
I'm just removing the offending "ax.set_aspect('equal')" from plotting.py to fix this.

jorgensd:
> I'm just removing the offending "ax.set_aspect('equal')" from plotting.py to fix this.
As the boundary mesh class is not existing in dolfinx, the code path of tdim=3, gdim=3 will not run even if you remove the set_aspect. Therefore I have moved my helper function from a test into dolfinx.mesh to cope with this.

garth-wells:
Maybe we should just remove Matplotlib and use vedo and/or pvista instead? See #1159.

jorgensd:
> Maybe we should just remove Matplotlib and use vedo and/or pvista instead? See #1159.

I can give vedo a go

jorgensd:
> Maybe we should just remove Matplotlib and use vedo and/or pvista instead? See #1159. 

I think we should merge this, and then think about #1159 (and not add any more effort into matplotlib)

python/dolfinx/mesh.py
```diff
@@ -126,6 +126,32 @@ def MeshTags(mesh, dim, indices, values):
     return fn(mesh, dim, indices, values)
 
 
+def create_boundary_mesh(mesh, comm, orient=False):
```
garth-wells, line 129: Let's put this back in the plotting module. We don't want to encourage its use without careful thought.

garth-wells (changes requested):

garth-wells (approved):

Issue #1159, opened 2020-10-12
Add demo using vedo and/or pyvista for visualisation


garth-wells:
It looks like Python/Jupyter tools for 3D visualisation have come along quite a bit recently. `vedo` (https://vedo.embl.es/) and `pyvista` (https://www.pyvista.org/) look nice and work with mesh data in VTK ordering, which should make it reasonably straightforward to interface to them. There is even support for high-order geometries.  

- [x] pyvista
- [ ] vedo

The design should avoid importing vdeo or pyvista in the DOLFINX core, but add interfaces for returning the data in a suitable format.

jorgensd:
As Vedo requires VTK<9.0.1, which is not pip installable, I suggest we test pyvista.

jorgensd:
@garth-wells I do not think we should proceed with either of these software just yet.
*vedo*: Does not use latest VTK available with pip, which requires user installation of an older VTK version
*pyvista*: Does not support higher order Lagrangian elements yet (https://github.com/pyvista/pyvista/issues/947)

garth-wells:
Both worked fine for me recently when I tested. Higher-order elements isn't really an issue since we don't support higher-order elements with Matplotlib.

Is there a good reason to support Matplotlib over vedo or pyvista?

jorgensd:
Matplotlib is not dependent on VTK. This means that visualization can be done without extra configuration in docker and on HPC systems (as long as matplotlib is installed). VTK can cause issues such as https://github.com/pyvista/pyvista/issues/155 when using docker containers, which would require extra overhead from the users side. I can of course make a demo using Pyvista on first order meshes (and add some io wrappers).

marcomusy:
> As Vedo requires VTK<9.0.0, which is not pip installable, 

vtk 8.* is pip installable.

jorgensd:
> > As Vedo requires VTK<9.0.0, which is not pip installable,
> 
> vtk 8.* is pip installable.

You are absolutely right. For some strange reason my pip installation didn’t want to find any of the other versions.

marcomusy:
It seems _you_ were actually right! Apparently vtk doesn't have wheels for python 3.8 - that might explain why your pip did not find the version.
I'll just remove the requirement as `vedo` can work with version 9 too (see https://github.com/marcomusy/vedo/issues/235).

jorgensd:
Currently, there is some work required for usage of vedo, see for instance: https://github.com/marcomusy/vedo/issues/297

[...]

PR #1161, opened 2020-10-13
Pyvista demos

jorgensd:
Addresses #1159 .

A minimal interface for plotting with pyvista.
The interface supports:

- Plotting of any dolfinx mesh (or sub-entities of the mesh) `dolfinx.plotting.pyvista_topology_from_mesh`, both first and second order geometries. This also makes it possible to plot parts of the mesh, for instance by using `mt.indices[mt.values==i]` is indicated in the demo.
- Plotting of (almost) arbitrary order CG and DG functions `dolfinx.plotting.pyvista_topology_from_function`. Can also be used to plot subsets of cells in the mesh (entities as input). This function creates a topology based on the dof coordinates of a CG/DG space, such that we can use the underlying array in the function directly to visualize function values.
- Plotting vector valued functions in 2D/3D with either Glyphs or Streamtracing.
 
The demo explains how to plot the following figures
2D_function_warp
2D_markers
3D_function
3D_wireframe_with_nodes
DG
vectors_0
streamlines_0
vectors_0


python/dolfinx/plotting.py
```diff
@@ -6,7 +6,7 @@
 
 import os
 import warnings
-
+import numba
```
garth-wells, line 9: Numba shouldn't become a DOLFIN-X dependency.

python/dolfinx/plotting.py
```diff
@@ -447,25 +447,33 @@ def plot(object, *args, **kwargs):
 _dolfin_to_vtk_cell = {cpp.mesh.CellType.point: 1, cpp.mesh.CellType.triangle: 69,
                        cpp.mesh.CellType.quadrilateral: 70, cpp.mesh.CellType.tetrahedron: 71,
                        cpp.mesh.CellType.hexahedron: 72}
-# Permutation for Dolfinx DG layout to VTK (triangles)
-_perm_dg = {1: [0, 1, 2], 2: [0, 2, 5, 1, 4, 3], 3: [0, 3, 9, 1, 2, 6, 8, 7, 4, 5],
-            4: [0, 4, 14, 1, 2, 3, 8, 11, 13, 12, 9, 5, 6, 7, 10]}
-_perm_dq = {1: [0, 1, 3, 2], 2: [0, 2, 8, 6, 1, 5, 7, 3, 4], 3: [0, 3, 15, 12, 1, 2, 7, 11, 13, 14, 4, 8, 5, 6, 9, 10]}
+# Permutation for Dolfinx DG layout to VTK
```
garth-wells, line 450: Don't we normally have this type of information in the C++ layer?

python/dolfinx/plotting.py
```diff
@@ -447,25 +447,33 @@ def plot(object, *args, **kwargs):
...
 
 
 # @numba.njit(cache=True)
-def create_pyvista_cell_topology(topology, perm_to_vtk, num_cells, num_dofs_per_cell, dofmap, offsets):
+def create_pyvista_cell_topology(topology, perm_to_vtk, num_cells, num_dofs_per_cell, dofmap, offsets, entities):
     """
     Creating a mesh topology given a dofmap (dolfin-X ordering).
     Return a pyvista mesh topology (vtk ordering)
     """
     for i in range(num_cells):
```
garth-wells, line 469: Is there a clever NumPy way to avoid the loop?

cpp/dolfinx/io/pyvistawriter.cpp
```diff
@@ -0,0 +1,48 @@
+// Copyright (C) 2020 Garth N. Wells
```
garth-wells, line 1: I don't think we should have these files. They key is returning the raw data in a flexible enough format that pyvista can process it.

python/dolfinx/plotting.py
jorgensd: Im moving all this down into C++. I'll let you know once its done

python/dolfinx/plotting.py
jorgensd: These are a new set of permutations, from the DG lexicographical ordering to VTK.

python/dolfinx/plotting.py
jorgensd: Removing this

cpp/dolfinx/io/pyvistawriter.cpp
jorgensd: This is why I first used numba, but moved everything into C++ to avoid getting numba as a dependency.

python/dolfinx/plotting.py
jorgensd: I can think about if there is another way of obtaining the topology without using loops.

python/dolfinx/plotting.py
jorgensd: They are not required for any other usage than making meshes out of DG dofmaps, so I'd rather keep them locally in this file.

python/dolfinx/plotting.py
```diff
@@ -439,3 +438,78 @@ def plot(object, *args, **kwargs):
...
+    num_nodes = geometry_entities.shape[1]
+    e_type = cpp.mesh.cell_entity_type(mesh.topology.cell_type, dim)
+    map_vtk = cpp.io.transpose_map(cpp.io.perm_vtk(e_type, num_nodes))
+    topology = np.zeros((num_cells, num_nodes + 1), dtype=np.int32)
+    topology[:, 0] = num_nodes
+    topology[:, 1:] = geometry_entities[:, map_vtk]
+    return topology.reshape(1, -1)[0], cell_types
+
+
+def pyvista_topology_from_function_space(u, entities=None):
```
garth-wells, line 479: Looks like this function only returns mesh data?

python/dolfinx/wrappers/io.cpp
```diff
@@ -31,9 +31,13 @@ namespace dolfinx_wrappers
 void io(py::module& m)
 {
 
+  // dolfinx::io::cell vtk cell type converter
+  m.def("get_vtk_cell_type", &dolfinx::io::cells::get_vtk_cell_type);
+
   // dolfinx::io::cell permutation functions
   m.def("perm_vtk", &dolfinx::io::cells::perm_vtk);
   m.def("perm_gmsh", &dolfinx::io::cells::perm_gmsh);
+  m.def("transpose_map", &dolfinx::io::cells::transpose);
```
garth-wells, line 40: This is a super cryptic function name. C++ is marginally better (since it's in an extra namespace).

I'd suggest removing it from Python and performing the operation locally. Avoid bloating the Python API will niche functions that can be easily done from Python.

python/dolfinx/plotting.py
jorgensd: Yes, it only returns the topology, as we can easily access the underlying geometry (with `u.function_space.tabulate_dof_coordinates`) and the nodal data as `u.vector.array`,
I could move the  return of this data into this function. However, as there are special reshaping options for vector data, and if you would like to use the real or complex part of the function, I've let these return types out of the function

jorgensd: Note that the `dofmap` used to make the topology in this function is the one of the function-space.

garth-wells: It would make sense to only pass in `topology` rather than `u`.

jorgensd: How would you suggest to do that?
The topology created from the function space is not the mesh topology. It is a topology based on the degrees of freedom of the function space, and the corresponding dof coordinates.
This means that if we supply a DG 1- DG N field, we create a mesh where no cell is connected to its neighbour.

For a CG-N space on a CG-1 mesh, it creates the topology of a CG-N mesh.

garth-wells: I think the docstring needs updating since it sounds like it's creating a new topology that isn't based on the mesh, whereas the docstring says that it returns the mesh topology (and geometry, which I don't see in the return args).

python/dolfinx/plot.py
```diff
@@ -0,0 +1,110 @@
...
+                                              3: [0, 3, 15, 12, 1, 2, 7, 11, 13, 14, 4, 8, 5, 6, 9, 10]},
+            cpp.mesh.CellType.hexahedron: {1: [0, 1, 3, 2, 4, 5, 7, 6],
+                                           2: [0, 2, 8, 6, 18, 20, 26, 24, 1, 5, 7, 3, 19,
+                                               23, 25, 21, 9, 11, 17, 15, 12, 14, 10, 16, 4, 22, 14]}}
+
+# NOTE: Edge visualization of higher order elements are sketchy, see:
+# https://github.com/pyvista/pyvista/issues/947
+
+
+def _transpose(map):
```
michalhabera, line 30: A.k.a `numpy.argsort`

python/dolfinx/plot.py
```diff
@@ -0,0 +1,110 @@
...
+
+    """
+    transpose = np.zeros(len(map), dtype=np.int32)
+    for i in range(len(map)):
+        transpose[map[i]] = i
+    return transpose
+
+
+@functools.singledispatch
+def create_pyvista_topology(mesh: cpp.mesh.Mesh, dim: int, entities=None):
```
michalhabera, line 42: Could this be called `create_vtk_topology`? It uses VTK ordering, produces pure arrays, has nothing to do with pyvista per se.

garth-wells:
The plain VTK cell type identifiers should be used. By using the unusual "Arbitrary order Lagrange elements" for simple triangles, tets, etc, we hit all sorts of bugs in other libraries. itkwidgets does not work properly with arbitrary order.

jorgensd:
> The plain VTK cell type identifiers should be used. By using the unusual "Arbitrary order Lagrange elements" for simple triangles, tets, etc, we hit all sorts of bugs in other libraries. itkwidgets does not work properly with arbitrary order.

I have now added special handling of first order meshes and functions. 
I think we should submit issues to the libraries where this reveals issues.

Issue #1110, opened 2020-08-07
Boundary condition lifting does not respect domain makers


jorgensd:
Lifting for problems with subdomains have unclear behavior, as it only uses the kernel for the first integral:
https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfinx/fem/assemble_vector_impl.h#L174

Consider a 3x1 unit square mesh with quadrilateral, where the cells are marked 1,0,2, from left to right.
Assume that we have a boundary condition on the left and right hand side of the unit square. 
Assembling each of the domains 1 and 2 separately, and applying the boundary condition can only reproduce the expected behavior if there is only one cell integral present. See the minimal example below:

```
import numpy as np
import ufl
from mpi4py import MPI
from petsc4py import PETSc
import dolfinx
import dolfinx.io
def left_side(x):
    return x[0] < 1/3
def right_side(x):
    return x[0] > 2/3
mesh = dolfinx.UnitSquareMesh(
    MPI.COMM_WORLD, 3, 1, dolfinx.cpp.mesh.CellType.quadrilateral)
V = dolfinx.FunctionSpace(mesh, ("Lagrange", 1))

tdim = mesh.topology.dim
num_cells = mesh.topology.index_map(tdim).size_local
cell_midpoints = dolfinx.cpp.mesh.midpoints(mesh, tdim,
                                            range(num_cells))
values = []
for cell_index in range(num_cells):
    if left_side(cell_midpoints[cell_index]):
        values.append(1)
    elif right_side(cell_midpoints[cell_index]):
        values.append(2)
    else:
        values.append(0)
ct = dolfinx.mesh.MeshTags(mesh, mesh.topology.dim,
                           range(num_cells),
                           np.array(values, dtype=np.intc))

u = ufl.TrialFunction(V)
v = ufl.TestFunction(V)
x = ufl.SpatialCoordinate(mesh)
dx = ufl.Measure("dx", domain=mesh, subdomain_data=ct)
a = x[0]*ufl.inner(u, v)*dx(1) + 2*ufl.inner(u, v) * dx(2)
a1 = x[0]*ufl.inner(u, v)*dx(1)
a2 = 2*ufl.inner(u, v)*dx(2)
l = 2*ufl.inner(x[0], v)*dx(1) +\
    ufl.inner(x[1], v)*dx(2)
l1 = 2*ufl.inner(x[0], v)*dx(1)
l2 = ufl.inner(x[1], v)*dx(2)


def right(x):
[...]

PR #1177, opened 2020-10-22
Fix BC application

michalhabera:
This assembles boundary conditions to the RHS, i.e. `b = b - A*b`, where `A` is now the correct local element tensor, including contributions from all subdomains and interior facet integrals. Fixes a critical bug #1110 .

I've restructured the code a bit so it has the same pattern as for assemble functions, meaning there is high level `apply_lifting()` which handles BC for all subdomains of the mesh and then there are specific `_lift_bc_cells, _lift_bc_exterior_facet, _lift_bc_interior_facet` which runs on a given subset of active mesh entities.

Some code repetition appears which could be maybe refactored in the future.

### DIFF
--- a/cpp/dolfinx/fem/Form.h
+++ b/cpp/dolfinx/fem/Form.h
@@ -162,7 +162,7 @@ public:
   /// Get the function for 'kernel' for integral i of given
   /// type
   /// @param[in] type Integral type
-  /// @param[in] i Integral number
+  /// @param[in] i Domain index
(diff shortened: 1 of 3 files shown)


garth-wells (commented):
Could you add a test?

garth-wells (approved):

michalhabera:
Test added. The test on the current master would fail, https://github.com/FEniCS/dolfinx/runs/1296595659?check_suite_focus=true

PR #1179, opened 2020-10-24
Add env vars to control cffi_verbose and cffi_libraries

michalhabera:
This solves https://github.com/FEniCS/ffcx/issues/241 in a way that user who set `DOLFINX_JIT_CFLAGS="-Ofast -march=native"` could now also set `DOLFINX_JIT_CFFI_LIBRARIES="m"`.

It is also helpful to set `DOLFINX_JIT_CFFI_VERBOSE=1` and see how and when C compiler is called.


garth-wells:
I think `DOLFINX_JIT_CFFI_VERBOSE` is handy. For the libraries, is it possible for the user to have full control via the API? In most cases `DOLFINX_JIT_CFFI_LIBRARIES` shouldn't be preferred to the API approach if users have kernels that have library calls.

michalhabera:
@garth-wells Both are easily controllable via API, i.e. pass `Form(a, jit_parameters={"cffi_verbose": True})` (shown in `demo_elasticity.py`)

The point here is the following: When you set `DOLFNX_JIT_CFLAGS="-Ofast -march=native"` you need to explicitly link against math library (see https://github.com/FEniCS/ffcx/issues/241, compiler does some transformation which suddenly requires to call math.h function).
So I would like to avoid that your environmental variable setting requires changing the code and adding `jit_parameters={"cffi_libraries": ["m"]}` to each Form...

I completely agree that passing these via Form's API is preferable if user is linking his special libraries.

garth-wells:
See https://github.com/FEniCS/ffcx/issues/286 for the cause of the underlying bug with the math library linking.

garth-wells (approved):
How should we document these variable? We should add that they override settings via the API.

michalhabera:
Could docs be added to the docstring of `ffcx_jit` method? Unfortunately, generated docs doesn't seem rich, https://fenicsproject.org/docs/dolfinx/dev/python/_autogenerated/dolfinx.jit.html

garth-wells:
Maybe we should summarise relevant environment variables somewhere at the top level of the docs.

Issue #1171, opened 2020-10-21
Re-write pack_constants to use aligned memory


jhale:
FFCX produces kernels that can require aligned memory using the `assume_aligned` option. For example, 32 byte alignment might allow the compiler to emit AVX2 intrinsics to operate on the passed memory.

However, `fem::pack_constants` is returning an `Eigen::Array` that is constructed by an `Eigen::Map` of an unaligned `std::vector`. Passing this unaligned data to an aligned FFCX kernel can lead to crashes.

This issue is not showing up in the wild very often because `assume_aligned` is set so that FFCX generates C code that assumes no alignment on passed memory.

jhale:
Fixed in PR #1181.

Also necessary for all to work seamlessly is https://github.com/FEniCS/ffcx/pull/285

PR #1181, opened 2020-10-27
Pack constants into aligned Eigen array

jhale:
The user can now safely set `assume_aligned` in FFCX.

Fix for issue #1171.

### DIFF
--- a/cpp/dolfinx/fem/utils.h
+++ b/cpp/dolfinx/fem/utils.h
@@ -448,15 +448,30 @@ pack_coefficients(const fem::Form<T>& form)
 template <typename T>
 Eigen::Array<T, Eigen::Dynamic, 1> pack_constants(const fem::Form<T>& form)
 {
-  std::vector<T> constant_values;
-  for (auto& constant : form.constants())
+  const auto& constants = form.constants();
+
+  // Calculate size of array needed to store packed constants.
+  Eigen::Index size
+      = std::accumulate(constants.begin(), constants.end(), 0,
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/fem/utils.h
```diff
@@ -448,15 +448,30 @@ pack_coefficients(const fem::Form<T>& form)
...
+
+  // Pack constants.
+  Eigen::Array<T, Eigen::Dynamic, 1> constant_values(size);
+  Eigen::Index offset = 0;
+  for (const auto& constant : constants)
   {
-    const std::vector<T>& array = constant->value;
-    constant_values.insert(constant_values.end(), array.begin(), array.end());
+    const auto& value = constant->value;
+    const Eigen::Index value_size = value.size();
```
michalhabera, line 466: I think the below loop could be just
```c++
constant_values.segment(offset, value.size()) = value;
```

However constant data will never be performance critical, so whatever is more readable is preferred.

michalhabera (approved):

cpp/dolfinx/fem/utils.h
jhale: `value` is concretely a `std::vector` so it's not possible. We can discuss if the underlying storage of `function::Constant` should be an `Eigen::Array`, it seems a bit out of place with the current design.

michalhabera: Ah ok, I see. Anyway, not too important.

PR #1196, opened 2020-11-05
Added a padded bounding box and corresponding test.

jorgensd:
To be able to do efficient collision detection, it is quite useful to be able to work with a padded `BoundingBoxTree`. This means that we can identify if two elements are almost in contact without having to use GJK on a large number of elements.


cpp/dolfinx/geometry/BoundingBoxTree.cpp
```diff
@@ -20,7 +20,8 @@ namespace
 //-----------------------------------------------------------------------------
 // Compute bounding box of mesh entity
 Eigen::Array<double, 2, 3, Eigen::RowMajor>
-compute_bbox_of_entity(const mesh::Mesh& mesh, int dim, std::int32_t index)
+compute_bbox_of_entity(const mesh::Mesh& mesh, int dim, std::int32_t index,
+                       double padding)
```
chrisrichardson, line 24: Do we need (or want?) to add padding here? I think it is better to just say "compute_bbox_of_entity" returns the bbox of the entity, and then you can pad it later, if you want to.

cpp/dolfinx/geometry/BoundingBoxTree.cpp
```diff
@@ -250,8 +256,10 @@ BoundingBoxTree::BoundingBoxTree(const mesh::Mesh& mesh, int tdim) : _tdim(tdim)
   const std::int32_t num_leaves = map->size_local() + map->num_ghosts();
   Eigen::Array<double, Eigen::Dynamic, 3, Eigen::RowMajor> leaf_bboxes(
       2 * num_leaves, 3);
+
   for (int e = 0; e < num_leaves; ++e)
-    leaf_bboxes.block<2, 3>(2 * e, 0) = compute_bbox_of_entity(mesh, tdim, e);
+    leaf_bboxes.block<2, 3>(2 * e, 0)
+        = compute_bbox_of_entity(mesh, tdim, e, padding);
```
chrisrichardson, line 262: Padding could be added here instead of in `compute_bbox_of_entity`

cpp/dolfinx/geometry/BoundingBoxTree.cpp
```diff
@@ -293,8 +293,12 @@ BoundingBoxTree::BoundingBoxTree(
   Eigen::Array<double, Eigen::Dynamic, 3, Eigen::RowMajor> leaf_bboxes(
       2 * entity_indices.size(), 3);
   for (std::size_t i = 0; i < entity_indices_sorted.size(); ++i)
+  {
     leaf_bboxes.block<2, 3>(2 * i, 0)
-        = compute_bbox_of_entity(mesh, tdim, entity_indices_sorted[i], padding);
+        = compute_bbox_of_entity(mesh, tdim, entity_indices_sorted[i]);
+    leaf_bboxes.block<1, 3>(2 * i, 0) -= padding;
```
chrisrichardson, line 299: More concise as `leaf_bboxes.row(2*i)` ?

PR #1206, opened 2020-11-11
Fix broken code for cell_normals and make it more efficient by using entitiy lists

jorgensd:
Current master is broken, as the following test:
```

def test_normals(cube, square):
    """ Test that cell normals of a set of facets """
    def left_side(x):
        return numpy.isclose(x[0], 0)
    fdim = cube.topology.dim - 1
    facets = locate_entities_boundary(cube, fdim, left_side)
    normals = cell_normals(cube, fdim)
    assert(numpy.isclose(normals[facets], [-1, 0, 0]).all(axis=1).all())

    fdim = square.topology.dim - 1
    facets = locate_entities_boundary(square, fdim, left_side)
    normals = cell_normals(square, fdim)
    assert(numpy.isclose(normals[facets], [-1, 0, 0]).all(axis=1).all())
```
would return gibberish if we exposed `cell_normals` to the python layer. This is because it is missing the topology->geometry mapping, which was created way back when we separated mesh topology and geometry.

This pull request adds in this mapping, and exploits the fact that in most cases one is only interested in a subset of facets.

### DIFF
--- a/python/test/unit/mesh/test_face.py
+++ b/python/test/unit/mesh/test_face.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2011 Garth N. Wells
+# Copyright (C) 2020 Garth N. Wells, Jørgen S. Dokken
 #
 # This file is part of DOLFINX (https://www.fenicsproject.org)
 #
@@ -6,7 +6,10 @@
 
 import dolfinx
 import pytest
+import numpy
 from dolfinx import UnitCubeMesh, UnitSquareMesh
+from dolfinx.mesh import locate_entities_boundary
+from dolfinx.cpp.mesh import cell_normals
 from dolfinx_utils.test.skips import skip_in_parallel
 from mpi4py import MPI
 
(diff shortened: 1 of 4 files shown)


python/test/unit/mesh/test_face.py
```diff
@@ -36,3 +39,18 @@ def test_area(cube, square):
     cube.topology.create_entities(1)
     area = dolfinx.cpp.mesh.volume_entities(square, range(num_faces), 2).sum()
     assert area == pytest.approx(1.0)
+
+
+def test_normals(cube, square):
+    """ Test that cell normals of a set of facets """
```
chrisrichardson, line 45: typo "test that"

python/test/unit/mesh/test_face.py
```diff
@@ -36,3 +39,18 @@ def test_area(cube, square):
...
+    def left_side(x):
+        return numpy.isclose(x[0], 0)
+    fdim = cube.topology.dim - 1
+    facets = locate_entities_boundary(cube, fdim, left_side)
+    normals = cell_normals(cube, fdim, facets)
+    assert(numpy.isclose(normals, [-1, 0, 0]).all(axis=1).all())
```
chrisrichardson, line 51: Can use `numpy.allclose()`

chrisrichardson (approved):

PR #1207, opened 2020-11-12
give priority to cray pkgconfig files when searching for PETSc, SLEPc

drew-parsons:
Following on from PR#1195, when looking for PETSc and SLEPC in pkg_search_module, pkgconfig priority was previously given to the cray version of the pkgconfig files. This patch restores that priority, after including lowercase petsc and slepc as alternative pkg names.


drew-parsons:
Not sure how effective the craypetsc_real and crayslepc_real pkg names actually are.  pkg-config operation on CRAY systems is a nightmare, doesn't work cleanly. But should keep the principle of priority the same as it was before in any case, putting the cray pkg names first.

PR #1211, opened 2020-11-13
Function move

### LINKED ISSUE #1210: Class Function is not trivially movable
Currently, the `Function` class has defaulted move constructors and move assignment operators.
However, the class stores an object `Vec _petsc_vector`, which is a pointer. This means that, when calling `Function(Function&& other)`, `other._petsc_vector` needs to be zeroed or the memory it points to will be released when `other` goes out of scope.

massimiliano-leoni:
Closes #1210 .

Writing explicit move constructor and move assignment operator for class Function, zeroing the stored `Vec` data member.

### DIFF
--- a/cpp/dolfinx/function/Function.h
+++ b/cpp/dolfinx/function/Function.h
@@ -101,6 +101,8 @@ public:
     _id = std::move(v._id);
     _function_space = std::move(v._function_space);
     _x = std::move(v._x);
+    if (_petsc_vector)
+      VecDestroy(&_petsc_vector);
     _petsc_vector = std::move(v._petsc_vector);
     v._petsc_vector = nullptr;
   }
--- a/cpp/dolfinx/function/Function.h
+++ b/cpp/dolfinx/function/Function.h
@@ -79,7 +79,7 @@ public:
   Function(const Function& v) = delete;
 
   /// Move constructor
-  Function(Function&& v)
+  Function(Function&& v) noexcept
       : name(std::move(v.name)), _id(std::move(v._id)),
         _function_space(std::move(v._function_space)), _x(std::move(v._x)),
         _petsc_vector(std::move(v._petsc_vector))
--- a/cpp/dolfinx/function/Function.h
+++ b/cpp/dolfinx/function/Function.h
@@ -97,16 +97,15 @@ public:
   /// Move assignment
   Function& operator=(Function&& v)
   {
-    if (this != &v)
+    name = std::move(v.name);
+    _id = std::move(v._id);
+    _function_space = std::move(v._function_space);
+    _x = std::move(v._x);
+    if (_petsc_vector)
     {
-      name = std::move(v.name);
-      _id = std::move(v._id);
-      _function_space = std::move(v._function_space);
-      _x = std::move(v._x);
-      if (_petsc_vector)
-        VecDestroy(&_petsc_vector);
-      _petsc_vector = std::exchange(v._petsc_vector, nullptr);
+      VecDestroy(&_petsc_vector);
     }
+    _petsc_vector = std::exchange(v._petsc_vector, nullptr);
 
     return *this;
   }
--- a/cpp/dolfinx/function/Function.h
+++ b/cpp/dolfinx/function/Function.h
@@ -79,7 +79,13 @@ public:
   Function(const Function& v) = delete;
 
   /// Move constructor
-  Function(Function&& v) = default;
+  Function(Function&& v)
+      : name(std::move(v.name)), _id(std::move(v._id)),
+        _function_space(std::move(v._function_space)), _x(std::move(v._x)),
+        _petsc_vector(std::move(v._petsc_vector))
+  {
+    v._petsc_vector = nullptr;
+  }
 
   /// Destructor
   virtual ~Function()
@@ -89,7 +95,15 @@ public:
   }
 
   /// Move assignment
-  Function& operator=(Function&& v) = default;
+  Function& operator=(Function&& v)
+  {
+    name = std::move(v.name);
+    _id = std::move(v._id);
+    _function_space = std::move(v._function_space);
+    _x = std::move(v._x);
+    _petsc_vector = std::move(v._petsc_vector);
+    v._petsc_vector = nullptr;
+  }
 
   // Assignment
   Function& operator=(const Function& v) = delete;
--- a/cpp/dolfinx/function/Function.h
+++ b/cpp/dolfinx/function/Function.h
@@ -95,17 +95,13 @@ public:
   }
 
   /// Move assignment
-  Function& operator=(Function&& v)
+  Function& operator=(Function&& v) noexcept
   {
     name = std::move(v.name);
     _id = std::move(v._id);
     _function_space = std::move(v._function_space);
     _x = std::move(v._x);
-    if (_petsc_vector)
-    {
-      VecDestroy(&_petsc_vector);
-    }
-    _petsc_vector = std::exchange(v._petsc_vector, nullptr);
+    std::swap(_petsc_vector, v._petsc_vector);
 
     return *this;
   }
--- a/cpp/dolfinx/function/Function.h
+++ b/cpp/dolfinx/function/Function.h
@@ -97,13 +97,16 @@ public:
   /// Move assignment
   Function& operator=(Function&& v)
   {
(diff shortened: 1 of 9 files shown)


chrisrichardson (approved):
I think this is fine.

cpp/dolfinx/function/Function.h
```diff
@@ -79,7 +79,13 @@ class Function
   Function(const Function& v) = delete;
 
   /// Move constructor
-  Function(Function&& v) = default;
+  Function(Function&& v) noexcept
+      : name(std::move(v.name)), _id(std::move(v._id)),
+        _function_space(std::move(v._function_space)), _x(std::move(v._x)),
+        _petsc_vector(std::move(v._petsc_vector))
+  {
+    v._petsc_vector = nullptr;
```
garth-wells, line 87: I think we can use `std::exchange` to do everything in the initialiser list.

cpp/dolfinx/function/Function.h
```diff
@@ -89,7 +95,15 @@ class Function
...
 
   /// Move assignment
-  Function& operator=(Function&& v) = default;
+  Function& operator=(Function&& v)
+  {
+    name = std::move(v.name);
+    _id = std::move(v._id);
+    _function_space = std::move(v._function_space);
+    _x = std::move(v._x);
+    _petsc_vector = std::move(v._petsc_vector);
```
garth-wells, line 104: Use `std::exchange` to compress two lines into one.

garth-wells (changes requested):
Looks good - just two small comments to compress some code.

massimiliano-leoni:
In the move assignment operator I understand that we need to clean the current vector before replacing it, or it'll leak. Is that correct?

cpp/dolfinx/function/Function.h
```diff
@@ -89,7 +95,21 @@ class Function
...
-  Function& operator=(Function&& v) = default;
+  Function& operator=(Function&& v)
+  {
+    if (this != &v)
+    {
+      name = std::move(v.name);
+      _id = std::move(v._id);
+      _function_space = std::move(v._function_space);
+      _x = std::move(v._x);
+      if (_petsc_vector)
```
garth-wells, line 106:
```
if (_petsc_vector)
        VecDestroy(&_petsc_vector);
```
isn't required. Ownership is just being transferred.

garth-wells (changes requested):

cpp/dolfinx/function/Function.h
massimiliano-leoni: Shouldn't we clean up the memory the destination function held before the assignment?

massimiliano-leoni: I mean, in case we have something like

```
// Both functions store data
u = function::Function<PetscScalar>(V);
v = function::Function<PetscScalar>(V);

u = std::move(v);
```
Then we should free the memory pointed to by `u._petsc_vector` before acquiring ownership.

jhale: Isn't this dealt with by the destructor and the RAII idiom in the usual way?

massimiliano-leoni: That data member is stored as a plain pointer.

cpp/dolfinx/function/Function.h
```diff
@@ -89,7 +95,21 @@ class Function
   }
 
   /// Move assignment
-  Function& operator=(Function&& v) = default;
+  Function& operator=(Function&& v)
+  {
+    if (this != &v)
```
michalhabera, line 100: Why are you checking this? Self-move-assignment will be most likely undefined behaviour in all std lib moves below anyway.

michalhabera (commented):
Up to self-move assignment check looks ok.

cpp/dolfinx/function/Function.h
massimiliano-leoni: It's a standard idiom of C++ that assignment operators should be safe with respect to self-assignment. In the case of move assignment there's no consensus: cppreference has snippets where it is used [e.g. [in a note here](https://en.cppreference.com/w/cpp/utility/exchange)] whereas isocpp mentions that it shouldn't be an issue. I thought I added it, "better safe than sorry".

cpp/dolfinx/function/Function.h
garth-wells: Yes, it should be cleaned up by RAII in the `Function` destructor. What needs to be done here is `v._petsc_vector` needs to be set to `nullptr` is there is nothing to cleanup. The simplest fix is probably to use `std::swap` (or remove PETSc!).

cpp/dolfinx/function/Function.h
garth-wells: I'm not super keen on the check. One needs to do really weird (and bad) things to trigger it.

cpp/dolfinx/function/Function.h
massimiliano-leoni: I verified with valgrind that there is, in fact, a memory leak without a call to `VecDestroy`.

michalhabera: I think @massimiliano-leoni is right, after `u = std::move(v)` the old `u._petsc_vector` got lost and no-one is responsible for it.

cpp/dolfinx/function/Function.h
michalhabera: It feels very weird to me to handle this, and makes the code ugly as well. Maybe just add and assert() if you want to be safe.

cpp/dolfinx/function/Function.h
garth-wells: What's needed is replacing `std::exchange` with `std::swap`.

michalhabera: So you suggest to swap `u._petsc_vector` and `v._petsc_vector` such that after `u = std::move(v)` the `v` will be managing the old u’s vector? That would work as well to avoid leaks, but I do not undertstand why bind these two together. I would expect after move assigning `v` into `u` the old u’s vector be freed. It could even happen that someone stores a reference to `v` so it won’t get freed.

jhale: Amazing @massimiliano-leoni, I learned something new. Thanks for running valgrind on it. C++ move semantics are so transparent and easy to understand ;).

garth-wells: @michalhabera a couple of points why destruction should be left the destructor:
1. The move assignment should just 'move', i.e. be essentially zero ops. Cleaning up memory in a move assignment is counter-intuitive
2. Destruction should take place in the destructor
3. Adding `VecDestroy` to the move assignment is code duplication - the step is already performed by `Function::~Function`

michalhabera: @massimiliano-leoni Giving it some thoughts I agree with Garth on doing `std::swap(u._petsc_vector, v._petsc_vector)` and not calling `VecDestroy` inside the move assignment.

massimiliano-leoni: @jhale I know the feeling!

@garth-wells , @michalhabera there are notable proponents of both implementations, and I'm fine with either, so I changed it to the one you prefer.

cpp/dolfinx/function/Function.h
massimiliano-leoni: I removed the check.

mrambausek: Consider some function that consumes a Function and either returns a new one or just forwards the given Function if no new Function is needed for whatever reason. Assign the result to the input, .e.g.
```
Function myfunc;
...
myfunc = modify_or_new(myfunc, other_args);
```
I believe, that there might be valid reasons for such or at least similar a constructions. A "practical" example might be
to append elements to a vector with std::accumulate instead of using std::transform. I have working code that does that so I would say that std containers support self-move assignment.

michalhabera: With the latest version of the code self-move-assignment of Function is possible as far as all std lib implementations will allow it.

mrambausek: Right, I forgot about the code that followed the check. Should behave correctly as it is. Beyond that it's probably just a matter of taste.

michalhabera (approved):

garth-wells (approved):

PR #1214, opened 2020-11-16
Mesh refinement code issues

jhale:
Currently contains a test that fails but shouldn't


jhale:
Ignore this, 'user' error. I will improve/correct the documentation.

chrisrichardson:
I think some changes happened to the interpretation of the markers when it transitioned from `MeshFunction` to `MeshTags`. Currently the values of the markers are ingored, and it just looks at which tags exist.

PR #1221, opened 2020-11-24
Fix igi_to_vertex mapping in xmdf_utils::extract_local_entities

### LINKED ISSUE #1220: Global geometry mapping to local index mapping acessing non-existing indices
In the `igi_to_vertex`-mapping in https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfinx/io/xdmf_utils.cpp#L544-L546, indices not existing in the map is accessed when creating the local entity map. This does not cause an error, as calling an non-initialized key returns 0. 

This causes additional entries in the local_entities sent into a MeshTag, and can cause additional entries to be tagged in the 
`MeshTag` in parallel

jorgensd:
Fixes #1220

### DIFF
--- a/cpp/dolfinx/io/xdmf_utils.cpp
+++ b/cpp/dolfinx/io/xdmf_utils.cpp
@@ -528,7 +528,9 @@ xdmf_utils::extract_local_entities(
   }
 
   std::vector<std::int32_t> entities_new;
+  entities_new.reserve(recv_ents.array().size());
   std::vector<std::int32_t> values_new;
+  values_new.reserve(recv_vals.array().size());
 
   for (Eigen::Index e = 0; e < recv_ents.array().rows() / num_vertices_per_entity; ++e)
   {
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/io/xdmf_utils.cpp
```diff
@@ -527,25 +527,40 @@ xdmf_utils::extract_local_entities(
...
+  entities_new.reserve(recv_ents.array().size());
+  std::vector<std::int32_t> values_new;
+  values_new.reserve(recv_vals.array().size());
 
-  for (Eigen::Index e = 0; e < entities_local.rows(); ++e)
+  for (Eigen::Index e = 0; e < recv_ents.array().rows() / num_vertices_per_entity; ++e)
   {
-    for (Eigen::Index i = 0; i < entities_local.cols(); ++i)
+    bool entity_found = true;
+    std::vector<std::int32_t> entities(num_vertices_per_entity);
```
michalhabera, line 538: This should be named entity, it only stores one entity.

cpp/dolfinx/io/xdmf_utils.cpp
```diff
@@ -527,25 +527,40 @@ xdmf_utils::extract_local_entities(
...
+    // First entity is guaranteed to be on processor
+    entities[0] = igi_to_vertex[recv_ents.array()[e * num_vertices_per_entity]];
```
michalhabera, line 540: No need to separate this as a special case, there is no gain.

michalhabera (approved):

PR #1222, opened 2020-11-25
Change dof finder return type for blocked degrees of freedom support

garth-wells:
This small change is for `locate_dofs_topological` and `locate_dofs_geometrical` to return two 1D arrays rather than one 2D array. This is preparatory work for supporting blocked degree-of-freedom maps, in which case the two returns arrays can possibly have different lengths.


garth-wells:
Superseded by #1225.

PR #1223, opened 2020-11-25
Fix local writing of meshtags

jorgensd:
Currently `XDMFFile.write_meshtags` writes ghost entities, which results in major duplication + visualization issues.
This PR changes `write_meshtags` such that it only writes owned entities (instead of owned + ghosted entities).

MWE:
```
from dolfinx import UnitCubeMesh, MeshTags
from mpi4py import MPI
import numpy as np

comm_world = MPI.COMM_WORLD
comm_self = MPI.COMM_SELF

if comm_world.rank == 0:
    mesh = UnitCubeMesh(comm_self, 10, 10, 10)
    mesh.name = "mesh"
    num_cells = mesh.topology.index_map(mesh.topology.dim).size_local
    indices = np.arange(num_cells, dtype=np.int32)
    values = np.ones(num_cells, dtype=np.int32)
    mt = MeshTags(mesh, mesh.topology.dim, indices, values)
    mt.name = "cells"
    with XDMFFile(comm_self, "mesh_serial.xdmf", "w") as xdmf:
        xdmf.write_mesh(mesh)
        xdmf.write_meshtags(mt)
    print("Serial indices:", len(mt.indices))
comm_world.barrier()


with XDMFFile(comm_world, "mesh_serial.xdmf", "r") as xdmf:
    mesh2 = xdmf.read_mesh(name="mesh")
    mt2 = xdmf.read_meshtags(mesh2, name="cells")
    print("Parallel indices:", len(mt2.indices))

with XDMFFile(comm_world, "mesh_parallel.xdmf", "w") as xdmf:
    xdmf.write_mesh(mesh2)
    xdmf.write_meshtags(mt2)
```

### DIFF
--- a/cpp/dolfinx/io/xdmf_meshtags.h
+++ b/cpp/dolfinx/io/xdmf_meshtags.h
@@ -38,21 +38,16 @@ void add_meshtags(MPI_Comm comm, const mesh::MeshTags<T>& meshtags,
   const std::vector<std::int32_t>& active_entities = meshtags.indices();
   const std::vector<T>& values = meshtags.values();
 
-  // Remove ghost indices (note that indices are sorted)
-  std::int32_t num_entities = 0;
-  const std::int32_t num_active_entities = active_entities.size();
-  for (std::int32_t i = 0; i < num_active_entities; ++i)
-  {
-    if (active_entities[i] < num_local_entities)
-      ++num_entities;
-    else
-      break;
-  }
+  // Find number of tagged entities in local range
+  const int num_active_entities
+      = std::lower_bound(active_entities.begin(), active_entities.end(),
+                         num_local_entities)
+        - active_entities.begin();
 
   const std::vector<std::int32_t> local_entities(
-      active_entities.begin(), active_entities.begin() + num_entities);
+      active_entities.begin(), active_entities.begin() + num_active_entities);
   const std::vector<T> local_values(values.begin(),
-                                    values.begin() + num_entities);
+                                    values.begin() + num_active_entities);
   const std::string path_prefix = "/MeshTags/" + name;
   xdmf_mesh::add_topology_data(comm, xml_node, h5_id, path_prefix,
                                mesh->topology(), mesh->geometry(), dim,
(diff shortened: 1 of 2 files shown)


cpp/dolfinx/io/xdmf_meshtags.h
```diff
@@ -32,11 +32,27 @@ void add_meshtags(MPI_Comm comm, const mesh::MeshTags<T>& meshtags,
   assert(meshtags.mesh());
   std::shared_ptr<const mesh::Mesh> mesh = meshtags.mesh();
   const int dim = meshtags.dim();
+  const int local_range = mesh->topology().index_map(dim)->size_local();
   const std::vector<std::int32_t>& active_entities = meshtags.indices();
+  const std::vector<std::int32_t>& values = meshtags.values();
+
+  // Remove ghost indices (note that indices are sorted)
+  int num_local_entities = 0;
```
michalhabera, line 40: You can find it without the loop,
```c++
const int num_local_entities = std::lower_bound(active_entities.begin(), active_entities.end(), local_range) - active_entities.begin();
```

Just note below you are copying both vectors into another vectors, just in order to cut their tails. Maybe there is a clever way how to avoid this copy. In theory you can prepare Eigen::Map and map only the active part, and pass just the eigen map to add_topology_data. But it would need API change there.

Any better ideas @garth-wells ?

garth-wells: Could just pass begin/end iterators. Agree that copying a big chuck of the vector isn't very nice.

garth-wells:
The PR description gives a MWE, but it doesn't summarise what it changes it introduces.

garth-wells:
Closed by mistake . . .

jorgensd:
> The PR description gives a MWE, but it doesn't summarise what it changes it introduces.

Summary added

cpp/dolfinx/io/xdmf_meshtags.h
jorgensd: Any suggestions on how we could change:
```
xdmf_utils::add_data_item(attribute_node, h5_id, path_prefix + "/Values",
                            local_values, offset, {global_num_values, 1}, "",
                            use_mpi_io);
```
to use iterators? As this is a templated function, changing it to iterators is not straightforward as far as I can tell.

michalhabera: @jorgensd @garth-wells I've made an attempt with Eigen::Map, see https://github.com/FEniCS/dolfinx/tree/michal/write_owned_entities

@jorgensd If you are happy with this then just extend some existing test, so it would catch the issue being addressed here.

jorgensd:
Superseded by #1228

PR #1226, opened 2020-11-27
Refinement updates

chrisrichardson:
* Updates to mesh refinement methods, allowing refinement in parallel with any ghost mode and no repartitioning
* Introduces a `std::function` interface as the partitioner in `Mesh::create_mesh`
* delete 100+ lines of code

### DIFF
--- a/cpp/dolfinx/io/xdmf_utils.cpp
+++ b/cpp/dolfinx/io/xdmf_utils.cpp
@@ -527,25 +527,38 @@ xdmf_utils::extract_local_entities(
       igi_to_vertex[nodes_g[x_dofs[cell_vertex_dofs[v]]]] = vertices[v];
   }
 
-  // Apply map and obtain entities defined with local vertex numbers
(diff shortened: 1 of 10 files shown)


IgorBaratta (approved):
It looks fine to me. 
This interface for partitioning seems flexible.

A minor question is, couldn't we capture the `index_map` to get the ghost owners [here](https://github.com/FEniCS/dolfinx/pull/1226/files#diff-7e60b3ff46629dee99093c159ae6a27a8742ab64d87a45e4d44f0522f50df74dR339)??

garth-wells:
Following on from @IgorBaratta's comment, we should avoid the non-scalable MPI_Allgather at https://github.com/FEniCS/dolfinx/pull/1226/files#diff-7e60b3ff46629dee99093c159ae6a27a8742ab64d87a45e4d44f0522f50df74dR339.

chrisrichardson:
I agree. That piece of code has an attached FIXME, because it is essentially reverse engineering some data from the GraphBuilder (ownership of ghost nodes in the graph). I am hoping to remove it all in a future PR, but I wanted to do so in manageable chunks.

PR #1228, opened 2020-11-28
XDMFFile.write_meshtags for owned entities

jorgensd:
Improved version of #1223, writing only owned entities of mesh tags to file to avoid duplication in parallel.

### DIFF
--- a/python/test/unit/io/test_xdmf_meshdata.py
+++ b/python/test/unit/io/test_xdmf_meshdata.py
@@ -1,15 +1,13 @@
-# Copyright (C) 2012-2020 Garth N. Wells, Jørgen S. Dokken
+# Copyright (C) 2012-2019 Garth N. Wells
 #
 # This file is part of DOLFINX (https://www.fenicsproject.org)
 #
 # SPDX-License-Identifier:    LGPL-3.0-or-later
 
 import os
-from xml.etree import ElementTree
 
-import numpy as np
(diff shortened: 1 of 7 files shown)


python/test/unit/io/test_xdmf_meshdata.py
```diff
@@ -62,3 +64,44 @@ def test_read_mesh_data(tempdir, tdim, n):
     assert cell_type[1] == 1
     assert mesh.topology.index_map(tdim).size_global == mesh.mpi_comm().allreduce(cells.shape[0], op=MPI.SUM)
     assert mesh.geometry.index_map().size_global == mesh.mpi_comm().allreduce(x.shape[0], op=MPI.SUM)
+
+
+def test_write_mesh_data(tempdir):
```
michalhabera, line 69: There is already a test in `test_xdmf_meshtags.py`. I think you just need to add that XMLParser and check that number of entities in the XML file is the same as owned entities for the MeshTags.

jorgensd: > There is already a test in `test_xdmf_meshtags.py`. I think you just need to add that XMLParser and check that number of entities in the XML file is the same as owned entities for the MeshTags.
Test updated

michalhabera (approved):

garth-wells (approved):

PR #1230, opened 2020-11-30
Remove redundant wrapper

michalhabera:
Importing cpp into python public layer instead. Less code to maintain.


jorgensd (approved):
The only issue I see with this is that we suddenly have no documentation of the usage in the Python layer. But that applies to most functions exposed directly through pybind. For instance the last argument in `select_colliding_cells` deserves an docstring.

michalhabera:
I agree, it is a general issue though.

jorgensd:
> I agree, it is a general issue though.

We could maybe use pybind11_mkdoc or Binder? See: https://github.com/pybind/pybind11/issues/1626

michalhabera:
Here is the valid link https://github.com/pybind/pybind11_mkdoc. I would love to see something like this in action.

jorgensd:
> Here is the valid link https://github.com/pybind/pybind11_mkdoc. I would love to see something like this in action.

I had a look at pybind11_mkdoc today. I am not convinced. It is easy to generate the headers, but the parser does not create newlines, thus the doc:
```
  /// Create contact constraint
  ///
  /// @param[in] V The function space
  /// @param[in] slaves List of local slave dofs
  /// @param[in] slaves_cells List of local slave cells
  /// @param[in] offsets Offsets for local slave cells
```
Became:
```
In [3]: mpc.MultiPointConstraint??
Init docstring:
__init__(self: dolfinx_mpc.cpp.mpc.MultiPointConstraint, arg0: dolfinx::function::FunctionSpace, arg1: numpy.ndarray[numpy.int32[m, 1]], arg2: int) -> None

Create contact constraint

@param[in] V The function space @param[in] slaves List of local slave
dofs @param[in] slaves_cells List of local slave cells @param[in]
offsets Offsets for local slave cells
```

michalhabera:
@jorgensd Can you make an issue for better pybind11 docs if doesn't exist? I'd like to merge&close this PR.

jorgensd:
> @jorgensd Can you make an issue for better pybind11 docs if doesn't exist? I'd like to merge&close this PR.

See: #1236

PR #1231, opened 2020-12-01
Remove scotch header from dolfinx public api

IgorBaratta:
PT-SCOTCH is not part of the public interface, so the header (_ptscotch.h_) should not be included in any h file.
This PR solves potential conflicts with parmetis (redefinition of `idx_t`).

The same can be done with parmetis, if the PR is merged.

### DIFF
--- a/.github/workflows/docker-end-user.yml
+++ b/.github/workflows/docker-end-user.yml
@@ -47,7 +47,7 @@ jobs:
           docker pull dolfinx/dev-env
       - name: Set default FFCX parameters
         run: |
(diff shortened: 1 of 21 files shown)


garth-wells:
I'd suggest a simpler fix of just copying the graph is the graph int type and the SCOTCH int type are different. This avoids messy with compiler definitions and the preprocessor. See https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfinx/function/Function.h#L174 for an example of using C++ rather than the preprocessor to make type decisions.

garth-wells (approved):

chrisrichardson (approved):

PR #1234, opened 2020-12-01
Restyle move operators

### LINKED ISSUE #1233: Restyle some move constructors
Similarly to what was done in #1211 , the move operators in `PETScVector` and `PETScOperator` can be restyled.

massimiliano-leoni:
Closes #1233 .

Inspired by #1211 , this is not a bug fix but just a restyling for PETScVector and PETScOperator.

### DIFF
--- a/cpp/dolfinx/la/PETScOperator.cpp
+++ b/cpp/dolfinx/la/PETScOperator.cpp
@@ -21,7 +21,7 @@ PETScOperator::PETScOperator(Mat A, bool inc_ref_count) : _matA(A)
     PetscObjectReference((PetscObject)_matA);
 }
 //-----------------------------------------------------------------------------
-PETScOperator::PETScOperator(PETScOperator&& A) noexcept
+PETScOperator::PETScOperator(PETScOperator&& A)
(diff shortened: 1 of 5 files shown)


cpp/dolfinx/la/PETScOperator.cpp
```diff
@@ -21,10 +21,9 @@ PETScOperator::PETScOperator(Mat A, bool inc_ref_count) : _matA(A)
     PetscObjectReference((PetscObject)_matA);
 }
 //-----------------------------------------------------------------------------
-PETScOperator::PETScOperator(PETScOperator&& A) noexcept : _matA(nullptr)
+PETScOperator::PETScOperator(PETScOperator&& A) noexcept
```
garth-wells, line 24: I don't think `std::exchange` is `noexcept`.

massimiliano-leoni: Right, sorry about that. Interestingly enough, cppreference uses it in a `noexcept` function, despite not marking it `noexcept` [cf. https://en.cppreference.com/w/cpp/utility/exchange ].

garth-wells (approved):

PR #1235, opened 2020-12-02
Allow control of non-affine NR iterations

michalhabera:
Some practitioner's meshes fail to do any IO, since it is calling eval() which needs physical-to-reference map. This small change allows users to do
```python
mesh.geometry.cmap.non_affine_atol = 1.0e-8
mesh.geometry.cmap.non_affine_max_its = 20
```
(or whatever works) before the respective NR calls.

### DIFF
--- a/cpp/dolfinx/fem/CoordinateElement.h
+++ b/cpp/dolfinx/fem/CoordinateElement.h
@@ -57,6 +57,12 @@ public:
   /// Return the dof layout
   const ElementDofLayout& dof_layout() const;
 
+  /// Absolute increment stopping criterium for non-affine Newton solver
+  double non_affine_atol = 1.0e-8;
+
+  /// Maximum number of iterations for non-affine Newton solver
+  int non_affine_max_its = 10;
+
   /// Compute physical coordinates x for points X  in the reference
   /// configuration
   /// @param[in,out] x The physical coordinates of the reference points X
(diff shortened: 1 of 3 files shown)


cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -57,6 +57,12 @@ class CoordinateElement
   /// Return the dof layout
   const ElementDofLayout& dof_layout() const;
 
+  /// Absolute increment stopping criterium for non-affine Newton solver
+  double non_affine_atol = 1.0e-8;
```
garth-wells, line 61: I'd prefer to see these passed at arguments to the relevant function and keep `CoordinateElement` stateless.

michalhabera: `CoordinateElement::compute_reference_geometry` is all based on reading the state of underlying CoordinateElement.
I though about the pure API approach too, but this means:
1. passing these parameters into `write_function()`, yuck, or
2. breaking the IO interface to work on lower level structures.

Probably the better approach would be 2., but it is an incomparable amount of work. See this PR as a hotfix for users to use any real-world hex meshes in IO.

michalhabera: I can add the `FIXME` note if that helps?

garth-wells: This feels like a top down (wrong) rather than a bottom-up fix (right). I don't mind seeing it merged now if it gets fixed later. It does seem to come back to the issue of magic interpolation behind the scenes with file output for visualisation.

michalhabera: I agree with you. I can alternatively just tweak the current hardcoded tolerance and number of iterations to be more robust?

garth-wells: See issue #1245.

PR #1246, opened 2020-12-10
Add option for plotting streamlines

lockyluc:
Vector fields can currently only be plotted as quiver plots. 
However, matplotlib has a utility for creating streamlines as well. [1]
I propose to add a keyword `style` in this case to use either plotting style.

```python
# Minimal Example of how to use
from fenics import *
import matplotlib.pyplot as plt
mesh = UnitSquareMesh(10,10)
V = FunctionSpace(mesh,'P',1)
u = Expression("x[0]+x[1]", degree=1)
v = interpolate(u,V)
plt.subplot('211') 
plot(grad(v), style='quiver')
plt.subplot('212')
plot(grad(v), style='streamline')
plt.show()
```

[1] https://matplotlib.org/3.3.3/gallery/images_contours_and_fields/plot_streamplot.html


jorgensd:
@lockyluc Make sure that your changes pass the CI tests:
https://github.com/FEniCS/dolfinx/runs/1533223104 
In this case you need to follow the flake8 formatting style

jorgensd:
Matplotlib is being deprecated in favor of pyvista, see  #1161

PR #1247, opened 2020-12-10
Graphbuilder checks

chrisrichardson:
This PR updates the dual graph builder, which processes raw mesh topology before partitioning.
* Check for error conditions (e.g. >2 cells sharing a facet)
* Use AdjacencyList more
* Provide better load balancing of the global match phase (speeds up overall computation)

### DIFF
--- a/cpp/dolfinx/mesh/GraphBuilder.cpp
+++ b/cpp/dolfinx/mesh/GraphBuilder.cpp
@@ -151,11 +151,6 @@ compute_nonlocal_dual_graph(
   LOG(INFO) << "Build nonlocal part of mesh dual graph";
   common::Timer timer("Compute non-local part of mesh dual graph");
 
-  // Get cell offset for this process and add to local graph
-  const std::int32_t num_local_cells = cell_vertices.num_nodes();
-  const std::int64_t cell_offset
-      = dolfinx::MPI::global_offset(mpi_comm, num_local_cells, true);
-
   // Get number of MPI processes, and return if mesh is not distributed
   const int num_processes = dolfinx::MPI::size(mpi_comm);
   if (num_processes == 1)
@@ -166,16 +161,9 @@ compute_nonlocal_dual_graph(
             local_graph.array().cast<std::int64_t>(), local_graph.offsets()),
         0, 0};
   }
+
   // At this stage facet_cell map only contains facets->cells with edge
   // facets either interprocess or external boundaries
-
-  // List of cell vertices
-  const int tdim = mesh::cell_dim(cell_type);
-  const int num_vertices_per_facet
-      = mesh::num_cell_vertices(mesh::cell_entity_type(cell_type, tdim - 1));
-
-  // Compute local edges (cell-cell connections) using global (internal
-  // to this function, not the user numbering) numbering
 
   // Find the global range of the first vertex index of each facet in the list
   // and use this to divide up the facets between all processes.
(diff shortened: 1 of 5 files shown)


cpp/dolfinx/mesh/GraphBuilder.cpp
```diff
@@ -130,138 +131,154 @@ compute_local_dual_graph_keyed(
...
-  }
-
   // Get number of MPI processes, and return if mesh is not distributed
   const int num_processes = dolfinx::MPI::size(mpi_comm);
   if (num_processes == 1)
-    return {graph, 0, 0};
+  {
+    // Convert graph to int64
+    return {
+        graph::AdjacencyList<std::int64_t>(
```
garth-wells, line 160: Does this need a `std::move` to avoid a copy, or will it be ok?

cpp/dolfinx/mesh/GraphBuilder.cpp
```diff
@@ -130,138 +131,154 @@ compute_local_dual_graph_keyed(
...
-    for (int i = 0; i < data_p.rows(); i += (num_vertices_per_facet + 1))
-    {
-      // Build map key
-      std::copy(data_p.data() + i, data_p.data() + i + num_vertices_per_facet,
-                key.first.begin());
-      key.second.first = p;
-      key.second.second = data_p[i + num_vertices_per_facet];
+    for (int j = received_buffer_offsets[p] / (num_vertices_per_facet + 1);
+         j < received_buffer_offsets[p + 1] / (num_vertices_per_facet + 1); ++j)
+      proc[j] = p;
```
garth-wells, line 235: Add braces.

cpp/dolfinx/mesh/GraphBuilder.cpp
```diff
@@ -130,138 +131,154 @@ compute_local_dual_graph_keyed(
...
+  }
+
+  // Get permutation that takes facets into sorted order
+  std::vector<int> perm(num_facets);
+  std::iota(perm.begin(), perm.end(), 0);
+  // Lambda with capture for sort comparison
+  const auto cmp
+      = [&received_buffer_array, num_vertices_per_facet](int a, int b) {
+          const auto facet_a = received_buffer_array.row(a);
+          const auto facet_b = received_buffer_array.row(b);
```
garth-wells, line 245: Need to take care that this doesn't create copies of the rows behind the scenes. Might be better to compute the offset manually.

cpp/dolfinx/mesh/GraphBuilder.cpp
```diff
@@ -271,32 +288,45 @@ compute_nonlocal_dual_graph(
...
-    if (it == graph[cell_list[i] - offset].end())
+    if (it == graph[cell_list[i] - cell_offset].end())
     {
       edges.push_back(cell_list[i + 1]);
       ++num_nonlocal_edges;
     }
+    else
+    {
+      LOG(ERROR) << "Received same edge twice in dual graph";
+      throw std::runtime_error("Inconsistent mesh data in GraphBuilder");
```
garth-wells, line 316: Make exception message more informative (like the log message).

Do we need log messages and exception messages?

garth-wells (changes requested):

cpp/dolfinx/mesh/GraphBuilder.cpp
chrisrichardson: When running from python, and reading from hdf5, the code just hangs, and the runtime error never appears. 
The logging has proved very useful.

cpp/dolfinx/mesh/GraphBuilder.cpp
chrisrichardson: I'm sure it doesn't, but I'll change to using explicit pointers to make it clearer.

garth-wells (approved):

PR #1262, opened 2020-12-16
Add Actions testing with Intel compiler

garth-wells:
Only tests C++ interface due to Intel compiler bugs that prevent build of Python/pybind11 interface.

### DIFF
--- a/cpp/dolfinx/graph/AdjacencyList.h
+++ b/cpp/dolfinx/graph/AdjacencyList.h
@@ -193,8 +193,10 @@ public:
   {
     if constexpr (std::is_same<X, T>::value)
       return *this;
+    else
+      return graph::AdjacencyList<X>(_array.template cast<X>(), _offsets);
 
-    return graph::AdjacencyList<X>(_array.template cast<X>(), _offsets);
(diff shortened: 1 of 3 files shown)


cpp/dolfinx/graph/AdjacencyList.h
```diff
@@ -191,8 +191,14 @@ class AdjacencyList
   template <typename X>
   decltype(auto) as_type() const
   {
+// Workaround for Intel compler bug, see
+// https://community.intel.com/t5/Intel-C-Compiler/quot-if-constexpr-quot-and-quot-missing-return-statement-quot-in/td-p/1154551
+#ifdef __INTEL_COMPILER
```
jhale, line 196: Indented preprocessor directives (should) be OK?

jhale: This is a style thing, ignore me, it's fine.

jhale (approved):

cpp/dolfinx/graph/AdjacencyList.h
garth-wells: I just let `clang-format` do its thing.

PR #1265, opened 2020-12-16
Python doc updates

chrisrichardson:


garth-wells (approved):

jorgensd:
If we could hold on this for a bit, Im currently updating the `solve` docs (will push something soon)

jorgensd:
@garth-wells I've now removed all the old docs in solve, and code-paths/options that are note used. Let me know what you think.

PR #1266, opened 2020-12-16
Expose ghosts of bc dof indices

jorgensd:
Currently, there is no way of accessing DirichletBC-dofs that are not owned by the current processor in the Python layer without using MPI gathering and filtering the results.
This is a problem for instance when using numba-assemblers, where you would like to remove all contributions in the matrix from bc-dofs (both owned and ghosts).
Itherefore propose changing `dof_indices` to returning the whole array  `_dofs0`.


garth-wells:
Having both feels a bit like duplication. I wonder if we could return the array of indices and the number of owned indices?

PR #1276, opened 2020-12-17
Add back DiscreteOperator

jorgensd:
Adding back discrete operator.
Works for function spaces were block size==1. Not sure how this should be done for blocked spaces.


michalhabera:
This is rather specific code. Why is it so much needed to have in core codebase? One can easily build any discrete operator (not just gradient) using ufc_expression and custom tabulation loop.

garth-wells:
> This is rather specific code. Why is it so much needed to have in core codebase? One can easily build any discrete operator (not just gradient) using ufc_expression and custom tabulation loop.

Example?

python/test/unit/fem/test_discrete_operators.py
```diff
@@ -1,60 +1,55 @@
...
-#     with pytest.raises(RuntimeError):
-#         DiscreteOperators.build_gradient(W._cpp_object, W._cpp_object)
-
-#     V = FunctionSpace(mesh, ("Lagrange", 2))
-#     with pytest.raises(RuntimeError):
-#         DiscreteOperators.build_gradient(W._cpp_object, V._cpp_object)
+import numpy
+import pytest
+from dolfinx import FunctionSpace, UnitCubeMesh, UnitSquareMesh
+from dolfinx.cpp.fem import build_discrete_operator
```
jpdean, line 11: Shouldn't this be `from dolfinx.cpp.fem import build_discrete_gradient`?

garth-wells: @jpdean can you push a fix?

jorgensd:
> > This is rather specific code. Why is it so much needed to have in core codebase? One can easily build any discrete operator (not just gradient) using ufc_expression and custom tabulation loop.
> 
> Example?



> > This is rather specific code. Why is it so much needed to have in core codebase? One can easily build any discrete operator (not just gradient) using ufc_expression and custom tabulation loop.
> 
> Example?

@michalhabera do you have an example of how to do this with ufc_expression?:)

michalhabera:
> > > This is rather specific code. Why is it so much needed to have in core codebase? One can easily build any discrete operator (not just gradient) using ufc_expression and custom tabulation loop.
> > 
> > 
> > Example?
> 
> > > This is rather specific code. Why is it so much needed to have in core codebase? One can easily build any discrete operator (not just gradient) using ufc_expression and custom tabulation loop.
> > 
> > 
> > Example?
> 
> @michalhabera do you have an example of how to do this with ufc_expression?:)

This concept is possible if expressions like https://github.com/FEniCS/ffcx/blob/master/test/test_jit_expression.py#L85 are assembled into PETSc Mat using INSERT_VALUES mode.

For C++ workflow a proof-of-concept for local operators is in https://github.com/FEniCS/ffcx/blob/michal/expressions-cpp/test/grad.cpp and https://github.com/FEniCS/ufl/compare/michal/expressions

It needs some work, and #1215 would help it. For this is not immediately available I agree this PR could be merged, but hopefully replaced with the use of `ufc_expressions` and insert mode assemblers, in very near future.

jorgensd:
Are we ready to merge this now?

jorgensd:
@garth-wells  The latest PR reduced the size of the sparsity pattern to what is actually needed, and speeds up the code quite a bit.
I am not convinced the calls to entity_dofs to get the local indices causes any significant slow-down. @jpdean has tested the code for an actual problem in parallel, and says that the new code works nicely.

cpp/dolfinx/fem/discreteoperators.cpp
```diff
@@ -0,0 +1,207 @@
...
+                                 + mesh->topology().index_map(1)->num_ghosts();
+  const std::shared_ptr<const fem::DofMap> dofmap0 = V0.dofmap();
+  assert(dofmap0);
+  for (std::int32_t e = 0; e < num_edges; ++e)
+  {
+    std::vector<std::int32_t> rows;
+    std::vector<std::int32_t> cols;
+  
+    // Find local index of edge in one of the cells it is part of
+    auto cells = e_to_c->links(e);
```
garth-wells, line 99: Avoid `auto` in this loop. It can hide expensive operations.

cpp/dolfinx/fem/discreteoperators.cpp
```diff
@@ -0,0 +1,207 @@
...
+    assert(cells.size() > 0);
+    const std::int32_t cell = cells[0];
+    auto edges = c_to_e->links(cell);
+    const auto* it = std::find(edges.data(), edges.data() + edges.size(), e);
+    assert(it != (edges.data() + edges.size()));
+    const int local_edge = std::distance(edges.data(), it);
+
+    // Find the dofs located on the edge
+    auto dofs0 = dofmap0->cell_dofs(cell);
+    auto local_dofs = layout0->entity_dofs(1, local_edge);
```
garth-wells, line 109: This call is expensive - take outside of loop.

cpp/dolfinx/fem/discreteoperators.cpp
```diff
@@ -0,0 +1,207 @@
...
+    // Find local index of each of the vertices and map to local dof
+    auto dofs1 = V1.dofmap()->cell_dofs(cell);
+    for (std::int32_t i = 0; i < 2; ++i)
+    {
+      const auto* it
+          = std::find(cell_vertices.data(),
+                      cell_vertices.data() + cell_vertices.size(), vertices[i]);
+      assert(it != (cell_vertices.data() + cell_vertices.size()));
+      const int local_vertex = std::distance(cell_vertices.data(), it);
+      auto local_v_dofs = layout1->entity_dofs(0, local_vertex);
```
garth-wells, line 125: Expensive call. Take outside loop.

cpp/dolfinx/fem/discreteoperators.cpp
```diff
@@ -0,0 +1,207 @@
...
+      assert(local_v_dofs.size() == 1);
+      cols.push_back(dofs1[local_v_dofs[0]]);
+    }
+    Eigen::Map<const Eigen::Array<std::int32_t, Eigen::Dynamic, 1>> _rows(
+      rows.data(), rows.size());
+    Eigen::Map<const Eigen::Array<std::int32_t, Eigen::Dynamic, 1>> _cols(
+      cols.data(), cols.size());
+  pattern.insert(_rows, _cols);
```
garth-wells, line 133: Looks old. I thought this now uses `span`?

cpp/dolfinx/fem/discreteoperators.cpp
```diff
@@ -0,0 +1,207 @@
...
+    auto edges = c_to_e->links(cell);
+    const auto* it = std::find(edges.data(), edges.data() + edges.size(), e);
+    assert(it != (edges.data() + edges.size()));
+    const int local_edge = std::distance(edges.data(), it);
+
+    // Find the dofs located on the edge
+    auto dofs0 = dofmap0->cell_dofs(cell);
+
+    // FIXME: avoid this expensive call
+    const std::vector<int32_t> local_dofs = layout0->entity_dofs(1, local_edge);
```
garth-wells, line 162: Take outside loop.

cpp/dolfinx/fem/discreteoperators.cpp
```diff
@@ -0,0 +1,207 @@
...
+    for (std::int32_t i = 0; i < 2; ++i)
+    {
+      const auto* it
+          = std::find(cell_vertices.data(),
+                      cell_vertices.data() + cell_vertices.size(), vertices[i]);
+      assert(it != (cell_vertices.data() + cell_vertices.size()));
+      const int local_vertex = std::distance(cell_vertices.data(), it);
+
+      // FIXME: avoid this expensive call
+      const std::vector<int32_t> local_v_dofs
```
garth-wells, line 182: Take outside loop.

garth-wells (changes requested):

garth-wells:
> @garth-wells The latest PR reduced the size of the sparsity pattern to what is actually needed, and speeds up the code quite a bit.
> I am not convinced the calls to entity_dofs to get the local indices causes any significant slow-down. @jpdean has tested the code for an actual problem in parallel, and says that the new code works nicely.

It's not good practice to be dynamically created `std::vector` inside a tight loop. I can be easily moved outside in this case.

garth-wells (approved):

garth-wells (changes requested):
Still has std::vector allocations in tight loops.

garth-wells (approved):

Issue #1277, opened 2020-12-17
Mismatch in Eigen aligment with other dependencies causes segfaults


jorgensd:
Setting alignment to zero in our CMakeLists removes the issue, but is it a solution?
In general, if more an more packages uses Eigen, how can we make sure all are compiled with the same alignment?

chrisrichardson:
I agree we should set to unaligned by default.

garth-wells:
See #1381.

PR #1278, opened 2020-12-17
GMSH eigen aligment + version update

jorgensd:
Setting default eigen alignment to zero removes the conflict between dolfin-x and gmsh eigen installations. 
One minor update due to version bump.
Adresses  #1277


IgorBaratta:
I am not sure about setting the alignment to 0.
But importing gmsh before dolfinx could be a temporary fix.

jhale:
Could you clarify where memory is being passed and in which direction? I.e. who is responsible for allocating it.

michalhabera (approved):

chrisrichardson:
If we set `EIGEN_MAX_ALIGN_BYTES=0` (I think we should) then the documentation around that needs correcting too.
Setting =0 is equivalent to disabling all alignment `EIGEN_DONT_ALIGN`.

chrisrichardson (approved):

jorgensd:
Close due to #1279

PR #1280, opened 2020-12-18
Add high level LinearProblem class (Python)

jorgensd:
As the solve-function has been misused by many people (me included) over the last decade, we have talked about replacing it with a `LinearProblem`-class.
We could also create a `NonlinearVariationalSolver`, which implements the Newton-method with boundary conditions, but I think that is something for a later PR.


python/demo/poisson/demo_poisson.py
```diff
@@ -175,14 +175,16 @@
...
-# <dolfinx.fem.solving.solve>` function with the arguments ``a == L``,
-# ``u`` and ``bc`` as follows: ::
+# a finite element function space. Next, we initialize a solver using the :py:class:`LinearVariationalSolver
+# <dolfinx.fem.linearvariationalsolver.LinearVariationalSolver>`.
+# This class is initialized with the arguments ``a == L``, ``u`` and ``bc`` as follows: ::
 
-# Compute solution
 u = Function(V)
-solve(a == L, u, bc, petsc_options={"ksp_type": "preonly", "pc_type": "lu"})
+lin_solver = LinearVariationalSolver(a == L, u, [bc])
```
garth-wells, line 183: `solver` or `problem` would be a neater name in place of `lin_solver`.

python/demo/poisson/demo_poisson.py
```diff
@@ -175,14 +175,16 @@
...
 
+# When we want to compute the solution to the problem, we can specify what kind of solver we want to use.
+# In this problem, we use a direct LU solver, which is defined through the dictionary ``petsc_options``.
+lin_solver.solve(petsc_options={"ksp_type": "preonly", "pc_type": "lu"})
```
garth-wells, line 187: These argument probably need to be passed to the `LinearVariationalSolver` object as the it should store its solver (ksp) object.

python/dolfinx/__init__.py
```diff
@@ -46,8 +46,7 @@
 
 from .fem.form import Form
 from .fem.dirichletbc import DirichletBC
-from .fem.solving import solve
-
+from .fem.linearvariationalsolver import LinearVariationalSolver
```
garth-wells, line 49: I'd like to avoid mangling the namespaces - remove, at least for now,

python/dolfinx/fem/linearvariationalsolver.py
```diff
@@ -0,0 +1,80 @@
...
+
+from petsc4py import PETSc
+from ufl.equation import Equation
+from dolfinx import fem
+
+
+class LinearVariationalSolver():
+    assemble_b = True
+
+    def __init__(self, variational_problem: Equation, u: fem.Function,
```
garth-wells, line 17: We want to remove `Equation` - it's an abuse of syntax. Pass two forms instead.

I'm not sure whether or not we want to pass a solution function, or let the object create it.

python/dolfinx/fem/linearvariationalsolver.py
```diff
@@ -0,0 +1,80 @@
...
+
+        """
+        if (len(variational_problem.lhs.arguments()) != 2):
+            raise ValueError(
+                "Variational problem has to consist of a bi-linear form,"
+                + " defined with a `ufl.TrialFunction` and `ufl.TestFunction`.")
+        # If a == 0 is inserted just assume that rhs is 0.
+        if (variational_problem.rhs == 0):
+            self.assemble_b = False
+            self.b = u.vector.copy()
```
garth-wells, line 35: The RHS vector should be created by the matrix object to ensure a consistent layout.

python/dolfinx/fem/linearvariationalsolver.py
```diff
@@ -0,0 +1,80 @@
...
+        .. code-block:: python
+
+            solver = LinearVariationalSolver(a==L, u, [bc0, bc1], form_compiler_parameters={"optimize": True})
+
+        """
+        if (len(variational_problem.lhs.arguments()) != 2):
+            raise ValueError(
+                "Variational problem has to consist of a bi-linear form,"
+                + " defined with a `ufl.TrialFunction` and `ufl.TestFunction`.")
+        # If a == 0 is inserted just assume that rhs is 0.
```
garth-wells, line 32: Don't allow this.

python/dolfinx/fem/linearvariationalsolver.py
```diff
@@ -0,0 +1,80 @@
...
+
+        self.solver = PETSc.KSP().create(u.function_space.mesh.mpi_comm())
+        self.a = fem.Form(variational_problem.lhs, form_compiler_parameters=form_compiler_parameters)
+        self.A = fem.create_matrix(self.a)
+        self.solver.setOperators(self.A)
+
+        self.bcs = bcs
+        self.u = u
+
+    def solve(self, petsc_options={}):
```
garth-wells, line 48: Solver options need to be passed to initialiser.

python/dolfinx/fem/linearvariationalsolver.py
```diff
@@ -0,0 +1,80 @@
...
+        self.solver.setOptionsPrefix("dolfinx_solve_")
+        opts = PETSc.Options()
+        opts.prefixPush("dolfinx_solve_")
+        for k, v in petsc_options.items():
+            opts[k] = v
+        opts.prefixPop()
+
+        self.solver.setFromOptions()
+        self.solver.solve(self.b, self.u.vector)
+        self.u.vector.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
```
garth-wells, line 80: return `self.u`.

garth-wells (changes requested):

python/dolfinx/fem/linearvariationalsolver.py
michalhabera: I agree with @garth-wells here.

python/dolfinx/fem/linearvariationalsolver.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, variational_problem: Equation, u: fem.Function,
+                 bcs: typing.List[fem.DirichletBC] = [], form_compiler_parameters={}):
+        """
+        Initialize the linear variational solver. By initialization we mean creating the underlying PETSc solver,
+        and the linear algebra structures (the matrix and vector), that can be reused for multiple solve calls.
+
+        .. code-block:: python
+
+            solver = LinearVariationalSolver(a==L, u, [bc0, bc1], form_compiler_parameters={"optimize": True})
```
michalhabera, line 25: `{"optimize": True}` is some very old legacy, doesn't exist anymore. Please pass `form_compiler_parameters` and `jit_parameters` as well.

python/dolfinx/fem/linearvariationalsolver.py
```diff
@@ -0,0 +1,80 @@
...
+
+        # Apply boundary conditions
+        fem.apply_lifting(self.b, [self.a], [self.bcs])
+        self.b.ghostUpdate(addv=PETSc.InsertMode.ADD, mode=PETSc.ScatterMode.REVERSE)
+        fem.set_bc(self.b, self.bcs)
+
+        # Solve linear algebra problem using PETSc
+        self.solver.setOptionsPrefix("dolfinx_solve_")
+        opts = PETSc.Options()
+        opts.prefixPush("dolfinx_solve_")
```
michalhabera, line 73: What if I have two instances of solver, both with different parameters? The prefix should be unique.

michalhabera (commented):
I would suggest a different name, `LinearSolver`? Variational is a bit weird, as we do not really define a functional which is to be minimized - we already give a weak form.

python/demo/helmholtz2D/demo_helmholtz_2d.py
```diff
@@ -15,8 +15,8 @@
 import numpy as np
 from mpi4py import MPI
 
-from dolfinx import (Function, FunctionSpace, UnitSquareMesh,
-                     has_petsc_complex, solve)
+from dolfinx import Function, FunctionSpace, UnitSquareMesh, has_petsc_complex
+from dolfinx.fem import LinearProblem
```
garth-wells, line 19: I would like to see `fem.LinearProblem` used the the demos. Using namespacing means we can keep names short, like 'LinearProblem', without losing the context.

python/demo/helmholtz2D/demo_helmholtz_2d.py
```diff
@@ -52,14 +52,14 @@
 L = inner(f, v) * dx
 
 # Compute solution
-u = Function(V)
-solve(a == L, u, [])
+problem = LinearProblem(a, L)
```
garth-wells, line 55: `problem = fem.LinearProblem`

python/dolfinx/fem/linearproblem.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class LinearProblem():
+    """
+    Class for solving a linear variational problem of the form a(u,v)=L(v)
+    using PETSc as a linear algebra backend.
+    """
+
+    def __init__(self, a: form.Form, L: form.Form, bcs: typing.List[fem.DirichletBC] = [],
+                 petsc_options={}, form_compiler_parameters={}, jit_parameters={}):
```
garth-wells, line 21: Use `foo_options=` for all?

python/dolfinx/fem/linearproblem.py
```diff
@@ -0,0 +1,101 @@
...
+#
+# SPDX-License-Identifier:    LGPL-3.0-or-later
+import typing
+
+from petsc4py import PETSc
+from ufl import form
+from dolfinx import fem
+
+
+class LinearProblem():
```
garth-wells, line 14: I'd suggest adding some methods to access the solver, solution, RHS, etc.

python/dolfinx/fem/linearproblem.py
```diff
@@ -0,0 +1,101 @@
+# -*- coding: utf-8 -*-
+# Copyright (C) 2020 Jørgen S. Dokken
+#
+# This file is part of DOLFINX (https://www.fenicsproject.org)
+#
+# SPDX-License-Identifier:    LGPL-3.0-or-later
+import typing
+
+from petsc4py import PETSc
+from ufl import form
```
garth-wells, line 10: Use module name for UFL function/classes, otherwise it get confusing with names like 'form'.

```
import ufl
foo = ufl.form()
```

garth-wells (changes requested):

python/dolfinx/fem/linearproblem.py
michalhabera: `form_compiler_parameters` and `jit_parameters` are standard names used in Form() and jit call.

python/dolfinx/fem/linearproblem.py
michalhabera: I think @jorgensd meant this as a direct access to member variables of the class, `self.solver`, ...

python/dolfinx/fem/linearproblem.py
```diff
@@ -0,0 +1,101 @@
...
+
+        # Extract function space from TrialFunction (which is at the end of the argument
+        # list as it is numbered as 1, while the Test function is numbered as 0)
+        self.u = fem.Function(a.arguments()[-1].ufl_function_space())
+        self.bcs = bcs
+
+        self.solver = PETSc.KSP().create(self.u.function_space.mesh.mpi_comm())
+        self.solver.setOperators(self.A)
+        # Give unique prefix linked to the dolfinx.fem.Function's unique ID.
+        self.solver.setOptionsPrefix("dolfinx_solve_{0:d}".format(self.u.id))
```
michalhabera, line 70: What if I change the function `u`, because I have decided to store solution in a different place? This should be unique to the very LinearProblem instance. You can achieve this with global ID counter which increments for each class instance in the current Python run -- just like ufl.Coefficient does, see implementation there.

python/dolfinx/fem/linearproblem.py
jorgensd: I can add getters if that is something we want. As @michalhabera I figured one could just access the obects as problem.solver, problem.u, problem.L etc. as they are public.

michalhabera: We just need to provent following:
```python
prob = LinearProblem(a, L)
prob.L = L2
prob.solve()
```
which would incorrectly lead to the impression one can change the form after LinearProblem construction.

Simplest way would be to store such variables into private names `self._L, self._b` and provide a getter for it.

python/dolfinx/fem/linearproblem.py
```diff
@@ -0,0 +1,106 @@
...
+                          jit_parameters=jit_parameters)
+        self.A = fem.create_matrix(self.a)
+
+        self.L = fem.Form(L, form_compiler_parameters=form_compiler_parameters,
+                          jit_parameters=jit_parameters)
+        self.b = fem.create_vector(self.L)
+
+        # Extract function space from TrialFunction (which is at the end of the argument
+        # list as it is numbered as 1, while the Test function is numbered as 0)
+        self.u = fem.Function(a.arguments()[-1].ufl_function_space())
```
michalhabera, line 65: Personally, I do not like this approach. I would suggest to create result function only if user didn't provide any. That is less forcing.

python/dolfinx/fem/linearproblem.py
jorgensd: Sure, I'll add that

python/dolfinx/fem/linearproblem.py
jorgensd: As long as @garth-wells are onboard with that (I added this after input from him on a previous iteration of the PR)

python/dolfinx/fem/linearproblem.py
garth-wells: 'options' sound better - I think we should change it elsewhere. Who uses 'compiler parameters' rather than 'compiler options'?

michalhabera: But this would be a separate PR with overlap in ffcx.

jorgensd: Created separate issue #1290

python/dolfinx/fem/linearproblem.py
```diff
@@ -0,0 +1,129 @@
...
+from dolfinx import fem
+from petsc4py import PETSc
+
+
+class LinearProblem():
+    """
+    Class for solving a linear variational problem of the form a(u,v)=L(v)
+    using PETSc as a linear algebra backend.
+    """
+    _globalcount = 0
```
garth-wells, line 19: What is this for?

jorgensd: As @michalhabera mentioned in a previous comment, to make the PETSc-options unique to the current solver (imagine having three `LinearProblem`s with different options, we need a unique prefix to the options. These are create using `         ufl.utils.counted.counted_init(self, countedclass=LinearProblem)` which is the class used to give for instance `ufl.Coefficient` a unique id.

garth-wells: I've pushed a simpler fix - just uses `id(self)`.

garth-wells:
I'm happy to merge this. Generalisation and refinements come later. The only think it removes is a long standing problem.

PR #1281, opened 2020-12-18
Add NonlinearVariationalSolver

jorgensd:
As we should have some support for solving non-linear problems without demanding the user to know how to create a Newton iteration, I suggest a `NonlinearVariationalSolver` class, which is based on a standard Newton solver.


jorgensd:
Close due to parallel work by @garth-wells  in https://github.com/FEniCS/dolfinx/tree/garth/nls-change

PR #1284, opened 2020-12-20
Use STL containers in AdjacencyList and add an implementation of (std::)span

garth-wells:
The PR changes internal storage of AdjacencyList from Eigen to std::vector. This reduced the verbosity of the code and makes it easier to apply STL algorithms. The code is also more transparent.

To get 'views' into the data, e.g. the list of edges of a graph node, an implementation of std::span has been dropped into the code. `std::span` is part of the C++20, so the drop in code can be removed in the future when we move to C++20.

Quite a bit of unnecessary Eigen code has been removed (1D arrays that are not used for computation) which makes code clearer. Also fixes the usage of `py::array_t` in the pybind11 layer and removed unused mesh hashing.

Testing with https://github.com/FEniCS/performance-test show no performance change.


cpp/dolfinx/fem/interpolate.h
```diff
@@ -109,10 +109,10 @@ void interpolate_from_any(Function<T>& u, const Function<T>& v)
   assert(bs == dofmap_u->bs());
   for (int c = 0; c < num_cells; ++c)
   {
-    auto dofs_v = dofmap_v->cell_dofs(c);
-    auto cell_dofs = dofmap_u->cell_dofs(c);
+    tcb::span<const std::int32_t> dofs_v = dofmap_v->cell_dofs(c);
```
michalhabera, line 112: Const span

cpp/dolfinx/fem/utils.h
```diff
@@ -437,10 +437,10 @@ pack_coefficients(const U& u)
     {
       for (std::size_t coeff = 0; coeff < dofmaps.size(); ++coeff)
       {
-        auto dofs = dofmaps[coeff]->cell_dofs(cell);
+        tcb::span<const std::int32_t> dofs = dofmaps[coeff]->cell_dofs(cell);
```
michalhabera, line 440: Const span

cpp/dolfinx/graph/AdjacencyList.h
```diff
@@ -129,62 +113,47 @@ class AdjacencyList
...
   }
 
-  /// Links (edges) for given node
+  /// Get the links (edges) for given node
   /// @param [in] node Node index
   /// @return Array of outgoing links for the node. The length will be
-  ///   AdjacencyList:num_links(node).
-  typename Eigen::Array<T, Eigen::Dynamic, 1>::SegmentReturnType links(int node)
+  /// AdjacencyList:num_links(node).
+  tcb::span<T> links(int node)
```
michalhabera, line 136: These could have also a `const` return variant?

python/dolfinx/wrappers/graph.cpp
```diff
@@ -25,22 +25,28 @@ void declare_adjacency_list(py::module& m, std::string type)
   py::class_<dolfinx::graph::AdjacencyList<T>,
              std::shared_ptr<dolfinx::graph::AdjacencyList<T>>>(
       m, pyclass_name.c_str(), "Adjacency List")
-      .def(py::init<std::int32_t>())
+      // .def(py::init<std::int32_t>())
```
michalhabera, line 28: Leftover

cpp/demo/hyperelasticity/main.cpp
```diff
@@ -28,11 +28,10 @@ class HyperElasticProblem : public nls::NonlinearProblem
     const int bs = L->function_spaces()[0]->dofmap()->index_map_bs();
     std::int32_t size_local = bs * map->size_local();
 
-    const Eigen::Array<PetscInt, Eigen::Dynamic, 1>& ghosts
-        = map->ghosts().cast<PetscInt>();
+    std::vector<PetscInt> ghosts(map->ghosts().begin(), map->ghosts().end());
```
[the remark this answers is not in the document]
garth-wells: Doesn't really matter here.

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -321,15 +319,15 @@ fem::locate_dofs_topological(
...
+    const auto* it
+        = std::find(entities_d.begin(), entities_d.end(), entities[e]);
+    assert(it != entities_d.end());
+    const int entity_local_index = std::distance(entities_d.begin(), it);
 
     // Get cell dofmap
-    auto cell_dofs0 = dofmap0->cell_dofs(cell);
-    auto cell_dofs1 = dofmap1->cell_dofs(cell);
-    assert(bs0 * cell_dofs0.rows() == bs1 * cell_dofs1.rows());
+    tcb::span<const std::int32_t> cell_dofs0 = dofmap0->cell_dofs(cell);
```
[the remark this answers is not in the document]
garth-wells: Could be, but doesn't have to be.

cpp/dolfinx/fem/Function.h
```diff
@@ -358,8 +358,8 @@ class Function
                                          X, J, detJ, K, cell_info[cell_index]);
 
       // Get degrees of freedom for current cell
-      auto dofs = dofmap->cell_dofs(cell_index);
-      for (Eigen::Index i = 0; i < dofs.size(); ++i)
+      tcb::span<const std::int32_t> dofs = dofmap->cell_dofs(cell_index);
```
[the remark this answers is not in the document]
garth-wells: See above comment.

cpp/dolfinx/graph/AdjacencyList.h
garth-wells: Doesn't really make sense to return by value and use `const`.

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -49,8 +50,7 @@ namespace fem
 /// V[0] and V[1]. The array[0](i) entry is the DOF index in the space
 /// V[0] and array[1](i) is the correspinding DOF entry in the space
 /// V[1]. The returned dofs are 'unrolled', i.e. block size = 1.
-std::array<Eigen::Array<std::int32_t, Eigen::Dynamic, 1>, 2>
-locate_dofs_topological(
```
michalhabera, line 53: These must be fixed also in pybind11 wrappers, otherwise they return List().

garth-wells: Fixed.

cpp/dolfinx/graph/AdjacencyList.h
mrambausek: But `tcb::span<const T> links(int node) const` maybe? (I guess span<T> works like a view that allows writing to the viewed memory if T is not const)

cpp/dolfinx/graph/AdjacencyList.h
```diff
@@ -32,83 +50,49 @@ class AdjacencyList
...
   {
-    const std::int32_t num_links = matrix.cols();
-    for (Eigen::Index e = 0; e < _offsets.rows(); e++)
-      _offsets[e] = e * num_links;
-
-    // NOTE: Do not directly copy data from matrix because it may be a
-    // view into a larger array
-    for (Eigen::Index i = 0; i < matrix.rows(); ++i)
-      _array.segment(_offsets(i), num_links) = matrix.row(i);
+    std::tie(_array, _offsets) = create_adjacency_data(matrix);
```
IgorBaratta, line 77: Is `std::tie` preferred to structured binding here?

cpp/dolfinx/graph/AdjacencyList.h
garth-wells: Yes, `tcb::span<const T> links(int node) const` has already been added.

cpp/dolfinx/graph/AdjacencyList.h
garth-wells: I'm not sure what the options are. Structured binding followed by a move?

This constructor will get removed at some point in favour of a helper function - keeps the constructors simple and computation light.

IgorBaratta: Ok, I looked for some options, and `std::tie` seems to be neater for variables already declared.

PR #1285, opened 2020-12-21
Change Newton solver to functional design

### LINKED ISSUE #1173: Redesign NewtonSolver to follow functional design
The `NewtonSolver` uses `NonlinearProblem`, with the latter following a classical object oriented design. It would be simpler and consistent with other parts of DOLFINX if `NewtonSolver` followed a functional design, i.e. asked for functions that compute the Jacobian and residual rathe than a `NonlinearProblem` instance. This also simplifies the pybind11 interfacing considerably as we don't need to muck about with trampoline classes.

garth-wells:
This removes the NonlinearProblem class, and instead the Newton solver accepts functions for computing the Jacobian and the residual. This avoids inheritance and simplifies the pybind11 interface (no trampoline classes!). The design is also much closer to the SNES design.

The NewtonSolver is not very polished as it wasn't very polished before this change.

Fixes #1173.


python/test/unit/nls/test_newton.py
```diff
@@ -100,23 +94,26 @@ def boundary(x):
         return np.logical_or(x[0] < 1.0e-8, x[0] > 1.0 - 1.0e-8)
 
     u_bc = fem.Function(V)
-    u_bc.vector.set(1.0)
-    u_bc.vector.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+    with u_bc.vector.localForm() as u_local:
```
jorgensd, line 97: I would suggest that we set the boundary condition in the following way:
```
boundary_dofs = fem.locate_dofs_geometrical(V, boundary)
with u_bc.vector.localForm() as u_local:
    u_local.setValues(boundary_dofs, np.full(len(boundary_dofs), 1))
```
This limits the number of values we have to set in each boundary condition (and the function can be used for other constant DirichletBCs).
```

File: python/test/unit/nls/test_newton.py [hunk 2]
```diff
@@ -100,23 +94,26 @@ def boundary(x):
...
+    n, converged = solver.solve(u.vector)
     assert converged
     assert n == 1
 
     # Increment boundary condition and solve again
-    u_bc.vector.set(2.0)
-    u_bc.vector.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
-    n, converged = solver.solve(problem, u.vector)
+    with u_bc.vector.localForm() as u_local:
+        u_local.set(2.0)
```
jorgensd, line 115: Same suggestion as above

python/test/unit/nls/test_newton.py
```diff
@@ -136,25 +133,28 @@ def boundary(x):
         return np.logical_or(x[0] < 1.0e-8, x[0] > 1.0 - 1.0e-8)
 
     u_bc = fem.Function(V)
-    u_bc.vector.set(1.0)
-    u_bc.vector.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+    with u_bc.vector.localForm() as u_local:
+        u_local.set(1.0)
```
jorgensd, line 137: Same as above

python/test/unit/nls/test_newton.py
```diff
@@ -136,25 +133,28 @@ def boundary(x):
...
+    n, converged = solver.solve(u.vector)
     assert converged
     assert n < 6
 
     # Modify boundary condition and solve again
-    u_bc.vector.set(0.5)
-    u_bc.vector.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
-    n, converged = solver.solve(problem, u.vector)
+    with u_bc.vector.localForm() as u_local:
+        u_local.set(0.5)
```
jorgensd, line 156: Same as above

python/test/unit/nls/test_newton.py
```diff
@@ -174,16 +174,15 @@ def boundary(x):
         return np.logical_or(x[0] < 1.0e-8, x[0] > 1.0 - 1.0e-8)
 
     u_bc = fem.Function(V)
-    u_bc.vector.set(1.0)
-    u_bc.vector.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+    with u_bc.vector.localForm() as u_local:
+        u_local.set(1.0)
```
jorgensd, line 178: Same as above

python/test/unit/nls/test_newton.py
```diff
@@ -204,70 +203,10 @@ def boundary(x):
     assert snes.getIterationNumber() < 6
 
     # Modify boundary condition and solve again
-    u_bc.vector.set(0.5)
-    u_bc.vector.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+    with u_bc.vector.localForm() as u_local:
```
jorgensd, line 206: Same as above

jorgensd (commented):
Only minor docs changes.

python/test/unit/nls/test_newton.py
jorgensd: Ignore this, as it will not be a problem when we introduce `dolfinx.Constant` as an input argument to `DirichletBC`

PR #1287, opened 2020-12-22
Support custom mesh distributions by pass partitioning function to mesh builders

IgorBaratta:
To deal with the `MPICommWrapper` translation to c++ in pybind, I've created a wrapper function allowing one to pass a python function `partitioner` with `MPICommWrapper` as argument to C++ (that accepst `MPI_Comm`):
```c++
auto partitioner_wrapper
    = [partitioner](
          MPI_Comm comm, int n, const dolfinx::mesh::CellType cell_type,
          const dolfinx::graph::AdjacencyList<std::int64_t>& cells,
          dolfinx::mesh::GhostMode ghost_mode) {
        return partitioner(MPICommWrapper(comm), n, cell_type, cells,
                           ghost_mode);
      };
```

To avoid copying back and forth the partitioning function signature, I've created a typedef 
```c++
using PartitioningFunction
    = std::function<const dolfinx::graph::AdjacencyList<std::int32_t>(
        MPI_Comm mpi_comm, int nparts, const dolfinx::mesh::CellType cell_type,
        const dolfinx::graph::AdjacencyList<std::int64_t>& cells,
        dolfinx::mesh::GhostMode ghost_mode)>;
```
which is documented with doxygen.

I'm still working on more elaborated tests, but it's already possible to assess the proposed interface.


garth-wells:
Feels to me that this need to go one level lower. `mesh::partition_cells` does more than just partition - it builds the dual graph.

Could `mesh::partition_cells` take a `std::function` that partitions a graph?

IgorBaratta:
> Feels to me that this need to go one level lower. `mesh::partition_cells` does more than just partition - it builds the dual graph.
> 
> Could `mesh::partition_cells` take a `std::function` that partitions a graph?

I thought about leaving it like this because not all partitioners need the dual graph. One might be interested in partition cells by position (cell midpoint) or directly via `ParMETIS V3 PartMeshKway`, for example. 

But I think `mesh::partition_cells` can still take a `std::function` with the actual graph partitioner.

garth-wells:
> I thought about leaving it like this because not all partitioners need the dual graph. One might be interested in partition cells by position (cell midpoint) or directly via `ParMETIS V3 PartMeshKway`, for example.

Good point.
 
> But I think `mesh::partition_cells` can still take a `std::function` with the actual graph partitioner.
> I will revise the interface.

Sounds good - then we have complete flexibility. Provide a partitioner to `mesh::partition_cells` or provide a user implementation of `mesh::partition_cells`.

garth-wells (approved):

PR #1300, opened 2021-01-03
More mesh graph construction speedup

garth-wells:


michalhabera:
Could you check if this issue was improved as well (possibly close)? https://github.com/FEniCS/dolfinx/issues/101

garth-wells:
These changes don't address #101 - the cause of #101 is that mesh creation follows the same pipeline in serial and parallel. An optimisation would be to skip some steps in serial.

PR #1309, opened 2021-01-07
Generalize and simplify compute_closest_entity

### LINKED ISSUE #1352: Remove geometry::node
This is a left over from when topology and geometry were not cleanly separated. Not clear what this function does for some cell types or for high-order geometry.

It's still used a bit in the bounding box tree code.

jorgensd:
The current implementation of `compute_closest_entity` has several issues:

-  **Bug not relevant after PR**: Does currently not use the midpoint tree as an initial guess (ignores output of `_compute_closest_point`). With this PR we remove the initial guess given by the midpoint tree from `compute_closest_entity`. Instead, we can give an initial search radius for which the entity should be within. This will accelerate the search through the boundingbox tree.
- **Restructuring**: It was unclear for any user why the midpoint tree was required in `compute_closest_entity`. With this PR the midpoint tree can be used as input to `compute_closest_entity` to get an initial radius estimate that can be used to speed up the search through the actual boundingbox tree. This is illustrated in `test_midpoint_tree`
-  **Limitation**: This PR extends `compute_closest_entity` to work for any mesh entity, not just cells.

Also closes #1352


cpp/dolfinx/geometry/BoundingBoxTree.cpp
```diff
@@ -268,6 +270,21 @@ BoundingBoxTree::BoundingBoxTree(const mesh::Mesh& mesh, int tdim,
             << " nodes for " << num_leaves << " entities.";
 }
 //-----------------------------------------------------------------------------
+void BoundingBoxTree::remap_entity_indices(
+    std::vector<std::int32_t> entity_indices)
```
chrisrichardson, line 274: Pass by reference

cpp/dolfinx/geometry/BoundingBoxTree.h
```diff
@@ -102,6 +102,14 @@ class BoundingBoxTree
...
   }
 
+  /// Remap entity indices for bounding box trees that does not span a whole
+  /// mesh. Each leaf node should contain the actual entity index, not a
+  /// reference to the index of the entity_indices list.
+  /// @param[in] entity_indices The list of entities used to build the bounding
+  /// box tree. They should be in the same order as the input to the bounding
+  /// box tree constructor.
+  void remap_entity_indices(std::vector<std::int32_t> entity_indices);
+
```
chrisrichardson, line 112: ref

cpp/dolfinx/geometry/utils.cpp
```diff
@@ -279,57 +248,34 @@ double geometry::compute_squared_distance_bbox(
...
+  // If initial search radius is 0 we estimate the initial distance to the point
+  // using a "random" node from the tree.
+  std::int32_t initial_guess;
+  if (R < 0)
   {
-    throw std::runtime_error("Cannot compute closest point. "
-                             "Search tree has not been built for point cloud");
+    initial_guess = 0;
+    R = std::sqrt((tree.get_bbox(initial_guess).row(0).transpose().matrix() - p)
+                      .squaredNorm());
```
chrisrichardson, line 267: Why take sqrt? (could just use `.norm()` method). But why not just keep R^2, it is cheaper...

cpp/dolfinx/geometry/utils.cpp
```diff
@@ -279,57 +248,34 @@ double geometry::compute_squared_distance_bbox(
...
-
-  // Call recursive find function
-  _compute_closest_point(tree, p, tree.num_bboxes() - 1, closest_point, R2);
-
-  return {closest_point, sqrt(R2)};
+  else
+    initial_guess = -1;
+
+  // Use GJK to find determine the actual closest entity
+  const auto [index, distance] = _compute_closest_entity(
```
chrisrichardson, line 273: rename `distance` as it is actually distance^2.

cpp/dolfinx/geometry/utils.cpp
chrisrichardson: Maybe just add a new variable
`const double R2 = (R < 0) ? tree.get_bbox(...).squaredNorm() : R*R; `

chrisrichardson (approved):

PR #1311, opened 2021-01-07
Fix int narrowing in std::partial_sum

garth-wells:


chrisrichardson (approved):
Approving because it fixes the issue, but this is ugly and shows up the flaws in current c++/stl.

PR #1315, opened 2021-01-12
Dofmapbuilder optimise

chrisrichardson:
* Remove `owned[]` array 
* Remove `pos[]` array and release `offsets[]` a bit earlier.
* Should save a bit of memory and maybe time in this code which is at the high-water mark of mesh generation

### DIFF
--- a/cpp/dolfinx/fem/dofmapbuilder.cpp
+++ b/cpp/dolfinx/fem/dofmapbuilder.cpp
@@ -226,24 +226,14 @@ std::pair<std::vector<std::int32_t>, std::int32_t> compute_reordering_map(
       offset[d] = map->size_local();
   }
 
-  // Count locally owned dofs
-  std::vector<bool> owned(dof_entity.size(), false);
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -226,24 +226,14 @@ std::pair<std::vector<std::int32_t>, std::int32_t> compute_reordering_map(
...
-  }
-
   // Create map from old index to new contiguous numbering for locally
   // owned dofs. Set to -1 for unowned dofs
   std::vector<int> original_to_contiguous(dof_entity.size(), -1);
   std::int32_t owned_size = 0;
   for (std::size_t i = 0; i < original_to_contiguous.size(); ++i)
   {
-    if (owned[i])
+    const auto& e = dof_entity[i];
```
garth-wells, line 235: Could you make the type explicit rather than using `auto`?

garth-wells (changes requested):

garth-wells (approved):

PR #1317, opened 2021-01-14
Improve modules names for plot

garth-wells:


jorgensd:
Matplotlib is being deprecated, see  #1161

PR #1323, opened 2021-01-15
Add execution policy to std::inclusive_scan call

jhale:
Slightly old GCC versiond do not have the call without the execution policy documented here:

https://en.cppreference.com/w/cpp/algorithm/inclusive_scan

Discussed here:

https://stackoverflow.com/questions/55771604/g-with-stdexclusive-scan-c17

### DIFF
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -6,6 +6,7 @@
 
 #include "IndexMap.h"
 #include <algorithm>
+#include <execution>
 #include <numeric>
 #include <unordered_map>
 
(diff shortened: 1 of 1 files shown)


jhale:
Didn't really work, requires tbb.

garth-wells:
We can replace `std::inclusive_scan` with a loop. We had to remove `std::partial_sum` because of an integer overflow issue.

jhale:
Yep I can do that. You really do need a very new gcc to use inclusive_scan e.g. 9.3.0. Even 9.2.0 does not work. I'll leave a comment to indicate that it could be swapped back in the future.

PR #1324, opened 2021-01-15
Transition to basix library for finite element definitions

### LINKED ISSUE #599: get_entity_vertices_"CellType" should use FIAT information.
In `dolfin/cpp/mesh/cell_types.cpp`there are several functions that manually set up the entity vertices for each celltype ( [Example for hexahedron](https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfin/mesh/cell_types.cpp#L76-L125) ). This should rather take use of [FIATs reference element](https://github.com/FEniCS/fiat/blob/master/FIAT/reference_element.py#L567-L590)

### LINKED ISSUE #736: Add dof permutations for spaces where the shape of the dof arrangement does not match the shape of the face
For some spaces (eg `("N2E", tetrahedron, 3)`), the arrangement of dofs on the faces/interior does not match the shape of the face. For these spaces, the dof_permuter will either fail an assert or incorrectly permute the dofs. This was causing #728.

Picture of `("N2E", tetrahedron, 3)` from [femtable](http://femtable.org/):
N2E

There are also dofs of different types on a face, which could cause further complications.

mscroggs:
Refactor everything to use Basix in the place of FIAT. This PR is twinned with FEniCS/ffcx#296.

Close #736 (that can now be done by basix). Close #599 (all done except the functions mentioned in #1338).


cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -74,42 +112,54 @@ void CoordinateElement::compute_reference_geometry(
...
+  std::vector<Eigen::ArrayXXd> tabulated_data;
 
   if (_is_affine)
   {
     Eigen::Matrix<double, Eigen::Dynamic, 1, Eigen::ColMajor, 3, 1> x0(_gdim);
-    Eigen::Matrix<double, Eigen::Dynamic, 1, Eigen::ColMajor, 3, 1> X0(_tdim);
-    X0.setZero();
+    Eigen::ArrayXXd X0 = Eigen::ArrayXXd::Zero(1, tdim);
+
+    tabulated_data = basix::tabulate(_basix_element_handle, 1, X0);
```
jorgensd, line 146: Shouldn't this be 0?

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -5,30 +5,59 @@
...
-    mesh::CellType cell_type, int topological_dimension,
-    int geometric_dimension, const std::string& signature,
-    const ElementDofLayout& dof_layout, bool is_affine,
-    const std::function<int(double*, int, int, const double*)>&
-        evaluate_basis_derivatives)
-    : _tdim(topological_dimension), _gdim(geometric_dimension),
-      _cell(cell_type), _signature(signature), _dof_layout(dof_layout),
-      _is_affine(is_affine),
-      _evaluate_basis_derivatives(evaluate_basis_derivatives)
+    int basix_element_handle, int geometric_dimension,
```
michalhabera, line 16: What is `basix_element_handle`?

cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -91,12 +101,16 @@ class CoordinateElement
                                           Eigen::Dynamic, Eigen::RowMajor>>&
           cell_geometry) const;
 
-private:
-  // Topological and geometric dimensions
-  int _tdim, _gdim;
+  /// TODO
```
michalhabera, line 104: Can you add the docstring?

cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -66,6 +55,28 @@ FiniteElement::FiniteElement(const ufc_finite_element& ufc_element)
...
+         {interval, "interval"},
+         {triangle, "triangle"},
+         {tetrahedron, "tetrahedron"},
+         {quadrilateral, "quadrilateral"},
+         {hexahedron, "hexahedron"}};
+  const std::string cell_shape = ufc_to_cell.at(ufc_element.cell_shape);
+
+  const std::string family = ufc_element.family;
+
+  if (family == "mixed element")
```
michalhabera, line 69: Can you explain this condition?

cpp/dolfinx/fem/Function.h
```diff
@@ -279,29 +278,31 @@ class Function
...
-        sub_f.eval(x, cells, sub_u);
-
-        for (int i = 0; i < sub_value_size; ++i)
-          u.col(offset + i) = sub_u.col(i);
-        offset += sub_value_size;
-      }
-
-      return;
+      throw std::runtime_error("Not supported");
+      // if (bs_element != 1)
```
michalhabera, line 282: Commented code.

cpp/dolfinx/fem/FunctionSpace.cpp
```diff
@@ -39,66 +39,86 @@ internal_tabulate_dof_coordinates(
...
+  if (needs_permutation_data)
+    mesh->topology_mutable().create_entity_permutations();
+  const std::vector<std::uint32_t>& cell_info
+      = needs_permutation_data ? mesh->topology().get_cell_permutation_info()
+                               : std::vector<std::uint32_t>(num_cells);
+
   for (int c = 0; c < num_cells; ++c)
   {
-    // Update cell
+    // X = element->dof_coordinates(cell_info[c]);
```
michalhabera, line 101: Leftover comment.

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -115,37 +113,80 @@ class FiniteElement
...
+  bool interpolation_ident() const noexcept;
+
+  /// Points on the reference cell at which an expression need to be
+  /// evaluated in order to interpolate the expression in the finite
+  /// element space. For Lagrange elements the points will just be the
+  /// nodal positions. For other elements the points will typically be
+  /// the quadrature points used to evaluate moment degrees of freedom.
+  /// @return Points on the reference cell. Shape is (num_points, tdim).
+  Eigen::Array<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
+  interpolation_points() const noexcept;
```
michalhabera, line 144: What about Hermite elements? You'd need values of derivatives at some points.

cpp/dolfinx/mesh/cell_types.cpp
```diff
@@ -280,24 +95,19 @@ mesh::CellType mesh::cell_facet_type(mesh::CellType type)
...
-  case mesh::CellType::quadrilateral:
-    return get_entity_vertices_quadrilateral(dim);
-  case mesh::CellType::hexahedron:
-    return get_entity_vertices_hexahedron(dim);
-  default:
-    throw std::runtime_error("Unsupported cell type.");
-    return Eigen::Array<int, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>();
-  }
+  assert(dim >= 0);
+  // assert(dim < basix::topological_dimension(celltype));
```
michalhabera, line 99: Leftover comment.

python/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
 cmake_minimum_required(VERSION 3.10)
 
+set (CMAKE_BUILD_TYPE Debug)
```
michalhabera, line 3: Please no. Debug must be only when requested.

python/test/unit/fem/test_fem_pipeline.py
```diff
@@ -328,22 +128,25 @@ def test_vector_P_tp(family, degree, cell_type, datadir):
...
 
-# TODO: Implement DPC spaces
 @parametrize_cell_types_quad
-@pytest.mark.parametrize("family", ["DQ"])
-# @pytest.mark.parametrize("family", ["DQ", "DPC"])
+@pytest.mark.parametrize("family", ["DQ", "DPC"])
 @pytest.mark.parametrize("degree", [1, 2, 3])
 def test_dP_quad(family, degree, cell_type, datadir):
+    if family == "DPC":
+        pytest.skip()  # These space currently not implemented in basix
```
michalhabera, line 136: pytest.skip() takes a string, do not put this into a comment. And below as well.

python/test/unit/function/test_function.py
```diff
@@ -26,9 +26,9 @@ def mesh():
     return UnitCubeMesh(MPI.COMM_WORLD, 3, 3, 3)
 
 
-@pytest.fixture
-def R(mesh):
-    return FunctionSpace(mesh, ('R', 0))
+# @pytest.fixture
```
michalhabera, line 29: Real element is longer gone, you can remove all related also below.

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -115,37 +113,80 @@ class FiniteElement
...
   std::shared_ptr<const FiniteElement>
   extract_sub_element(const std::vector<int>& component) const;
 
+  /// Check if interpolation into the finite element space is an
+  /// identity operation given the evaluation on an expression at
+  /// specific points, i.e. the degree-of-freedom are equal to point
+  /// evaluations. The function will return `true` for Lagrange
+  /// elements.
+  ///  @return True is interpolation is an identity operation
+  bool interpolation_ident() const noexcept;
```
michalhabera, line 135: This is long referred to as the element mapping being "identity", in UFL. Or do you mean something else?

python/CMakeLists.txt
chrisrichardson: I think this just got left in by mistake!

cpp/dolfinx/fem/CoordinateElement.cpp
jorgensd: The docs for this function can be found at: https://github.com/FEniCS/basix/blob/main/cpp/finite-element.h#L178-L189
I guess we need both 0th and 1st order derivatives to compute the jacobian, so I guess it is correct as is,

cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -24,18 +26,22 @@ class CoordinateElement
...
-  /// @param[in] is_affine Boolean flag indicating affine mapping
-  /// @param[in] evaluate_basis_derivatives
-  CoordinateElement(mesh::CellType cell_type, int topological_dimension,
-                    int geometric_dimension, const std::string& signature,
-                    const ElementDofLayout& dof_layout, bool is_affine,
-                    const std::function<int(double*, int, int, const double*)>&
-                        evaluate_basis_derivatives);
+  /// @param[in] needs_permutation_data Indicates whether or not the element
+  /// needs permutation data (for higher order elements)
+  /// @param[in] get_dof_permutation TODO
```
jorgensd, line 35: Add documentation

cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -51,6 +57,10 @@ class CoordinateElement
   /// Return the topological dimension of the cell shape
   int topological_dimension() const;
 
+  /// TODO
```
jorgensd, line 60: Add docs

cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -79,62 +90,78 @@ FiniteElement::FiniteElement(const ufc_finite_element& ufc_element)
...
-  assert(_evaluate_reference_basis_derivatives);
-  const int num_points = X.rows();
-  int ret = _evaluate_reference_basis_derivatives(reference_values.data(),
-                                                  order, num_points, X.data());
-  if (ret == -1)
-  {
-    throw std::runtime_error("Generated code returned error "
-                             "in evaluate_reference_basis_derivatives");
-  }
+  assert(order == 1); // TODO: fix this for order > 1
```
jorgensd, line 155: Is it particularly hard to generalize this? Seems like the old version supports arbitrary order.

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -154,7 +195,8 @@ class FiniteElement
   int _tdim, _space_dim, _value_size, _reference_value_size;
 
   // Dof coordinates on the reference element
-  Eigen::Array<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor> _refX;
+  //   Eigen::Array<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>
```
jorgensd, line 198: Leftover? should be removed?

cpp/dolfinx/fem/interpolate.h
```diff
@@ -278,7 +309,9 @@ void interpolate_c(
       x.rows(), value_size);
   f(values, x);
 
-  detail::interpolate_values<T>(u, values);
+  // FIXME: This looks strange. Check and add comments
```
jorgensd, line 312: Seems like an appropriate thing to check

cpp/dolfinx/mesh/cell_types.cpp
```diff
@@ -313,17 +123,18 @@ mesh::get_sub_entities(CellType type, int dim0, int dim1)
     throw std::runtime_error(
         "mesh::get_sub_entities supports getting edges (d=1) at present.");
   }
+  // TODO: get this data from basix
```
jorgensd, line 126: Should this be done before a merge?

python/test/unit/function/test_function.py
```diff
@@ -275,7 +275,7 @@ def eval(self, x):
 
 
 @skip_in_parallel
-def test_near_evaluations(R, mesh):
+def xtest_near_evaluations(R, mesh):
```
jorgensd, line 278: Is there a reason for this going from test to xtest?

cpp/dolfinx/fem/CoordinateElement.cpp
garth-wells: It's the identifier for a basix finite element.

jorgensd: As far I can understand Basix only creates the finite element once, as it has to invert a matrix for each element, thus this is cached and identified through an integer.

cpp/dolfinx/fem/FiniteElement.h
garth-wells: But what is 'the element mapping'?

This function returns true for a point evaluation element. I_h(u) = A u(x_i), where  I_h(u) are the dofs, u(x) is the function u evaluated at points {x_i} and A is the interpolation matrix. Function returns true if A = Identity.

cpp/dolfinx/fem/FiniteElement.h
garth-wells: And we support Hermite elements?

This is already a big step forward from 'dof_coordinates', so we don't need to complicate for cases that aren't yet supported. It can be easily generalised later for other cases. There are several ways this could be done.

cpp/dolfinx/fem/FiniteElement.h
michalhabera: It is a map which tells how degrees of freedom transform under a coordinate transformation (reference to physical, or physical to reference, I don't remember the direction). This was included in transform dofs or similar in the past.

Why is this not fetched from the types of dofs then, if it should be all about point-eval dofs?

I think one should be able to show - at least for affine elements - that if all dofs are point-eval then they transform via the "identity".

garth-wells: I think this is nicer than a check on dofs, which we don't have anyway in this class. This function just says that the interpolation operator is the identify, so no need to call the `interpolate` function. I could do something whacky and use 2\delta_{ij} for defining the dofs, in which case one would need to call the interpolate function as it would apply the scaled identity I/2.

python/test/pipeline_tests.py
```diff
@@ -0,0 +1,213 @@
+# Copyright (C) 2019 Jorgen Dokken, Matthew Scroggs and Garth N. Wells
```
jorgensd, line 1: Remember to revert splitting of test files, as we've discussed in person.

cpp/dolfinx/fem/Function.h
```diff
@@ -353,9 +339,13 @@ class Function
       // Compute basis on reference element
       element->evaluate_reference_basis(basis_reference_values, X);
 
+      element->apply_dof_transformation(basis_reference_values.data(),
```
jorgensd, line 342: Add comment on what this does and why it is important.

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -142,14 +142,10 @@ void assemble_matrix(
     // FIXME: cleanup these calls? Some of the happen internally again.
     mesh->topology_mutable().create_entities(tdim - 1);
```
jorgensd, line 143: Could you remove create_entities from here? In `create_connectivity(tdim-1, tdim)` the connectivity is created, ref:
https://github.com/FEniCS/dolfinx/blob/c6ba579c0c09e7ec8c07d55c8c82696d55a8c961/cpp/dolfinx/mesh/Topology.cpp#L196-L197
Would address the fixme

cpp/dolfinx/fem/FiniteElement.cpp
mscroggs: basix does not support mixed elements, so the subelements should be handled separately (I've added this as a comment in the file)

cpp/dolfinx/fem/FiniteElement.h
mscroggs: I guess in future basix could return an int telling which many derivatives are needed at each point. But we can do this in a future PR

python/test/unit/function/test_expression_evaluation.py
```diff
@@ -200,8 +201,7 @@ def test_assembly_into_quadrature_function():
     mesh = dolfinx.UnitSquareMesh(MPI.COMM_WORLD, 3, 6)
 
     quadrature_degree = 2
-    quadrature_points = FIAT.create_quadrature(FIAT.ufc_simplex(
-        mesh.topology.dim), quadrature_degree, scheme="default").get_points()
+    quadrature_points = basix.make_quadrature(basix.CellType.triangle, quadrature_degree)
```
jorgensd, line 204: Even if this is skipped, this is not the correct way of calling `make_quadrature`, ref: https://github.com/FEniCS/basix/blob/main/cpp/quadrature.h#L66

mscroggs: I've created an issue to remind me to fix this test once quadrature elements are fixed: https://github.com/FEniCS/ffcx/issues/298

cpp/dolfinx/mesh/cell_types.cpp
mscroggs: I've made an issue for this one (https://github.com/FEniCS/dolfinx/issues/1338). Not done yet as I think this data is not immediately available in basix yet

cpp/dolfinx/fem/interpolate.h
mscroggs: @garth-wells Did you leave this FIXME?

cpp/dolfinx/fem/FiniteElement.cpp
mscroggs: I've made issue #1339

cpp/dolfinx/fem/CoordinateElement.cpp
mscroggs: I've requested documentation of this in FEniCS/basix#98

cpp/dolfinx/fem/CoordinateElement.cpp
mscroggs: I've requested documentation of this in FEniCS/basix#98

cpp/dolfinx/fem/interpolate.h
garth-wells: I thought this go removed. I have another branch where it will get cleaned up.

jorgensd: This code path is untested as far as I can tell. Should we add a minimal test (interpolation of CG N <-> CG 1 for instance)?

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -5,30 +5,57 @@
...
+    std::function<int(int*, const uint32_t)> permute_dofs,
+    std::function<int(int*, const uint32_t)> unpermute_dofs)
+    : _gdim(geometric_dimension), _signature(signature),
+      _dof_layout(dof_layout), _basix_element_handle(basix_element_handle),
+      _needs_permutation_data(needs_permutation_data),
+      _permute_dofs(permute_dofs), _unpermute_dofs(unpermute_dofs)
 {
+  const mesh::CellType cell = cell_shape();
+  int degree = basix::degree(basix_element_handle);
+  _is_affine
```
michalhabera, line 28: We do export `is_affine` in ufc.h, could it be reused? Preferably both should come from basix alone?

cpp/dolfinx/fem/ElementDofLayout.h
```diff
@@ -43,27 +43,12 @@ class ElementDofLayout
   /// @param[in] parent_map TODO
   /// @param[in] sub_dofmaps TODO
   /// @param[in] cell_type The cell type of the mesh.
-  /// @param[in] base_permutations The base permutations for the dofs on
```
michalhabera, line 46: Is this description now in `basix` somewhere?

cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -79,62 +93,84 @@ FiniteElement::FiniteElement(const ufc_finite_element& ufc_element)
...
-  assert(_evaluate_reference_basis);
-  const int num_points = X.rows();
-  int ret = _evaluate_reference_basis(reference_values.data(), num_points,
-                                      X.data());
-  if (ret == -1)
-  {
-    throw std::runtime_error("Generated code returned error "
-                             "in evaluate_reference_basis");
-  }
+  Eigen::ArrayXXd basix_data = basix::tabulate(_basix_element_handle, 0, X)[0];
```
michalhabera, line 134: Could be const?

cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -79,62 +93,84 @@ FiniteElement::FiniteElement(const ufc_finite_element& ufc_element)
...
+  if (order != 1)
   {
-    throw std::runtime_error("Generated code returned error "
-                             "in evaluate_reference_basis_derivatives");
+    throw std::runtime_error(
+        "FiniteElement::evaluate_reference_basis_derivatives only supports "
+        "order 1 at the moment.");
   }
+
+  std::vector<Eigen::ArrayXXd> basix_data
```
michalhabera, line 166: Could be const vector of const Array?

cpp/dolfinx/fem/Function.h
```diff
@@ -279,29 +278,8 @@ class Function
...
-        sub_f.eval(x, cells, sub_u);
-
-        for (int i = 0; i < sub_value_size; ++i)
-          u.col(offset + i) = sub_u.col(i);
-        offset += sub_value_size;
-      }
-
-      return;
+      throw std::runtime_error("Function::eval is not supported for mixed "
+                               "elements. Extract subsopaces.");
```
michalhabera, line 282: Typo, "subspaces"

cpp/dolfinx/mesh/cell_types.cpp
```diff
@@ -280,24 +95,18 @@ mesh::CellType mesh::cell_facet_type(mesh::CellType type)
...
-    return get_entity_vertices_tetrahedron(dim);
-  case mesh::CellType::quadrilateral:
-    return get_entity_vertices_quadrilateral(dim);
-  case mesh::CellType::hexahedron:
-    return get_entity_vertices_hexahedron(dim);
-  default:
-    throw std::runtime_error("Unsupported cell type.");
-    return Eigen::Array<int, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor>();
-  }
+  assert(dim >= 0);
```
michalhabera, line 98: This should be runtime error, not assert, if you want to check for it. But I would just skip the check.

cpp/dolfinx/mesh/cell_types.cpp
```diff
@@ -280,24 +95,18 @@ mesh::CellType mesh::cell_facet_type(mesh::CellType type)
...
+
+  std::vector<std::vector<int>> topology
```
michalhabera, line 100: Could be const vector of const?

cpp/dolfinx/fem/CoordinateElement.cpp
mscroggs: I've added this to issue #1338

cpp/dolfinx/fem/ElementDofLayout.h
mscroggs: Yes, there's a very thorough description there with examples: https://github.com/FEniCS/basix/blob/a633ad1b5a95f5d1fb62d68a9cd3795104bba96e/cpp/finite-element.h#L230-L307

python/test/pipeline_tests.py
michalhabera, line 1: What are these tests for? They doesn't seem to match pytest layout.

mscroggs: I meant to delete that file (it's leftover from when I separated out some of the slower tests and is now unused)

PR #1327, opened 2021-01-16
Update matplotlib plotting

jorgensd:
The current `dolfinx.plotting.plot` module is heavily outdated, for instance:
- **Bad documentation**: Documentation indicates that both `DirichletBC` and the non-existent `Expression` can be plotted with our backend, which is no longer `True`. This PR updates the docs to match the usage of matplotlib keywords.
- **Unused functions/variables**: Currently, several paths such as `mplot_dirichletbc`, `create_cg1_function_space` and `mplot_expression` are no longer used and should be removed, as done in this PR.
- **No support for quadrilateral meshes**: Prior to this PR, there was no support for plotting quadrilateral meshes. This PR addes limited support. We can now plot the mesh, and plot the function values with scatter plots.
- **Return bools rather than errors**: Many places we have previously returned a mixture of strings and booleans. I suggest that we return errors when a function is not used properly.
- **Minor typos and fixes**: This PR also fixes some minor bugs in the plotting module as well.


garth-wells:
I would happily see Matplotlib removed and placed with vedo (https://vedo.embl.es/) and/or pyvista (https://docs.pyvista.org/). Matplotlib isn't a great experience for meshes, etc.

jorgensd:
> I would happily see Matplotlib removed and placed with vedo (https://vedo.embl.es/) and/or pyvista (https://docs.pyvista.org/). Matplotlib isn't a great experience for meshes, etc.

Then we get back to having a VTK dependency in our code, which we removed from old dolfin. We have a PR for using pyvista:
#1161, which I can try to update. However, I wasn't very happy with the solution, as you needed [special treatment](https://github.com/FEniCS/dolfinx/pull/1161/files#diff-17efd24f40ba6dc4f0840f6baaacd385307e142f60979e0fd1db6c432881bc99R25-R49) for dockerized containers. 

I could give vedo a go over the weekend.

garth-wells:
> > I would happily see Matplotlib removed and placed with vedo (https://vedo.embl.es/) and/or pyvista (https://docs.pyvista.org/). Matplotlib isn't a great experience for meshes, etc.
> 
> Then we get back to having a VTK dependency in our code, which we removed from old dolfin. 

It's only a runtime dependency.

jorgensd:
> I would happily see Matplotlib removed and placed with vedo (https://vedo.embl.es/) and/or pyvista (https://docs.pyvista.org/). Matplotlib isn't a great experience for meshes, etc.

I played around with vedo yesterday, and I like the interface for meshing more than I do for pyvista.
However, some bugs needs to get squashed before getting a general demo working with 3D elements, see https://github.com/marcomusy/vedo/issues/297

I'll give the pyvista model another go tonight, moving some of the modules of #1161 to C++. However, I still think this PR should be merged, as it cleans up the docs and usage of plot (even if we will remove it in the future).

jorgensd:
Matplotlib is being deprecated, see:  #1161

PR #1331, opened 2021-01-18
Fix reference counting of nested PETSc Mat

garth-wells:


garth-wells:
Closing for now to stop triggering the CI twice.

PR #1334, opened 2021-01-18
Return refinement data

chrisrichardson:


cpp/dolfinx/refinement/utils.cpp
```diff
@@ -137,12 +137,15 @@ refinement::compute_edge_sharing(const mesh::Mesh& mesh)
   for (std::size_t i = 0; i < neighbors.size(); ++i)
     proc_to_neighbor.insert({neighbors[i], i});
 
-  std::map<std::int32_t, std::set<int>> shared_edges;
+  std::map<std::int32_t, std::vector<int>> shared_edges;
```
garth-wells, line 140: Could this be made more efficient by using an AdjacencyList?

chrisrichardson: Not really, as it is only over a subset

PR #1342, opened 2021-01-21
Test intel compilers

IgorBaratta:
Tests with classic Intel compilers are disabled due to an issue with pybind.
Once that is solved I can re-enable the tests.

Fix expected to pybind version 2.6.2:
https://github.com/pybind/pybind11/issues/2771

### DIFF
--- a/.github/workflows/intel.yml
+++ b/.github/workflows/intel.yml
@@ -2,9 +2,9 @@ name: C++ testing with Intel compilers
 
 on:
   # Uncomment the below 'push' to trigger on push
-  # push:
-  #  branches:
-  #    - "**"
+  push:
+   branches:
+     - "**"
   schedule:
     # '*' is a special character in YAML, so string must be quoted
     - cron: "0 2 * * TUE"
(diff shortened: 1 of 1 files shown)


garth-wells (commented):
Have we lost the C++ build test with the classic Intel compiler?

jhale:
Last time I tried, there were some template type inferences that we use
quite a lot that Intel doesn't seem to support.

garth-wells:
> Actually, the issue is building basix, it also uses pybind..
> We can install it using gcc, to allow the C++ build test with the classic compiler.

Ah, ok. Let's create an issue so we don't forget it.

garth-wells (approved):

garth-wells:
@jhale There's been quite a bit of work on pybind11 for the Intel classic compilers: https://github.com/pybind/pybind11/pull/2573.

Also see https://github.com/pybind/pybind11/pull/2769 and https://github.com/pybind/pybind11/issues/2771.

jhale:
Last time I checked there was an issue simply building C++ DOLFINX with Intel Classic, e.g. lines like this:

https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfinx/fem/assemble_vector_impl.h#L542

Intel classic couldn't infer the template type of the function from the argument.

garth-wells:
> Last time I checked there was an issue simply building C++ DOLFINX with Intel Classic, e.g. lines like this:
> 
> https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfinx/fem/assemble_vector_impl.h#L542
> 
> Intel classic couldn't infer the template type of the function from the argument.

Intel class does work now for DOLFIN C++. Testing has been disabled because basix uses pybind11, which causes a problem.

PR #1345, opened 2021-01-21
Higher order meshes

### LINKED ISSUE #54: Improved tests for higher order geometry meshes
Need some unit tests for meshes of Tri-6 and Tet-10.

mscroggs:
Correct a few bugs in how higher order meshes are handled.

Added scalar pipeline tests using Unit meshes to be sure there are not broken.

Simplified higher order mesh tests and added 3D tests.

Close #54

### DIFF
--- a/python/test/unit/mesh/test_higher_order_mesh.py
+++ b/python/test/unit/mesh/test_higher_order_mesh.py
@@ -254,7 +254,7 @@ def test_triangle_mesh_vtk(order):
     def coord_to_vertex(x, y):
         return y * (2 * order + 3 - y) // 2 + x
 
-    # Make the cell, following https://blog.kitware.com/modeling-arbitrary-order-lagrange-finite-elements-in-the-visualization-toolkit/
+    # Make the cell, following https://blog.kitware.com/modeling-arbitrary-order-lagrange-finite-elements-in-the-visualization-toolkit/  # noqa: E501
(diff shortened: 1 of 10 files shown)


python/test/unit/mesh/test_higher_order_mesh.py
```diff
@@ -386,168 +296,6 @@ def test_xdmf_input_tri(datadir):
...
-
-    nodes = [0, 3, 10, 11]
-    ref = sympy_scipy(points, nodes, 2 * L, H)
-    assert ref == pytest.approx(intu, rel=1e-6)
-
-
-@skip_in_parallel
-@pytest.mark.parametrize('L', [1, 2])
-@pytest.mark.parametrize('H', [1])
-@pytest.mark.parametrize('Z', [0, 0.3])
```
jorgensd, line 489: Its quite dangerous to remove all the VTK ordering tests here. We should probably and some GMSH and VTK tests to make sure all dof permuations are correct

garth-wells (approved):

python/test/unit/mesh/test_higher_order_mesh.py
mscroggs: VTK tests added in 1a2d470bef1849fa4f49bd44514ed7ca75cce6ae

jorgensd (approved):

PR #1349, opened 2021-01-22
Using STL vector in locate_dofs_topological

### LINKED ISSUE #1348: locate_dofs_topological still using Eigen vector
A recent API change introduced the use of `std::vector` in place of Eigen arrays. `locate_dofs_topological` was left out and still takes an Eigen array as a parameter. This makes it impossible, for example, to pass directly the value returned by `MeshTags::find`, which is a natural concatenation.

massimiliano-leoni:
Closes #1348 

Changing `locate_dofs_topological` to take an `std::vector` as an argument.

### DIFF
--- a/cpp/dolfinx/fem/DirichletBC.cpp
+++ b/cpp/dolfinx/fem/DirichletBC.cpp
@@ -248,7 +248,7 @@ get_remote_bcs2(const common::IndexMap& map0, int bs0,
 //-----------------------------------------------------------------------------
 std::array<std::vector<std::int32_t>, 2> fem::locate_dofs_topological(
     const std::array<std::reference_wrapper<const fem::FunctionSpace>, 2>& V,
-    const int dim, const std::vector<std::int32_t>& entities, bool remote)
+    const int dim, tcb::span<const std::int32_t>& entities, bool remote)
 {
   const fem::FunctionSpace& V0 = V.at(0).get();
   const fem::FunctionSpace& V1 = V.at(1).get();
@@ -374,7 +374,7 @@ std::array<std::vector<std::int32_t>, 2> fem::locate_dofs_topological(
 //-----------------------------------------------------------------------------
 std::vector<std::int32_t>
 fem::locate_dofs_topological(const fem::FunctionSpace& V, const int dim,
-                             const std::vector<std::int32_t>& entities,
+                             tcb::span<const std::int32_t>& entities,
                              bool remote)
 {
   assert(V.dofmap());
--- a/cpp/dolfinx/fem/DirichletBC.cpp
+++ b/cpp/dolfinx/fem/DirichletBC.cpp
@@ -374,7 +374,7 @@ std::array<std::vector<std::int32_t>, 2> fem::locate_dofs_topological(
 //-----------------------------------------------------------------------------
 std::vector<std::int32_t>
 fem::locate_dofs_topological(const fem::FunctionSpace& V, const int dim,
-                             const tcb::span<const int32_t>& entities,
+                             const tcb::span<const std::int32_t>& entities,
                              bool remote)
 {
   assert(V.dofmap());
--- a/cpp/dolfinx/fem/DirichletBC.cpp
+++ b/cpp/dolfinx/fem/DirichletBC.cpp
@@ -248,7 +248,7 @@ get_remote_bcs2(const common::IndexMap& map0, int bs0,
 //-----------------------------------------------------------------------------
 std::array<std::vector<std::int32_t>, 2> fem::locate_dofs_topological(
     const std::array<std::reference_wrapper<const fem::FunctionSpace>, 2>& V,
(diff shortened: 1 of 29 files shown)


cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -409,7 +408,7 @@ fem::locate_dofs_topological(const fem::FunctionSpace& V, const int dim,
   const int num_entity_closure_dofs
       = dofmap->element_dof_layout->num_entity_closure_dofs(dim);
   std::vector<std::int32_t> dofs;
-  for (Eigen::Index i = 0; i < entities.rows(); ++i)
+  for (size_t i = 0; i < entities.size(); ++i)
```
IgorBaratta, line 411: I think the C ++ version `std :: size_t` is preferable.

massimiliano-leoni: Of course, that was unintentional :D
Thanks!

python/dolfinx/wrappers/fem.cpp
```diff
@@ -457,12 +457,14 @@ void fem(py::module& m)
...
-         const int dim, const Eigen::Ref<const Eigen::ArrayXi>& entities,
+         const int dim, const py::array_t<std::int32_t>& entities,
          bool remote) -> std::array<py::array, 2> {
         if (V.size() != 2)
           throw std::runtime_error("Expected two function spaces.");
+        auto sp
+            = tcb::span<const std::int32_t>(entities.data(), entities.size());
         std::array<std::vector<std::int32_t>, 2> dofs
-            = dolfinx::fem::locate_dofs_topological({V[0], V[1]}, dim, entities,
+            = dolfinx::fem::locate_dofs_topological({V[0], V[1]}, dim, sp,
```
garth-wells, line 467: Replace `sp` with `tcb::span(entities.data(), entities.size())`

python/dolfinx/wrappers/fem.cpp
```diff
@@ -471,9 +473,11 @@ void fem(py::module& m)
   m.def(
       "locate_dofs_topological",
       [](const dolfinx::fem::FunctionSpace& V, const int dim,
-         const Eigen::Ref<const Eigen::ArrayXi>& entities, bool remote) {
+         const py::array_t<std::int32_t>& entities, bool remote) {
+        auto sp
+            = tcb::span<const std::int32_t>(entities.data(), entities.size());
         return as_pyarray(
-            dolfinx::fem::locate_dofs_topological(V, dim, entities, remote));
+            dolfinx::fem::locate_dofs_topological(V, dim, sp, remote));
```
garth-wells, line 480: See comment above. Uses modern C++ type inference.

garth-wells (changes requested):
Looks good just some small style comments.

python/dolfinx/wrappers/fem.cpp
massimiliano-leoni: That's how I first tried it, but it complained that

```
fem.cpp:467:15: error: no matching function for call to 'locate_dofs_topological'
DirichletBC.h:53:42: note: candidate function not viable: expects an l-value for 3rd argument
```

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -78,7 +77,7 @@ std::array<std::vector<std::int32_t>, 2> locate_dofs_topological(
 /// with V.
 std::vector<std::int32_t>
 locate_dofs_topological(const fem::FunctionSpace& V, const int dim,
-                        const Eigen::Ref<const Eigen::ArrayXi>& entities,
+                        tcb::span<const std::int32_t>& entities,
```
garth-wells, line 80: Need to add const here.

python/dolfinx/wrappers/fem.cpp
garth-wells: There is a missing `const` - I just posted a comment.

python/dolfinx/wrappers/fem.cpp
```diff
@@ -457,23 +457,26 @@ void fem(py::module& m)
...
   m.def(
       "locate_dofs_topological",
       [](const dolfinx::fem::FunctionSpace& V, const int dim,
-         const Eigen::Ref<const Eigen::ArrayXi>& entities, bool remote) {
-        return as_pyarray(
-            dolfinx::fem::locate_dofs_topological(V, dim, entities, remote));
+         const py::array_t<std::int32_t>& entities, bool remote) {
+        return as_pyarray(dolfinx::fem::locate_dofs_topological(
+            V, dim,
+            tcb::span<const std::int32_t>(entities.data(), entities.size()),
```
garth-wells, line 478: Use `tcb::span(entities.data(), entities.size())`

massimiliano-leoni: Was already on the way!

garth-wells (approved):

PR #1350, opened 2021-01-22
Remove inclusive_scan, does not work on GCC before 9.3.0.

jhale:
As per subject.


IgorBaratta:
I think `std::inclusive_scan` is also used to create topology in `Topology.cpp` ...

garth-wells:
@jhale could you drop in a comment before each change saying the we don't use `std::partial_sum` because of an integer narrowing issue? Otherwise someone might think that `std::partial_sum` is a tidy simplification and make a change.

jhale:
@garth-wells That is already done, also I mention the precise issue why `std::inclusive_scan` isn't use, so that someone can switch back once we're well passed GCC 9 series being in production use.

Thanks @IgorBaratta, I adjusted `Topology.cpp` as well.

garth-wells (approved):

PR #1351, opened 2021-01-22
Just minor style fixes I ran into

massimiliano-leoni:


cpp/dolfinx/fem/utils.cpp
```diff
@@ -244,7 +244,7 @@ fem::create_coordinate_map(ufc_coordinate_mapping* (*fptr)())
 std::shared_ptr<fem::FunctionSpace>
 fem::create_functionspace(ufc_function_space* (*fptr)(const char*),
                           const std::string function_name,
-                          std::shared_ptr<mesh::Mesh> mesh)
+                          std::shared_ptr<dolfinx::mesh::Mesh> mesh)
```
garth-wells, line 247: I don't think we want the `dolfinx` here - we're already in the `dolfinx` namespace. Similarly below.

garth-wells (changes requested):

cpp/dolfinx/fem/utils.cpp
massimiliano-leoni: I agree; I originally wrote it like that but it failed the CI step on documentation [which it shouldn't]. See https://github.com/FEniCS/dolfinx/runs/1750732173 .

massimiliano-leoni: It's fixed now.

garth-wells (approved):

PR #1359, opened 2021-01-25
Move installation of CMake file to ${CMAKE_INSTALL_LIBDIR}

### LINKED ISSUE #1147: C++ tests fail on 32 bit systems
C++ tests and demos are failing on 32 bit systems (i386, armhf).  But python tests are passing, which suggests the 32 bit systems themselves are fine and the problem is in the config file for the tests.

Logs can be found at https://ci.debian.net/packages/d/dolfinx/
e.g. i386 https://ci.debian.net/data/autopkgtest/testing/i386/d/dolfinx/7272770/log.gz
armhf  https://ci.debian.net/data/autopkgtest/unstable/armhf/d/dolfinx/7268765/log.gz

Tests are run through `cmake $srcdir` and  `make tests`
see https://salsa.debian.org/science-team/fenics/dolfinx/-/blob/master/debian/tests/test-dolfinx-cpp

The error message is
```
==== running C++ tests ====
-- The C compiler identification is GNU 10.2.0
-- The CXX compiler identification is GNU 10.2.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/mpicc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at CMakeLists.txt:5 (find_package):
  Could not find a configuration file for package "DOLFINX" that is
  compatible with requested version "".

  The following configuration files were considered but not accepted:

    /usr/share/dolfinx/cmake/DOLFINXConfig.cmake, version: 2019.2.9.99 (64bit)



-- Configuring incomplete, errors occurred!
```
The reference to "64 bit" in the error message suggests to me that the 32/64-bit test in `/usr/share/dolfinx/cmake/DOLFINXConfigVersion.cmake` is not correct for 32-bit systems:
```
# check that the installed version has the same 32/64bit-ness as the one which is currently searching:
if(NOT CMAKE_SIZEOF_VOID_P STREQUAL "8")
  math(EXPR installedBits "8 * 8")
  set(PACKAGE_VERSION "${PACKAGE_VERSION} (${installedBits}bit)")
  set(PACKAGE_VERSION_UNSUITABLE TRUE)
endif()
```

garth-wells:
The location `${CMAKE_INSTALL_LIBDIR}/cmake/dolfinx` is more common and it can contain arch-specific data.

Fixes #1147.


drew-parsons:
Looks good. I'll apply it and confirm things are still running at my end.

drew-parsons:
So there's no misunderstanding, I think the original destination ${CMAKE_INSTALL_DATAROOTDIR}/dolfinx/cmake (i.e. `/usr/share/dolfinx/cmake`) is used automatically by cmake. In https://cmake.org/cmake/help/latest/command/find_package.html#search-procedure it's the entry

    <prefix>/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/
(/usr/share being the 3rd alternative there). So I was wrong to to say in #1147  it wasn't searched without being added manually to CMAKE_MODULE_PATH.

But I think this PR is still sensible since the cmake files contain references to arch-dependent paths (specifically, library paths in INTERFACE_LINK_LIBRARIES. Also, arch dependent paths appear in dolfinx.conf. So CMAKE_INSTALL_LIBDIR is more consistent for these files than CMAKE_INSTALL_DATAROOTDIR

drew-parsons:
There may be an issue with `DOLFINXTargets.cmake`.  It first sets `_IMPORT_PREFIX` based on `_realOrig` and `_realCurr`, and would typically get set to the location of the dolfinx cmake files.

Then we have `get_filename_component` repeated 4 times to pare back the value of `_IMPORT_PREFIX`
```
get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
get_filename_component(_IMPORT_PREFIX "${_IMPORT_PREFIX}" PATH)
```
`_IMPORT_PREFIX` is validated (`/` converts to `""``, or "/usr" is taken as default. Then it is used in `INTERFACE_INCLUDE_DIRECTORIES` as `${_IMPORT_PREFIX}/include`

That fourfold `get_filename_component` looks dodgy. In My tests it results in `/include` being provided for include directories instead of `/usr/include`. I'm looking further into it.

drew-parsons:
As far as I can tell the multiple `get_filename_component` lines are some cmake assery. DOLFINXTargets.cmake is generated by `install(EXPORT DOLFINXTargets)` in cpp/dolfinx/CMakeLists.txt.  Evidentally cmake have made their generated files fragile with respect to relocation.  Bad design if you ask me, but fixing cmake is out of scope for us.

This last trouble is self-inflicted (or if you prefer, cmake-inflicted) by me wanting to move the cmake files between  CMAKE_INSTALL_LIBDIR and CMAKE_INSTALL_DATAROOTDIR in order to keep the debian libdolfinx-dev package arch-independent. I'll just fix my part of the problem in the Debian build scripts by eliminating one of the `get_filename_component` lines in the Debian installed cmake file.

In other words, thumbs up to this PR. It makes sense for _normal_ installation to use CMAKE_INSTALL_LIBDIR.

drew-parsons (approved):
Tested for Debian, cf. https://buildd.debian.org/status/package.php?p=dolfinx

PR #1362, opened 2021-01-27
Safest possible Eigen3 settings

jhale:
Obviously no vectorisation is not ideal but I do not see any easy alternatives for now. Will make same change in basix.


garth-wells:
We're close to eliminating Eigen from the pubic interface, which should fix this issue.

PR #1364, opened 2021-01-27
Remove timing assert

jorgensd:
Remove assert on timing search acceleration by midpointtrees. Seems like it is not a good measure


jorgensd:
Preceded by #1363

PR #1370, opened 2021-01-29
Removed Eigen::Tensor

mscroggs:
Eigen Tensors can't be sliced (only single items can be accessed), and are a huge pain to wrap with pybind. Until a new version of Eigen comes out with proper Tensor support std::vector might be better?


garth-wells (approved):
Fine with me. We just need to be super clear in the docs on the size and shape, and data layout. For interfacing to NumPy it might be better to use `span`, but that can be changed later. Depends a bit on how we want to interface to these function from Python.

PR #1372, opened 2021-01-30
Improve sparsity pattern

chrisrichardson:
* Simplifies internals, by caching owned and unowned _rows_ separately (instead of _columns_)
* Use int32 indices for off-diagonal (unowned columns)
* Creates a new IndexMap for columns after assembly (with full local-to-global map)

At the moment, PETSc Mat just uses nnz per row to initialise, but this change will make it easier to use other libraries which need the full graph to initialise their matrix structures.


cpp/dolfinx/la/SparsityPattern.cpp
```diff
@@ -212,27 +165,20 @@ void SparsityPattern::insert(const tcb::span<const std::int32_t>& rows,
...
-      {
-        if (cols[j] < local_size1)
-          _diagonal_cache[rows[i]].push_back(cols[j]);
-        else
-        {
-          _off_diagonal_cache[rows[i]].push_back(
-              ghosts1[cols[j] - local_size1]);
-        }
-      }
+      for (std::int32_t col : cols)
```
garth-wells, line 175: Could use `insert` here.

cpp/dolfinx/la/SparsityPattern.cpp
```diff
@@ -212,27 +165,20 @@ void SparsityPattern::insert(const tcb::span<const std::int32_t>& rows,
...
+        _cache_owned[row].push_back(col);
+    }
+    else if (row < size0)
+    {
+      for (std::int32_t col : cols)
```
garth-wells, line 180: Could use `insert` here.

cpp/dolfinx/la/SparsityPattern.cpp
```diff
@@ -348,35 +303,74 @@ void SparsityPattern::assemble()
...
       }
       else
       {
-        assert(row_local < (std::int32_t)_off_diagonal_cache.size());
-        _off_diagonal_cache[row_local].push_back(col);
+        // Column index may not exist in column indexmap
+        auto it = global_to_local.insert({col, local_i});
+        if (it.second)
+        {
+          ++local_i;
```
garth-wells, line 314: Would be clearer if this as remove to the bottom of the code block.

cpp/dolfinx/la/SparsityPattern.cpp
```diff
@@ -348,35 +303,74 @@ void SparsityPattern::assemble()
...
-    row.erase(std::unique(row.begin(), row.end()), row.end());
+    const std::vector<std::int32_t>::iterator it_end
+        = std::unique(row.begin(), row.end());
+
+    // Find position of first "off-diagonal" column
+    const std::vector<std::int32_t>::iterator it_diag
+        = std::lower_bound(row.begin(), it_end, local_size1);
+
+    adj_data.insert(adj_data.end(), row.begin(), it_diag);
+    adj_counts[i] += (it_diag - row.begin());
```
garth-wells, line 341: Use `std::distance`.

cpp/dolfinx/la/SparsityPattern.cpp
```diff
@@ -348,35 +303,74 @@ void SparsityPattern::assemble()
...
+    adj_data_off.insert(adj_data_off.end(), it_diag, it_end);
+    adj_counts_off[i] += (it_end - it_diag);
```
garth-wells, line 343: Use `std::distance`.

cpp/dolfinx/la/SparsityPattern.cpp
```diff
@@ -348,35 +303,74 @@ void SparsityPattern::assemble()
...
-      _off_diagonal_cache);
-  std::vector<std::vector<std::int64_t>>().swap(_off_diagonal_cache);
+  std::vector<std::int32_t> adj_offsets(local_size0 + 1);
+  std::partial_sum(adj_counts.begin(), adj_counts.end(),
+                   adj_offsets.begin() + 1);
+  std::vector<std::int32_t> adj_offsets_off(local_size0 + 1);
+  std::partial_sum(adj_counts_off.begin(), adj_counts_off.end(),
+                   adj_offsets_off.begin() + 1);
+
+  // FIXME: after transferring rows, the "correct" row map has no ghosts,
```
garth-wells, line 355: This is a bit cryptic. What's the "original" l2g? I would expect this is just one.

cpp/dolfinx/la/SparsityPattern.cpp
```diff
@@ -348,35 +303,74 @@ void SparsityPattern::assemble()
...
+  // but PETSc Mat needs to know original L2G map.
+  // _index_maps[0] = std::make_shared<common::IndexMap>(comm, local_size0);
+  _diagonal = std::make_shared<graph::AdjacencyList<std::int32_t>>(
+      std::move(adj_data), std::move(adj_offsets));
+
+  LOG(INFO) << "Column ghost size increased from "
```
garth-wells, line 361: Why has it increased? Add reason to log comment.

garth-wells (changes requested):

cpp/dolfinx/la/SparsityPattern.cpp
chrisrichardson: Due to received new columns from other processes - this is noted earlier in the code.

cpp/dolfinx/la/SparsityPattern.cpp
chrisrichardson: I'll expand the explanation here

PR #1373, opened 2021-01-30
Purge PETSc from IO code

garth-wells:


cpp/dolfinx/io/xdmf_function.h
```diff
@@ -30,7 +30,11 @@ namespace xdmf_function
 {
 
 /// TODO
-void add_function(MPI_Comm comm, const fem::Function<PetscScalar>& u,
+void add_function(MPI_Comm comm, const fem::Function<double>& u, const double t,
+                  pugi::xml_node& xml_node, const hid_t h5_id);
+
+/// TODO
```
chrisrichardson, line 36: What is TODO?

chrisrichardson:
I've been keen on doing this for a while... can we now mix I/O of real and complex in the same build then?

cpp/dolfinx/io/xdmf_function.h
garth-wells: No idea it - was already there.

garth-wells:
There is one tiny issue with `FiniteElement` and `ufc_scalar_type_t` still.

PR #1383, opened 2021-02-04
Removed unused or misleading includes

IgorBaratta:
Using  `include-what-you-use`.

Also:
 - Moved some eigen includes to cpp (when it's not part of the interface).


garth-wells (approved):

jhale:
How easy would it be to have this in the CI?

garth-wells:
> How easy would it be to have this in the CI?

I'd favour just running it just periodically. I'm not sure that it's 100% bullet proof or correct, so wouldn't want to get caught trying to work around it to get tests to pass.

IgorBaratta:
> How easy would it be to have this in the CI?

`iwyu` makes some weird recommendations. 
We still need to manually filter the suggestions before applying them.

PR #1388, opened 2021-02-05
Interpolation of functions into abritrary spaces

mscroggs:
Using the mapping information that basix provides directly, we can now interpolation into an arbitrary space.

There is currently quite a lot of expensive memory assignment and copying in `interpolate`, but this will be corrected once we have finished updating basix's public interface.

This PR is coupled with FEniCS/basix#113 and FEniCS/ffcx#307.


cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -194,6 +192,81 @@ class FiniteElement
       _apply_dof_transformation_to_scalar(data, cell_permutation, block_size);
   }
 
+  /// Apply inverse transpose permutation to some data
+  ///
+  /// @param[in,out] data The data to be transformed
+  /// @param[in] cell_permutation Permutation data fro the cell
```
jorgensd, line 198: Typo.

python/test/unit/function/test_interpolation.py
```diff
@@ -47,168 +47,192 @@ def random_point_in_cell(cell_type):
...
     # Randomly number the points and create the mesh
     order = list(range(num_points))
     random.shuffle(order)
     ordered_points = np.zeros(points.shape)
-    ordered_points = np.zeros(points.shape)
     for i, j in enumerate(order):
         ordered_points[j] = points[i]
     cells = np.array([order])
+
+    print(ordered_points, cells)
```
jorgensd, line 111: Remove print

python/test/unit/function/test_interpolation.py
```diff
@@ -47,168 +47,192 @@ def random_point_in_cell(cell_type):
...
     else:
         def f(x):
-            return (x[1], 2 * x[0] ** order, 3 * x[2])
+            return (x[1] ** min(poly_order, 1), 2 * x[0] ** poly_order, 3 * x[2] ** min(poly_order, 2))
 
+    v = Function(V)
     v.interpolate(f)
-    points = [random_point_in_cell(cell_type) for count in range(5)]
+    for d in V.dofmap.cell_dofs(0):
+        print(v.vector[d])
```
jorgensd, line 160: Remove print

cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -292,3 +296,39 @@ bool FiniteElement::needs_permutation_data() const noexcept
   return _needs_permutation_data;
 }
 //-----------------------------------------------------------------------------
+void FiniteElement::map_push_forward(
```
garth-wells, line 299: I's suggest templating these functions on the C++ side, see https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfinx/fem/FiniteElement.h#L187.

garth-wells (changes requested):

PR #1389, opened 2021-02-05
Remove constant from hyperelasticity ufl

jorgensd:
As described in the comments of the PR, we have to carefully introduce dolfinx.Constant to the users, as it should only be used in specific cases, which will vary in the C++ and python layer.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,9 +3,9 @@ version: 2.1
 install-python-components: &install-python-components
   name: Install FEniCS Python components
   command: |
-    pip3 install git+https://github.com/FEniCS/basix.git --upgrade
-    pip3 install git+https://github.com/FEniCS/ufl.git --upgrade
-    pip3 install git+https://github.com/FEniCS/ffcx.git --upgrade
+    pip3 install git+https://github.com/FEniCS/basix.git
+    pip3 install git+https://github.com/FEniCS/ufl.git
+    pip3 install git+https://github.com/FEniCS/ffcx.git
 
 flake8-python-code: &flake8-python-code
   name: Flake8 checks on Python code
--- a/cpp/cmake/post-install/CMakeLists.txt
+++ b/cpp/cmake/post-install/CMakeLists.txt
@@ -4,7 +4,7 @@ DOLFINX has now been installed in
 
     ${CMAKE_INSTALL_PREFIX}
 
-and the demo programs have been installed in
+and demo programs have been installed in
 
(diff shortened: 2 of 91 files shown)


chrisrichardson (approved):

garth-wells:
I there a particular motivation for this change?

jorgensd:
> I there a particular motivation for this change?

As you can in the diff, there was commented code with constant usage (probably from prior to dolfinx::Constant existing. So I figured it is nice to highlight that one can and should use constants, also in C++

michalhabera:
In principle, we should only use `Constant` in places, where user changes the value of it. Or explicitly state this is the intention. We had `Constant` all over in old-FEniCS code and users inherited this from our demos and examples, just putting Constant() around any "constant" value in their forms.
Constant is not recommended for performance reasons as it 1. increases the amount of data that has to be pre-fetched and passed to kernel and 2. is a black-box object for form compiler, so `Constant(2.0) + Constant(1.0)` is computed at each cell, while `2.0 + 1.0` would be evaluated immediately to `3.0` by UFL.

jorgensd:
> In principle, we should only use `Constant` in places, where user changes the value of it. Or explicitly state this is the intention. We had `Constant` all over in old-FEniCS code and users inherited this from our demos and examples, just putting Constant() around any "constant" value in their forms.
> 
> Constant is not recommended for performance reasons as it 1. increases the amount of data that has to be pre-fetched and passed to kernel and 2. is a black-box object for form compiler, so `Constant(2.0) + Constant(1.0)` is computed at each cell, while `2.0 + 1.0` would be evaluated immediately to `3.0` by UFL.

Isn't wrapping E and nu and not lambda and mu exactly enforcing that principle?
These are the two parameters a user would vary in this demo?

garth-wells:
`Constant` (constant in space) is intended for cases where it changes during a simulation, e.g. time `t`, or is not known at compile time, e.g. in a library with precompiled forms or a problem involving a sweep over numerous values of the constants.

jorgensd:
> `Constant` (constant in space) is intended for cases where it changes during a simulation, e.g. time `t`, or is not known at compile time, e.g. in a library with precompiled forms or a problem involving a sweep over numerous values of the constants.

So we have two options then:
1. close this PR as the parameters are known at compilation time
2. Change this PR to remove the commented lines relating to constant in the ufl file.

michalhabera:
With this state of demo I'd vote for 2., just remove the commented `Constant`. If you want to keep the `Constant` then demo must make use of it, either change its value and assemble two problems, or very strongly advertise this.

PR #1390, opened 2021-02-06
Remove Eigen::Vector3d from public interface and simplify BoundingBoxTree

garth-wells:


cpp/dolfinx/geometry/BoundingBoxTree.cpp
```diff
@@ -20,381 +20,331 @@ using namespace dolfinx::geometry;
...
+                                 const tcb::span<const std::int32_t>& entities,
                                  double padding)
     : _tdim(tdim)
 {
-  // Check dimension
-  if (tdim > mesh.topology().dim())
+  if (tdim < 0 or tdim > mesh.topology().dim())
   {
-    throw std::runtime_error("Dimension must be a number between 0 and "
+    throw std::runtime_error("Dimension must be a number between 1 and "
```
jorgensd, line 239: This should still be zero? It is allowed to build a vertex-tree? At least the if test does not fail if tdim==0.

garth-wells: Tests fail if `tdim==0` is not allow, e.g. the test at https://github.com/FEniCS/dolfinx/blob/master/python/test/unit/geometry/test_bounding_box_tree.py#L177 fails.

jorgensd: So what I meant is that the runtime error should not say that dimensions is a number between 1 and tdim, but 0 and tdim

jorgensd (changes requested):

python/dolfinx/wrappers/geometry.cpp
```diff
@@ -54,19 +55,25 @@ void geometry(py::module& m)
...
+                  double padding) {
+                 return dolfinx::geometry::BoundingBoxTree(
+                     mesh, tdim, tcb::span(entities.data(), entities.size()),
+                     padding);
+               }),
            py::arg("mesh"), py::arg("tdim"), py::arg("entity_indices"),
-           py::arg("padding") = 0)
-      .def(py::init<const std::vector<Eigen::Vector3d>&>())
-      .def("num_bboxes", &dolfinx::geometry::BoundingBoxTree::num_bboxes)
+           py::arg("padding") = 0.0)
```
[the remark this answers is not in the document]
garth-wells: Because it's not used in any tests or demos.

jorgensd (approved):

chrisrichardson (approved):

PR #1394, opened 2021-02-08
Remove Eigen::Array<bool, Eigen::Dynamic, 1>  from interface

IgorBaratta:
`std::vector<bool>` is already used extensively for markers.

### DIFF
--- a/python/dolfinx/wrappers/fem.cpp
+++ b/python/dolfinx/wrappers/fem.cpp
@@ -474,7 +474,7 @@ void fem(py::module& m)
       "locate_dofs_geometrical",
       [](const std::vector<
              std::reference_wrapper<const dolfinx::fem::FunctionSpace>>& V,
-         const std::function<Eigen::Array<bool, Eigen::Dynamic, 1>(
+         const std::function<std::vector<bool>(
              const Eigen::Ref<const Eigen::Array<double, 3, Eigen::Dynamic,
                                                  Eigen::RowMajor>>&)>& marker)
           -> std::array<py::array, 2> {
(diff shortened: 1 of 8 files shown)


python/dolfinx/wrappers/fem.cpp
```diff
@@ -474,7 +474,7 @@ void fem(py::module& m)
       "locate_dofs_geometrical",
       [](const std::vector<
              std::reference_wrapper<const dolfinx::fem::FunctionSpace>>& V,
-         const std::function<Eigen::Array<bool, Eigen::Dynamic, 1>(
+         const std::function<std::vector<bool>(
```
michalhabera, line 477: In all python interfaces (also below), we need to have `py::array` of bools, for performance reasons.

michalhabera (changes requested):

garth-wells:
I think we need to get the 2d-array class merged to make this change really work. Otherwise we get a mish mash of EIgen and STL - we can't use pure Eigen functionality and we can't use STL functionality, e.g. `std::transform`.

python/dolfinx/wrappers/fem.cpp
IgorBaratta: `std::vector` allocates its own memory, and doesn't have a `reference` constructor.
I'm having trouble understanding how it could help.

garth-wells: > `std::vector` allocates its own memory, and doesn't have a `reference` constructor.
> I'm having trouble understanding how it could help.

We might have to make a copy in the wrappers, which is probably ok. The point @michalhabera is making is that pybind11 is really slow in handling `list` <-> `std::vector`, so it needs to managed explicitly.

IgorBaratta:
handled elsewhere.

PR #1397, opened 2021-02-08
Create two dimensional array structure

IgorBaratta:
Also removed `Eigen::Array` from mesh (cpp/dolfinx/mesh) interface.
I just worked mesh module (yet), the other files were changed to allow compilation and testing.

We can use the numpy buffer protocol to wrap the 2d array seamlessly as a NumPy array:
https://pybind11.readthedocs.io/en/stable/advanced/pycpp/numpy.html


cpp/dolfinx/common/array_2d.h
```diff
@@ -0,0 +1,187 @@
...
+    _rows = rows;
+    _cols = cols;
+    _storage.resize(rows * cols, val);
+  }
+
+  /// Copies a block of data from other container.
+  template <class Container,
+            typename
+            = typename std::enable_if<is_2d_container<Container>::value>::type>
+  void copy(const Container& other)
```
garth-wells, line 168: Not sure about the name - I associated `copy` with a new object. Maybe `assign` is better?

cpp/dolfinx/common/array_2d.h
```diff
@@ -0,0 +1,187 @@
...
+
+  /// Returns a reference to the row at specified location (i).
+  tcb::span<value_type> row(int i) const
+  {
+    size_type offset = i * _cols;
+    return tcb::span<value_type>(pointer(&_storage[0] + offset), _cols);
+  }
+
+  /// Returns a pointer to the first element of the array.
+  value_type* data() noexcept { return _storage.data(); }
```
garth-wells, line 117: I'd like to put a big warning on the `data` functions in case we introduce offsets to align each row.

garth-wells (changes requested):
- Do move constructors/assignment need to be added?
- Could we introduce a stride to give the stride of the length of each row to allow offsets by row for memory alignment?

cpp/dolfinx/common/array_2d.h
```diff
@@ -0,0 +1,187 @@
...
+           const Allocator& alloc = Allocator())
+      : _rows(rows), _cols(cols)
+  {
+    _storage = std::vector<T, Allocator>(_rows * _cols, value, alloc);
+  }
+
+  /// Constructs a two dimensional array with size = rows * cols, and initialize
+  /// elements with value value.
+  template <class Iterator>
+  array_2d(const std::vector<size_t>& shape, Iterator first, Iterator last)
```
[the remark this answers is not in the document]
IgorBaratta: No, it's a temporary constructor for use with pybind.
I'll add a comment that it's going to be removed.

cpp/dolfinx/common/array_2d.h
```diff
@@ -0,0 +1,187 @@
...
+  using size_type = typename std::vector<T, Allocator>::size_type;
+  using reference = typename std::vector<T, Allocator>::reference;
+  using const_reference = typename std::vector<T, Allocator>::const_reference;
+  using pointer = typename std::vector<T, Allocator>::pointer;
+  using iterator = typename std::vector<T, Allocator>::iterator;
+  using const_iterator = typename std::vector<T, Allocator>::const_iterator;
+  /// \endcond
+
+  /// Constructs an empty two dimensional array.
+  array_2d() : _rows(0), _cols(0) { _storage = std::vector<T, Allocator>(); }
```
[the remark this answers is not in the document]
IgorBaratta: We can have `array_2d() = default;`, or explicitly create and empty array ..
It's used in a few places to return an empty array. 
eg: https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfinx/mesh/cell_types.cpp#L139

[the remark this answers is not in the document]
garth-wells: > We can have `array_2d() = default;`, or explicitly create and empty array ..
> It's used in a few places to return an empty array.
> eg: https://github.com/FEniCS/dolfinx/blob/master/cpp/dolfinx/mesh/cell_types.cpp#L139

It would be handy to be explicit with the row/column size - sometimes even if the array is empty knowing the number of columns (or rows) carries useful information.

cpp/dolfinx/common/array_2d.h
```diff
@@ -0,0 +1,187 @@
...
+      : _rows(rows), _cols(columns)
+  {
+    if (_rows * _cols != list.size())
+      throw std::runtime_error("Dimension mismatch");
+    _storage = std::vector<T, Allocator>(list, alloc);
+  }
+
+  /// Returns a reference to the element at specified location (i, j).
+  /// No bounds checking is performed.
+  reference operator()(size_type i, size_type j)
```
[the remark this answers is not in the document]
IgorBaratta: I was following the std library convention and container access operators are not specified as `noexcept`.
Eg. Accessing out-of-bound index in a vector is undefined behavior (might throw).

But we still  can add  `noexcept`.

cpp/dolfinx/common/array_2d.h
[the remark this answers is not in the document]
IgorBaratta: removed

cpp/dolfinx/common/array_2d.h
```diff
@@ -0,0 +1,187 @@
...
+  const_reference operator()(size_type i, size_type j) const
+  {
+    return _storage[i * _cols + j];
+  }
+
+  /// Returns a reference to the row at specified location (i).
+  tcb::span<value_type> row(int i) const
+  {
+    size_type offset = i * _cols;
+    return tcb::span<value_type>(pointer(&_storage[0] + offset), _cols);
```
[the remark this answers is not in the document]
IgorBaratta: since it's not possible to convert a stl iterator to a pointer, I used instead;

```c++
return tcb::span<const value_type>(std::next(_storage.data(), offset), _cols);
```

cpp/dolfinx/common/array_2d.h
```diff
@@ -0,0 +1,187 @@
...
+  size_type rows() const noexcept { return _rows; }
+
+  /// Returns the array dimensions {cols, rows}.
+  std::pair<size_type, size_type> shape() const noexcept
+  {
+    return {_rows, _cols};
+  }
+
+  /// Changes the number of elements stored.
+  void resize(size_type rows, size_type cols, value_type val = value_type())
```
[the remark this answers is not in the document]
IgorBaratta: removed

cpp/dolfinx/common/array_2d.h
```diff
@@ -0,0 +1,187 @@
...
+  size_type size() const noexcept { return _storage.size(); }
+
+  /// Returns the number of cols in the two-dimensional array.
+  size_type cols() const noexcept { return _cols; }
+
+  /// Returns the number of rows in the two-dimensional array.
+  size_type rows() const noexcept { return _rows; }
+
+  /// Returns the array dimensions {cols, rows}.
+  std::pair<size_type, size_type> shape() const noexcept
```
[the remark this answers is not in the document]
IgorBaratta: I added `rows()` and `cols()` to facilitate the transition from eigen.
But I've jsut added a note that they should be removed.

cpp/dolfinx/common/array_2d.h
IgorBaratta: Changed function name to `assign`.

cpp/dolfinx/common/array_2d.h
IgorBaratta: I think this function is handy for interoperability with eigen and numpy.

For example, iIf we intend to use the buffer protocol to avoid copies, `*ptr=array.data()`
```c++
struct buffer_info {
    void *ptr;
    py::ssize_t itemsize;
    std::string format;
    py::ssize_t ndim;
    std::vector<py::ssize_t> shape;
    std::vector<py::ssize_t> strides;
};
```

garth-wells: Agree, but if we have a stride we don't want the function being used carelessly.

cpp/dolfinx/common/array_2d.h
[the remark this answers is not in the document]
IgorBaratta: added `constexpr`.

cpp/dolfinx/common/array_2d.h
IgorBaratta: A warning in the documentation, then?

IgorBaratta:
> * Do move constructors/assignment need to be added?

I think the default versions do exactly what we want (move all members or copy all members).
We might need to add them if we change the storage type (for example `new`).

IgorBaratta:
> * Could we introduce a stride to give the stride of the length of each row to allow offsets by row for memory alignment?

I'll add the stride, it'll also simplify interoperability with pybind.

Memory alignment by row can be obtained manually by:
```c++
#include <boost/align/aligned_allocator.hpp>

using allocator = boost::alignment::aligned_allocator<double, 3>;
common::array2d<double, allocator> arr(x, 3);
```
We can integrate the allocator and always align by row (if it's known at compile time).

michalhabera:
> Memory alignment by row can be obtained manually by:
> 
> ```c++
> #include <boost/align/aligned_allocator.hpp>
> 
> using allocator = boost::alignment::aligned_allocator<double, 3>;
> common::array2d<double, allocator> arr(x, 3);
> ```
> 
> We can integrate the allocator and always align by row (if it's known at compile time).

What we need is to be able to specify alignment in bytes, such that every row, regardless of array's dimensions is aligned to this boundary. This could be achieved with aligning the entire underlying memory and padding the rows. For example, an array of three doubles with rows aligned to 32 bytes would need to be padded,
```
d1 d1 d1 d1 d1 d1 d1 d1, d2 d2 d2 d2 d2 d2 d2 d2, d3 d3 d3 d3 d3 d3 d3 d3, 0 0 0 0 0 0 0 0 \newrow
d4 d4 d4 d4 d4 d4 d4 d4, d5 d5 d5 d5 d5 d5 d5 d5, d6 d6 d6 d6 d6 d6 d6 d6, 0 0 0 0 0 0 0 0 \newrow
```
where `d1, d2, ...` represent bytes of the double entries. If you use the aligned allocator and make sure first `d1` is aligned to 32 bytes, then also all rows starting with `d4, d7, d10, ...` will be.

What @garth-wells probably means with stride is to encode this padding above. For the above say: chunks of size 3 doubles with offsets/padding of 1 double as a view into 1-D memory of 8 doubles.

garth-wells:
> What @garth-wells probably means with stride is to encode this padding above. For the above say: chunks of size 3 doubles with offsets/padding of 1 double as a view into 1-D memory of 8 doubles.

Yes, that's what I had in mind. I'm hoping that with stride info that NumPy can still wrap the array and it will appear as nothing more that a 2D array in the Python interface.

cpp/dolfinx/common/array_2d.h
```diff
@@ -0,0 +1,187 @@
...
+    return tcb::span<value_type>(pointer(&_storage[0] + offset), _cols);
+  }
+
+  /// Returns a pointer to the first element of the array.
+  value_type* data() noexcept { return _storage.data(); }
+
+  /// Returns a const pointer to the first element of the array.
+  const value_type* data() const noexcept { return _storage.data(); };
+
+  /// Returns an iterator to the first element of the array.
```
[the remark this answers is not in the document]
IgorBaratta: Removed.
I think it's possible but we need to define what the deference of a row iterator would return. 
`array2d::row(i)` returns a span to row i..

IgorBaratta:
@michalhabera   thanks for the clarification.

The boost allocator can align the data arbitrarily (32 bytes = aligned_allocator <double, 4>;). We would just need to pad.
Regarding the stride it would be just the current stride + padding size;

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -294,17 +284,30 @@ void mesh(py::module& m)
 
   m.def("locate_entities",
         [](const dolfinx::mesh::Mesh& mesh, int dim,
-           const std::function<std::vector<bool>(
-               const dolfinx::common::array2d<double>&)>& marker) {
-          return as_pyarray(dolfinx::mesh::locate_entities(mesh, dim, marker));
+           const std::function<py::array_t<bool>(
+               py::array_t<double, py::array::c_style>&)>& marker) {
+          auto cpp_marker = [&](const dolfinx::common::array2d<double>& x) {
```
garth-wells, line 312: Let's be explicit with the capture - look like only `marker` needs to be captured here.

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -98,8 +98,10 @@ void mesh(py::module& m)
       "cell_normals",
       [](const dolfinx::mesh::Mesh& mesh, int dim,
          const py::array_t<std::int32_t, py::array::c_style>& entities) {
-        return dolfinx::mesh::cell_normals(
+        auto normals = dolfinx::mesh::cell_normals(
             mesh, dim, tcb::span(entities.data(), entities.size()));
+        return py::array_t<double>(normals.shape, normals.strides(),
```
garth-wells, line 103: If we add something like the `as_pyarray` function but for `common::array2d` we can avoid copying the data.

cpp/dolfinx/geometry/BoundingBoxTree.cpp
```diff
@@ -39,23 +40,27 @@ compute_bbox_of_entity(const mesh::Mesh& mesh, int dim, std::int32_t index)
 {
   // Get the geometrical indices for the mesh entity
   const int tdim = mesh.topology().dim();
-  const Eigen::Array<double, Eigen::Dynamic, 3, Eigen::RowMajor>& geom_dofs
-      = mesh.geometry().x();
+  // Use eigen map for now.
```
michalhabera, line 43: Are these meant to be more like `FIXME` notes?

cpp/dolfinx/mesh/cell_types.cpp
```diff
@@ -120,25 +119,22 @@ mesh::get_sub_entities(CellType type, int dim0, int dim1)
...
-  const static Eigen::Array<int, 4, 3, Eigen::RowMajor> tetrahedron
-      = (Eigen::Array<int, 4, 3, Eigen::RowMajor>() << 0, 1, 2, 0, 3, 4, 1, 3,
-         5, 2, 4, 5)
-            .finished();
-  const static Eigen::Array<int, 6, 4, Eigen::RowMajor> hexahedron
-      = (Eigen::Array<int, 6, 4, Eigen::RowMajor>() << 0, 1, 3, 5, 0, 2, 4, 8,
-         1, 2, 6, 9, 3, 4, 7, 10, 5, 6, 7, 11, 8, 9, 10, 11)
-            .finished();
-
+  dolfinx::common::array2d<int> triangle({{0, 1, 2}});
```
michalhabera, line 122: Could this be just a `std::array<std::array>`? No need for dynamic memory.

cpp/dolfinx/geometry/BoundingBoxTree.cpp
IgorBaratta: Yes. I'll add `FIXME` flag.

cpp/dolfinx/mesh/cell_types.cpp
IgorBaratta: Yep,  but I think it would complicate the return type `std::array<std::array<int, N>, M>` with M, N known at compile time (auto would fail).
Maybe we could handle this in a different PR?

garth-wells (approved):

garth-wells:
I think this is good to go - some refinements can be made later as we go.

PR #1398, opened 2021-02-08
Wrap some mesh/form data as numpy arrays

jorgensd:
To avoid copying data when using these functions with numba, we expose them as numpy arrays. This is because  numba is not fond of "reflected lists", aka. []/std::vector.

### DIFF
--- a/python/dolfinx/wrappers/fem.cpp
+++ b/python/dolfinx/wrappers/fem.cpp
@@ -65,10 +65,10 @@ void fem(py::module& m)
         "Create a sparsity pattern for bilinear form.");
   m.def("pack_coefficients",
         &dolfinx::fem::pack_coefficients<dolfinx::fem::Form<PetscScalar>>,
-        "Pack coefficients for a UFL form.");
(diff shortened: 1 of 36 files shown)


garth-wells (approved):

python/dolfinx/wrappers/fem.cpp
```diff
@@ -69,12 +69,22 @@ void fem(py::module& m)
...
-        &dolfinx::fem::pack_constants<dolfinx::fem::Expression<PetscScalar>>,
-        "Pack constants for a UFL expression.");
+  m.def(
+      "pack_constants",
+      [](const dolfinx::fem::Form<PetscScalar>& form) {
+        return as_pyarray(
+            dolfinx::fem::pack_constants<dolfinx::fem::Form<PetscScalar>>(
+                form));
+      },
+      "Pack constants for a UFL form.");
```
garth-wells, line 79: It's just a Form - not a UFL form.

garth-wells (approved):

PR #1405, opened 2021-02-12
Fix returntypes for dolfinx::array2d

jorgensd:
To be able to use dolfinx::array2d in python the need to be wrapped up as py:array_t.


garth-wells (changes requested):
Need to use `as_pyarray2d` to avoid unnecessary copies.

IgorBaratta (approved):

garth-wells (approved):

PR #1410, opened 2021-02-15
Remove all Eigen from the public interface

### LINKED ISSUE #1381: Remove Eigen from public interface
Having Eigen in the public interface can cause all sorts of problems around alignment. Eigen has some compiler-time settings that set the alignment, but these are not necessarily common across libraries that are linked. Aligned Eigen objects use different 'free' functions, and mixing alignments leads to segfaults.

Since two upstream compiled dependencies may use different alignments the only robust solution appears to be removing Eigen from the public interface.

garth-wells:
Fixes #1381.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,6 +6,7 @@ install-python-components: &install-python-components
     pip3 install git+https://github.com/FEniCS/basix.git
     pip3 install git+https://github.com/FEniCS/ufl.git
     pip3 install git+https://github.com/FEniCS/ffcx.git
+    pip3 install sphinx==3.4.3
 
 flake8-python-code: &flake8-python-code
   name: Flake8 checks on Python code
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -40,6 +40,7 @@ jobs:
           python3 -m pip install git+https://github.com/FEniCS/basix.git
           python3 -m pip install git+https://github.com/FEniCS/ufl.git
           python3 -m pip install git+https://github.com/FEniCS/ffcx.git
+          python3 -m pip install sphinx --upgrade
 
       - name: Flake8 checks
         run: |
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -40,7 +40,7 @@ jobs:
           python3 -m pip install git+https://github.com/FEniCS/basix.git
           python3 -m pip install git+https://github.com/FEniCS/ufl.git
           python3 -m pip install git+https://github.com/FEniCS/ffcx.git
-          python3 -m pip install sphinx --upgrade
+          python3 -m pip install sphinx==3.4.3
 
       - name: Flake8 checks
         run: |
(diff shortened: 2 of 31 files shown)


python/dolfinx/wrappers/geometry.cpp
```diff
@@ -39,7 +37,46 @@ void geometry(py::module& m)
...
+              for (py::ssize_t j = 0; j < qx.shape(1); j++)
+                _q(i, j) = qx(i, j);
+          }
+          else
+            throw std::runtime_error("Array has wrong ndim.");
+
+          return dolfinx::geometry::compute_distance_gjk(_p, _q);
+        });
+
+  //   m.def("compute_distance_gjk", &dolfinx::geometry::compute_distance_gjk);
```
jorgensd, line 79: Leftover comment

.circleci/config.yml
```diff
@@ -6,6 +6,7 @@ install-python-components: &install-python-components
     pip3 install git+https://github.com/FEniCS/basix.git
     pip3 install git+https://github.com/FEniCS/ufl.git
     pip3 install git+https://github.com/FEniCS/ffcx.git
+    pip3 install sphinx==3.4.3
```
jorgensd, line 9: What do we need sphinx for?

.github/workflows/ccpp.yml
```diff
@@ -40,6 +40,7 @@ jobs:
           python3 -m pip install git+https://github.com/FEniCS/basix.git
           python3 -m pip install git+https://github.com/FEniCS/ufl.git
           python3 -m pip install git+https://github.com/FEniCS/ffcx.git
+          python3 -m pip install sphinx==3.4.3
```
jorgensd, line 43: Same as above

cpp/dolfinx/geometry/BoundingBoxTree.cpp
```diff
@@ -54,9 +51,9 @@ compute_bbox_of_entity(const mesh::Mesh& mesh, int dim, std::int32_t index)
       = mesh::entities_to_geometry(mesh, dim, entity, false);
   tcb::span<const int> entity_vertex_indices = vertex_indices.row(0);
 
-  const Eigen::Vector3d x0 = geom_dofs.row(entity_vertex_indices[0]);
+  // const Eigen::Vector3d x0 = geom_dofs.row(entity_vertex_indices[0]);
```
jorgensd, line 54: Leftover comment

cpp/dolfinx/mesh/utils.cpp
```diff
@@ -448,25 +434,31 @@ mesh::entities_to_geometry(const mesh::Mesh& mesh, int dim,
...
-      for (auto j : xc)
-        midpoint += geom_dofs.row(j).matrix().transpose();
+      for (std::int32_t j : xc)
+        for (int k = 0; k < 3; ++k)
+          midpoint[k] += geom_dofs(j, k);
       midpoint /= xc.size();
 
       // Compute vector triple product of two edges and vector to midpoint
-      Eigen::Vector3d p0 = geom_dofs.row(entity_geometry(i, 0));
+      // Eigen::Vector3d p0 = geom_dofs.row(entity_geometry(i, 0));
```
jorgensd, line 443: Leftover comment

jorgensd (changes requested):

.circleci/config.yml
garth-wells: Building the docs on the CI.

Sphinx was upgraded yesterday and containers have new version, but the new release has a bug (it's on the Sphinx issue tracker).

PR #1412, opened 2021-02-15
Make eigen dependency private

IgorBaratta:
should `EIGEN_MAX_ALIGN_BYTES` be private too?
**EDIT**: EIGEN_MAX_ALIGN_BYTES removed.


garth-wells:
I think we can remove `EIGEN_MAX_ALIGN_BYTES` now and let Eigen do its default thing.

python/test/unit/fem/test_assemble_cppimport.py
```diff
@@ -25,10 +25,11 @@
 def test_eigen_assembly(tempdir):  # noqa: F811
     """Compare assembly into scipy.CSR matrix with PETSc assembly"""
     def compile_eigen_csr_assembler_module():
+        eigen_dir = "/usr/include/eigen3"
```
garth-wells, line 28: This will break when Eigen is installed elsewhere. Does Eigen support `pkgconfig`?

jhale:
Forcing the eigen alignment bytes can be removed now. People can use eigen
data structures internally as they please.

python/test/unit/fem/test_assemble_cppimport.py
IgorBaratta: I think so, 
`eigen_dir = dolfinx.pkgconfig.parse("eigen3")["include_dirs"]` provides the eigen include dir.

garth-wells: My system doesn't have a pkgconfig file for Eigen, so we your elegantly skip the test if `dolfinx.pkgconfig.parse` doesn't find Eigen.

IgorBaratta:
Needs some work on `basix` first.
The header `basix.h` doesn't include eigen, but  the basix library has it on its interface.

python/test/unit/fem/test_assemble_cppimport.py
garth-wells: I found `eigen3.pc` - pkg-config path was messed up.

IgorBaratta:
I think this is now ready for review since `eigen` has been removed from `basix`.

garth-wells (approved):

PR #1417, opened 2021-02-17
Allow overlapping boundaries in boundary conditions

garth-wells:
This is an updated version of #1215 by @massimiliano-leoni.

It handles the case the a Dirichlet bc is set twice, which is not uncommon at corners. This change preserves the old-DOLIN behaviour where the last set boundary condition in the list rules.

This change requires a PETSc flush in switching from add to set. There is scope for some small changes that will eliminate a communication set.


python/dolfinx/wrappers/fem.cpp
```diff
@@ -309,40 +309,42 @@ void fem(py::module& m)
...
           dolfinx::fem::assemble_matrix(
-              dolfinx::la::PETScMatrix::add_block_expand_fn(
+              dolfinx::la::PETScMatrix::set_block_expand_fn(
                   A, a.function_spaces()[0]->dofmap()->bs(),
-                  a.function_spaces()[1]->dofmap()->bs()),
+                  a.function_spaces()[1]->dofmap()->bs(), ADD_VALUES),
               a, rows0, rows1);
         });
-  m.def("add_diagonal",
+  m.def("set_diagonal",
```
michalhabera, line 340: This should be called "insert_diagonal" here in python, since "set_diagonal" is reserved in c++ for the method which accepts INSERT_MODE. Then add a little python docstring to this pybind wrapper, since it does something new comparing to c++.

python/dolfinx/wrappers/la.cpp
```diff
@@ -151,6 +152,10 @@ void la(py::module& m)
         return ret;
       },
       "Gather an (ordered) list of sub vectors from a block vector.");
+
```
michalhabera, line 155: Remove commented code.

michalhabera (changes requested):

python/dolfinx/wrappers/fem.cpp
michalhabera: @massimiliano-leoni  Can you fix these?

massimiliano-leoni: @michalhabera on it :D

cpp/dolfinx/la/PETScMatrix.cpp
```diff
@@ -212,20 +212,20 @@ PETScMatrix::add_fn(Mat A)
...
-  return [A, cache = std::vector<PetscInt>()](
+  return [A, mode, cache = std::vector<PetscInt>()](
              std::int32_t m, const std::int32_t* rows, std::int32_t n,
              const std::int32_t* cols, const PetscScalar* vals) mutable {
     PetscErrorCode ierr;
 #ifdef PETSC_USE_64BIT_INDICES
     cache.resize(m + n);
     std::copy_n(rows, m, cache.begin());
-    std::copy_n(cols, n, cache.begin() + m);
+    std::copy_n(cols, n, std::next(cache.begin(), m));
```
michalhabera, line 224: @garth-wells Do you know if the `std::next` is needed here?

michalhabera, line 224: @garth-wells Do you know if the `std::next` is needed here?

michalhabera (approved):

cpp/dolfinx/la/PETScMatrix.cpp
chrisrichardson: It's seldom really "needed", just a way of showing you are adding to an iterator.

PR #1422, opened 2021-02-18
Interpolate into arbitrary spaces

mscroggs:
This allows for the interpolation of functions into arbitrary spaces, using the data made available by Basix.

(This is the same as #1388 but using the new Basix interface.)

### DIFF
--- a/python/test/unit/function/test_interpolation.py
+++ b/python/test/unit/function/test_interpolation.py
@@ -21,10 +21,10 @@ parametrize_cell_types = pytest.mark.parametrize(
                   CellType.quadrilateral, CellType.hexahedron])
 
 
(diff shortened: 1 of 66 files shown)


python/test/unit/function/test_interpolation.py
```diff
@@ -21,10 +21,10 @@
                   CellType.quadrilateral, CellType.hexahedron])
 
 
-def random_point_in_cell(cell_type):
+def random_point_in_reference(cell_type):
```
michalhabera, line 24: I think tests should be deterministic.

mscroggs: I've added `random.seed` to the tests, which I think does this

Issue #350, opened 2019-03-20
Writing DG vectors (dim >= 3) to XDMF gives node-centered fields


tianyikillua:
Correct behavior when writing a **scalar** DG field to a XDMF file

``` xml
...
<Attribute Name="Test" AttributeType="Scalar" Center="Cell">
    <DataItem Dimensions="8 1" Format="HDF">u.h5:/VisualisationVector/0</DataItem>
</Attribute>
...
```

![scalar](https://user-images.githubusercontent.com/4027283/54682911-8259e480-4b10-11e9-9f13-224f0b2e8258.png)

However writing a **vector** DG field returns an **interpolated** nodal field

``` xml
...
<Attribute Name="Test" AttributeType="Vector" Center="Node">
    <DataItem Dimensions="8 3" Format="HDF">u.h5:/VisualisationVector/0</DataItem>
</Attribute>
...
```

![vector](https://user-images.githubusercontent.com/4027283/54682917-86860200-4b10-11e9-9ca7-eadeb26348ec.png)

MWE à la `dolfinx`
``` python
from dolfin import *
from dolfin.io import XDMFFile
import numpy as np

mesh = RectangleMesh(
    MPI.comm_world,
    [Point(0, 0)._cpp_object, Point(1, 1)._cpp_object], [2, 2],
    CellType.Type.triangle, dolfin.cpp.mesh.GhostMode.none)
cmap = dolfin.fem.create_coordinate_map(mesh.ufl_domain())
mesh.geometry.coord_mapping = cmap
num_cells = 8

# Write a scalar DG field to a XDMF file
# Correct behavior, Center="Cell" in XDMF
V = FunctionSpace(mesh, ("DG", 0))
u = Function(V, name="Test")
values = np.arange(num_cells)
u.vector()[:] = values.flatten()

with XDMFFile(MPI.comm_world, "u.xdmf") as f:
    f.write(u, 0)

# Write a vector (dim >= 3) DG field to a XDMF file
# With dim = 2, correct behavior, Center="Cell" in XDMF
# With dim >= 3, wrong behavior, Center="Node" in XDMF
dim = 3
V = VectorFunctionSpace(mesh, ("DG", 0), dim=dim)
v = Function(V, name="Test")
values = np.zeros((num_cells, dim))
values[:, 0] = np.arange(num_cells)
values[:, 1] = -np.arange(num_cells)

v.vector()[:] = values.flatten()

[...]

PR #1423, opened 2021-02-18
Fix block-size for cell-centered data in XDMF IO

michalhabera:
Fixes a bug which forced any function with block size > 1 be evaluated into "point values" (i.e. interpolated into CG1) even if the underlying element was DG0.


jorgensd:
Would close https://github.com/FEniCS/dolfinx/issues/350

garth-wells (approved):

PR #1425, opened 2021-02-23
Avoid expensive MPI call in sparsity pattern

IgorBaratta:


IgorBaratta:
> Do we still need `ghost_owners1` if we are not making an indexmap?

Good point. I don't think we need it.
And we can reduce the size of data to be sent.

garth-wells (approved):

PR #1428, opened 2021-03-02
add adjacency nested list to python for mesh partitioning

jorgensd:
Needed to specify ghost cells in shared facet mode with custom partitioner

### DIFF
--- a/python/dolfinx/wrappers/graph.cpp
+++ b/python/dolfinx/wrappers/graph.cpp
@@ -32,6 +32,9 @@ void declare_adjacency_list(py::module& m, std::string type)
         std::vector<T> data(adj.data(), adj.data() + adj.size());
         return dolfinx::graph::build_adjacency_list<T>(std::move(data), dim);
       }))
+      .def(py::init([](const std::vector<std::vector<T>> adj) {
+        return dolfinx::graph::AdjacencyList<T>(adj);
+      }))
       .def(
           "links",
           [](const dolfinx::graph::AdjacencyList<T>& self, int i) {
--- a/python/dolfinx/wrappers/graph.cpp
+++ b/python/dolfinx/wrappers/graph.cpp
@@ -32,9 +32,14 @@ void declare_adjacency_list(py::module& m, std::string type)
         std::vector<T> data(adj.data(), adj.data() + adj.size());
         return dolfinx::graph::build_adjacency_list<T>(std::move(data), dim);
       }))
-      .def(py::init([](const std::vector<std::vector<T>> adj) {
-        return dolfinx::graph::AdjacencyList<T>(adj);
-      }))
+      .def(py::init(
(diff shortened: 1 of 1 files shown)


garth-wells:
The PR introduced a very expensive operation. What is it you need to do?

jorgensd:
> The PR introduced a very expensive operation. What is it you need to do?
Given a mesh of `14` cells, we would like to have to the following partitioning: 

ghost_mesh
To be able to specify this custom partitioning with `shared_facet` as ghost mode, we need to supply information about ghost cells.
Below is the partitioner for this current mesh.
```
def partitioner(*args):
    midpoints = np.mean(x_global[cells], axis=1)
    dest = []
    for i, midpoint in enumerate(midpoints):
        if MPI.COMM_WORLD.rank == 0:
            dest = [[0], [0, 1], [0], [0, 1], [0], [0], [1, 0]]

        elif MPI.COMM_WORLD.rank == 1:
            dest = [[1], [0], [0], [0], [0], [0], [1, 0]]

    return dolfinx.cpp.graph.AdjacencyList_int32(dest)


ghost_mode = dolfinx.cpp.mesh.GhostMode.shared_facet
new_mesh = dolfinx.mesh.create_mesh(MPI.COMM_WORLD, cells, x, domain, ghost_mode, partitioner)
``` 
We observe that the number of links per cell varies, if the cell is a ghost on another process, it has two entries, if only local on one process only one entry. Therefore we need an adjacency-list with variable length.

garth-wells:
Use a data array and an offset array.

Making available `std::vector`, and even worse `std::vector<std::vector>`, interfaces for things that can be big will lead to people writing very slow code.

python/dolfinx/wrappers/graph.cpp
```diff
@@ -32,6 +32,14 @@ void declare_adjacency_list(py::module& m, std::string type)
         std::vector<T> data(adj.data(), adj.data() + adj.size());
         return dolfinx::graph::build_adjacency_list<T>(std::move(data), dim);
       }))
+      .def(py::init(
+          [](const py::array_t<T, py::array::c_style>& array,
+             const py::array_t<std::int32_t, py::array::c_style>& displ) {
+            std::vector<T> data(array.data(), array.data() + array.size());
```
garth-wells, line 38: Create the vectors directly in the AdjacencyList argument list or use std::move to void two copies.

garth-wells (changes requested):

python/dolfinx/wrappers/graph.cpp
IgorBaratta: Changed to:
> ```c++
> std::vector data(array.data(), array.data() + array.size());
> std::vector offsets(displ.data(), displ.data() + displ.size());
> return dolfinx::graph::AdjacencyList<T>(std::move(data), std::move(offsets));
> ```

garth-wells (approved):

PR #1429, opened 2021-03-03
Use xtensor in topology computation

chrisrichardson:
Uses fixed width array2d instead of AdjacencyList, which should be more efficient.


jorgensd:
> Uses fixed width array2d instead of AdjacencyList, which should be more efficient.

It should be more efficient, but has to be reverted once/if we consider mixed-topologies, right?

chrisrichardson:
Everything would need to be revised for mixed topologies, but in any case, triangular and quadrilateral facets would be treated as separate (fixed-width) lists.

PR #1431, opened 2021-03-04
Add ndspan and improve array2d

IgorBaratta:
- add `ndarray` (up to 3 dimensions)
- add some tests
- remove unnecessary copies from python binders


garth-wells:
Started using xtensor instead.

PR #1435, opened 2021-03-08
Scatter la:Vector

chrisrichardson:
Allows forward and reverse scatter of `la::Vector` performing ghost updates without need for external libraries

### DIFF
--- a/cpp/dolfinx/fem/DofMap.cpp
+++ b/cpp/dolfinx/fem/DofMap.cpp
@@ -97,7 +97,7 @@ fem::DofMap build_collapsed_dofmap(MPI_Comm comm, const DofMap& dofmap_view,
   std::vector<std::int64_t> global_index_remote(
       dofmap_view.index_map->num_ghosts());
   dofmap_view.index_map->scatter_fwd(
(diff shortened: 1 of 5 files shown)


IgorBaratta (approved):

cpp/dolfinx/fem/DofMap.cpp
```diff
@@ -93,8 +93,12 @@ fem::DofMap build_collapsed_dofmap(MPI_Comm comm, const DofMap& dofmap_view,
 
   // Send new global indices for owned dofs to non-owning process, and
   // receive new global indices from owner
-  std::vector global_index_remote
-      = dofmap_view.index_map->scatter_fwd(global_index, 1);
+
+  std::vector<std::int64_t> global_index_remote(
+      dofmap_view.index_map->num_ghosts());
+  dofmap_view.index_map->scatter_fwd(
+      tcb::span<const std::int64_t>(global_index),
```
garth-wells, line 100: I think this could be just `global_index`.

chrisrichardson: gcc won't compile that.

PR #1438, opened 2021-03-09
Use tcb::span in local_to_global and global_to_local

IgorBaratta:


cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -446,11 +446,12 @@ const std::vector<std::int64_t>& IndexMap::ghosts() const noexcept
   return _ghosts;
 }
 //-----------------------------------------------------------------------------
-void IndexMap::local_to_global(const std::int32_t* local, int n,
-                               std::int64_t* global) const
+void IndexMap::local_to_global(tcb::span<const std::int32_t> local,
```
garth-wells, line 449: Use `const` references here - save creating a new span if it's not needed.

IgorBaratta: done!

garth-wells: I think you can make both arguments const references.

chrisrichardson: span is very light, just 2 x int64 - is it worth taking a const ref?

garth-wells (approved):

chrisrichardson (approved):

PR #1440, opened 2021-03-10
Add ADIOS2 to Dockerfile

jorgensd:
Build ADIOS in Dockerfile.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -352,7 +352,10 @@ ONBUILD ARG DOLFINX_CMAKE_CXX_FLAGS
 
 # The dolfinx-onbuild container expects to have folders basix/ ufl/ ffcx/ and
 # dolfinx/ mounted/shared at /src.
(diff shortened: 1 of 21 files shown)


docker/Dockerfile
```diff
@@ -158,6 +159,16 @@ RUN cd /usr/local && \
     cmake -G Ninja -DCMAKE_INSTALL_PREFIX=${KAHIP_DIR} -DNONATIVEOPTIMIZATIONS=On .. && \
     ninja install
 
+# Install ADIOS2
+ENV ADIOS2_DIR=/usr/local/ADIOS2/deploy
```
garth-wells, line 163: The steps should be modelled on the PETSc/SLEPc install and cleanup steps rather than Gmsh (Gmsh is a binary install). Also get just the desired tag directly and the git --single-branch option.

garth-wells (changes requested):

PR #1442, opened 2021-03-10
Update basix installation

jorgensd:
Use the new two-stage installation

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -331,7 +331,10 @@ ONBUILD ARG DOLFINX_CMAKE_CXX_FLAGS
 
 # The dolfinx-onbuild container expects to have folders basix/ ufl/ ffcx/ and
 # dolfinx/ mounted/shared at /src.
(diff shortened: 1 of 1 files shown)


jhale:
I already tried but got it wrong anyway. I'll clean up the mess and push back to here.

jhale:
We're getting a runtime linking error on importing dolfinx.cpp.

Should we be linking to `basix` here?

https://github.com/FEniCS/dolfinx/blob/master/python/CMakeLists.txt#L45

The tests seem to run so I'm a little confused.

garth-wells:
It might be a problem with JIT - probably need to add a Basix pkg-config file, and have dolfinx.pc use it too.

Can you run a test that doesn't call any JIT? (JIT is called to build the coordinate map for meshes).

jhale:
That may be an issue, but this error message is quite explicit that it comes from simply importing `dolfinx.cpp`:

https://github.com/FEniCS/dolfinx/runs/2076621407?check_suite_focus=true#step:10:13

PR #1443, opened 2021-03-10
Try linking with basix explicitly

jhale:


garth-wells:
This doesn't look like the correct fix. I can't see that basix is exposed in the Python wrappers, so the DOLFINX target should be bringing in any targets that it depends on.

PR #1448, opened 2021-03-11
Use an AjacencyList for shared indices in IndexMap

IgorBaratta:
- expose `shared_disp` (offsets in the index map).

Can one propose a better name `shared indices`?

### DIFF
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -784,7 +784,7 @@ void IndexMap::scatter_rev(tcb::span<T> local_data,
   for (std::size_t i = 0; i < _ghosts.size(); ++i)
   {
     const int np = _ghost_owners[i];
-    for (std::int32_t j = 0; j < n; ++j)
+    for (int j = 0; j < n; ++j)
       send_data[displs[np] + j] = remote_data[i * n + j];
     displs[np] += n;
   }
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -560,7 +560,7 @@ std::map<std::int32_t, std::set<int>> IndexMap::compute_shared_indices() const
   for (std::int32_t p = 0; p < _shared_indices->num_nodes(); ++p)
   {
     const int rank_global = neighbors_out[p];
-    for (const auto& idx : _shared_indices->links(p))
+    for (const std::int32_t& idx : _shared_indices->links(p))
       shared_indices[idx].insert(rank_global);
   }
 
@@ -573,7 +573,7 @@ std::map<std::int32_t, std::set<int>> IndexMap::compute_shared_indices() const
   std::vector<int> fwd_sharing_offsets{0};
   for (std::int32_t p = 0; p < _shared_indices->num_nodes(); ++p)
   {
-    for (const auto& idx : _shared_indices->links(p))
+    for (const std::int32_t& idx : _shared_indices->links(p))
     {
       assert(shared_indices.find(idx) != shared_indices.end());
       if (auto it = shared_indices.find(idx); it->second.size() > 1)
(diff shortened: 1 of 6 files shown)


cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -306,7 +306,7 @@ common::stack_index_maps(
 //-----------------------------------------------------------------------------
 IndexMap::IndexMap(MPI_Comm comm, std::int32_t local_size)
     : _comm_owner_to_ghost(MPI_COMM_NULL), _comm_ghost_to_owner(MPI_COMM_NULL),
-      _comm_symmetric(MPI_COMM_NULL)
+      _comm_symmetric(MPI_COMM_NULL), _shared_indices(0)
```
IgorBaratta, line 309: `AdjacencyList` doesn't have a default constructor.
Any alternative to avoid this weird initialization?

garth-wells: Make it a `std::unique_ptr`.

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -343,6 +343,7 @@ IndexMap::IndexMap(MPI_Comm comm, std::int32_t local_size)
   _comm_owner_to_ghost = dolfinx::MPI::Comm(comm0, false);
   _comm_ghost_to_owner = dolfinx::MPI::Comm(comm1, false);
   _comm_symmetric = dolfinx::MPI::Comm(comm2, false);
+  _shared_indices = std::make_unique<graph::AdjacencyList<std::int32_t>>(0);
```
garth-wells, line 346: I don't follow what's happening here - why is `_shared_indices` being initialised? Are there no 'shared indices'?

cpp/dolfinx/common/IndexMap.h
```diff
@@ -145,9 +147,12 @@ class IndexMap
 
   /// @todo Reconsider name
   /// Local (owned) indices shared with neighbor processes, i.e. are
-  /// ghosts on other processes
+  /// ghosts on other processes, grouped by sharing (neighbor)
+  /// process(destination ranks in forward communicator and source ranks in the
+  /// reverse communicator)
   /// @return List of indices that are ghosted on other processes
-  const std::vector<std::int32_t>& shared_indices() const noexcept;
+  const std::unique_ptr<graph::AdjacencyList<std::int32_t>>&
```
garth-wells, line 154: If `_shared_indices` is always created in the constructor it would be better to return a reference here.

cpp/dolfinx/common/IndexMap.cpp
IgorBaratta: This initialization is in the non-overlapping index-map constructor, so there are no 'shared indices', and no ghosts.
It's being initialized to avoid the need for checking if the unique_ptr is "empty".

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -694,9 +695,10 @@ void IndexMap::scatter_fwd(tcb::span<const T> local_data,
 
   // Copy into sending buffer
   std::vector<T> data_to_send(displs_send.back());
-  for (std::size_t i = 0; i < _shared_indices.size(); ++i)
+  const std::vector<std::int32_t>& indices = _shared_indices->array();
+  for (std::size_t i = 0; i < indices.size(); ++i)
   {
-    const int index = _shared_indices[i];
+    const int index = indices[i];
```
garth-wells, line 701: Let's be explicit here with `std::int32_t` vs `std::int64_t` and in other places where there could be confusion.

IgorBaratta: After checking `IndexMap`, we now have:
Fixed-width integers used for differentiating local and global indices and quantities.  Only a few places had minor inconsistencies. 
And `int`s used for block size, ranks and number of neighbors, and size of data to send/receive.

garth-wells (approved):

PR #1449, opened 2021-03-12
Fix array return type in python

IgorBaratta:


python/dolfinx/wrappers/la.cpp
```diff
@@ -96,8 +96,17 @@ void la(py::module& m)
...
              std::shared_ptr<dolfinx::la::Vector<PetscScalar>>>(m, "Vector")
-      .def_property_readonly("array",
-                             &dolfinx::la::Vector<PetscScalar>::mutable_array);
+      .def_property_readonly(
+          "array",
+          [](dolfinx::la::Vector<PetscScalar>& self) {
+            std::vector<PetscScalar>& array = self.mutable_array();
+            return py::array(array.size(), array.data(), py::cast(self));
+          })
+      .def_property_readonly(
```
garth-wells, line 105: Why are there two near-identical functions?

IgorBaratta: One for `dolfinx::la::Vector<PetscScalar>`  and the other for `std::shared_ptr<dolfinx::la::Vector<PetscScalar>>`.
There is possibly a smarter way to do this.. but I'm not aware of one alternative.

garth-wells: I see that, but why is it needed?

IgorBaratta: It's not actually needed.
It was just my misunderstanding of how smart pointers work in pybind.
https://pybind11.readthedocs.io/en/stable/advanced/smart_ptrs.html#std-shared-ptr

Just removed the first function.

python/dolfinx/wrappers/la.cpp
```diff
@@ -96,8 +96,11 @@ void la(py::module& m)
   // dolfinx::la::Vector
   py::class_<dolfinx::la::Vector<PetscScalar>,
              std::shared_ptr<dolfinx::la::Vector<PetscScalar>>>(m, "Vector")
-      .def_property_readonly("array",
-                             &dolfinx::la::Vector<PetscScalar>::mutable_array);
+      .def_property_readonly(
```
garth-wells, line 99: Keep the reference version - it's simpler and consistent with what we do elsewhere.

garth-wells (approved):

PR #1450, opened 2021-03-14
Fix python lambda assembler

chrisrichardson:
Fixes current broken pybind, which translates `int*` as `int` for python.


garth-wells:
Is this something that we want? I presume it is very slow. If we keep it and it's slow, the docstring should make clear that it's slow and make clear why it's been added.

chrisrichardson:
It's not that slow, in fact (about 3x slower than PETSc) - but anyway, it's certainly not for general use. I'll make that clear.

PR #1460, opened 2021-03-17
Bump GMSH to 4.8.0

jorgensd:
As Eigen is out of basix, we can bump gmsh without getting alignment segfaults.

Fix one interface change from 4.6.0 -> 4.8.0


garth-wells:
Eigen is still in DOLFINX - have you checked for issues?

PR #1467, opened 2021-04-04
Dual graph mixed topology support

chrisrichardson:
Allowws mixed-topology mesh to be partitioned by creating a dual graph.
* Uses topological dimension as a parameter, instead of cell type
* Assumes: 1D meshes are only interval, 2D may be triangle or quad, 3D may have tet, pyramid, prism or hex
* This is a first step towards mixed mesh support, a lot more needs to be done elsewhere.
* Not tested. Should we add a python interface to dual graph creation, for testing?

### DIFF
--- a/cpp/dolfinx/mesh/graphbuild.cpp
+++ b/cpp/dolfinx/mesh/graphbuild.cpp
@@ -22,7 +22,7 @@ namespace
 //-----------------------------------------------------------------------------
 // Compute local part of the dual graph, and return return (local_graph,
 // facet_cell_map, number of local edges in the graph (undirected)
-std::pair<graph::AdjacencyList<std::int32_t>, std::vector<std::int64_t>>
+std::pair<graph::AdjacencyList<std::int32_t>, dolfinx::array2d<std::int64_t>>
 compute_local_dual_graph_keyed(
     const graph::AdjacencyList<std::int64_t>& cell_vertices, int tdim)
 {
@@ -32,12 +32,12 @@ compute_local_dual_graph_keyed(
   if (num_local_cells == 0)
   {
     // Empty mesh on this process
-    std::vector<int> offsets = {0};
-    std::vector<int> values;
-    return {graph::AdjacencyList<std::int32_t>(values, offsets), {}};
+    dolfinx::array2d<std::int64_t> m(0, 0);
+    return {graph::AdjacencyList<std::int32_t>(0), m};
   }
 
   // Count number of cells with n vertices from 0-8
+  // covering interval(2) through to hex(8)
   std::vector<int> count(9, 0);
   for (int i = 0; i < num_local_cells; ++i)
     ++count[cell_vertices.num_links(i)];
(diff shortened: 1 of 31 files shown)


cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -22,124 +22,194 @@ namespace
...
+    return {graph::AdjacencyList<std::int32_t>(0), m};
+  }
+
+  // Count number of cells with n vertices from 0-8
+  // covering interval(2) through to hex(8)
+  std::vector<int> count(9, 0);
+  for (int i = 0; i < num_local_cells; ++i)
+  {
+    assert(cell_vertices.num_links(i) < 9);
+    ++count[cell_vertices.num_links(i)];
```
garth-wells, line 45: This line is rather cryptic.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -22,124 +22,194 @@ namespace
...
+  }
+
+  int num_facets = 0;
+  int num_facet_vertices = 0;
+
+  // Map from number of cell vertices -> facet vertex list
```
garth-wells, line 51: Explanation is a bit cryptic, would be better to expand. Start with high level intent with words.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -22,124 +22,194 @@ namespace
...
+  // This is unique for each dimension
+  // 1D (interval: 2 vertices -> 2 facets)
+  // 2D (triangle: 3v->3f, quad: 4v->4f)
+  // 3D (tet: 4v->4f, pyramid: 5v->5f, prism: 6v->5f, hex: 8v->6f)
+  std::vector<graph::AdjacencyList<int>> nv_to_facets(
+      9, graph::AdjacencyList<int>(0));
+
+  if (tdim == 1)
```
garth-wells, line 59: Use switch rather than if.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -174,9 +242,14 @@ compute_nonlocal_dual_graph(
             0};
   }
 
-  const int tdim = mesh::cell_dim(cell_type);
-  const int num_vertices_per_facet
-      = mesh::num_cell_vertices(mesh::cell_entity_type(cell_type, tdim - 1));
+  // Some processes may have empty map
+
+  std::int32_t num_vertices_per_facet;
+  MPI_Allreduce(&facet_cell_map.shape[1], &num_vertices_per_facet, 1,
```
garth-wells, line 248: The first argument looks unsafe to me - the type is isn't clear. Assign to a variable with the correct type first, then pass address.

cpp/dolfinx/mesh/topologycomputation.cpp
```diff
@@ -587,26 +588,28 @@ compute_from_map(const graph::AdjacencyList<std::int32_t>& c_d0_0,
...
   const auto e_vertices_ref = mesh::get_entity_vertices(cell_type_d0, d1);
-  std::vector<int> keys(e_vertices_ref.size());
+  std::vector<int> keys(e_vertices_ref.array().size());
   for (int e = 0; e < c_d0_0.num_nodes(); ++e)
   {
     auto e0 = c_d0_0.links(e);
-    for (std::size_t i = 0; i < e_vertices_ref.shape[0]; ++i)
-      for (std::size_t j = 0; j < e_vertices_ref.shape[1]; ++j)
-        keys[i * e_vertices_ref.shape[1] + j] = e0[e_vertices_ref(i, j)];
+    for (int i = 0; i < e_vertices_ref.num_nodes(); ++i)
```
garth-wells, line 595: Add braces.

garth-wells (commented):
Any change in performance?

chrisrichardson:
Performance largely depends on `std::sort` and the compare and swap methods it uses. Now should be pretty good.

PR #1473, opened 2021-04-08
Refactor expression interpolation

IgorBaratta:
Co-author: @jorgensd 

- Remove tabulation from loops:
   - `CoordinateElement::push_forward` now receives the tabulated data.
   - Created a function to compute the jacobian and related data `compute_jacobian_data` given the tabulated data at the interpolation points (reference domain). In main, a newton to back engineer the interpolation point using compute reference geometry, which is not necessary because the we know the interpolation points.
 - Introduce xtensor to communicate directly with `basix::FiniteElement`
 - Use xtensor to allocate large datasets (instead of array2d or vector because they need initialization).
 - Remove Eigen::Replicate and avoid some lazy computation, which were expensive.

Further optimizations (in a future PR):
- Special treatment for Lagrange elements  (straightforward).
- Optimize `FiniteElement::map_pull_back`, which corresponds of 50-60% of the interpolation time. Currently it incurs in unnecessary copies and transposition, the idea is no interface directly with `basix::FiniteElement` using xtensor.
- Reconsider the number of interpolation points in basix (currently 5*degree for non-Lagrange elements). 
   - Maybe use the (max degree of the polynomial set)**2+ 1

Some comparisons with main:
For each function space (family +  degree + cel ltype) we increased the number of cells  
- #Tetrahedral elements: 3072, 24576, 196608
- #Hexahedral elements: 512, 4096, 32768 

**Lagrange**
lagrange_interpolate_speedup

**Raviart–Thomas**
rt_interpolate_speedup

**Nédélec**
n1curl_interpolate_speedup

**Tabulate dof coordinates (byproduct)**
Due to the refactoring of push-forward we see also a boost in the performance of tabulate dof coordinates:
lagrange_tabulate_coords_speedup

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -142,6 +142,12 @@ set_package_properties(Eigen3 PROPERTIES TYPE REQUIRED
   DESCRIPTION "Lightweight C++ template library for linear algebra"
   URL "http://eigen.tuxfamily.org")
 
+# Check for required package xtensor
+find_package(xtensor REQUIRED)
+set_package_properties(xtensor PROPERTIES TYPE REQUIRED
+  DESCRIPTION "C++ library for numerical analysis with multi-dimensional array expressions."
+  URL "https://xtensor.readthedocs.io/")
+
 find_package(PETSc 3.10)
 set_package_properties(PETSc PROPERTIES TYPE REQUIRED
   DESCRIPTION "Portable, Extensible Toolkit for Scientific Computation (PETSc)"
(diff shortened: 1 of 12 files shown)


jorgensd:
Updated results after introducing major simplification for elements with identity interpolation matrices. We have currently assumed that this only happens in cases where the dofs are point evaluations.

Lagrange on hexahedral mesh, 32768 cells (speedup compared to main from yesterday)
lagrange_hex_speedup
Lagrange on tetrahedral mesh, 196608 cells (speedup compared to master)
lagrange_tet_speedup

garth-wells (approved):

Issue #1303, opened 2021-01-05
Cahn-Hillard demo documentation out of date


jorgensd:
After #1285 , all references to `NonlinearProblem <dolfinx.cpp.NonlinearProblem>` should be removed, and more documentation on what the different functions (`form`, `F`, `J`, `matrix`, `vector`) does.

PR #1478, opened 2021-04-11
Create Nonlinear problem and Python interface of NewtonSolver.

jorgensd:
Suggested lightweight wrapper in Python for standard use-cases.

Update documentation of Cahn Hilliard demo  (adressing: #1303 ) and use new classes


chrisrichardson (approved):
Looks fine to me

jhale:
I'd prefer to have the Jacobian derivation done outside the class, increases the flexibility. The old pattern where it was passed J as an optional argument was OK too.

python/dolfinx/fem/nonlinearproblem.py
```diff
@@ -0,0 +1,122 @@
...
+        du = ufl.TrialFunction(V)
+        self.L = fem.form.Form(F, form_compiler_parameters=form_compiler_parameters, jit_parameters=jit_parameters)
+        # Create the Jacobian matrix, dF/du
+        self.a = fem.form.Form(ufl.derivative(F, u, du), form_compiler_parameters=form_compiler_parameters,
+                               jit_parameters=jit_parameters)
+        self.bcs = bcs
+
+        # Create matrix and vector to be used for assembly
+        # of the non-linear problem
+        self.matrix = fem.create_matrix(self.a)
```
michalhabera, line 59: Not sure if this fits here well. I would say to either store the jacobian and residual state in the NonlinearProblem internally, and then do not ask in F() nor J() to have vector/matrix provided, but assemble into with internal state mat/vec.

Or maybe even better, do not store this state in the class, but cache it in the NewtonSolver. `NonlinerProblem` -- if you want to make it independent on the actual solver, should just hold residual and jacobian callbacks.

jorgensd: > Not sure if this fits here well. I would say to either store the jacobian and residual state in the NonlinearProblem internally, and then do not ask in F() nor J() to have vector/matrix provided, but assemble into with internal state mat/vec.
> 
> Or maybe even better, do not store this state in the class, but cache it in the NewtonSolver. `NonlinerProblem` -- if you want to make it independent on the actual solver, should just hold residual and jacobian callbacks.

For this we would need to change the underlying C++ code. What do you think about this @garth-wells ?

We should probably also think about the design with respect to the SNESSolver (https://github.com/FEniCS/dolfinx/blob/4cdc251c325adbf884150aa00914da5eec1c2f40/python/test/unit/nls/test_newton.py#L190-L192) which uses the same design as our current NewtonSolver.

michalhabera: If SNESSolver uses the same design as NewtonSolver that is good.

I prefer more how `NonlinearPDE_SNESProblem` looks. I.e. it computes jacobian form and residual form (if not provided explicitly), and user then calls `create_matrix` on this form fetched from the `NonlinearProblem`. It is just less error-prone, to keep less variables as the object state.

You can make NonlinearProblem look similar to  `NonlinearPDE_SNESProblem` and then both, our NewtonSolver and SNESSolver should be easily usable, no?

jorgensd: I've now moved the Matrix and Vector construction into the python initializer of NewtonSolver. If the user wants to create this matrix themselves, they can always use the `cpp.nls.NewtonSolver`.

I figured I'd make a pull request with a SNESProblem after this is merged.

python/dolfinx/fem/nonlinearproblem.py
```diff
@@ -0,0 +1,125 @@
...
+            DOLFINX.
+        jit_parameters
+            Parameters used in CFFI JIT compilation of C code generated by FFCX.
+            See `python/dolfinx/jit.py` for all available parameters.
+            Takes priority over all other parameter values.
+
+        .. code-block:: python
+            problem = LinearProblem(F, u, [bc0, bc1])
+        """
+        self.L = fem.form.Form(F, form_compiler_parameters=form_compiler_parameters, jit_parameters=jit_parameters)
```
michalhabera, line 49: Can you make `self.L, self.a` just read only private properties? The same as in `LinearProblem`.

jorgensd:
@garth-wells @michalhabera  any remaining issues?

python/dolfinx/fem/nonlinearproblem.py
michalhabera: @jorgensd  I meant to make it exactly as in `LinearProblem`, so add property getters for private variables. Private viariables `_foo` should be accessed only via getters. It is more clean and makes it clear what is private/public API.

python/dolfinx/fem/nonlinearproblem.py
```diff
@@ -0,0 +1,145 @@
...
+class NewtonSolver(cpp.nls.NewtonSolver):
+    def __init__(self, comm: mpi4py.MPI.Intracomm, problem: NonlinearProblem):
+        """
+        Create a Newton solver for a given MPI communicator and non-linear problem.
+        """
+        super().__init__(comm)
+
+        # Create matrix and vector to be used for assembly
+        # of the non-linear problem
+        self._A = fem.create_matrix(problem._a)
```
michalhabera, line 122: `problem.a`

python/dolfinx/fem/nonlinearproblem.py
```diff
@@ -0,0 +1,145 @@
...
+        self.setJ(problem.J, self._A)
+        self._b = fem.create_vector(problem._L)
```
michalhabera, line 124: `problem.L`

python/dolfinx/wrappers/nls.cpp
```diff
@@ -27,17 +27,61 @@ void nls(py::module& m)
...
         return std::make_unique<dolfinx::nls::NewtonSolver>(comm.get());
       }))
-      .def("setF", &dolfinx::nls::NewtonSolver::setF)
-      .def("setJ", &dolfinx::nls::NewtonSolver::setJ)
-      .def("setP", &dolfinx::nls::NewtonSolver::setP)
-      .def("set_form", &dolfinx::nls::NewtonSolver::set_form)
-      .def("solve", &dolfinx::nls::NewtonSolver::solve)
-      .def_readwrite("atol", &dolfinx::nls::NewtonSolver::atol)
-      .def_readwrite("rtol", &dolfinx::nls::NewtonSolver::rtol)
+      .def("setF", &dolfinx::nls::NewtonSolver::setF,
```
garth-wells, line 30: I'm not keen on extensive docs in the `cpp` wrappers. For me it makes it hard to read the code because the code becomes hard to find amongst the docstrings. I prefer to see it documented from Python.

python/demo/cahn-hilliard/demo_cahn-hilliard.py
```diff
@@ -323,7 +266,7 @@ def vector(self):
     T = 50 * dt
 
 u.vector.copy(result=u0.vector)
-u0.vector.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+scatter_forward(u.x)
```
garth-wells, line 269: Where did this function come from?

garth-wells (changes requested):
Looks ok to me from a quick scan, but I don't see how to control the inner linear solver. The design of the nonlinear solvers needs a careful look so there may in the future be significant changes to how nonlinear solvers are handled.

python/demo/cahn-hilliard/demo_cahn-hilliard.py
jorgensd: This is the function from IndexMap used to update Ghost (introduced by @chrisrichardson ). It is nicer than using PETSc directly, and faster.

jorgensd:
> Looks ok to me from a quick scan, but I don't see how to control the inner linear solver. The design of the nonlinear solvers needs a careful look so there may in the future be significant changes to how nonlinear solvers are handled.

We could merge: https://github.com/FEniCS/dolfinx/pull/1487
into this PR, as it allows these changes in the NewtonSolver API.
What do you think?

python/demo/cahn-hilliard/demo_cahn-hilliard.py
jorgensd: @garth-wells I've removed most of the documentation from the Pybind layer. In general we need to think about how we can customize the underlying linear solver. However, I feel that is out of the scope of this PR.

jhale:
I am happy with the PR as it is. 

We should look at how to control the settings of inner solver but it can be done in a separate PR.

jorgensd:
@jhale has added in exposure of the underlying PETSc solver in the NewtonSolver, superseeding #1487 .
I think this PR is ready to be merged in now.

cpp/dolfinx/la/PETScKrylovSolver.h
```diff
@@ -35,16 +35,16 @@ class PETScKrylovSolver
   /// @param[in] inc_ref_count Increment the reference count on `ksp` if true
   PETScKrylovSolver(KSP ksp, bool inc_ref_count);
 
-  // Delete copy constructor
```
garth-wells, line 38: The deleted lines are better - we don't need to document something that is deleted.

cpp/dolfinx/nls/NewtonSolver.h
```diff
@@ -35,8 +35,20 @@ class NewtonSolver
   /// @param[in] comm The MPI communicator for the solver
   explicit NewtonSolver(MPI_Comm comm);
 
+  /// Move constructor (deleted)
```
garth-wells, line 38: Don't document (`///`) deleted constructors.

cpp/dolfinx/nls/NewtonSolver.h
```diff
@@ -35,8 +35,20 @@ class NewtonSolver
...
+  NewtonSolver(NewtonSolver&& solver) = delete;
+
+  /// Copy constructor (deleted)
```
garth-wells, line 41: See comment above.

cpp/dolfinx/nls/NewtonSolver.h
```diff
@@ -35,8 +35,20 @@ class NewtonSolver
...
+  NewtonSolver(const NewtonSolver& solver) = delete;
+
+  /// Assignment operator (deleted)
```
garth-wells, line 44: See comment above.

cpp/dolfinx/nls/NewtonSolver.h
```diff
@@ -35,8 +35,20 @@ class NewtonSolver
...
+  NewtonSolver& operator=(const NewtonSolver& solver) = delete;
+
+  /// Move assignment constructor (deleted)
```
garth-wells, line 47: See comment above.

python/dolfinx/fem/nonlinearproblem.py
```diff
@@ -0,0 +1,109 @@
+# -*- coding: utf-8 -*-
```
garth-wells, line 1: Could this be combined with `LinearProblem` in `problem.py`? Is nice to keep the number of files and not have a separate file for every class.

python/dolfinx/nls.py
```diff
@@ -0,0 +1,58 @@
+# Copyright (C) 2021 Jørgen S. Dokken
+#
+# This file is part of DOLFINx (https://www.fenicsproject.org)
+#
+# SPDX-License-Identifier:    LGPL-3.0-or-later
+"""Methods for solving nonlinear equations."""
+
+import types
+from petsc4py import PETSc
```
garth-wells, line 9: Do we need to import if we only want type hints?

python/dolfinx/wrappers/nls.cpp
```diff
@@ -27,17 +27,30 @@ void nls(py::module& m)
       .def(py::init([](const MPICommWrapper comm) {
         return std::make_unique<dolfinx::nls::NewtonSolver>(comm.get());
       }))
+      .def("get_krylov_solver",
```
garth-wells, line 30: Maybe this should be a propery?

python/dolfinx/nls.py
jorgensd: It will be possible from python 3.10: https://docs.python.org/pt-br/3.10/whatsnew/3.10.html#pep-613-typealias

cpp/dolfinx/nls/NewtonSolver.h
jhale: I think this is a limitation of our Doxygen settings (all methods must be documented.)

jorgensd:
@garth-wells  @jhale are we ready to merge this PR?

PR #1487, opened 2021-04-15
Customise newton solver

massimiliano-leoni:
The `NewtonSolver` class defaults to using LU for its linear system solves. This PR adds two optional parameters to its constructor to specify custom solvers and preconditioners.


jhale:
I agree it's limiting being hard coded to LU, but I'm not sure about the implementation. Just being able to set the Krylov algorithm and the preconditioner isn't really going to be enough for a user wanting customisation. BoomerAMG typically needs a few extra options to get good performance, for example.

It might be best to have a method to return the underlying `dolfinx::PETScKrylovSolver`, then the user can do as they please.

massimiliano-leoni:
I agree with you, Jack. I wanted to make sure there was interest in the feature before I wrote too much code.
Maybe it would be better to provide an access method to the solver. What about that?
This way, the user can access the underlying PETSc object and configure the solver as needed.

jorgensd:
> I agree it's limiting being hard coded to LU, but I'm not sure about the implementation. Just being able to set the Krylov algorithm and the preconditioner isn't really going to be enough for a user wanting customisation. BoomerAMG typically needs a few extra options to get good performance, for example.
> 
> It might be best to have a method to return the underlying `dolfinx::PETScKrylovSolver`, then the user can do as they please.

How exactly should this be done in the Python interface? As the petsc4py options and PETScOptions in C++ are separate objects, I do not see any nice way of exposing it.

massimiliano-leoni:
> How exactly should this be done in the Python interface? As the petsc4py options and PETScOptions in C++ are separate objects, I do not see any nice way of exposing it.

I don't have an answer to that question, I have very little experience with the Python bindings.

jorgensd:
> > How exactly should this be done in the Python interface? As the petsc4py options and PETScOptions in C++ are separate objects, I do not see any nice way of exposing it.
> 
> I don't have an answer to that question, I have very little experience with the Python bindings.

@massimiliano-leoni, this question was more directed towards @jhale's comment.

massimiliano-leoni:
@jorgensd I suspected that, but I answered just to yo make sure :_-D

PR #1488, opened 2021-04-15
More simplifications in Coordinate Element

IgorBaratta:


cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -95,161 +104,96 @@ void CoordinateElement::push_forward(array2d<double>& x,
...
-  Eigen::Map<const Eigen::Array<double, Eigen::Dynamic, Eigen::Dynamic,
-                                Eigen::RowMajor>>
-      _cell_geometry(cell_geometry.data(), cell_geometry.shape[0],
-                     cell_geometry.shape[1]);
-
   // in-argument checks
   const int tdim = this->topological_dimension();
-  const int gdim = _x.cols();
-  assert(_cell_geometry.cols() == gdim);
+  const int gdim = x.shape(1);
```
garth-wells, line 119: If you make this `std::size_t`,  the `(int)` casts below can be removed.

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -95,161 +104,96 @@ void CoordinateElement::push_forward(array2d<double>& x,
...
-    // Compute Jacobian and inverse
-    dphi = tabulated_data.block(1, 0, tdim, d).transpose();
-    Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor, 3, 3>
-        J0(gdim, tdim);
-    J0 = _cell_geometry.matrix().transpose() * dphi;
-    Eigen::Matrix<double, Eigen::Dynamic, Eigen::Dynamic, Eigen::RowMajor, 3, 3>
-        K0(tdim, gdim);
+    // Tabulate shape function and first derivative at the origin
+    xt::xtensor<double, 2> X0 = xt::zeros<double>({1, tdim});
+    xt::xtensor<double, 4> tabulated_data = _element->tabulate(1, X0);
```
garth-wells, line 133: Can we find a better variable name?

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -95,161 +104,96 @@ void CoordinateElement::push_forward(array2d<double>& x,
...
-      else
-        detJ[ip] = std::sqrt((Jview.transpose() * Jview).determinant());
     }
   }
 }
 //-----------------------------------------------------------------------------
-void CoordinateElement::permute_dofs(std::int32_t* dofs,
-                                     std::uint32_t cell_perm) const
+void CoordinateElement::permute_dofs(tcb::span<std::int32_t> dofs,
+                                     const uint32_t cell_perm) const
```
garth-wells, line 186: Use C++ `std::uint32_t`. `const` is not normally used when passing by value.

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -95,161 +104,96 @@ void CoordinateElement::push_forward(array2d<double>& x,
...
 {
-  basix::permute_dofs(_basix_element_handle, dofs, cell_perm);
+  assert(_element);
+  _element->permute_dofs(dofs, cell_perm);
 }
 //-----------------------------------------------------------------------------
-void CoordinateElement::unpermute_dofs(std::int32_t* dofs,
-                                       std::uint32_t cell_perm) const
+void CoordinateElement::unpermute_dofs(tcb::span<std::int32_t> dofs,
+                                       const uint32_t cell_perm) const
```
garth-wells, line 193: See comment above.

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -264,101 +208,105 @@ CoordinateElement::tabulate(int n, const array2d<double>& X) const
...
 }
-//-----------------------------------------------------------------------------
-
-void CoordinateElement::compute_jacobian_data(
-    const xt::xtensor<double, 4>& tabulated_data, const array2d<double>& X,
-    const array2d<double>& cell_geometry, std::vector<double>& J,
-    xtl::span<double> detJ, std::vector<double>& K) const
+//--------------------------------------------------------------------------------
+void CoordinateElement::compute_jacobian(
+    const xt::xtensor<double, 4>& tabulated_data,
```
garth-wells, line 213: Use a more meaningful argument name.

cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -49,18 +49,47 @@ class CoordinateElement
...
-  /// Tabulate shape functions up to n-th order derivative at points X in the
-  /// reference geometry
-  /// Note: Dynamic allocation.
+  /// Compute basis values and derivatives at set of points.
+  /// @param[in] n The order of derivatives, up to and including, to
+  /// compute. Use 0 for the basis functions only.
+  /// @param[in] X The points at which to compute the basis functions.
+  /// The shape of x is (number of points, geometric dimension).
+  /// @return The basis functions (and derivatives). The shape is
+  /// (derivative, number point, number of basis fn, value size).
```
garth-wells, line 58: Use `xtensor<double, x>` in place of `array2d<double>`?

cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -49,18 +49,47 @@ class CoordinateElement
...
+  /// Compute Jacobian for a cell with given geometry using the
+  /// basis functions and first order derivatives.
+  /// @param[in] tabulated_data Pre-computed basis values and derivatives at set
+  /// of points. Need at least first order derivative.
+  /// @param[in] cell_geometry Coordinates/geometry
+  /// @param[in,out] J The Jacobian
+  /// The shape of J is (number of points, geometric dimension, topological
+  /// dimenson).
+  void compute_jacobian(const xt::xtensor<double, 4>& tabulated_data,
+                        const xt::xtensor<double, 2>& cell_geometry,
```
garth-wells, line 70: Maybe allow data for multiple cells to be passed? Common operation would be evaluating data at same reference points on multiple cells. Would there be performance benefits?

cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -81,17 +110,19 @@ class CoordinateElement
 
   /// Compute reference coordinates X, and J, detJ and K for physical
   /// coordinates x
-  void compute_reference_geometry(array2d<double>& X, std::vector<double>& J,
-                                  xtl::span<double> detJ,
-                                  std::vector<double>& K,
-                                  const array2d<double>& x,
-                                  const array2d<double>& cell_geometry) const;
+  void compute_reference_geometry(
```
garth-wells, line 113: I think a better name would be `compute_pull_back`.

cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -81,17 +110,19 @@ class CoordinateElement
...
+      xt::xtensor<double, 2>& X, xt::xtensor<double, 3>& J,
+      xt::xtensor<double, 1>& detJ, xt::xtensor<double, 3>& K,
+      const xt::xtensor<double, 2>& x,
+      const xt::xtensor<double, 2>& cell_geometry) const;
 
   /// Permutes a list of DOF numbers on a cell
-  void permute_dofs(std::int32_t* dofs, std::uint32_t cell_perm) const;
+  void permute_dofs(tcb::span<std::int32_t> dofs,
+                    const uint32_t cell_perm) const;
```
garth-wells, line 121: Use `std::`

cpp/dolfinx/fem/Function.h
```diff
@@ -335,7 +336,19 @@ class Function
         xp(0, j) = x(p, j);
 
       // Compute reference coordinates X, and J, detJ and K
-      cmap.compute_reference_geometry(X, J, detJ, K, xp, coordinate_dofs);
+      xt::xtensor<double, 2> Xview = xt::adapt(X.data(), X.shape);
```
garth-wells, line 339: Declare the xtensor arrays outside of the loop to avoid memory allocations.

Also, these aren't views.

Alternative is to use `auto` and template `compute_reference_geometry`, etc.

garth-wells (changes requested):

cpp/dolfinx/fem/CoordinateElement.h
IgorBaratta: I think it was some merging conflict (it was uint32_t before).

cpp/dolfinx/fem/CoordinateElement.cpp
IgorBaratta: Changed to dphi. We only need the first derivative of the shape functions.

cpp/dolfinx/fem/CoordinateElement.h
IgorBaratta: I'm not sure if there would be performance benefits.
The shape function and derivatives are computed only once per mesh (or coordinate element), and the main computation is just a dot product.

garth-wells (approved):

PR #1492, opened 2021-04-16
Add docstrings to dofmap python interface

jorgensd:


python/dolfinx/fem/dofmap.py
```diff
@@ -19,24 +19,46 @@ def __init__(self, dofmap: cpp.fem.DofMap):
...
+        Parameters
+        ----------
+        cell
+          The cell index
+        """
         return self._cpp_object.cell_dofs(cell_index)
 
     @property
     def bs(self):
+        """
```
garth-wells, line 33: Can we put the one-line doc strings on one line? Three lines is a waste of space and reduced code readability.

PR #1500, opened 2021-04-18
Gmsh bump

jorgensd:
As we have removed Eigen, it should be possible to bump gmsh to 4.8.3


godsic:
I've built docker container with gmsh 4.8.3 and it works just fine for my limited user case.

jorgensd:
> I've built docker container with gmsh 4.8.3 and it works just fine for my limited user case.

As you observe here, the CI is not happy with the bump (Also fine locally for me). I need to sit down and build the CI containers locally

jorgensd:
Seems like the CI has some incompatibility with the GMSH binaries, as I cannot reproduce the crashes locally with `fenicsproject/test-env:openmpi/mpich`.

jorgensd:
> Is the recipe for how the gmsh binaries are built published? It could well be they are using different flags, emitting instructions that the CI does not have. Could also be an Eigen alignment issue.

I have not been able to find them. I guess we could add what they use in: https://gitlab.onelab.info/gmsh/gmsh/-/blob/master/utils/docker/Dockerfile.ubuntu20.04 and build gmsh from source

jorgensd:
> Is the recipe for how the gmsh binaries are built published? It could well be they are using different flags, emitting instructions that the CI does not have. Could also be an Eigen alignment issue.

@jhale Building from source now passes the CI. A change to the Dockerfile would be along the lines of
```
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -41,7 +41,7 @@
 #    docker build --target dev-env --file dolfinx/docker/Dockerfile --build-arg PETSC_SLEPC_OPTFLAGS="-O2 -march=native" .
 #
 
-ARG GMSH_VERSION=4.8.3
+ARG GMSH_VERSION=4_8_4
 ARG PYBIND11_VERSION=2.6.2
 ARG PETSC_VERSION=3.15.0
 ARG SLEPC_VERSION=3.15.0
@@ -192,18 +192,25 @@ RUN wget -nc --quiet https://github.com/ornladios/ADIOS2/archive/v${ADIOS2_VERSI
     cmake --install build-dir && \
     rm -rf /tmp/*
 
-# Download Install gmsh SDK
+# Install GMSH
 # Only compatible with x86-64 (amd64)
 RUN dpkgArch="$(dpkg --print-architecture)"; \
     case "$dpkgArch" in amd64) \
     cd /usr/local && \
-    wget -nc --quiet http://gmsh.info/bin/Linux/gmsh-${GMSH_VERSION}-Linux64-sdk.tgz && \
-    tar -xf gmsh-${GMSH_VERSION}-Linux64-sdk.tgz && \
-    rm gmsh-${GMSH_VERSION}-Linux64-sdk.tgz ;; \
+    export DEBIAN_FRONTEND=noninteractive && \
+    apt-get -qq update && \
+    apt-get -yq --with-new-pkgs -o Dpkg::Options::="--force-confold" upgrade && \
+    apt-get -y install libfltk1.3-dev libfreetype6-dev libgl1-mesa-dev libocct-foundation-dev libocct-data-exchange-dev && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* && \
+    git clone -b gmsh_${GMSH_VERSION} --single-branch https://gitlab.onelab.info/gmsh/gmsh.git && \
+    cd gmsh && \
+    mkdir build && cd build && \
+    cmake -DCMAKE_BUILD_TYPE=Release -DENABLE_BUILD_DYNAMIC=1  -DENABLE_OPENMP=1  .. && \
+    make install ;; \
     esac;
 
-ENV PATH=/usr/local/gmsh-${GMSH_VERSION}-Linux64-sdk/bin:$PATH \
-    PYTHONPATH=/usr/local/gmsh-${GMSH_VERSION}-Linux64-sdk/lib:$PYTHONPATH
+ENV PYTHONPATH=/usr/local/lib:$PYTHONPATH
```
Should we  make a new PR changing this in the Dockerfile?

jhale:
It's fine for me, particularly as it would presumably fix the missing gmsh
in the ARM builds?

jorgensd:
Fixed in #1559

PR #1503, opened 2021-04-19
Use xtensor in graphbuilder

chrisrichardson:


garth-wells:
Is there a reason to use `xarray` rather than `xtensor`?

chrisrichardson:
I'm still learning xt, but I assumed xarray was more like Eigen::Array, and xtensor more like Eigen::Matrix, i.e. xtensor should be used more for linear algebra objects, rather than just arrayed data. I realise the difference is probably largely semantic.

garth-wells:
> I'm still learning xt, but I assumed xarray was more like Eigen::Array, and xtensor more like Eigen::Matrix, i.e. xtensor should be used more for linear algebra objects, rather than just arrayed data. I realise the difference is probably largely semantic.

`xtensor` is for fixed rank. `xarray` is for a priori unknown rank or dynamic rank. If the rank is fixed then `xtensor`  should be used.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -266,10 +268,12 @@ compute_nonlocal_dual_graph(
   std::int64_t local_min = std::numeric_limits<std::int64_t>::max();
   std::int64_t local_max = 0;
 
-  for (std::size_t i = 0; i < facet_cell_map.shape[0]; ++i)
+  if (facet_cell_map.shape()[0] > 0)
   {
-    local_min = std::min(local_min, facet_cell_map(i, 0));
-    local_max = std::max(local_max, facet_cell_map(i, 0));
+    xt::xtensor<std::array<std::int64_t, 2>, 0> p
```
garth-wells, line 273: This looks odd with `std::array` mixed in. Have you checked that it does what is expected?

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -285,7 +289,7 @@ compute_nonlocal_dual_graph(
 
   // Count number of item to send to each rank
   std::vector<int> p_count(num_processes, 0);
-  for (std::size_t i = 0; i < facet_cell_map.shape[0]; ++i)
+  for (std::size_t i = 0; i < facet_cell_map.shape()[0]; ++i)
```
garth-wells, line 292: `facet_cell_map.shape(0)`

cpp/dolfinx/mesh/graphbuild.cpp
IgorBaratta: `xt::xtensor_fixed<std::int64_t, xt::xshape<2>> p` ?

chrisrichardson: @IgorBaratta - Does that need another header? I can't get it to compile.
I agree the declaration looks odd, but it does return a scalar (0) dimension xtensor containing a std::array correctly.
I would be very happy with something else, but it is hard to determine the correct return type.

IgorBaratta: I think it needs `xtensor/xfixed.hpp`, but std::array is simpler

PR #1506, opened 2021-04-19
Remove Eigen from Dockerfile

garth-wells:


jhale:
I'll merge this into my branch jhale/docker-arm as I've made some extensive changes.

Issue #1509, opened 2021-04-21
Interpolation into DG 0 space fails


jhale:
```
import numpy as np

import mpi4py.MPI as MPI

import dolfinx
import ufl

def f_e(x):
    values = np.ones(x.shape[1])
    return values

mesh = dolfinx.UnitSquareMesh(MPI.COMM_WORLD, 3, 3)
V_DG = dolfinx.FunctionSpace(
    mesh, ufl.FiniteElement("DG", mesh.ufl_cell(), 0))
f = dolfinx.Function(V_DG)
f.interpolate(f_e)
```

Fails with:
```
python3: /src/dolfinx/cpp/dolfinx/fem/CoordinateElement.cpp:203: void dolfinx::fem::CoordinateElement::push_forward(xt::xtensor<double, 2>&, xt::xtensor<double, 2>&, xt::xtensor<double, 2>&) const: Assertion `phi.shape(2) == cell_geometry.shape(0)' failed.

Loguru caught a signal: SIGABRT
Stack trace:
37            0x5f927e _start + 46
36      0x7f4ca54bd0b3 __libc_start_main + 243
35            0x6b64bd Py_BytesMain + 45
34            0x6b6132 Py_RunMain + 530
33            0x67da6e PyRun_SimpleFileExFlags + 382
32            0x67d6db PyRun_FileExFlags + 155
31            0x67d61f python3() [0x67d61f]
30            0x67d5a1 python3() [0x67d5a1]
29            0x68c1e7 PyEval_EvalCode + 39
28            0x56822a _PyEval_EvalCodeWithName + 618
27            0x56a136 _PyEval_EvalFrameDefault + 2294
26            0x5f6033 _PyFunction_Vectorcall + 915
25            0x56822a _PyEval_EvalCodeWithName + 618
24            0x569f5e _PyEval_EvalFrameDefault + 1822
23            0x5f6033 _PyFunction_Vectorcall + 915
22            0x56822a _PyEval_EvalCodeWithName + 618
21            0x56b7b0 _PyEval_EvalFrameDefault + 8048
20            0x5f2b87 PyObject_Call + 503
19            0x5f6033 _PyFunction_Vectorcall + 915
18            0x56822a _PyEval_EvalCodeWithName + 618
17            0x56ef97 _PyEval_EvalFrameDefault + 22359
16            0x50aa8b python3() [0x50aa8b]
15            0x5f3446 _PyObject_MakeTpCall + 662
[...]

PR #1511, opened 2021-04-21
Bug fixes in CoordinateElement

jhale:
- `const double` is intepreted by pybind11 as longdouble.
- The assert on shape(2) (3rd dimension) makes no sense on an array of dimension 2.


jhale:
Fixes #1509.

jorgensd (approved):

IgorBaratta (approved):

IgorBaratta:
The shape(2) is reminiscent of when the input data was 4d. But now we only pass 2d:
```c++
const xt::xtensor<double, 2> phi = xt::view(cmap.tabulate(0, X), 0, xt::all(), xt::all(), 0);
```

PR #1513, opened 2021-04-22
Performance updates in graph construction

garth-wells:


cpp/dolfinx/graph/partition.cpp
```diff
@@ -42,9 +42,10 @@ build::distribute(MPI_Comm comm, const graph::AdjacencyList<std::int64_t>& list,
   std::vector<int> num_per_dest_send(size, 0);
   for (int i = 0; i < destinations.num_nodes(); ++i)
   {
-    const auto& dests = destinations.links(i);
-    for (int j = 0; j < destinations.num_links(i); ++j)
-      num_per_dest_send[dests[j]] += list.num_links(i) + 3;
+    auto list_num_links = list.num_links(i) + 3;
```
chrisrichardson, line 45: Better to use int type than auto

cpp/dolfinx/graph/partition.cpp
```diff
@@ -200,8 +202,8 @@ build::compute_ghost_indices(MPI_Comm comm,
                                  MPI_INFO_NULL, false, &neighbor_comm);
 
   std::vector<int> send_offsets = {0};
-  for (std::size_t i = 0; i < ghost_index_count.size(); ++i)
-    send_offsets.push_back(send_offsets.back() + ghost_index_count[i]);
+  for (auto index_count : ghost_index_count)
```
chrisrichardson, line 205: int instead of auto?

PR #1518, opened 2021-04-23
Remove upper bound on ufl

jorgensd:
UFL version bump


chrisrichardson:
I think I fixed this in another branch

PR #1524, opened 2021-04-29
Create and store parent map when refining a mesh

mscroggs:
When a mesh is refined, this makes a "parent map" that contains a pair (dim, number) for each vertex in the refined mesh. This tells us which entity the new vertex is at the midpoint of.

For example, (0, 5) tells us that the vertex in the refined mesh is at vertex 5 of the parent mesh; (1, 3) tells us that the vertex in the refined mesh is the midpoint of edge 3 of the parent mesh.

This currently only works in serial. Someone who knows more about how the partitioning works might have to take a look at this in parallel. (Currently in serial, no map is created, so the refined mesh stores and empty map.)


garth-wells:
We will need to get this working in parallel before it can be merged (a DOLFINx design principle was consistent behaviour in serial and parallel).

PR #1540, opened 2021-05-11
Use xtensor in pack_coefficients

IgorBaratta:
`std::vector` can potentially initialize data with zeros (expensive), using `xt::empty` guarantees that data is allocated but not initialized.

### DIFF
--- a/cpp/dolfinx/fem/assemble_matrix_impl.h
+++ b/cpp/dolfinx/fem/assemble_matrix_impl.h
@@ -1,6 +1,6 @@
 // Copyright (C) 2018-2019 Garth N. Wells
 //
-// This file is part of DOLFINX (https://www.fenicsproject.org)
+// This file is part of DOLFINx (https://www.fenicsproject.org)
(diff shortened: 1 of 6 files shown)


garth-wells:
I'm not sure we want to do this. I had in mind using array2d to align rows for SIMD loading.

IgorBaratta:
Some data:
```
mpirun -np 56 ./dolfinx-scaling-test --order 1 --problem_type poisson --scaling_type weak --ndofs 500'000
```

```
ZZZ Assemble vector                                            |    1    0.648508 
ZZZ Pack Coefficients                                          |    1    0.246505

ZZ Assemble vector (xt)                                        |    1    0.556826
ZZZ Pack Coefficients (xt)                                     |    1    0.165185 
```

Pack coefficients is domintinating the assembly time.

IgorBaratta:
Alternative:
https://github.com/FEniCS/dolfinx/compare/igor/no_init_adaptor

PR #1552, opened 2021-05-21
Template over block size for common cases for vector (RHS) assembly

garth-wells:
Uses `constexpr` to fix block size at compile time for common cases.


michalhabera:
Why not classical
```c++
template <typename T>
assemble(args, bs);

template <typename T, int bs>
assemble(args){
    assemble<T>(args, bs);
}
``` 
?

I do not like you are repeating that code bit. Also, the entire function has compile time known block size, but you only take advatange of it the in repeated code block. In this case it could be the only use of `bs`, but if it is needed elsewhere, do you duplicate that part again?

garth-wells:
> Why not classical
> 
> ```c++
> template <typename T>
> assemble(args, bs);
> 
> template <typename T, int bs>
> assemble(args){
>     assemble<T>(args, bs);
> }
> ```
> 
> ? (but not sure if the above would get inlined)
> 

Because that requires two functions.

> I do not like you are repeating that code bit. Also, the entire function has compile time known block size, but you only take advantage of it the in repeated code block. In this case it could be the only use of `bs`, but if it is needed elsewhere, do you duplicate that part again?

Where else is it needed that would make a performance difference?

michalhabera:
> > Why not classical
> > ```c++
> > template <typename T>
> > assemble(args, bs);
> > 
> > template <typename T, int bs>
> > assemble(args){
> >     assemble<T>(args, bs);
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ? (but not sure if the above would get inlined)
> 
> Because that requires two functions.
> 
> > I do not like you are repeating that code bit. Also, the entire function has compile time known block size, but you only take advantage of it the in repeated code block. In this case it could be the only use of `bs`, but if it is needed elsewhere, do you duplicate that part again?
> 
> Where else is it needed that would make a performance difference?

Here is a runtime `bs1`, https://github.com/FEniCS/dolfinx/pull/1552/files#diff-f5ccf3dfd046113bbec39c9a335aeb682c7510fde079966e2dd3f56fcbd67eabR216
(and I do not mean a solution is to make that bit duplicate).

It would be more natural (in my opinion), if this dispatch happens at the level of entire function - or separate the dispatched call into a function.

IgorBaratta (approved):

jhale:
I see @michalhabera point but I think using constexpr approach is fine on balance.

PR #1555, opened 2021-05-22
Fix geometry dimension to 3 in assembler

garth-wells:
The change makes the geometric dimension a compile time constant (3) for fetching geometry data to pass into kernels. This gives a helpful performance benefit.

Requires https://github.com/FEniCS/ffcx/pull/358 first.


chrisrichardson (approved):
Satisfyingly simple

PR #1557, opened 2021-05-25
Expose underlying basix element in coordinate element.

jorgensd:
Reimplement `evaluate_reference_basis_derivatives`.
Simplify `evaluate_reference_basis`.

As `evaluate_reference_basis_derivatives` now just directly calls basix, there shouldn't be any need for additional tests of the functionality in DOLFINx.

Exposing these functions will make it easier to create custom assemblers, as one does not have to backwards engineer the basix element from the Mesh/FunctionSpace

### DIFF
--- a/cpp/dolfinx/fem/CoordinateElement.h
+++ b/cpp/dolfinx/fem/CoordinateElement.h
@@ -123,6 +123,9 @@ public:
   /// passing in (for higher order geometries)
   bool needs_permutation_data() const;
 
+  /// Return shared pointer to Basix
+  std::shared_ptr<basix::FiniteElement> element() const;
+
   /// Absolute increment stopping criterium for non-affine Newton solver
   double non_affine_atol = 1.0e-8;
 
--- a/cpp/dolfinx/fem/FiniteElement.cpp
+++ b/cpp/dolfinx/fem/FiniteElement.cpp
@@ -121,7 +121,6 @@ FiniteElement::FiniteElement(const ufc_finite_element& ufc_element)
     ufc_finite_element* ufc_sub_element = ufc_element.sub_elements[i];
     _sub_elements.push_back(std::make_shared<FiniteElement>(*ufc_sub_element));
   }
-
 }
 //-----------------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -123,6 +123,9 @@ class CoordinateElement
   /// passing in (for higher order geometries)
   bool needs_permutation_data() const;
 
+  /// Return shared pointer to Basix
```
IgorBaratta, line 126:
```suggestion
  /// Return shared pointer to Basix element
```

cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -164,25 +163,18 @@ void FiniteElement::evaluate_reference_basis(
...
-// void FiniteElement::evaluate_reference_basis_derivatives(
-//     std::vector<double>& /*values*/, int /*order*/,
-//     const xt::xtensor<double, 2>& /*X*/) const
-// {
-//   // NOTE: This function is untested. Add tests and re-active
-//   throw std::runtime_error(
-//       "FiniteElement::evaluate_reference_basis_derivatives required
-//       updating");
-// }
+void FiniteElement::evaluate_reference_basis_derivatives(
```
IgorBaratta, line 169: The function allocates and copies the data to the input.
Wouldn't it be simpler to return it?

garth-wells (commented):
It would be better to not expose Basix. The design idea is that a user could eventually provide their own element.

Is there a reason why we can't now use a `FiniteElement` in place of `CoordinateElement`?

jorgensd:
> It would be better to not expose Basix. The design idea is that a user could eventually provide their own element.
> 
> Is there a reason why we can't now use a `FiniteElement` in place of `CoordinateElement`?

There are some specific differences between a FiniteElement and a CoordinateElement, as the coordinate element has functionality to compute the Jacobian (and should be able to return the reference facet Jacobians), that I am not sure we should expose through the FiniteElement interface. We could make CoordinateElement store a shared pointer to the finite element (and only have the coordinate-element specific functions in CoordinateElement).

PR #1559, opened 2021-06-01
Gmsh bump + src install

### LINKED ISSUE #1328: Bump gmsh version
Currently, we are not using the latest release of gmsh due to the Eigen aligment issues, #1277 .
Once we have sorted this out, we should get on to the latest version of gmsh. 
With this, we should remove the download of gmsh binaries, and rather use `pip3`, as only the following is required to make gmsh installable with pip3 (tested using ubuntu:20.04 as base):
```
apt-get update
apt-get install libglu1 libxcursor-dev libxinerama1libxft2 python3-pip
pip3 install gmsh --user
export PATH=$PATH:/root/.local/bin
python3 -c "import gmsh; print(gmsh.__version__)"
```

jorgensd:
Resolves #1328.
GMSH built from source with OpenCascade and OpenMP support.
Instruction set tested locally and in #1500

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -41,7 +41,8 @@
 #    docker build --target dev-env --file dolfinx/docker/Dockerfile --build-arg PETSC_SLEPC_OPTFLAGS="-O2 -march=native" .
 #
 
-ARG GMSH_VERSION=4.6.0
+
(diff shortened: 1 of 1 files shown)


jhale (approved):
Let's get this merged and see if there are any failures in a build overnight.

PR #1564, opened 2021-06-09
Add coefficient/constant input to Python assemblers

jorgensd:
Add pybindings for sending in coeffs and constants to assembly, to avoid re-computation in certain cases.


garth-wells:
The benefit here is limited because the coefficient array still gets copied at each call. Would be better if we could make changes to avoid the copy.

garth-wells:
Closing because the data is copied.

PR #1565, opened 2021-06-10
Simplify tabulation through dolfinx::FiniteElement

jorgensd:
As DOLFINx uses `xtensor` there is no point of distinguishing between tabulation of 0th order and nth order derivatives of the basis function. 

Also makes it possible to tabulate the derivative of basis function in DOLFINx (useful for custom kernels).


cpp/dolfinx/fem/Function.h
```diff
@@ -341,9 +343,11 @@ class Function
...
       // Get cell geometry (coordinate dofs)
       auto x_dofs = x_dofmap.links(cell_index);
-      for (std::size_t i = 0; i < num_dofs_g; ++i)
-        for (std::size_t j = 0; j < gdim; ++j)
-          coordinate_dofs(i, j) = x_g(x_dofs[i], j);
+      for (std::size_t i = 0; i < x_dofs.size(); ++i)
+      {
+        std::copy_n(xt::row(x_g, x_dofs[i]).begin(), 3,
+                    std::next(coordinate_dofs.begin(), 3 * i));
+      }
```
jorgensd, line 350:
```suggestion
      for (std::size_t i = 0; i < num_dofs_g; ++i)
        for (std::size_t j = 0; j < gdim; ++j)
          coordinate_dofs(i, j) = x_g(x_dofs[i], j);
```

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -81,10 +81,17 @@ class FiniteElement
...
+  /// Evaluate all derivatives of the basis functions up to given order at given
+  /// points in reference cell
+  /// @param[in,out] values Four dimensional xtensor that will be filled with
+  /// the tabulated values. Should be of shape {num_derivatives, num_points,
+  /// num_dofs, reference_value_size}
+  /// @param[in] X Two dimensional xtensor of shape [num_points, geometric
+  /// dimension] containing the points at the reference element
+  /// @param[in] order The number of derivatives (up to and including this
+  /// order) to tabulate for.
+  void tabulate(xt::xtensor<double, 4>& values, const xt::xtensor<double, 2>& X,
```
IgorBaratta, line 93: Should we also change the order of parameters in `CoordinateElement::tabulate` for consistency?

cpp/dolfinx/fem/Function.h
```diff
@@ -300,8 +300,10 @@ class Function
     xt::xtensor<double, 1> detJ = xt::zeros<double>({1});
 
     // Prepare basis function data structures
-    xt::xtensor<double, 3> basis_reference_values(
-        {1, space_dimension, reference_value_size});
+    xt::xtensor<double, 4> basis_derivatives_reference_values(
+        {1, 1, space_dimension, reference_value_size});
+    auto basis_reference_values = xt::view(basis_derivatives_reference_values,
+                                           0, xt::all(), xt::all(), xt::all());
```
garth-wells, line 306: A little issue is that this will be copied further down in each loop iteration when passed to `element->transform_reference_basis` since `transform_reference_basis` expects a `xt::xarray`.

jorgensd: That is not correct?
Transform reference basis takes an xtensor: https://github.com/FEniCS/dolfinx/blob/13b06ce99e79a6972b55cdc4710df05b4a93d79b/cpp/dolfinx/fem/FiniteElement.cpp#L200-L208
and 
`map_push_forward_m` takes an xtensor: https://github.com/FEniCS/basix/blob/266c9b486d6a8e6d00c118cb05f499690cd98679/cpp/basix/finite-element.h#L317

garth-wells: Yes, it takes a `xt::xtensor<double, 3>`, but `xt::view` isn't a `xt::xtensor<double, 3>` hence a copy will be made.

PR #1576, opened 2021-06-19
Bump dependency versions in Dockerfile

garth-wells:


jhale (approved):
Will be built automatically late tonight/early tomorrow morning.

PR #1586, opened 2021-06-23
Get entity closure dofs from ffc rather than recomputing

mscroggs:


cpp/dolfinx/fem/utils.cpp
```diff
@@ -83,22 +83,28 @@ fem::create_element_dof_layout(const ufc_dofmap& dofmap,
...
     entity_dofs[dim].resize(num_entities);
+    entity_closure_dofs[dim].resize(num_entities);
     for (int i = 0; i < num_entities; ++i)
     {
       work_array.resize(num_entity_dofs[dim]);
       dofmap.tabulate_entity_dofs(work_array.data(), dim, i);
       entity_dofs[dim][i] = std::set<int>(work_array.begin(), work_array.end());
-      dof_count += num_entity_dofs[dim];
+
+      work_array.resize(num_entity_closure_dofs[dim]);
```
garth-wells, line 105: I wouldn't worry about using a work array here - the loop is small and only called once or twice in a computation. Fine to just allocate inside the loop.

garth-wells (approved):

PR #1589, opened 2021-06-24
Fix array2d for USM allocators

IgorBaratta:
```c++
_storage = std::vector<T, Allocator>(shape[0] * shape[1], value, alloc);
```
Calls a deleted copy constructor with USM allocator.


garth-wells:
I'm wondering if we should remove `array2d` and just pass a `std::vector`/`std::span` and the row offset.

garth-wells (approved):

PR #1590, opened 2021-06-24
Expose map pull_back and test it with dolfinx.Expression

jorgensd:
Required for fast point evaluation for time dependent problems.


garth-wells (approved):

jorgensd:
@garth-wells can we merge this once Ci is green?

PR #1598, opened 2021-06-30
Interpolation between different meshes

massimiliano-leoni:
### Note: The post below describes the first proposed implementation. I later developed a second one, so this post is no longer relevant. See https://github.com/FEniCS/dolfinx/pull/1598#issuecomment-915942471

In this branch, I implemented functionalities to interpolate discrete functions between different meshes.

I will now give a brief explanation of the algorithm that I implemented.

### The serial algorithm
Consider two discrete functions `v`, the function to be interpolated [source function], and `u`, the interpolating function [target function].
The idea behind my implementation is to reuse the existing code that interpolates from a user-provided analytical expression, namely
```
void interpolate(
    Function<T>& u,
    const std::function<xt::xarray<T>(const xt::xtensor<double, 2>&)>& f,
    const xt::xtensor<double, 2>& x, const xtl::span<const std::int32_t>& cells)
```
In there, this method computes `auto values = f(x)` and then uses these values to define the interpolating function `u`.
I figured that if I can obtain those same values from `v`, as opposed to from `f`, then it will all be downhill from there. To this end, I recycle existing code to compute the points at which `v` needs to be sampled in order to define `u` [`fem::interpolation_coords`] with the idea of using `v.eval` on those points to sample `v`.
Before I can do that, I need to find out which cell of `v`'s mesh contains each of these points. I use the existing `BoundingBoxTree` and collision detection functionalities to accomplish this quickly.

### The parallel algorithm
When running in parallel, the biggest issue is that, since meshes are partitioned independently, the cell that contains a given point will not, in general, reside on the same process that contains that point.
The memory-optimal way of doing this would be, for each process `p`, to find the processes owning the cells that contain the points `p` owns, to later arrange the optimal amount of data exchange.
My current implementation follows a different philosophy.
Initially, each process contains the list of points at which it needs to evaluate `v`in order to define `u`.
First of all, I gather all these lists into a big list on every process: now, every process contains the list of all the points needed by all the processes.
I then proceed to find, on every process, which points lie in cells that that process owns. Although these are many points, for large numbers of processes it is likely that most of these points will lie outside the mesh that a single process contains, and those will be very fast checks.
Next, I evaluate `v` at all those points on every process. This way, every point is evaluated on at least one process. Again, since most points will likely lie outside of a process's local mesh, most of these will be ignored.
Finally, I can combine these values back on all processes, so that each process gets the full list of values matching the full list of points. The portion of interest can then be extracted and passed to the existing `interpolate` function for local interpolation.

### Note
I removed the Work-in-progress tag although some performance optimisation should still be possible.


massimiliano-leoni:
I forgot to mention, I added a demo case in the `demo` folder to demonstrate the interpolation, so you can use that for testing -- it doesn't have to stay there, I just didn't know where to put it.

jorgensd:
I would strongly suggest that you use: `compute_global_tree` and `compute_collisions_point` to reduce the number of points shared with each process. See for instance:
https://github.com/FEniCS/dolfinx/blob/d316ef77c58883f09d0e68486aa1caa24a10f6ec/python/test/unit/geometry/test_bounding_box_tree.py#L431
With such an implementation, you can gather all ranks that might intersect with the points, and create a neighbourhood communicator that only communicates the subset of points that are relevant

massimiliano-leoni:
Thanks @jorgensd for the feedback! The one you suggest seems to be the other approach I mentioned in the first post, is it not? I have been wondering if the overhead of sending more nodes than necessary outweighs the overhead of sending several messages: in case I wanted to only send the relevant nodes, I'm thinking I would need to orchestrate several independent transfers, having one communicator per process, is this correct?

jorgensd:
Imagine you have two sets of meshes with 500 millions vertices in each (and assuming that we have a CG-1 space that would be the same number of dofs). If we want to interpolate these between processes, and we have the unfortunate scenario that there is no process overlap between the partitions, we would need to send these to all processes (but only one or a few actually needs it).

You can use unsymmetric MPI neighborhood communciators to send the subset information back and forth between processes https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/common/MPI.h#L92-L93 (it is used through dolfin-x, and also in my MPC library https://github.com/jorgensd/dolfinx_mpc). 

One thing to note is that by using process bounding boxes, it is still not guaranteed that the process has the point in it. However, it is a very easy way to reduce the amount of data that is communicated.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -107,8 +147,495 @@ void interpolate_from_any(Function<T>& u, const Function<T>& v)
...
+    std::vector<std::vector<double>> pointsToSend(nProcs);
+    for (decltype(candidates.size()) i = 0; i < candidates.size(); ++i)
+    {
+      const auto xp = x_t(i, 0);
+      const auto yp = x_t(i, 1);
+      const auto zp = x_t(i, 2);
+
+      for (const auto& p : candidates[i])
+      {
+        pointsToSend[p].push_back(xp);
```
jorgensd, line 198: Using the insert function, as in: https://github.com/FEniCS/dolfinx/blob/3f8fa273315b904a9f394ea3d7072e7b940997fd/cpp/dolfinx/la/SparsityPattern.cpp#L59
i.e.
```
auto point = xt::row(x_t, i);
pointsToSend[p].insert(pointsToSend[p].end(), point.begin(), point.end())
```

cpp/dolfinx/fem/interpolate.h
```diff
@@ -107,8 +147,495 @@ void interpolate_from_any(Function<T>& u, const Function<T>& v)
...
+    for (decltype(x_t.shape(0)) i = 0; i < x_t.shape(0); ++i)
+    {
+      const auto xp = x_t(i, 0);
+      const auto yp = x_t(i, 1);
+      const auto zp = x_t(i, 2);
+      candidates[i]
+          = dolfinx::geometry::compute_collisions(globalBB, {xp, yp, zp});
+    }
+
+    std::vector<std::vector<double>> pointsToSend(nProcs);
```
jorgensd, line 189: You should use reserve on all the sub-vectors, otherwise the push back in the next loop is going to be quite expensive. I would suggest that each sub-vector reserves the size `3*candidates.size()`

cpp/dolfinx/fem/interpolate.h
```diff
@@ -107,8 +147,495 @@ void interpolate_from_any(Function<T>& u, const Function<T>& v)
...
+          const auto z4 = xv[3 * vtx[3] + 2];
+
+          // We compute the barycentric coordinates of the given point
+          // in the cell at hand
+          const xt::xarray<double> A = {{x1 - x4, x2 - x4, x3 - x4},
+                                        {y1 - y4, y2 - y4, y3 - y4},
+                                        {z1 - z4, z2 - z4, z3 - z4}};
+          const xt::xarray<double> b = {xp - x4, yp - y4, zp - z4};
+
+          const xt::xarray<double> l = xt::linalg::solve(A, b);
```
jorgensd, line 295: Note that xt::linalg was removed from dolfinx in https://github.com/FEniCS/dolfinx/pull/1703

cpp/dolfinx/fem/interpolate.h
```diff
@@ -107,8 +147,495 @@ void interpolate_from_any(Function<T>& u, const Function<T>& v)
...
+        const double xp = pointsToReceive(i, 0);
+        const double yp = pointsToReceive(i, 1);
+        const double zp = pointsToReceive(i, 2);
+
+        // For each cell that might contain that point
+        for (const auto& j :
+             dolfinx::geometry::compute_collisions(bbt, {xp, yp, zp}))
+        {
+          // Get the vertexes that belong to that cell
+          const auto vtx = connectivity.links(j);
```
jorgensd, line 273: Is there as specific reason for you are not using: https://github.com/FEniCS/dolfinx/blob/d316ef77c58883f09d0e68486aa1caa24a10f6ec/cpp/dolfinx/geometry/utils.h#L83-L94

Especially as the code below seems to only work for tetrahedrons.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -107,8 +147,495 @@ void interpolate_from_any(Function<T>& u, const Function<T>& v)
...
+      {
+        // Get its coordinates
+        const double xp = pointsToReceive(i, 0);
+        const double yp = pointsToReceive(i, 1);
+
+        // For each cell that might contain that point
+        for (const auto& j :
+             dolfinx::geometry::compute_collisions(bbt, {xp, yp, 0}))
+        {
+          // Get the vertexes that belong to that cell
```
jorgensd, line 326: Similar comment as above

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: Is 3 an heuristic guess or did you mean `nProcs`?

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: I request it explicitly in the `CMakeLists.txt` file, but I agree this should be addressed better.

cpp/dolfinx/fem/interpolate.h
jorgensd: 3 because you send 3 points, and candidates.size as it is the maximum number of possible points

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: I didn't know of this function, I'll check it out. Thanks!

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: Ah, I understand what you mean now. That would mean allocating enough memory to send each point to every process, right?

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: And yes, the code is only meant to work for simplicial meshes in 2 and 3 dimensions. Does FEniCS-X support non-simplicial meshes too? I never thought of it as I only ever used triangles and tetras :_D

cpp/dolfinx/fem/interpolate.h
jorgensd: Yes, but most of it might be unused. We should at least reserve some memory, as using push back does several reserves.

cpp/dolfinx/fem/interpolate.h
jorgensd: Yes, FEniCSx supports, quads and hexes (and soon prisms as @chrisrichardson  is working on it)

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: It is supposed to be logarithmically amortised, but I don't mind, this is an implementation detail that can always be changed. I wonder if anyone else will chip in into the discussion :thinking:

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: Ok, I looked into that function and it turns out I had reinvented the wheel once again :D
Thanks for pointing that out, I switched to using it of course.

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: That piece of code was removed.

massimiliano-leoni:
I did some quick timing tests on a cluster with InfiniBand interconnect, nodes with 24 threads each, with a mesh consisting of 24M tetras.
```
  n       t        decrease factor
  24   371.889
  48   190.165   -0.9676203339015955
  96   158.629   -0.26159519798458825
 144   109.428   -0.9157417895107853
 192    85.247   -0.8680202825388009
 240    70.104   -0.8764450956489861
```

Where the decrease factor is the  in  and perfect strong scalability corresponds to , so it seems to be scaling at least decently.

massimiliano-leoni:
More timings, just for the fun of it, with a P2 function on the same mesh.

```
n    t              decrease factor
24   1007.359714
48   506.031528    -0.9932797623666415
96   426.84532     -0.24551391257479313
144  298.413884    -0.882789410980575
192  223.71631     -1.0014618462533544
240  183.774946    -0.8813469419348123
```

cpp/dolfinx/fem/interpolate.h
```diff
@@ -598,8 +629,252 @@ void interpolate(Function<T>& u, const Function<T>& v,
...
+        {
+          x_t(i, j) = x(j, i);
+        }
+      }
+
+      // Create a global bounding-box tree to quickly look for processes that
+      // might be able to evaluate the interpolating function at a given point
+      dolfinx::geometry::BoundingBoxTree bb(*v.function_space()->mesh(), tdim,
+                                            0.0001);
+      auto globalBB = bb.create_global_tree(MPI_COMM_WORLD);
```
jorgensd, line 661: I would use `mesh->comm()` here to be on the safe side (in case mesh is defined with `MPI_COMM_SELF`). Do you agree?

cpp/dolfinx/fem/interpolate.h
```diff
@@ -598,8 +629,252 @@ void interpolate(Function<T>& u, const Function<T>& v,
...
+      // that might be able to do it
+      std::vector<std::vector<std::int32_t>> candidates(x_t.shape(0));
+      for (decltype(x_t.shape(0)) i = 0; i < x_t.shape(0); ++i)
+      {
+        const auto collisions = dolfinx::geometry::compute_collisions(
+            globalBB,
+            xt::xtensor<double, 2>({{x_t(i, 0), x_t(i, 1), x_t(i, 2)}}));
+        candidates[i] = std::vector<std::int32_t>(collisions.links(0).cbegin(),
+                                                  collisions.links(0).cend());
+      }
```
jorgensd, line 673: As `compute_collisions` (https://github.com/FEniCS/dolfinx/blob/c30adac64e5f5a19e7909666d18a8d6d8d38ebfb/cpp/dolfinx/geometry/utils.h#L50-L53) gives you an adjacency list, this is not necessary.
I would call:
```
const auto collisions = dolfinx::geometry::compute_collisions(globalBB, x_t);
```
where `collisions.links(i)` gives you the processes which collide with the ith point.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -598,8 +629,252 @@ void interpolate(Function<T>& u, const Function<T>& v,
...
+            xt::xtensor<double, 2>({{xp, yp, zp}}));
+        const std::vector<std::int32_t> intersectingCells(
+            collidingCells.links(0).cbegin(), collidingCells.links(0).cend());
+
+        // If there is any -- there should be at most one -- note it down
+        if (not intersectingCells.empty())
+        {
+          evaluationCells[i] = intersectingCells[0];
+        }
+      }
```
jorgensd, line 782: Similar comment as above. The compute_colliding cells and compute collisions works on sets of points, and should not be looped over. There is alot unecessary copying here as a result.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -598,8 +629,252 @@ void interpolate(Function<T>& u, const Function<T>& v,
...
+
+      // This transposition is a quick and dirty solution and should be avoided
+      xt::xarray<T> myVals_t = xt::zeros_like(xt::transpose(myVals));
+      for (decltype(myVals.shape(1)) i = 0; i < myVals.shape(1); ++i)
+      {
+        for (decltype(myVals.shape(0)) j = 0; j < myVals.shape(0); ++j)
+        {
+          myVals_t(i, j) = myVals(j, i);
+        }
+      }
```
jorgensd, line 872: Same comment as previously.

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: Yes, of course. Good catch.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -598,8 +629,252 @@ void interpolate(Function<T>& u, const Function<T>& v,
...
+      std::vector<std::int32_t> cells(num_cells, 0);
+      std::iota(cells.begin(), cells.end(), 0);
+
+      // Collect all the points at which values are needed to define the
+      // interpolating function
+      const xt::xtensor<double, 2> x = fem::interpolation_coords(
+          *u.function_space()->element(), *u.function_space()->mesh(), cells);
+
+      // This transposition is a quick and dirty solution and should be avoided
+      auto x_t = xt::zeros_like(xt::transpose(x));
```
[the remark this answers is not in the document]
massimiliano-leoni: I think previously I had some code that accessed `x_t`'s raw storage and thus I needed it to be stored contiguously. I must have removed that code though.
I might remove `x_t` altogether at this point.

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: Same answer, I guess :thinking:

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: Done. I was temporarily overkilling it with the copying to be on the safe side, thanks!

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: I think when I first wrote this piece of code those functions only accepted a single point. In any case, I applied the change you suggested.

cpp/demo/interpolation/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+      xt::xtensor<PetscScalar, 1> r(xt::shape({x.shape()[1]}), 0);
+      for (std::size_t i = 0; i < x.shape(1); ++i)
+      {
+        r(0, i) = std::cos(10 * x(0, i)) * std::sin(10 * x(2, i));
+        //              r(0, i) = x(1, i);
+      }
+      return r;
+    });
+    // Is this needed?
+    VecGhostUpdateBegin(uL->vector(), INSERT_VALUES, SCATTER_FORWARD);
```
jorgensd, line 54: This should not be needed, and if needed you should rather use: https://github.com/FEniCS/dolfinx/blob/d933045f09153c2e1f88682fe562440131f3f94c/cpp/dolfinx/la/Vector.h#L127-L131

cpp/demo/interpolation/main.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+  {
+    auto meshL = std::make_shared<mesh::Mesh>(generation::BoxMesh::create(
+        MPI_COMM_WORLD, {{{0.0, 0.0, 0.0}, {1.0, 1.0, 1.0}}}, {16, 16, 16},
+        mesh::CellType::tetrahedron, mesh::GhostMode::shared_facet));
+
+    auto meshR = std::make_shared<mesh::Mesh>(generation::BoxMesh::create(
+        MPI_COMM_WORLD, {{{0.25, 0.25, 0.25}, {0.75, 0.75, 0.75}}},
+        {16, 16, 16}, mesh::CellType::tetrahedron,
+        mesh::GhostMode::shared_facet));
```
jorgensd, line 33: What is the interpolation behaviour if you would do `interpolate(uL, uR)` where several points will find no candidate on the other mesh?

cpp/dolfinx/fem/interpolate.h
```diff
@@ -602,8 +633,217 @@ void interpolate(Function<T>& u, const Function<T>& v,
...
+
+      // Collect all the points at which values are needed to define the
+      // interpolating function
+      const xt::xtensor<double, 2> x = fem::interpolation_coords(
+          *u.function_space()->element(), *u.function_space()->mesh(), cells);
+
+      // Create a global bounding-box tree to quickly look for processes that
+      // might be able to evaluate the interpolating function at a given point
+      dolfinx::geometry::BoundingBoxTree bb(*v.function_space()->mesh(), tdim,
+                                            0.0001);
```
jorgensd, line 654: Is the `padding` of 0.0001 needed here?

cpp/demo/interpolation/main.cpp
massimiliano-leoni: Thanks, this was again written before the new API appeared. I changed it, even though this test was not really meant to stay.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -602,8 +633,217 @@ void interpolate(Function<T>& u, const Function<T>& v,
...
+      std::vector<std::int32_t> allPointsToSend(nProcs * nProcs);
+      MPI_Allgather(nPointsToSend.data(), nProcs, MPI_INT32_T,
+                    allPointsToSend.data(), nProcs, MPI_INT32_T,
+                    MPI_COMM_WORLD);
+
+      std::size_t nPointsToReceive = 0;
+      for (int i = 0; i < nProcs; ++i)
+      {
+        nPointsToReceive += allPointsToSend[mpi_rank + i * nProcs];
+      }
```
jorgensd, line 694:
```suggestion
        nPointsToReceive += allPointsToSend[mpi_rank + i * nProcs];
```

cpp/demo/interpolation/main.cpp
massimiliano-leoni: If a point in the target mesh cannot be evaluated by the source mesh, its value is set to zero.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -602,8 +633,217 @@ void interpolate(Function<T>& u, const Function<T>& v,
...
+      // Compute the offset sendingOffsets[i] at which data will be sent to
+      // process i
+      std::vector<std::int32_t> sendingOffsets(nProcs, 0);
+      for (int i = 0; i < nProcs; ++i)
+      {
+        for (int j = 0; j < mpi_rank; ++j)
+        {
+          sendingOffsets[i] += allPointsToSend[nProcs * j + i];
+        }
+      }
```
jorgensd, line 705:
```suggestion
        for (int j = 0; j < mpi_rank; ++j)
          sendingOffsets[i] += allPointsToSend[nProcs * j + i];
```

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: It might be needed if the meshes only share a boundary. Or at least this is my interpretation of the padding.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -602,8 +633,217 @@ void interpolate(Function<T>& u, const Function<T>& v,
...
+
+      for (decltype(pointsToReceive.shape(0)) i = 0;
+           i < pointsToReceive.shape(0); ++i)
+      {
+        // If any cell was found at all -- there should be at most one -- note
+        // it down
+        if (not collidingCells.links(i).empty())
+        {
+          evaluationCells[i] = collidingCells.links(i)[0];
+        }
```
jorgensd, line 759:
```suggestion
          evaluationCells[i] = collidingCells.links(i)[0];
```

cpp/dolfinx/fem/interpolate.cpp
```diff
@@ -63,3 +63,9 @@ fem::interpolation_coords(const fem::FiniteElement& element,
   return x;
 }
 //-----------------------------------------------------------------------------
+/// This should be hidden somewhere
```
chrisrichardson, line 66: See `dolfinx::MPI::mpi_type()`

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: The source code is inconsistent with respect to this particular detail of the coding style [both versions appear here and there]. If this is the official convention, however, just say so and I'll change it.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -602,8 +633,217 @@ void interpolate(Function<T>& u, const Function<T>& v,
...
+      std::transform(pointsToSend.cbegin(), pointsToSend.cend(),
+                     nPointsToSend.begin(),
+                     [](const auto& el) { return el.size(); });
+
+      // Share a matrix where element (i, j) is the number of coordinates that
+      // process i sends to process j
+      std::vector<std::int32_t> allPointsToSend(nProcs * nProcs);
+      MPI_Allgather(nPointsToSend.data(), nProcs, MPI_INT32_T,
+                    allPointsToSend.data(), nProcs, MPI_INT32_T,
+                    MPI_COMM_WORLD);
```
chrisrichardson, line 688: `MPI_COMM_WORLD` should be replaced with a user defined comm, e.g. from the mesh.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -602,8 +633,217 @@ void interpolate(Function<T>& u, const Function<T>& v,
     // Get mesh and check that functions share the same mesh
     if (mesh != v.function_space()->mesh())
     {
-      throw std::runtime_error(
-          "Interpolation on different meshes not supported (yet).");
+      const int nProcs = dolfinx::MPI::size(MPI_COMM_WORLD);
```
chrisrichardson, line 636: Don't use MPI_COMM_WORLD

jorgensd:
We would need some tests to make sure this is correct. I.e. Interpolate a known function onto a sub-set (similar to your C++ demo), and check that the interpolated function has the correct values.

chrisrichardson:
There are questions about scalability of this code. When every process communicates with every other process, it becomes a quadratic scaling, which is not good. Obviously, it can be difficult to avoid, but you should have a good idea of the communication pattern from the initial global BBtree collisions, and can then limit the communication to the processes that actually need it, e.g. by using a neighborhood communicator.

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: Thanks, I changed all occurrences. Should I also check that the two meshes have the same communicator?

cpp/dolfinx/fem/interpolate.cpp
massimiliano-leoni: Done, thanks!

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: Thanks, I changed all of them :)

massimiliano-leoni:
> We would need some tests to make sure this is correct. I.e. Interpolate a known function onto a sub-set (similar to your C++ demo), and check that the interpolated function has the correct values.

I can prepare a unit test that also validates the interpolated values, if you'd like me to.

jorgensd:
> > There are questions about scalability of this code. When every process communicates with every other process, it becomes a quadratic scaling, which is not good. Obviously, it can be difficult to avoid, but you should have a good idea of the communication pattern from the initial global BBtree collisions, and can then limit the communication to the processes that actually need it, e.g. by using a neighborhood communicator.
> 
> I think this already happens. Please be advised that the currently implemented algorithm is the one described at [#1598 (comment)](https://github.com/FEniCS/dolfinx/pull/1598#issuecomment-915942471) , according to which a process only communicates with processes that can possibly evaluate its points.

I guess that is not strictly true, as you would have a send->recv of a an array with 0 data size for all processes with no incoming points.

massimiliano-leoni:
> I guess that is not strictly true, as you would have a send->recv of a an array with 0 data size for all processes with no incoming points.

The current implementation with one-sided communication has no send and receive, it just copies data onto the target memory space. This can be implementation-dependent, but I would imagine that a copy of size zero is actually a no-op, since no synchronisation with the target process is needed. We can enforce this with an `if` statement anyway.

massimiliano-leoni:
> The unit test isn't working and it seems to be an issue with MPICH [it works on my machine with OpenMPI]. How do you see temporarily ignoring the unit test until this issue is fixed?

Of course it wasn't an issue in MPICH, it was my fault [and mine alone!]. I see green light from the unit tests now :D

garth-wells:
Closing, see updated version in #2245.

PR #1601, opened 2021-07-05
Cleanup helmholtz demo

jorgensd:
Fix a few things in the Helmholtz demo:
- Use solver that is consistent in serial and parallel (reported by @adeebkor ).
- Remove unused function
- <s>Use scatter forward after interpolation

### DIFF
--- a/python/demo/helmholtz2D/demo_helmholtz_2d.py
+++ b/python/demo/helmholtz2D/demo_helmholtz_2d.py
@@ -47,7 +47,6 @@ u = TrialFunction(V)
 v = TestFunction(V)
 f = Function(V)
 f.interpolate(lambda x: A * k0**2 * np.cos(k0 * x[0]) * np.cos(k0 * x[1]))
-f.x.scatter_forward()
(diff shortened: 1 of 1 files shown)


python/demo/helmholtz2D/demo_helmholtz_2d.py
```diff
@@ -47,18 +47,18 @@
 v = TestFunction(V)
 f = Function(V)
 f.interpolate(lambda x: A * k0**2 * np.cos(k0 * x[0]) * np.cos(k0 * x[1]))
+f.x.scatter_forward()
```
garth-wells, line 50: Is this call needed? I would have expected the dofs for `f` to all be correctly computed because they just depend on `x`, which available on all ranks.

jorgensd: You are right. I thought we only did interpolation on local cells, but we also do it on ghost cells.

garth-wells (approved):

PR #1602, opened 2021-07-06
Add MPI_Type_dup to copy and move

chrisrichardson:
Fixes a double-free memory problem in la::Vector

### DIFF
--- a/cpp/dolfinx/la/Vector.h
+++ b/cpp/dolfinx/la/Vector.h
@@ -46,17 +46,17 @@ public:
         _buffer_recv_fwd(x._buffer_recv_fwd), _x(x._x)
   {
     MPI_Type_dup(x._datatype, &_datatype);
-  };
+  }
 
   /// Move constructor
   Vector(Vector&& x)
       : _map(std::move(x._map)), _bs(std::move(x._bs)),
-        _request(std::move(x._request)),
+        _datatype(std::exchange(x._datatype, MPI_DATATYPE_NULL)),
+        _request(std::exchange(x._datatype, MPI_REQUEST_NULL)),
         _buffer_send_fwd(std::move(x._buffer_send_fwd)),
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/la/Vector.h
```diff
@@ -40,15 +40,28 @@ class Vector
...
 
   /// Move constructor
-  Vector(Vector&& x) noexcept = default;
+  Vector(Vector&& x)
+      : _map(std::move(x._map)), _bs(std::move(x._bs)),
+        _request(std::move(x._request)),
+        _buffer_send_fwd(std::move(x._buffer_send_fwd)),
+        _buffer_recv_fwd(std::move(x._buffer_recv_fwd)), _x(std::move(x._x))
+  {
+    MPI_Type_dup(x._datatype, &_datatype);
```
garth-wells, line 58: I don't think this should duplicate. Should be similar to https://github.com/FEniCS/dolfinx/blob/43847f184f3aa69f166de5fafe9584de914b4e11/cpp/dolfinx/la/PETScVector.cpp#L202.

cpp/dolfinx/la/Vector.h
```diff
@@ -40,15 +40,28 @@ class Vector
...
+        _buffer_recv_fwd(x._buffer_recv_fwd), _x(x._x)
+  {
+    MPI_Type_dup(x._datatype, &_datatype);
+  };
 
   /// Move constructor
-  Vector(Vector&& x) noexcept = default;
+  Vector(Vector&& x)
+      : _map(std::move(x._map)), _bs(std::move(x._bs)),
+        _request(std::move(x._request)),
```
garth-wells, line 54: This should probably use `std::exchange`.

garth-wells (approved):

garth-wells (approved):

IgorBaratta (approved):

PR #1611, opened 2021-07-09
Push docker images with tag prefix

### LINKED ISSUE #1610: Create docker images of version releases
For easy access to debugging and end-users

mscroggs:
When a tag is created, this will push docker images with that tag (eg `dolfinx/dolfinx:v0.1.1-arm64` `dolfinx/lab:v0.1.1-amd64` etc).

Fixes #1610

### DIFF
--- a/.github/workflows/docker-end-user.yml
+++ b/.github/workflows/docker-end-user.yml
@@ -11,7 +11,7 @@ on:
   push:
     # This will run every time a version tag is created
     tags:
-      - 'v*'
+      - '*'
   schedule:
     # '*' is a special character in YAML, so string must be quoted
     - cron: "0 5 * * *"
--- a/.github/workflows/docker-end-user.yml
+++ b/.github/workflows/docker-end-user.yml
@@ -8,6 +8,10 @@ on:
   #push:
   #  branches:
   #    - "**"
+  push:
+    # This will run every time a version tag is created
+    tags:
+      - 'v*'
   schedule:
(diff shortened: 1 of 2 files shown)


.github/workflows/docker-end-user.yml
```diff
@@ -8,6 +8,10 @@ on:
   #push:
   #  branches:
   #    - "**"
+  push:
+    # This will run every time a version tag is created
+    tags:
+      - 'v*'
```
jhale, line 14: Are we going to use the `v` prefix? It does seem to be the custom on GitHub. @chrisrichardson?

jhale (commented):
Thanks this is really helpful. It's also useful to have the base development images around. Could you add the same trick there?

https://github.com/FEniCS/dolfinx/blob/main/.github/workflows/docker-build-env.yml

.github/workflows/docker-end-user.yml
mscroggs: Good spot. The first release doesn't have a v. I guess we can actually just make this push for any tags

mscroggs: I've remove the v prefix

jhale: Well, I guess that's the reason to have `v`, to distinguish a release from any other tag. I think @chrisrichardson decided not to have `v` for a reason but I forgot why. 

On balance, I think we can trust people not to push tags.

PR #1618, opened 2021-07-15
Expose degree of CoordinateElement

jorgensd:
This variable is needed if we want to recreate the coordinate element for higher order geometries in custom assemblers


garth-wells:
I think this should be put on hold - there may be a better way, and it's not needed immediately so there is time to think about it.

PR #1622, opened 2021-07-15
Add radix sort and argsort (lexicographic)

IgorBaratta:
- Adds a function to sort integers in O(n) time with O(n) storage.
- Adds a function to sort (lexicographic) an array of ints.

Example of the impact on create entities:
image

Comparision with std::sort, boost::sort::integer_sort for sorting `cell indices`:
sort


cpp/dolfinx/common/sort.h
```diff
@@ -0,0 +1,146 @@
...
+#include <numeric>
+#include <vector>
+
+namespace dolfinx
+{
+
+// Sort a vector with radix sorting algorithm.
+// The bucket size is determined by the number of bits to sort at a time.
+template <typename T, int BITS = 8>
+void radix_sort(std::vector<T>& array)
```
garth-wells, line 23: Could we make this a span?

garth-wells (approved):

garth-wells (approved):

PR #1629, opened 2021-07-16
Add missing overloaded BBox collisions method

nate-sime:
Simple addition of missing method from the wrapper interface.


python/dolfinx/wrappers/geometry.cpp
```diff
@@ -46,6 +46,10 @@ void geometry(py::module& m)
         py::overload_cast<const dolfinx::geometry::BoundingBoxTree&,
                           const dolfinx::geometry::BoundingBoxTree&>(
             &dolfinx::geometry::compute_collisions));
+  m.def("compute_collisions",
```
garth-wells, line 49: How big will the returned `std::vector` typically be?

nate-sime: Less than or equal to the MPI comm size. Is there a better data structure?

garth-wells: See https://github.com/FEniCS/dolfinx/blob/28edfc1d74bb9029bf51396b52d9ae88ede9f10e/python/dolfinx/wrappers/mesh.cpp#L324 for an example of how to return a NumPy array without the very expensive copy of `std::vector` into a `list`.

garth-wells (approved):

PR #1648, opened 2021-07-30
ADIOS2 output (FIDES support)

jorgensd:
As Paraview now comes with FIDES support (For mac it should have the Fides Reader, for linux use: https://data.kitware.com/api/v1/file/61016c222fa25629b943e716/download) we should think about using it to support output.

Current PR introduces ADIOSFile (could maybe be called FIDESWriter?)
which writes meshes and lists of functions to file.
**Functionality**
- Write mesh to file (first order geometries)
- Write function (P1 interpolation to vertices, both real and complex mode)
  - Scalar fields
  - Vector fields


jorgensd:
Needs a rewrite for time dependent problems

PR #1651, opened 2021-08-02
Update Basix interface to allow variants

mscroggs:
In FEniCS/basix#252, we allow the user to create variants of Lagrange spaces. This PR allows these to be used in DOLFINx.

This also replaces the initialisation of Basix elements with strings with tidier initialisation with enums.

This is coupled with FEniCS/ffcx#376 and FEniCS/basix#252.

### DIFF
--- a/cpp/dolfinx/fem/FiniteElement.cpp
+++ b/cpp/dolfinx/fem/FiniteElement.cpp
@@ -144,8 +144,19 @@ FiniteElement::FiniteElement(const ufc_finite_element& ufc_element)
   if (is_basix_element(ufc_element))
   {
     // FIXME: Find a better way than strings to initialise this Basix element
-    _element = std::make_unique<basix::FiniteElement>(basix::create_element(
-        _family.c_str(), cell_shape.c_str(), ufc_element.degree));
+    if (ufc_element.needs_lattice_type)
+    {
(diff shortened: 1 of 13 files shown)


garth-wells (commented):
Do we have a way yet to know which Lagrange element type was used in FFCx?

mscroggs:
Not yet I opened an issue in Basix (https://github.com/FEniCS/basix/issues/253) about this

cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -143,19 +143,18 @@ FiniteElement::FiniteElement(const ufc_finite_element& ufc_element)
 
   if (is_basix_element(ufc_element))
   {
-    // FIXME: Find a better way than strings to initialise this Basix element
     if (ufc_element.needs_lattice_type)
     {
       _element = std::make_unique<basix::FiniteElement>(basix::create_element(
-          basix::element::str_to_type(_family.c_str()),
-          basix::cell::str_to_type(cell_shape.c_str()), ufc_element.degree,
+          (basix::element::family)ufc_element.basix_family,
```
garth-wells, line 149: The 'proper' C++ way to do this would be `static_cast<basix::element::family>(ufc_element.basix_family)`.

garth-wells (approved):

PR #1655, opened 2021-08-06
ADIOS2 support (VTX and Fides)

jorgensd:
This PR introduces ADIOS2 as an engine for outputting meshes and functions.
## VTX 
Format described by ADIOS2 (https://adios2.readthedocs.io/en/latest/ecosystem/visualization.html) using VTK as backend.
Only uses local to process geometry/topology for outputting
Supports:
- Time dependent output of meshes and functions
- Arbitrary order Lagrange/Discontinuous Lagrange as output (For a single function at the time)
- Interpolated data to mesh nodes for lists of functions (or for functions of other spaces than Lagrange/Disc. Lagrange).
## Fides
New visualization engine using VTK-M, docs available at: https://fides.readthedocs.io/en/latest/?badge=latest
Currently only supports first order Lagrange visualization in Paraview (Linux binary available at: https://data.kitware.com/api/v1/file/61016c222fa25629b943e716/download)
Some bugs in visualization of time dependent meshes/functions (file needs to be refreshed in paraview after moving from one step to the other).


cpp/dolfinx/io/adios2_utils.h
```diff
@@ -0,0 +1,150 @@
...
+}
+
+namespace io
+{
+// Convenience functions for declaring attributes and variables with adios2
+namespace adios2_utils
+{
+
+//-----------------------------------------------------------------------------
+/// Extract the mesh connectivity for all cells of a given mesh
```
garth-wells, line 36: I don't understand what's returned here, and it doesn't look to be adios2 specific.

cpp/dolfinx/io/cells.cpp
```diff
@@ -275,6 +275,74 @@ std::vector<std::uint8_t> gmsh_quadrilateral(int num_nodes)
     throw std::runtime_error("Higher order GMSH quadrilateral not supported");
   }
 }
+//-----------------------------------------------------------------------------
+std::vector<std::uint8_t> discontinuous_triangle(int num_nodes)
```
garth-wells, line 279: Could this function name be more expressive?

cpp/dolfinx/io/cells.cpp
```diff
@@ -341,6 +409,33 @@ std::vector<std::uint8_t> io::cells::perm_gmsh(const mesh::CellType type,
   return io::cells::transpose(map);
 }
 //-----------------------------------------------------------------------------
+std::vector<std::uint8_t> io::cells::perm_discontinuous(mesh::CellType type,
```
garth-wells, line 412: I think most of the added functions need to be made more expressive.

cpp/dolfinx/io/adios2_utils.h
jorgensd: It is extracting the mesh connectivity for each cell. It is not the topology. i.e. for two triangles you could get
[[0, 1, 2], [1,2,3]]

cpp/dolfinx/io/cells.cpp
jorgensd: DG triangle, as it is directly related to the DG function space?

cpp/dolfinx/io/cells.cpp
jorgensd: `perm_DG`?

cpp/dolfinx/io/adios2_utils.h
garth-wells: I don't see how that's not the cell topology.

jorgensd: The DOLFINx topology is first order.
The connectivity here supports higher order cells. 
We could try to extend https://github.com/FEniCS/dolfinx/blob/d316ef77c58883f09d0e68486aa1caa24a10f6ec/cpp/dolfinx/mesh/utils.h instead, or make a similar function in utils.

cpp/dolfinx/io/cells.cpp
jorgensd: No longer needed

PR #1666, opened 2021-08-13
Bump min CMake version to 3.16. Earlier versions are not tested.

garth-wells:
Earlier versions can cause problems with 'modern' configuration of dependencies.


jhale (approved):
It's OK for me, I've basically come to terms with building my own cmake on HPC systems anyway!

IgorBaratta (approved):

PR #1676, opened 2021-08-19
Support time-dependent meshes in XDMF files

massimiliano-leoni:
In this branch I added functionalities to save time-dependent meshes to XDMF files. In the created XML file, `Grid` nodes are now temporal collections, and each grid has a `Time` node, similarly to how it's done for `Function`s. Compatibility is not broken as the mesh time defaults to zero and writing a function still automatically picks the correct mesh if there is only one.
When multiple meshes are present, it is the user's responsibility to pass the correct XPath to `XDMFFile::write_function`, which is not too hard.


garth-wells:
Closing because efforts are moving away from XDMF.

PR #1680, opened 2021-08-20
Compress index map

jorgensd:
Function that takes an index map and set of local entities, and creates a new smaller index map only containing these entities (including ghosts from other processes).

Returns the new index map and a vector mapping from the new local index to the original global index.

This function is step 1 towards having a MeshView class


cpp/dolfinx/common/utils.cpp
```diff
@@ -5,10 +5,17 @@
 // SPDX-License-Identifier:    LGPL-3.0-or-later
 
 #include "utils.h"
+#include <numeric>
 #include <sstream>
+#include <unordered_map>
+#include <xtensor/xadapt.hpp>
+#include <xtensor/xio.hpp>
```
IgorBaratta, line 12:
```suggestion
```

cpp/dolfinx/common/utils.cpp
```diff
@@ -5,10 +5,17 @@
 // SPDX-License-Identifier:    LGPL-3.0-or-later
 
 #include "utils.h"
+#include <numeric>
 #include <sstream>
+#include <unordered_map>
+#include <xtensor/xadapt.hpp>
```
IgorBaratta, line 11:
```suggestion
```

cpp/dolfinx/common/utils.cpp
```diff
@@ -24,3 +29,137 @@ std::string dolfinx::common::indent(std::string block)
...
+  // Split indices into local and ghost indices
+  const std::int32_t local_size = map->size_local();
+  std::vector<std::int32_t> local_indices;
+  std::vector<std::int32_t> ghost_indices;
+  std::vector<std::int32_t> num_ghosts(source_ranks.size());
+
+  for (std::size_t i = 0; i < indices.size(); ++i)
+  {
+    if (indices[i] < local_size)
+      local_indices.push_back(indices[i]);
```
IgorBaratta, line 60: Could we preallocate the vector? 
Either reserving `local_indices(indices.size())` or by counting ?

python/test/unit/common/test_index_map.py
```diff
@@ -0,0 +1,65 @@
...
+    sl = vertex_map.size_local
+    org_ghosts = vertex_map.ghosts
+    org_ghost_owners = vertex_map.ghost_owner_rank()
+
+    # Create an index map where only every third local index is saved
+    num_owned = 0
+    entities = []
+    for i in range(sl):
+        if i % 4 == 0:
+            entities.append(i)
```
IgorBaratta, line 25: Should we use `numpy.arange(0, sl, 4)` instead of a loop?

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -130,13 +130,6 @@ compute_owned_shared(MPI_Comm comm, const xtl::span<const std::int64_t>& ghosts,
     }
   }
 
-  // Work-around for OpenMPI bug. Some version of OpenMPI crash if
```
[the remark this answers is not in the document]
jorgensd: I think they are still needed.

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -682,4 +661,198 @@ std::map<std::int32_t, std::set<int>> IndexMap::compute_shared_indices() const
...
+    }
+  }
+
+  // Compute ghost_pos_recv_fwd
+  std::vector<std::int32_t> tmp_displs = displs_recv_fwd;
+  std::vector<std::int32_t> ghost_pos_recv_fwd(ghost_owner.size());
+  for (std::size_t i = 0; i < ghost_owner.size(); ++i)
+    ghost_pos_recv_fwd[i] = tmp_displs[ghost_owner[i]]++;
+
+  // Step 8: Create neighbourhood communicators for thw new map
```
jorgensd, line 822:
```suggestion
  // Step 8: Create neighbourhood communicators for the new map
```

jpdean:
OK, I think I might know why this is failing. When running on `n` processes, the size of `ranks_old_to_new_recv` will be `n-1`, but it's creation here
https://github.com/FEniCS/dolfinx/blob/dokken/compress_index_map/cpp/dolfinx/common/IndexMap.cpp#L774
does not take into account the current rank index. This means that when doing
`submap.ghost_owner_rank()` for rank r, the entries will only be correct for the ghosts owned by ranks < r. In other words, for rank 0, all entries in `submap.ghost_owner_rank()` will be off by 1. For rank 1, `submap.ghost_owner_rank()` will return the correct values for ghosts owned by rank 0, but ghosts owned by ranks 2, 3, ..., r will be off by 1. `submap.ghost_owner_rank()` will always return the correct owners on the highest rank.

I have to go out for an hour, but I can try to fix this when I'm back.

garth-wells:
I've push at least one fix in 64c431efaa627b64c93fe3a115b46e63984de4ff. The neighbourhood rank wasn't being mapped to the global rank.

garth-wells (approved):

PR #1682, opened 2021-08-24
Pin numpy to numba compatible version

### LINKED ISSUE #1681: Downgrade numpy in Docker images
Latest release of numpy is not compatible with latest release of numba, see for instance: 
https://github.com/FEniCS/dolfinx/runs/3406497989?check_suite_focus=true
and 
https://github.com/numba/numba/blob/0.54.0/numba/__init__.py#L138

Thus we need to downgrade numpy to have numba working.

jorgensd:
Resolves #1681


chrisrichardson:
This *should* fix the problem. Would be good to get CI working again.

PR #1683, opened 2021-08-24
MeshView

jorgensd:
Adds the functionality of a MeshView.
Given a mesh, dimension and set of entities, create:
- A new topology for the entities (`_topology`), that is renumbered to use contiguous indexing for both entities and vertices.
- A map from the new (local) vertex number to the parent mesh (`_parent_vertex_map`)
- A map from the new (local) entity index to the parent mesh (`_parent_vertex_map`)
- A new geometry, directly mapping to dofs in the parent mesh geometry.


garth-wells:
Should this be closed now that https://github.com/FEniCS/dolfinx/pull/1907 has been merged?

PR #1684, opened 2021-08-25
Extend and clean-up entities to geometry

jorgensd:
The current entities to geometry has a boolean flag; orient, that only works for the case of facets of a tetrahedron. 
I propose to split the orient function into a separate function, specifically designed for facets of tetrahedra. This function is not used alot in dolfinx, but for instance in the interface to BEMPP.

Additionally, there are many instances where we would like to get all the closure dofs of an entity in the geometry (such as in a mesh-view). Therefore, `entities_to_geometry` has been changed to return the geometry nodes for all closure dofs of the entity.
The original function (without the orient flag) has been renamed `entities_to_vertex_geometry`.


jorgensd:
Close due to major changes in the C++ code since this was proposed. Should be reconsidered in the future.

PR #1685, opened 2021-08-25
Allows dofmap creation on a meshview

jpdean:


python/dolfinx/mesh_view.py
```diff
@@ -0,0 +1,20 @@
+# FIXME This seems a bit of a hack
```
jorgensd, line 1: Add header with copyright and your name here

python/test/unit/fem/test_mesh_view_spaces.py
```diff
@@ -0,0 +1,23 @@
+# TODO When cell / facet spaces are working properly, improve these tests.
```
jorgensd, line 1: Add header to file with copyright,name and licence

python/test/unit/mesh/test_meshview.py
```diff
@@ -7,6 +7,7 @@
 import dolfinx
 from mpi4py import MPI
 import numpy as np
+import ufl
```
jorgensd, line 10: Is this needed?

cpp/dolfinx/fem/FunctionSpace.h
```diff
@@ -37,6 +38,10 @@ class FunctionSpace
   FunctionSpace(std::shared_ptr<const mesh::Mesh> mesh,
                 std::shared_ptr<const fem::FiniteElement> element,
                 std::shared_ptr<const fem::DofMap> dofmap);
+  
+  FunctionSpace(std::shared_ptr<const mesh::MeshView> meshView,
```
jorgensd, line 42: Maybe we should call it `mesh_view` when it is a variable, as it is closer to how we define other variables (such as mesh, element and dofmap)

PR #1687, opened 2021-08-26
Encapsulate petsc4py and mpi4py detection in CmakeLists.txt file

garth-wells:
Less code and will support forthcoming 'isolated' Python builds with `pyproject.toml`. It's helpful for isolated builds to not have non-standard 'build' files in the source tree. We do have `CMakeFiles.txt`, but this is so common that there are tools to handle it, e.g. `scikit-build`, with isolated builds.


jhale (approved):
It's ok from my point of view.

python/CMakeLists.txt
```diff
@@ -57,18 +57,28 @@ if("XTENSOR_USE_XSIMD" IN_LIST BASIX_DEFN OR "XTENSOR_USE_XSIMD" IN_LIST DOLFINX
...
+  COMMAND "${PYTHON_EXECUTABLE}" "-c" "import mpi4py; print(mpi4py.get_include())"
+  OUTPUT_VARIABLE MPI4PY_INCLUDE_DIR
+  RESULT_VARIABLE MPI4PY_COMMAND_RESULT
+  ERROR_QUIET  OUTPUT_STRIP_TRAILING_WHITESPACE)
+if(NOT MPI4PY_COMMAND_RESULT)
+  message(STATUS "Found mpi4py include directory at ${MPI4PY_INCLUDE_DIR}")
   target_include_directories(cpp PRIVATE ${MPI4PY_INCLUDE_DIR})
-  target_compile_definitions(cpp PRIVATE HAS_PYBIND11_MPI4PY)
+else()
+  message(FATAL_ERROR "petsc4y could not be found.")
```
drew-parsons, line 83: this should be mpi4py here, not petsc4py

PR #1693, opened 2021-08-30
Use pip install pybind11[global]

jhale:
This change allows downstream packages to find pybind11's cmake find module without passing custom hints.


garth-wells:
I'm not sure this is a good idea. Outside of a virtual env it installs files in `/usr/local`. The pybind11 docs don't recommend it outside of virtual environments, see https://pybind11.readthedocs.io/en/stable/installing.html#include-with-pypi.

jhale:
You're right, it's not a good solution.

PR #1701, opened 2021-09-06
Enable xfailed test

### LINKED ISSUE #1134: Jump integrals not correct for Hdiv and Hcurl
Dof permuting has not been implemented yet to cover these, as I discovered when trying to add more tests for Dof permuting:

https://github.com/FEniCS/dolfinx/blob/13fab1e29b748de77bbc8fb1def7f9ff2227e8d4/python/test/unit/fem/test_dof_permuting.py#L524-L781

mscroggs:
It appears that whatever was causing this to fail previously is now fixed.

Close #1134.


jhale:
I'll check this, I've been having some problems with jump terms on curl and div spaces on recent commits. I have a set of commits where it did work from earlier in the summer.

jhale:
I posted a new issue here: https://github.com/FEniCS/dolfinx/issues/1708

PR #1702, opened 2021-09-07
Prism cell mesh

chrisrichardson:
Although a BoxMesh of prisms is not in itself terribly useful, this is a stepping stone to mixed-topology, and will allow us to work out some of the issues, for example: how to deal with different facet types on the same cell.

* BoxMesh of prisms
* Returns cell topology as an AdjacencyList 
* Updates topologycomputation to deal with triangle and quad facets
* Simple test


cpp/dolfinx/mesh/topologycomputation.cpp
```diff
@@ -534,61 +554,50 @@ compute_from_transpose(const graph::AdjacencyList<std::int32_t>& c_d1_d0,
...
-    {
-      auto keys_begin
-          = std::next(keys.cbegin(), i * e_vertices_ref.num_links(i));
-      auto keys_end
-          = std::next(keys.cbegin(), (i + 1) * e_vertices_ref.num_links(i));
-      std::partial_sort_copy(keys_begin, keys_end, key.begin(), key.end());
-      const auto it = entity_to_index.find(key);
-      assert(it != entity_to_index.end());
+      for (int j = 0; j < 2; ++j)
+        key[j] = e0[vref->links(i)[j]];
```
garth-wells, line 595: Would be better to move `vref->links(i)` outside of the `j` loop.

cpp/dolfinx/mesh/cell_types.h
```diff
@@ -41,21 +41,24 @@ std::string to_string(CellType type);
...
+/// Return type of cell for entity of dimension d at given entity index.
+CellType cell_entity_type(CellType type, int d, int index);
 
 /// Return facet type of cell
+/// For simplex and hypercube cell types, this is independent of the facet
+/// index, but for prism and pyramid, it can be triangle or quadrilateral.
 /// @param[in] type The cell type
-/// @return The type of the cell's facets
-CellType cell_facet_type(CellType type);
+/// @param[in] index The facet index
```
garth-wells, line 51: Do we have the 'reference' prism documented somewhere?

cpp/dolfinx/mesh/topologycomputation.cpp
```diff
@@ -404,38 +404,48 @@ compute_entities_by_key_matching(
   // Initialize local array of entities
   const std::int8_t num_entities_per_cell
       = mesh::cell_num_entities(cell_type, dim);
-  const std::size_t num_vertices_per_entity
-      = mesh::num_cell_vertices(mesh::cell_entity_type(cell_type, dim));
+
+  // For some cells, the num_vertices varies per facet (3 or 4)
+  int max_vertices_per_entity = 0;
+  for (int i = 0; i < num_entities_per_cell; ++i)
```
garth-wells, line 410: This is a bit ugly - could it be neater to use `get_sub_entities`?

cpp/dolfinx/mesh/topologycomputation.cpp
```diff
@@ -404,38 +404,48 @@ compute_entities_by_key_matching(
...
       const std::int32_t idx = c * num_entities_per_cell + i;
       auto ev = e_vertices.links(i);
 
-      // Get entity vertices
-      assert(e_vertices.num_links(i) == (int)num_vertices_per_entity);
-      for (std::size_t j = 0; j < num_vertices_per_entity; ++j)
+      // Get entity vertices padding with -1 if fewer than
+      // max_vertices_per_entity
+      entity_list(idx, max_vertices_per_entity - 1) = -1;
+      for (int j = 0; j < e_vertices.num_links(i); ++j)
```
garth-wells, line 438:
```suggestion
      for (std::size_t j = 0; j < ev.size(); ++j)
```

garth-wells (changes requested):

cpp/dolfinx/mesh/cell_types.h
chrisrichardson: Should be in basix

cpp/dolfinx/mesh/topologycomputation.cpp
chrisrichardson: I don't think we can use that, because we need something more general (`get_sub_entities` only works from face->edge).

chrisrichardson: Could do:
```
  std::vector<mesh::CellType> entity_type(num_entities_per_cell);
  for (int i = 0; i < num_entities_per_cell; ++i)
    entity_type[i] = mesh::cell_entity_type(cell_type, dim, i);

  int max_vertices_per_entity = std::transform_reduce(
      entity_type.begin(), entity_type.end(), 0,
      [](int a, int b) { return std::max(a, b); }, &mesh::num_cell_vertices)
```
or
```
int max_vertices_per_entity = 0;
 if (dim == 1)
   max_vertices_per_entity = 2;
 else if (dim == 2 and cell_type == mesh::CellType::prism)
   max_vertices_per_entity = 4;
 else if (dim == 2)
   max_vertices_per_entity = mesh::num_cell_vertices(mesh::cell_entity_type(cell_type, dim, 0);
```

I think I'm inclined to stick with the original version...

PR #1706, opened 2021-09-09
Add plotting exceptions for prisms

jorgensd:
#1702 did not address `cell_entity_type` in plotting. Currently raising exception for unsupported cell.


jorgensd:
Fixed in #1706

PR #1709, opened 2021-09-10
Remove strings from DOLFINx <-> Basix cell type conversion

mscroggs:
This allows FEniCS/Basix#270 to be done.

### DIFF
--- a/cpp/dolfinx/fem/CoordinateElement.cpp
+++ b/cpp/dolfinx/fem/CoordinateElement.cpp
@@ -33,6 +33,29 @@ double compute_determinant(Matrix& A)
     return std::sqrt(math::det(BA));
   }
 }
+//-----------------------------------------------------------------------------
+basix::cell::type cell_type_to_basix_type(mesh::CellType celltype)
+{
+  switch (celltype)
+  {
+  case mesh::CellType::interval:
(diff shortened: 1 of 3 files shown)


cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -33,6 +33,29 @@ double compute_determinant(Matrix& A)
     return std::sqrt(math::det(BA));
   }
 }
+//-----------------------------------------------------------------------------
+basix::cell::type cell_type_to_basix_type(mesh::CellType celltype)
```
garth-wells, line 37: Could the code in `cell_types.cpp` be re-used instead of this function?

mscroggs: I can't see a way to do it without including Basix in the header file, as we'd need the basix cell enum to declare the function there, unless I'm missing a good idea?

mscroggs: Just noticed that we're including basix/finite-element.h in the header of FiniteElement.h, so it's not really a problem to include basix/cell.h in cell_types.h, so I've done that and removed the duplicated code.

garth-wells (approved):

PR #1723, opened 2021-09-24
Fix finding manylinux built wheel of basix

jhale:
Suggestions welcome. Ubuntu's cmake doesn't look in lib64.


garth-wells:
I'd like to take some more time to figure this out. Doesn't feel right hardwiring the path name.

PR #1725, opened 2021-09-27
Add test for packing coeffs for form without coeffs

jorgensd:
The commit 1b1df970f13acfe022975a2eea225f1df1e50b2b introduced a bug that causes form without coefficients to segfault, as we divide by cstride (0) in the python layer.

This PR fixes the issue, and adds an additional test


IgorBaratta (approved):

garth-wells:
I've fixed this already in https://github.com/FEniCS/dolfinx/tree/garth/assembler-python-coeff. Probably better to let the branch I'm working on land to avoid merge conflicts.

PR #1726, opened 2021-09-27
Extra test for dokken/compress_index_map

jpdean:
Add a new test to https://github.com/FEniCS/dolfinx/pull/1680 to check that ghosts in the sub index map are mapped properly.

### DIFF
--- a/python/test/unit/common/test_index_map.py
+++ b/python/test/unit/common/test_index_map.py
@@ -52,8 +52,10 @@ def test_sub_index_map():
 
     # Check that ghost indices are correct in submap
     # NOTE This assumes size_local is the same for all ranks
+    # TODO Consider renaming to something shorter
     submap_global_to_map_global_map = np.concatenate([local_indices[rank] + map_local_size * rank
                                                       for rank in range(comm.size)])
+    # FIXME Do this more elegantly
     submap_ghosts = []
     for map_ghost in map.ghosts:
         submap_ghost = np.where(submap_global_to_map_global_map == map_ghost)[0]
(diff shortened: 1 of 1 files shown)


python/test/unit/common/test_index_map.py
```diff
@@ -10,54 +10,55 @@
...
+    my_rank = comm.rank
 
     # Create index map with one ghost from each other process
-
     n = 7
     assert comm.size < n + 1
-    size_local = np.math.factorial(n)
+    map_local_size = np.math.factorial(n)
 
     # FIXME: document which indices are ghosted
```
jorgensd, line 21:
```suggestion
    # The ghosts added is the ith ghost from the ith process relative to the current rank, i.e.
    # rank 0 contains the first index of rank 2, second of rank 3 etc.
    # rank 1 contains the first index of rank 0, the second of rank 2 etc.
```

python/test/unit/common/test_index_map.py
```diff
@@ -10,54 +10,55 @@
...
+    # Check that ghost indices are correct in submap
+    # NOTE This assumes size_local is the same for all ranks
+    # TODO Consider renaming to something shorter
+    submap_global_to_map_global_map = np.concatenate([local_indices[rank] + map_local_size * rank
+                                                      for rank in range(comm.size)])
+    # FIXME Do this more elegantly
+    submap_ghosts = []
+    for map_ghost in map.ghosts:
+        submap_ghost = np.where(submap_global_to_map_global_map == map_ghost)[0]
+        if submap_ghost.size != 0:
```
jorgensd, line 62: If we cannot find the ghost I guess we should raise an error?

jorgensd (changes requested):

python/test/unit/common/test_index_map.py
jpdean: You mean raise an error if `submap_ghost.size == 0`? This shouldn't raise an error, as a ghost in the original map might not be in the submap

jorgensd (approved):

PR #1728, opened 2021-09-27
Support passing of form coefficients to assemblers from Python

garth-wells:
The adds support to the Python interface that has existed in the C++ interface.

### DIFF
--- a/python/dolfinx/fem/assemble.py
+++ b/python/dolfinx/fem/assemble.py
@@ -9,12 +9,11 @@ import contextlib
 import functools
 import typing
 
-from petsc4py import PETSc
-
 import ufl
 from dolfinx import cpp
 from dolfinx.fem.dirichletbc import DirichletBC
 from dolfinx.fem.form import Form
+from petsc4py import PETSc
 
 
 def _create_cpp_form(form):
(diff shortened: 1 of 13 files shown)


michalhabera:
Maybe there could be a Python unit test if this adds the functionality to Python?

IgorBaratta (approved):

python/dolfinx/fem/assemble.py
```diff
@@ -68,52 +106,64 @@ def create_matrix_nest(a: typing.List[typing.List[typing.Union[Form, cpp.fem.For
...
+
 # -- Scalar assembly ---------------------------------------------------------
 
 
-def assemble_scalar(M: typing.Union[Form, cpp.fem.Form]) -> PETSc.ScalarType:
-    """Assemble functional. The returned value is local and not accumulated
-    across processes.
+def assemble_scalar(M: typing.Union[Form, cpp.fem.Form],
+                    coeffs=Coefficients(None, None)) -> PETSc.ScalarType:
+    """Assemble functional. The returned value is local and not
```
jorgensd, line 116: Should the documentation contain some hints on what shape or type the coefficients/constants should have, or do you think this is for expert usage only.

jorgensd (approved):

python/dolfinx/fem/assemble.py
garth-wells: I think that's more for experts, and it might change.

PR #1733, opened 2021-10-02
Fix CMake Basix detection

garth-wells:
This is for when the Basix C++ library is installed via pip.


jhale:
This line is also essential for pre-built wheels of basix: https://github.com/FEniCS/basix/pull/286/files#diff-8f5e0e0fa93b882fdb8563a1d81147a3ebfeb8628cb8f22e07413603f2925d3aR24

garth-wells:
> This line is also essential for pre-built wheels of basix: https://github.com/FEniCS/basix/pull/286/files#diff-8f5e0e0fa93b882fdb8563a1d81147a3ebfeb8628cb8f22e07413603f2925d3aR24

We should make these settings an option - not ideal to hardwire rpath settings.

PR #1742, opened 2021-10-06
(Fast?) Interpolation between spaces

IgorBaratta:
- No geometry information required.

### DIFF
--- a/cpp/dolfinx/fem/interpolate.h
+++ b/cpp/dolfinx/fem/interpolate.h
@@ -165,15 +165,14 @@ void interpolate_from_any(Function<T>& u, const Function<T>& v)
         = element_to->compute_interpolation_operator(*element_from);
 
     const int num_cells = map->size_local() + map->num_ghosts();
-
     std::vector<T> v_local(element_from->space_dimension());
     std::vector<T> u_local(element_to->space_dimension());
 
     const auto apply_dof_transformation
-        = element_from->get_dof_transformation_function<T>(false, false, false);
+        = element_from->get_dof_transformation_function<T>(false, true, false);
 
     const auto apply_inverse_dof_transform
-        = element_from->get_dof_transformation_function<T>(true, true, false);
+        = element_to->get_dof_transformation_function<T>(true, true, false);
 
     for (int c = 0; c < num_cells; ++c)
     {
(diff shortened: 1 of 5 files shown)


cpp/dolfinx/fem/interpolate.h
```diff
@@ -15,6 +16,7 @@
 #include <variant>
 #include <xtensor/xadapt.hpp>
 #include <xtensor/xarray.hpp>
+#include <xtensor/xio.hpp>
```
jorgensd, line 19: I guess xio is not needed after debugging?

cpp/dolfinx/fem/interpolate.h
```diff
@@ -110,32 +104,97 @@ void interpolate_from_any(Function<T>& u, const Function<T>& v)
     throw std::runtime_error(
         "Interpolation on different meshes not supported (yet).");
   }
+
```
jorgensd, line 107:
```suggestion
```

cpp/dolfinx/fem/interpolate.h
```diff
@@ -110,32 +104,97 @@ void interpolate_from_any(Function<T>& u, const Function<T>& v)
...
-  std::vector<T>& coeffs = u.x()->mutable_array();
-
-  // Iterate over mesh and interpolate on each cell
-  const auto dofmap_u = u.function_space()->dofmap();
-  const std::vector<T>& v_array = v.x()->array();
-  const int num_cells = map->size_local() + map->num_ghosts();
-  const int bs = dofmap_v->bs();
-  assert(bs == dofmap_u->bs());
-  for (int c = 0; c < num_cells; ++c)
+  if (element_to->hash() == element_from->hash())
```
jorgensd, line 118: This should be simplified to a copy of array, instead of looping over all dofs, as hash0==has1 means same function space?

cpp/dolfinx/fem/interpolate.h
```diff
@@ -110,32 +104,97 @@ void interpolate_from_any(Function<T>& u, const Function<T>& v)
...
+
+    // Iterate over mesh and interpolate on each cell
+    const auto dofmap_u = u.function_space()->dofmap();
+    const auto dofmap_v = v.function_space()->dofmap();
+
+    const std::vector<T>& v_array = v.x()->array();
+    std::vector<T>& u_array = u.x()->mutable_array();
+
+    // Compute interpolation operator
+    xt::xtensor<double, 2> i_m
```
jorgensd, line 159: `i_m` could be const?

python/test/unit/function/test_interpolation.py
```diff
@@ -218,3 +219,19 @@ def test_mixed_interpolation():
     v = Function(FunctionSpace(mesh, ufl.MixedElement([A, B])))
     with pytest.raises(RuntimeError):
         v.interpolate(lambda x: (x[1], 2 * x[0], 3 * x[1]))
+
```
jorgensd, line 222: Maybe we should have the following tests:
- vector-DG <-> Nedelec
- vector-DG <-> RT
- CG -> DG
- Same space to same space (copy-assign)

cpp/dolfinx/fem/interpolate.h
IgorBaratta: the data can be strided for sub-elements.

PR #1743, opened 2021-10-06
Concatenating strings in a more portable way -- for some unclear reason

massimiliano-leoni:
Normal string concatenation was observed to fail on some machines for unclear reasons. This implementation seems to be more robust.


jhale:
This is unexpected! It would be nice to have a bit of clarity on why this happens before continuing though. What C++ standard library version are you building against?

garth-wells:
@massimiliano-leoni Does https://stackoverflow.com/a/59823692 help? (adding `s`)

PR #1744, opened 2021-10-07
Geometry with multiple maps

chrisrichardson:
Simple change to `mesh::Geometry` to accept multiple Coordinate Maps. 
Existing code often expects one map, so there are many places where we need to check and raise an error.


garth-wells:
Superseded by #2597.

PR #1782, opened 2021-10-28
Use interpreter variable passed to cmake to find python modules

renefritze:
This now matches the one used in installed checks in
[python/setup.py](https://github.com/FEniCS/dolfinx/blob/main/python/setup.py#L45)

Without this change cmake would fail for me to find petsc4py when I have a Python installation in /usr/local/

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -2,10 +2,10 @@ cmake_minimum_required(VERSION 3.16)
 
 PROJECT(dolfinx_pybind11)
 
+find_package(Python3 COMPONENTS Interpreter Development REQUIRED)
 find_package(pybind11 REQUIRED CONFIG HINTS ${PYBIND11_DIR} ${PYBIND11_ROOT}
   $ENV{PYBIND11_DIR} $ENV{PYBIND11_ROOT})
 
-find_package(Python3 COMPONENTS Interpreter REQUIRED)
 execute_process(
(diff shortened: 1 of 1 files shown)


python/CMakeLists.txt
```diff
@@ -2,10 +2,10 @@ cmake_minimum_required(VERSION 3.16)
 
 PROJECT(dolfinx_pybind11)
 
+find_package(Python COMPONENTS Interpreter Development REQUIRED)
```
garth-wells, line 5:
```suggestion
find_package(Python3 COMPONENTS Interpreter Development REQUIRED)
```
I think this should be `Python3` to avoid falling back to Python2 (which we don't support).

garth-wells (changes requested):
Thanks, just a small suggestion to avoid accidental fallback to Python2.

python/CMakeLists.txt
renefritze: Right, I've amended that. Thanks!

jhale:
@renefritze Could you try my branch https://github.com/FEniCS/dolfinx/pull/1815 on your machine?

renefritze:
> @renefritze Could you try my branch #1815 on your machine?

That seems unrelated. Do you mean https://github.com/FEniCS/basix/pull/343?

renefritze:
> It's tangentially related, I had to make some modifications to your PR to get what I wanted working, and I wanted to know if my new branch #1815 still works for you.

Yes, that still works for me. Thank you for being this considerate.

PR #1784, opened 2021-11-02
Add [[maybe_unused]] to variables unused in "Release" mode.

jorgensd:
Running Release mode with -Werror -Wall -pedantic shows quite a few possible unused variables, as they are only used in asserts. This commit fixes it.

### DIFF
--- a/cpp/dolfinx/io/HDF5Interface.h
+++ b/cpp/dolfinx/io/HDF5Interface.h
@@ -185,9 +185,6 @@ inline void HDF5Interface::write_dataset(
   // Dataset dimensions
   const std::vector<hsize_t> dimsf(global_size.begin(), global_size.end());
 
-  // Generic status report
-  herr_t status;
-
   // Create a global data space
   const hid_t filespace0 = H5Screate_simple(rank, dimsf.data(), nullptr);
   assert(filespace0 != HDF5_FAIL);
@@ -219,6 +216,9 @@ inline void HDF5Interface::write_dataset(
       = H5Dcreate2(file_handle, dataset_path.c_str(), h5type, filespace0,
                    H5P_DEFAULT, chunking_properties, H5P_DEFAULT);
   assert(dset_id != HDF5_FAIL);
+
+  // Generic status report
+  [[maybe_unused]] herr_t status;
 
   // Close global data space
   status = H5Sclose(filespace0);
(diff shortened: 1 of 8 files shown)


cpp/dolfinx/io/HDF5Interface.h
```diff
@@ -220,6 +217,9 @@ inline void HDF5Interface::write_dataset(
                    H5P_DEFAULT, chunking_properties, H5P_DEFAULT);
   assert(dset_id != HDF5_FAIL);
 
+  // Generic status report
+  [[maybe_unused]] herr_t status;
```
garth-wells, line 221: These should all be changed from assertions to exceptions.

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -118,7 +118,8 @@ XDMFFile::XDMFFile(MPI_Comm comm, const std::string filename,
   if (_file_mode == "r")
   {
     // Load XML doc from file
-    pugi::xml_parse_result result = _xml_doc->load_file(_filename.c_str());
+    [[maybe_unused]] pugi::xml_parse_result result
```
garth-wells, line 121: Change to exceptions.

cpp/dolfinx/la/PETScMatrix.cpp
```diff
@@ -190,8 +191,9 @@ PETScMatrix::set_fn(Mat A, InsertMode mode)
 {
   return [A, mode, cache = std::vector<PetscInt>()](
              std::int32_t m, const std::int32_t* rows, std::int32_t n,
-             const std::int32_t* cols, const PetscScalar* vals) mutable {
-    PetscErrorCode ierr;
+             const std::int32_t* cols, const PetscScalar* vals) mutable
+  {
+    [[maybe_unused]] PetscErrorCode ierr;
```
garth-wells, line 196: Change to exceptions.

garth-wells (changes requested):
A few of the assertion should be changed to exceptions, comments inline.

cpp/dolfinx/la/PETScMatrix.cpp
jorgensd: @garth-wells  I've now changed quite a few of them. Do you want any more of them to be changed?

cpp/dolfinx/la/PETScMatrix.cpp
```diff
@@ -201,12 +201,7 @@ PETScMatrix::set_fn(Mat A, InsertMode mode)
 #else
     ierr = MatSetValuesLocal(A, m, rows, n, cols, vals, mode);
 #endif
-
-#ifdef DEBUG
```
michalhabera, line 205: I found these useful in cases where `MatSetValuesLocal` went silently wrong in Release mode.

garth-wells: I think they should be caught at the higher level by the caller.

The original PR turned on the check in all modes, which isn't appropriate.

PR #1798, opened 2021-11-11
Interpolation for elements with different Piola maps

### LINKED ISSUE #1747: Support interpolation operator between elements with different geometric maps
Recently added interpolation operator support works only between elements that use the same geometric map type. It would be helpful to support interpolations between elements with different maps.

mscroggs:
Implemented interpolation for element with different map types.
Added more tests for interpolation between elements.

Fixes #1747.


cpp/dolfinx/fem/interpolate.h
```diff
@@ -396,6 +397,111 @@ void interpolate(Function<T>& u, const Function<T>& v)
...
+      cell[0] = c;
+      for (std::size_t i = 0; i < cells.size(); ++i)
+        cells[i] = c;
+      const xt::xtensor<double, 2> X
+          = interpolation_coords(*element_to, *mesh, cell);
+      xt::xtensor<T, 2> _v({v_values.shape(0), v_values.shape(2)});
+      v.eval(xt::transpose(X), cells, _v);
+      for (std::size_t i = 0; i < v_values.shape(0); ++i)
+        for (std::size_t j = 0; j < v_values.shape(2); ++j)
+          v_values(i, 0, j) = _v(i, j);
```
mscroggs, line 475: I can't get `xt::view` to correctly pass a reshaped version of `v_values` into `v.eval`.

Either this should be done, or assigning memory for `_v` should move out of loop.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -396,6 +397,111 @@ void interpolate(Function<T>& u, const Function<T>& v)
...
+      for (std::size_t i = 0; i < num_dofs_g; ++i)
+        for (std::size_t j = 0; j < gdim; ++j)
+          coordinate_dofs(i, j) = x_g(x_dofs[i], j);
+
+      // Evaluate v at interpolation points
+      cell[0] = c;
+      for (std::size_t i = 0; i < cells.size(); ++i)
+        cells[i] = c;
+      const xt::xtensor<double, 2> X
+          = interpolation_coords(*element_to, *mesh, cell);
```
mscroggs, line 470: Could we push points to the cell instead of using `interpolation_coords` here so that we don't need to transpose later?

mscroggs: This has been resolved by adding a `transpose` argument to `interpolation_coords`.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -396,6 +399,104 @@ void interpolate(Function<T>& u, const Function<T>& v)
...
+        cmap.compute_jacobian(dphi, coordinate_dofs, _J);
+        cmap.compute_jacobian_inverse(_J, xt::view(K, p, xt::all(), xt::all()));
+        detJ[p] = cmap.compute_jacobian_determinant(_J);
+      }
+
+      // Pull back to the reference and interpolate
+      element_to->map_pull_back(v_values, J, detJ, K, v_mapped_values);
+
+      xt::xtensor<T, 2> _vm
+          = xt::transpose(xt::view(v_mapped_values, xt::all(), 0, xt::all()));
```
mscroggs, line 488: This is probably doing a copy. Is there a way to avoid it?

cpp/dolfinx/fem/interpolate.h
IgorBaratta: The issue is that `eval` takes an `xt::xtensor<T, 2>&`, which has data ownership.
We can template eval, or use `xt::xexpression`. I'd suggest doing the former.

Maybe add a note so we can fix it in a new PR?

jorgensd: We will not use eval after all, as it does an unnecessary pull back, as we already know the points in the reference element. So I guess we should just create a separate issue

mscroggs: eval is gone

PR #1800, opened 2021-11-14
Pack coefficients only for the active entities for each kernel

jpdean:
This PR modifies `pack_coefficients` so that coefficients are packed separately for each kernel in a `Form` and only for the active entities of that kernel. This is in contrast to the implementation in main, where all coefficients in a `Form` are packed for every cell in the mesh.

With this PR, given a form `a`, calling `pack_coefficients(a)` from python returns a dictionary with keys `(integral_type, domain_id)`. The value for each key is an array of shape `(num_active_entities, cstride)` i.e. the coefficients for each active entity for that kernel.

There are a few advantages to this approach, the main being that it is a step towards supporting assembly of forms with coefficients that are defined on a submesh (see https://github.com/FEniCS/dolfinx/projects/9#card-67987072). Secondly, in some common use cases, significantly fewer coefficients need to be packed with this implementation, for instance, in the case of an exterior facet integral over the domain boundary or a cell integral over a small subset of cells. Thirdly, it is a step towards packing coefficients at quadrature points.

Unfortunately, there are some disadvantages to this approach too, where more data is stored compared to main. This would be the case, for instance, for forms containing coefficients that appear in both a cell integral over the whole domain and a facet integral, or in the case of interior facet integrals.

No modifications to FFCx have been made yet, meaning the generated kernels expect all coefficients in the form to be packed, even if they are not used in that particular kernel. The next step would be to modify this behavior so that only coefficients needed by the kernel are packed.

Any feedback would be very much appreciated!


cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -72,8 +73,12 @@ void assemble_cells(
   std::vector<T> Ae(ndim0 * ndim1);
   const xtl::span<T> _Ae(Ae);
   std::vector<double> coordinate_dofs(3 * num_dofs_g);
-  for (std::int32_t c : cells)
+
+  // Iterate over active cells
+  for (std::size_t index = 0; index < cells.size(); ++index)
   {
+    auto c = cells[index];
```
chrisrichardson, line 80: Do we need auto here? Maybe std::int32_t is better.

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -41,8 +41,10 @@ T assemble_cells(const mesh::Geometry& geometry,
 
   // Iterate over all cells
   T value(0);
-  for (std::int32_t c : cells)
+  for (std::size_t index = 0; index < cells.size(); ++index)
   {
+    auto c = cells[index];
```
chrisrichardson, line 46: `std::int32_t`

cpp/dolfinx/fem/utils.h
```diff
@@ -437,53 +438,111 @@ fem::FunctionSpace create_functionspace(
...
+    std::int32_t offset, int space_dim,
+    const std::function<void(const xtl::span<T>&,
+                             const xtl::span<const std::uint32_t>&,
+                             std::int32_t, int)>& transformation)
 {
-  using T = typename U::scalar_type;
+  const int bs = dofmap.bs();
+  for (std::size_t index = 0; index < active_facets.size(); ++index)
+  {
+    auto cell = active_facets[index].first;
```
chrisrichardson, line 500: `std::int32_t`

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -33,8 +33,9 @@ void assemble_matrix(
     const std::function<int(std::int32_t, const std::int32_t*, std::int32_t,
                             const std::int32_t*, const T*)>& mat_set_values,
     const Form<T>& a, const xtl::span<const T>& constants,
-    const xtl::span<const T>& coeffs, int cstride, const std::vector<bool>& bc0,
-    const std::vector<bool>& bc1);
+    const std::map<std::pair<IntegralType, int>,
+                   std::pair<std::vector<T>, int>>& coefficients,
```
garth-wells, line 37: I think this will need to use a `span` rather than `std::vector` to avoid forcing the storage type (which would be problematic from Python).

jhale:
Tangential but perhaps related point: It would also be good to think about whether this implementation extends nicely to the use case of having memory-intensive coefficients, e.g. a tensor-valued coefficient on a high-order quadrature space on the entire domain. In that setting it would be advantageous to batch coefficient computation across a user-defined number of active cells prior to assembling the FE kernel contributions.

IgorBaratta:
> Tangential but perhaps related point: It would also be good to think about whether this implementation extends nicely to the use case of having memory-intensive coefficients, e.g. a tensor-valued coefficient on a high-order quadrature space on the entire domain. In that setting it would be advantageous to batch coefficient computation across a user-defined number of active cells prior to assembling the kernel contributions.

Good point. There are some trade-offs to think about.
I think this PR's approach is potentially beneficial for memory-intensive coefficients. Since we are packing only on active cells for a given domain, and domains do not overlap (currently), there is little redundancy.  This would not necessarily be true if we grouped kernels by the quadrature rule. 

Packing coefficients on quadrature points is a bit more complicated since it would involve analysing the form (grad(u) vs u). It's not clear to me if it's advantageous for CPUs (but it's definitely beneficial for GPUs).

michalhabera:
Just note there is no guarantee that domains do not overlap. You can have `dx(1)` marking entire mesh and `dx(2)` marking part of it. Then `...*dx(1) + ...*dx(2)` would do some extra packing in your case on the intersection.

Packing coefficients per integral sounds as good idea, but I would recommend to primarily focus on `enabled_coefficients` member of `ufc_integral`. It tells you which coefficients in the form are used in this integral kernel (I just do not remember if it uses original coefficient positions or position of coefficients after preprocessing). This leads even more speedup in forms `...*dx + ...*ds` since it avoids packing cell coefficients for boundary facets and vice versa.

I would also suggest to have a low-level interface `pack_coefficients(std::vector<int>& active_entities, std::vector<Function&> enabled_coefficients)` which is then used in high level interfaces as you've added. This is more in the dolfinx spirit, as you could use this low-level interface for your custom assemblers and kernels - which do not have any Form associated.

jpdean:
> Just note there is no guarantee that domains do not overlap. You can have `dx(1)` marking entire mesh and `dx(2)` marking part of it. Then `...*dx(1) + ...*dx(2)` would do some extra packing in your case on the intersection.
> 
> Packing coefficients per integral sounds as good idea, but I would recommend to primarily focus on `enabled_coefficients` member of `ufc_integral`. It tells you which coefficients in the form are used in this integral kernel (I just do not remember if it uses original coefficient positions or position of coefficients after preprocessing). This leads even more speedup in forms `...*dx + ...*ds` since it avoids packing cell coefficients for boundary facets and vice versa.
> 
> I would also suggest to have a low-level interface `pack_coefficients(std::vector<int>& active_entities, std::vector<Function&> enabled_coefficients)` which is then used in high level interfaces as you've added. This is more in the dolfinx spirit, as you could use this low-level interface for your custom assemblers and kernels - which do not have any Form associated.

Thanks for the feedback Michal. I hadn't thought about the case of overlapping domains; there are definitely some pros and cons to this approach of packing coefficients. I was planning on looking at `enabled_coefficients` as a next step (i.e. in another PR), as I didn't want this PR to get too large. I will also look into adding a lower level interface as you have suggested.

garth-wells (approved):

PR #1802, opened 2021-11-15
Use standard terminology for pull back/push forward

garth-wells:


mscroggs (approved):
I'm very happy with this change, as the old names were confusing me on Friday

PR #1803, opened 2021-11-15
Avoid dividing by zero in NewtonSolver

massimiliano-leoni:
At `NewtonSolver.cpp:31` there is a line that sometimes performs a division by zero. Although this shouldn't be a concern, it just happened to me that a Newton solver object was not working properly without this edit, in the sense that the convergence criterion was satisfied at the 0-th iteration.


massimiliano-leoni:
Never mind, I was completely mislead by a different bug. Closing.

PR #1805, opened 2021-11-16
Progress towards removing import of `cpp` in user and test code

garth-wells:
- Import `cpp` as `_cpp` to avoid polluting namespaces
- Make MPI comm a property of mesh
- Introduce a Python Mesh that inherits from the C++ version
- Add some Python factory functions to avoid direct imports on wrapped C++ objects
- Add imported wrapped C++ classes to Python API docs where appropriate

There are some cases that can be looked at in more details later (particularly objects that are templated over a type)


python/demo/mixed-elasticity-sc/static-condensation-elasticity.py
```diff
@@ -60,19 +61,19 @@ def left(x):
 
 # Locate all facets at the free end and assign them value 1
 free_end_facets = locate_entities_boundary(mesh, 1, free_end)
-mt = dolfinx.mesh.MeshTags(mesh, 1, free_end_facets, 1)
```
jorgensd, line 63: Should we use `mesh.topology.dim-1` here instead of 1?
I’ve found that most people on discourse does not understand the idea of the dimension of the entity.

python/demo/stokes-taylor-hood/demo_stokes-taylor-hood.py
```diff
@@ -427,7 +429,7 @@ def lid_velocity_expression(x):
 (u, p) = ufl.TrialFunctions(W)
 (v, q) = ufl.TestFunctions(W)
 f = Function(W0)
-zero = dolfinx.Constant(mesh, 0.0)
+zero = Constant(mesh, 0.0)
```
jorgensd, line 432: This zero should probably be wrapped with a `PETSc.ScalarType`

python/dolfinx/geometry.py
```diff
@@ -4,6 +4,28 @@
...
+from dolfinx import cpp as _cpp
+from dolfinx.cpp.geometry import (compute_closest_entity, compute_collisions,
+                                  compute_distance_gjk, create_midpoint_tree)
+
+__all__ = ["compute_colliding_cells", "squared_distance", "compute_closest_entity", "compute_collisions",
+           "compute_distance_gjk", "create_midpoint_tree"]
+
+
+class BoundingBoxTree(_cpp.geometry.BoundingBoxTree):
+    def __init__(self, mesh, dim, entities=None, padding=0.0):
```
jorgensd, line 16: Do you want me to add some documentation to the functions/classes in this file?

python/dolfinx/mesh.py
```diff
@@ -7,23 +7,48 @@
...
 
 
-def locate_entities(mesh: cpp.mesh.Mesh,
-                    dim: int,
-                    marker: types.FunctionType):
+class Mesh(_cpp.mesh.Mesh):
+    """A class for representing meshes. Mesh objects should not be
+    created using this class directly."""
+
+    def __init__(self, comm, topology, geometry, domain):
```
jorgensd, line 25: Documentation of input types would be nice here.

python/dolfinx/mesh.py
```diff
@@ -80,89 +102,64 @@ def locate_entities_boundary(mesh: cpp.mesh.Mesh,
...
-    return mesh
+    return Mesh.from_cpp(mesh, domain)
 
 
-def MeshTags(mesh, dim, indices, values):
+def MeshTags(mesh: Mesh, dim: int, indices, values):
 
     if isinstance(values, int):
-        values = numpy.full(indices.shape, values, dtype=numpy.int32)
+        values = np.full(indices.shape, values, dtype=np.int32)
```
jhale, line 156: This cast is not safe according to numpy:

`np.can_cast(int, np.int32)`

Should avoid doing hidden unsafe casts.

jhale: Cast from int to np.int64 is safe, possible alternative?

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -268,8 +268,8 @@ void mesh(py::module& m)
       .def_property_readonly("cell_type", &dolfinx::mesh::Topology::cell_type)
       .def("cell_name", [](const dolfinx::mesh::Topology& self)
            { return dolfinx::mesh::to_string(self.cell_type()); })
-      .def("mpi_comm", [](dolfinx::mesh::Mesh& self)
-           { return MPICommWrapper(self.mpi_comm()); });
+      .def_property_readonly("mpi_comm", [](dolfinx::mesh::Mesh& self)
```
jhale, line 271: Is this near zero-cost? It's been a function for a long time.

python/test/unit/fem/test_element_integrals.py
```diff
@@ -8,11 +8,13 @@
 import random
 from itertools import combinations, product
 
+import dolfinx
 import numpy as np
 import pytest
 import ufl
-from dolfinx import (Constant, Function, FunctionSpace, VectorFunctionSpace,
-                     cpp, fem)
+from dolfinx.fem import (Constant, Function, FunctionSpace,
```
jhale, line 15: General comment: Are `assemble_*` used frequently enough that they should be in the root namespace? I do not really see why eg. `Form`, `Function` and `UnitSquareMesh` are in the root namespace, but `assemble_*` are not.

jhale: Equally, perhaps some things are currently not justified in the root namespace.

jhale (changes requested):

python/dolfinx/mesh.py
garth-wells: This PR didn't change the details of this code. Could you create a separate PR if it needs to be changed? There is already a lot going on in this PR.

python/dolfinx/wrappers/mesh.cpp
garth-wells: It should just be some messing about with pointers.

Everything was a function in the first version of the pybind11 wrappers. We should, at some point, check the whole interface for what should be a property.

python/test/unit/fem/test_element_integrals.py
garth-wells: `Form`, `Function`, etc, have been removed from the root namespace in this PR. They were in the root namespace for legacy reasons and just never got moved out.

python/dolfinx/mesh.py
jhale: Will do.

jhale (approved):

python/demo/pyvista/demo_pyvista.py
```diff
@@ -176,8 +173,8 @@ def in_circle(x):
 # Create a dolfinx.MeshTag for all cells. If midpoint is inside the
 # circle, it gets value 1, otherwise 0.
 num_cells = mesh.topology.index_map(mesh.topology.dim).size_local
-midpoints = dolfinx.mesh.midpoints(mesh, mesh.topology.dim, list(np.arange(num_cells, dtype=np.int32)))
-cell_tags = dolfinx.MeshTags(mesh, mesh.topology.dim, np.arange(num_cells), in_circle(midpoints))
+midpoints = midpoints(mesh, mesh.topology.dim, list(np.arange(num_cells, dtype=np.int32)))
```
michalhabera, line 176: This overwrites the `dolfinx.mesh.midpoints` placeholder.

python/test/unit/graph/test_adjacencylist.py
```diff
@@ -0,0 +1,12 @@
+
```
michalhabera, line 1: Add license header.

michalhabera (approved):

python/demo/pyvista/demo_pyvista.py
garth-wells: Good point. I wanted to change the function name because it should be a verb.

garth-wells: `mesh.midpoints` renamed to `mesh.compute_midpoints`

python/dolfinx/mesh.py
```diff
@@ -80,89 +102,109 @@ def locate_entities_boundary(mesh: cpp.mesh.Mesh,
...
-    """Refine a mesh"""
+def refine(mesh: Mesh, cell_markers: _cpp.mesh.MeshTags_int8 = None, redistribute: bool = True):
+    """Refine a mesh
+
+    Parameters
+    ----------
+    mesh
+        The mesh from which to build a refined mesh
+    cell_markers
+        Optional argument to specify which cells should be refined. If not supplied uniform refinement.
```
chrisrichardson, line 132: This should be "entity markers" not "cell markers". The fundamental operation of refinement is on edges, so if you supply cells, it will refine all the edges of the cell - but you can also just supply the edges.

jorgensd: I'll fix that when merging in: https://github.com/FEniCS/dolfinx/issues/1834 afterwards

chrisrichardson (approved):

jorgensd (approved):

PR #1807, opened 2021-11-16
Fix places where data is only used in assertions, causing -Wall -Werr -pedantic to fail in Release mode

jorgensd:


garth-wells:
I think we need to add a section to the CI to test for this.

garth-wells (approved):

PR #1815, opened 2021-11-18
Build Linux x86_64 binary wheels

jhale:


python/CMakeLists.txt
```diff
@@ -2,10 +2,10 @@ cmake_minimum_required(VERSION 3.16)
 
 PROJECT(dolfinx_pybind11)
 
-find_package(Python3 COMPONENTS Interpreter Development REQUIRED)
```
garth-wells, line 5: Is there a reason for undoing the earlier PR? It seems it followed the recommended practice at https://pybind11.readthedocs.io/en/stable/cmake/index.html#new-findpython-mode.

jhale: https://github.com/pybind/pybind11/pull/2689/files

Recently updated documentation re: development.

However, it still fails if I have find python before find pybind11. Not clear why at the moment, still investigating.

jhale: This is fixed now. The earlier PR was on the right track.

In short:

- Development.Module is enough for our purposes (we do not embed the Python interpreter).
- More hints needed to be passed from `setup.py` for Development to work robustly in e.g. virtual environments.
- `pybind11_add_module` properly handles adding the necessary target_link_library calls to `cpp`, no extra call is necessary.
- `pybind11_add_module` can suffice with `MODULE` rather than `SHARED` (the default infact, but I've left it for clarity).

PR #1822, opened 2021-11-21
Get dolfinx_jit_parameters.json location from envvar

matsievskiysv:
When running systemd service it's customary run it as a system user without home directory. In such a setup default JIT cache directory (`~/.cache/fenics`) cannot be created and user configuration file (`~/.config/dolfinx/dolfinx_jit_parameters.json`) cannot be set. This PR adds ability to override user configuration file location by `DOLFINX_JIT_PARAMETERS_FILE` environment variable.

### DIFF
--- a/python/dolfinx/jit.py
+++ b/python/dolfinx/jit.py
@@ -22,7 +22,7 @@ __all__ = ["ffcx_jit", "get_parameters"]
 DOLFINX_DEFAULT_JIT_PARAMETERS = {
     "cache_dir":
         (os.getenv("XDG_CACHE_HOME", default=Path.home().joinpath(".cache")) / Path("fenics")
-         "Path for storing DOLFINx JIT cache. " + 
+         "Path for storing DOLFINx JIT cache. " + \
          "Default prefix ~/.cache/ can be changed using XDG_CACHE_HOME environment variable."),
     "cffi_debug":
         (False, "CFFI debug mode"),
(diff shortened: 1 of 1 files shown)


jhale:
This is a fine idea. Rather then reinventing the wheel though, I would rather we support the $XDG_CONFIG_HOME environment variable according to the XDG config specification.

https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html

matsievskiysv:
I've updated to code to use `XDG_CACHE_HOME` and `XDG_CONFIG_HOME`. I guess, [ffcx/parameters.py](https://github.com/FEniCS/ffcx/blob/main/ffcx/parameters.py) needs to be updated as well.

python/dolfinx/jit.py
```diff
@@ -21,7 +21,9 @@
 
 DOLFINX_DEFAULT_JIT_PARAMETERS = {
     "cache_dir":
-        (Path.joinpath(Path.home(), ".cache", "fenics"), "Path for storing DOLFINx JIT cache"),
+        (os.getenv("XDG_CACHE_HOME", default=Path.home().joinpath(".cache")) / Path("fenics"),
+         "Path for storing DOLFINx JIT cache."
+         + "Default prefix ~/.cache/ can be changed using XDG_CACHE_HOME environment variable."),
```
matsievskiysv, line 26: `+` is not necessary

```python3
a = (1, "hello" "world")
print(a)
```
```
(1, 'helloworld')
```

jhale: Thanks, flake8 had me running round in circles.

matsievskiysv: +)

PR #1826, opened 2021-11-23
Remedy errant boost behaviour

chrisrichardson:
* Ensure that the filename string does not contain any `\0` characters by going through `.c_str()` from boost


michalhabera:
This is the solution for, https://github.com/FEniCS/dolfinx/pull/1743, right?

PR #1830, opened 2021-11-24
Simplify MeshTag construction

garth-wells:
Use a 2D array rather than AdjacencyList in the constructor.


jorgensd:
@garth-wells This simplification does not support MeshTags facets on prisms, or MeshTags on mixed-cell type meshes?

garth-wells:
> @garth-wells This simplification does not support MeshTags facets on prisms, or MeshTags on mixed-cell type meshes?

We don't support mixed-meshes yet. The storage layout for mixed meshes is still be determined. This function could use an `AdjacencyList` or it could use a collection of 2D arrays (one for each entity type).

PR #1831, opened 2021-11-25
Simplify packing coefficients

SarahRo:
- replace pack_coefficient_cell, pack_coefficient_exterior_facet and pack_coefficient_interior_facet by one function pack_coefficient_entity
- add test with non-constant coefficients
- simplify pack_coefficients

PR #1832, opened 2021-11-25
Simplify map construction when packing coefficients

garth-wells:


jorgensd:
Should this be merged with #1831 ?

garth-wells:
Change made in #1831.

PR #1835, opened 2021-11-26
Update refinement docs

### LINKED ISSUE #1834: Unclear behavior of mesh refinement
According to the documentation, we should only refine cells marked with value==1 if a meshtag is supplied:
https://github.com/FEniCS/dolfinx/blob/3f8fa273315b904a9f394ea3d7072e7b940997fd/cpp/dolfinx/refinement/refine.h#L36-L38

However, the code does not check these values:
https://github.com/FEniCS/dolfinx/blob/3f8fa273315b904a9f394ea3d7072e7b940997fd/cpp/dolfinx/refinement/plaza.h#L54-L61
MWE:
```python
import dolfinx.io
import dolfinx
from mpi4py import MPI
import numpy as np

mesh = dolfinx.UnitSquareMesh(MPI.COMM_WORLD, 10, 10)
cell_map = mesh.topology.index_map(mesh.topology.dim)
num_cells = cell_map.size_local + cell_map.num_ghosts

refine_cells = np.array(
    [num_cells//2, num_cells//3, num_cells/4], dtype=np.int32)
values = np.arange(len(refine_cells), dtype=np.int8)
marker = dolfinx.MeshTags(mesh, mesh.topology.dim, refine_cells, values)
mesh.topology.create_entities(mesh.topology.dim-2)
mesh.topology.create_entities(mesh.topology.dim-1)
new_mesh = dolfinx.mesh.refine(mesh, marker)


with dolfinx.io.XDMFFile(mesh.mpi_comm(), "mesh.xdmf", "w") as xdmf:
    xdmf.write_mesh(new_mesh)
```

jorgensd:
Resolves #1834

### DIFF
--- a/cpp/dolfinx/mesh/cell_types.cpp
+++ b/cpp/dolfinx/mesh/cell_types.cpp
@@ -17,102 +17,98 @@
 using namespace dolfinx;
 
 //-----------------------------------------------------------------------------
-std::string mesh::to_string(mesh::CellType type)
+std::string mesh::to_string(CellType type)
 {
   switch (type)
   {
-  case mesh::CellType::point:
+  case CellType::point:
     return "point";
-  case mesh::CellType::interval:
+  case CellType::interval:
     return "interval";
-  case mesh::CellType::triangle:
+  case CellType::triangle:
     return "triangle";
-  case mesh::CellType::tetrahedron:
+  case CellType::tetrahedron:
     return "tetrahedron";
-  case mesh::CellType::quadrilateral:
+  case CellType::quadrilateral:
     return "quadrilateral";
-  case mesh::CellType::pyramid:
+  case CellType::pyramid:
     return "pyramid";
-  case mesh::CellType::prism:
+  case CellType::prism:
     return "prism";
-  case mesh::CellType::hexahedron:
+  case CellType::hexahedron:
     return "hexahedron";
   default:
     throw std::runtime_error("Unknown cell type.");
-    return std::string();
   }
 }
 //-----------------------------------------------------------------------------
 mesh::CellType mesh::to_type(const std::string& cell)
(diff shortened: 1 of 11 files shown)


chrisrichardson:
Actually, the wording still leaves something rto be desired "entity indices to be split". In fact, it is the edges incident to the entities referenced by the indices that will be split (obviously this is simplest when tdim=1, edges).

jhale:
I don't think these routines support directly marking edges for refinement, so the entities always being cells is appropriate I think?

jorgensd:
> I don't think these routines support directly marking edges for refinement, so the entities always being cells is appropriate I think?

According to the source code https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/refinement/plaza.cpp#L620-L634 
it does, as it maps everything down to the edges

chrisrichardson:
Yes, that's what I was trying to say.

jhale:
My mistake, thanks for the correction!

garth-wells:
I found this confusing, so changed it to take the list of edges to be split, and added a utility function to compute the edges incident to a list of entities of some some other dimension.

jorgensd:
> I found this confusing, so changed it to take the list of edges to be split, and added a utility function to compute the edges incident to a list of entities of some some other dimension.

Fine by me, but should we add a test/demo to illustrate the usage?

garth-wells:
> > I found this confusing, so changed it to take the list of edges to be split, and added a utility function to compute the edges incident to a list of entities of some some other dimension.
> 
> Fine by me, but should we add a test/demo to illustrate the usage?

Seems the marked refinement is completely untested. I's suggests tests in a separate PR.

garth-wells (approved):

PR #1836, opened 2021-11-26
Add diagonal enum for RectangleMesh

### LINKED ISSUE #1833: Change diagonal argument to enum in mesh construction
We should always avoid the usage of strings. used in `RectangleMesh`
https://github.com/FEniCS/dolfinx/blob/c30adac64e5f5a19e7909666d18a8d6d8d38ebfb/cpp/dolfinx/generation/RectangleMesh.h#L34-L37

jorgensd:
Fixes #1833.

### DIFF
--- a/cpp/dolfinx/generation/RectangleMesh.cpp
+++ b/cpp/dolfinx/generation/RectangleMesh.cpp
@@ -8,11 +8,10 @@
 #include <cfloat>
 #include <dolfinx/common/MPI.h>
 #include <dolfinx/graph/AdjacencyList.h>
+#include <dolfinx/mesh/utils.h>
 #include <xtensor/xfixed.hpp>
 #include <xtensor/xtensor.hpp>
 #include <xtensor/xview.hpp>
-#include <dolfinx/mesh/utils.h>
-
 
 using namespace dolfinx;
 using namespace dolfinx::generation;
@@ -25,7 +24,7 @@ mesh::Mesh build_tri(MPI_Comm comm,
                      std::array<std::size_t, 2> n,
                      const mesh::GhostMode ghost_mode,
                      const mesh::CellPartitionFunction& partitioner,
-                     const std::string& diagonal)
+                     const generation::DiagonalType diagonal)
 {
   fem::CoordinateElement element(mesh::CellType::triangle, 1);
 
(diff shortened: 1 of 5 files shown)


cpp/dolfinx/generation/RectangleMesh.cpp
```diff
@@ -25,7 +24,7 @@ mesh::Mesh build_tri(MPI_Comm comm,
                      std::array<std::size_t, 2> n,
                      const mesh::GhostMode ghost_mode,
                      const mesh::CellPartitionFunction& partitioner,
-                     const std::string& diagonal)
+                     const generation::DiagonalType diagonal)
```
garth-wells, line 27: In the DOLFINx style we don't normally use `const` when passing enums.

cpp/dolfinx/generation/RectangleMesh.cpp
```diff
@@ -82,7 +74,7 @@ mesh::Mesh build_tri(MPI_Comm comm,
 
   // Create vertices and cells
   std::size_t nv, nc;
-  if (diagonal == "crossed")
+  if (diagonal == DiagonalType::crossed)
```
garth-wells, line 77: The `if`s can be changed to switches now that we have enums.

cpp/dolfinx/generation/RectangleMesh.h
```diff
@@ -12,6 +12,21 @@
 #include <mpi.h>
 #include <string>
 
+namespace dolfinx::generation
+{
+
+/// Enum for different diagonal types
+enum class DiagonalType : int
```
garth-wells, line 19: Is `int` required here?

garth-wells (changes requested):

cpp/dolfinx/generation/RectangleMesh.cpp
jorgensd: We use it for GhostMode in this function. Do you want me to change that as well?

cpp/dolfinx/generation/RectangleMesh.h
jorgensd: I could not get the code to compile without int

garth-wells (approved):

PR #1840, opened 2021-11-26
Assert cast from specific Python int to int32 is safe

jhale:


garth-wells:
Do we really need this given that we use `py::array<T>` in the pybind11 layer? I thought that this raises an exception if there is a type problem?

jhale:
Yes. MWE, nothing that pybind11 could fix:

```
import numpy as np
x = 2147483647 + 1 # no overflow
y = np.int32(x)
print(x)
print(y) # overflow int32
print(np.can_cast(x, np.int32)) # false
```

PR #1846, opened 2021-11-28
Re-enable pybind11/eigen test.

### LINKED ISSUE #1757: Remove possibly untested test
As we no longer depend on eigen, I do not think the test:
https://github.com/FEniCS/dolfinx/blob/main/python/test/unit/fem/test_assemble_cppimport.py#L26
is ever checked for correctness. 

We could either:
1. Remove it
2. Add eigen to testing suite

garth-wells:
Fixes #1757.


jhale (commented):
Would be good to put the libeigen-dev into the Dockerfile under the 'required for tests' section.

garth-wells:
> Would be good to put the libeigen-dev into the Dockerfile under the 'required for tests' section.

I considered that, but didn't want to bloat to dev image.

jhale:
It's 3500kb installed so I wouldn't worry about it too much.

PR #1858, opened 2021-12-01
Update Expression

michalhabera:
Along with https://github.com/FEniCS/ffcx/pull/432

This PR allows Expression contain Argument, which makes possible to build global interpolation operators, see added rank-1 test.


jorgensd:
I think you need to add something along the lines of: 
https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/assemble_vector_impl.h#L542
to https://github.com/FEniCS/dolfinx/blob/c3497c8de348f528aeb520d3dcb51f8b325a14a1/cpp/dolfinx/fem/Expression.h#L140

cpp/dolfinx/fem/Expression.h
```diff
@@ -86,23 +92,18 @@ class Expression
   /// Evaluate the expression on cells
   /// @param[in] active_cells Cells on which to evaluate the Expression
   /// @param[out] values A 2D array to store the result. Caller
-  /// responsible for correct sizing which should be (num_cells,
-  /// num_points * value_size columns).
+  /// is responsible for correct sizing which should be (num_cells,
+  /// num_points * value_size * num_all_argument_dofs columns).
```
jhale, line 96: I don't understand what `num_all_argument_dofs` is.

Wondering if this is the best layout now that it's a 4D tensor. Would it be better to pass e.g. a span of the correct size and document the layout when reinterpreted into a 4D tensor?

cpp/dolfinx/fem/Expression.h
```diff
@@ -114,8 +115,13 @@ class Expression
     // Create data structures used in evaluation
     std::vector<double> coordinate_dofs(3 * num_dofs_g);
 
+    const int num_all_argument_dofs
```
jhale, line 118: If the caller needs this to interpret the output of the function then it should go in a public member function?

python/dolfinx/fem/function.py
```diff
@@ -104,67 +106,56 @@ def __init__(self,
...
+        coeffs = [original_coefficients[self._ufc_expression.original_coefficient_positions[i]]._cpp_object
+                  for i in range(self._ufc_expression.num_coefficients)]
 
         ufl_constants = ufl.algorithms.analysis.extract_constants(ufl_expression)
-        constants = [ufl_constant._cpp_object for ufl_constant in ufl_constants]
+        constants = [constant._cpp_object for i, constant in enumerate(ufl_constants)]
 
-        # Getcpp Expression type
-        def expressiontype(dtype):
+        def create_expression(dtype):
```
jhale, line 119: It would be nice to support float32, useful for GPU-based computations in TF/JAX etc.

python/dolfinx/fem/function.py
```diff
@@ -104,67 +106,56 @@ def __init__(self,
...
+                               self.num_points * self.value_size * num_all_argument_dofs), dtype=self.dtype)
+            self._cpp_object.eval(cells, values)
         else:
-            assert u.ndim < 3
-            assert u.size == num_cells * self.num_points * self.value_size
-            assert u.shape[0] == num_cells
-            assert u.shape[1] == self.num_points * self.value_size
-            self._cpp_object.eval(cells, u)
+            if values.ndim >= 3 or values.shape[0] != _cells.shape[0] or values.shape[1] != _cells.shape[1]:
+                raise TypeError("Provided vaules array has incorrect shape.")
```
jhale, line 155: What would be the most convenient shape for the caller across the Python interface?

python/dolfinx/wrappers/fem.cpp
```diff
@@ -1023,5 +1035,74 @@ void fem(py::module& m)
       .def("tabulate_dof_coordinates",
            [](const dolfinx::fem::FunctionSpace& self)
            { return xt_as_pyarray(self.tabulate_dof_coordinates(false)); });
+
+  // template <typename T>
```
jhale, line 1039: Delete?

jhale (changes requested):
Still need to add the stress computational example to the `demo_elasticity.py`. Very busy this week, I might not get round to it till the back end of next week.

jorgensd:
I guess we should have a demo illustrating how to use this functionality from the C++ layer, with an ufl file (an extended version of what is in the ffcx tests).

cpp/dolfinx/fem/Expression.h
```diff
@@ -156,34 +166,60 @@ class Expression
...
 
   // Function to evaluate the Expression
-  std::function<void(T*, const T*, const T*, const double*)> _fn;
+  std::function<void(T*, const T*, const T*, const double*, const int*,
+                     const uint8_t*)>
+      _fn;
 
   // Evaluation points on reference cell
-  xt::xtensor<double, 2> _x;
+  xt::xtensor<double, 2> _X;
```
garth-wells, line 213: C++ peculiarity - underscore followed by a capital is not valid C++ (https://en.cppreference.com/w/cpp/language/identifiers).

python/dolfinx/fem/function.py
```diff
@@ -9,8 +9,7 @@
 
 import typing
 
-if typing.TYPE_CHECKING:
-    from dolfinx.mesh import Mesh
+from dolfinx.mesh import Mesh
```
jhale, line 12: Docs don't build with this if statement

python/dolfinx/wrappers/fem.cpp
```diff
@@ -453,68 +453,68 @@ void declare_objects(py::module& m, const std::string& type)
...
+      .def_property_readonly("mesh", &dolfinx::fem::Expression<T>::mesh)
+      .def_property_readonly("num_points",
+                             &dolfinx::fem::Expression<T>::num_points)
+      .def_property_readonly("value_size",
+                             &dolfinx::fem::Expression<T>::value_size)
+      .def_property_readonly("value_shape",
+                             &dolfinx::fem::Expression<T>::value_shape)
+      .def_property_readonly("num_argument_dofs",
+                             &dolfinx::fem::Expression<T>::num_argument_dofs)
+      .def_property_readonly("X", [](const dolfinx::fem::Expression<T>& self){ return dolfinx_wrappers::xt_as_pyarray(std::move(self.X())); } );
```
jhale, line 500: Nasty hack when I was tired, needs removing

python/dolfinx/wrappers/la.cpp
```diff
@@ -15,6 +15,7 @@
 #include <memory>
 #include <petsc4py/petsc4py.h>
 #include <pybind11/complex.h>
+#include <pybind11/functional.h>
```
jhale, line 18: Remove?

python/dolfinx/fem/function.py
```diff
@@ -120,80 +121,86 @@ def __init__(self, ufl_expression: ufl.core.expr.Expr, x: np.ndarray,
...
         ufl_constants = ufl.algorithms.analysis.extract_constants(ufl_expression)
-        constants = [ufl_constant._cpp_object for ufl_constant in ufl_constants]
+        constants = [constant._cpp_object for i, constant in enumerate(ufl_constants)]
 
-        # Getcpp Expression type
-        def expressiontype(dtype):
+        def create_expression(dtype):
             if dtype is np.float64:
-                return _cpp.fem.Expression_float64
+                return _cpp.fem.create_expression_float64
```
jhale, line 136: Could probably copy DirichletBCMetaClass and FormMetaClass approach for Expression.

cpp/dolfinx/fem/Expression.h
jhale: Thanks, will change. I'll probably be learning new C++ peculiarities until the end of my days...

python/dolfinx/fem/function.py
garth-wells: See https://github.com/FEniCS/dolfinx/pull/1957.

cpp/demo/hyperelasticity/hyperelasticity.ufl
```diff
@@ -59,13 +59,22 @@ lmbda = E*nu/((1 + nu)*(1 - 2*nu))
...
+J_form = derivative(F_form, u, du)
+
+# Compute von Mises stress in current configuration
+sigma = (1/J)*diff(psi, F)*F.T
+sigma_deviatoric = sigma - (1 / 3) * tr(sigma) * I
+sigma_von_mises = sqrt((3 / 2) * inner(sigma_deviatoric, sigma_deviatoric))
+
+forms = [F_form, J_form]
+elements = [(element)]
+expressions = [(sigma_von_mises, [[0.25, 0.25, 0.25]])]
```
jorgensd, line 77: As mentioned in our Slack discussion, I think it would be even better to use this demo to highlight that we can do interpolation into suitable function spaces.

Rephrasing this as:
```python
 expressions = [(sigma_von_mises, TensorElement("Discontinous Lagrange", "DG", (3,3))]
 ```
and using the data from: https://github.com/FEniCS/dolfinx/pull/1889
to be able to visualize the stress on the whole mesh without a projection

Comment (jhale) [reply cpp/demo/hyperelasticity/hyperelasticity.ufl:77]: Agreed, will change. In the meantime we've discovered that FFCx cannot support forms, elements and expressions simultaneously in it's output, so we'll change that first.

File: python/test/unit/fem/test_expression.py [hunk 12]
```diff
@@ -251,61 +357,3 @@ def e_exact(x):
             x = mesh.geometry.cmap.push_forward(quadrature_points, xg)
             e_exact_eval[Q_dofs_unrolled[cell]] = e_exact(x.T).T.flatten()
         assert np.allclose(local.array, e_exact_eval)
-
-
-@pytest.mark.skipif(np.issubdtype(PETSc.ScalarType, np.complexfloating),
```
jorgensd, line 256: I think this test should be added back in the final PR request.

cpp/demo/hyperelasticity/hyperelasticity.ufl
michalhabera: Processing of multiple different objects is now fixed. I'd prefer to go with @jorgensd 's suggestion in a separate PR. We should have this PR merged asap.

jorgensd: Thats fine

python/test/unit/fem/test_expression.py
michalhabera: It would be better to include this into a demo. I agree it demonstrates useful technique, but it is not a test of untested code path itself, right?

jorgensd: The code path is already tested, that is true.
There are quite a lot of things we should put in demos, as already discussed earlier in this PR.
I guess we need to make a plan of what to show where, as currently, expression is only used in 
elasticity demo, and it does not really highlight any use-case, as it only evaluates the von mises stresses on the reference element (without a pullback from the physical coordinate system as done in this test).
The demo should either do as in this test, or use interpolate and put it in a suitable function space.

cpp/dolfinx/fem/Expression.h
```diff
@@ -112,11 +114,39 @@ class Expression
...
+    {
+      num_argument_dofs
+          = _argument_function_space->dofmap()->element_dof_layout().num_dofs();
+      const auto element = _argument_function_space->element();
+
+      if (element->needs_dof_transformations())
+      {
+        _mesh->topology_mutable().create_entity_permutations();
+        cell_info = xtl::span(_mesh->topology().get_cell_permutation_info());
+        dof_transform = element->get_dof_transformation_function<T>();
```
jhale, line 138: @michalhabera On my new version of clang this gave me an error unless I removed the <T>?

cpp/demo/hyperelasticity/main.cpp
```diff
@@ -194,9 +196,36 @@ int main(int argc, char* argv[])
...
 
+    // Calculate Cauchy stress
+    // Construct appropriate Basix element for stress
+    const auto family = basix::element::family::P;
+    const auto cell_type = basix::cell::type::tetrahedron;
+    const auto variant = basix::element::lagrange_variant::equispaced;
+    const int k = 0;
+    const bool discontinuous = true;
+
+    const auto S_element
```
garth-wells, line 207: I'd suggest
```
basix::FiniteElement  S_element = basix::create_element(basix::element::family::P, basix::cell::type::tetrahedron, 0, basix::element::lagrange_variant::equispaced, true);
```
Use `auto` can hinder readability.

Might be nice to get the Basix element type from the mesh. We have `mesh::cell_type_to_basix_type` for this.

cpp/demo/hyperelasticity/main.cpp
```diff
@@ -194,9 +196,36 @@ int main(int argc, char* argv[])
...
+        = basix::create_element(family, cell_type, k, variant, discontinuous);
+    auto S = std::make_shared<fem::FunctionSpace>(
+        fem::create_functionspace(mesh, S_element, 9));
+
+    const auto sigma_expression
```
garth-wells, line 212: Does this need to be a pointer?

cpp/demo/hyperelasticity/main.cpp
```diff
@@ -194,9 +196,36 @@ int main(int argc, char* argv[])
...
+        = std::make_shared<fem::Expression<T>>(fem::create_expression<T>(
+            *expression_hyperelasticity_sigma, {{"u", u}}, {}, S));
+
+    auto sigma = std::make_shared<fem::Function<T>>(S);
```
garth-wells, line 216: Does this need to be a pointer?

cpp/demo/hyperelasticity/main.cpp
```diff
@@ -194,9 +196,36 @@ int main(int argc, char* argv[])
...
+    sigma->name = "cauchy_stress";
+    sigma->interpolate(*sigma_expression);
+
     // Save solution in VTK format
-    io::VTKFile file(MPI_COMM_WORLD, "u.pvd", "w");
-    file.write({*u}, 0.0);
+    io::VTKFile file_u(MPI_COMM_WORLD, "u.pvd", "w");
```
garth-wells, line 221: Get the communicator from the mesh.

cpp/demo/hyperelasticity/main.cpp
```diff
@@ -194,9 +196,36 @@ int main(int argc, char* argv[])
...
+    const auto family = basix::element::family::P;
+    const auto cell_type = basix::cell::type::tetrahedron;
+    const auto variant = basix::element::lagrange_variant::equispaced;
+    const int k = 0;
+    const bool discontinuous = true;
+
+    const auto S_element
+        = basix::create_element(family, cell_type, k, variant, discontinuous);
+    auto S = std::make_shared<fem::FunctionSpace>(
+        fem::create_functionspace(mesh, S_element, 9));
```
garth-wells, line 210: Could the `9` be made dimension independent? I.e. `gdim**2`?

cpp/dolfinx/fem/Expression.h
```diff
@@ -72,7 +78,7 @@ class Expression
   std::vector<int> coefficient_offsets() const
   {
     std::vector<int> n{0};
-    for (auto& c : _coefficients)
+    for (const auto& c : _coefficients)
```
garth-wells, line 102: No need to `const` here.

cpp/dolfinx/fem/Expression.h
```diff
@@ -82,22 +88,18 @@ class Expression
   }
 
   /// Evaluate the expression on cells
-  /// @param[in] cells Cells on which to evaluate the Expression
+  /// @param[in] active_cells Cells on which to evaluate the Expression
```
garth-wells, line 112: `cells` is nicer than `active_cells`. Not sure what 'active' means in this context.

cpp/dolfinx/fem/Expression.h
```diff
@@ -112,11 +114,43 @@ class Expression
     // Create data structures used in evaluation
     std::vector<double> coordinate_dofs(3 * num_dofs_g);
 
+    int num_argument_dofs = 1;
+    xtl::span<const std::uint32_t> cell_info;
+    std::function<void(const xtl::span<T>&,
```
garth-wells, line 140: Use `auto` here.

cpp/dolfinx/fem/Expression.h
```diff
@@ -112,11 +114,43 @@ class Expression
...
+          = _argument_function_space->dofmap()->element_dof_layout().num_dofs();
+      const auto element = _argument_function_space->element();
+
+      if (element->needs_dof_transformations())
+      {
+        _mesh->topology_mutable().create_entity_permutations();
+        cell_info = xtl::span(_mesh->topology().get_cell_permutation_info());
+        dof_transform_to_transpose
+            = element
+                  ->template get_dof_transformation_to_transpose_function<T>();
```
garth-wells, line 162: What's does `template` do here?

cpp/dolfinx/fem/Expression.h
```diff
@@ -125,28 +159,32 @@ class Expression
...
+
+      for (std::size_t j = 0; j < values_local.size(); ++j)
+        values(c, j) = values_local[j];
     }
   }
 
   /// Get function for tabulate_expression.
-  /// @return Function to tabulate expression.
-  const std::function<void(T*, const T*, const T*, const double*)>&
+  /// @param[out] fn Function to tabulate expression.
```
garth-wells, line 195: This should be `@return` rather than `@param[out]`.

cpp/dolfinx/fem/utils.h
```diff
@@ -682,6 +686,117 @@ pack_coefficients(const Form<T>& form, IntegralType integral_type, int id)
...
+  {
+    throw std::runtime_error(
+        "Expression has Argument but no Argument function space was provided");
+  }
+
+  if (argument_function_space)
+  {
+    if (!mesh)
+      mesh = argument_function_space->mesh();
+    if (argument_function_space->mesh()->id() != mesh->id())
```
garth-wells, line 708: No need to use `id` to check. Can use `if (argument_function_space->mesh() != mesh)`.

cpp/dolfinx/fem/utils.h
```diff
@@ -682,6 +686,117 @@ pack_coefficients(const Form<T>& form, IntegralType integral_type, int id)
...
+  }
+  else if constexpr (std::is_same<T, double>::value)
+    tabulate_tensor = expression.tabulate_tensor_float64;
+  else if constexpr (std::is_same<T, std::complex<double>>::value)
+  {
+    tabulate_tensor
+        = reinterpret_cast<void (*)(T*, const T*, const T*, const double*,
+                                    const int*, const unsigned char*)>(
+            expression.tabulate_tensor_complex128);
+  }
```
garth-wells, line 743: Do we need a final `else` with an exception?

cpp/dolfinx/fem/utils.h
```diff
@@ -682,6 +686,117 @@ pack_coefficients(const Form<T>& form, IntegralType integral_type, int id)
...
+  assert(tabulate_tensor);
+
+  return fem::Expression<T>(coefficients, constants, mesh, points,
```
garth-wells, line 746: Will it compile without `<T>`?

python/dolfinx/fem/function.py
```diff
@@ -327,10 +327,16 @@ def _interpolate(u, cells):
         def _(u_ptr, cells):
             self._cpp_object.interpolate_ptr(u_ptr, cells)
 
+        @_interpolate.register(Expression)
+        def _(expr: Expression, cells: np.ndarray = None):
+            # Interpolate expression for the set of cells
```
garth-wells, line 341: Use `"""Interpolate expression for the set of cells"""`.

garth-wells (changes requested):
Looks good. The part I'm not clear on is how to determine the locations on the reference cell when I want to interpolate the Expression. I think we could quite easily get these points from Basix/FFCx. The UFL file could have the element and the expression - the user could create a FiniteElement and Expression using UFCx data, and then interpolate.

jorgensd:
> Looks good. The part I'm not clear on is how to determine the locations on the reference cell when I want to interpolate the Expression. I think we could quite easily get these points from Basix/FFCx. The UFL file could have the element and the expression - the user could create a FiniteElement and Expression using UFCx data, and then interpolate.

I thought we could have an interface as follows:
```

> Looks good. The part I'm not clear on is how to determine the locations on the reference cell when I want to interpolate the Expression. I think we could quite easily get these points from Basix/FFCx. The UFL file could have the element and the expression - the user could create a FiniteElement and Expression using UFCx data, and then interpolate.

```
# ...Define forms
forms = [F_form, J_form]
expressions = [(sigma, i_element)]
```
and then ffcx could fetch the appropriate interpolation points using basix.

File: cpp/dolfinx/fem/Expression.h (hunk 22 above)
Comment (jhale) [reply cpp/dolfinx/fem/Expression.h:159]: ISO C++03 14.2/4:

When the name of a member template specialization appears after . or -> in a postfix-expression, or after nested-name-specifier in a qualified-id, and the postfix-expression or qualified-id explicitly depends on a template-parameter (14.6.2), the member template name must be prefixed by the keyword template. Otherwise the name is assumed to name a non-template.

**Review by jhale — APPROVED** [review]

File: cpp/demo/hyperelasticity/main.cpp [hunk 28]
```diff
@@ -160,43 +164,73 @@ int main(int argc, char* argv[])
...
+                          - (x2 - x2_c) * std::cos(theta) - x2));
           return values;
         });
 
     // Create Dirichlet boundary conditions
     auto bdofs_left
         = fem::locate_dofs_geometrical({*V},
                                        [](auto& x) -> xt::xtensor<bool, 1> {
-                                         return xt::isclose(xt::row(x, 0), 0.0);
+                                         auto x0 = xt::row(x, 0);
```
garth-wells, line 182: Any reason for introducing extra lines?

cpp/demo/hyperelasticity/main.cpp
```diff
@@ -160,43 +164,73 @@ int main(int argc, char* argv[])
...
     newton_solver.setF(problem.F(), problem.vector());
     newton_solver.setJ(problem.J(), problem.matrix());
     newton_solver.set_form(problem.form());
 
     la::petsc::Vector _u(la::petsc::create_vector_wrap(*u->x()), false);
     newton_solver.solve(_u.vec());
 
+    // Compute Cauchy stress
+    // Construct appropriate Basix element for stress
+    const auto family = basix::element::family::P;
```
garth-wells, line 207: If you want `const` here and for `k`, `variant`, etc, better to use `constexpr`.

cpp/dolfinx/fem/Expression.h
```diff
@@ -28,29 +28,45 @@ class Constant;
...
       const xt::xtensor<double, 2>& X,
-      const std::function<void(T*, const T*, const T*, const double*)> fn,
-      std::size_t value_size)
-      : _coefficients(coefficients), _constants(constants), _mesh(mesh), _x(X),
-        _fn(fn), _value_size(value_size)
+      const std::function<void(T*, const T*, const T*, const double*,
+                               const int*, const uint8_t*)>
+          fn,
+      const std::vector<int>& value_shape,
+      const std::shared_ptr<const mesh::Mesh> mesh = nullptr,
```
garth-wells, line 56: It's unusual to pass by value with `const`. Normally I'd expect non-const, or `const` and pass by reference.

cpp/dolfinx/fem/Expression.h
```diff
@@ -28,29 +28,45 @@ class Constant;
...
+      = nullptr)
+      : _coefficients(coefficients), _constants(constants), _mesh(mesh),
+        _x_ref(X), _fn(fn), _value_shape(value_shape),
+        _argument_function_space(argument_function_space)
   {
-    // Do nothing
+    // Extract mesh from argument's function space
+    if (!_mesh and argument_function_space)
+      _mesh = argument_function_space->mesh();
+    else if (argument_function_space and _mesh != argument_function_space->mesh())
```
garth-wells, line 66: The logic here around `_mesh` is a bit confusing. I would suggest making the `else if` a separate check before extracting. The first `if` changes `_mesh` and the second check involves `_mesh` but doesn't see the possible change. Correct but potentially confusing.

cpp/dolfinx/fem/Expression.h
```diff
@@ -112,8 +141,39 @@ class Expression
...
+             std::int32_t, int)
+    {
+      // Do nothing
+    };
+
+    if (_argument_function_space)
+    {
+      num_argument_dofs
+          = _argument_function_space->dofmap()->element_dof_layout().num_dofs();
+      const auto element = _argument_function_space->element();
```
garth-wells, line 160: No need for `const` here.

Add `assert(element)` before using `element`.

cpp/dolfinx/fem/utils.h
```diff
@@ -723,6 +728,116 @@ void pack_coefficients(const Form<T>& form, IntegralType integral_type, int id,
...
 }
 
+/// Create Expression from UFC
+template <typename T>
+fem::Expression<T> create_expression(
+    const ufcx_expression& expression,
+    const std::vector<std::shared_ptr<const fem::Function<T>>>& coefficients,
+    const std::vector<std::shared_ptr<const fem::Constant<T>>>& constants,
+    std::shared_ptr<const mesh::Mesh> mesh = nullptr,
+    const std::shared_ptr<const fem::FunctionSpace> argument_function_space
```
garth-wells, line 738: See earlier comment on pass by value/reference and `const`.

cpp/dolfinx/fem/utils.h
```diff
@@ -723,6 +728,116 @@ void pack_coefficients(const Form<T>& form, IntegralType integral_type, int id,
...
+/// Create Expression from UFC input
+/// (with named coefficients and constants)
+template <typename T>
+fem::Expression<T> create_expression(
+    const ufcx_expression& expression,
+    const std::map<std::string, std::shared_ptr<const fem::Function<T>>>&
+        coefficients,
+    const std::map<std::string, std::shared_ptr<const fem::Constant<T>>>&
+        constants,
+    const std::shared_ptr<const mesh::Mesh> mesh = nullptr,
```
garth-wells, line 798: See earlier comments.

python/demo/elasticity/demo_elasticity.py
```diff
@@ -157,11 +158,26 @@ def sigma(v):
 solver.solve(b, u.vector)
 solver.view()
 
+# Compute von Mises stress via interpolation
+sigma_deviatoric = sigma(u) - (1 / 3) * tr(sigma(u)) * Identity(len(u))
+sigma_von_mises = sqrt((3 / 2) * inner(sigma_deviatoric, sigma_deviatoric))
+
+W = FunctionSpace(mesh, ("Discontinuous Lagrange", 0))
+print(W.element.interpolation_points)
```
garth-wells, line 166: Left over debug.

python/dolfinx/fem/function.py
```diff
@@ -99,110 +99,139 @@ def __init__(self, ufl_expression: ufl.core.expr.Expr, x: np.ndarray,
...
 
     @property
-    def x(self) -> np.ndarray:
+    def X(self) -> np.ndarray:
         """Evaluation points on the reference cell"""
-        return self._cpp_object.x
+        return self._cpp_object.X
+
+    @property
+    def num_points(self) -> int:
```
garth-wells, line 207: Could this be removed and users does `expr.X.shape[0]`?

python/dolfinx/fem/function.py
```diff
@@ -99,110 +99,139 @@ def __init__(self, ufl_expression: ufl.core.expr.Expr, x: np.ndarray,
...
+        """Number of points on reference cell"""
+        return self._cpp_object.X.shape[0]
```
garth-wells, line 209: The array still 2D for a single point?

python/dolfinx/fem/function.py
```diff
@@ -99,110 +99,139 @@ def __init__(self, ufl_expression: ufl.core.expr.Expr, x: np.ndarray,
...
 
+    @property
+    def argument_function_space(self) -> typing.Optional[FunctionSpace]:
+        """Return argument function space if expression has argument"""
+        return self._argument_function_space
+
     @property
     def ufcx_expression(self):
-        """The compiled ufcx_expression object"""
+        """Return the compiled ufcx_expression object"""
```
garth-wells, line 223: Old docstring was better - same info, less text.

python/dolfinx/fem/function.py
```diff
@@ -330,10 +359,16 @@ def _interpolate(u, cells):
         def _(u_ptr, cells):
             self._cpp_object.interpolate_ptr(u_ptr, cells)
 
+        @_interpolate.register(Expression)
+        def _(expr: Expression, cells: np.ndarray = None):
```
garth-wells, line 363: Type hint is wrong `None` is not a `np.ndarray`.

Could remove type hint since this is a 'private' function.

python/dolfinx/fem/function.py
```diff
@@ -330,10 +359,16 @@ def _interpolate(u, cells):
...
+            """Interpolate expression for the set of cells"""
```
garth-wells, line 364: `Expression` or `expression`?

python/dolfinx/wrappers/fem.cpp
```diff
@@ -449,67 +449,81 @@ void declare_objects(py::module& m, const std::string& type)
...
+              py::init(
+                  [](const std::vector<std::shared_ptr<
+                         const dolfinx::fem::Function<T>>>& coefficients,
+                     const std::vector<std::shared_ptr<
+                         const dolfinx::fem::Constant<T>>>& constants,
+                     const std::shared_ptr<const dolfinx::mesh::Mesh>& mesh,
+                     const py::array_t<double, py::array::c_style>& X,
+                     std::uintptr_t fn_addr,
+                     const std::vector<int>& value_shape,
+                     const std::shared_ptr<const fem::FunctionSpace>
```
garth-wells, line 462: See earlier comments on pass by value/reference.

python/dolfinx/wrappers/fem.cpp
```diff
@@ -449,67 +449,81 @@ void declare_objects(py::module& m, const std::string& type)
...
+                        tabulate_expression_ptr, value_shape, mesh,
+                        function_space);
+                  }),
+              py::arg("coefficients"), py::arg("constants"), py::arg("mesh"),
+              py::arg("X"), py::arg("fn"), py::arg("value_shape"),
+              py::arg("function_space"))
+          .def(
+              "eval",
+              [](const dolfinx::fem::Expression<T>& self,
+                 const py::
```
garth-wells, line 480: Unusual line break.

python/dolfinx/wrappers/fem.cpp
```diff
@@ -449,67 +449,81 @@ void declare_objects(py::module& m, const std::string& type)
...
+      = std::string("create_expression_") + type;
+  m.def(
+      pymethod_create_expression.c_str(),
+      [](const std::uintptr_t expression,
+         const std::vector<std::shared_ptr<const dolfinx::fem::Function<T>>>&
+             coefficients,
+         const std::vector<std::shared_ptr<const dolfinx::fem::Constant<T>>>&
+             constants,
+         const std::shared_ptr<const dolfinx::mesh::Mesh> mesh,
+         const std::shared_ptr<const dolfinx::fem::FunctionSpace>
```
garth-wells, line 518: See earlier comments.

python/test/unit/fem/test_expression.py
```diff
@@ -1,102 +1,202 @@
...
+from petsc4py import get_config as PETSc_get_config
+import dolfinx.cpp
+dolfinx.cpp.common.init_logging(["-v"])
+# Get details of PETSc install
+petsc_dir = PETSc_get_config()['PETSC_DIR']
+petsc_arch = petsc4py.lib.getPathArchPETSc()[1]
+petsc_lib_name = ctypes.util.find_library("petsc")
+
+# Get PETSc int and scalar types
+if np.dtype(PETSc.ScalarType).kind == 'c':
```
garth-wells, line 34: This are a few nicer NumPy ways to check if a type is complex, e.g.
```
np.issubdtype(PETSc.ScalarType, np.complexfloating)
```

python/test/unit/fem/test_expression.py
```diff
@@ -1,102 +1,202 @@
...
+    dofmap = vdP1.dofmap.list.array
+    scatter(b.vector.array, array_evaluated, dofmap)
 
-    # Prepare mesh and dofmap data
-    pos = mesh.geometry.dofmap.offsets
-    x_dofs = mesh.geometry.dofmap.array
-    x = mesh.geometry.x
-    coeff_dofmap = P2.dofmap.list.array
-    dofmap = vP1.dofmap.list.array
+    def grad_expr1(x):
```
garth-wells, line 128:
```
def grad_expr1(x):
   return np.vstack((2.0 * x[0],  4.0 * x[1])) 
```
is nicer.

python/test/unit/fem/test_expression.py
```diff
@@ -200,9 +300,12 @@ def test_assembly_into_quadrature_function():
     Q_element = ufl.VectorElement("Quadrature", ufl.triangle, quadrature_degree, quad_scheme="default")
     Q = FunctionSpace(mesh, Q_element)
 
+    def T_exact(x):
+        return x[0] + 2.0 * x[1]
+
     P2 = FunctionSpace(mesh, ("P", 2))
     T = Function(P2)
-    T.interpolate(lambda x: x[0] + 2.0 * x[1])
```
garth-wells, line 205: This is nicer than the change.

garth-wells (changes requested):

cpp/demo/hyperelasticity/main.cpp
```diff
@@ -160,43 +164,73 @@ int main(int argc, char* argv[])
           auto x1 = xt::row(x, 1);
           auto x2 = xt::row(x, 2);
           xt::xarray<double> values = xt::zeros_like(x);
-          xt::row(values, 1) = scale
-                               * (x1_c + (x1 - x1_c) * std::cos(theta)
-                                  - (x2 - x2_c) * std::sin(theta) - x1);
-          xt::row(values, 2) = scale
-                               * (x2_c + (x1 - x1_c) * std::sin(theta)
-                                  - (x2 - x2_c) * std::cos(theta) - x2);
+          auto _row1 = xt::row(values, 1);
```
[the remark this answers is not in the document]
michalhabera: I followed https://github.com/FEniCS/dolfinx/pull/1646. I'll revert the changes and see if Intel is green.

cpp/dolfinx/fem/Expression.h
michalhabera: I think we should use non-const shared_ptr by value. Const ptr by reference means it only "might" retain refcount, https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-sharedptrparam-const, but we know we do certainly retain refcount.

I'll use const ptr by reference here to be consistent with the rest codebase.

garth-wells: I think in our use, const ref does mean that an object will be created/returned that will bump the ref count. Otherwise we would pass a plain reference.

michalhabera: Const ref doesn't tell you if the shared_ptr is later assigned (constructed) into member variable. I think thats what guideline is trying to say. But since we know we always do `_mesh(mesh)` then we should pass ptr by value, so interface makes that clear.

python/test/unit/fem/test_expression.py
```diff
@@ -1,102 +1,202 @@
...
+                         VectorFunctionSpace, create_sparsity_pattern, form)
 from dolfinx.mesh import create_unit_square
 
+import petsc4py.lib
 from mpi4py import MPI
 from petsc4py import PETSc
+from petsc4py import get_config as PETSc_get_config
+import dolfinx.cpp
+dolfinx.cpp.common.init_logging(["-v"])
+# Get details of PETSc install
```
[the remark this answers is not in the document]
michalhabera: This whole chunk for PETSc handling is copy&paste from `test_custom_assembler.py`. PETSc is required here for matrix data structure into which we assemble global interpolation operator.

python/dolfinx/fem/function.py
michalhabera: Yes.

python/test/unit/fem/test_expression.py
```diff
@@ -1,102 +1,202 @@
...
-    # Data structure for the result
-    b = Function(vP1)
-    assemble_expression(b.vector.array, compiled_expr.tabulate_expression,
-                        (pos, x_dofs, x), dofmap, f.vector.array, coeff_dofmap)
+        return values
 
-    b2 = Function(vP1)
-    b2.interpolate(lambda x: np.vstack((2.0 * x[0], 4.0 * x[1])))
+    b2 = Function(vdP1)
+    b2.interpolate(grad_expr1)
```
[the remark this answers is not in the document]
michalhabera: I've made `grad_expr1` according to your suggestion above. Why the test can't be working?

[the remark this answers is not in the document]
garth-wells: Left over comment - git diff separated the return line from the function body, looked like no return. 'above issue' comment was removed n draft.

garth-wells (approved):

PR #1861, opened 2021-12-02
Fix refinement and add tests for sub refinement

jorgensd:
Fixing bug in Python interface reported at: https://fenicsproject.discourse.group/t/what-is-the-latest-api-for-mesh-refinement-in-dolfinx/7201

Fixing bug in C++ refinement when number of new vertices on a process is 0.

Add tests for refinement using list of edges.

Reactivate gdim refinement test

Expose `compute_incident_entities` to use with refinement

### DIFF
--- a/cpp/dolfinx/refinement/utils.cpp
+++ b/cpp/dolfinx/refinement/utils.cpp
@@ -92,25 +92,24 @@ xt::xtensor<double, 2> create_new_geometry(
     for (std::size_t j = 0; j < x_g.shape(1); ++j)
       new_vertex_coordinates(v, j) = x_g(vertex_to_x[v], j);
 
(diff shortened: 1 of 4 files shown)


cpp/dolfinx/refinement/utils.cpp
```diff
@@ -99,13 +99,19 @@ xt::xtensor<double, 2> create_new_geometry(
 
   const xt::xtensor<double, 2> midpoints
       = mesh::compute_midpoints(mesh, 1, edges);
-  // The below should work, but misbehaves with the Intel icpx compiler
```
chrisrichardson, line 102: Could you just put an `if (num_new_vertices > 0)` here?

chrisrichardson (approved):

PR #1862, opened 2021-12-02
Enable setting of graph partitioner used in cell partitioning from Python

### LINKED ISSUE #1859: Enable choice of ParMETIS, SCOTCH or KaHIP partitioners from Python
It is possible to pass the chosen partitioner to the mesh creation functions from C++, but this is not possible from Python. What is needed is to expose the different partitioning functions in Python.

garth-wells:
This was previously only possible from the C++ interface.

Fixes #1859.


python/dolfinx/wrappers/caster_mpi.h
```diff
@@ -38,14 +38,23 @@ class type_caster<dolfinx_wrappers::MPICommWrapper>
   // Python to C++
   bool load(handle src, bool)
   {
-    // Simplified version of isinstance(src, mpi4py.MPI.Comm) - avoids
-    // segfault when pybind11 tries to convert some other random type to
-    // MPICommWrapper
-    if (not hasattr(src, "Allgather"))
+    // NOTE: This check would be safer than hasattr but does not evaluate to
+    // true.
+    // if (PyObject_TypeCheck(src.ptr(), &PyMPIComm_Type))
```
jhale, line 43: @chrisrichardson Any idea why this check doesn't work? Perhaps you've been here before.

python/test/unit/mesh/test_mesh_partitioners.py
```diff
@@ -20,13 +20,37 @@
 assert (tempdir)
 
 
-@pytest.mark.parametrize("partitioner", [partition_cells_graph])
+partitioners = [dolfinx.cpp.graph.partitioner()]
```
jhale, line 23: Should be empty list? Existence of default partitioner covered in next try/except statement.

python/test/unit/mesh/test_mesh_partitioners.py
```diff
@@ -20,13 +20,37 @@
...
+try:
+    from dolfinx.cpp.graph import partitioner
+    partitioners.append(partitioner())
+except ImportError:
+    raise
+try:
+    from dolfinx.cpp.graph import partitioner_scotch
+    partitioners.append(partitioner_scotch())
+except ImportError:
+    raise
```
jhale, line 33: Should be pass? We can build just with e.g. Parmetis or KaHiP now?

jhale (changes requested):

python/test/unit/mesh/test_mesh_partitioners.py
garth-wells: Not yet.

python/setup.py
```diff
@@ -47,7 +47,8 @@ def build_extension(self, ext):
                       f'-DPython3_LIBRARIES={sysconfig.get_config_var("LIBDEST")}',
                       f'-DPython3_INCLUDE_DIRS={sysconfig.get_config_var("INCLUDEPY")}']
 
-        cfg = 'Debug' if self.debug else 'Release'
+        # cfg = 'Debug' if self.debug else 'Release'
```
jhale, line 50: Remember to backout.

jhale (approved):

PR #1864, opened 2021-12-03
Add isort configuration and non-blocking CI step

jhale:


jorgensd:
Could we add an `.isort.cfg` file to DOLFINx with the ruleset you use, as the default sort import command (in for instance vscode) will not pick up these options: https://code.visualstudio.com/docs/python/editing#_sort-imports

jorgensd:
@jhale Adding:
```
[isort]
known_first_party = dolfinx
```
to setup.cfg seems to do the trick for vscode

jhale:
That's the plan!

I'm still playing around with it. @garth-wells pointed out to me that isort has some undesirable 'clever' default behaviour depending on the environment. For example, what is in site-packages seems to get put in third-party automatically.

jhale:
@jorgensd Does this branch work properly in vscode?

jorgensd:
> @jorgensd Does this branch work properly in vscode?

Yes, this seems to be working with vscode (Tested demo_elasticity.py and demo_gmsh.py)

PR #1875, opened 2021-12-08
Interpolation of ufl expressions

jorgensd:
This is a prototype implementation of interpolation of ufl expressions into a suitable function space. 
The implementation leverages `dolfinx.Expression`, and can also be used to interpolate over sub-sets of cells effectively.
This can for instance be used for setting spatially varying boundary conditions.

Various tests added


cpp/dolfinx/fem/interpolate.h
```diff
@@ -640,4 +644,124 @@ void interpolate(Function<T>& u, const Function<T>& v)
   }
 }
 
+/// Interpolate from a dolfinx Expression into a compatible Function on the same
+/// mesh
+/// @param[in, out] u The function to interpolate into
+/// @param[in] expr The expression to be interpolated
+/// @param[in] cells List of cell indices (local to process) to interpolate on
+template <typename T>
+void interpolate(Function<T>& u, const Expression<T>& expr,
```
garth-wells, line 653: This looks like a lot of duplicated code. I think with some refactoring it will be possible to increase code re-use.

jorgensd: Code has now been refactored to re-use interpolate(u, f, cells)

jhale:
Could you merge or branch off of PR #1858?

jorgensd:
> Could you merge or branch off of PR #1858?
I would prefer it the other way around, as this code does not change the current implementation of expression, it only uses the functionality that exists.

It also adds several tests (previously there were very few tests for expressions).

Also, it is unclear to me what `num_argument_dofs` in the new initializer is.

michalhabera:
> > Could you merge or branch off of PR #1858?
> 
> I would prefer it the other way around, as this code does not change the current implementation of expression, it only uses the functionality that exists.
> 
> It also adds several tests (previously there were very few tests for expressions).
> 
> Also, it is unclear to me what `num_argument_dofs` in the new initializer is.

I will push updates to #1858 . Num_argument_dofs is replaced with pointer functions space for argument. For matrix-free interpolation you are writing this is just nullptr.

python/dolfinx/fem/function.py
```diff
@@ -312,6 +312,19 @@ def _(u_ptr):
 
         _interpolate(u)
 
+    def interpolate_cells(self, ufl_expression: ufl.core.expr.Expr, cells: np.ndarray = None):
```
garth-wells, line 315: I'm not sure what this function name means?

garth-wells: I still don't understand what `interpolate_cells` means. What is the meaning of `_cells`?

jorgensd: That you can interpolate on a subset of cells. If you supply the cell indices (local to process) as a list, you only perform integration over those entities

jorgensd: Could be `interpolate_over_cells`, or simply `interpolate`. As i said, I only gave it a distinct name to avoid unclear behavior in the Python layer

michalhabera: > I.e.
> 
> ```
> u = Function(V)
> w = Function(W)
> u.interpolate(w)
> u.interpolate_cells(w)
> ```
> 
> would produce the same result with very different code paths.

I think it would be better if you do not put this interpolation into Function's class. If you keep it `Expression.interpolate` it makes things more clear.

jorgensd: How would you make that work? Then the user has to supply the quadrature points for the space they are interpolating in to, which I think is rather unnecessary.

cpp/dolfinx/fem/Function.h
```diff
@@ -225,6 +228,15 @@ class Function
     fem::interpolate(*this, fx, cells);
   }
 
+  /// Interpolate an Expression (based on ufl)
+  /// @param[in] expr The function to be interpolated.
+  /// @param[in] cells The cells (local to process) to interpolate into
+  void interpolate(const Expression<T>& expr,
```
michalhabera, line 234: I'd remove this. It is confusing and should be executed with dolfinx.Expression.interpolate directly.

jorgensd: See my comment above. As long as expression need to take in the interpolation points, I thinks the interface becomes unnecessary explicit.

python/dolfinx/fem/function.py
michalhabera: I would like to have `Expression.interpolate(target_function, cells)`.

The explicitness of creating an Expression is good, because construction of Expression (at least in Python) involves potentially expensive JIT compilation. It is very dangerous to for user since
```
u.interpolate(f)
u.interpolate_cells(f)  # This calls FFCx and JIT!
```

jorgensd: My main issue is that you have to pass in the interpolation points to the `Expression`. What happens if a user passes in something else than the interpolation points? The code would then randomly crash at application of the interpolation matrix.

In clear-text, the interface you suggest would look like:
```
expr = Expression(ufl_form, interpolation_points)
expr.interpolate(u, cells)
```
which leaves it up to the user to define the correct interpolation points.

jhale: Hi Jorgen, this is a good patch, but it's brought up some tangential issues about where everything should be located and what the right interface should be. I'm currently having a think about it.

michalhabera: > My main issue is that you have to pass in the interpolation points to the `Expression`. What happens if a user passes in something else than the interpolation points? The code would then randomly crash at application of the interpolation matrix.
> 
> In clear-text, the interface you suggest would look like:
> 
> ```
> expr = Expression(ufl_form, interpolation_points)
> expr.interpolate(u, cells)
> ```
> 
> which leaves it up to the user to define the correct interpolation points.

I do not see the point, your free standing interpolate() has these issues anyway (just needs to add proper sanity checks) . But my main concern is only `Function.interpolate_cells()`, because it is not Function which you interpolate, it is the Function wrapped as an Expression.

We need to encourage users to do the right and fast thing. 
If you want to interpolate dolfinx.Function into another dolfinx.Function just call `u.interpolate(f)` - doesn't involve compilation and is faster.

jorgensd: > I do not see the point, your free standing interpolate() has these issues anyway (just needs to add proper sanity checks)

The free-standing interpolation function is not exposed to python, so users won't be able to access it. It can be put in an anonymous namespace if we don't want people to use it.


> We need to encourage users to do the right and fast thing.
> If you want to interpolate dolfinx.Function into another dolfinx.Function just call `u.interpolate(f)` - doesn't involve compilation and is faster.

I am experimenting with how to get single-dispatch to handle this nicely. i.e. default to the non-JIT interpolate if it is only a function that is supplied, but use JIT for anything else, such as linear combinations of functions from different spaces.

jhale: My two cents:
```
V = FunctionSpace(mesh, ("Lagrange", 2))
f = Function(V)
fem.function.interpolate(lambda x: A * k0**2 * np.cos(k0 * x[0]) * np.cos(k0 * x[1]), f)
```

```
V = FunctionSpace(mesh, ("Lagrange", 2))
f = Function(V)
g = Function(V)
fem.function.interpolate(g, f)
```

```
V = FunctionSpace(mesh, ("Lagrange", 2))
f = Function(V)
# TODO: Add explicit code for interpolation_points
expr = Expression(ufl_form, interpolation_points)
fem.expression.interpolate(expr, f)
```

jorgensd:
After discussions, the interface has been updated to the following:
- `interpolate` now takes in either a Function, Expression or lambda function.
- For the case of Expression, it also takes in which cells we should perform interpolation on.

The latest commit fixes an issue in interpolate for when not all cells are supplied on a process, as well as changing interpolation_points to a read_only property.

jhale (approved):

cpp/dolfinx/fem/interpolate.h
```diff
@@ -640,4 +645,43 @@ void interpolate(Function<T>& u, const Function<T>& v)
...
+  assert(dofmap);
+  const int element_vs = element->value_size();
+  assert(expr_value_size == element_vs);
+  auto X = expr.x();
+  assert(X.shape() == element->interpolation_points().shape());
+
+  // Evaluate expression at quadrature points for the subset of cells
+  expr.eval(cells, expr_values);
+
+  // Reshape evaluated data to fit interpolate
```
garth-wells, line 676: Can this be simplified, or the doc string expanded? It’s hard to follow the multiple reshaping steps.

jorgensd: Expression takes in a matrix of shape (num_cells, num_points*value_size), where the last dimension varies with xyzxyz.
Interpolate takes in a matrix of shape (value_size, num_cells*num_points), i.e xxyyzz ordering where all points of cell0 comes first, then cell1 etc.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -510,9 +514,10 @@ void interpolate(Function<T>& u, xt::xarray<T>& f,
                           xt::xall<std::size_t>>;
     auto pull_back_fn = element->map_fn<U_t, U_t, J_t, J_t>();
 
-    for (std::int32_t c : cells)
+    for (std::size_t c = 0; c < cells.size(); ++c)
```
garth-wells, line 517: Why have the loops been changed from range-based?

jorgensd: Because you need to access both c and cells[c] (see [line 547](https://github.com/FEniCS/dolfinx/blob/d2add3693f50c8ba798e1ff8b8368757115de453/cpp/dolfinx/fem/interpolate.h#L547))
The code in main does not work if you send in a set of cells not equal to the local range.
Changing the loop here fixes this bug

jorgensd: Similar comment is relevant for 453 and the first for at 446

cpp/dolfinx/fem/interpolate.h
garth-wells: This is still confusing - why is the reshaping done in more than one step?

cpp/dolfinx/fem/interpolate.h
```diff
@@ -646,4 +650,51 @@ void interpolate(Function<T>& u, const Function<T>& v)
   }
 }
 
+/// Interpolate from an Expression into a compatible Function on the
+/// same mesh
+/// @param[out] u The function to interpolate into
+/// @param[in] expr The expression to be interpolated
```
garth-wells, line 656: The conditions on the Expression need to be clear and explicit, e.g. the evaluation point locations.

cpp/dolfinx/fem/interpolate.h
jorgensd: How would you do it in one step?

garth-wells (approved):

PR #1881, opened 2021-12-13
Fix various bugs reported by sonarcloud

jorgensd:
Various fixes reported by sonarcloud: https://sonarcloud.io/project/issues?id=FEniCS_dolfinx&resolved=false&types=BUG

Removes two tests that have been skipped for a long time, and that I do not think is going to ever get fixed.


cpp/dolfinx/generation/RectangleMesh.cpp
```diff
@@ -197,9 +197,7 @@ mesh::Mesh build_tri(MPI_Comm comm,
 
           switch (diagonal)
           {
-          case DiagonalType::right_left:
-            local_diagonal = DiagonalType::right;
-            break;
+          case DiagonalType::right_left:
```
garth-wells, line 200: This bit of code is confusing.

jorgensd: We could use an if (x==a or x==b) instead of a switch ?

garth-wells: I think that would be better in this case. The present looks like a bug to me, even if it isn't.

cpp/dolfinx/generation/RectangleMesh.cpp
```diff
@@ -195,15 +195,9 @@ mesh::Mesh build_tri(MPI_Comm comm,
                                 xt::all());
           _cell.assign(c);
 
-          switch (diagonal)
-          {
-          case DiagonalType::right_left: 
-          case DiagonalType::left_right:
+          if (diagonal == DiagonalType::right_left
+              or diagonal == DiagonalType::left_right)
             local_diagonal = DiagonalType::right;
```
garth-wells, line 200: We use curly braces for multiline conditionals.

garth-wells (approved):

PR #1829, opened 2021-11-24
DirichletBC with dolfinx.Constant

jorgensd:
Resolves #507

Issue #1816, opened 2021-11-18
Ordering of shared indices in `IndexMap`


garth-wells:
It could be helpful to group the (1) exclusively owned, (2) owned and shared and (3) ghost indices, and the case (2) and (3) order by rank. At the moment we just have group (3) at the end of the array.

IgorBaratta:
```
||||||||||||||||||||||||||||||||||||||||||||||||||||||   |||||||||||||||||||||||||   |||||||||||
     (1)                                                      (2)                       (3)
```
Currently (1) and (2) are mixed, but it would be good to group so one could have a range for (2) with contiguous memory.
That would allow for example :+1: 
```c++
la::Vector<double, DeviceAllocator> x(indexmap, bs);
std::array range = x.shared_range();
int size = range[1] -range[0];

double* ptr = x.data();
double* shared_buffer = AllocateSharedMemory(size);
double* ghosts_buffer = AllocateSharedMemory(ghost_size);


// Device to device memcopy
memcopy(ptr+range[0], shared_buffer , size);
memcopy(ptr+range_ghosts[0], ghosts_buffer, ghost_size);

// If a GPU-aware mpi implementation is available no copy from device to host.
// Otherwise only copy the shared and ghost buffers.
indexmap.scatter_fwd(shared_buffer, ghosts_buffer);


```

chrisrichardson:
This might affect data-locality adversely.  I think, for now, it would be worth making it an optional choice, so we can test out the performance implications both ways.

garth-wells:
@IgorBaratta do you think we should worry about this?

garth-wells:
I think this is possibly to restrictive. Creator can always control order for data locality when taking slices, but `IndexMap` doesn't need to know about this.

PR #1883, opened 2021-12-13
Refactor DirichletBC

### LINKED ISSUE #507: Allow DirichletBC to be set with a Constant
At present, a DirichletBC can only be set with a `Function`, but it is often just a `Constant` value.

jorgensd:
Superseeds: #1829 and resolves #507 adding `dolfinx.Constant` as a valid input to the DirichletBC.

Refactoring of getting degrees of freedom from remote for topological searches, all different combinations:
`V`, `V.sub(i)`, `(V.sub(i)`, `V.sub(i).collapse())` uses the same functionality to fetch dofs that are not on an owned entity (facet), but is a dof on the process.


cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -168,6 +170,59 @@ class DirichletBC
...
+  /// @param[in] dofs Degree-of-freedom block indices (@p
+  /// std::vector<std::int32_t>) from the input function space to constrain
+  /// @note The dofs will be unrolled with the dofmap block size
+  template <typename U>
+  DirichletBC(const std::shared_ptr<const fem::Constant<T>>& g, U&& dofs,
+              const std::shared_ptr<const fem::FunctionSpace>& V)
+      : _function_space(V), _g(g), _dofs0(std::forward<U>(dofs))
+  {
+    assert(_function_space);
+    std::visit(
```
garth-wells, line 187: I don't understand what this is for. The Function interface specifies the type for `g`, so why is there a visitor pattern to check?

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -254,16 +327,46 @@ class DirichletBC
...
-    xtl::span<const T> g = _g->x()->array();
-    for (std::size_t i = 0; i < _dofs0.size(); ++i)
-    {
-      if (_dofs0[i] < (std::int32_t)x.size())
-      {
-        assert(_dofs1_g[i] < (std::int32_t)g.size());
-        x[_dofs0[i]] = scale * g[_dofs1_g[i]];
-      }
-    }
+    std::visit(
```
garth-wells, line 330: There are much simpler ways to do this, see https://www.cppstories.com/2018/06/variant/.

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -107,6 +169,8 @@ get_remote_bcs1(const common::IndexMap& map,
   const std::vector<std::int64_t>& ghosts = map.ghosts();
 
   // Build map from ghost to local position
+  // NOTE: Should we use map here or just one vector with ghosts and
```
IgorBaratta, line 172: Sorting takes `O(n * log(n))` , which is the same cost as the creation of the map.
`map.find` and `lower_bound`  are both `O(log(n)` . 

So if the ghosts were sorted I would use a vector intead of a map.

jorgensd: Currently ghosts are not sorted, so I guess this is a thing we can improve if we change the indexmap, ref #1816

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -50,55 +51,119 @@ dolfinx::MPI::Comm create_symmetric_comm(MPI_Comm comm)
...
+
+  if (bs_map == 1)
+    map.local_to_global(dofs_local, dofs_global);
+  else
+  {
+    // Convert dofs indices to 'block' map indices
+    std::vector<std::int32_t> dofs_local_block;
+    dofs_local_block.reserve(num_dofs_block);
+    for (std::size_t i = 0; i < dofs_local.size(); i += bs)
+      dofs_local_block.push_back(dofs_local[i] / bs_map);
```
jorgensd, line 146: @garth-wells here you have added the assumption that the dofs are sorted and contain no duplicates (which the input currently do) . Should we add that as a `@note for the input arguments`?

garth-wells: > @garth-wells here you have added the assumption that the dofs are sorted and contain no duplicates 

Not sure that is entirely true on ordering. There is an assumption that blocks remained blocked.

> (which the input currently do) . Should we add that as a `@note for the input arguments`?

The docs already say the dofs must be sorted (in docs for in argument). Feel free to add more if you think it's needed.

jorgensd: Agree that inside `get_remote_dofs` it does not have to be fully sorted, but the input has to be "blocked".
I will add a note on `get_remote_dofs` that we assume that the input dofs are blocked, but not necessarily sorted.

garth-wells (approved):

PR #1884, opened 2021-12-13
Rename mesh generation functions

jhale:


cpp/dolfinx/mesh/CMakeLists.txt
```diff
@@ -4,9 +4,12 @@ set(HEADERS_mesh
   ${CMAKE_CURRENT_SOURCE_DIR}/Geometry.h
   ${CMAKE_CURRENT_SOURCE_DIR}/Topology.h
   ${CMAKE_CURRENT_SOURCE_DIR}/MeshTags.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/box_mesh.h
```
garth-wells, line 7: Use the underscore in the header name is inconsistent with the style we use for all other headers.

I would suggest putting all 'generator' functions in a single header. The multiple files is legacy.

cpp/dolfinx/mesh/interval_mesh.cpp
```diff
@@ -73,7 +73,7 @@ mesh::Mesh build(MPI_Comm comm, std::size_t nx, std::array<double, 2> x,
 } // namespace
 
 //-----------------------------------------------------------------------------
-mesh::Mesh IntervalMesh::create(MPI_Comm comm, std::size_t n,
+mesh::Mesh interval_mesh::create(MPI_Comm comm, std::size_t n,
```
garth-wells, line 76: No need for a namespace here. `mesh::create_interval` would be fine. Same for other cases.

python/dolfinx/mesh.py
```diff
@@ -218,3 +221,166 @@ def MeshTags(mesh: Mesh, dim: int, indices: np.ndarray, values: np.ndarray) -> t
 
     fn = _meshtags_types[dtype]
     return fn(mesh, dim, indices.astype(np.int32), values)
+
+
+def create_interval_mesh(comm: _MPI.Comm, nx: int, points: list, ghost_mode=GhostMode.shared_facet,
```
garth-wells, line 226: No need for 'mesh' in the name when in the mesh module. See https://gist.github.com/sloria/7001839.

garth-wells (changes requested):
Main comments are:
- I think all the built-in generators should go in a one header and one implementation file.
- There is some redundant naming, with the module name appearing in functions names.
- No need for each generating to have its own namespace.

garth-wells (approved):

PR #1885, opened 2021-12-14
Remove PETSc Vec from `fem::Function`

garth-wells:
This change isolated PETSc in the C++ interface. It should now be possible to build the C++ interface without PETSc.

For now, the 'caching' behaviour of a PETSc Vec that wraps the a Functions data has been moved into the Python layer.

Change also makes more use of access the `la::Vector` object for less verbose setting of owned and ghost values in a vector.

### DIFF
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -1,0 +1,68 @@
+name: SonarCloud
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    container: fenicsproject/test-env:latest-mpich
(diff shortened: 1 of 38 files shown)


jorgensd:
Tangential comment: Should there be added a use-case where `scatter_reverse` is used in the python layer?
If one wraps the output from `assemble_vector` in a `dolfinx.Function`, we could avoid using PETSc for ghost-updates after applying lifting.

garth-wells:
> Tangential comment: Should there be added a use-case where `scatter_reverse` is used in the python layer? If one wraps the output from `assemble_vector` in a `dolfinx.Function`, we could avoid using PETSc for ghost-updates after applying lifting.

We need to give the Python layer some thought. It's more tightly (and less clearly) coupled with PETSc.

PR #1891, opened 2021-12-16
Update docker basic test to make images build

jorgensd:
Fix docker image builds


jhale:
I'll hit the docker image build trigger.

PR #1892, opened 2021-12-16
Fix GJK loop access

jorgensd:
Use known size in loop as reported by Sonarcloud


jhale:
Already discussed with Chris earlier in the week.

Issue #1872, opened 2021-12-07
RectangleMesh ignores z-component of Points


BenjaminRodenberg:
I'm currently migrating a FEniCS-based code to dolfinx. In my FEniCS code I used the following for creating a rectangle in a 2D domain:
```
from fenics import Point, RectangleMesh

p0 = Point(0, 0)
p1 = Point(1, 1)
nx, ny = 10, 10
mesh = RectangleMesh(p0, p1, nx, ny, diagonal="left")
```
I translated this to
```
from dolfinx import Point, RectangleMesh
from mpi4py import MPI

p0 = (0, 0, 0)
p1 = (1, 1, 0)
nx, ny = 10, 10
mesh = RectangleMesh(MPI.COMM_WORLD, [p0, p1], [nx, ny], diagonal="left")
```
Two things I found irritating:

## 1. Why do I have to provide a z-component, even if the mesh is in 2D?

If I only provide two components, I get an error:
```
>>> m = dolfinx.RectangleMesh(MPI.COMM_WORLD, [(0,0), (1,1)], [10,10], diagonal="left")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/benjamin/.local/lib/python3.8/site-packages/dolfinx/generation.py", line 76, in RectangleMesh
    mesh = cpp.generation.create_rectangle_mesh(comm, points, n, cell_type, ghost_mode, partitioner, diagonal)
TypeError: create_rectangle_mesh(): incompatible function arguments. The following argument types are supported:
    1. (comm: MPICommWrapper, p: List[List[float[3]][2]], n: List[int[2]], celltype: dolfinx.cpp.mesh.CellType, ghost_mode: dolfinx.cpp.mesh.GhostMode, partitioner: Callable[[MPICommWrapper, int, int, dolfinx.cpp.graph.AdjacencyList_int64, dolfinx.cpp.mesh.GhostMode], dolfinx.cpp.graph.AdjacencyList_int32], diagonal: str) -> dolfinx.cpp.mesh.Mesh

Invoked with: <mpi4py.MPI.Intracomm object at 0x7fb6a82c2f90>, [(0, 0), (1, 1)], [10, 10], <CellType.triangle: 3>, <GhostMode.shared_facet: 1>, <built-in method partition_cells_graph of PyCapsule object at 0x7fb69fdb96f0>, 'left'
```

## 2. I tried to add a non-zero z-component and found out that it is just ignored.

Is this a bug?
[...]

Issue #1873, opened 2021-12-07
Improve dolfinx.generation.RectangleMesh interface 


jorgensd:
Currently `dolfinx.generation.RectangleMesh` takes in a list of two 3D points, even if the last dimension of each point is ignored. 
Related to #1872 . This could be improved with some padding in the python wrapper.

jhale:
Fixed by #1894

PR #1894, opened 2021-12-16
Change create_rectangle to take two dimensional points

jhale:
Discussed in #1873. Consistent with `create_interval` and `create_box`.

### DIFF
--- a/cpp/demo/poisson/main.cpp
+++ b/cpp/demo/poisson/main.cpp
@@ -120,7 +120,7 @@ int main(int argc, char* argv[])
   {
     // Create mesh and function space
     auto mesh = std::make_shared<mesh::Mesh>(mesh::create_rectangle(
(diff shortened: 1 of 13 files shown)


garth-wells:
Are we sure we don't want to allow intervals and rectangles embedded in 3D?

jhale:
Personally, I'd use a mesh generator for anything more complex like that.

A rectangle in 3D would also require specification of a normal to the surface as well.

What do you think?

garth-wells:
> Personally, I'd use a mesh generator for anything more complex like that.
> 
> A rectangle in 3D would also require specification of a normal to the surface as well.
> 
> What do you think?

I has assumed that we did support built-in rectangles in 3D. If we didn't and we don't want to, that's fine with me.

jhale:
You and the user who reported https://github.com/FEniCS/dolfinx/issues/1872 !

jhale:
Fixes #1872 and #1873.

PR #1899, opened 2021-12-17
Flatten geometry storage

garth-wells:
We work with a fixed dimension of 3 for the geometry data storage, but by using `xt::xtensor<double, 2>` we don't exploit the known size at compile time. The change flattens the storage to  `std::vector<double>`. This avoids the use of xtensor views in performance critical parts of the code (xtensor views have known performance issues).

When it's convenient to have an xtensor objects, the underlying data can be easily wrapped using `xt::adapt`.

No changes to the Python interface for accessing the geometry.


IgorBaratta (approved):

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -231,37 +230,39 @@ void mesh(py::module& m)
...
-            {
-              std::array shape
-                  = {std::size_t(x.shape(0)), std::size_t(x.shape(1))};
-              auto _x
-                  = xt::adapt(x.data(), x.size(), xt::no_ownership(), shape);
-              return dolfinx::mesh::Geometry(map, dofmap, element, _x,
-                                             std::move(indices));
-            }
-          }))
+      //   .def(py::init(
```
IgorBaratta, line 233: Just a minor question:
Is this constructor being intentionally disabled?

garth-wells: It's been removed. It's not used, so I didn't bother updating it.

PR #1900, opened 2021-12-17
Update special functions tests

jorgensd:
Previous test was outdated and would not execute.
Test updated to run in both serial and parallel.


python/test/unit/fem/test_special_functions.py
```diff
@@ -1,35 +1,56 @@
...
-#        assert round(assemble(c1*ds(mesh)) - ref_int, 7) == 0
+    ds = ufl.Measure("ds", domain=mesh)
+    a0 = mesh.comm.allreduce(assemble_scalar(c * ds), op=MPI.SUM)
+    a = mesh.comm.allreduce(assemble_scalar(c0 * ds), op=MPI.SUM)
+    assert numpy.isclose(a.real, 2)
+    assert numpy.isclose(a0.real, 2)
+
+
+@pytest.mark.parametrize('mesh_factory', [(create_unit_square, (MPI.COMM_WORLD, 3, 3), 1. / 3),
+                                          #   (create_unit_square,
```
garth-wells, line 35: @jorgensd could you add a comment why these are commented out?

jorgensd: Ufl does not support facetnormal of non-simplices for now. I guess we should add support for it

garth-wells: Can you put the comment in the code?

jorgensd: Ill do it tomorrow once Im at a computer

PR #1901, opened 2021-12-17
Implement DOF permutations for VectorElement

### LINKED ISSUE #1895: Higher-order mixed elements not work (dof permutation error)
The code
```python
import numpy as np
from dolfinx.fem import FunctionSpace
from mpi4py import MPI
import ufl

mesh = create_unit_square(MPI.COMM_WORLD, 2, 2)
P3 = ufl.VectorElement("Lagrange", mesh.ufl_cell(), 3)
P1 = ufl.FiniteElement("Lagrange", mesh.ufl_cell(), 1)
TH = P2 * P1
W = FunctionSpace(mesh, TH)
```
fails with the error
```
RuntimeError: Permuting DOFs for vector elements not implemented.
```
Changing the degree of `P3` from 3 to 2, the code works as expected.

mscroggs:
Fixes #1895.

### DIFF
--- a/cpp/dolfinx/fem/FiniteElement.cpp
+++ b/cpp/dolfinx/fem/FiniteElement.cpp
@@ -403,8 +403,25 @@ FiniteElement::get_dof_permutation_function(bool inverse,
     }
     else if (!scalar_element)
     {
-      throw std::runtime_error(
-          "Permuting DOFs for vector elements not implemented.");
+      // Vector element
+      std::function<void(const xtl::span<std::int32_t>&, std::uint32_t)>
+          sub_element_function
+          = _sub_elements[0]->get_dof_permutation_function(inverse);
+      int dim = _sub_elements[0]->space_dimension();
+      int bs = _bs;
+      return [dim, sub_element_function, bs,
+              subdofs](const xtl::span<std::int32_t>& doflist,
+                       std::uint32_t cell_permutation) {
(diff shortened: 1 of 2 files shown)


cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -403,8 +403,25 @@ FiniteElement::get_dof_permutation_function(bool inverse,
...
+      // Vector element
+      std::function<void(const xtl::span<std::int32_t>&, std::uint32_t)>
+          sub_element_function
+          = _sub_elements[0]->get_dof_permutation_function(inverse);
+      int dim = _sub_elements[0]->space_dimension();
+      int bs = _bs;
+      return [dim, sub_element_function,
+              bs](const xtl::span<std::int32_t>& doflist,
+                  std::uint32_t cell_permutation) {
+        std::vector<std::int32_t> subdofs(dim);
```
garth-wells, line 415: Does this give a dynamic memory allocation every time the function is called?

mscroggs: Yes, but it's difficult to get rid of unless we get Basix to handle block size when doing permutations

garth-wells: How about 'capturing' the vector like we do for the lambda that holds PETSc matrices?

garth-wells (approved):

PR #1903, opened 2021-12-17
Add function to update ghosts using a list of destinations

IgorBaratta:
This is part 1 of a series of PR (2 or 3)

In this PR we add a function update the ghosts given the destination of owned cells.
User can remove or add new ghosts by updating the destination of the owned cells.

Also we have added a function to compute the interface facets between adjacent processes.
This function can be used to apply interface conditions in additive schwarz methods, or to compute a new layer of ghosts (will be used in the next PR).


cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -442,6 +442,59 @@ std::vector<bool> mesh::compute_boundary_facets(const Topology& topology)
...
+
+  const std::vector<int32_t>& fwd_shared_facets
+      = facets->scatter_fwd_indices().array();
+
+  std::int32_t num_facets = facets->size_local() + facets->num_ghosts();
+  std::vector<bool> _interface_facets(num_facets, false);
+
+  // If a shared facet is connected to a single cell it's a candidate to be on
+  // the interface between adjacent subdomains
+  for (const std::int32_t& f : fwd_shared_facets)
```
garth-wells, line 464: No need for a reference here.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -442,6 +442,59 @@ std::vector<bool> mesh::compute_boundary_facets(const Topology& topology)
...
+  // the interface between adjacent subdomains
+  for (std::int32_t f = facets->size_local(); f < num_facets; ++f)
+    _interface_facets[f] = fc->num_links(f) == 1;
+
+  // Remove local exterior domain facets from candidates
+  // Fixme: This is problably unnecessary and going to be removed.
+  std::vector<bool> boundary_facet = mesh::compute_boundary_facets(topology);
+  std::transform(boundary_facet.begin(), boundary_facet.end(),
+                 _interface_facets.begin(), _interface_facets.begin(),
+                 [](const auto& f0, const auto& f1)
```
garth-wells, line 477: No need for `const` or `&`.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -442,6 +442,59 @@ std::vector<bool> mesh::compute_boundary_facets(const Topology& topology)
...
+
+  // If a ghost facet is connected to a single cell it's a candidate to be on
+  // the interface between adjacent subdomains
+  for (std::int32_t f = facets->size_local(); f < num_facets; ++f)
+    _interface_facets[f] = fc->num_links(f) == 1;
+
+  // Remove local exterior domain facets from candidates
+  // Fixme: This is problably unnecessary and going to be removed.
+  std::vector<bool> boundary_facet = mesh::compute_boundary_facets(topology);
+  std::transform(boundary_facet.begin(), boundary_facet.end(),
```
garth-wells, line 475: Use `cbegin/cend` if you want to show 'const-ness'.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -442,6 +442,59 @@ std::vector<bool> mesh::compute_boundary_facets(const Topology& topology)
...
+                 _interface_facets.begin(), _interface_facets.begin(),
+                 [](const auto& f0, const auto& f1)
+                 { return f0 ? false : f1; });
+
+  // Remove remote exterior domain facets from candidates
+  // Note: std::vector<bool> cannot be converted to span<bool>, using
+  // std::int8_t instead
+  std::vector<std::int8_t> remote_boundary_facets(facets->num_ghosts(), 0);
+  std::vector<std::int8_t> _boundary_facet(boundary_facet.size());
+  std::copy(boundary_facet.begin(), boundary_facet.end(),
```
garth-wells, line 485: Can this be replaced by
```cpp
std::vector<std::int8_t> _boundary_facet(boundary_facet.begin(), boundary_facet.end());
```
?

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -442,6 +442,59 @@ std::vector<bool> mesh::compute_boundary_facets(const Topology& topology)
...
+            _boundary_facet.begin());
+  facets->scatter_fwd<std::int8_t>(tcb::make_span(_boundary_facet),
+                                   tcb::make_span(remote_boundary_facets), 1);
+  std::transform(remote_boundary_facets.begin(), remote_boundary_facets.end(),
+                 _interface_facets.begin() + facets->size_local(),
```
garth-wells, line 490: Use `std::next` to advance the iterator.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -442,6 +442,59 @@ std::vector<bool> mesh::compute_boundary_facets(const Topology& topology)
...
+                 _interface_facets.begin() + facets->size_local(),
+                 [](const std::int8_t& f0, const bool& f1)
```
garth-wells, line 492: Remove references and `const`

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -442,6 +442,59 @@ std::vector<bool> mesh::compute_boundary_facets(const Topology& topology)
...
+                 { return f0 ? false : f1; });
```
garth-wells, line 493: This is a bit cryptic. Could you be more explicit with the check on `f0`?

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -442,6 +442,59 @@ std::vector<bool> mesh::compute_boundary_facets(const Topology& topology)
...
+  // Remove local exterior domain facets from candidates
+  // Fixme: This is problably unnecessary and going to be removed.
+  std::vector<bool> boundary_facet = mesh::compute_boundary_facets(topology);
+  std::transform(boundary_facet.begin(), boundary_facet.end(),
+                 _interface_facets.begin(), _interface_facets.begin(),
+                 [](const auto& f0, const auto& f1)
+                 { return f0 ? false : f1; });
+
+  // Remove remote exterior domain facets from candidates
+  // Note: std::vector<bool> cannot be converted to span<bool>, using
```
garth-wells, line 481: Feel free to change functions that return `std::vector<bool>` to `std::vector<std::int8_t>` if it makes code simpler. It will probably be faster too.

cpp/dolfinx/mesh/Topology.h
```diff
@@ -40,6 +40,15 @@ class Topology;
...
 std::vector<bool> compute_boundary_facets(const Topology& topology);
 
+/// Compute marker for facets that are on the interface between adjacent
+/// subdomains, i.e. are connected to only one cell and are either shared or
+/// ghosts;
+/// @param[in] topology The topology
+/// @return Vector with length equal to the number of facets (owned + ghosts) on
+///   this this process. True if the ith facet (local index) is on the
+///   interface.
+std::vector<bool> compute_interface_facets(const Topology& topology);
```
garth-wells, line 50: Maybe `partition_boundary_facet_markers` would be a better name (although it's a bit long)? Probably don't need 'compute' in the name for a free function.

cpp/dolfinx/mesh/Topology.h
```diff
@@ -40,6 +40,15 @@ class Topology;
 ///   exterior of the domain.
 std::vector<bool> compute_boundary_facets(const Topology& topology);
 
+/// Compute marker for facets that are on the interface between adjacent
+/// subdomains, i.e. are connected to only one cell and are either shared or
+/// ghosts;
+/// @param[in] topology The topology
+/// @return Vector with length equal to the number of facets (owned + ghosts) on
+///   this this process. True if the ith facet (local index) is on the
+///   interface.
```
garth-wells, line 49: 'interface' -> 'partition boundary'

cpp/dolfinx/mesh/utils.cpp
```diff
@@ -581,3 +582,63 @@ mesh::compute_incident_entities(const Mesh& mesh,
   return entities1;
 }
 //-----------------------------------------------------------------------------
+mesh::Mesh mesh::update_ghosts(const mesh::Mesh& mesh,
+                               graph::AdjacencyList<std::int32_t>& dest)
```
garth-wells, line 586: Should this be `const`?

cpp/dolfinx/mesh/utils.h
```diff
@@ -153,4 +153,12 @@ compute_incident_entities(const Mesh& mesh,
                           const xtl::span<const std::int32_t>& entities, int d0,
                           int d1);
 
+/// Update mesh with a new set of ghosts.
+// Ghost can be removed or added according to the new ghost list.
+/// @param[in] mesh The mesh
+/// @param[in] dest Destination rank for mesh cells owned by this rank
+/// @return Mesh with new ghosts
+mesh::Mesh update_ghosts(const mesh::Mesh& mesh,
```
garth-wells, line 161: This function name doesn't see right to me. It's not updating a mesh, but creating a new one.

I'm also not sure why one would use the function. Should it really work on a Topology object and return a new Topology object that can then me used to build a Mesh?

garth-wells (changes requested):

cpp/dolfinx/mesh/utils.h
IgorBaratta: At first I thought about modifying the mesh object, but the topology  has some states (connectivities, entities) that would be invalidated.  
I'll modify (simplify) the function to return a new topology, which could be used later for constructing a new mesh.

PR #1904, opened 2021-12-18
Consolidate PETSc linear algebra wrapper into single h/cpp files

garth-wells:
The merges all the classes functions in the namespace `dolfinx::la::petsc` into `petsc.h/cpp`.

The reduces the number of files helps avoid leakage of PETSc into other parts of the code. Much of the PETSc helper functionality had become (intentionally) very light, with users using petsc4py from Python and users using the 'native' PETSc interface from C++. The only real purpose of the wrapper classes is now is to ease some lifetime management from the C++ side.


jhale:
Looks good. 

Couple of comments:

* nls::NewtonSolver is totally reliant on PETSc. Should it go in a namespace petsc?
* The norm computation functions seem lightweight/unnecessary if users should use PETSc C interface for most tasks.

garth-wells:
> Looks good.
> 
> Couple of comments:
> 
> * nls::NewtonSolver is totally reliant on PETSc. Should it go in a namespace petsc?

I thought about this, but I think the `NetwonSolver` solver design needs to be looked at carefully and ideally be made independent of the backend. I think we can leave it where it is for now.

> * The norm computation functions seem lightweight/unnecessary if users should use PETSc C interface for most tasks.

Yes, these could be removed.

jhale (approved):

PR #1905, opened 2021-12-20
Allow BCs to be set with literals and arrays

garth-wells:


jorgensd:
Should we add a demo highlighting when to use a Constant as opposed to array's (i.e. time dependent, but constant in space Dirichlet conditions).

garth-wells:
> Should we add a demo highlighting when to use a Constant as opposed to array's (i.e. time dependent, but constant in space Dirichlet conditions).

Can do, but let's use a separate PR. The use case is covered by Newton solver unit test.

PR #1909, opened 2021-12-27
Dynamically create Form classes to inherit from appropriate typed cpp base

### LINKED ISSUE #1209: Non-optimal performance of repeated processing or UFL forms in assembly
I encountered a non-optimal performance of `dolfinx.fem.assemble_*` routines on HPC infrastructures recently. I prepared the following test with some explanatory description in the contained README file: https://github.com/mrehor/dolfinx-hpc-assembly/

The thing is that there are some redundant calls of `_create_cpp_form` in [assemble.py](https://github.com/FEniCS/dolfinx/blob/c5f975ba76d047e09298021163dfb21a93a54b0f/python/dolfinx/fem/assemble.py) (e.g. L174 and L180) which seem to cause a potential I/O overhead.  

**I understand that this is technically not a bug as it is still possible to do things efficiently with the current code.** I wanted to address this HPC-related issue with the suggestion to at least expose `dolfinx.fem.assemble._create_cpp_form` within API or to rewrite assembly-related routines on the Python side to accept only forms of `cpp.fem.Form` type. Not sure if the latter option isn't too "drastic". Any thoughts?

garth-wells:
This change creates `fem.Form` instances with the correct `fem::Form<T>` base. This removes the need to access '`_cpp_object`'. Forms can be passed directly into pybind11 wrapped functions that expect a `fem::Form<T>`.

This PR also requires that forms passed into lower-level functions, e.g. assembly, be explicitly instantiated. This fixes #1209 and is consistent with the DOLFINx design principle of favouring explicit over implicit (especially in lower level interfaces). Being explicit highlighted some cases in the test where the logic was not transparent.

Previously one could do:
```cpp
...
from dolfinx.fem import assemble_matrix

a = inner(u, v) * dx
A = assemble_matrix(a)
```
this change requires:
```cpp
...
from dolfinx.fem import assemble_matrix, form

a = form(inner(u, v) * dx)
A = assemble_matrix(a)
```
Subsequent calls to `assemble_matrix(a)` will no longer involve the overhead of UFL and FFCx processing of the form, Even if the form is in a cache, the cost of pre-processing can in cases be noticeable.

Moreover, this change is consistent with forthcoming changes for `Expression` that will deliberately not hide the behind-the-scenes UFL/FFCx processing.

Higher level interfaces, e.g. `fem.LinearProblem`, are not affected.


python/demo/elasticity/demo_elasticity.py
```diff
@@ -101,11 +101,11 @@ def sigma(v):
...
 u = TrialFunction(V)
 v = TestFunction(V)
-a = inner(sigma(u), grad(v)) * dx
-L = inner(f, v) * dx
+a = form(inner(sigma(u), grad(v)) * dx)
+L = form(inner(f, v) * dx)
 
 # Set up boundary condition on inner surface
-bc = DirichletBC(np.array([0, 0, 0], dtype=PETSc.ScalarType),
+bc = dirichletbc(np.array([0, 0, 0], dtype=PETSc.ScalarType),
```
jorgensd, line 108: Should this be `Dirichletbc` as Dirichlet is a name?

python/dolfinx/fem/__init__.py
```diff
@@ -13,6 +13,15 @@
 from dolfinx.cpp.fem import transpose_dofmap  # noqa
 from dolfinx.cpp.fem import IntegralType
 from dolfinx.cpp.fem import create_sparsity_pattern as _create_sparsity_pattern
+from dolfinx.fem.dirichletbc import (DirichletBC, dirichletbc, bcs_by_block,
```
jorgensd, line 16: Should we remove `DirichletBC` and `Form` from import, as they cannot be called without
```python
     formcls = type("DirichletBC", (DirichletBC, bctype), {})
```
and 
```python
     formcls = type("Form", (Form, ftype), {})
```
as the class Form and DirichletBC has no initial base class to inherit from, and call `super()` with.

python/demo/elasticity/demo_elasticity.py
garth-wells: No - we use lower case for functions and reserve upper case for classes.

python/dolfinx/fem/__init__.py
garth-wells: I'm not sure - we still use it with type hints and we want to document the methods.

The analogy to NumPy is `ndarray`. `ndarray` is the 'main' NumPy class, but is created used factory functions.

jhale: It would be good to signpost to users that these classes are now special metaclasses, either through the name e.g. appending Meta, or improve the docstrings.

python/dolfinx/fem/forms.py
```diff
@@ -80,44 +79,62 @@ def __init__(self, form: ufl.Form, dtype: np.dtype = PETSc.ScalarType,
...
     def ufc_form(self):
-        """Return the compiled ufc_form object"""
+        """The compiled ufc_form object"""
         return self._ufc_form
 
     @property
     def code(self):
-        """Return C code strings"""
+        """C code strings"""
         return self._code
```
jhale, line 98: Can also take collection of forms, as well as form.

python/dolfinx/fem/__init__.py
```diff
@@ -53,6 +53,6 @@ def create_sparsity_pattern(a: typing.Union[_FormComplex, _FormReal]):
     "assemble_vector_block", "assemble_vector_nest",
     "assemble_matrix_block", "assemble_matrix_nest",
     "assemble_matrix", "set_bc", "set_bc_nest",
-    "DirichletBC", "bcs_by_block", "DofMap", "Form", "IntegralType",
```
jhale, line 56: Using a verb e.g. create_form would make the purpose of the function clearer at the expense of brevity in the code.

python/dolfinx/fem/forms.py
```diff
@@ -19,9 +19,8 @@
 
 
 class Form:
-    def __init__(self, form: ufl.Form, dtype: np.dtype = PETSc.ScalarType,
-                 form_compiler_parameters: dict = {}, jit_parameters: dict = {}):
-        """Create DOLFINx Form
```
jhale, line 24: Why remove typing on parameters?

python/demo/mixed-elasticity-sc/static-condensation-elasticity.py
```diff
@@ -20,8 +20,8 @@
 import ufl
 from dolfinx import geometry
 from dolfinx.cpp.fem import Form_complex128, Form_float64
```
jhale, line 22: Is it possible to remove this and use the factory function?

python/dolfinx/fem/forms.py
jhale: Can also take collrction of forms

python/demo/mixed-elasticity-sc/static-condensation-elasticity.py
garth-wells: Maybe, but it's a bit separate form this change. Might be possible to so something neater soon.

python/dolfinx/fem/forms.py
garth-wells: Fixed (partially, some types depend on the metaclass type).

PR #1916, opened 2022-01-06
Fix function copy option

### LINKED ISSUE #1912: dolfinx.fem.Function.copy raises AttributeError
Minimal example
```
import dolfinx.fem
import dolfinx.mesh
import mpi4py
import mpi4py.MPI

mesh = dolfinx.mesh.create_unit_square(mpi4py.MPI.COMM_WORLD, 2, 2)
V = dolfinx.fem.FunctionSpace(mesh, ("Lagrange", 1))
u = dolfinx.fem.Function(V)
u.copy()
```
fails with

```
  File "dolfinx/fem/function.py", line 345, in copy
    self._cpp_object.vector.copy())
AttributeError: 'dolfinx.cpp.fem.Function_float64' object has no attribute 'vector'
```

Coverage of such method seems to be missing from `python/test/unit/fem/test_function.py`.

jorgensd:
Resolves #1912

### DIFF
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -348,8 +348,7 @@ class Function(ufl.Coefficient):
         degree-of-freedom vector is copied.
 
         """
-        return Function(self.function_space,
-                        self._cpp_object.vector.copy())
+        return Function(self.function_space, type(self.x)(self.x))
 
     @property
     def vector(self):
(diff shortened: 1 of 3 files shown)


garth-wells (changes requested):
Looks good. Could we add a test that would have picked up the issue?

francesco-ballarin (approved):

jorgensd:
> Looks good. Could we add a test that would have picked up the issue?

Test added

garth-wells (approved):

PR #1917, opened 2022-01-07
Fix Python wrappers when SCOTCH not installed

jhale:


python/dolfinx/wrappers/graph.cpp
```diff
@@ -103,6 +103,7 @@ void graph(py::module& m)
           MPICommWrapper, int,
           const dolfinx::graph::AdjacencyList<std::int64_t>&, std::int32_t,
           bool)>;
+#ifdef HAS_PTSCOTCH
```
garth-wells, line 106: The logic here looks wrong to me. It adds multiple functions called `partition`. I think there need to be some `#else` statements.

garth-wells (changes requested):

python/dolfinx/wrappers/graph.cpp
jhale: Chris and I are looking at it now. The question is, for a given combination of included partitioners, what should the default partitioner be?

garth-wells: Inherit it from the C++ interface - see
- https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/graph/partition.h#L47
- https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/graph/partition.cpp#L24

garth-wells (approved):

PR #1925, opened 2022-01-11
Simplify `ufcx.h` detection

garth-wells:


cpp/cmake/modules/FindUFCx.cmake
```diff
@@ -0,0 +1,66 @@
+# - Try to find UFCx
+#
+# Once done this will define
```
jhale, line 3: Missing, good place to explain the logic of where it looks?

python/doc/source/installation.rst
```diff
@@ -26,7 +26,7 @@ C++ core
 - xtensor (https://xtensor.readthedocs.io/)
 - pkg-config (https://www.freedesktop.org/wiki/Software/pkg-config/)
 - Python 3 [build dependency]
-- FFCx [build dependency, for ``ufc.h`` header]
+- FFCx [build dependency, for ``ufcx.h`` header]
```
jhale, line 29: Still true?

garth-wells: Fixed.

Issue #1612, opened 2021-07-09
complex number tests generate zero solution with PETSc 3.12 (ubuntu focal 20.04 PPA build)


drew-parsons:
A problem was reported at https://fenicsproject.discourse.group/t/helmholtz-complex-xdmf-contains-zeros-only/6204 for dolfinx 0.1.0 in complex number mode. The problem is with the PPA build for ubuntu focal 20.04. The helmholtz2D demo (demo_helmholtz_2d.py) generates a null solution (0 in both real and imaginary parts).  I can reproduce the null result on a fresh cloud installation of focal.

The "equivalent" build on Debian unstable does not have the problem: it generates a nontrivial periodically repeating solution for the helmholtz2D demo.

The PPA build log is https://launchpadlibrarian.net/546658626/buildlog_ubuntu-focal-amd64.fenics-dolfinx_1%3A0.1.0-2~ppa1~focal3_BUILDING.txt.gz
I haven't spotted any obvious problems there though.

focal provides PETSc 3.12 while my working system (Debian unstable+experimental) provides PETSc 3.15.  Do we know the circumstances in which dolfinx complex might appear to run successfully, but generate a trivial zero solution?  Are there any known problems with PETSc 3.12 that might explain the zero result?

drew-parsons:
I can reproduce this error with a fresh build of PETSc 3.12 on Debian unstable (where PETSc 3.14 or 3.15 is otherwise used).

That supports the hypothesis that PETSc 3.12 is the problem here, so it should be possible to reproduce in docker builds using petsc 3.12.

Updating the title accordingly.

francesco-ballarin:
@drew-parsons is this still happening with dolfinx 0.7.0 and PETSc 3.20?

drew-parsons:
PETSc 3.12 fails on debian unstable:
```
/projects/petsc/build/3.12/petsc-3.12.4+dfsg1/src/mat/impls/hypre/mhypre.c: In function ‘MatCreateFromParCSR’:
/projects/petsc/build/3.12/petsc-3.12.4+dfsg1/src/mat/impls/hypre/mhypre.c:1506:43: error: assignment to expression with array type
 1506 |       hypre_ParCSRMatrixColStarts(parcsr) = (HYPRE_BigInt*)(map->range + myid);
      |                                           ^
[...]

PR #1927, opened 2022-01-12
Major simplification of PETSc and SLEPc detection

garth-wells:
Use CMake to automatically process the PETSc and SLEPc pkg-config files and generate targets.


francesco-ballarin:
Heads up to @drew-parsons that this PR raises the minimum required PETSc/SLEPc to 3.15, while current versions on impish are 3.14, so that he is aware of this before his next build of the PPA.

garth-wells:
We could maybe drop to 3.14, but it's completely untested.

drew-parsons:
Thanks for the notice.  The next oldest supported Ubuntu release is focal with PETSc 3.12.  Do we know definitely that current dolfinx (if adjusted against the 3.15 detection in this PR) will not build against PETSc 3.12?

As for testing PETSc 3.14, we can consider the impish build to be the test. If it fails, we can revert the PPA and leave impish on dolfinx 0.3.0.

cpp/CMakeLists.txt
```diff
@@ -168,13 +169,29 @@ if("XTENSOR_USE_XSIMD" IN_LIST BASIX_DEFN)
...
 find_package(xtensor 0.23.10 REQUIRED)
 set_package_properties(xtensor PROPERTIES TYPE REQUIRED
   DESCRIPTION "C++ library for numerical analysis with multi-dimensional array expressions."
   URL "https://xtensor.readthedocs.io/")
 
-find_package(PETSc 3.10)
+# Check for PETSc
+find_package(PkgConfig REQUIRED)
+set(ENV{PKG_CONFIG_PATH} "$ENV{PETSC_DIR}/$ENV{PETSC_ARCH}/lib/pkgconfig:$ENV{PETSC_DIR}/lib/pkgconfig:$ENV{PKG_CONFIG_PATH}")
+pkg_search_module(PETSC REQUIRED IMPORTED_TARGET PETSc>=3.15 petsc>=3.15)
```
drew-parsons, line 181: Closes #1612

Issue #1926, opened 2022-01-12
Expression failing for N1Curl -> vector-DG1 due to zero tables.


jorgensd:

michalhabera:
Ah, I see where the problem is. Zero tables should not make it to codegen phase, because when IR (expression tree) is built it removes trivial (zero) branches here https://github.com/FEniCS/ffcx/blob/9846c2e67e51672545f08e77ed597fed26e38855/ffcx/ir/integral.py#L99

However, it does that only if the Expression has single root node (is only scalar valued). That line needs changing so that for each `S_targets` the new subgraph is rebuild.

PR #1928, opened 2022-01-12
Zero table elimination for non-scalar graph

jorgensd:
Testing resolution for  #1926.
https://github.com/FEniCS/ffcx/pull/438 has to be merged first

### DIFF
--- a/python/test/unit/fem/test_expression.py
+++ b/python/test/unit/fem/test_expression.py
@@ -309,3 +309,69 @@ def test_expression():
     for i, cell in enumerate(closest_cell):
         point = points[local_map[i]]
         assert np.allclose(grad_u_at_x[i, i], gradf(point))
+
+
+def test_zero_elimination():
+    """ Testing that ffcx eliminates zeros correctly when compiling expressions"""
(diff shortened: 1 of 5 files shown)


python/test/unit/fem/test_expression.py
```diff
@@ -309,3 +309,72 @@ def gradf(x):
     for i, cell in enumerate(closest_cell):
         point = points[local_map[i]]
         assert np.allclose(grad_u_at_x[i, i], gradf(point))
+
+
+def test_zero_elimination():
```
garth-wells, line 314: Looks like this test should be in FFCx.

jorgensd: I'm not certain on how to easily check this in FFCx (as I am not too familiar with the expression interface there). Maybe @michalhabera has an idea of how to nicely test this.

jorgensd: I've now moved the scalar and vector test to FFCx. However, I am reluctant to move the tensor test, as it is rather involved (using Nedelec elements, making creation of the coefficient array rather involved).

jorgensd:
Tests added to FFCx instead

PR #1930, opened 2022-01-12
Add native dolfinx sparse matrix

chrisrichardson:
Provides a distributed sparse matrix using CSR storage
* Templated to support multiple types (double, float, complex, etc)
* Possible to interface to external packages
* Future possible memory allocation to devices
* Useful for experimenting on linear algebra outside petsc


cpp/dolfinx/la/Matrix.h
```diff
@@ -0,0 +1,372 @@
...
+#include <vector>
+#include <xtensor/xtensor.hpp>
+#include <xtl/xspan.hpp>
+
+namespace dolfinx::la
+{
+
+/// Distributed sparse Matrix
+template <typename T, class Allocator = std::allocator<T>>
+class Matrix
```
garth-wells, line 20: I think we should name this `MatrixCSR` or simlar.

cpp/dolfinx/la/Matrix.h
```diff
@@ -0,0 +1,372 @@
...
+
+  /// Create a distributed matrix
+  Matrix(const SparsityPattern& p, const Allocator& alloc = Allocator())
+      : _index_maps({p.index_map(0), p.column_index_map()}),
+        _data(p.num_nonzeros(), 0, alloc), _cols(p.num_nonzeros()),
+        _row_ptr(
+            _index_maps[0]->size_local() + _index_maps[0]->num_ghosts() + 1, 0)
+  {
+    // Store sparsity
+    const graph::AdjacencyList<std::int32_t>& pg = p.graph();
```
garth-wells, line 35: This can be done in the initialiser list.

cpp/dolfinx/la/Matrix.h
```diff
@@ -0,0 +1,372 @@
...
+    const std::int32_t local_size0 = _index_maps[0]->size_local();
+    const std::array local_range0 = _index_maps[0]->local_range();
+    const std::int32_t num_ghosts0 = _index_maps[0]->num_ghosts();
+    const std::vector<std::int64_t>& ghosts0 = _index_maps[0]->ghosts();
+    const std::vector<int> ghost_owners0 = _index_maps[0]->ghost_owner_rank();
+    const std::int32_t local_size1 = _index_maps[1]->size_local();
+    const std::array local_range1 = _index_maps[1]->local_range();
+    const std::vector<std::int64_t>& ghosts1 = _index_maps[1]->ghosts();
+
+    // Get ghost->owner communicator for rows
```
garth-wells, line 55: The below lines aren't getting a communicator, despite comment.

cpp/dolfinx/la/Matrix.h
```diff
@@ -0,0 +1,372 @@
...
+    const auto dest_ranks = dolfinx::MPI::neighbors(_nbr_comm)[1];
+    const int num_neighbors = dest_ranks.size();
+
+    // Global-to-neigbourhood map for destination ranks
```
garth-wells, line 59: Comment doesn't make clear what the map is. 'global-to-local ranks for neighbourhood'?

cpp/dolfinx/la/Matrix.h
```diff
@@ -0,0 +1,372 @@
...
+    std::map<int, std::int32_t> dest_proc_to_neighbor;
+    for (std::size_t i = 0; i < dest_ranks.size(); ++i)
+      dest_proc_to_neighbor.insert({dest_ranks[i], i});
+
+    // Get ownership of each ghost row by neighbor
+    _ghost_row_to_neighbor_rank.resize(num_ghosts0, -1);
+    for (int i = 0; i < num_ghosts0; ++i)
+    {
+      // Find rank on neigbourhood comm of ghost owner
```
garth-wells, line 68: // Find rank on neigbourhood comm of ghost row owner

garth-wells (changes requested):
Looking promising. I think the code comments need to be check for clarity. The implementation is inevitable technical we need the comments to be really sharp.

cpp/dolfinx/la/Matrix.h
```diff
@@ -0,0 +1,375 @@
...
+    {
+      for (int j = _row_ptr[r]; j < _row_ptr[r + 1]; ++j)
+        A(r, _cols[j]) = _data[j];
+    }
+
+    return A;
+  }
+
+  /// Copy cached ghost values to row owner and add.
+  void finalize()
```
garth-wells, line 296: This need to be broken into two non-blocking functions (start/end).

chrisrichardson (commented):
probably should just delete this function

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,415 @@
...
+  // // Block size
+  std::array<int, 2> _bs;
+
+  // Data
+  std::vector<T, Allocator> _data;
+  std::vector<std::int32_t> _cols, _row_ptr;
+
+  // Precomputed data for finalize/update
+  // Neighborhood communicator
+  MPI_Comm _neighbor_comm;
```
garth-wells, line 401: This should be a `dolfinx::MPI::Comm`. At the moment there is a leak because the communicator isn't being freed.

IgorBaratta: Currently we have an issue with `MPI_Comm_dup` and MPICH for unweighted neighbor communicators.
Possible solution add weights to graphs, see #1944 .

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -162,8 +162,8 @@ class MatrixCSR
   /// Set all non-zero local entries to a value
   void set(T x)
   {
-    std::int32_t local_size[0] = _index_maps[0]->size_local();
-    std::fill(_data.begin(), std::next(_data.begin(), _row_ptr[local_size[0]]),
+    std::int32_t local_size0 = _index_maps[0]->size_local();
+    std::fill(_data.begin(), std::next(_data.begin(), _row_ptr[local_size0]),
```
IgorBaratta, line 166:
```suggestion
    std::fill_n(_data.begin(), local_size0,
```

cpp/test/unit/vector.cpp
```diff
@@ -55,9 +61,72 @@ void test_vector()
   CHECK(v.norm(la::Norm::linf) == static_cast<PetscScalar>(mpi_size - 1));
 }
 
+void test_matrix()
```
chrisrichardson, line 64: Move this out of `vector.cpp`

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,490 @@
...
+      }
+      else
+        throw std::runtime_error("Local row out of range");
+    }
+  }
+
+  /// Get the number of local rows
+  /// @param[in] ghost_rows Set to true to include ghost rows in the
+  /// number of local rows
+  std::int32_t rows(bool ghost_rows = false) const
```
chrisrichardson, line 311: Maybe should be `num_rows()` (compare with `cols()` which returns something quite different).

cpp/dolfinx/la/SparsityPattern.cpp
```diff
@@ -224,11 +198,48 @@ void SparsityPattern::insert_diagonal(const std::vector<int32_t>& rows)
...
+common::IndexMap SparsityPattern::column_index_map() const
+{
+  if (!_graph)
+    throw std::runtime_error("Sparsity pattern has not been finalised.");
+
+  std::array range = _index_maps[1]->local_range();
+  const std::int32_t local_size = range[1] - range[0];
+  return common::IndexMap(
+      _comm.comm(), local_size,
+      dolfinx::MPI::compute_graph_edges(
```
garth-wells, line 231: @chrisrichardson `compute_graph_edges` should be avoided whenever possible. It involves global all-to-all communication. We must have computed the required information already during construction?

chrisrichardson: I am very happy to look at this in a future PR (soon).

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+      std::size_t d = std::distance(_cols.begin(), cit);
+      _unpack_pos.push_back(d);
+    }
+  }
+
+  /// Move constructor
+  /// @todo Check handling of MPI_Request
+  MatrixCSR(MatrixCSR&& A) = default;
+
+  /// Set all non-zero local entries to a value
```
garth-wells, line 220: Should it also set ghost rows?

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+  /// @param[in] x The value to set non-zero matrix entries to
+  /// @todo This should probably also set ghost rows
+  void set(T x) { std::fill_n(_data.begin(), _index_maps[0]->size_local(), x); }
+
+  /// Set values in the matrix
```
garth-wells, line 225: Expand docstring on calling `finalise` before/after, etc.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+          std::size_t d = std::distance(_cols.begin(), it);
+          _data[d] = xr[c];
+        }
+      }
+      else
+        throw std::runtime_error("Local row out of range");
+    }
+  }
+
+  /// Accumulate values in the matrix
```
garth-wells, line 267: Expand docstring on calling finalise before/after, etc.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+  void add(const xtl::span<const T>& x,
+           const xtl::span<const std::int32_t>& rows,
+           const xtl::span<const std::int32_t>& cols)
+  {
+    assert(x.size() == rows.size() * cols.size());
+    const std::int32_t local_size0 = _index_maps[0]->size_local();
+    const std::int32_t num_ghosts0 = _index_maps[0]->num_ghosts();
+    for (std::size_t r = 0; r < rows.size(); ++r)
+    {
+      // Columns indices for row
```
garth-wells, line 284: Comment looks wrong.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+      std::int32_t row = rows[r];
+
+      // Current data row
+      const T* xr = x.data() + r * cols.size();
+
+      if (row < local_size0 + num_ghosts0)
```
garth-wells, line 290: This confusing. No need for a if/else - just put the check and exception at the top.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+      }
+      else
+        throw std::runtime_error("Local row out of range");
+    }
+  }
+
+  /// Get the number of local rows
+  /// @param[in] ghost_rows Set to true to include ghost rows in the
+  /// number of local rows
+  std::int32_t num_rows(bool ghost_rows = false) const
```
garth-wells, line 312: Default args can lead to bugs. I'd suggest using two functions.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+          std::size_t d = std::distance(_cols.begin(), it);
+          _data[d] += xr[c];
+        }
+      }
+      else
+        throw std::runtime_error("Local row out of range");
+    }
+  }
+
+  /// Get the number of local rows
```
garth-wells, line 309: This is not very clear, esp. with the `ghost_rows` parameter.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+    const std::int32_t ncols
+        = _index_maps[1]->size_local() + _index_maps[1]->num_ghosts();
+    xt::xtensor<T, 2> A = xt::zeros<T>({nrows, ncols});
+    for (std::size_t r = 0; r < nrows; ++r)
+      for (int j = _row_ptr[r]; j < _row_ptr[r + 1]; ++j)
+        A(r, _cols[j]) = _data[j];
+    return A;
+  }
+
+  /// Communicate ghost row data to the owning ranks
```
garth-wells, line 335: But what's the operation? Accumulate?

Expand doc string to explain relationship to `finalize_begin`/`finalize_end`.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+  void finalize()
+  {
+    finalize_begin();
+    finalize_end();
+  }
+
+  /// Begin communication of ghost row data to owning ranks
```
garth-wells, line 342: What does it do with the communicated data?

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+  /// @note Calls to this function must be followed by
+  /// MatrixCSR::finalize_end(). Between the two calls matrix values
+  /// must not be changed.
+  void finalize_begin()
+  {
+    const std::int32_t local_size0 = _index_maps[0]->size_local();
+    const std::int32_t num_ghosts0 = _index_maps[0]->num_ghosts();
+
+    // For each ghost row, pack and send values to send to neighborhood
+    std::vector<int> insert_pos(_val_send_disp);
```
garth-wells, line 352: I think the memory allocations should be moved to the constructor. The advantage of requiring an exact sparsity pattern is that we can avoid dynamic memory allocations after initial construction.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+    std::vector<T> ghost_value_data(_val_send_disp.back());
+    for (int i = 0; i < num_ghosts0; ++i)
+    {
+      const int neighbor_rank = _ghost_row_to_neighbor_rank[i];
+
+      // Get position in send buffer
```
garth-wells, line 358: Position of what?

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+  /// @return Row (0) and column (1) index maps
+  const std::array<std::shared_ptr<const common::IndexMap>, 2>&
+  index_maps() const
+  {
+    return _index_maps;
+  }
+
+  /// Get local data values
+  /// @param[in] ghost_rows Set to true to include data of ghost rows
+  xtl::span<T> values(bool ghost_rows = false)
```
garth-wells, line 417: This is a low-level function so I think we can let the caller deal with ghosts by having a function that returns he pointer to the owned values.

Best to avoid default arguments whenever reasonably possible.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+  {
+    const std::int32_t nrows = num_rows(ghost_rows);
+    return xtl::span<T>(_data.data(), _row_ptr.at(nrows));
+  }
+
+  /// Get local values (const version)
+  /// @param[in] ghost_rows Set to true to include data of ghost rows
+  xtl::span<const T> values(bool ghost_rows = false) const
```
garth-wells, line 425: See other comments on ghost parameter.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+  {
+    const std::int32_t nrows = num_rows(ghost_rows);
+    return xtl::span<const T>(_data.data(), _row_ptr.at(nrows));
+  }
+
+  /// Get local row pointers
+  /// @param[in] ghost_rows Set to true to include data of ghost rows
+  xtl::span<const std::int32_t> row_ptr(bool ghost_rows = false) const
```
garth-wells, line 433: See other comments on ghost parameter.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -0,0 +1,491 @@
...
+
+  /// Get local column indices
+  /// @param[in] ghost_rows Set to true to include data of ghost rows
+  xtl::span<const std::int32_t> cols(bool ghost_rows = false) const
+  {
+    const std::int32_t nrows = num_rows(ghost_rows);
+    return xtl::span<const std::int32_t>(_cols.data(), _row_ptr.at(nrows));
+  }
+
+  /// Get start of off-diagonal (unowned columns) on each row
```
garth-wells, line 447: Expand explanation. Which array is it relevant to?

cpp/dolfinx/la/SparsityPattern.cpp
```diff
@@ -24,8 +24,7 @@ SparsityPattern::SparsityPattern(
     : _comm(comm), _index_maps(maps), _bs(bs)
 {
   assert(maps[0]);
-  _cache_owned.resize(maps[0]->size_local());
-  _cache_unowned.resize(maps[0]->num_ghosts());
+  _row_cache.resize(maps[0]->size_local() + maps[0]->num_ghosts());
```
garth-wells, line 27: Move to initialiser list.

cpp/dolfinx/la/SparsityPattern.cpp
```diff
@@ -77,8 +76,9 @@ SparsityPattern::SparsityPattern(
           comm, std::set<int>(ghost_owners1.begin(), ghost_owners1.end())),
       ghosts1, ghost_owners1);
 
-  _cache_owned.resize(_index_maps[0]->size_local());
-  _cache_unowned.resize(_index_maps[0]->num_ghosts());
+  _row_cache.resize(_index_maps[0]->size_local()
```
garth-wells, line 79: Move to initialiser list.

cpp/dolfinx/la/SparsityPattern.cpp
```diff
@@ -182,14 +162,9 @@ void SparsityPattern::insert(const xtl::span<const std::int32_t>& rows,
...
-    if (row < local_size0)
-      _cache_owned[row].insert(_cache_owned[row].end(), cols.begin(),
-                               cols.end());
-    else if (row < size0)
-    {
-      _cache_unowned[row - local_size0].insert(
-          _cache_unowned[row - local_size0].end(), cols.begin(), cols.end());
-    }
+    assert(row >= 0);
+    if (row < size0)
```
garth-wells, line 166: Simplify to a single if/exception.

cpp/dolfinx/la/SparsityPattern.cpp
```diff
@@ -224,11 +198,48 @@ void SparsityPattern::insert_diagonal(const std::vector<int32_t>& rows)
...
+  const std::int32_t local_size = range[1] - range[0];
+  const std::int32_t num_ghosts = _col_ghosts.size();
+  std::vector<std::int64_t> global(local_size + num_ghosts);
+  std::iota(global.begin(), std::next(global.begin(), local_size), range[0]);
+  std::copy(_col_ghosts.begin(), _col_ghosts.end(),
+            global.begin() + local_size);
+  return global;
+}
+//-----------------------------------------------------------------------------
+common::IndexMap SparsityPattern::column_index_map() const
```
garth-wells, line 222: Should this be computed at construction rather than re-computed at each call?

cpp/dolfinx/la/SparsityPattern.h
```diff
@@ -69,46 +70,54 @@ class SparsityPattern
...
   SparsityPattern& operator=(SparsityPattern&& pattern) = default;
 
+  /// Insert non-zero locations using local (process-wise) indices
+  void insert(const xtl::span<const std::int32_t>& rows,
+              const xtl::span<const std::int32_t>& cols);
+
+  /// Insert non-zero locations on the diagonal
+  /// @param[in] rows The rows in local (process-wise) indices. The
+  /// indices must exist in the row IndexMap.
+  void insert_diagonal(const std::vector<std::int32_t>& rows);
```
garth-wells, line 80: Use `span` rather than `vector`.

cpp/dolfinx/la/SparsityPattern.h
```diff
@@ -69,46 +70,54 @@ class SparsityPattern
...
   ///
-  /// @return The global index non-zero columns on this process, including
-  /// ghosts
+  /// @note The ghosts are computed only once SparsityPattern::assemble
+  /// has been called
+  /// @return The global index non-zero columns on this process,
+  /// including ghosts
   std::vector<std::int64_t> column_indices() const;
 
+  /// Compute IndexMap for columns after SparsityPattern assembly
```
garth-wells, line 99:
```
/// Compute IndexMap for columns in this SparsityPattern after assembly
```

cpp/dolfinx/la/SparsityPattern.h
```diff
@@ -69,46 +70,54 @@ class SparsityPattern
...
 
-  /// Insert non-zero locations using local (process-wise) indices
-  void insert(const xtl::span<const std::int32_t>& rows,
-              const xtl::span<const std::int32_t>& cols);
-
-  /// Insert non-zero locations on the diagonal
-  /// @param[in] rows The rows in local (process-wise) indices. The
-  ///   indices must exist in the row IndexMap.
-  void insert_diagonal(const std::vector<std::int32_t>& rows);
+  /// Return total number of local nonzeros
```
garth-wells, line 107: Does the sparsity pattern have ghost rows?

cpp/dolfinx/la/SparsityPattern.h
```diff
@@ -69,46 +70,54 @@ class SparsityPattern
...
+  std::int64_t num_nonzeros() const;
 
-  /// Finalize sparsity pattern and communicate off-process entries
-  void assemble();
+  /// Number of non-zeros in owned columns (diagonal block) on a given row
+  std::int32_t nnz_diag(int row) const;
```
garth-wells, line 111: Use `std::int32_t` for row parameter.

Does row need to be owned?

cpp/dolfinx/la/SparsityPattern.h
```diff
@@ -69,46 +70,54 @@ class SparsityPattern
...
 
-  /// Return number of local nonzeros
-  std::int64_t num_nonzeros() const;
+  /// Number of non-zeros in unowned columns (off-diagonal block) on a given row
+  std::int32_t nnz_off_diag(int row) const;
```
garth-wells, line 114: See above comment.

cpp/dolfinx/la/SparsityPattern.h
```diff
@@ -69,46 +70,54 @@ class SparsityPattern
...
 
-  /// Sparsity pattern for the owned (diagonal) block. Uses local
-  /// indices for the columns.
-  const graph::AdjacencyList<std::int32_t>& diagonal_pattern() const;
+  /// Sparsity pattern graph. Uses local indices for the columns.
```
garth-wells, line 116: Add comment on column IndexMap - can be used to get global column indices.

garth-wells (changes requested):

cpp/dolfinx/la/MatrixCSR.h
chrisrichardson: Yes, I agree. Does it make sense to 'set to a value' or is it always just going to be zero?

garth-wells: I think it will almost always be zero, but it doesn't introduce any complexity allowing to be some other value. I can image maybe creating an identity matrix for testing.

chrisrichardson: OK, but you need to be careful with that, if multiple entries exist on different processes for the same global location.

cpp/dolfinx/la/SparsityPattern.cpp
chrisrichardson: Not possible for this one.

cpp/dolfinx/la/SparsityPattern.cpp
chrisrichardson: It can't be computed until after all insertions.

IgorBaratta (approved):

PR #1941, opened 2022-01-16
Fix narrowing of global indices in graph builder

chrisrichardson:
* int64 bug fix


jhale:
Good spot. My view is that we should consider disabling implicit narrowing casts. SonarCloud picked up a lot of potential issues like this.

PR #1944, opened 2022-01-17
Add uniform weights to MPI neighbor communicator

IgorBaratta:
Workaround to issue https://github.com/pmodels/mpich/issues/5764 .
`MPI_Comm_dup` fails for unweighted graph communicators.

The issue has been fixed in https://github.com/pmodels/mpich/pull/5629 and it's available in mpich 4.0rc3( a release candidate).

### DIFF
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -296,12 +296,17 @@ common::stack_index_maps(
   const std::vector<int> out_neighbors(out_neighbor_set.begin(),
                                        out_neighbor_set.end());
 
+  // NOTE: create uniform weights as a workaround to issue
+  // https://github.com/pmodels/mpich/issues/5764
+  const std::vector<int> in_weights(in_neighbors.size(), 1);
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -296,12 +296,17 @@ common::stack_index_maps(
   const std::vector<int> out_neighbors(out_neighbor_set.begin(),
                                        out_neighbor_set.end());
 
+  // NOTE: create uniform weights as a workaround to issue
+  // https://github.com/pmodels/mpich/issues/5764
+  const std::vector<int> in_weights(in_neighbors.size(), 1);
+  const std::vector<int> out_weights(out_neighbors.size(), 1);
+
```
chrisrichardson, line 303: It's not really needed here, as we soon free the comm anyway, and it never gets copied.

chrisrichardson (approved):

cpp/dolfinx/common/IndexMap.cpp
garth-wells: It can be duplicated (and probably will be) by functions that access the IndexMap comms.

PR #1948, opened 2022-01-18
Refactor `Function::eval` for performance

jorgensd:
As `basix.tabulate` allocates alot of data structures, it cannot be called in loops. 
This PR splits `eval` into three steps:
1. Pull back physical coordinates and compute Jacobian, Inverse and determinant (for all input points in a for-loop).
2. Tabulate finite element at all pulled back points (outside loop)
3. Transform, push_forward and expand coefficients (for all input points in a for-loop).

Following is the results for `compute_point_values` (which uses eval at all mesh vertices) for the MWE:
```python
from time import perf_counter
from dolfinx import fem
from dolfinx.mesh import create_unit_cube, CellType
from mpi4py import MPI
import ufl

N = 20
mesh = create_unit_cube(MPI.COMM_WORLD, N, N, N, cell)
num_cells_local = mesh.topology.index_map(mesh.topology.dim).size_local

ufl_cell = mesh.ufl_cell()
if space == "Lagrange":
    el = ufl.VectorElement(space, ufl_cell, degree)
else:
    el = ufl.FiniteElement(space, ufl_cell, degree)

V = fem.FunctionSpace(mesh, el)
u = fem.Function(V)

start = perf_counter()
u.compute_point_values()
end = perf_counter()
```
compare_5
compare_4
compare_3
compare_2
compare_1
compare_0


garth-wells:
See https://github.com/FEniCS/basix/issues/313 for Basix issue.

garth-wells:
I think it would help if the PR description was tightened up a bit.
- Is the performance issue only for evaluations at 'arbitrary' physical points? Cases where the reference points are the same on all cells should not use `eval`, right?
- From the description it seems all point are passed to Basix at once. Does Basix memory scale linearly with the number of points?

jorgensd:
>     * Is the performance issue only for evaluations at 'arbitrary' physical points? Cases where the reference points are the same on all cells should not use `eval`, right?

The issue only affects `eval`, which evaluates at arbitrary physical points. 
This is avoided for cases where the reference points are the same in all elements by using `dolfinx.fem.Expression`.

>     * From the description it seems all point are passed to Basix at once. Does Basix memory scale linearly with the number of points?
I have not investigated how the memory in basix scale with the number of points. Maybe at @mscroggs could shed some light on this.

garth-wells (approved):

PR #1949, opened 2022-01-19
Make timing MPI reduction easily selectable, and MPI_MAX by default

chrisrichardson:
Currently, `list_timings()` produces the average time of a `Timer` across all MPI ranks. This is misleading, and the maximum (or minimum) are more useful.
This PR makes it possible to easily select, and makes max the default.


jorgensd:
Could you expose this to the Python interface as well?

garth-wells (approved):

PR #1950, opened 2022-01-19
Add optional `cells` argument in interpolate functions

garth-wells:
This sets the cells that interpolation is performed over.

### DIFF
--- a/cpp/dolfinx/fem/Function.h
+++ b/cpp/dolfinx/fem/Function.h
@@ -225,12 +225,41 @@ public:
   }
 
   /// Interpolate an Expression (based on UFL)
-  /// @param[in] e The function to be interpolated
+  /// @param[in] e The Expression to be interpolated. The Expression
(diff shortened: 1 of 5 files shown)


cpp/dolfinx/fem/Function.h
```diff
@@ -195,13 +210,29 @@ class Function
...
+    std::int32_t num_cells = cell_map->size_local() + cell_map->num_ghosts();
+    std::vector<std::int32_t> cells(num_cells, 0);
+    std::iota(cells.begin(), cells.end(), 0);
+    interpolate(f, cells);
+  }
+
+  /// Interpolate an Expression (based on UFL)
+  /// @param[in] e The function to be interpolated
+  /// @param[in] cells The cells to interpolate into
+  void interpolate(const Expression<T>& e,
```
jorgensd, line 232: We should have an interpolate for `Expression` over the whole domain as well I guess.

PR #1952, opened 2022-01-19
Enable CMAKE_INCLUDE_CURRENT_DIR in cpp/demos

IgorBaratta:
Currently the following build command doesn't work:

```bash
cd cpp/demo
mkdir build
cd build
cmake ..
make -j4
```
``` bash
dolfinx/cpp/demo/poisson/main.cpp:90:10: fatal error: poisson.h: No such file or directory
   90 | #include "poisson.h"
```


garth-wells (changes requested):
This needs to be changed in the script that generates the demo CMake files.

garth-wells:
I'm not sure if all these changes are needed - compare with https://github.com/FEniCS/performance-test/blob/main/src/CMakeLists.txt.

IgorBaratta:
> I'm not sure if all these changes are needed - compare with https://github.com/FEniCS/performance-test/blob/main/src/CMakeLists.txt.

Adding 
```
set(CMAKE_INCLUDE_CURRENT_DIR ON)
```
does the trick.

garth-wells (approved):

PR #1955, opened 2022-01-19
Remove compute point values from demos

jorgensd:
Add example of how to visualize second order N1Curl in Python.

Added example of how to use interpolation on subsets of cells (and remove `dolfinx.fem.LinearProblem` from pyvista demo).

Bump pyvistaqt to 0.7.0

The function is still exposed to python as it is used for XDMFFile.write_function, as should therefore be tested.


.github/workflows/pyvista.yml
```diff
@@ -53,7 +51,7 @@ jobs:
           apt-get install -y --no-install-recommends libgl1-mesa-dev xvfb # pyvista
           apt-get install -y --no-install-recommends python3-pyqt5 libgl1-mesa-glx  # pyvistaqt
           pip3 install pyvista==${PYVISTA_VERSION}
-          python3 -m pip install git+https://github.com/pyvista/pyvistaqt.git@${PYVISTA_QT_VERSION}
+          pup3 install pyvistaqt==${PYVISTA_QT_VERSION}
```
jorgensd, line 54:
```suggestion
          pip3 install pyvistaqt==${PYVISTA_QT_VERSION}
```

PR #1956, opened 2022-01-19
Move data allocation out of `scatter_**_begin`

IgorBaratta:


cpp/dolfinx/la/Vector.h
```diff
@@ -43,6 +43,12 @@ class Vector
       MPI_Type_contiguous(bs, dolfinx::MPI::mpi_type<T>(), &_datatype);
       MPI_Type_commit(&_datatype);
     }
+
+    // Pre-allocate buffers data
```
garth-wells, line 47: Can this be moved into the initialiser list?

garth-wells (approved):

PR #1964, opened 2022-01-24
Rmove span of view with first index always being 0

jorgensd:
Instead of creating a new xtensor for pushing forward basis values in interpolation, 
use view of original data structure from basix (as `basis_derivatives_reference0.shape(0)=1`).


jorgensd:
Copy is needed as reference values are only computed once, and the transformation would keep if we work on the view

PR #1966, opened 2022-01-25
Refactor pack coefficients

IgorBaratta:
- Separate memory allocation and computation
- Reduce code duplication

Main idea is that one can initialize coefficients once and reuse the allocated memory:

```c++
std::map<std::pair<IntegralType, int>, std::pair<std::vector<T>, int>> coeff = fem::allocate_coefficient_memory(L);

while (something){
 /// do some work
 fem::pack_coefficients(L, coeff);
 // Assemble vector, matrix, etc
 ...
}
```


garth-wells (changes requested):
I think this change is fine, but I don't like that there are now two ways to pack coefficients (return by value and pass by reference). This is a low-level and performance oriented interface, so I'd be happy to have just the new in/out argument version and remove the old version.

IgorBaratta:
> I think this change is fine, but I don't like that there are now two ways to pack coefficients (return by value and pass by reference). This is a low-level and performance oriented interface, so I'd be happy to have just the new in/out argument version and remove the old version.

Removed the return by value function.

garth-wells (approved):

PR #1978, opened 2022-02-01
Add support for `pathlib`

finsberg:
When working with python paths, it is very common to use `pathlib`, i.e

```python
from pathlib import Path
path = Path("poisson.xdmf")
```
There are [many good reasons](https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/) for using `pathlib` instead of `os.path`
However it is not currently supported to pass in a `pathlib` object to `XDMFFile`. 

The following code

```python
with io.XDMFFile(comm, Path("poisson.xdmf"), "w") as file:
    file.write_mesh(mesh)
    file.write_function(uh)
```

will produce the following error

```
  File "/Users/henriknf/local/src/fenicsx/dolfinx/python/demo/poisson/demo_poisson.py", line 212, in <module>
    with io.XDMFFile(comm, Path("poisson.xdmf"), "w") as file:
TypeError: __init__(): incompatible constructor arguments. The following argument types are supported:
    1. dolfinx.cpp.io.XDMFFile(comm: MPICommWrapper, filename: str, file_mode: str, encoding: dolfinx.cpp.io.XDMFFile.Encoding = <Encoding.HDF5: 0>)

Invoked with: <mpi4py.MPI.Intracomm object at 0x107900f30>, PosixPath('poisson.xdmf'), 'w'
```

In this PR I add support for passing in `pathlib` objects by declaring the `filename` argument in the `XDMFFile` constructor as type [`std::filesystem::path`](https://en.cppreference.com/w/cpp/filesystem/path). It was also added support for this conversion in [pybind11 v2.7.0](https://newreleases.io/project/github/pybind/pybind11/release/v2.7.0)

I wasn't able to run the tests locally, so I hope the CI will catch any potentially bugs that I introduced.


garth-wells:
Looks fine, but we just need to check for any issues with `std::filesystem`. We used `boost::filesystem` over `std::filesystem` because until some point `std::filesystem` support in gcc and clang was clunky. I've created #1980, which switches `boost::filesystem` -> `std::filesystem` internally. We'll see if anyone has concerns with the switch. I expect that Debian/Ubuntu 20.04 compilers are now new enough

garth-wells (changes requested):
@finsberg could you resolve the merge conflicts with `main`?

garth-wells (approved):

Issue #1976, opened 2022-02-01
Feedback on `demo_poisson.py`


finsberg:
Hi,

I am a FEniCS user and I am currently exploring the FEniCSx demos in order to start learning the new syntax for FEniCSx. 
I started with the [`demo_possion.py`](https://github.com/FEniCS/dolfinx/blob/main/python/demo/poisson/demo_poisson.py), and I am sharing my thoughts here in order to express what I think could be be done in a different way, and what I think is currently missing. If this is not the correct format to provide such feedback please let me know. I would also like to say that many of the things I highlight here expresses my opinions which is definitely not necessarily the best for FEniCSx.

I would be happy to help out with a pull request if you think it is appropriate.

## Imports

1. Do you have any guidelines in terms of import order? I think it is worth consider to have a consistent way of importing modules, such as first python std library, then third party and then dolfinx. Futhermore, within each group it should be alphabetical, and you can use a tool such as `isort` or `reorder-python-imports` to do this automatically (or as a pre-commit hook).
2. It should be some standard way in how you import modules and functions. For example, I think doing both
   ```python
   from dolfinx import fem
   ```
   and
   ```python
   from dolfinx.fem import FunctionSpace
   ```
    is a bit redundant. Why not just use `fem.FunctionSpace`? 
    Similarly I think it would by good to just use e.g
    ```python
    import ufl
    ufl.grad
    ```
    I think this also aligns well with how people use other modules, such as `numpy`. This way there is no ambiguity when you use `np.exp` and `ufl.exp` for example.

## Positional vs Keyword arguments
I don't know if you want to stick to using positional arguments. In my opinion it is more clear if one uses keyword arguments for most things. As an example, I think it is better to write
```python
mesh = create_rectangle(
    comm=MPI.COMM_WORLD, 
    points=((0.0, 0.0), (2.0, 1.0)), 
    n=(32, 16), 
    cell_type=CellType.triangle
)
```
than
```python
mesh = create_rectangle(
    MPI.COMM_WORLD,
    ((0.0, 0.0), (2.0, 1.0)),
    (32, 16),
    CellType.triangle
)
```

## FunctionSpace
I think it is great that you have started to use type hint in dolfinx. This will make it much better to work with in an IDE.

I think the FunctionSpace class need some better documentation. When I look up the signature for FunctionSpace, I get the following
```
fem.FunctionSpace(
    mesh: 'Mesh',
[...]

PR #1979, opened 2022-02-01
Update poisson demo

finsberg:
In https://github.com/FEniCS/dolfinx/issues/1976 I added some feedback on the Poisson demo, and in this PR I try to incorporate some of these ideas.

First of all, I realised that the demo was written for a unit square, but the code in the demo was using a rectangle. I don't know what you prefer here? I guess if this is the very first demo, then a unit square it probably OK. I tried to change the wording so that it uses the rectangle for now.

When it comes to **imports** I try to use the namespaces for sub-packages (`fem`, `io` etc). I definitely see the problem with the `mesh` sub-package,  and I solved it here by importing it as `_mesh` (which is also a bit clumsy I think).

I also wrote a bit more verbose code so that it is a bit easier to understand for the beginner. 

I also updated the figure so that the figure is what you see when you plot with `pyvista`.


chrisrichardson:
I don't know how everyone else feels, but I find `ufl.grad` `ufl.inner` `ufl.dx` all very intrusive in a form, 'a' or 'L' and ir reduces the mathematical elegance with `ufl.` everywhere...

steffen-sch:
Hi, 

thanks for your effort. I want to share some thoughts purely from a user perspective, because I think this demo will be one of the most read examples. Therefore, it ist important for the reader to understand what is going on.

```
# Create an MPI communicator
comm = MPI.COMM_WORLD
```
I have no idea what an MPI communicator is, therefore the comment doesn't help me at all. The user gets no idea what is going on.

```
# Create mesh
mesh = _mesh.create_rectangle(
```
In Python, the leading underscore marks things as protected. This code looks like an access to a protected module. 
I would recommend to use "mesh" as the module name: "rect_mesh = mesh.create_rectangle(".

```
# Define boundary condition on x = 0 or x = 1
facets = _mesh.locate_entities_boundary(mesh, dim=1, marker=marker)
dofs = fem.locate_dofs_topological(V=V, entity_dim=1, entities=facets)
bc = fem.dirichletbc(value=ScalarType(0), dofs=dofs, V=V)
```
"locate_entities_boundary"  doesn't read well.  For me "locate_boundary_entities" sounds much more natural. But "locate" is what the programmer of the function has thought about.  The user "locates" it in the marker function above. As a user I want to "get_boundary_entities". She same applies for the "locate_dofs_topological" function name.
One other Python convention that improves readability is to separate words by an underscore.  "dirichletbc" should be at least "dirichlet_bc".

And as a user, it confuses me, why is fem, plot, io and mesh imported from dolfinx and ufl doesn't. Is there a practical relevance (for the user) to do so?

jorgensd:
> I have no idea what an MPI communicator is, therefore the comment doesn't help me at all. The user gets no idea what is going on.
> 

We could add something along the lines of: 


> We also have to supply the MPI-communicator. This is to specify how we would like the program to behave in parallel. If we supply MPI.COMM_WORLD we create a single mesh, whose data is distributed over the number of processors we would like to use. We can for instance run the program in parallel on two processors by using mpirun, as:
> 
>  mpirun -n 2 python3 t1.py
> 
> However, if we would like to create a separate mesh on each processor, we can use MPI.COMM_SELF. This is for instance useful if we run a small problem, and would like to run it with multiple parameters.
> 



The text above is taken from my[ dolfinx tutorial](https://jorgensd.github.io/dolfinx-tutorial/chapter1/fundamentals_code.html)


> And as a user, it confuses me, why is fem, plot, io and mesh imported from dolfinx and ufl doesn't. Is there a practical relevance (for the user) to do so?

ufl is a separate python module: https://github.com/FEniCS/ufl
It provides the form language that DOLFINx interprets.
In the old version of dolfin, it was a constant source of confusion that certain ufl functions where loaded when loading dolfin, and others not.

steffen-sch:
@jorgensd in regard to  MPI: Maybe for the user it's enough to know that MPI controls the parallel processing and its default configuration. For me the question is, if it is trying to use multiple processors in this case.

jorgensd:
> @jorgensd in regard to MPI: Maybe for the user it's enough to know that MPI controls the parallel processing and its default configuration. For me the question is: does it try to use multiple processors in this case?

That was the whole point of my quote above. If you run your problem as usual with `python3 name_of_program.py`, you are spawning one process, i.e. the mesh is not distributed. However, if you would use `mpirun -n 5 python3 name_of_program.py` the data will be distributed over 5 processes if you use `MPI.COMM_WORLD`. If you use `MPI.COMM_SELF`, 5 separate meshes will be created, one on each process.

steffen-sch:
> That was the whole point of my quote above. If you run your problem as usual with `python3 name_of_program.py`, you are spawning one process, i.e. the mesh is not distributed. However, if you would use `mpirun -n 5 python3 name_of_program.py` the data will be distributed over 5 processes if you use `MPI.COMM_WORLD`. If you use `MPI.COMM_SELF`, 5 separate meshes will be created, one on each process.
For me, the default case is not to use mpirun. This means the MPI code has no effect.  This would be good to know as  a reader of the demo.

finsberg:
Would it be possible to make the communicator take a default value of `MPI.COMM_WORLD`, so that the users at the beginner level doesn't need to care about the MPI communication, or would this go against some fundamental principles? Personally, I didn't understand this in the beginning when I started with FEniCS, and I actually ended up written my own wrappers that had `COMM_WORLD` as the default communicator.

garth-wells:
> Would it be possible to make the communicator take a default value of `MPI.COMM_WORLD`, so that the users at the beginner level doesn't need to care about the MPI communication, or would this go against some fundamental principles? Personally, I didn't understand this in the beginning when I started with FEniCS, and I actually ended up written my own wrappers that had `COMM_WORLD` as the default communicator.

Past experience is that default communicators lead to all manner of library bugs. We end up with code that only works on `COMM_WORLD`.

If we were to consider a default communicator, we would probably need to move it to the end of the argument list like `petsc4py` does.

python/demo/poisson/demo_poisson.py
```diff
@@ -76,30 +76,37 @@
...
-mesh = create_rectangle(MPI.COMM_WORLD, ((0.0, 0.0), (2.0, 1.0)), (32, 16), CellType.triangle)
-V = FunctionSpace(mesh, ("Lagrange", 1))
+# function space :math:`V` relative to this mesh. We create a
+# rectangular mesh using built-in function provided by
+# class :py:class:`create_rectangle<dolfinx.mesh.create_rectangle>`.
+# In order to create a mesh consisting of 32 x 16 squares with each
+# square divided into two triangles, we do as follows ::
+
+# Create an MPI communicator
+comm = MPI.COMM_WORLD
```
garth-wells, line 91: I don't feel that this extra assignment add much.

On a technical level, we're not creating a communicator here. `MPI.COMM_WORLD` already exists.

meg-simula:
> I don't know how everyone else feels, but I find `ufl.grad` `ufl.inner` `ufl.dx` all very intrusive in a form, 'a' or 'L' and it reduces the mathematical elegance with `ufl.` everywhere...

Absolutely agree. Recommend avoiding the ufl. prefix.

python/demo/poisson/demo_poisson.py
```diff
@@ -76,30 +76,37 @@
...
+# Create mesh
+mesh = _mesh.create_rectangle(
+    comm=comm,
+    points=((0.0, 0.0), (2.0, 1.0)),
+    n=(32, 16),
+    cell_type=_mesh.CellType.triangle
+)
+# Define element which is a tuple with (family, degree, form_degree).
+# We can also omit the form_degree and use the default value
+element = ("Lagrange", 1)
```
garth-wells, line 102: I don't find the extra assignment here so helpful. I prefer
```
V = fem.FunctionSpace(mesh, ("Lagrange", 1))
```

python/demo/poisson/demo_poisson.py
```diff
@@ -134,10 +141,19 @@
...
 
+# Create a function that returns True if the point is on the boundary
+
+
+def marker(x):
+    left = np.isclose(x[0], 0.0)
+    right = np.isclose(x[0], 2.0)
+    return np.logical_or(left, right)
+
+
```
garth-wells, line 152: I like the lambda, and I like introducing users to it - it emphasises the data-centric design that distinguishes DOLFINx.

meg-simula:
> In Python, the leading underscore marks things as protected. This code looks like an access to a protected module. I would recommend to use "mesh" as the module name: "rect_mesh = mesh.create_rectangle(".
> 

An idea would be to call the module  'meshing' instead (of mesh).

python/demo/poisson/demo_poisson.py
```diff
@@ -134,10 +141,19 @@
...
 # Define boundary condition on x = 0 or x = 1
-facets = locate_entities_boundary(mesh, 1, lambda x: np.logical_or(np.isclose(x[0], 0.0),
-                                                                   np.isclose(x[0], 2.0)))
-bc = dirichletbc(ScalarType(0), locate_dofs_topological(V, 1, facets), V)
+facets = _mesh.locate_entities_boundary(mesh, dim=1, marker=marker)
```
garth-wells, line 154: These few steps will need some extra explanation in the comments.

python/demo/poisson/demo_poisson.py
```diff
@@ -134,10 +141,19 @@
...
+dofs = fem.locate_dofs_topological(V=V, entity_dim=1, entities=facets)
+bc = fem.dirichletbc(value=ScalarType(0), dofs=dofs, V=V)
```
garth-wells, line 156: `ScalarType(0)` is clumsy and really just for the CI. Would be good if we could avoid it somehow, without reducing the functionality.

python/demo/poisson/demo_poisson.py
```diff
@@ -212,9 +229,11 @@
     # If pyvista environment variable is set to off-screen (static)
     # plotting save png
     if pyvista.OFF_SCREEN:
+        # Run demo with 'PYVISTA_OFF_SCREEN=true python demo_poisson.py'
```
garth-wells, line 232: I wonder it we want this. It's a complication introduced by CI and docker.

python/demo/poisson/demo_poisson.py
```diff
@@ -212,9 +229,11 @@
...
         pyvista.start_xvfb(wait=0.1)
-        plotter.screenshot("uh.png")
+        plotter.screenshot("poisson_u.png")
     else:
         plotter.show()
 except ModuleNotFoundError:
-    print("pyvista is required to visualise the solution")
+    print("'pyvista' is required to visualise the solution")
+    print("Install 'pyvista' with pip: 'python3 -m pip install pyvista'")
```
garth-wells, line 239: Could the command be just `pip3 install pyvista`?

python/demo/poisson/demo_poisson.py
```diff
@@ -76,30 +76,37 @@
...
+# rectangular mesh using built-in function provided by
+# class :py:class:`create_rectangle<dolfinx.mesh.create_rectangle>`.
+# In order to create a mesh consisting of 32 x 16 squares with each
+# square divided into two triangles, we do as follows ::
+
+# Create an MPI communicator
+comm = MPI.COMM_WORLD
+
+# Create mesh
+mesh = _mesh.create_rectangle(
```
garth-wells, line 94: I'm not sure the extra vertical spacing helps readability.

garth-wells (changes requested):
Thanks for this!

I've added comments where I have suggestions for changes. Unfortunately the bits that I think look fine as-is get no comments!

I'd like for now to remove the png. I think we can discuss later whether or not we want images files in the repo. What I'd like to avoid is an accumulation of binary junk in the repo as we refine the demos.

garth-wells:
> I don't know how everyone else feels, but I find `ufl.grad` `ufl.inner` `ufl.dx` all very intrusive in a form, 'a' or 'L' and it reduces the mathematical elegance with `ufl.` everywhere...

I think we need `ufl.` for UFL objects (`TestFunction`, etc). Maybe the compromise is we can get away without `ufl` for most operators/functions that operate on UFL objects?

python/demo/poisson/demo_poisson.py
finsberg: I belive the main benefit of using `python3 -m pip` as opposed to `pip3` is that if you run the demo with `python3 demo_poisson.py` then you are sure that the dependency you install  (in the case `pyvista`) is installed in the correct environment. You can see a more thorough discussion about this topic here: https://snarky.ca/why-you-should-use-python-m-pip/

python/demo/poisson/demo_poisson.py
finsberg: I guess this also depends if you want to use keyword arguments or not, and if you want to keep the line length below a certain value.

python/demo/poisson/demo_poisson.py
finsberg: Sounds good. I reverted this change in https://github.com/FEniCS/dolfinx/pull/1979/commits/b48e86cbdefc812f07b0e2f131a7d799853168fe

python/demo/poisson/demo_poisson.py
finsberg: Fixed in https://github.com/FEniCS/dolfinx/pull/1979/commits/b48e86cbdefc812f07b0e2f131a7d799853168fe

python/demo/poisson/demo_poisson.py
finsberg: I added a few more explanations in https://github.com/FEniCS/dolfinx/pull/1979/commits/b48e86cbdefc812f07b0e2f131a7d799853168fe but these should probably be revised by someone who know the intentions behind the choice of API here. 

Personally I think it would be good not having to call `locate_entities_boundary` and `locate_dofs_topological`, but I guess this is also me not being used to the new syntax in FEniCSx.

python/demo/poisson/demo_poisson.py
finsberg: I removed the saving to png part. I guess there will be a separate demo about `pyvista` where this can be explained in more detail.

python/demo/poisson/demo_poisson.py
jorgensd: The reason for splitting entities and degrees of freedom is that the entities can be used for a lot of other things than applying Dirichlet boundary conditions.
For Dirichlet bcs there are two functions; locate_dofs_topological and locate_dofs_geometrical.
Locate_dofs_geometrical only make sense for spaces that has point evaluations as functionals (i.e. Lagrange), and cannot be used with nedelec spaces and others (as a degree of freedom is associated with a facet).
For such spaces you can use locate_dofs_topological.

jhale:
@finsberg I used your PR as a start for mine https://github.com/FEniCS/dolfinx/pull/1992. You can merge main onto your branch using 'theirs' to resolve the conflicts.

garth-wells:
This has been super helpful, and the advice has also helped us get the 'machinery' of creating the demo docs in place.

Closing this PR since the changes have been cherry picked into merged changes for how the docs are generated.

PR #1982, opened 2022-02-02
Remove `dolfinx_utils`

garth-wells:
Moves functionality into the 'proper' location (`conftest.py`).


francesco-ballarin:
Some of the fixtures and/or skips are actually quite useful while preparing tests for libraries that depend on `dolfinx`. I guess downstream developers (e.g., me) can live with duplicating those fixtures/skip in their code, but maybe can you consider packaging them in a place like `dolfinx.test`, rather than removing `dolfinx_utils` altogether?

garth-wells:
> Some of the fixtures and/or skips are actually quite useful while preparing tests for libraries that depend on `dolfinx`. I guess downstream developers (e.g., me) can live with duplicating those fixtures/skip in their code, but maybe can you consider packaging them in a place like `dolfinx.test`, rather than removing `dolfinx_utils` altogether?

We're aiming to have only one 'skip' (when number of parallel ranks is greater than 1). The only other skip we're using is for complex mode, which is used in only ~2 places and we'll aim to soon remove (most objects now accept a `dtype` arg).  There is only one fixture, which with some luck we can remove.

Give the very small number of test fixtures and skips, I think it is better to reproduce locally because there is no guarantee that these won't be changed or removed.

PR #1986, opened 2022-02-03
Adding in optional communicator for PETSc init

adrianjhpc:
This change allows users to specify an MPI communicator when initialising PETSc so that this can be called from a sub-set of MPI processes when used from another application. Prior to this change, users had to manually specify:

PETSC_COMM_WORLD =subcomm

(where subcomm is the specific MPI communicator) prior to any PETSc functionaity in the dolfinx code being used to avoid deadlock. This formalises it by passing the communicator as a specific parameter to the PETSc init, and defaulting to MPI_COMM_WORLD if no communicator is passed.


cpp/dolfinx/common/subsystem.cpp
```diff
@@ -64,18 +63,26 @@ void subsystem::init_logging(int argc, char* argv[])
...
 //-----------------------------------------------------------------------------
-void subsystem::init_petsc(int argc, char* argv[])
+void subsystem::init_petsc(int argc, char* argv[], MPI_Comm comm)
 {
   if (argc > 1)
     LOG(INFO) << "Initializing PETSc with given command-line arguments.";
 
+  // Ensure MPI has been initialised, otberwise setting PETSC_COMM_WORLD can 
+  // call MPI not initialised errors.
+  init_mpi(argc, argv);
```
garth-wells, line 80: In the interests of being explicit, I would suggest checking that MPI has been initialised and throwing and error if it hasn't, and suggest that MPI needs to be initialised by the user first. Risk at the moment is confusion around who is responsible for finalising MPI.

cpp/dolfinx/common/subsystem.cpp
```diff
@@ -64,18 +63,26 @@ void subsystem::init_logging(int argc, char* argv[])
...
+  // Ensure the PETSC_COMM_WORLD is not MPI_COMM_WORLD if the user has specified 
+  // a communicator.
+  //
+  PETSC_COMM_WORLD = comm;
```
garth-wells, line 84: Is this safe? Looks wrong to assign to `PETSC_COMM_WORLD`.

cpp/dolfinx/common/subsystem.h
```diff
@@ -27,11 +29,11 @@ void init_logging(int argc, char* argv[]);
 
 /// Initialize PETSc (and SLEPc, if configured) without command-line
 /// arguments
-void init_petsc();
+void init_petsc(MPI_Comm comm = MPI_COMM_WORLD);
```
garth-wells, line 32: Would it be better to make the default `MPI_COMM_NULL` == let PETSc manage the MPI initialisation?

garth-wells (commented):
The support that this changes provides is good. I'm just wondering if the implementation is a bit 'implicit'. I've left a few comments inline.

cpp/dolfinx/common/subsystem.cpp
adrianjhpc: Good point. The reason I didn't do that is that `PetscInitialize` also will implicitly initialise MPI if it's not initialised. If you're happy moving to the model that users always have to initialise MPI first before calling PETSc (which would be a change to the current behaviour, but would be much more explicit) then that's fine.

cpp/dolfinx/common/subsystem.cpp
adrianjhpc: Yes, it looks weird doesn't it, but it's the "PETSc" way, i.e. https://petsc.org/main/docs/manualpages/Sys/PETSC_COMM_WORLD.html

cpp/dolfinx/common/subsystem.h
adrianjhpc: I'm not sure which is best. If `PETSC_COMM_WORLD` == `MPI_COMM_WORLD` then it's just the same as letting PETSc manage the initialisation. 

If however, you want people to be able to let `init_petsc` do MPI initialisation when people aren't specifying explicit communicators, but require people to initialise MPI manually prior to `init_petsc` if they are specifying communicators, this would be the way to go.

I guess this is a design decision that is probably best not made by me.

cpp/dolfinx/common/subsystem.cpp
adrianjhpc: Checking `PetscFinalize` I see it will only call `MPI_Finalize` if it also called `MPI_Init` but will not call `MPI_Finalize` if the user called `MPI_Init` prior to calling `PetscInitialize` (i.e. `init_petsc`)

garth-wells:
This PR prompted me to look closely again at the code in `subsystem.h`. I think `subsystem.h/cpp` can just be removed, which handles the cases that this PR sets out to fix, forcing the user to be explicit. See PR #2094

adrianjhpc:
Great

garth-wells:
Superseded by #2094.

Issue #1989, opened 2022-02-05
Support the VTK HDF5 IO format


garth-wells:
VTK now has an HDF5 format for unstructured grids - https://kitware.github.io/vtk-examples/site/VTKFileFormats/#unstructured-grid. This format exploits more VTKs features than XDMF, and the future of XMDF is uncertain.

It would be helpful to support this format, and it should be straightforward as it builds on the existing VTK/VTU data layout.

garth-wells:
See #3674.

PR #1990, opened 2022-02-05
Improve VTK output

garth-wells:
This change removes calls to `Function::compute_point_values` for VTK output, which was very slow. VTK output now preserves exactly the Function being output - no hidden interpolation or projection.
- `Functions` sent to a VTKFile must all use the same `FiniteElement` type (up to the block size), unless a `Function` is DG0 (CellData). Cell-wise constant Functions can be mixed with point-wise Functions,
- The `FiniteElement` must be (discontinuous) Lagrange. It can be any degree.
- Output `Functions` cannot be sub-functions. Sub-functions should be collapsed or interpolated before output
- VTU output now contains information on ghost cells and nodes, and 'global' cell and nodes ids. The information us used by VTK/ParaView.
- VTU and VTX/ADIOS2 output now share more code
- Extra tests added

The changes will support further improvements to VTX/ADIOS2 output and future VTK/HDF5 input/output (#1989).


cpp/dolfinx/io/vtk_utils.h
```diff
@@ -0,0 +1,49 @@
...
+} // namespace dolfinx::mesh
+
+namespace dolfinx::io
+{
+/// Given a FunctionSpace, create a topology and geometry based on the
+/// dof coordinates.
+/// @note Only supports (discontinuous) Lagrange functions
+/// @param[in] V The function space
+std::tuple<xt::xtensor<double, 2>, std::vector<std::int64_t>,
+           std::vector<std::uint8_t>, xt::xtensor<std::int32_t, 2>>
```
jorgensd, line 33: Maybe add a description of the output. Why are there 4 arrays in the tuple?

I guess it should be:
Returns the mesh geometry `(block_size*(num_dofs_local+num_ghosts, 3)`, the global geometry index, ghost_level marker and the topology (mesh connectivity) using local indices

python/test/unit/io/test_vtk.py
```diff
@@ -51,18 +50,12 @@ def test_save_3d_mesh(tempdir, cell_type):
...
     u = Function(FunctionSpace(mesh, ("Lagrange", 2)))
-    u.interpolate(f)
-    u.vector.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
-
+    u.interpolate(lambda x: x[0])
     filename = os.path.join(tempdir, "u.pvd")
     with VTKFile(MPI.COMM_WORLD, filename, "w") as vtk:
         vtk.write_function(u, 0.)
-        vtk.write_function(u, 1.)
+        # vtk.write_function(u, 1.)
```
jorgensd, line 58: Remove comment or remove line?

jorgensd (changes requested):

cpp/dolfinx/io/vtk_utils.cpp
```diff
@@ -0,0 +1,207 @@
...
+      std::int32_t dof = dofs[i];
+      for (std::size_t j = 0; j < gdim; ++j)
+        coords(dof, j) = x(i, j);
+    }
+  }
+
+  // Origina points IDs
+  std::vector<std::int64_t> x_id(num_nodes);
+  std::array<std::int64_t, 2> range = map_dofs->local_range();
+  std::int32_t size_local = range[1] - range[0];
```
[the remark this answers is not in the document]
garth-wells: Not sure this helps readability.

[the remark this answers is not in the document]
jorgensd: I added this to make it an explicit and not implicit cast

python/test/unit/io/test_adios2.py
```diff
@@ -234,3 +234,22 @@ def vel(x):
         f.write(t)
 
     f.close()
+
+
+@pytest.mark.skipif(not has_adios2, reason="Requires ADIOS2.")
```
jorgensd, line 239: I guess this test should be marked with X-fail, as the output currently is not readable with Paraview?

garth-wells: It won't produce output. The test is checking that an exception is raised.

cpp/dolfinx/io/vtk_utils.h
garth-wells: Return object docs added.

PR #1996, opened 2022-02-08
Extend pyvista plotting to higher order meshes.

jorgensd:
Reuse some functionality from #1990 to properly plot higher order meshes with Pyvista.

Plotting of sub entities (edges, facets) are linearised based on the topology.

### DIFF
--- a/python/dolfinx/plot.py
+++ b/python/dolfinx/plot.py
@@ -1,4 +1,4 @@
-# Copyright (C) 2021 Jørgen S. Dokken and Garth N. Wells
+# Copyright (C) 2021-2022 Jørgen S. Dokken and Garth N. Wells
 #
 # This file is part of DOLFINx (https://www.fenicsproject.org)
 #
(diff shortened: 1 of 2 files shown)


python/dolfinx/plot.py
```diff
@@ -32,39 +33,44 @@ def create_vtk_mesh(mesh: _mesh.Mesh, dim: int, entities=None):
...
         num_cells = len(entities)
 
-    # Get the indices in the geometry array that correspong to the
-    # topology vertices
-    # NOTE: This linearizes higher order geometries
-    geometry_entities = _cpp.mesh.entities_to_geometry(mesh, dim, entities, False)
+    if dim == mesh.topology.dim:
+        geometry_entities = _cpp.io.extract_vtk_connectivity(mesh)[entities]
+        de = mesh.geometry.cmap.create_dof_layout()
+        num_nodes_per_cell = de.num_dofs
```
garth-wells, line 51: `num_nodes_per_cell =  geometry_entities.shape[1]`

python/dolfinx/plot.py
```diff
@@ -32,39 +33,44 @@ def create_vtk_mesh(mesh: _mesh.Mesh, dim: int, entities=None):
...
         entities = np.arange(num_cells, dtype=np.int32)
     else:
         num_cells = len(entities)
 
-    # Get the indices in the geometry array that correspong to the
-    # topology vertices
-    # NOTE: This linearizes higher order geometries
-    geometry_entities = _cpp.mesh.entities_to_geometry(mesh, dim, entities, False)
+    if dim == mesh.topology.dim:
+        geometry_entities = _cpp.io.extract_vtk_connectivity(mesh)[entities]
```
garth-wells, line 49: I find this variable name confusing. Something like `vtk_cells` or  `vtk_topology` would be clearer.

python/dolfinx/wrappers/io.cpp
```diff
@@ -37,6 +38,10 @@ void io(py::module& m)
   // dolfinx::io::cell vtk cell type converter
   m.def("get_vtk_cell_type", &dolfinx::io::cells::get_vtk_cell_type);
 
+  // dolfinx::io VTK mesh topology converter
```
garth-wells, line 41: No need for a code comment. Simple docstring could be added.

python/dolfinx/plot.py
```diff
@@ -32,39 +33,44 @@ def create_vtk_mesh(mesh: _mesh.Mesh, dim: int, entities=None):
...
+        geometry_entities = geometry_entities[:, map_vtk]
+
+    # Create mesh topology
+    topology = np.empty((num_cells, num_nodes_per_cell + 1), dtype=np.int32)
+    topology[:, 0] = num_nodes_per_cell
+    topology[:, 1:] = geometry_entities
 
-    e_type = _cpp.mesh.cell_entity_type(cell_type, dim, 0)
+    # Array holding the cell type (shape) for each cell
     vtk_type = _first_order_vtk[cell_type] if degree == 1 else _cpp.io.get_vtk_cell_type(cell_type, mesh.topology.dim)
```
garth-wells, line 70: Why are both `_first_order_vtk` and  `_cpp.io.get_vtk_cell_type` needed?

jorgensd: See the comment at the declaration of the dictionary:
```
# NOTE: This dictionary and following function should be revised when
 # pyvista has better support for arbitrary lagrangian elements, see:
 # https://github.com/pyvista/pyvista/issues/947

 # Cell types can be found at
 # https://vtk.org/doc/nightly/html/vtkCellType_8h_source.html
 ```

garth-wells: But the type doesn't fundamentally change the order. `get_vtk_cell_type` effectively give the ID for the cell shape. The 'order' is defined by the number of points per cells.

jorgensd: It does not, but it gives nicer rendering of  edges on first order meshes with the traditional cell type than with the arbitrary lagrangian one.

PR #1997, opened 2022-02-08
Add dtype to constant to avoid importing petsc in demos

jorgensd:
Similar to function, make it possible to specify dtype through constant construction.
Defaults to `PETSc.ScalarType` as Function and Expression.
This cleans up the user-API for most use-cases, and avoids errors when specifying constants such as `dolfinx.fem.Constant(mesh, 1)`.


garth-wells (commented):
The code was designed to take the `dtype` from the argument, e.g.
```
f0 =  Constant(mesh, np.array([0, 0], dtype=np.float32))
f1 =  Constant(mesh, 0.0)
f1 =  Constant(mesh, 0.0 + 1.0j)
```

garth-wells:
Closing since `dtype` is inferred from the bc value type.

Note that the concept of 'real' and 'complex' modes is a PETSc concept and not a DOLFINx concept.

PR #2009, opened 2022-02-11
Neigh comm with null_ptr needs 0 as input size

jorgensd:
Causes segfault with MPICH 4.0.
Has also been an issue with OPENMPI for a longer time.


jorgensd:
Fixed in other PRs

PR #2014, opened 2022-02-14
Return interprocess boundary vertices from non-local dual graph computation

chrisrichardson:
* Reduces code in non-local dual graph
* Returns additional data: list of vertices which lie on the inter-process boundary

This PR is a step towards a new strategy for computing the vertex IndexMap (i.e. vertex ownership and sharing)

### DIFF
--- a/cpp/dolfinx/mesh/graphbuild.cpp
+++ b/cpp/dolfinx/mesh/graphbuild.cpp
@@ -38,8 +38,10 @@ namespace
 /// cell.
 /// @param[in] local_graph The dual graph for cells on this MPI rank
 /// @return (0) Extended dual graph to include ghost edges (edges to
-/// off-rank cells) and (1) the number of ghost edges
-std::pair<graph::AdjacencyList<std::int64_t>, std::int32_t>
+/// off-rank cells), (1) the number of ghost edges, (2) the vertices lying on
+/// the inter-process boundary
+std::tuple<graph::AdjacencyList<std::int64_t>, std::int32_t,
+           std::vector<std::int64_t>>
 compute_nonlocal_dual_graph(
     const MPI_Comm comm, const xt::xtensor<std::int64_t, 2>& unmatched_facets,
     const graph::AdjacencyList<std::int32_t>& local_graph)
@@ -56,7 +58,8 @@ compute_nonlocal_dual_graph(
                 std::vector<std::int64_t>(local_graph.array().begin(),
                                           local_graph.array().end()),
                 local_graph.offsets()),
-            0};
+            0,
+            {}};
   }
 
   // Get cell offset for this process to create global numbering for
@@ -68,8 +71,8 @@ compute_nonlocal_dual_graph(
               dolfinx::MPI::mpi_type<std::int64_t>(), MPI_SUM, comm,
               &request_cell_offset);
 
-  // At this stage facet_cell map only contains facets->cells with edge
-  // facets either interprocess or external boundaries
+  // At this stage unmatched_facets only contains edge
+  // facets which are either interprocess or external boundaries
 
   // TODO: improve scalability, possibly by limiting the number of
   // processes which do the matching, and using a neighbor comm?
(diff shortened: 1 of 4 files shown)


cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -38,8 +38,10 @@ namespace
 /// cell.
 /// @param[in] local_graph The dual graph for cells on this MPI rank
 /// @return (0) Extended dual graph to include ghost edges (edges to
-/// off-rank cells) and (1) the number of ghost edges
-std::pair<graph::AdjacencyList<std::int64_t>, std::int32_t>
+/// off-rank cells), (1) the number of ghost edges, (2) the vertices lying on
```
garth-wells, line 41: Could you clarify the jargon? The docstring has 'edges' which I presume are edges in the graph sense (and not mesh sense), and also 'nodes' which seems to be in the mesh sense (and not graph sense).

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -147,50 +150,44 @@ compute_nonlocal_dual_graph(
...
-  std::vector<int> proc(num_facets_rcvd);
-  for (int p = 0; p < num_ranks; ++p)
-  {
-    for (int f = recvd_disp[p] / (max_num_vertices_per_facet + 1);
-         f < recvd_disp[p + 1] / (max_num_vertices_per_facet + 1); ++f)
-    {
-      proc[f] = p;
-    }
-  }
+  // Create a return buffer with one entry per received facet
```
garth-wells, line 153: Do you mean receive buffer?

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -201,103 +198,86 @@ compute_nonlocal_dual_graph(
...
+  // Copy existing local graph, adding cell_offset
   pos.assign(graph.num_nodes(), 0);
-  std::vector<std::int64_t> ghost_edges;
   for (int i = 0; i < local_graph.num_nodes(); ++i)
   {
-    auto local_graph_i = local_graph.links(i);
-    auto graph_i = graph.links(i);
-    for (std::size_t j = 0; j < local_graph_i.size(); ++j)
-      graph_i[pos[i]++] = local_graph_i[j] + cell_offset;
+    std::transform(local_graph.links(i).begin(), local_graph.links(i).end(),
```
garth-wells, line 259: Create `local_graph.links(i)` before std::transform to avoid multiple constructions?

(might be handy to add some iterators to Adjacency list)

cpp/dolfinx/mesh/graphbuild.h
```diff
@@ -42,9 +42,10 @@ build_local_dual_graph(const xtl::span<const std::int64_t>& cells,
 /// @param[in] cells Collection of cells, defined by the cell vertices
 /// from which to build the dual graph
 /// @param[in] tdim The topological dimension of the cells
-/// @return The (0) dual graph and (1) number of  ghost edges
+/// @return The (0) dual graph, (1) number of  ghost edges, and (2)
```
garth-wells, line 45: See earlier comment on graph vs mesh terminology.

cpp/dolfinx/mesh/graphbuild.cpp
chrisrichardson: It's a reply to the received data. Maybe `reply_buffer` is better. The sequence is something like send -> recv, process data, then reverse the communication, sending back a reply to the originator.

cpp/dolfinx/mesh/graphbuild.cpp
chrisrichardson: Not sure I understand what you mean. But  `pos` could be created using `std::adjacent_difference`, I think.

garth-wells: `local_graph.links(i).begin()` and `local_graph.links(i).end()` creates span twice.

garth-wells:
@chrisrichardson has this been superseded by the recent changes to parallel mesh construction?

chrisrichardson:
Superseded by recent ghosting changes

PR #2016, opened 2022-02-15
Move PETSc assembly functions into submodule

garth-wells:
This makes clear which functions in the assembly interfaces operate on and return PETSc objects.

Plus many doc improvements.


python/dolfinx/fem/assemble.py
```diff
@@ -56,345 +55,175 @@ def _pack(form):
...
+    _cpp.fem.assemble_matrix(A, a, constants, coeffs, bcs)
 
+    # If matrix is a 'diagonal'block, set diagonal entry for constrained
+    # dofs
+    if a.function_spaces[0].id == a.function_spaces[1].id:
+        if len(bcs) > 0:
+            warnings.warn("Setting of matrix bc diagonals not yet implemented.")
+    #     A.assemblyBegin(PETSc.Mat.AssemblyType.FLUSH)
+    #     A.assemblyEnd(PETSc.Mat.AssemblyType.FLUSH)
+    #     _cpp.fem.petsc.insert_diagonal(A, a.function_spaces[0], bcs, diagonal)
```
jorgensd, line 210: Remove commented lines?

garth-wells: These are left in as a reminder of what needs doing next.

PR #2029, opened 2022-02-23
Add more logging to XDMF meshtags

chrisrichardson:


cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -321,6 +321,7 @@ mesh::MeshTags<std::int32_t>
 XDMFFile::read_meshtags(const std::shared_ptr<const mesh::Mesh>& mesh,
                         const std::string name, const std::string xpath)
 {
+  LOG(INFO) << "XDMF read meshtags (" << name << ")";
```
garth-wells, line 324: Does `xdmf_utils::distribute_entity_data` below just read, or is it doing a lot of distribution work? If it's doing a lot of work it would be helpful to add another log message just before it.

cpp/dolfinx/io/xdmf_utils.cpp
```diff
@@ -523,8 +528,10 @@ xdmf_utils::distribute_entity_data(const mesh::Mesh& mesh, int entity_dim,
     entities_send[p].insert(entities_send[p].end(), entity.begin(),
                             entity.end());
     data_send[p].push_back(data[e]);
+    ++total_size;
   }
 
+  LOG(INFO) << "XDMF send entity keys (" << total_size << ")";
```
garth-wells, line 534: The size is just ` entities_vertices.shape(0)`? No need to add a counter?

cpp/dolfinx/io/xdmf_utils.cpp
```diff
@@ -575,10 +583,12 @@ xdmf_utils::distribute_entity_data(const mesh::Mesh& mesh, int entity_dim,
       send_nodes_owned[p1].insert(send_nodes_owned[p1].end(), e_recv.begin(),
                                   e_recv.end());
       send_vals_owned[p1].push_back(_data_recv[e]);
+      ++total_size;
     }
   }
 
   // TODO: Pack into one MPI call
+  LOG(INFO) << "XDMF return entity and value data (" << total_size << ")";
```
garth-wells, line 591: Do we nee to add a size? Counter is a bit ugly, and the log message doesn't say what the number is.

garth-wells (changes requested):
See inline comments.

cpp/dolfinx/io/XDMFFile.cpp
chrisrichardson: It does a lot of redistribution, but there are further log messages inside it now, before each MPI call.

cpp/dolfinx/io/xdmf_utils.cpp
chrisrichardson: size is useful to see where we are using lots of memory (which is the main issue). I'll try to do it without a counter.

PR #2038, opened 2022-03-03
Fix interpolation-io demo [continuous element]

IgorBaratta:
Bool "discontinuous" should be false for continuous element.

It might be useful to use an enum in Basix ...


chrisrichardson (approved):

garth-wells (approved):
Agree this should become an enum.

PR #2039, opened 2022-03-05
Add block size support to MatrixCSR<T>

chrisrichardson:
* Adds block size

Could possibly use templating for `bs` in `add_csr` and `set_csr`?
How to support insertion of 'block data' into an unblocked matrix?


cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -53,14 +57,25 @@ void set_csr(U&& data, const V& cols, const V& row_ptr, const W& x,
...
       auto it = std::lower_bound(cit0, cit1, xcols[c]);
       assert(it != cit1);
       std::size_t d = std::distance(cols.begin(), it);
       assert(d < data.size());
-      data[d] = xr[c];
+      if constexpr (BS > 0)
+      {
+        for (int i = 0; i < BS; ++i)
+          for (int j = 0; j < BS; ++j)
+            data[d * BS * BS + i * BS + j] = xr[(i * nc + c) * BS + j];
```
garth-wells, line 71: `d * BS * BS + i` can be hoisted outside of the `j` loop. Like wise for `i * nc + c) * BS`.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -53,14 +57,25 @@ void set_csr(U&& data, const V& cols, const V& row_ptr, const W& x,
...
+      }
+      else
+      {
+        for (int i = 0; i < bs[0]; ++i)
+          for (int j = 0; j < bs[1]; ++j)
+            data[d * nbs + i * bs[1] + j] = xr[(i * nc + c) * bs[1] + j];
```
garth-wells, line 77: See above comment on hoisting.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -74,17 +89,22 @@ void set_csr(U&& data, const V& cols, const V& row_ptr, const W& x,
 /// to the matrix
 /// @param[in] xrows The row indices of `x`
 /// @param[in] xcols The column indices of `x`
-template <typename U, typename V, typename W, typename X>
+/// @param[in] bs Number of entries in each block
```
garth-wells, line 92: See comment on above function.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -74,17 +89,22 @@ void set_csr(U&& data, const V& cols, const V& row_ptr, const W& x,
...
+template <int BS = -1, typename U, typename V, typename W, typename X>
```
garth-wells, line 93: Document `BS`.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -94,17 +114,79 @@ void add_csr(U&& data, const V& cols, const V& row_ptr, const W& x,
...
       assert(it != cit1);
       std::size_t d = std::distance(cols.begin(), it);
       assert(d < data.size());
-      data[d] += xr[c];
+
+      if constexpr (BS > 0)
+      {
+        for (int i = 0; i < BS; ++i)
+          for (int j = 0; j < BS; ++j)
+            data[d * BS * BS + i * BS + j] += xr[(i * nc + c) * BS + j];
```
garth-wells, line 129: Loop hoisting (see above comment).

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -94,17 +114,79 @@ void add_csr(U&& data, const V& cols, const V& row_ptr, const W& x,
...
+      }
+      else
+      {
+        for (int i = 0; i < bs[0]; ++i)
+          for (int j = 0; j < bs[1]; ++j)
+            data[d * nbs + i * bs[1] + j] += xr[(i * nc + c) * bs[1] + j];
```
garth-wells, line 135: Loop hoisting.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -94,17 +114,79 @@ void add_csr(U&& data, const V& cols, const V& row_ptr, const W& x,
...
+}
+
+/// @brief Add block data to a CSR matrix of blocksize 1
+///
+/// @param[out] data The CSR matrix data
+/// @param[in] cols The CSR column indices
+/// @param[in] row_ptr The pointer to the ith row in the CSR data
+/// @param[in] x The `m` by `n` dense block of values (row-major) to add
+/// to the matrix
+/// @param[in] xrows The block row indices of `x`
```
garth-wells, line 148: Docs need to be expanded to make block size role clear.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -94,17 +114,79 @@ void add_csr(U&& data, const V& cols, const V& row_ptr, const W& x,
...
+      {
+        for (int i = 0; i < bs[0]; ++i)
+          for (int j = 0; j < bs[1]; ++j)
+            data[d * nbs + i * bs[1] + j] += xr[(i * nc + c) * bs[1] + j];
+      }
+    }
+  }
+}
+
+/// @brief Add block data to a CSR matrix of blocksize 1
```
garth-wells, line 141: 'with block size 1'?

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -94,17 +114,79 @@ void add_csr(U&& data, const V& cols, const V& row_ptr, const W& x,
...
+/// @param[out] data The CSR matrix data
+/// @param[in] cols The CSR column indices
+/// @param[in] row_ptr The pointer to the ith row in the CSR data
+/// @param[in] x The `m` by `n` dense block of values (row-major) to add
+/// to the matrix
+/// @param[in] xrows The block row indices of `x`
+/// @param[in] xcols The block column indices of `x`
+/// @param[in] bs Number of entries in each block
+template <typename U, typename V, typename W, typename X>
+void add_csr_blocked(U&& data, const V& cols, const V& row_ptr, const W& x,
```
garth-wells, line 152: It's not clear how this function relates to `add_csr`, which also has a block size.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -133,31 +215,111 @@ class MatrixCSR
 
   /// Insertion functor for setting values in matrix. It is typically
   /// used in finite element assembly functions.
-  /// @param A Matrix to insert into
-  /// @return Function for inserting values into `A`
+  /// @return Function for inserting values into the matrix
   /// @todo clarify setting on non-owned enrties
-  auto mat_set_values()
+  std::function<int(const xtl::span<const std::int32_t>& rows,
```
garth-wells, line 220: This should be changed back to `auto` to allow inlining of the anonymous function. Will need redesign of the function body.

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -133,31 +215,111 @@ class MatrixCSR
...
+      impl::set_csr(_data, _cols, _row_ptr, data, rows, cols, _bs, local_size);
       return 0;
     };
   }
 
   /// Insertion functor for accumulating values in matrix. It is
   /// typically used in finite element assembly functions.
-  /// @param A Matrix to insert into
-  /// @return Function for inserting values into `A`
-  auto mat_add_values()
```
garth-wells, line 154: See above comment on `auto`,

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -368,10 +528,17 @@ class MatrixCSR
...
+    xt::xtensor<T, 2> A = xt::zeros<T>({nrows * _bs[0], ncols * _bs[1]});
     for (std::size_t r = 0; r < nrows; ++r)
       for (int j = _row_ptr[r]; j < _row_ptr[r + 1]; ++j)
-        A(r, _cols[j]) = _data[j];
+      {
+        int row = r * _bs[0];
+        int col = _cols[j] * _bs[1];
+        for (int k = 0; k < _bs[0]; ++k)
+          for (int m = 0; m < _bs[1]; ++m)
+            A(row + k, col + m) = _data[j * _bs[0] * _bs[1] + k * _bs[1] + m];
```
garth-wells, line 548: Hoist loop variables.

cpp/dolfinx/la/SparsityPattern.h
```diff
@@ -70,6 +70,11 @@ class SparsityPattern
   /// Move assignment
   SparsityPattern& operator=(SparsityPattern&& pattern) = default;
 
+  /// Take a SparsityPattern with a block size greater than 1, and
+  /// expand the entries, repeating rows and columns in each block.
+  /// @return SparsityPattern with bs={1, 1}
+  SparsityPattern expand();
```
garth-wells, line 76: Can this be a free function? Would be better if it's not a class member.

What are preconditions? Should the pattern be finalised first? Is there a reason why the function isn't `const`?

garth-wells (changes requested):
First round of comments.

cpp/dolfinx/la/MatrixCSR.h
chrisrichardson: redesign? Not sure how to. If I put `auto` then I get `inconsistent deduction for auto return type`.

chrisrichardson:
```
inconsistent deduction for auto return type: 
‘dolfinx::la::MatrixCSR<T, Allocator>::mat_add_values() [with T = double; Allocator = std::allocator<double>]::<lambda(const tcb::span<const int>&, const tcb::span<const int>&, const tcb::span<const double>&)>’ 
and then 
‘dolfinx::la::MatrixCSR<T, Allocator>::mat_add_values() [with T = double; Allocator = std::allocator<double>]::<lambda(const tcb::span<const int>&, const tcb::span<const int>&, const tcb::span<const double>&)>’
```
which doesn't really make sense, as it says the two identical types are inconsistent. I suppose it really treats each lambda function as a "type" so you can only return one actual lambda function with `auto`, and not choose which one to return.

cpp/dolfinx/la/SparsityPattern.h
chrisrichardson: Yes, maybe better as a free function. Then I don't need to worry about being `const` 😄

cpp/dolfinx/la/MatrixCSR.h
IgorBaratta: I don't think it's possible to have auto there in C++ 17.  Every lambda has a unique type and it's not defined at compile time. 
One alternative I can think of is to template the Matrix<T, BS>. So if BS is known at compile time we can use `cosntexpr if`.

garth-wells:
@chrisrichardson should this PR be closed? Superseded by other changes?

PR #2040, opened 2022-03-05
Remove UniqueIdGenerator class

garth-wells:
Not needed. Can just use object address, or from Python the `id()` function.


garth-wells:
Superseded by an upcoming change.

PR #2051, opened 2022-03-09
Replace discrete operator with arbitrary order operator

jorgensd:
Create general matrix operator $G$ mapping $u\in P^k/Q^k \mapsto  \nabla \phi \in ND_k$.

Generalizes the previous discrete operator, and works for arbitrary order spaces.


jpdean:
Great! This functionality is very useful for solving Maxwell's equations. I'll look at this PR in detail tonight.

It would also be good to implement the discrete curl and discrete divergence operators too at some point in the future.

jorgensd:
> Great! This functionality is very useful for solving Maxwell's equations. I'll look at this PR in detail tonight.
> 
> It would also be good to implement the discrete curl and discrete divergence operators too at some point in the future.

I'll start by implementing the vector Lagrange->Nedelec operator, ref: https://hypre.readthedocs.io/en/latest/solvers-ams.html#high-order-discretizations

The gradient operator has some neat tricks to it, as the inverse covariant piola makes the contribution only dependent on the dof transformation, and not the jacobian, inverse jacobian and detJ.

It would be useful if you could make a list of what operators (and between what spaces) we need for different solvers, it would be very helpful. I guess they follow the de Rahm complex https://defelement.com/de-rham.html

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -68,128 +66,78 @@ void assemble_discrete_gradient(
...
+                           const xtl::span<const std::uint32_t>&, std::int32_t,
+                           int)>
+      apply_inverse_dof_transform
+      = e0->get_dof_transformation_function<T>(true, true, false);
+  mesh->topology_mutable().create_entity_permutations();
+  xtl::span<const std::uint32_t> cell_info
+      = xtl::span(mesh->topology().get_cell_permutation_info());
+  auto dofmap0 = V0.dofmap();
+  auto dofmap1 = V1.dofmap();
+  // Create element kernel
```
garth-wells, line 125: This comment is a bit vague.

In general, comments should have a black line before them for readability.

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -68,128 +66,78 @@ void assemble_discrete_gradient(
...
-
-  // Build discrete gradient operator/matrix
-  const std::shared_ptr<const fem::DofMap> dofmap1 = V1.dofmap();
-  assert(dofmap1);
-  const std::vector<std::int64_t>& global_indices
-      = mesh->topology().index_map(0)->global_indices();
-  std::array<T, 2> Ae;
-  for (std::int32_t e = 0; e < num_edges; ++e)
+  std::array<std::size_t, 4> t_shape
+      = {(std::size_t)tdim + 1, X.shape(0), (std::size_t)ndofs_cell_1,
```
garth-wells, line 98: Use `std::size_t` for the variable to reduce the boiler plate code.

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -68,128 +66,78 @@ void assemble_discrete_gradient(
...
+         (std::size_t)ref_vs_1};
+  xt::xtensor<double, 4> l_basis(t_shape);
+  e1->tabulate(l_basis, X, 1);
+  auto l_dphi
```
garth-wells, line 102: Better variable name?

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -68,128 +66,78 @@ void assemble_discrete_gradient(
...
+      = xt::view(l_basis, xt::xrange(1, tdim + 1), xt::all(), xt::all(), 0);
+  auto dphi = xt::reshape_view(l_dphi, {tdim * t_shape[1], t_shape[2]});
+
+  // Compute local element interpolation matrix
```
garth-wells, line 106: // Get the element interpolation matrix

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -68,128 +66,78 @@ void assemble_discrete_gradient(
...
+  const int bs_0 = e0->block_size();
+  assert(bs_0 == 1);
+  const int ndofs_cell_0 = e0->space_dimension() / bs_0;
+  xt::xtensor<T, 2> Ae = xt::zeros<double>({ndofs_cell_0, ndofs_cell_1});
+  const xt::xtensor<double, 2> Pi = e0->interpolation_operator();
+  math::dot(Pi, dphi, Ae);
+
+  // Get inverse transform
```
garth-wells, line 114: // Get inverse DOF transform function

python/test/unit/fem/test_discrete_operators.py
```diff
@@ -51,6 +52,47 @@ def test_incompatible_spaces():
...
 
-    V = FunctionSpace(mesh, ("Lagrange", 2))
-    with pytest.raises(RuntimeError):
-        create_discrete_gradient(W._cpp_object, V._cpp_object)
+
+@pytest.mark.skip_in_parallel
+@pytest.mark.parametrize("p", range(1, 5))
+@pytest.mark.parametrize("cell_type", [
+    CellType.quadrilateral, CellType.triangle, CellType.tetrahedron, CellType.hexahedron])
+def test_interpolation_matrix(cell_type, p):
```
garth-wells, line 60: Can a parallel check be added? Even if just on the Hypre AMS iteration count.

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -68,128 +66,78 @@ void assemble_discrete_gradient(
...
-
-  // Copy index maps from dofmaps
-  std::array<std::shared_ptr<const common::IndexMap>, 2> index_maps
-      = {{V0.dofmap()->index_map, V1.dofmap()->index_map}};
-  std::vector<std::array<std::int64_t, 2>> local_range
-      = {index_maps[0]->local_range(), index_maps[1]->local_range()};
-  assert(V0.dofmap()->index_map_bs() == V1.dofmap()->index_map_bs());
-
-  // Initialize required connectivities
+  // Tabulate Lagrange space at H(curl) interpolation points
```
garth-wells, line 89: Break up steps with comments. Looks like first step is getting the interpolation points for Lagrange element.

garth-wells (changes requested):
Looks good. Let's iterate on it a bit.

garth-wells:
I think the strings checks should be removed and replaced with a clear warning in the doc strings.

Strings are fragile and it introduced more work for later to remove them.

jorgensd:
> I think the string checks should be removed and replaced with a clear warning in the doc strings.
> 
> Strings are fragile and it introduced more work for later to remove them. We don't need to protect the user from themselves for advanced functionality.

What about the following:
- We check that the mapping-type of the output space is the covariant piola map
- We check that the mapping-type of the input space is the identity
The whole code relies on these assumptions, and would produce garbage if this is not true.
We can get this using `dolfinx::fem::FiniteElement::map_type` (https://github.com/FEniCS/dolfinx/blob/9d11d4408975b8b959be19f4b6c0f08cf4e289a7/cpp/dolfinx/fem/FiniteElement.h#L182)

jorgensd:
> > What about the following:
> > 
> > * We check that the mapping-type of the output space is the covariant piola map
> > * We check that the mapping-type of the input space is the identity
> >   The whole code relies on these assumptions, and would produce garbage if this is not true.
> >   We can get this using `dolfinx::fem::FiniteElement::map_type` (https://github.com/FEniCS/dolfinx/blob/9d11d4408975b8b959be19f4b6c0f08cf4e289a7/cpp/dolfinx/fem/FiniteElement.h#L182
> >   
> >   )
> 
> Anything but strings.
> 
> The point still holds that we don't need to stop a user from shooting themselves in the foot if they mis-use things. Most important is that the docs state the preconditions.

I've added a suggest documentation and assertion.

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -7,56 +7,54 @@
...
-/// and V1 is the P1 Lagrange space.
-///
-/// @param[in] V0 A H(curl) space
-/// @param[in] V1 A P1 Lagrange space
-/// @return The sparsity pattern
+/// @param[in] V0 A q-th order Nedelec (first kind) space
+/// @param[in] V1 A p-th order Lagrange space
+/// @param[in] V0 A degree q Nedelec (first kind) space
+/// @param[in] V1 A degere p Lagrange space
 la::SparsityPattern
```
garth-wells, line 37: This seems like just a re-implementation of `fem::sparsitybuild::cells`, and it doesn't see to exploit any special non-zero structure for the discrete gradient problem.

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -7,56 +7,45 @@
...
-/// and V1 is the P1 Lagrange space.
+/// @warning This function relies on the user supplying appropriate
+/// input and output spaces. See parameter descriptions.
 ///
-/// @param[in] mat_set A function (or lambda capture) to set values in a matrix
-/// @param[in] V0 A H(curl) space
-/// @param[in] V1 A P1 Lagrange space
+/// @param[in] mat_set A functor that sets values in a matrix
+/// @param[in] V0 A Nedelec (first kind) space
+/// @param[in] V1 A Lagrange space
```
jorgensd, line 48: @garth-wells should we say:
```
V1 A scalar Lagrange space
```
?

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -26,75 +26,73 @@ namespace dolfinx::fem
...
 /// @brief Assemble a discrete gradient operator.
 ///
-/// The discrete gradient operator \f$A\f$ takes the gradient of a
-/// Lagrange finite element function \f$v \in V_h \subset H^1\f$ to a
-/// Nedelec (first kind) function \f$w \in W_h \subset H({\rm curl})\f$.
-/// i.e. \f$w=Av\f$ where \f$v\f$ are the Lagrange space
-/// degrees-of-freedom and \f$w\f$ are the curl space
-/// degrees-of-freedom.
-///
+/// The discrete gradient operator \f$A\f$ interpolates take the
```
jorgensd, line 29: This sentence doesn’t sound quite right anymore.
`…  \f$A\f$ interpolates take the …`

garth-wells (approved):

PR #2056, opened 2022-03-11
Add Python interface for ADIOS2-based output formats

jorgensd:
Both writers are now nicely exposed in Python.

Make sure Fides writer throws unsupported error for DG elements.

Fix a typo in coordinate element


python/dolfinx/io.py
```diff
@@ -15,10 +15,55 @@
 from dolfinx import cpp as _cpp
 from dolfinx.cpp.io import distribute_entity_data
 from dolfinx.cpp.io import perm_gmsh as cell_perm_gmsh
+from dolfinx.common import has_adios2
```
garth-wells, line 18: I don't think we this. Python way is try/except rather than checking that ADIOS2 is availably.

python/dolfinx/io.py
```diff
@@ -15,10 +15,55 @@
...
+        """Interface to VTK files for ADIOS 2
+
+        VTX supports arbitrary order Lagrangian finite elements for the geometry description.
+        VTX supports arbitrary order (discontinous) Lagrangian finite elements for functions.
+        All functions has to be of the same element family and same order.
+
+        """
+
+        def __init__(self, comm: _MPI.Comm, filename: str, output: typing.Union[Mesh, typing.List[Function], Function]):
+            """
```
garth-wells, line 36: Document arguments (Google Doc style).

python/dolfinx/io.py
```diff
@@ -15,10 +15,55 @@
...
+            Initialize a writer for outputting a mesh, a single (discontinuous) Lagrange function or
+            list of (discontinuous Lagrange functions sharing the same element familty and degree
+            """
+            if isinstance(output, Mesh):
```
garth-wells, line 40: Avoid `isinstance`. Use try/except instead.

python/dolfinx/io.py
```diff
@@ -15,10 +15,55 @@
...
+                super().__init__(comm, filename, output)
+            elif isinstance(output, (list, tuple)):
+                super().__init__(comm, filename, [getattr(u, "_cpp_object", u) for u in output])
+            else:
+                super().__init__(comm, filename, getattr(output, "_cpp_object", output))
+
+    class FidesWriter(_cpp.io.FidesWriter):
+        """Interface to Fides data structures for ADIOS 2
```
garth-wells, line 48: Document args.

python/dolfinx/io.py
```diff
@@ -15,10 +15,55 @@
...
+
+        Fides supports first order Lagrangian finite elements for the geometry description.
+        Fides supports first order Lagrangian finite elements for functions.
+        All functions has to be of the same element family and same order.
+
+        """
+
+        def __init__(self, comm: _MPI.Comm, filename: str, output: typing.Union[Mesh, typing.List[Function], Function]):
+            """
```
garth-wells, line 57: Document args.

python/dolfinx/io.py
```diff
@@ -15,10 +15,55 @@
...
+            Initialize a writer for outputting a mesh, a single Lagrange function or
+            list of Lagrange functions sharing the same element familty and degree
+            """
+            if isinstance(output, Mesh):
```
garth-wells, line 61: Avoid `isinstance`.

python/dolfinx/io.py
```diff
@@ -15,10 +15,55 @@
...
+                super().__init__(comm, filename, output)
+            elif isinstance(output, (list, tuple)):
+                super().__init__(comm, filename, [getattr(u, "_cpp_object", u) for u in output])
```
garth-wells, line 64: What id the list is mixed Python and C++ `Function` objects? I would suggest a utility to extract a list of pybind11 wrapped `Functions`.

garth-wells (changes requested):

python/dolfinx/io.py
jorgensd: All functions (python or C++) can be captured with `getattr()`. I've added a standalone function that is reused in the different classes.

PR #2059, opened 2022-03-12
Use scalable distribution of graph data

garth-wells:


cpp/dolfinx/common/MPI.h
```diff
@@ -125,8 +178,120 @@ std::vector<int> compute_graph_edges_pcx(MPI_Comm comm,
 std::vector<int> compute_graph_edges_nbx(MPI_Comm comm,
                                          const xtl::span<const int>& edges);
 
+/// @brief Distribute row data to 'post office' rank.
+///
+/// The post office rank for a row is determined by applying
+/// MPI::index_owner global index of the row.
```
chrisrichardson, line 184:
```suggestion
/// MPI::index_owner to the global index of the row.
```

cpp/dolfinx/common/MPI.h
```diff
@@ -125,8 +178,120 @@ std::vector<int> compute_graph_edges_pcx(MPI_Comm comm,
...
+template <typename T>
+std::pair<std::vector<std::int32_t>, std::vector<T>>
+distribute_to_postoffice(MPI_Comm comm, const xtl::span<const T>& x,
+                         std::array<std::int64_t, 2> shape,
+                         std::int64_t rank_offset);
+
+/// @brief Distribute rows of a rectangular data array from post office
+/// ranks to ranks where they are required.
+///
+/// This functions determines local neighborhoods for communication, and
```
chrisrichardson, line 204:
```suggestion
/// This function determines local neighborhoods for communication, and
```

cpp/dolfinx/common/MPI.h
```diff
@@ -125,8 +178,120 @@ std::vector<int> compute_graph_edges_pcx(MPI_Comm comm,
...
+std::vector<T> distribute_data(MPI_Comm comm,
+                               const xtl::span<const std::int64_t>& indices,
+                               const xtl::span<const T>& x, int shape1);
+
+/// @brief Distribute rows of a rectangular data array to ranks where
+/// they are required (non-scalable version).
+///
+/// See MPI::distribute_data description.
+///
+/// @note This functions used MPI all-to-all collectives and is
```
chrisrichardson, line 266:
```suggestion
/// @note This function uses MPI all-to-all collectives and is
```

cpp/dolfinx/common/MPI.h
```diff
@@ -229,6 +343,466 @@ constexpr MPI_Datatype mpi_type()
...
+distribute_to_postoffice(MPI_Comm comm, const xtl::span<const T>& x,
+                         std::array<std::int64_t, 2> shape,
+                         std::int64_t rank_offset)
+{
+  const int size = dolfinx::MPI::size(comm);
+  const int rank = dolfinx::MPI::rank(comm);
+  assert(x.size() % shape[1] == 0);
+  const std::int32_t shape0_local = x.size() / shape[1];
+
+  LOG(2) << "Sending data to post offices (distribute_to_postoffice)";
```
chrisrichardson, line 356:
```suggestion
  LOG(INFO) << "Sending data to post offices (distribute_to_postoffice)";
```

chrisrichardson, line 356:
```suggestion
  LOG(INFO) << "Sending data to post offices (distribute_to_postoffice)";
```

cpp/dolfinx/common/MPI.h
```diff
@@ -229,6 +343,466 @@ constexpr MPI_Datatype mpi_type()
...
+  std::vector<T> recv_buffer_data(shape[1] * recv_disp.back());
+  MPI_Neighbor_alltoallv(send_buffer_data.data(), num_items_per_dest.data(),
+                         send_disp.data(), compound_type,
+                         recv_buffer_data.data(), num_items_recv.data(),
+                         recv_disp.data(), compound_type, neigh_comm);
+  MPI_Type_free(&compound_type);
+
+  MPI_Comm_free(&neigh_comm);
+
+  LOG(2) << "Completed send data to post offices.";
```
chrisrichardson, line 477:
```suggestion
  LOG(INFO) << "Completed send data to post offices.";
```

chrisrichardson (commented):
I think `distribute_to_postoffice` needs more explanation. I can't see the purpose of it from the description. It seems as if the data is already distributed across processes. If each process already has the same number of rows of `x` then `distribute_to_postoffice` is a no-op?

garth-wells:
> I think `distribute_to_postoffice` needs more explanation. I can't see the purpose of it from the description. It seems as if the data is already distributed across processes. If each process already has the same number of rows of `x` then `distribute_to_postoffice` is a no-op?

Yes, which is why I split it into a separate function. An optimisation would be to skip the send-to-postoffice step if the data is know to already be distributed.

Issue #60, opened 2018-05-04
Use scalable dynamic discovery of sparse communication graphs and use sparse communication (MPI)


chrisrichardson:
In a few places, we need to find out where several processes share the same index.
This is done by dividing the index range, and using all_to_all to send to the "index_owner", followed by another all_to_all to send the results back to the origin. 
Can we find a better algorithm?
Can we optimise based on sending to a subset of "post offices", instead of all processes?

wence-:
I think this paper is relevant:
```
@inproceedings{hoefler-dsde,
  author={T. Hoefler and C. Siebert and A. Lumsdaine},
  title={{Scalable Communication Protocols for Dynamic Sparse Data Exchange}},
  year={2010},
  month={Jan.},
  pages={159--168},
  booktitle={Proceedings of the 2010 ACM SIGPLAN Symposium on Principles and Practice of Parallel Programming (PPoPP'10)},
  location={Bangalore, India},
  publisher={ACM},
  isbn={978-1-60558-708-0},
  source={http://www.unixer.de/~htor/publications/},
}```

garth-wells:
The algorithms from Hoefler et al have been added in #2050. This issue can be closed once the new function are used in more places, removing unnecessary all-to-all calls.

PR #2067, opened 2022-03-15
Scalable dual graph construction

garth-wells:
See comments on #2059.

Another step towards #60.


garth-wells:
Superseded by #2080.

PR #2071, opened 2022-03-16
Return empty permutations on ranks with no cells or facets

jpdean:
This PR allows empty permutation vectors to be returned on ranks with no cells or facets, and adds a test which fails on main.


garth-wells (approved):
We should look at this again later - maybe we need to return a pointer. A pointer can carry the  extra information of whether or not something has been done.

PR #2073, opened 2022-03-17
Avoid using count/id to name `fem.Functions` in Python

### LINKED ISSUE #2061: paraview crashed when loading hdf5 file
Hello,

I created a singularity container based on the docker container on the docker hub (both dolfinx/dolfinx and dolfinx/lab). I can run the python demos, such as poisson and elasticity and generate the output files. But I cannot open them in paraview. I tried different versions of paraview from 5.8 to 5.10 and have similar errors. It will crash when I try to load the output file.

What are the possible reasons for this? Is this related to the transfer from docker container to singularity container? I am running simulations on computer clusters and thus can only use singularity. 

Thanks.

Best,
Teng

jorgensd:
Resolves #2061

### DIFF
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -249,7 +249,7 @@ class Function(ufl.Coefficient):
 
         # Set name
         if name is None:
-            self.name = "f_{}".format(self.count())
+            self.name = "f"
         else:
             self.name = name
 
--- a/python/test/unit/fem/test_function.py
+++ b/python/test/unit/fem/test_function.py
@@ -45,7 +45,7 @@ def Q(mesh):
 def test_name_argument(W):
     u = Function(W)
     v = Function(W, name="v")
-    assert u.name == "f_{}".format(u.count())
+    assert u.name == "f"
     assert v.name == "v"
     assert str(v) == "v"
 


garth-wells:
What happens if one tries to write two Functions in the same file, e.g. with `VTXFile`?

jorgensd:
> What happens if one tries to write two Functions in the same file, e.g. with `VTXFile`?

We already have this issue when writing multiple functions from the C++ layer, as we have a default name there. It would lead to non-readable files. I think we should revise the Io interface to take in the name and the function (dictionary style, as we do with dolfinx::fem::Form, and not rely on the name associated with the function, as it would lead to having string comparison.

garth-wells (approved):

garth-wells:
> > What happens if one tries to write two Functions in the same file, e.g. with `VTXFile`?
> 
> We already have this issue when writing multiple functions from the C++ layer, as we have a default name there. It would lead to non-readable files. I think we should revise the Io interface to take in the name and the function (dictionary style, as we do with dolfinx::fem::Form, and not rely on the name associated with the function, as it would lead to having string comparison.

Nothing wrong with a string comparison in this case since the name is defined by a string. No need for a complicated solution - we can just check in the IO function that string identifiers are unique and throw an error is they are not.

PR #2078, opened 2022-03-18
Fix interpolation into serendipity

### LINKED ISSUE #2076: Interpolation into serendipity spaces is incorrect
Dashed line is sin(pi * x / 2). Solid line is current interpolation of this in DOLFINx using order 2 serendipity on a single cell.

The interpolation is correct when done with Basix alone.

See failing test in the branch https://github.com/FEniCS/dolfinx/tree/mscroggs/serendipity-interpolate

mscroggs:
Added a new interpolation code path for elements with value size 1 whose DOFs are not all point evaluations.

Fixes #2076.


cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -189,6 +189,11 @@ class FiniteElement
   /// @return True if interpolation is an identity operation
   bool interpolation_ident() const noexcept;
 
+  /// Check if the map from the reference to a physical cell for this element is
```
garth-wells, line 192: I don't quite follow. When is the map ever the identity? The reference and physically cells have different geometry.

mscroggs: I mean the push forward or pull back map is the identity. Perhaps a better name is needed

garth-wells: @jpdean might have a suggestion for a name.

mscroggs: I've updated the docstring to say this is the push forward/pull back map we're talking about. Happy to take suggestions for a better name

jpdean: Sorry, I didn't see this! I think calling it a pullback/pushforward is fine.

PR #2081, opened 2022-03-20
Do not overwrite UFL id in Function

### LINKED ISSUE #2074: #2044 may have made UFL forms non deterministic
Hi,
in one of the downstream libraries I develop I do some manipulations of UFL forms, and I have some unit tests that check the output of the postprocessed forms.

One upstream change that happened between March 5 and March 10 has made this process non deterministic. Looking at the commit history, I think #2044 may have been the culprit.

Consider the following MWE
```
import dolfinx
import mpi4py
import mpi4py.MPI
import ufl

mesh = dolfinx.mesh.create_unit_square(mpi4py.MPI.COMM_WORLD, 2, 2)
V = dolfinx.fem.FunctionSpace(mesh, ("Lagrange", 1))

u = ufl.TrialFunction(V)
v = ufl.TestFunction(V)
f1 = dolfinx.fem.Function(V, name="coefficient 1")
f2 = dolfinx.fem.Function(V, name="coefficient 2")

dx = ufl.dx
grad = ufl.grad
inner = ufl.inner

a1 = f1 * f2 * inner(grad(u), grad(v)) * dx
print(a1)
```

Since March 10, the print may randomly be either
```
{ coefficient 1 * coefficient 2 * (conj(((grad(v_0)) : (grad(v_1))))) } * dx(<Mesh #0>[everywhere], {})
```
or
```
{ coefficient 2 * coefficient 1 * (conj(((grad(v_0)) : (grad(v_1))))) } * dx(<Mesh #0>[everywhere], {})
```

Surely for FFCx purposes the two outputs are equivalent, but still it would be nice to restore a deterministic behavior.

Will #2073 further change this behavior, since two unnamed `dolfinx.fem.Function` will end up with the same name `f`?

michalhabera:
Should fix #2074 and #2061 . UFL id generated with unique global generator in UFL should not be overwritten with
dolfinx's Function id. Equality of UFL objects is different to equality of dolfinx cpp objects.

This will make UFL form in #2074 deterministic again (python `id()` is not) and file IO should work as well, as Function names contain the deterministic UFL id.

### DIFF
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -245,7 +245,7 @@ class Function(ufl.Coefficient):
             self._cpp_object = functiontype(dtype)(V._cpp_object)
 
         # Initialize the ufl.FunctionSpace
-        super().__init__(V.ufl_function_space(), count=id(self))
+        super().__init__(V.ufl_function_space())
 
         # Set name
         if name is None:
--- a/python/test/unit/io/test_xdmf_function.py
+++ b/python/test/unit/io/test_xdmf_function.py
@@ -145,7 +145,6 @@ def test_save_3d_vector_series(tempdir, encoding, cell_type):
         u.vector.set(2.0 + (2j if np.issubdtype(PETSc.ScalarType, np.complexfloating) else 0))
         file.write_function(u, 0.2)
 
-    # NOTE: This is an issue with appending in parallel
-    # with XDMFFile(mesh.comm, filename, "a", encoding=encoding) as file:
-    #     u.vector.set(3.0 + (3j if np.issubdtype(PETSc.ScalarType, np.complexfloating) else 0))
-    #     file.write_function(u, 0.3)
+    with XDMFFile(mesh.comm, filename, "a", encoding=encoding) as file:
+        u.vector.set(3.0 + (3j if np.issubdtype(PETSc.ScalarType, np.complexfloating) else 0))
+        file.write_function(u, 0.3)


garth-wells:
See #2073 and the comment https://github.com/FEniCS/dolfinx/pull/2073#issuecomment-1073064767.

michalhabera:
> See #2073 and the comment [#2073 (comment)](https://github.com/FEniCS/dolfinx/pull/2073#issuecomment-1073064767).

Dolfinx IO uses Function name as unique identifier, which uses UFL id. Two functions could be written into a file (XDMFFile for sure) without issues, unlike in the current `main` where function names could be different on different MPI ranks due to non-deterministic names (and required you to comment the respective unit test...). What is the problem?

garth-wells:
> > See #2073 and the comment [#2073 (comment)](https://github.com/FEniCS/dolfinx/pull/2073#issuecomment-1073064767).
> 
> Dolfinx IO uses Function name as unique identifier, which uses UFL id. Two functions could be written into a file (XDMFFile for sure) without issues, unlike in the current `main` where function names could be different on different MPI ranks due to non-deterministic names (and required you to comment the respective unit test...). What is the problem?

Do the two function get given the same name in the output file?

michalhabera:
> > > See #2073 and the comment [#2073 (comment)](https://github.com/FEniCS/dolfinx/pull/2073#issuecomment-1073064767).
> > 
> > 
> > Dolfinx IO uses Function name as unique identifier, which uses UFL id. Two functions could be written into a file (XDMFFile for sure) without issues, unlike in the current `main` where function names could be different on different MPI ranks due to non-deterministic names (and required you to comment the respective unit test...). What is the problem?
> 
> Do the two function get given the same name in the output file?

In python, the default name is the one in the form `f_{ufl_id}`, so two functions with default names could be stored automatically into the same output file.
In C++, the default name naturally does not have access to UFL global id, so that one is the same and would cause IO issues unless name is explicitly specified to C++ Function constructor.

We can solve the IO naming issues later, but my PR mainly resolves the equality notion for UFL objects. As @francesco-ballarin said, UFL's id should preferably not be non-deterministic because it randomly fails various UFL expressions equality checks.

garth-wells (approved):

PR #2084, opened 2022-03-22
Change empty default lists to `None`

jorgensd:
Non-compliant python.


francesco-ballarin:
@jorgensd should the same change be applied to https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/fem/petsc.py as well?

jorgensd:
> @jorgensd should the same change be applied to https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/fem/petsc.py as well?

Yes, somehow i missed those when i was search-replacing

PR #2086, opened 2022-03-22
Fix for exterior facet integrals on a submesh

jpdean:
This PR fixes a bug where some exterior facets are not detected in `set_default_domains` when one process owns all the cells in a submesh but other processes ghost some of those cells.  It also adds a test that fails on main.

### DIFF
--- a/cpp/dolfinx/fem/Form.h
+++ b/cpp/dolfinx/fem/Form.h
@@ -577,9 +577,10 @@ private:
         assert(c_to_f);
 
         // Only need to consider shared facets when there are no ghost
-        // cells and none of cells owned by this process are shared (the
+        // cells and none of cells owned by this process are shared. The
         // latter check is required because a submesh could have no ghost
-        // cells but another process could share one of this process cells)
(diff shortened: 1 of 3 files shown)


cpp/dolfinx/fem/Form.h
```diff
@@ -577,10 +577,15 @@ class Form
...
 
         // Only need to consider shared facets when there are no ghost
-        // cells
+        // cells and none of cells owned by this process are shared (the
+        // latter check is required because a submesh could have no ghost
+        // cells but another process could share one of this process cells)
         std::set<std::int32_t> fwd_shared_facets;
         assert(topology.index_map(tdim - 1));
-        if (topology.index_map(tdim)->num_ghosts() == 0)
+        const bool owned_cells_shared =
```
garth-wells, line 585:
```cpp
const bool owned_cells_shared = !topology.index_map(tdim)->scatter_fwd_indices().array().empty();
```

garth-wells (approved):

PR #2094, opened 2022-03-26
Remove `common::subsystem`

garth-wells:
This was a left over from when we had a singleton class that automatically (mis)managed initialisation and finalisation of MPI, PETSc, etc. What was left was no less verbose than calling `MPI_Init`, `PetscInitialize`, etc, but obscured what is actually happening. This PR removes the wrappers around MPI and PETSc initialisation.

The Python interface is unaffected.


IgorBaratta (approved):

cpp/dolfinx/common/log.cpp
```diff
@@ -6,3 +6,27 @@
...
+//-----------------------------------------------------------------------------
+void dolfinx::init_logging(int argc, char* argv[])
+{
+  loguru::g_stderr_verbosity = loguru::Verbosity_WARNING;
+
+#ifndef NDEBUG
+  loguru::SignalOptions signals;
+#else
+  loguru::SignalOptions signals = loguru::SignalOptions::none();
+#endif
```
IgorBaratta, line 20: I found it a bit hard to read `if not defined "do not debug"`:
```suggestion
#ifdef NDEBUG
  loguru::SignalOptions signals = loguru::SignalOptions::none();
#else
  loguru::SignalOptions signals;
#endif
```

chrisrichardson (approved):

PR #2101, opened 2022-03-29
Add access to original cell index

chrisrichardson:
Simple access to stored "original cell index"


python/dolfinx/wrappers/mesh.cpp
```diff
@@ -293,6 +293,8 @@ void mesh(py::module& m)
            })
       .def_property_readonly("dim", &dolfinx::mesh::Topology::dim,
                              "Topological dimension")
+      .def_readonly("original_cell_index",
```
garth-wells, line 296: This needs to be wrapped properly as a NumPy array.

PR #2107, opened 2022-03-30
Submesh entity map fix

jpdean:
Currently, the list of entities passed to `create_submesh` acts as the map from submesh entities to mesh entities. However, this can be problematic depending on how the list of entities is created. For instance, if this is obtained using `locate_entities_boundary`, then the list will only contained owned entities, and not ghost entities that might be present in the submesh. This can lead to, for instance, `cell_index_map.size_local + cell_index_map.num_ghosts` not being equal to  `cell_to_vertex_connectivity.num_nodes`.

This PR fixed this issue by returning an entity map which includes any ghost entities present in the submesh. It also adds a test that fails on main.


garth-wells (approved):

cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -258,6 +259,20 @@ mesh::create_submesh(const Mesh& mesh, int dim,
...
+  // entities in the mesh.
+  std::vector<int32_t> submesh_to_mesh_entity_map(
+      submesh_owned_entities.begin(), submesh_owned_entities.end());
+  submesh_to_mesh_entity_map.reserve(submesh_entity_index_map->size_local()
+                                     + submesh_entity_index_map->num_ghosts());
+  // Add ghost vertices to the map
+  std::transform(submesh_entity_index_map_pair.second.begin(),
+                 submesh_entity_index_map_pair.second.end(),
+                 std::back_inserter(submesh_to_mesh_entity_map),
+                 [mesh_entity_index_map](std::int32_t entity_index) {
```
garth-wells, line 272: `[mesh_entity_index_map]` -> `[size_local=mesh_entity_index_map->size_loca()]`

garth-wells, line 272: `[mesh_entity_index_map]` -> `[size_local=mesh_entity_index_map->size_loca()]`

cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -258,6 +259,20 @@ mesh::create_submesh(const Mesh& mesh, int dim,
...
+                   return mesh_entity_index_map->size_local() + entity_index;
```
garth-wells, line 273: `mesh_entity_index_map->size_local()` -> `size_local`.

garth-wells, line 273: `mesh_entity_index_map->size_local()` -> `size_local`.

PR #2108, opened 2022-03-31
Minor improvements and removal of unused includes

jorgensd:


cpp/dolfinx/refinement/plaza.cpp
```diff
@@ -318,9 +318,9 @@ face_long_edge(const mesh::Mesh& mesh)
   for (std::size_t e = 0; e < edge_length.size(); ++e)
   {
     // Get first attached cell
-    assert(e_to_c->num_links(e) > 0);
-    const std::int32_t c = e_to_c->links(e)[0];
-    auto cell_vertices = c_to_v->links(c);
+    auto cells = e_to_c->links(e);
+    assert(!cells.empty());
+    auto cell_vertices = c_to_v->links(cells[0]);
```
garth-wells, line 323:
```suggestion
    auto cell_vertices = c_to_v->links(cells.front());
```

cpp/dolfinx/refinement/plaza.cpp
```diff
@@ -333,10 +333,16 @@ face_long_edge(const mesh::Mesh& mesh)
     assert(it1 != cell_vertices.end());
     const std::size_t local1 = std::distance(cell_vertices.begin(), it1);
 
-    auto x_dofs = x_dofmap.links(c);
+    auto x_dofs = x_dofmap.links(cells[0]);
```
garth-wells, line 336:
```suggestion
    auto x_dofs = x_dofmap.links(cells.front());
```

cpp/dolfinx/refinement/utils.cpp
```diff
@@ -217,7 +216,7 @@ refinement::create_new_vertices(
   std::map<std::int32_t, std::int64_t> local_edge_to_new_vertex;
   for (int local_i = 0; local_i < edge_index_map->size_local(); ++local_i)
   {
-    if (marked_edges[local_i] == true)
+    if (marked_edges[local_i])
```
garth-wells, line 219: I'm not sure about this given that `marked_edges` is not `bool`.

garth-wells (changes requested):

cpp/dolfinx/refinement/utils.cpp
jorgensd: We use this logic everywhere else in the code. See for instance
https://github.com/FEniCS/dolfinx/blob/a91c7fbb4e61f99b7aa20b519021e38d051a9aab/cpp/dolfinx/refinement/plaza.cpp#L72-L83

cpp/dolfinx/refinement/plaza.cpp
```diff
@@ -333,10 +328,16 @@ face_long_edge(const mesh::Mesh& mesh)
...
+    auto x_dofs = x_dofmap.links(cells.front());
+    xtl::span<const double, 3> x0(
+        mesh.geometry().x().data() + 3 * x_dofs[local0], 3);
+    xtl::span<const double, 3> x1(
+        mesh.geometry().x().data() + 3 * x_dofs[local1], 3);
+
+    // Compute length of edge between vertex x0 and x1
+    edge_length[e] = std::sqrt(std::transform_reduce(
+        x0.begin(), x0.end(), x1.begin(), 0.0, std::plus<>(),
+        [](auto x0, auto x1) { return (x0 - x1) * (x0 - x1); }));
```
chrisrichardson, line 340: Can we skip the `sqrt` ?

garth-wells: > Can we skip the `sqrt` ?

Is it not needed?

jorgensd: I dont think so, it is just a measure of the longest edge. I guess this would just change the quality ratio down the line when deciding if we do uniform refinement of a cell in 2D

chrisrichardson: Yes, I suppose so. But maybe we should just remove this feature, as it is not very important, adds complexity, and only works in 2D.

PR #2111, opened 2022-03-31
Move to C++20

garth-wells:
Requires https://github.com/FEniCS/basix/pull/563.

Removes `xtl` dependency in favour of `std::span`. 

GCC >= 10 required. Will allow code simplifications by using ranges and concepts from C++20.


cpp/dolfinx/geometry/BoundingBoxTree.cpp
```diff
@@ -283,7 +280,7 @@ BoundingBoxTree::BoundingBoxTree(
   }
 
   LOG(INFO) << "Computed bounding box tree with " << num_bboxes()
-            << " nodes for " << num_leaves << " points.";
+            << " nodes for " << points.size() << " points.";
```
michalhabera, line 283: This makes `num_leaves` above unused.

michalhabera (approved):

cpp/dolfinx/geometry/BoundingBoxTree.cpp
garth-wells: It's used in the `if` statement just above.

IgorBaratta (approved):
Looks good to me.

My only concern are features that are not well supported across compilers, see
https://en.cppreference.com/w/cpp/compiler_support.
Which is not the case of this PR.

PR #2120, opened 2022-04-05
Fix locate entities on SubMesh where one process owns all cells

jorgensd:
Similar fix to #2086

### DIFF
--- a/cpp/dolfinx/mesh/Topology.cpp
+++ b/cpp/dolfinx/mesh/Topology.cpp
@@ -705,9 +705,17 @@ std::vector<std::int8_t> mesh::compute_boundary_facets(const Topology& topology)
   auto facets = topology.index_map(tdim - 1);
   if (!facets)
     throw std::runtime_error("Facets have not been computed.");
+  auto cells = topology.index_map(tdim);
 
+  // Only need to consider shared facets when there are no ghost
+  // cells and none of cells owned by this process are shared. The
+  // latter check is required because a submesh could have no ghost
+  // cells on this process but another process could ghost some of
+  // those cells)
   std::vector<std::int32_t> fwd_shared_facets;
-  if (dolfinx::MPI::size(facets->comm()) > 1 and facets->num_ghosts() == 0)
+  if (dolfinx::MPI::size(facets->comm()) > 1 and facets->num_ghosts() == 0
+      and topology.index_map(tdim)->num_ghosts() == 0
+      and topology.index_map(tdim)->scatter_fwd_indices().array().empty())
   {
     fwd_shared_facets.assign(facets->scatter_fwd_indices().array().begin(),
                              facets->scatter_fwd_indices().array().end());
--- a/python/test/unit/mesh/test_mesh.py
+++ b/python/test/unit/mesh/test_mesh.py
@@ -18,7 +18,8 @@ from dolfinx import mesh as _mesh
 from dolfinx.cpp.mesh import (create_cell_partitioner, entities_to_geometry,
                               is_simplex)
 from dolfinx.fem import assemble_scalar, form
-from dolfinx.mesh import (CellType, DiagonalType, GhostMode, create_box,
+from dolfinx.mesh import (CellType, DiagonalType, GhostMode,
+                          compute_boundary_facets, create_box,
                           create_rectangle, create_submesh, create_unit_cube,
                           create_unit_interval, create_unit_square,
                           locate_entities, locate_entities_boundary)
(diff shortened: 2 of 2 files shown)


cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -705,9 +705,17 @@ std::vector<std::int8_t> mesh::compute_boundary_facets(const Topology& topology)
...
 
+  // Only need to consider shared facets when there are no ghost
+  // cells and none of cells owned by this process are shared. The
+  // latter check is required because a submesh could have no ghost
+  // cells on this process but another process could ghost some of
+  // those cells)
   std::vector<std::int32_t> fwd_shared_facets;
-  if (dolfinx::MPI::size(facets->comm()) > 1 and facets->num_ghosts() == 0)
+  if (dolfinx::MPI::size(facets->comm()) > 1 and facets->num_ghosts() == 0
+      and topology.index_map(tdim)->num_ghosts() == 0
```
garth-wells, line 717: `topology.index_map(tdim)` has been assigned to `cells` above, so `cells` can be used here.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -705,9 +705,17 @@ std::vector<std::int8_t> mesh::compute_boundary_facets(const Topology& topology)
   auto facets = topology.index_map(tdim - 1);
   if (!facets)
     throw std::runtime_error("Facets have not been computed.");
+  auto cells = topology.index_map(tdim);
```
garth-wells, line 708: Add `assert(cells)`.

python/test/unit/mesh/test_mesh.py
```diff
@@ -578,3 +579,16 @@ def test_original_index():
     s = sum(mesh.topology.original_cell_index)
     s = MPI.COMM_WORLD.allreduce(s, MPI.SUM)
     assert(s == nx**3 * 6 * (nx**3 * 6 - 1) // 2)
+
+
+def test_boundary_identification():
```
garth-wells, line 584: Add a docstring - it's not clear from the function name what precisely it's testing.

python/test/unit/mesh/test_mesh.py
```diff
@@ -578,3 +579,16 @@ def test_original_index():
...
+    nx = 8
+    mesh = create_unit_square(MPI.COMM_WORLD, nx, nx, ghost_mode=GhostMode.shared_facet)
+    left_cells = locate_entities(mesh, mesh.topology.dim, lambda x: x[0] <= 0.5 + 5e-16)
```
garth-wells, line 587: Use `np.isclose` in the lambda.

python/test/unit/mesh/test_mesh.py
```diff
@@ -578,3 +579,16 @@ def test_original_index():
...
+    sub_mesh = create_submesh(mesh, mesh.topology.dim, left_cells)
```
garth-wells, line 588: `sub_mesh, _, _, _ = create....` would make the code clearer.

garth-wells (changes requested):

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -705,9 +705,17 @@ std::vector<std::int8_t> mesh::compute_boundary_facets(const Topology& topology)
...
 
+  // Only need to consider shared facets when there are no ghost
+  // cells and none of cells owned by this process are shared. The
+  // latter check is required because a submesh could have no ghost
+  // cells on this process but another process could ghost some of
+  // those cells)
   std::vector<std::int32_t> fwd_shared_facets;
-  if (dolfinx::MPI::size(facets->comm()) > 1 and facets->num_ghosts() == 0)
+  if (dolfinx::MPI::size(facets->comm()) > 1 and facets->num_ghosts() == 0
```
jpdean, line 716: I don't think the check `dolfinx::MPI::size(facets->comm()) > 1` is actually necessary.

What is more concerning is that I think the check `facets->num_ghosts() == 0` is actually wrong. For a `GhostMode.none` mesh, `fwd_shared_facets` should be populated, otherwise some facets will incorrectly be identified as being on the boundary. However, a `GhostMode.none` mesh can have ghost facets. Am I missing something? If not, this bug will be affecting main too.

python/test/unit/mesh/test_mesh.py
```diff
@@ -578,3 +579,16 @@ def test_original_index():
...
+    sub_mesh[0].topology.create_entities(sub_mesh[0].topology.dim - 1)
+    sub_mesh[0].topology.create_connectivity(sub_mesh[0].topology.dim - 1, sub_mesh[0].topology.dim)
+
+    right_facets = compute_boundary_facets(sub_mesh[0].topology)
```
jpdean, line 592: I believe, because of the issue I mentioned above, some facets will be incorrectly marked as being on the boundary on some ranks for a `GhostMode.none` mesh. This would be the case even for a mesh created using `create_unit_square`.

jpdean (changes requested):

jpdean:
See https://github.com/FEniCS/dolfinx/pull/2124

PR #2124, opened 2022-04-12
Fix `compute_boundary_facets` for `GhostMode.none`

jpdean:
This PR fixes a bug in `compute_boundary_facets` for `GhostMode.none` meshes, where some facets would be incorrectly identified as boundary facets. It also addresses the same issue  as  https://github.com/FEniCS/dolfinx/pull/2120, and removes some duplicate code.

@jorgensd perhaps we could add a more general version of the test in https://github.com/FEniCS/dolfinx/pull/2120 to make sure this isn't accidentally broken again?

### DIFF
--- a/cpp/dolfinx/mesh/Topology.cpp
+++ b/cpp/dolfinx/mesh/Topology.cpp
@@ -710,12 +710,18 @@ std::vector<std::int8_t> mesh::compute_boundary_facets(const Topology& topology)
   // Should always have cell index map
   assert(cell_imap);
 
-  // TODO UPDATE COMMENT
-  // Only need to consider shared facets when there are no ghost
-  // cells and none of cells owned by this process are shared. The
-  // latter check is required because a submesh could have no ghost
-  // cells on this process but another process could ghost some of
-  // those cells)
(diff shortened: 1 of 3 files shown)


garth-wells (approved):

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -702,36 +702,49 @@ graph::AdjacencyList<std::int32_t> convert_to_local_indexing(
...
+  // In case (i), checking that a facet is connected to only one cell is
+  // sufficient to identify it as a boundary facet. In case (ii), we must
+  // additionally check that the facet is not shared with another process to
+  // differentiate between the partition boundary and the physical boundary.
+  //
+  // NOTE: It is not sufficient to only check that a mesh has no ghost cells
+  // to determine if it falls into category (i) or (ii). This is because a
+  // submesh could have no ghost cells and no shared facets, but could share
+  // some cells with other processes.
+  std::set<std::int32_t> fwd_shared_facets;
```
garth-wells, line 725: Can we avoid `std::set` and use `std::vector` + sort + unique?

PR #2131, opened 2022-04-14
Create submodule for gmsh tools

### LINKED ISSUE #1152: Move gmsh Python utility functions to separate module

jorgensd:
Resolves #1152


jorgensd:
@garth-wells Is anything stopping this from getting merged?

python/dolfinx/io/__init__.py
```diff
@@ -0,0 +1,18 @@
+# Copyright (C) 2022 Jørgen S. Dokken
+#
+# This file is part of DOLFINx (https://www.fenicsproject.org)
+#
+# SPDX-License-Identifier:    LGPL-3.0-or-later
+"""Tools for input/output (IO)."""
+
+from dolfinx import cpp as _cpp
+from dolfinx.cpp.io import distribute_entity_data  # noqa: F401
+from dolfinx.io import gmsh  # noqa: F401
```
garth-wells, line 10: Maybe `from dolfinx.io import gmsh as _gmsh`? I think there is a risk of confusing `gmsh` with the actual Gmsh module.

python/dolfinx/io/gmsh.py
```diff
@@ -0,0 +1,120 @@
...
+
+import ufl
+from dolfinx import cpp as _cpp
+from dolfinx.mesh import CellType
+
+__all__ = ["extract_topology_and_markers", "extract_geometry", "ufl_mesh", "cell_perm"]
+
+
+def extract_topology_and_markers(gmsh_model, model_name=None):
+    """Extract all entities tagged with a physical marker in the gmsh
```
garth-wells, line 19: Add `Arguments` and `Returns` field to the doc string. Can typing be introduced without importing gmsh?

python/dolfinx/io/gmsh.py
```diff
@@ -0,0 +1,120 @@
...
+    model, and collect the data per cell type. Returns a nested
+    dictionary where the first key is the gmsh MSH element type integer.
+    Each element type present in the model contains the cell topology of
+    the elements and corresponding markers.
+
+    """
+    if model_name is not None:
+        gmsh_model.setCurrent(model_name)
+
+    # Get the physical groups from gmsh on the form [(dim1, tag1),(dim1,
```
garth-wells, line 29: '. . . in the form . . . '?

python/dolfinx/io/gmsh.py
```diff
@@ -0,0 +1,120 @@
...
+    # tag2), (dim2, tag3),...]
+    phys_grps = gmsh_model.getPhysicalGroups()
+    topologies = {}
+    for dim, tag in phys_grps:
+        # Get the entities for a given dimension:
```
garth-wells, line 34: "Get entities of dimension 'dim'"?

python/dolfinx/io/gmsh.py
```diff
@@ -0,0 +1,120 @@
...
+        # dim=0->Points, dim=1->Lines, dim=2->Triangles/Quadrilaterals,
+        # etc.
+        entities = gmsh_model.getEntitiesForPhysicalGroup(dim, tag)
+
+        for entity in entities:
+            # Get data about the elements on a given entity:
+            # NOTE: Assumes that each entity only have one cell-type
```
garth-wells, line 41: What does this mean? Can an entity have more than one type?

python/dolfinx/io/gmsh.py
```diff
@@ -0,0 +1,120 @@
...
+    phys_grps = gmsh_model.getPhysicalGroups()
+    topologies = {}
+    for dim, tag in phys_grps:
+        # Get the entities for a given dimension:
+        # dim=0->Points, dim=1->Lines, dim=2->Triangles/Quadrilaterals,
+        # etc.
+        entities = gmsh_model.getEntitiesForPhysicalGroup(dim, tag)
+
+        for entity in entities:
+            # Get data about the elements on a given entity:
```
garth-wells, line 40: What data?

python/dolfinx/io/gmsh.py
```diff
@@ -0,0 +1,120 @@
...
+            # topology of the elements
+            properties = gmsh_model.mesh.getElementProperties(element_type)
+            name, dim, order, num_nodes, local_coords, _ = properties
+
+            # 2D array of shape (num_elements,num_nodes_per_element)
+            # containing the topology of the elements on this entity
+            # NOTE: GMSH indexing starts with 1 and not zero
+            element_topology = node_tags[0].reshape(-1, num_nodes) - 1
+
+            # Gather data for each element type and the
```
garth-wells, line 60: Gather implies communication in the DOLFINx context. Should it be 'Get data on each . . .'

python/dolfinx/io/gmsh.py
```diff
@@ -0,0 +1,120 @@
...
+    """
+    if model_name is not None:
+        gmsh_model.setCurrent(model_name)
+
+    # Get the unique tag and coordinates for nodes
+    # in mesh
+    indices, points, _ = gmsh_model.mesh.getNodes()
+    points = points.reshape(-1, 3)
+
+    # GMSH indices starts at 1
```
garth-wells, line 87: And? Explain how this relates to what the code is doing.

garth-wells (changes requested):

python/dolfinx/io/__init__.py
jorgensd: I am simply following the style we use for `dolfinx.fem.petsc`
I think it should be quite clear that this is not the gmsh module, as we in no part of any code import other software as submodules.

I think it would look really weird to use
```python
import dolfinx
dolfinx.io._gmsh.some_function(…)
```

garth-wells: The different with PETSc is subtle but is there - in `petsc4py` the name is `PETSc`. We use `petsc`.

I had in mind with the underscore keeping `io.gmsh` private, but if it's imported here only to go into `__all__` that doesn't make sense. I just think it's confusing using the same name as the actual gmsh module.

python/demo/demo_gmsh.py
```diff
@@ -166,16 +167,16 @@
     marked_facets, facet_values = np.empty((0, 6)).astype(np.int64), np.empty((0,)).astype(np.int32)
 
 # Permute the topology from GMSH to DOLFINx ordering
-domain = ufl_mesh_from_gmsh(gmsh_cell_id, 3)
+domain = ufl_mesh(gmsh_cell_id, 3)
 
-gmsh_tetra10 = cell_perm_gmsh(CellType.tetrahedron, 10)
+gmsh_tetra10 = cell_perm(CellType.tetrahedron, 10)
```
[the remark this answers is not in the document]
jorgensd: `topology_permutation`? As what the function does is giving you the permutation of each cell in the topology from gmsh (msh) ordering to DOLFINx ordering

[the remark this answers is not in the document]
garth-wells: It would help if the function name made clearer what is returned. Maybe `cell_perm_array`?

python/dolfinx/io/gmsh.py
jorgensd: A facet of a prism/pyramid is both quads and triangles, and the current interface would not support that

python/dolfinx/io/gmsh.py
```diff
@@ -0,0 +1,120 @@
...
+    indices -= 1
+
+    # Sort nodes in geometry according to the unique index
```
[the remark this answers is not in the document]
jorgensd: Every node in GMSH has a unique index. When retrieving the physical coordinates of the model, they are not necessarily in the same order as the actual index, i.e. `points[i]` refers to the `indices[i]`th node in the mesh.

garth-wells (approved):

PR #2135, opened 2022-04-19
Fix expression evaluation on subset of cells

### LINKED ISSUE #2134: Interpolation of ufl expression containing a dolfinx Function on subset of cells
Using a `Function` in a `dolfinx.fem.Expression` yields unexpected results when interpolating on a subset of the mesh's cells.

Consider the following:
```
from mpi4py import MPI
import dolfinx
import numpy as np

V = dolfinx.fem.FunctionSpace(mesh, ("CG", 1))

c = dolfinx.fem.Constant(mesh, 1.0)
u1 = dolfinx.fem.Function(V)

u1.interpolate(
	dolfinx.fem.Expression(c, V.element.interpolation_points))

print(f"u1: {u1.vector.array_r}")
for j in range(2):
	u2 = dolfinx.fem.Function(V)
	u2.interpolate(
		dolfinx.fem.Expression(u1, V.element.interpolation_points),
		cells=np.array([j], dtype=np.int32))
	print(f"j={j}, u2: {u2.vector.array_r}")
```

which gives me:

```
u1: [1. 1. 1. 1. 1. 1. 1. 1. 1.]
j=0, u2: [1. 1. 1. 0. 0. 0. 0. 0. 0.]
j=1, u2: [1.63041663e-322 0.00000000e+000 0.00000000e+000 1.05609417e-316
 0.00000000e+000 0.00000000e+000 0.00000000e+000 0.00000000e+000
 0.00000000e+000]
```
I only get the expected results if the optional `cells=` argument to `dolfinx.Function.interpolate` is a list of cell indices starting at zero and incrementing only by one, e.g. `[0, 1, 2]`. I get unexpected results with, e.g. `[0, 2]` and `[2]`.

jorgensd:
Fixes #2134

### DIFF
--- a/cpp/dolfinx/fem/Expression.h
+++ b/cpp/dolfinx/fem/Expression.h
@@ -187,7 +187,7 @@ public:
                                 std::next(coordinate_dofs.begin(), 3 * i));
       }
 
-      const T* coeff_cell = coeffs.data() + cell * cstride;
+      const T* coeff_cell = coeffs.data() + c * cstride;
       std::fill(values_local.begin(), values_local.end(), 0.0);
       _fn(values_local.data(), coeff_cell, constant_data.data(),
           coordinate_dofs.data(), nullptr, nullptr);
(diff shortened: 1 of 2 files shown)


nate-sime (approved):

garth-wells:
Could a test be added that would have picked up the reported bug?

garth-wells (approved):

PR #2137, opened 2022-04-20
Remove an AllGather from refinement code

chrisrichardson:
* removes an `MPI_Allgather` call and replaces with neighbor operations.

### DIFF
--- a/cpp/dolfinx/refinement/utils.cpp
+++ b/cpp/dolfinx/refinement/utils.cpp
@@ -389,6 +389,8 @@ refinement::adjust_indices(const common::IndexMap& index_map, std::int32_t n)
 
   // Communicate offset to neighbors
   std::vector<std::int64_t> neighbor_offsets(neighbors.size(), 0);
+  // Ensure allocation, in case where neighbors.size() == 0, needed for some MPI
+  neighbor_offsets.reserve(1);
   MPI_Neighbor_allgather(&global_offset, 1, MPI_INT64_T,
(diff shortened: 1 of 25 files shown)


cpp/dolfinx/refinement/utils.cpp
```diff
@@ -372,24 +374,42 @@ refinement::adjust_indices(const common::IndexMap& index_map, std::int32_t n)
...
+  // Ensure allocation, in case where neighbors.size() == 0, needed for some MPI
+  neighbor_offsets.reserve(1);
+  MPI_Neighbor_allgather(&global_offset, 1, MPI_INT64_T,
+                         neighbor_offsets.data(), 1, MPI_INT64_T, comm_fwd);
 
   const std::vector<int>& ghost_owners = index_map.ghost_owner_rank();
   int local_size = index_map.size_local();
-  for (int i = 0; i < local_size; ++i)
-    global_indices[i] += global_offsets[mpi_rank];
+  std::vector<std::int64_t> global_indices = index_map.global_indices();
```
garth-wells, line 401: This looks very circular going to-and-from global and neighbourhood ranks several times. I think it can be simpler and the map construction avoided.

garth-wells (changes requested):

cpp/dolfinx/refinement/utils.cpp
chrisrichardson: IndexMap should have a function that returns the `ghost_owners_neighbor_rank`

PR #2142, opened 2022-04-22
Raise on DirichletBC for None block

michalhabera:
This fixes a potentially dangerous implicit behaviour, where None in list of forms for assemble_matrix_block ignores all DirichletBC applied there. Typical case: pin pressure in the Stokes problem.

### DIFF
--- a/python/dolfinx/fem/petsc.py
+++ b/python/dolfinx/fem/petsc.py
@@ -369,6 +369,12 @@ def _(A: PETSc.Mat, a: typing.List[typing.List[FormMetaClass]],
             if a_block is not None:
                 Asub = A.getNestSubMatrix(i, j)
                 assemble_matrix(Asub, a_block, bcs, diagonal, const, coeff)
+            elif i == j:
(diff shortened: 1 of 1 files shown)


garth-wells (approved):

francesco-ballarin:
I am not quite sure I follow the reasoning behind lines 374 and 416.

I have problems with more than two blocks in which the diagonal form is `None`, but also `a_row[0]` is `None`, yet there exists `k > 0` such that `a_row[k]` is not `None`. Should `a_row[0]` be replaced with the first non `None` form?

michalhabera:
> I am not quite sure I follow the reasoning behind lines 374 and 416.
> 
> I have problems with more than two blocks in which the diagonal form is `None`, but also `a_row[0]` is `None`, yet there exists `k > 0` such that `a_row[k]` is not `None`. Should `a_row[0]` be replaced with the first non `None` form?

Yes, indeed, any non `None` could be used to check the function space. I got narrowed to 2x2 block systems.

PR #2152, opened 2022-04-26
Support assembly of forms where coefficients are defined on different meshes

jpdean:
This PR adds support for the assembly of forms where coefficients are defined on different meshes.This is useful for several applications, including multiphysics problems, and is another step towards https://github.com/FEniCS/dolfinx/projects/9. The user must provide entity maps relating the entities in the integration domain mesh to the entities in the meshes that the coefficient(s) are defined on. In most cases, this functionality would be used in conjunction with `create_submesh`, but any meshes can be used provided the user can supply the required entity maps.

Currently, the trial and test functions must be defined over the same mesh, and all meshes are assumed to have the same cell type. I will generalise this soon in an upcoming PR.


cpp/dolfinx/fem/Form.h
```diff
@@ -77,6 +77,12 @@ class Form
   /// @param[in] mesh The mesh of the domain. This is required when
   /// there are not argument functions from which the mesh can be
   /// extracted, e.g. for functionals
+  /// @param[in] entity_maps If any coefficients in the form are not
```
garth-wells, line 80: Would be nice to say first what `entity_maps` is, and then when it's needed.

I'm struggling bit with the explanation.

cpp/dolfinx/fem/utils.h
```diff
@@ -156,14 +156,18 @@ std::vector<std::string> get_constant_names(const ufcx_form& ufcx_form);
 /// @param[in] constants Spatial constants in the form
 /// @param[in] subdomains Subdomain markers
 /// @param[in] mesh The mesh of the domain
+/// @param[in] entity_maps The entity maps for the form
```
garth-wells, line 159: Expand comment?

cpp/dolfinx/fem/utils.h
```diff
@@ -663,21 +655,39 @@ void pack_coefficients(const Form<T>& form, IntegralType integral_type, int id,
...
+        // If the coefficient is defined over a different mesh to the
+        // integration domain mesh, we must map entities from the
+        // integration domain mesh to the coefficient mesh to pack them
+        // properly
+        auto coeff_mesh = coefficients[coeff]->function_space()->mesh();
+        if (coeff_mesh != form.mesh())
+        {
+          auto fetch_cell
+              = [&entity_map = form.entity_maps().at(coeff_mesh)](auto entity)
+          { return entity_map[entity]; };
```
garth-wells, line 679: Is ` entity_map` a `std::map`? If it is we should use `find` rather than `[]` for the look up.

garth-wells (commented):
A few first-pass comments.

PR #2160, opened 2022-04-28
Set blas directory for RedHat CI

IgorBaratta:
Openblas is not installed in the standard directory and findpackage(BLAS) doesn't provide the "include directories".

### DIFF
--- a/.github/workflows/redhat.yml
+++ b/.github/workflows/redhat.yml
@@ -1,9 +1,9 @@
 name: Red Hat clone
 
 on:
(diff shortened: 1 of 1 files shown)


mscroggs:
Intel is probably failing because of https://github.com/FEniCS/ffcx/pull/484

garth-wells:
How is it that this CI worked previously?

mscroggs:
Basix wasn't using cblas.h until today

PR #2161, opened 2022-04-28
Add version consistency checking test

mscroggs:
Added a test that checks that version numbers in different locations agree


mscroggs:
Yes, perhaps this would be better if reworked into a test that finds the version numbers and asserts that they're all the same/consistent

garth-wells:
Closing for now, new CI for running specific tags against each other should pick up issues.

PR #2162, opened 2022-04-28
Generalize computation of `mesh::h` to arbitrary order and entity

jorgensd:
Fix computation of `h` for arbitrary order geometries and sub entities (maximum distance between any vertex in the entity).

The current implementation throws a memory error on higher order geometries.

The new implementation works for all cell types, and any sub-dimension (topological dimension 0 returns 0).

The new implementation removes all copying and avoids xtensor

### DIFF
--- a/cpp/dolfinx/mesh/utils.cpp
+++ b/cpp/dolfinx/mesh/utils.cpp
@@ -331,7 +331,7 @@ std::vector<std::int32_t> mesh::locate_entities_boundary(
 
   // Get geometry data
   const graph::AdjacencyList<std::int32_t>& x_dofmap = mesh.geometry().dofmap();
-  tcb::span<const double> x_nodes = mesh.geometry().x();
+  xtl::span<const double> x_nodes = mesh.geometry().x();
 
   // Build vector of boundary vertices
   const std::vector<std::int32_t> vertices(boundary_vertices.begin(),
--- a/cpp/dolfinx/mesh/utils.cpp
+++ b/cpp/dolfinx/mesh/utils.cpp
@@ -14,6 +14,7 @@
 #include <dolfinx/common/IndexMap.h>
 #include <dolfinx/common/log.h>
 #include <dolfinx/common/math.h>
+#include <dolfinx/common/utils.h>
 #include <dolfinx/fem/ElementDofLayout.h>
 #include <dolfinx/graph/partition.h>
 #include <stdexcept>
@@ -72,11 +73,7 @@ std::vector<double> mesh::h(const Mesh& mesh,
   // Get geometry dofmap and dofs
   const Geometry& geometry = mesh.geometry();
   const graph::AdjacencyList<std::int32_t>& x_dofmap = geometry.dofmap();
-  xtl::span<const double> x_g = geometry().x();
-
-  auto geom_dofs
-      = xt::adapt(geometry.x().data(), geometry.x().size(), xt::no_ownership(),
-                  std::vector({geometry.x().size() / 3, std::size_t(3)}));
+  const xtl::span<const double> x_g = mesh.geometry().x();
 
   // Lambda function to compute norm
   auto l2_norm = [](auto p0, auto p1)
@@ -88,26 +85,20 @@ std::vector<double> mesh::h(const Mesh& mesh,
   };
 
   std::vector<double> h_cells(entities.size(), 0);
-  xt::xtensor<double, 2> coordinate_dofs({num_nodes, (std::size_t)3});
   for (std::size_t e = 0; e < entities.size(); ++e)
   {
     // Get cell coordinates/geometry
     auto x_dofs = x_dofmap.links(entities[e]);
-    assert(dofs.size() == num_nodes);
-    for (std::size_t i = 0; i < num_nodes; ++i)
-    {
-      common::impl::copy_N<3>(std::next(x_g.begin(), 3 * x_dofs[i]),
-                              std::next(coordinate_dofs.begin(), 3 * i));
-    }
+    assert(x_dofs.size() == num_nodes);
 
     // Get maximum edge length
     for (std::size_t i = 0; i < num_nodes; ++i)
     {
       for (std::size_t j = i + 1; j < num_nodes; ++j)
       {
-        auto p0 = xt::row(points, i);
-        auto p1 = xt::row(points, j);
-        h_cells[e] = std::max(h_cells[e], l2_norm(p0, p1));
+        h_cells[e] = std::max(h_cells[e],
+                              l2_norm(std::next(x_g.begin(), 3 * x_dofs[i]),
+                                      std::next(x_g.begin(), 3 * x_dofs[j])));
       }
     }
   }
--- a/cpp/dolfinx/mesh/utils.cpp
+++ b/cpp/dolfinx/mesh/utils.cpp
@@ -61,43 +61,53 @@ std::vector<double> mesh::h(const Mesh& mesh,
                             int dim)
 {
   if (dim != mesh.topology().dim())
-    throw std::runtime_error("Cell size when dim ne tdim  requires updating.");
+    throw std::runtime_error("Cell size when dim != tdim  requires updating.");
 
   if (mesh.topology().cell_type() == CellType::prism and dim == 2)
     throw std::runtime_error("More work needed for prism cell");
 
-  // Get number of cell vertices
-  const CellType type = cell_entity_type(mesh.topology().cell_type(), dim, 0);
-  const int num_vertices = num_cell_vertices(type);
+  // Get number of cell nodes
+  const std::size_t num_nodes = mesh.geometry().cmap().dim();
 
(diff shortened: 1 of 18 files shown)


garth-wells:
I think the PR title and description could be improved. The PR title will appear in the ChangeLog.

PR description should explain what 'generalise' means. What is computed for higher order element?

jorgensd:
> I think the PR title and description could be improved. The PR title will appear in the ChangeLog.
> 
> PR description should explain what 'generalise' means. What is computed for higher order element?

Generalize means that it takes the maximum distance between any node in the cell, i.e. For higher order cells, the distance between any of the nodes on the facets and a vertex can be larger than the distance between any set of vertices.

jorgensd:
> I think the PR title and description could be improved. The PR title will appear in the ChangeLog.
> 
> PR description should explain what 'generalise' means. What is computed for higher order element?

@garth-wells I've now updated the description of the Pull request, and extended it.
Now `dolfinx::mesh::h` can compute the size of any mesh entity (vertex, edge, facet, cell) for any cell type.

garth-wells:
I'm not sure about the definition of `h`. Is this a standard definition?

jorgensd:
> I'm not sure about the definition of `h`. Is this a standard definition?

`h` is usually the `CellDiameter`, which would be the largest distance between any two points in a cell.
If the cell is sufficiently curved, simply computing the distance between the vertices is not going to be sufficient.

We currently do not support this through the form compiler (FFCx), and neither does TSFC, see: https://github.com/firedrakeproject/tsfc/issues/224

jorgensd:
@garth-wells Code has been rewritten to only use the vertices of the entity.

cpp/dolfinx/mesh/utils.cpp
```diff
@@ -335,7 +331,7 @@ std::vector<std::int32_t> mesh::locate_entities_boundary(
 
   // Get geometry data
   const graph::AdjacencyList<std::int32_t>& x_dofmap = mesh.geometry().dofmap();
-  xtl::span<const double> x_nodes = mesh.geometry().x();
+  tcb::span<const double> x_nodes = mesh.geometry().x();
```
jorgensd, line 334:
```suggestion
  xtl::span<const double> x_nodes = mesh.geometry().x();
```

cpp/dolfinx/mesh/utils.cpp
```diff
@@ -60,48 +60,44 @@ std::vector<double> mesh::h(const Mesh& mesh,
...
+  const xtl::span<const double> x_g = geometry.x();
+
+  // Lambda function to compute norm
+  auto l2_norm = [](auto p0, auto p1)
   {
-    // Get the coordinates  of the vertices
-    auto dofs = x_dofs.links(entities[e]);
+    double norm = 0;
+    for (std::size_t k = 0; k < 3; ++k)
+      norm += std::pow(p0[k] - p1[k], 2);
```
garth-wells, line 78: `std::pow` should be used for an exponent of 2. It's not optimised for this case. Use `(p0[k] - p1[k])*(p0[k] - p1[k])`.

cpp/dolfinx/mesh/utils.cpp
```diff
@@ -60,48 +60,44 @@ std::vector<double> mesh::h(const Mesh& mesh,
...
+  if (dim > 0)
+  {
+    for (std::size_t e = 0; e < entities.size(); ++e)
     {
-      for (int j = i + 1; j < num_vertices; ++j)
+      // Get maximum distance between any vertex
+      auto e_vertices = xt::row(geometry_vertices, e);
+      for (std::size_t i = 0; i < num_vertices_per_entity; ++i)
       {
-        auto p0 = xt::row(points, i);
```
garth-wells, line 98: I don't quite see the point of this change while `e_vertices` is still an xtensor type. It seems a but muddled being a mixture of xtensor std STL.

jorgensd: There are no problems with taking row-views og xtensors, as they are contiguous in memory.
Eventually `entities_to_geometry` should return an adjacency-list, such that it can support prisms and pyramids, and that would remove the need for this view.

The main improvements in this PR is that the code now works for higher order geometries, sub entities of cells, and does not do any copying when computing `h`.

cpp/dolfinx/mesh/utils.h
```diff
@@ -64,7 +64,10 @@ graph::AdjacencyList<std::int64_t>
 extract_topology(const CellType& cell_type, const fem::ElementDofLayout& layout,
                  const graph::AdjacencyList<std::int64_t>& cells);
 
-/// Compute greatest distance between any two vertices
+/// Compute greatest distance between any two vertices in the entity
+/// @param[in] mesh The mesh
+/// @param[in] entities List of entities (local to process)
+/// @param[in] dim The topological dimension of the entity
```
garth-wells, line 70: Needs a `@return` in doc.

cpp/dolfinx/mesh/utils.h
```diff
@@ -64,7 +64,10 @@ graph::AdjacencyList<std::int64_t>
 extract_topology(const CellType& cell_type, const fem::ElementDofLayout& layout,
                  const graph::AdjacencyList<std::int64_t>& cells);
 
-/// Compute greatest distance between any two vertices
+/// Compute greatest distance between any two vertices in the entity
+/// @param[in] mesh The mesh
+/// @param[in] entities List of entities (local to process)
```
garth-wells, line 69: `/// @param[in] entities List of entity indices to compute the size measure for.`

cpp/dolfinx/mesh/utils.h
```diff
@@ -64,7 +64,10 @@ graph::AdjacencyList<std::int64_t>
 extract_topology(const CellType& cell_type, const fem::ElementDofLayout& layout,
                  const graph::AdjacencyList<std::int64_t>& cells);
 
-/// Compute greatest distance between any two vertices
+/// Compute greatest distance between any two vertices in the entity
+/// @param[in] mesh The mesh
```
garth-wells, line 68: `/// @param[in] mesh Mesh that entities belong to.`

python/test/unit/mesh/test_mesh.py
```diff
@@ -316,6 +316,26 @@ def test_cell_h(c0, c1, c5):
...
+    N = 3
+    mesh = create_unit_cube(MPI.COMM_WORLD, N, N, N, cell_type=CellType.prism)
+    tdim = mesh.topology.dim
+    num_cells = mesh.topology.index_map(tdim).size_local
+    cells = np.arange(num_cells, dtype=np.int32)
+    h = _cpp.mesh.h(mesh, tdim, cells)
+    assert(np.allclose(h, np.sqrt(3 / (N**2))))
+
+
+@pytest.mark.parametrize("ct", [CellType.hexahedron, CellType.tetrahedron])
```
garth-wells, line 329: `ct` is cryptic.

cpp/dolfinx/mesh/utils.cpp
```diff
@@ -60,48 +60,44 @@ std::vector<double> mesh::h(const Mesh& mesh,
...
-  if (dim != mesh.topology().dim())
-    throw std::runtime_error("Cell size when dim ne tdim  requires updating.");
 
-  if (mesh.topology().cell_type() == CellType::prism and dim == 2)
-    throw std::runtime_error("More work needed for prism cell");
-
-  // Get number of cell vertices
-  const CellType type = cell_entity_type(mesh.topology().cell_type(), dim, 0);
-  const int num_vertices = num_cell_vertices(type);
+  // Get entity vertices in geometry dofs
```
garth-wells, line 64: This comment doesn't make sense to me.

cpp/dolfinx/mesh/utils.cpp
garth-wells: I think it's confusing to be mixing concepts.

cpp/dolfinx/mesh/utils.cpp
```diff
@@ -60,48 +60,44 @@ std::vector<double> mesh::h(const Mesh& mesh,
...
-                  std::vector({geometry.x().size() / 3, std::size_t(3)}));
-  std::vector<double> h_cells(entities.size(), 0);
-  assert(num_vertices <= 8);
-  xt::xtensor_fixed<double, xt::xshape<8, 3>> points;
-  for (std::size_t e = 0; e < entities.size(); ++e)
+  const mesh::Geometry& geometry = mesh.geometry();
+  const xtl::span<const double> x_g = geometry.x();
+
+  // Lambda function to compute norm
+  auto l2_norm = [](auto p0, auto p1)
```
garth-wells, line 74: The `auto` make things a bit unclear. The return type is always `double`, so it would be clearer if it was explicit.

I'm not sure what `p0` and `p1` are. There are index into, but from where the function is called it looks like an xtensor iterator.

garth-wells (changes requested):

garth-wells:
Merged into #2193.

PR #2172, opened 2022-05-10
Bump Docker base to Ubuntu 22.04 and other package updates

garth-wells:
Update Gmsh, PETSc, SLEPc, HDF5 , xtensor snd MPICH versions.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -42,26 +42,26 @@
 #
 
 
-ARG GMSH_VERSION=4_9_3
+ARG GMSH_VERSION=4_10_1
 ARG HDF5_SERIES=1.12
-ARG HDF5_PATCH=1
+ARG HDF5_PATCH=2
 ARG PYBIND11_VERSION=2.9.2
-ARG PETSC_VERSION=3.17.0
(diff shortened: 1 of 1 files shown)


francesco-ballarin:
FYI, sadly this breaks pyvista, because VTK 9.2 is late and VTK 9.1 on pypi has no python 3.10 support.

Relevant links, the first of which provides a workaround for affected users:
https://github.com/pyvista/pyvista/discussions/2064
https://discourse.vtk.org/t/vtk-9-2-0-release-cycle/8149/2
https://gitlab.kitware.com/vtk/vtk/-/issues/18335

@jhale may want to double check if this affects the jupyter images
@jorgensd will find that the visualization demo workflow is broken due to this change

francesco-ballarin:
It gets even worse for people (like me ;) ) who are using `pyvista.PlotterITK` in their notebooks, see https://github.com/InsightSoftwareConsortium/itkwidgets/issues/440 and https://github.com/InsightSoftwareConsortium/ITKMeshToPolyData/issues/45 . Fortunately, there isn't any in dolfinx.

PR #2180, opened 2022-05-17
Generalize vtk_triangle permutation logic

tiagovla:


tiagovla:
I think gmsh supports [higher order triangles](https://gitlab.onelab.info/gmsh/gmsh/-/blob/master/src/geo/MTriangle.h#L314) too.

PR #2182, opened 2022-05-18
Revised IndexMap and new Scatterer class

garth-wells:
This PR simplifies `IndexMap`. `IndexMap` had gotten complicated by a supporting the parallel layout of data and parallel communication. This included computing communication related data at construction that was not always used. This changes reduces `IndexMap` back top describing the parallel layout, and a new `Scatterer` class supports communication 'on' index maps.


python/test/unit/common/test_index_map.py
```diff
@@ -31,7 +31,7 @@ def test_sub_index_map():
     src_ranks = dest_ranks
 
     # Create index map
-    map = dolfinx.common.IndexMap(comm, map_local_size, dest_ranks, map_ghosts, src_ranks)
+    map = dolfinx.common.IndexMap(comm, map_local_size, map_ghosts, src_ranks)
```
IgorBaratta, line 34: Should we add an interface for when the dest ranks are known?

IgorBaratta (approved):

PR #2183, opened 2022-05-18
Update for name of KaHIP library

garth-wells:
KaHIP changed its library name from `interface` to `kahip`.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -70,18 +70,11 @@ add_feature_info(XTENSOR_OPTIMIZE XTENSOR_OPTIMIZE "Enable architecture-specific
 # Enable or disable optional packages
 
 # List optional packages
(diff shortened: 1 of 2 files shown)


tiagovla:
Shouldn't this be on by default?https://github.com/FEniCS/dolfinx/blob/8e5bc85c15f19b4bb01bbc30411633e2d0995101/cpp/CMakeLists.txt#L77

IgorBaratta:
> Shouldn't this be on by default?
> 
> https://github.com/FEniCS/dolfinx/blob/8e5bc85c15f19b4bb01bbc30411633e2d0995101/cpp/CMakeLists.txt#L77

KaHIP's build process is quite fragile.
So, if a user wants to use kahip, she/he should set `DOLFINX_ENABLE_KAHIP=ON` explicitly.

PR #2193, opened 2022-05-22
Remove some xtensor from mesh utility code

garth-wells:
Generally simpler and avoid hidden costs in the xtensor code.

### DIFF
--- a/cpp/dolfinx/fem/FunctionSpace.h
+++ b/cpp/dolfinx/fem/FunctionSpace.h
@@ -82,9 +82,9 @@ public:
   /// `(num_points, gdim)`, otherwise it is transposed and has shape
   /// `(gdim, num_points)`
   /// @return The dof coordinates `[([x0, y0, z0], [x1, y1, z1], ...)`
-  /// if @p transpose is false, and otherwise the returned data is
-  /// transposed.
-  xt::xtensor<double, 2> tabulate_dof_coordinates(bool transpose) const;
+  /// if `transpose` is false, and otherwise the returned data is
+  /// transposed. Storage is row-major.
+  std::vector<double> tabulate_dof_coordinates(bool transpose) const;
 
   /// The mesh
   std::shared_ptr<const mesh::Mesh> mesh() const;
--- a/cpp/dolfinx/fem/FunctionSpace.h
+++ b/cpp/dolfinx/fem/FunctionSpace.h
@@ -79,8 +79,8 @@ public:
   ///
   /// @todo Remove - see function in interpolate.h
   /// @param[in] transpose If false the returned data has shape
-  /// `(num_points, gdim)`, otherwise it is transposed and has shape
-  /// `(gdim, num_points)`
+  /// `(num_points, 3)`, otherwise it is transposed and has shape `(3,
+  /// num_points)`.
   /// @return The dof coordinates `[([x0, y0, z0], [x1, y1, z1], ...)`
   /// if `transpose` is false, and otherwise the returned data is
   /// transposed. Storage is row-major.
--- a/cpp/dolfinx/geometry/BoundingBoxTree.cpp
+++ b/cpp/dolfinx/geometry/BoundingBoxTree.cpp
@@ -39,10 +39,7 @@ std::array<std::array<double, 3>, 2>
 compute_bbox_of_entity(const mesh::Mesh& mesh, int dim, std::int32_t index)
 {
   // Get the geometrical indices for the mesh entity
-  const int tdim = mesh.topology().dim();
   xtl::span<const double> xg = mesh.geometry().x();
-
-  mesh.topology_mutable().create_connectivity(dim, tdim);
 
   // FIXME: return of small dynamic array is expensive
   xtl::span<const std::int32_t> entity(&index, 1);
@@ -241,6 +238,7 @@ BoundingBoxTree::BoundingBoxTree(const mesh::Mesh& mesh, int tdim,
 
   // Initialize entities of given dimension if they don't exist
   mesh.topology_mutable().create_entities(tdim);
+  mesh.topology_mutable().create_connectivity(tdim, mesh.topology().dim());
 
   // Create bounding boxes for all mesh entities (leaves)
   std::vector<std::pair<std::array<std::array<double, 3>, 2>, std::int32_t>>
--- a/cpp/dolfinx/geometry/BoundingBoxTree.cpp
+++ b/cpp/dolfinx/geometry/BoundingBoxTree.cpp
@@ -46,17 +46,19 @@ compute_bbox_of_entity(const mesh::Mesh& mesh, int dim, std::int32_t index)
 
   // FIXME: return of small dynamic array is expensive
   const std::array<std::int32_t, 1> entity = {index};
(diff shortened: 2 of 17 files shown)


cpp/dolfinx/geometry/BoundingBoxTree.cpp
```diff
@@ -39,24 +39,20 @@ std::array<std::array<double, 3>, 2>
...
   xtl::span<const double> xg = mesh.geometry().x();
 
-  mesh.topology_mutable().create_connectivity(dim, tdim);
-
   // FIXME: return of small dynamic array is expensive
-  const std::array<std::int32_t, 1> entity = {index};
-  const xt::xtensor<std::int32_t, 2> vertex_indices
+  xtl::span<const std::int32_t> entity(&index, 1);
+  const std::vector<std::int32_t> vertex_indices
       = mesh::entities_to_geometry(mesh, dim, entity, false);
```
jorgensd, line 47: I believe this isn’t a good idea. Entities to geometry should be an AdjacencyList, as it would have variable width for sub entities of pyramids/prisms

cpp/dolfinx/mesh/utils.cpp
```diff
@@ -435,64 +457,84 @@ mesh::entities_to_geometry(const Mesh& mesh, int dim,
   assert(e_to_v);
   const auto c_to_v = topology.connectivity(tdim, 0);
   assert(c_to_v);
-  for (std::size_t i = 0; i < entity_list.size(); ++i)
+
+  const std::size_t num_vertices
+      = num_cell_vertices(cell_entity_type(cell_type, dim, 0));
```
jorgensd, line 462: Here we need a FIXME, as this is not correct for prisms and pyramids.

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -396,21 +400,25 @@ void mesh(py::module& m)
...
-           bool orient)
+           py::array_t<std::int32_t, py::array::c_style> entities, bool orient)
         {
-          return xt_as_pyarray(dolfinx::mesh::entities_to_geometry(
-              mesh, dim, xtl::span(entity_list.data(), entity_list.size()),
-              orient));
+          std::vector<std::int32_t> idx = dolfinx::mesh::entities_to_geometry(
+              mesh, dim, entities, orient);
+          dolfinx::mesh::CellType cell_type = mesh.topology().cell_type();
+          std::size_t num_vertices = dolfinx::mesh::num_cell_vertices(
```
jorgensd, line 408: Same comment as above

python/setup.py
```diff
@@ -47,7 +47,8 @@ def build_extension(self, ext):
                       f'-DPython3_LIBRARIES={sysconfig.get_config_var("LIBDEST")}',
                       f'-DPython3_INCLUDE_DIRS={sysconfig.get_config_var("INCLUDEPY")}']
 
-        cfg = 'Debug' if self.debug else 'Release'
+        cfg = 'Debug'
+        # cfg = 'Debug' if self.debug else 'Release'
```
jorgensd, line 51: I guess this should be reverted

cpp/dolfinx/geometry/BoundingBoxTree.cpp
garth-wells: > I believe this isn’t a good idea. Entities to geometry should be an AdjacencyList, as it would have variable width for sub entities of pyramids/prisms

There is no functionality change in this PR. Just an improvement in implementation.

jorgensd: I think by flattening the data you make it more tedious to go back to an AdjacencyList if we want to support prisms and pyramid facets. However, do you want me to make an issue that can deal with it after this PR?

cpp/dolfinx/fem/FunctionSpace.h
```diff
@@ -82,9 +82,9 @@ class FunctionSpace
   /// `(num_points, gdim)`, otherwise it is transposed and has shape
```
jorgensd, line 82: I guess the documentation regarding shape is not really correct any longer.
Similarly the return type is no longer correct.

cpp/dolfinx/mesh/utils.cpp
```diff
@@ -60,52 +60,54 @@ std::vector<double> mesh::h(const Mesh& mesh,
                             const xtl::span<const std::int32_t>& entities,
                             int dim)
 {
-  if (dim != mesh.topology().dim())
-    throw std::runtime_error("Cell size when dim ne tdim  requires updating.");
+  if (entities.empty())
+    return std::vector<double>();
+  if (dim > 0)
```
jorgensd, line 65:
```suggestion
  if (dim == 0)
```

cpp/dolfinx/mesh/utils.cpp
```diff
@@ -60,52 +60,54 @@ std::vector<double> mesh::h(const Mesh& mesh,
...
+    return std::sqrt(norm);
+  };
 
-    // The below should work, but misbehaves with the Intel icpx compiler
-    // xt::view(points, xt::range(0, num_vertices), xt::all())
-    //     = xt::view(geom_dofs, xt::keep(dofs), xt::all());
-    auto points_view = xt::view(points, xt::range(0, num_vertices), xt::all());
-    points_view.assign(xt::view(geom_dofs, xt::keep(dofs), xt::all()));
+  // Compute greatest distance between any to vertices
+  assert(dim > 0);
```
jorgensd, line 87:
```suggestion
```
Can be removed if combined with the change above

cpp/dolfinx/mesh/utils.h
```diff
@@ -118,20 +127,22 @@ std::vector<std::int32_t> locate_entities_boundary(
...
+/// geometry indices for
 /// @param[in] orient If true, in 3D, reorients facets to have
 /// consistent normal direction
-/// @return Indices in the geometry array for the mesh entity vertices, i.e.
-/// indices(i, j) is the position in the geometry array of the j-th vertex of
-/// the entity entity_list[i].
-xt::xtensor<std::int32_t, 2>
+/// @return Indices in the geometry array for the entity vertices. The
+/// shape is `(num_entities, num_vertices_per_entity)` and the storage
+/// is row-major. The index `indices[i, j]` is the position in the
```
jorgensd, line 143:
```suggestion
/// is row-major. The index `indices[i * num_vertices_per_entity + j]` is the position in the
```

cpp/dolfinx/geometry/BoundingBoxTree.cpp
garth-wells: We are still miles from supported pyramid and mixed topologies. No need for hypothetical changes yet.

cpp/dolfinx/mesh/utils.h
garth-wells: Less clear to me.

jorgensd: But indices[i,j] assumes a 2D structure, while what we have is flat. I agree it is more verbose, but it correct

garth-wells: That's why I give the shape and the storage layout (row-major) in the docstring.

PR #2200, opened 2022-05-25
Rename folder for dolfinx python installation in docker images.

jorgensd:
As the docker images uses python 3.10, the corresponding folders in the docker images should be renamed.


garth-wells:
Maybe a variable with a comment at the top of the Dockerfile would be helpful?

jhale:
For the Dockerfile it would be best to have a manual variable to tweak at the top as @garth-wells says.

For the JSON file, I think we're out of luck, it seems to be static https://github.com/jupyter/notebook/issues/3469

jhale:
Fixes issue #2201

PR #2202, opened 2022-05-25
Revise plotting with matplotlib in parallel

jorgensd:
- Lagrange variants demo should only be executed in real mode
- Skip plotting in parallel for Lagrange variants demo (as they try different to write to the same file).
- TNT demo: Only plot on rank zero as all processes have the same data
- Testing the ubuntu-22.04 runner (https://github.com/actions/virtual-environments/issues/5490)

### DIFF
--- a/.github/workflows/pyvista.yml
+++ b/.github/workflows/pyvista.yml
@@ -2,18 +2,18 @@ name: Test visualisation demos
 
 on:
   # Uncomment the below to trigger tests on push
-  # push:
-  #   branches:
-  #     - "**"
+  push:
+    branches:
+      - "**"
   schedule:
     # '*' is a special character in YAML, so string must be quoted
     - cron: "0 1 * * *"
   workflow_dispatch: ~
 
 jobs:
-  build:
+  pyvista:
     if: "!(contains(github.event.head_commit.message, '[ci skip]') || contains(github.event.head_commit.message, '[skip ci]'))"
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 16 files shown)


python/demo/demo_lagrange_variants.py
```diff
@@ -19,23 +19,29 @@
...
 from petsc4py.PETSc import ScalarType
 
+if np.issubdtype(ScalarType, np.complexfloating):
+    print("Demo should only be executed with DOLFINx real mode")
+    exit(0)
+
 import matplotlib.pylab as plt
+
+import basix
+import basix.ufl_wrapper
```
mscroggs, line 37: These two imports should maybe be after the next comment block that explains them?

.github/workflows/pyvista.yml
```diff
@@ -2,18 +2,18 @@ name: Test visualisation demos
 
 on:
   # Uncomment the below to trigger tests on push
-  # push:
-  #   branches:
-  #     - "**"
+  push:
+    branches:
+      - "**"
```
mscroggs, line 7: Should this be on for every push or just for PRs to main?

mscroggs (approved):

.github/workflows/pyvista.yml
jorgensd: It is just to be able to check that the CI passes, once it is green I'll revert it

jorgensd, line 7:
```suggestion
  # push:
  #   branches:
  #     - "**"
```

garth-wells (approved):

PR #2208, opened 2022-05-25
Remove pugixml source from repository

### LINKED ISSUE #2005: Remove inclusion of pugixml source code
`pugixml` is now widely available through package managers (Debian/Ubuntu, Spack, homebrew) and it's not generally good practice to add third-party library code.

Should we remove the `pugixml` code from the git repository and use an installed version?

garth-wells:
Remove `pugixml` source from the repository. `pugixml` is now widely available in package managers, e.g. homebrew, Debian/Ubuntu, Spack.

`pugixml` is installed in the CI yaml files at the moment. This will move to the Docker images if this PR gets merged.

Fixes #2005.


jorgensd (approved):

chrisrichardson (approved):
I'm fairly ambivalent about this. Adding more installation dependencies vs. a single file which needs updating from time-to-time.

PR #2209, opened 2022-05-26
Expose function `MeshTags.find` to Python

jorgensd:


python/dolfinx/wrappers/mesh.cpp
```diff
@@ -123,7 +123,13 @@ void declare_meshtags(py::module& m, std::string type)
                                return py::array_t<std::int32_t>(
                                    self.indices().size(), self.indices().data(),
                                    py::cast(self));
-                             });
+                             })
+      .def("find",
+           [](dolfinx::mesh::MeshTags<T>& self, T value)
+           {
+             std::vector<std::int32_t> indices = self.find(value);
+             return py::array_t<std::int32_t>(indices.size(), indices.data());
```
garth-wells, line 131: This can be `return as_pyarray(self.find(value));`, which is more efficient.

garth-wells (approved):

PR #2210, opened 2022-05-26
Fix for `VTXWriter` for meshes with no cells on some ranks

jpdean:


python/test/unit/io/test_adios2.py
```diff
@@ -238,3 +239,33 @@ def test_save_vtkx_cell_point(tempdir):
         f = VTXWriter(mesh.comm, filename, [u])
         f.write(0)
         f.close()
+
+
+def test_empty_rank_mesh(tempdir):
+    """Test VTXWrite on mesh where some ranks have no cells"""
```
jorgensd, line 245:
```suggestion
    """Test VTXWriter on mesh where some ranks have no cells"""
```

garth-wells (approved):

Issue #2201, opened 2022-05-25
broken complex mode in docker image


francesco-ballarin:
```
$ docker run -it --rm dolfinx/dolfinx
root@0693c5ef81e6:~# . /usr/local/bin/dolfinx-complex-mode
root@0693c5ef81e6:~# python3 -c 'import dolfinx.cpp'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/usr/local/dolfinx-complex/lib/python3.8/dist-packages/dolfinx/__init__.py", line 33, in <module>
    from dolfinx import common
  File "/usr/local/dolfinx-complex/lib/python3.8/dist-packages/dolfinx/common.py", line 10, in <module>
    from dolfinx import cpp as _cpp
ImportError: /usr/local/dolfinx-complex/lib/python3.8/dist-packages/dolfinx/cpp.cpython-310-x86_64-linux-gnu.so: undefined symbol: _ZN7dolfinx3fem5petsc18create_matrix_nestERKSt6vectorIS2_IPKNS0_4FormIdEESaIS6_EESaIS8_EERKS2_IS2_INSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESaISI_EESaISK_EE
root@0693c5ef81e6:~# 
```

jhale:
Fixed in PR #2200

francesco-ballarin:
The actual fix will be in #2211

francesco-ballarin:
Neither PR fix the issue

```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/__w/multiphenicsx/multiphenicsx/multiphenicsx/cpp/__init__.py", line 12, in <module>
    from multiphenicsx.cpp.compile_code import compile_code
  File "/__w/multiphenicsx/multiphenicsx/multiphenicsx/cpp/compile_code.py", line 15, in <module>
    import dolfinx.jit
  File "/usr/local/dolfinx-complex/lib/python3.10/dist-packages/dolfinx/__init__.py", line 29, in <module>
    from dolfinx import common
  File "/usr/local/dolfinx-complex/lib/python3.10/dist-packages/dolfinx/common.py", line 10, in <module>
    from dolfinx import cpp as _cpp
[...]

PR #2217, opened 2022-05-28
Fix complex mode in dolfinx/dolfinx images

jorgensd:
As far as I can tell, the issue is that the `pip3 install` command creates a build directory that caches some info when we build the dolfinx real and then complex mode.

I've added a test to the docker.yml CI that picks up this issue, and suggested a new fix (removing the build directories from `dolfinx/python` after installation.

Should hopefully resolve https://github.com/FEniCS/dolfinx/issues/2201


francesco-ballarin:
can we manually run the docker pipeline to test this?

jhale:
The underlying reason for this might have been a change in pip related to switching from out-of-source to in-source builds, although I'm not certain.

Regarding the patch, I think it would be best to source own own configuration script `dolfinx-complex-mode` rather than manually setup the environment.

jorgensd:
> The underlying reason for this might have been a change in pip related to switching from out-of-source to in-source builds, although I'm not certain.
> 
> Regarding the patch, I think it would be best to source our own configuration script `dolfinx-complex-mode` rather than manually setup the environment.

I wasnt able to make a one-line command that did that with `docker run`. If you could show me how to do that I’d be grateful.

jhale:
The real underlying issue is that the DOLFINx Python object is linked with real mode PETSc objects, e.g.

```
	libamd.so.2 => /usr/local/petsc/linux-gnu-real-32/lib/libamd.so.2 (0x0000ffff8e3d0000)
	libscalapack.so.2.2 => /usr/local/petsc/linux-gnu-real-32/lib/libscalapack.so.2.2 (0x0000ffff8df70000)
	libsuperlu.so.5 => /usr/local/petsc/linux-gnu-real-32/lib/libsuperlu.so.5 (0x0000ffff8def0000)
```

jhale:
I've triggered a build with this branch at [b7e508b](https://github.com/FEniCS/dolfinx/pull/2217/commits/b7e508becd518a7257ad81623919be5d61e692ca).

jhale:
This patch hasn't fixed the issue. DOLFINx Python still seems to link to PETSc real mode libraries.

jhale:
https://github.com/FEniCS/dolfinx/runs/6652047258?check_suite_focus=true

docker/Dockerfile
```diff
@@ -418,16 +418,16 @@ ONBUILD RUN cd dolfinx && \
...
     cd ../ && \
     mkdir -p build-complex && \
     cd build-complex && \
     PETSC_ARCH=linux-gnu-complex-32 cmake -G Ninja -DCMAKE_INSTALL_PREFIX=/usr/local/dolfinx-complex -DCMAKE_BUILD_TYPE=${DOLFINX_CMAKE_BUILD_TYPE} -DCMAKE_CXX_FLAGS=${DOLFIN_CMAKE_CXX_FLAGS} ../cpp && \
     ninja install && \
     . /usr/local/dolfinx-complex/lib/dolfinx/dolfinx.conf && \
     cd ../python && \
-    CXXFLAGS=${DOLFINX_CMAKE_CXX_FLAGS} PETSC_ARCH=linux-gnu-complex-32 pip3 install -v --target /usr/local/dolfinx-complex/lib/python${PYTHON_VERSION}/dist-packages --no-dependencies . && \
+    CXXFLAGS=${DOLFINX_CMAKE_CXX_FLAGS} PETSC_ARCH=linux-gnu-complex-32 pip3 install -v --target /usr/local/dolfinx-complex/lib/python${PYTHON_VERSION}/dist-packages --no-dependencies --no-cache-dir . && \
     rm -rf build
```
jorgensd, line 431: Do want to remove this one as well?

PR #2222, opened 2022-05-31
Add missing function parameter names of pybind11 generated functions

### LINKED ISSUE #2220: Function argument names are wrong in the pybind11 generated functions
Hello, quick feedback from a user perspective.

Take any function from the python module `dolfinx.cpp`, for instance `dolfinx.cpp.fem.CoordinateElement.pull_back` that can be called as a method (without explicitly importing the cpp module) `mesh.geometry.cmap.pull_back(x, cell_geom)`.

The argument names for these functions are not explicit enough, for instance the generated prototype for `pull_back` is
`(self: dolfinx.cpp.fem.CoordinateElement, arg0: numpy.ndarray[numpy.float64], arg1: numpy.ndarray[numpy.float64]) -> numpy.ndarray[numpy.float64]`.
From the user perspective, that makes it hard to understand how to call this function, and thus one has to read the C++ documentation or source code.

A fix for this would be to add the original argument names in the pybind11 code, as explained [here](https://stackoverflow.com/questions/71443908/how-to-have-pybind11-positional-argument-names-show-up-on-help) for example.

atouminet:
Fixes #2220

### DIFF
--- a/python/dolfinx/wrappers/common.cpp
+++ b/python/dolfinx/wrappers/common.cpp
@@ -54,31 +54,36 @@ void common(py::module& m)
   py::class_<dolfinx::common::IndexMap,
              std::shared_ptr<dolfinx::common::IndexMap>>(m, "IndexMap")
       .def(py::init(
-          [](const MPICommWrapper comm, std::int32_t local_size)
-          { return dolfinx::common::IndexMap(comm.get(), local_size); }))
+               [](const MPICommWrapper comm, std::int32_t local_size)
+               { return dolfinx::common::IndexMap(comm.get(), local_size); }),
+           py::arg("comm"), py::arg("local_size"))
       .def(py::init(
-          [](const MPICommWrapper comm, std::int32_t local_size,
-             const py::array_t<std::int64_t, py::array::c_style>& ghosts,
-             const py::array_t<int, py::array::c_style>& ghost_owners)
-          {
-            return dolfinx::common::IndexMap(comm.get(), local_size, ghosts,
-                                             ghost_owners);
-          }))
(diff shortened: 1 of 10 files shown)


IgorBaratta:
Hi @Armand-Touminet,
could you reformat the modified files using the `.clang-format` provided in the top-level directory?

It seems that your "code formatter" is using a different style.

Armand-Touminet:
I guess it is just not possible. From the pybind11 perspective, the function to wrap is just an object wrapping a function pointer. In C++ these pointers do not store the names of the function arguments.

One thing that helps with maintainability though is that pybind11 will run some static asserts to check if the number of arguments matches the number of `py::arg` defined in the code, such that if some arguments are added or removed from the C++ functions, not updating the wrappers will result in a build error. However swapping or renaming arguments will require manual intervention.

The same could be said for the C++ API documentation that is not translated automatically to the python API.
I guess that's just how it is when working with multiple languages.

nate-sime:
I like the idea. Also for some reason your IDE or code linting has over formatted your code.

Armand-Touminet:
Are you sure about the formatting? I made sure to run clang-format on each file. Just checked again and not seeing any diffs compared to the current version.

jhale:
https://github.com/FEniCS/dolfinx/runs/6676578989?check_suite_focus=true

The formatting is OK according to the CI.

nate-sime (approved):
This is definitely helpful for users, I just worry about maintenance.

garth-wells:
@atouminet could you resolve the merge conflict?

garth-wells (approved):

PR #2223, opened 2022-05-31
Change pugixml CMake target to be backwards compatible

jorgensd:
Ref discussion on slack. @massimiliano-leoni could you verify that this works for you?


nate-sime:
This also is needed for the spack build.

nate-sime (approved):

PR #2224, opened 2022-06-01
Fix incorrect `assert` in `Scatterer`

jpdean:
After discussing with @jorgensd, it looks like [this](https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/common/Scatterer.h#L288) `assert` statement might be incorrect. `local_data` has size `size_local() *  bs()`, but the indices in `_local_inds` are grouped by the processes in the neighbourhood, so for some index maps, `_local_inds.size()` may be larger than `local_data.size()`.

This PR removes the `assert` statement, but perhaps it could be replaced.


cpp/dolfinx/common/Scatterer.h
```diff
@@ -285,7 +285,6 @@ class Scatterer
                        MPI_Request& request)
   {
     assert(local_buffer.size() == _local_inds.size());
-    assert(_local_inds.size() <= local_data.size());
```
garth-wells, line 288: Can this check be fixed rather the remove (see @IgorBaratta's comment)?

garth-wells (changes requested):

garth-wells (approved):

PR #2226, opened 2022-06-01
Minor improvements to `transfer_cell/facet_meshtags`

jorgensd:
- Add asserts to make it possible to find missing connectivities in DEBUG mode.
- Move mesh to avoid it being copied in `MeshTags` constructor.

The last point would cause issues when refining the mesh, then refining the tag and trying to write it to file.
You cannot create the required connectivities on the refined mesh, but currently have to access the copy through the meshtag.

### DIFF
--- a/cpp/dolfinx/refinement/utils.cpp
+++ b/cpp/dolfinx/refinement/utils.cpp
@@ -458,6 +458,7 @@ mesh::MeshTags<std::int32_t> refinement::transfer_facet_meshtag(
 
   // Get global index for each refined cell, before reordering in Mesh
   // construction
+  assert(refined_mesh);
   const std::vector<std::int64_t>& original_cell_index
       = refined_mesh->topology().original_cell_index;
   assert(original_cell_index.size() == cell.size());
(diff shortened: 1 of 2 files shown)


cpp/dolfinx/refinement/utils.cpp
```diff
@@ -448,7 +448,9 @@ mesh::MeshTags<std::int32_t> refinement::transfer_facet_meshtag(
 
   auto parent_c_to_f
       = parent_meshtag.mesh()->topology().connectivity(tdim, tdim - 1);
+  assert(parent_c_to_f);
```
garth-wells, line 451: These asserts should probably be errors.

cpp/dolfinx/refinement/utils.cpp
```diff
@@ -652,6 +654,6 @@ mesh::MeshTags<std::int32_t> refinement::transfer_cell_meshtag(
   }
 
   return mesh::MeshTags<std::int32_t>(
-      std::make_shared<mesh::Mesh>(refined_mesh), tdim,
+      std::make_shared<mesh::Mesh>(std::move(refined_mesh)), tdim,
```
garth-wells, line 657: `move` for `refined_mesh` doesn't make sense here. It won't move, and if it could it would cause a corruption. The old code is wrong too. It makes a copy of the mesh.

The argument `refined_mesh` should probably be a `shared_ptr`, or a more granular design would (i) have a function to get the new `MeshTag` data, then (ii) create the new `MeshTag`.

cpp/dolfinx/refinement/utils.cpp
```diff
@@ -555,7 +557,7 @@ mesh::MeshTags<std::int32_t> refinement::transfer_facet_meshtag(
   }
 
   return mesh::MeshTags<std::int32_t>(
-      std::make_shared<mesh::Mesh>(refined_mesh), tdim - 1,
+      std::make_shared<mesh::Mesh>(std::move(refined_mesh)), tdim - 1,
```
garth-wells, line 560: See comments on below funtion.

garth-wells (changes requested):

chrisrichardson (approved):

PR #2232, opened 2022-06-06
Fix for `tabulate_dof_coordinates` when some dofs don't belong to a cell on some processes

jpdean:
It is possible to create a function space where some dofs don't belong to a cell on some processes. In this case, the implementation of `tabulate_dof_coordinates` in `main` does not correctly tabulate the coordinates of those dofs.

This PR fixes this issue, and adds a test that fails on `main`.


cpp/dolfinx/common/Scatterer.h
```diff
@@ -285,7 +285,9 @@ class Scatterer
                        MPI_Request& request)
   {
     assert(local_buffer.size() == _local_inds.size());
-    assert(_local_inds.size() <= local_data.size());
+    if (_local_inds.size() > 0)
```
garth-wells, line 288: This isn't correct - the `assert` gets removed in non-debug mode, with the `if` then probably applied to the next statement.

cpp/dolfinx/fem/FunctionSpace.cpp
```diff
@@ -221,17 +220,61 @@ FunctionSpace::tabulate_dof_coordinates(bool transpose) const
...
+    for (std::size_t i = 0; i < dofs.size(); ++i)
+      for (std::size_t j = 0; j < gdim; ++j)
+        coords[dofs[i] * 3 + j] = x(i, j);
+  }
+
+  // If this process owns or ghosts dofs that do not belong to a cell
+  // on this process, the coordinates of those dofs will not have been
+  // tabulated by the above. Hence, these must be communicated.
+  const int size_local = index_map->size_local();
+  const int bs = 3;
```
garth-wells, line 232: Use `constexpr`.

garth-wells (commented):
I'm not sure about this. It introduces a new MPI communication, which is a bit hidden/implicit my taste.

Also, I can't follow with ambiguity the issue from the title. Could the title be improved and some text added to the PR description?

jorgensd:
The MPI communication is hidden/implicit as it is needed when:
- A process owns a DOF, but does not have any cells connected to the DOF.
This is the case when we allow `SubMeshes` as we keep the ownership from the parent_mesh.
This communication is an implementation detail, and is not something a user should be expected to be aware of.

jpdean:
Yes, I appreciate that it isn't ideal. Unfortunately, I think it's the only practical way to deal with this type of corner case. A similar fix is needed for interpolate, but that should be it.

I'll improve the title.

garth-wells:
> Yes, I appreciate that it isn't ideal. Unfortunately, I think it's the only practical way to deal with this type of corner case. A similar fix is needed for interpolate, but that should be it.

I think we need to step back and figure our what we want. What's point of having a dof if there is no cell in the general case? Would it be desirable to insist that a dof must always have a cell if this is possible?

Can you add a sketch to the PR an example where this problem arises?

jpdean:
Closing for now to investigate alternative approaches

PR #2233, opened 2022-06-08
Extend submesh to higher order geometries

jorgensd:
Current code does not actually work for higher order geometries (and does not throw an error).

This PR extends submeshes to higher order geometries (edges,facets,cell based submeshes).


jpdean:
My assembly tests all pass with these changes

cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -209,13 +208,14 @@ mesh::create_submesh(const Mesh& mesh, int dim,
                      const xtl::span<const std::int32_t>& entities)
 {
   // -- Submesh topology
+  const mesh::Topology& topology = mesh.topology();
```
garth-wells, line 211: `mesh::` not required.

cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -330,22 +329,61 @@ mesh::create_submesh(const Mesh& mesh, int dim,
   submesh_topology.set_connectivity(submesh_e_to_v, dim, 0);
 
   // -- Submesh geometry
+  const dolfinx::mesh::Geometry& geometry = mesh.geometry();
```
garth-wells, line 332: `dolfinx::mesh::` not required.

cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -330,22 +329,61 @@ mesh::create_submesh(const Mesh& mesh, int dim,
...
 
   // Get the geometry dofs in the submesh based on the entities in
   // submesh
-  const std::vector<std::int32_t> e_to_g
-      = entities_to_geometry(mesh, dim, submesh_to_mesh_entity_map, false);
-  const std::size_t num_vertices = mesh::num_cell_vertices(
-      cell_entity_type(mesh.topology().cell_type(), dim, 0));
+  const dolfinx::fem::ElementDofLayout layout
```
garth-wells, line 336: `dolfinx::` not required.

cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -330,22 +329,61 @@ mesh::create_submesh(const Mesh& mesh, int dim,
...
+        = topology.connectivity(dim, tdim);
+    assert(e_to_c);
+    const std::shared_ptr<const dolfinx::graph::AdjacencyList<int>> c_to_e
+        = topology.connectivity(tdim, dim);
+    assert(c_to_e);
+    for (std::size_t i = 0; i < submesh_to_mesh_entity_map.size(); ++i)
+    {
+      const std::int32_t idx = submesh_to_mesh_entity_map[i];
+      assert(!e_to_c->links(idx).empty());
+      // Always pick the last cell to be consistent with the e_to_v connectivity
```
garth-wells, line 363: Is this guaranteed? Or is it exploiting implicit behaviour that could change in the future?

cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -330,22 +329,61 @@ mesh::create_submesh(const Mesh& mesh, int dim,
...
+      const std::int32_t cell = e_to_c->links(idx).back();
+      const tcb::span<const int> cell_entities = c_to_e->links(cell);
```
garth-wells, line 365: This should be be `xtl::span`. Better would be `auto`.

cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -330,22 +329,61 @@ mesh::create_submesh(const Mesh& mesh, int dim,
...
+      auto it = std::find(cell_entities.begin(), cell_entities.end(), idx);
+      assert(it != cell_entities.end());
+      const auto local_entity = std::distance(cell_entities.begin(), it);
+      const std::vector<std::int32_t>& entity_dofs
+          = closure_dofs[dim][local_entity];
+
+      const tcb::span<const int> xc = xdofs.links(cell);
```
garth-wells, line 372: Use `auto`.

cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -413,14 +451,13 @@ mesh::create_submesh(const Mesh& mesh, int dim,
...
-      submesh_coord_cell, mesh.geometry().cmap().degree());
+      = cell_entity_type(geometry.cmap().cell_shape(), dim, 0);
+  auto submesh_coord_ele
+      = fem::CoordinateElement(submesh_coord_cell, geometry.cmap().degree());
 
   // Submesh geometry input_global_indices
   // TODO Check this
-  const std::vector<std::int64_t>& mesh_igi
-      = mesh.geometry().input_global_indices();
+  const std::vector<std::int64_t>& mesh_igi = geometry.input_global_indices();
```
garth-wells, line 460: Maybe a more instructive name than `mesh_igi`?

cpp/dolfinx/mesh/utils.cpp
```diff
@@ -470,7 +470,8 @@ mesh::entities_to_geometry(const Mesh& mesh, int dim,
   for (std::size_t i = 0; i < entities.size(); ++i)
   {
     const std::int32_t idx = entities[i];
-    const std::int32_t cell = e_to_c->links(idx).front();
+    // Always pick the second cell to be consistent with the e_to_v connectivity
```
garth-wells, line 473: See earlier comment on ordering.

python/test/unit/mesh/test_higher_order_mesh.py
```diff
@@ -36,10 +36,80 @@ def check_cell_volume(points, cell, domain, volume):
...
 
 
+@pytest.mark.skip_in_parallel
+@pytest.mark.parametrize('order', range(1, 5))
+def test_submesh(order):
+    points = []
+    points += [[i / order, j / order, 0] for j in range(order + 1)
+               for i in range(order + 1 - j)]
+    for k in range(1, order):
+        points += [[i / order, j / order + 0.1, k / order] for j in range(order + 1 - k)
```
garth-wells, line 49: Removing line break would improve readability.

python/test/unit/mesh/test_higher_order_mesh.py
```diff
@@ -36,10 +36,80 @@ def check_cell_volume(points, cell, domain, volume):
 
         mesh = create_mesh(MPI.COMM_WORLD, [ordered_cell], ordered_points, domain)
         area = assemble_scalar(form(1 * dx(mesh)))
-
         assert np.isclose(area, volume)
 
 
+@pytest.mark.skip_in_parallel
+@pytest.mark.parametrize('order', range(1, 5))
+def test_submesh(order):
```
garth-wells, line 44: This test is hard to follow. Explanation on what is being done would be helpful.

garth-wells (changes requested):

cpp/dolfinx/mesh/Mesh.cpp
jorgensd: This is how the implementation of `entity_to_vertex` is currently done. If we change how we create those connectivities, this would change how the local ordering of vertices is computed.

For exterior facet integrals, this has no effect as there is only one facet.
For interior facet integrals, one facet needs a permutation, to align the vertices of the child facet with the vertices of the parent cell.

The old implementation gives the correct vertices, but in a permuted order. This implementation makes sure we get the entities in the consistent order defined in basix.

cpp/dolfinx/mesh/Mesh.cpp
jorgensd, line 211:
```suggestion
  const Topology& topology = mesh.topology();
```

cpp/dolfinx/mesh/Mesh.cpp
jorgensd, line 332:
```suggestion
  const Geometry& geometry = mesh.geometry();
```

cpp/dolfinx/mesh/Mesh.cpp
jorgensd, line 336:
```suggestion
  const fem::ElementDofLayout layout
```

cpp/dolfinx/mesh/Mesh.cpp
jorgensd, line 365:
```suggestion
      auto cell_entities = c_to_e->links(cell);
```

cpp/dolfinx/mesh/Mesh.cpp
jorgensd, line 372:
```suggestion
      auto xc = xdofs.links(cell);
```

cpp/dolfinx/mesh/Mesh.cpp
jorgensd: Could we do that in a separate PR, as I haven't actually changed anything wrt. `mesh_igi`, except of some formatting.

python/test/unit/mesh/test_higher_order_mesh.py
jorgensd: The mesh generator is the same as we use everywhere else in the `test_higher_order_mesh.py`.

python/test/unit/mesh/test_higher_order_mesh.py
```diff
@@ -36,10 +36,80 @@ def check_cell_volume(points, cell, domain, volume):
...
+    points = []
```
jorgensd, line 45:
```suggestion
    # Generate a single cell higher order mesh
    points = []
```

python/test/unit/mesh/test_higher_order_mesh.py
```diff
@@ -36,10 +36,80 @@ def check_cell_volume(points, cell, domain, volume):
...
+    domain = ufl.Mesh(ufl.VectorElement(
+        "Lagrange", ufl.Cell("tetrahedron", geometric_dimension=3), order))
+
+    mesh = create_mesh(MPI.COMM_WORLD, [cell], points, domain)
+    for i in range(mesh.topology.dim):
+        mesh.topology.create_entities(i)
+    md = {"quadrature_degree": 10}
+    measures = (ufl.ds(mesh, metadata=md), ufl.dx(mesh, metadata=md))
+    dimensions = (mesh.topology.dim - 1, mesh.topology.dim)
+    for dim, dC in zip(dimensions, measures):
```
jorgensd, line 104: # Check that creating a submesh of single cell mesh, consisting of:
# 1. The cell
# 2. The facets of the cell
# Gives the correct computation of: volume (case 1) or surface area (case 2)

jorgensd, line 104:
```suggestion
    # Check that creating a submesh of single cell mesh, consisting of:
    # 1. The cell
    # 2. The facets of the cell
    # Gives the correct computation of: volume (case 1) or surface area (case 2)
    for dim, dC in zip(dimensions, measures):
```

jpdean (approved):
Looks good to me, and all my assembly tests still pass with these changes

chrisrichardson (approved):

jorgensd:
@garth-wells Can we merge this now? It has been thoroughly tested by @jpdean and adds important functionality to the class

garth-wells (approved):

PR #2234, opened 2022-06-09
Add CFFI to setup.py

jorgensd:
Add cffi to setup.py (as suggested by @minrk), required to make sure we have `libffi` (https://fenicsproject.discourse.group/t/installation-fenicsx-with-conda-on-macos/8492/5)


garth-wells (approved):
I think we should aim to remove `cffi` as a DOLFINx dependency and get what is needed from the returned FFCx (cffi) jitted objects. DOLFINx is using `cffi` directly for the tiniest things.

PR #2235, opened 2022-06-09
Fix for `interpolate` when some dofs don't belong to a cell on some processes

jpdean:
This fix is needed for the same reason as https://github.com/FEniCS/dolfinx/pull/2232.


jpdean:
Closing for now to investigate alternative approaches

PR #2244, opened 2022-06-16
Allow sparsity pattern to be built from integration entities

jpdean:
This PR makes it possible to build a sparsity pattern from a list of integration entities i.e. cell indices for cell integrals, (cell, local_facet) pairs for exterior facet integrals etc. It also modifies `create_sparsity_pattern` so that it only adds entries for entities in the form's integration domains, rather than for all entities of a particular integration type.

This PR is a step towards allowing the user to specify integration domains by passing a list of integration entities to `Form`, which would allow, for example, one-sided integrals on the interior facets of the mesh. I plan to put in another PR soon to add this functionality.

### DIFF
--- a/cpp/dolfinx/fem/sparsitybuild.cpp
+++ b/cpp/dolfinx/fem/sparsitybuild.cpp
@@ -30,6 +30,18 @@ void sparsitybuild::cells(
   }
 }
 //-----------------------------------------------------------------------------
+void cells(la::SparsityPattern& pattern,
+           const xtl::span<const std::int32_t>& cells,
+           const std::array<const std::reference_wrapper<const fem::DofMap>, 2>&
+               dofmaps)
+{
+  for (std::int32_t c : cells)
+  {
+    pattern.insert(dofmaps[0].get().cell_dofs(c),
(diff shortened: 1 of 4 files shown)


cpp/dolfinx/fem/sparsitybuild.cpp
```diff
@@ -30,6 +30,18 @@ void sparsitybuild::cells(
...
+void sparsitybuild::cells(
+    la::SparsityPattern& pattern, const xtl::span<const std::int32_t>& cells,
+    const std::array<const std::reference_wrapper<const fem::DofMap>, 2>&
+        dofmaps)
+{
+  for (std::int32_t c : cells)
+  {
+    pattern.insert(dofmaps[0].get().cell_dofs(c),
+                   dofmaps[1].get().cell_dofs(c));
+  }
```
jorgensd, line 42: Is it better with
```c++
  std::for_each(cells.cbegin(), cells.cend(),  auto [&dofmaps](auto c){
                       pattern.insert(dofmaps[0].get().cell_dofs(c), dofmaps[1].get().cell_dofs(c));
                       );
```
?

garth-wells (approved):

PR #2255, opened 2022-06-28
Remove some xtensor from refinement code.

jorgensd:
- Use plain loops to copy. 
- Keep midpoint structure flat
- Reduce size of new_geometry from the start, instead of doing a view copy at the end.
- Remove comment about things that should work with icpx


cpp/dolfinx/refinement/utils.cpp
```diff
@@ -86,16 +86,16 @@ xt::xtensor<double, 2> create_new_geometry(
...
   const std::size_t num_new_vertices = local_edge_to_new_vertex.size();
   xt::xtensor<double, 2> new_vertex_coordinates(
-      {num_vertices + num_new_vertices, 3});
+      {num_vertices + num_new_vertices, gdim});
 
   for (std::size_t v = 0; v < num_vertices; ++v)
   {
     const int pos = 3 * vertex_to_x[v];
-    for (std::size_t j = 0; j < 3; ++j)
+    for (std::size_t j = 0; j < gdim; ++j)
```
garth-wells, line 98: It's better for performance to have a fixed loop bound here.

jorgensd: Even if it means that we need to do a strided copy when returning the data?

garth-wells: Make everything 3D.

jorgensd: We need to remove the last dimensions before passing this to the geometry constructor, as it computes the gdim from the shape of the input.

garth-wells (approved):

PR #2262, opened 2022-07-08
Fix custom kernel subdomains

michalhabera:


garth-wells:
Could we add a test?

michalhabera:
> Could we add a test?

Done.

garth-wells (approved):

PR #2266, opened 2022-07-11
Add Real Space

jhale:
Builds on previous work of @mscroggs.

Currently brings in tensor related branches from @garth-wells so would need to wait before merging.


cpp/demo/real/main.cpp
```diff
@@ -0,0 +1,27 @@
+#include "real.h"
```
jhale, line 1: I will remove this, was useful for debugging and looking at generated code.

cpp/demo/real/real.py
```diff
@@ -0,0 +1,7 @@
+from ufl import FiniteElement, triangle, TrialFunction, TestFunction, dx, VectorElement, inner
```
jhale, line 1: To remove.

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -50,20 +51,23 @@ xt::xtensor<double, 4>
...
-  return _element->tabulate(n, X, {X.shape(0), X.shape(1)});
+  auto [tab, shape] = _element->tabulate(n, X, {X.shape(0), X.shape(1)});
+  return xt::adapt(tab, shape);
 }
 //--------------------------------------------------------------------------------
 void CoordinateElement::tabulate(int n, const xt::xtensor<double, 2>& X,
                                  xt::xtensor<double, 4>& basis) const
 {
-  _element->tabulate(n, X, basis);
+  assert(_element);
```
jhale, line 61: From Garth's branch.

cpp/demo/real/main.cpp
jhale, line 1: To remove.

michalhabera:
What is the parallel ownership of real element dofs?

For performance reasons we could pack coefficients of Real element in a similar way to how Constants are packed - outside of the main assembly loop, just once per domain. Legacy approach to Real element has many bottlenecks, I thought we opted for redesign.

tbrandvik:
I have some old Dolfin-based code that makes use of real spaces that I am trying to port to DolfinX. I was wondering if the capability in this branch is likely to be merged with the main branch in the near future or if I am better off using this branch directly? Thanks!

jhale:
Perhaps not @tbrandvik, because I agree with @michalhabera that there are better ways of doing this. I have a research problem that required this implementation and I had to get it done quickly (it worked for my purposes).

I would add that this branch does not currently work in parallel, and the inner product of a real function and a finite element function in a form is also not giving the right output. So be careful!

tbrandvik:
Thanks @jhale - that is useful information. Hopefully support for real spaces become part of the main branch at some point!

GSadashivaiah27:
Hi,

As I understand real space is still not available within DolfinX? or Is there a way to use it.

thisch:
I'm also interested in this feature. BTW Are there plans when the tensor related branches by @garth-wells will be merged?

ramrebol:
I'm also interested in this feature. Do you think it will be incorporated?

garth-wells:
It will be added at some point (hopefully soon). I still needs some careful thought on how to do it while maintaining decent performance.

ramrebol:
thanks @garth-wells for answer. May you suggest me how to impose the integral of the pressure equal to zero in a Stokes problem? (Typical stationary Stokes problem with Dirichlet boundary condition only).

jkrokowski:
Any status updates on this issue? I have an application in which I would like to use real type spaces.

garth-wells:
We need to come back to this with a focused effort. Closing for now.

PR #2269, opened 2022-07-15
Let integration entities be specified manually

jpdean:
Currently, `Form` computes which entities to integrate over from a `MeshTags` object. This is fine for many common use cases, but in more specialised applications, it is useful to be able to specify the entities to assemble over manually. For instance, the user may want to perform a "one sided" facet integral on some interior facets of the mesh, which is not currently possible.

This PR refactors `Form` so that the entities to integrate over can be specified manually. A free function has been added to compute integration entities from meshtags, rather than this being computed in  the form `Form` class.

There are still a few issues to fix:
1. In python, since a plain dictionary does not have a `ufl_id`, it causes [this](https://github.com/FEniCS/ufl/blob/main/ufl/form.py#L438)  check to fail. However, the check may not be necessary. If it is, `id` could be called on the data instead.
2. There are a couple of places where data is being copied. There are a few ways to avoid these copies, but I'm not sure which approach would be best. Any suggestions would be much appreciated.
3. Any suggestions on the best way to do the python bindings would be very helpful. At the moment, facets are flattened in both python and C++. I think it would be better to only flatten them in C++, but I'm open to suggestions!


cpp/dolfinx/fem/Form.h
```diff
@@ -93,12 +93,11 @@ class Form
   Form(const std::vector<std::shared_ptr<const FunctionSpace>>& function_spaces,
        const std::map<
            IntegralType,
-           std::pair<
-               std::vector<std::pair<
-                   int, std::function<void(T*, const T*, const T*,
-                                           const scalar_value_type_t*,
-                                           const int*, const std::uint8_t*)>>>,
-               const mesh::MeshTags<int>*>>& integrals,
+           std::map<int, std::pair<std::function<
```
garth-wells, line 96: It's not entirely clear in the doc string what the map key is.

cpp/dolfinx/fem/Form.h
```diff
@@ -118,38 +117,25 @@ class Form
     if (!_mesh)
       throw std::runtime_error("No mesh could be associated with the Form.");
 
-    // Store kernels, looping over integrals by domain type (dimension)
-    for (auto& integral_type : integrals)
+    // Store kernels, looping over integrals by type
+    // NOTE Could probably simplify Form by combining integral types
+    for (const auto& integral_type : integrals)
```
garth-wells, line 122: This would be easy to read if it was something like 
```cpp
for (const auto& [key, integral] : integrals)
```

cpp/dolfinx/fem/Form.h
```diff
@@ -118,38 +117,25 @@ class Form
...
       const IntegralType type = integral_type.first;
-      // Loop over integrals kernels and set domains
+      // Loop over integrals kernels and set domains (intentially copying
+      // the domain data)
       switch (type)
       {
       case IntegralType::cell:
-        for (auto& integral : integral_type.second.first)
-          _cell_integrals.insert({integral.first, {integral.second, {}}});
+        _cell_integrals = integrals.at(type);
```
garth-wells, line 130: Could this be:
```cpp
_cell_integrals = integral_type.second;
```

cpp/dolfinx/fem/Form.h
```diff
@@ -93,12 +93,11 @@ class Form
...
-               std::vector<std::pair<
-                   int, std::function<void(T*, const T*, const T*,
-                                           const scalar_value_type_t*,
-                                           const int*, const std::uint8_t*)>>>,
-               const mesh::MeshTags<int>*>>& integrals,
+           std::map<int, std::pair<std::function<
+                                       void(T*, const T*, const T*,
+                                            const scalar_value_type_t*,
+                                            const int*, const std::uint8_t*)>,
+                                   std::vector<std::int32_t>>>>& integrals,
```
garth-wells, line 100: Is `std::vector` the best container, it could it be a span? (or possibly a C++20 range)

cpp/dolfinx/fem/Form.h
```diff
@@ -598,7 +405,8 @@ class Form
         {
           // There will only be one pair for an exterior facet integral
           std::array<std::int32_t, 2> pair
-              = get_cell_local_facet_pairs<1>(f, f_to_c->links(f), *c_to_f)[0];
+              = mesh::get_cell_local_facet_pairs<1>(f, f_to_c->links(f),
```
garth-wells, line 408: Not changed in this PR, but docstring for `get_cell_local_facet_pairs` is wrong.

cpp/dolfinx/fem/utils.cpp
```diff
@@ -277,3 +277,110 @@ fem::FunctionSpace fem::create_functionspace(
...
+                         topology.index_map(dim)->size_local());
+
+  switch (integral_type)
+  {
+  case fem::IntegralType::cell:
+  {
+    for (auto c = tagged_entities.cbegin(); c != entity_end; ++c)
+    {
+      const std::size_t index = std::distance(tagged_entities.cbegin(), c);
+      integrals[tags[index]].push_back(*c);
```
garth-wells, line 311: This does a map lookup every time. Move map lookup outside of tight loop.

cpp/dolfinx/fem/utils.cpp
```diff
@@ -277,3 +277,110 @@ fem::FunctionSpace fem::create_functionspace(
...
+    case IntegralType::exterior_facet:
+    {
+      // When a mesh is not ghosted by cell, it is not straightforward
+      // to distinguish between (i) exterior facets and (ii) interior
+      // facets that are on a partition boundary. If there are no
+      // ghost cells, build a set of owned facts that are ghosted on
+      // another process to help determine if a facet is on an
+      // exterior boundary.
+      assert(topology.index_map(tdim - 1));
+      const std::vector<std::int32_t> fwd_shared_facets
```
garth-wells, line 333: I think we now have a function that determined exterior facets without relying on the ghosting type.

cpp/dolfinx/fem/utils.cpp
```diff
@@ -277,3 +277,110 @@ fem::FunctionSpace fem::create_functionspace(
...
+                                  fwd_shared_facets.end(), *f))
+          {
+            const std::size_t index
+                = std::distance(tagged_entities.cbegin(), f);
+
+            // There will only be one pair for an exterior facet integral
+            const std::array<std::int32_t, 2> pair
+                = mesh::get_cell_local_facet_pairs<1>(*f, f_to_c->links(*f),
+                                                      *c_to_f)[0];
+            integrals[tags[index]].insert(integrals[tags[index]].end(),
```
garth-wells, line 353: Avoid map lookup.

cpp/dolfinx/fem/utils.h
```diff
@@ -226,13 +238,15 @@ std::vector<std::string> get_constant_names(const ufcx_form& ufcx_form);
...
-    const std::vector<std::shared_ptr<const Constant<T>>>& constants,
-    const std::map<IntegralType, const mesh::MeshTags<int>*>& subdomains,
-    const std::shared_ptr<const mesh::Mesh>& mesh = nullptr)
+Form<T>
+create_form(const ufcx_form& ufcx_form,
+            const std::vector<std::shared_ptr<const FunctionSpace>>& spaces,
+            const std::vector<std::shared_ptr<const Function<T>>>& coefficients,
+            const std::vector<std::shared_ptr<const Constant<T>>>& constants,
+            const std::map<IntegralType,
+                           std::map<std::int32_t, std::vector<std::int32_t>>>&
```
garth-wells, line 247: Check what the best type is for the list of entities.

cpp/dolfinx/fem/utils.h
```diff
@@ -305,19 +320,26 @@ Form<T> create_form(
     }
     assert(k);
 
-    integral_data[IntegralType::cell].first.emplace_back(cell_integral_ids[i],
-                                                         k);
+    // FIXME Avoid copy (see also int and ext facets)
+    const int id = cell_integral_ids[i];
+    if (id == -1)
+    {
+      integral_data[IntegralType::cell][id] = {k, {}};
```
garth-wells, line 327: Use best to avoid `[]` access for maps. There are alternative for add entries.

cpp/dolfinx/mesh/Mesh.h
```diff
@@ -147,4 +147,32 @@ std::tuple<Mesh, std::vector<std::int32_t>, std::vector<std::int32_t>,
 create_submesh(const Mesh& mesh, int dim,
                const std::span<const std::int32_t>& entities);
 
+/// Helper function to get a std::vector of (cell, local_facet) pairs
```
garth-wells, line 150: Return type isn't a `std::vector`.

Best to avoid using types in the doctstring. The function signature specifies the type.

cpp/dolfinx/mesh/Mesh.h
```diff
@@ -147,4 +147,32 @@ std::tuple<Mesh, std::vector<std::int32_t>, std::vector<std::int32_t>,
...
+/// corresponding to a given facet index.
+/// @param[in] f Facet index
+/// @param[in] cells The cells connected to the facet
+/// @param[in] c_to_f Cell to facet connectivity
+/// @return Vector of (cell, local_facet) pairs
```
garth-wells, line 155: This needs some more explanation.

cpp/dolfinx/mesh/Mesh.h
```diff
@@ -147,4 +147,32 @@ std::tuple<Mesh, std::vector<std::int32_t>, std::vector<std::int32_t>,
...
+template <int num_cells>
```
garth-wells, line 156: Best to avoid `int` here and be explicit with the integer size.

python/dolfinx/wrappers/fem.cpp
```diff
@@ -774,56 +774,52 @@ void declare_form(py::module& m, const std::string& type)
...
       m, pyclass_name_form.c_str(), "Variational form object")
-      .def(
-          py::init(
-              [](const std::vector<std::shared_ptr<
-                     const dolfinx::fem::FunctionSpace>>& spaces,
-                 const std::map<
+      .def(py::init(
+               [](const std::vector<std::shared_ptr<
+                      const dolfinx::fem::FunctionSpace>>& spaces,
+                  const std::map<
```
garth-wells, line 780: I think it would be better to use `py::dict` here, and `py::array_t`. `std::vector` doesn't cast efficiently for large arrays, and the `std::map` isn't passed directly to the C++ layer so the hidden cast from `py::dict` to `std::map` could be avoided.

python/dolfinx/wrappers/fem.cpp
```diff
@@ -1007,6 +1007,15 @@ void fem(py::module& m)
   m.def("transpose_dofmap", &dolfinx::fem::transpose_dofmap,
         "Build the index to (cell, local index) map from a "
         "dofmap ((cell, local index ) -> index).");
+  // FIXME Where is the best place to put this?
+  m.def(
+      "compute_integration_domains",
```
garth-wells, line 1012: This will need a more elaborate wrapper to return `py::array_t`s rather than `std::vector`

python/test/unit/fem/test_assemble_domains.py
```diff
@@ -221,3 +221,112 @@ def test_additivity(mode):
...
+    # Create measures
+    dx_mt = ufl.Measure("dx", subdomain_data=mt_cells, domain=msh)
+    ds_mt = ufl.Measure("ds", subdomain_data=mt_facets, domain=msh)
+    dS_mt = ufl.Measure("dS", subdomain_data=mt_facets, domain=msh)
+
+    g = Function(V)
+    g.interpolate(lambda x: x[1]**2)
+
+    def create_forms(dx, ds, dS):
+        a = form(ufl.inner(g * u, v) * (dx(0) + dx(7) + ds(6))
```
garth-wells, line 274: I think would be easier to read on a single line.

garth-wells (changes requested):
Looking good - first pass of comments.

python/dolfinx/fem/forms.py
```diff
@@ -148,11 +149,29 @@ def _form(form):
...
-        # Subdomain markers (possibly None for some dimensions)
-        subdomains = {_cpp.fem.IntegralType.cell: subdomains.get("cell"),
-                      _cpp.fem.IntegralType.exterior_facet: subdomains.get("exterior_facet"),
-                      _cpp.fem.IntegralType.interior_facet: subdomains.get("interior_facet"),
-                      _cpp.fem.IntegralType.vertex: subdomains.get("vertex")}
+        # NOTE Could remove this and let the user convert mashtags by calling
+        # compute_integration_domains themselves
+        def get_integration_domains(integral_type, subdomains):
+            subdomain = subdomains.get(integral_type.name)
+            if isinstance(subdomain, dolfinx.mesh.MeshTagsMetaClass) \
```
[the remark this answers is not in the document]
jpdean: Alternatively, I could remove this function altogether and let the user call `compute_integration_domains` themselves if needed. This would be more explicit. What're your thoughts?

jorgensd:
What is missing to get this merged?

jpdean:
> What is missing to get this merged?

There are still a few comments from the code review that need to be addressed

garth-wells:
See #2522.

garth-wells:
@jpdean could you fix the merge conflict?

python/dolfinx/wrappers/fem.cpp
```diff
@@ -413,11 +413,13 @@ void declare_form(py::module& m, const std::string& type)
...
                       const dolfinx::fem::Constant<T>>>& constants,
-                  const std::map<dolfinx::fem::IntegralType,
-                                 const dolfinx::mesh::MeshTags<int>*>&
+                  const std::map<
+                      dolfinx::fem::IntegralType,
+                      std::vector<std::pair<std::int32_t, std::vector<std::int32_t>>>>&
                       subdomains,
                   std::shared_ptr<const dolfinx::mesh::Mesh> mesh)
                {
+                 // FIXME Use py::array instead of std::vector?
```
garth-wells, line 422: `std::vector` can be very slow in pybind11. We need 
```cpp
const std::map<dolfinx::fem::IntegralType,
               std::vector<std::pair<std::int32_t, py:array_t<std::int32_t>>>>&
               subdomains;
```

garth-wells (changes requested):
Last task is to replace some `std::vector<std::int32_t>` with `py::array_t`.

Might need to do some copies, but maybe some `std::vector` can be changed to `std::span` in C++ layer. Depends on how clever casting is.

garth-wells (approved):

PR #2270, opened 2022-07-15
Remove xtensor from geometry code

garth-wells:
Makes operations, and performance implications, clearer.

Also removes xtensor in other places where the benefit is not compelling.

### DIFF
--- a/cpp/dolfinx/common/TimeLogger.cpp
+++ b/cpp/dolfinx/common/TimeLogger.cpp
@@ -7,6 +7,7 @@
 #include "TimeLogger.h"
 #include <dolfinx/common/MPI.h>
 #include <dolfinx/common/log.h>
+#include <iostream>
 #include <variant>
 
 using namespace dolfinx;
--- a/cpp/dolfinx/geometry/gjk.cpp
+++ b/cpp/dolfinx/geometry/gjk.cpp
@@ -8,38 +8,28 @@
 #include <dolfinx/common/math.h>
 #include <numeric>
 #include <stdexcept>
-#include <xtensor/xadapt.hpp>
-#include <xtensor/xbuilder.hpp>
-#include <xtensor/xfixed.hpp>
-#include <xtensor/xnorm.hpp>
-#include <xtensor/xsort.hpp>
-#include <xtensor/xtensor.hpp>
-#include <xtensor/xview.hpp>
 
 using namespace dolfinx;
 
 namespace
 {
 
-// Find the resulting sub-simplex of the input simplex which is nearest to the
-// origin. Also, return the shortest vector from the origin to the resulting
-// simplex.
+// Find the resulting sub-simplex of the input simplex which is nearest
+// to the origin. Also, return the shortest vector from the origin to
+// the resulting simplex.
 std::pair<std::vector<double>, std::array<double, 3>>
-nearest_simplex(const std::span<const double>& ss)
+nearest_simplex(const std::span<const double>& s)
 {
-  auto s = xt::adapt(ss.data(), ss.size(), xt::no_ownership(),
-                     std::vector<std::size_t>{ss.size() / 3, 3});
-
-  assert(ss.size() % 3 == 0);
-  const std::size_t s_rows = s.shape(0);
+  assert(s.size() % 3 == 0);
+  const std::size_t s_rows = s.size() / 3;
   switch (s_rows)
   {
   case 2:
   {
-    auto s0 = ss.subspan(0, 3);
-    auto s1 = ss.subspan(3, 3);
+    auto s0 = s.subspan(0, 3);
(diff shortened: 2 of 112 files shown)


cpp/dolfinx/common/TimeLogger.cpp
```diff
@@ -7,6 +7,7 @@
 #include "TimeLogger.h"
 #include <dolfinx/common/MPI.h>
 #include <dolfinx/common/log.h>
+#include <iostream>
```
jorgensd, line 10: Is this include needed?

cpp/dolfinx/geometry/gjk.cpp
```diff
@@ -6,61 +6,74 @@
...
   {
   case 2:
   {
-    auto s0 = xt::row(s, 0);
-    auto s1 = xt::row(s, 1);
-    const double lm = -xt::sum(s0 * (s1 - s0))() / xt::norm_sq(s1 - s0)();
+    auto s0 = s.subspan(0, 3);
+    auto s1 = s.subspan(3, 3);
+    std::array ds = {s1[0] - s0[0], s1[1] - s0[1], s1[2] - s0[2]};
+    const double lm = -(s0[0] * ds[0] + s0[1] * ds[1] + s0[2] * ds[2])
```
jorgensd, line 32: Could we add a comment here stating what `lm` is, i.e. dot(s0, ds / |ds|)

cpp/dolfinx/geometry/gjk.cpp
```diff
@@ -6,61 +6,74 @@
...
+    {
+      return std::transform_reduce(
+          x.begin(), x.end(), y.begin(), 0.0, std::plus{},
+          [](auto x, auto y) { return (x - y) * (x - y); });
+    };
+    const double ab2 = length(a, b);
+    const double ac2 = length(a, c);
+    const double bc2 = length(b, c);
+
+    auto helper = [](auto& x, auto& y)
```
jorgensd, line 63: Maybe add a note saying that this is equivalent to `dot(x, x-y)`?

cpp/dolfinx/common/TimeLogger.cpp
garth-wells: Yes, because `std::cout` is used in the file.

cpp/dolfinx/geometry/gjk.cpp
```diff
@@ -69,90 +84,122 @@ nearest_simplex(const xt::xtensor<double, 2>& s)
...
-    xt::xtensor_fixed<double, xt::xshape<3>> vmin = xt::row(s, i);
-    double qmin = xt::norm_sq(vmin)();
+    std::size_t pos = 0;
+    {
+      double norm0 = std::numeric_limits<double>::max();
+      for (std::size_t i = 0; i < s.size(); i += 3)
+      {
+        double norm = 0.0;
+        for (std::size_t k = i; k < i + 3; ++k)
+          norm += s[k] * s[k];
```
jorgensd, line 113: Is there a reason for not using `transform_reduce` here?

cpp/dolfinx/geometry/gjk.cpp
```diff
@@ -161,89 +208,88 @@ nearest_simplex(const xt::xtensor<double, 2>& s)
...
     throw std::runtime_error("Number of rows defining simplex not supported.");
   }
-  }
 }
 //----------------------------------------------------------------------------
-// Support function, finds point p in bd which maximises p.v
-xt::xtensor_fixed<double, xt::xshape<3>>
-support(const xt::xtensor<double, 2>& bd,
-        const xt::xtensor_fixed<double, xt::xshape<3>>& v)
+// Helper function, finds point p in bd which maximises p.v
```
jorgensd, line 218: It should be called the "Support function", as this is the name of it in the actual algorithm

cpp/dolfinx/geometry/gjk.cpp
```diff
@@ -161,89 +208,88 @@ nearest_simplex(const xt::xtensor<double, 2>& s)
...
+                              const std::array<double, 3>& v)
 {
   int i = 0;
-  double qmax = xt::sum(xt::row(bd, 0) * v)();
-  for (std::size_t m = 1; m < bd.shape(0); ++m)
+  double qmax = bd[0] * v[0] + bd[1] * v[1] + bd[2] * v[2];
+  for (std::size_t m = 1; m < bd.size() / 3; ++m)
   {
-    double q = xt::sum(xt::row(bd, m) * v)();
+    // q = xt::sum(xt::row(bd, m) * v)();
```
jorgensd, line 226: Remove leftover comment?

cpp/dolfinx/geometry/gjk.cpp
garth-wells: A bit too heavyweight.

PR #2271, opened 2022-07-18
Support mixed domain assembly

jpdean:
This PR adds support for mixed domain assembly. Forms with test functions, trial functions, and coefficients defined over different meshes can be assembled. The user must provide a map to relate mesh entities in the integration domain to mesh entities in the mesh the function is defined over; this is best illustrated by the new tests.

The codimension 1 case requires [this](https://github.com/FEniCS/ffcx/tree/jpdean/mixed_dim) branch of FFCX. In a future PR, I will add permutations which are required in some cases.

There are still lots of parts of the code that need to be tidied up / refined. Any feedback on the implementation approach would be much appreciated.


cpp/dolfinx/fem/sparsitybuild.h
```diff
@@ -36,10 +36,16 @@ namespace sparsitybuild
 /// over
 /// @param[in] dofmaps The dofmap to use in building the sparsity
 /// pattern
+/// @param[in] cell_maps Maps from the cells in the integration domain to
+/// the cells in each function space in the form.
```
chrisrichardson, line 40: Add a description of the default.

cpp/dolfinx/fem/sparsitybuild.h
```diff
@@ -48,9 +54,15 @@ void cells(
 /// @param[in] dofmaps The dofmap to use in building the sparsity
 /// pattern
 /// @note The sparsity pattern is not finalised
+/// @param[in] cell_maps Maps from the cells in the integration domain to
+/// the cells in each function space in the form.
```
chrisrichardson, line 58: ditto: describe default

cpp/dolfinx/fem/Form.h
```diff
@@ -102,18 +108,24 @@ class Form
...
     if (!_mesh and !function_spaces.empty())
       _mesh = function_spaces[0]->mesh();
     for (const auto& V : function_spaces)
     {
-      if (_mesh != V->mesh())
-        throw std::runtime_error("Incompatible mesh");
+      if (_mesh != V->mesh()
+          and entity_maps.find(V->mesh()) == entity_maps.end())
+        throw std::runtime_error(
+            "Incompatible mesh. entity_maps must be provided.");
```
chrisrichardson, line 128:
```suggestion
            "Incompatible mesh: entity_maps must be provided.");
```
Just a minor tweak

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -80,6 +85,12 @@ void assemble_cells(
   for (std::size_t index = 0; index < cells.size(); ++index)
   {
     std::int32_t c = cells[index];
+    // Map the cell in the integration domain to the cell in the mesh
+    // each function space is defined over
+    std::int32_t c_0 = cell_map_0(cells.subspan(index, 1));
```
chrisrichardson, line 90: What is the point of `subspan` here? Isn't it just the same as `cells[index]`?

chrisrichardson: OK, I see that `cell_map_0` has to take a `span` argument... But if it is always length 1, does that make sense? I guess for facets, it is length 2, so it is consistent...

garth-wells:
Closing for now, re-open later when ready.

PR #2276, opened 2022-07-20
Add PML demo - GSoC 2022

mikics:
Currently, there are no demos in DOLFINx showing how to implement perfectly matched layers (PMLs) for time-harmonic electromagnetic problems. 

The aim of this pull request is to add such demo. In particular, the demo will show how to implement rectangular PML for a simple electromagnetic scattering problem in two dimensions (GSoC 2022).


python/demo/demo_pml.py
```diff
@@ -0,0 +1,225 @@
...
+
+def pml_coordinates(x, alpha, k0, l_dom, l_pml):
+
+    # Define boolean functions for the PML regions
+    inside_pml_x = (ufl.sign(ufl.sign(x[0])*x[0] - l_dom/2) + 1)/2
+    inside_pml_y = (ufl.sign(ufl.sign(x[1])*x[1] - l_dom/2) + 1)/2
+
+    # Define the coordinate transformation for PML regions
+    x_pml = x[0] + 1j*alpha/k0*x[0]*(ufl.sign(x[0])*x[0] - l_dom/2)/(l_pml/2 - l_dom/2)**2*inside_pml_x
+    y_pml = x[1] + 1j*alpha/k0*x[1]*(ufl.sign(x[1])*x[1] - l_dom/2)/(l_pml/2 - l_dom/2)**2*inside_pml_y
```
jorgensd, line 65:
```suggestion
    x_pml = x[0] + 1j*alpha/k0*x[0]*(ufl.sign(x[0])*x[0] - l_dom/2)/(l_pml/2 - l_dom/2)**2*inside_pml_x
    y_pml = x[1] + 1j*alpha/k0*x[1]*(ufl.sign(x[1])*x[1] - l_dom/2)/(l_pml/2 - l_dom/2)**2*inside_pml_y
```
Instead of repeating `x_pml` and `y_pml`, you can use `ufl.Index()`, as illustrated in the following minimal example:
```python
import dolfinx
from mpi4py import MPI
import ufl

V = dolfinx.fem.FunctionSpace(mesh, ("CG", 1))
u = dolfinx.fem.Function(V)
u.interpolate(lambda x: x[0])
x = ufl.SpatialCoordinate(mesh)


def F(u, x):
    qx = 2 * u * x[0]
    qy = 2 * u * x[1]
    return ufl.as_vector((qx, qy))


form1 = dolfinx.fem.form(ufl.inner(F(u, x), F(u, x))*ufl.dx)
print(dolfinx.fem.assemble_scalar(form1))


def G(u, x):
    i = ufl.Index()
    qi = 2 * u * x[i]
    return ufl.as_tensor(qi, i)


Gi = G(u, x)
form2 = dolfinx.fem.form(ufl.inner(Gi, Gi)*ufl.dx)
print(dolfinx.fem.assemble_scalar(form2))
```
returning
```bash
1.244444444444446
1.244444444444446
```

python/demo/demo_pml.py
```diff
@@ -0,0 +1,225 @@
...
+        ax = np.sin(self.theta)
+        ay = np.cos(self.theta)
+
+        return (-ax * np.exp(1j * phi), ay * np.exp(1j * phi))
+
+def curl_2d(a):
+    """Returns the curl of two 2D vectors as a 3D vector"""
+    ay_x = a[1].dx(0)
+    ax_y = a[0].dx(1)
+    return as_vector((0, 0, ay_x - ax_y))
```
jorgensd, line 55: Do this as a single line without defining ay_x and ax_y

python/demo/demo_pml.py
```diff
@@ -0,0 +1,225 @@
...
+
+problem = fem.petsc.LinearProblem(a, L, bcs=[], petsc_options={
+                                  "ksp_type": "preonly", "pc_type": "lu"})
+Esh = problem.solve()
+
+V_dg = fem.VectorFunctionSpace(mesh, ("DG", degree))
+Esh_dg = fem.Function(V_dg)
+Esh_dg.interpolate(Esh)
+
+with VTXWriter(MPI.COMM_WORLD, "Esh.bp", Esh_dg) as f:
```
jorgensd, line 183: Instead of MPI.COMM_WORLD use mesh.comm whenever you need a communicator relating to a function or a mesh

python/demo/demo_pml.py
```diff
@@ -0,0 +1,225 @@
...
+Es = ufl.TrialFunction(V)
+v = ufl.TestFunction(V)
+
+Es_3d = as_vector((Es[0], Es[1], 0))
+v_3d = as_vector((v[0], v[1], 0))
+
+dx = ufl.Measure("dx", mesh, subdomain_data=cell_tags)
+dAu = dx(au_tag)
+dBkg = dx(bkg_tag)
+dDom = dAu + dBkg
```
jorgensd, line 151:
```suggestion
dDom = dx((au_tag, bkg_tag))
```

python/demo/demo_pml.py
```diff
@@ -0,0 +1,225 @@
...
+reps_au = -1.0782
+ieps_au = 5.8089
+eps_au = reps_au + ieps_au * 1j
+
+D = fem.FunctionSpace(mesh, ("DG", 0))
+eps = fem.Function(D)
+au_cells = cell_tags.find(au_tag)
+bkg_cells = cell_tags.find(bkg_tag)
+eps.x.array[au_cells] = np.full_like(
+    au_cells, reps_au + ieps_au * 1j, dtype=np.complex128)
```
jorgensd, line 163: Isn't this `eps_au`?

python/demo/demo_pml.py
```diff
@@ -0,0 +1,225 @@
...
+
+dx = ufl.Measure("dx", mesh, subdomain_data=cell_tags)
+dAu = dx(au_tag)
+dBkg = dx(bkg_tag)
+dDom = dAu + dBkg
+dPml = dx(pml_tag)
+
+reps_au = -1.0782
+ieps_au = 5.8089
+eps_au = reps_au + ieps_au * 1j
```
jorgensd, line 156: Define this as a single number, and use `np.real` and `np.imag` when you explicitly need the individual components.

python/demo/demo_pml.py
```diff
@@ -0,0 +1,225 @@
...
+
+E_dg = fem.Function(V_dg)
+E_dg.interpolate(E)
+
+with VTXWriter(MPI.COMM_WORLD, "E.bp", E_dg) as f:
+    f.write(0.0)
+
+q_abs_analyt, q_sca_analyt, q_ext_analyt = calculate_analytical_efficiencies(
+    reps_au,
+    ieps_au,
```
jorgensd, line 197: Combine these into one input argument (complex number)

jhale:
@mikics I'm getting a fail on this for complex64 in the efficiency assertions. In the eye norm the numerical calculations don't look too bad. Is the tolerance a bit tight?

mikics:
> @mikics I'm getting a fail on this for complex64 in the efficiency assertions. In the eye norm the numerical calculations don't look too bad. Is the tolerance a bit tight?

On my laptop (with complex32) the demo passes the test both in serial and in parallel (even though the error increases in the parallel version). I can try to refine the mesh around the boundary, and see if the error diminishes in the complex64 version.

jhale:
Few things you could investigate:

* What's changed since [8d06b41](https://github.com/FEniCS/dolfinx/pull/2276/commits/8d06b414732dce63bbc823af732ed42d01f05318) when you were running green?
* We're moving the images back to PETSc 3.17 this morning, had a lot of problems with 3.18. Worth waiting to see if that changes anything.
* 32 and 64 simply relate to the global indexing, not to the data type of the complex scalar value. I wouldn't expect to see any difference.
* The docker image `fenicsproject/test-env:nightly` has complex32 and complex64 versions of PETSc which you can build DOLFINx against and experiment yourself.

jhale:
OK, it's running green now. Would still be interested to know if you can locate the cause of this error is before merging:

https://github.com/FEniCS/dolfinx/actions/runs/3733081716/jobs/6333395279

mikics:
> OK, it's running green now. Would still be interested to know if you can locate the cause of this error is before merging:
> 
> https://github.com/FEniCS/dolfinx/actions/runs/3733081716/jobs/6333395279

Will investigate that! We also have a duplicate module error now, since the same function for the calculation of analytical efficiencies is used for the demo for scattering boundary conditions. I will take care of that as well.

jhale:
Thanks! Also, on this PR could you add doc links to the two that will be merged?

https://docs.fenicsproject.org/dolfinx/main/python/demos.html#all-demos

mikics:
> Few things you could investigate:
> 
>  - What's changed since [8d06b41](https://github.com/FEniCS/dolfinx/pull/2276/commits/8d06b414732dce63bbc823af732ed42d01f05318) when you were running green?
> 

I have analyzed the ouputs of the CI checks for the [8d06b41](https://github.com/FEniCS/dolfinx/pull/2276/commits/8d06b414732dce63bbc823af732ed42d01f05318) (working commit) and [0eca2f2](https://github.com/FEniCS/dolfinx/pull/2276/commits/0eca2f2f3f0fcb9fe11104ad9625e44cce05f007) (first non-working commit). By looking at the `petsc4py` version, it seems that the two commits were checked with two different PETSc versions:

- [3.17.4](https://github.com/FEniCS/dolfinx/actions/runs/3521482312/jobs/5903375751#step:19:37) (working commit)
- [3.18.1](https://github.com/FEniCS/dolfinx/actions/runs/3576063279/jobs/6013408784#step:19:35) (not working commit)

I have also quickly checked the `git diff` of the two commits, and it doesn't seem there are major changes in the `dolfinx` code, but I have to better look into that to have a clear idea.

I would also like to check if the error obtained with PETSc 3.17.4 ("working" version) is affected in any way by the number of bits for the indexing, since the `assert` only tells us that the error is below 1%, but not if it changes. This may give us further hints on what is the problem.

jhale:
Thanks for investigating. The problems related to PETSc 3.18 are more to do with how our CI system is setup (using different compilers for PETSc and DOLFINx). 

You are running green now on 3.17 so I think it's OK for us to merge.

python/demo/demo_pml/demo_pml.py
```diff
@@ -0,0 +1,606 @@
...
+v = ufl.TestFunction(V)
+
+# Definition of 3d fields
+Es_3d = ufl.as_vector((Es[0], Es[1], 0))
+v_3d = ufl.as_vector((v[0], v[1], 0))
+
+# Measures for subdomains
+dx = ufl.Measure("dx", domain, subdomain_data=cell_tags)
+dDom = dx((au_tag, bkg_tag))
+dPml_xy = dx(pml_tag)
```
thisch, line 273: Here you're defining three domains for the three parts of the PML, which you then use in your definition of the weak form. Was there a particular reason why you have done it like that instead of defining cellwise-constant `fem.Function`s over the entire domain? I'm asking because in my own implementation of an UPML, I'm using `fem.Function`s.

For `eps` you are even defining a `fem.Function`, but you are not using it in the PML region(s).

python/demo/demo_pml/demo_pml.py
```diff
@@ -0,0 +1,606 @@
...
+# PML rectangles along x
+x_pml = ufl.as_vector((pml_coordinates(x[0], alpha, k0, l_dom, l_pml), x[1]))
+
+# PML rectangles along y
+y_pml = ufl.as_vector((x[0], pml_coordinates(x[1], alpha, k0, l_dom, l_pml)))
+# -
+
+# We can then express this coordinate systems as
+# a material transformation within the PML region. In other words,
+# the PML region can be interpreted as a material
```
thisch, line 324: This is known as a UPML (uniaxial PML). Maybe this could be mentioned here somewhere.

python/demo/demo_pml/demo_pml.py
```diff
@@ -0,0 +1,606 @@
...
+# A^{-1} \mathbf{A} {\boldsymbol{\varepsilon}_b}\mathbf{A}^{T},\\
+# & {\boldsymbol{\mu}_{pml}} =
+# A^{-1} \mathbf{A} {\boldsymbol{\mu}_b}\mathbf{A}^{T},
+# \end{align}
+# $$
+#
+# with $A^{-1}=\operatorname{det}(\mathbf{J})$.
+#
+# In DOLFINx, we use
+# `ufl.grad` to calculate the Jacobian of our
```
thisch, line 372: For UPML implementations `\varepsilon` and `\mu` are given by

image

(http://www.mit.edu/~wsshin/pdf/shin2012jcp.pdf)

I think you don't have to use `ufl.grad`, but it is nice to see that this is possible using the `ufl`.

python/demo/demo_pml/demo_pml.py
```diff
@@ -0,0 +1,606 @@
...
+
+
+# -
+
+# For convenience, we define the $\nabla\times$ operator for a 2D vector
+
+# +
+
+def curl_2d(a: fem.Function):
+    return ufl.as_vector((0, 0, a[1].dx(0) - a[0].dx(1)))
```
thisch, line 118: Can't you use `ufl.curl` instead of defining your own 2D version of `curl`?

thisch (commented):
Thx a lot for working on this! I would love to see a proper support for PMLs at some point in `dolfinx`.  `ngsolve` e.g. has a `SetPML` method, which is documented here https://docu.ngsolve.org/latest/i-tutorials/unit-1.7-helmholtz/pml.html.

I have my own 2D version of an UPML, which I implemented in a slightly different way and therefore have some questions and feedback.

mikics:
> Thx a lot for working on this! I would love to see a proper support for PMLs at some point in `dolfinx`. `ngsolve` e.g. has a `SetPML` method, which is documented here https://docu.ngsolve.org/latest/i-tutorials/unit-1.7-helmholtz/pml.html.
> 
> I have my own 2D version of an UPML, which I implemented in a slightly different way and therefore have some questions and feedback.

Thanks for giving precious comments on this PR! I will look at them right now.

python/demo/demo_pml/demo_pml.py
mikics: I think that using `ufl.curl` in this case is not possible since the problem is two-dimensional, and `ufl.curl` would try to evaluate the derivative of `Es` along the third dimension, resulting in an error. Therefore we need to define this version of the $\nabla \times$ operator.

python/demo/demo_pml/demo_pml.py
mikics: Did not know about this definition of uniaxial PML, thank you!

I am not familiar with the difference between uniaxial PML and stretched-coordinate PML. The calculation reported in the demo are based [on this work
](http://people.ee.duke.edu/~cummer/reprints/138_Popa11_PRA_ComplexTransformationOptics.pdf), where the authors propose a complex coordinate transformation to implement PML for irregularly shaped domain.

In this demo the domain has a regular shape (a square), and therefore what you propose makes sense. However, with `ufl.grad` we are not constrained to any particular shape, and therefore I would say that this approach is "more general". For instance, in [this other PR](https://github.com/FEniCS/dolfinx/blob/a543704a60f7c6869456e34e423b414555ee19a8/python/demo/demo_axis/demo_axis.py#L304-L322) a spherical PML is expressed with the same formulation, and, in principle, it could be used for more complicated shapes as well.

python/demo/demo_pml/demo_pml.py
mikics: The main reason is that we wanted to explicitly show that we have three different transformations in the six different regions of the PML.

Besides, I think that using `fem.Function` would make the computation a bit slower, since you need to interpolate the function over a certain function space (but I might be wrong).

python/demo/demo_pml/demo_pml.py
```diff
@@ -0,0 +1,606 @@
...
+# coordinate transformation for the different PML regions,
+# and then we can implement this Jacobian for
+# calculating $\boldsymbol{\varepsilon}_{pml}$
+# and $\boldsymbol{\mu}_{pml}$. The here below function
+# named `create_eps_mu()` serves this purpose:
+
+# +
+
+
+def create_eps_mu(
```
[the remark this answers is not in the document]
mikics: Do you mean something like this?

```python
def create_eps_mu(
        domain,
        pml_x: bool,
        pml_y: bool,
        eps_bkg: Union[float, ufl.tensors.ListTensor],
        mu_bkg: Union[float, ufl.tensors.ListTensor],
        k0: float,
        l_dom: float,
        l_pml: float,
        alpha: float = 1) -> Tuple[ufl.tensors.ComponentTensor,
                                   ufl.tensors.ComponentTensor]:

    x = ufl.SpatialCoordinate(domain)

    if pml_x:
        x_pml = (x[0] + 1j * alpha / k0 * x
                 * (ufl.algebra.Abs(x[0]) - l_dom / 2)
                 / (l_pml / 2 - l_dom / 2)**2)
    else:
        x_pml = x[0]

    if pml_y:
        y_pml = (x[1] + 1j * alpha / k0 * x
                 * (ufl.algebra.Abs(x[1]) - l_dom / 2)
                 / (l_pml / 2 - l_dom / 2)**2)
    else:
        y_pml = x[1]

    pml_coordinates = ufl.as_vector((x_pml, y_pml))

    J = ufl.grad(pml_coordinates)

    J = ufl.as_matrix(((J[0, 0], 0, 0),
                       (0, J[1, 1], 0),
                       (0, 0, 1)))

    A = ufl.inv(J)
    eps_pml = ufl.det(J) * A * eps_bkg * ufl.transpose(A)
    mu_pml = ufl.det(J) * A * mu_bkg * ufl.transpose(A)

    return eps_pml, mu_pml
```

python/demo/demo_pml/demo_pml.py
thisch: At least for me it works in a 2D problem.

python/demo/demo_pml/demo_pml.py
thisch: > Besides, I think that using fem.Function would make the computation a bit slower, since you need to interpolate the function over a certain function space (but I might be wrong).

I agree with you that it would probably be slower, but I haven't verified if using `fem.Function` has a measurable impact. In my application the user has to provide a permittivity function and a permeability function for the entire domain (including the PML). The solver in my application is not aware that there is a PML in the geometry (In 1D the outgoing boundary condition can be imposed at the boundary of your geometry, i.e., a PML is not needed.). I think both approaches (yours and mine) have their pros and cons.

python/demo/demo_pml/demo_pml.py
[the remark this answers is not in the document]
thisch: Yes exactly! I think the variable names (pml_x and pml_y) could be more descriptive, e.g., `x_coord_in_pml`  and `y_coord_in_pml`.

python/demo/demo_pml/demo_pml.py
mikics: To be clearer, if I use this weak form:

```python
F = - ufl.inner(ufl.curl(Es_3d), ufl.curl(v_3d)) * dDom \
    + eps * (k0**2) * ufl.inner(Es, v) * dDom \
    + (k0**2) * (eps - eps_bkg) * ufl.inner(Eb, v) * dDom \
    - ufl.inner(ufl.inv(mu_x) * ufl.curl(Es_3d), ufl.curl(v_3d)) * dPml_x \
    - ufl.inner(ufl.inv(mu_y) * ufl.curl(Es_3d), ufl.curl(v_3d)) * dPml_y \
    - ufl.inner(ufl.inv(mu_xy) * ufl.curl(Es_3d), ufl.curl(v_3d)) * dPml_xy \
    + (k0**2) * ufl.inner(eps_x * Es_3d, v_3d) * dPml_x \
    + (k0**2) * ufl.inner(eps_y * Es_3d, v_3d) * dPml_y \
    + (k0**2) * ufl.inner(eps_xy * Es_3d, v_3d) * dPml_xy
```

I get this error:

```bash
  File "/root/dolfinx/python/demo/demo_pml/demo_pml.py", line 444, in <module>
    a, L = ufl.lhs(F), ufl.rhs(F)
  File "/root/ufl/ufl/formoperators.py", line 69, in lhs
    form = expand_derivatives(form)
  File "/root/ufl/ufl/algorithms/ad.py", line 25, in expand_derivatives
    form = apply_algebra_lowering(form)
  File "/root/ufl/ufl/algorithms/apply_algebra_lowering.py", line 175, in apply_algebra_lowering
    return map_integrand_dags(LowerCompoundAlgebra(), expr)
  File "/root/ufl/ufl/algorithms/map_integrands.py", line 46, in map_integrand_dags
    return map_integrands(lambda expr: map_expr_dag(function, expr, compress),
  File "/root/ufl/ufl/algorithms/map_integrands.py", line 27, in map_integrands
    mapped_integrals = [map_integrands(function, itg, only_integral_type)
  File "/root/ufl/ufl/algorithms/map_integrands.py", line 27, in <listcomp>
    mapped_integrals = [map_integrands(function, itg, only_integral_type)
  File "/root/ufl/ufl/algorithms/map_integrands.py", line 35, in map_integrands
    return itg.reconstruct(function(itg.integrand()))
  File "/root/ufl/ufl/algorithms/map_integrands.py", line 46, in <lambda>
    return map_integrands(lambda expr: map_expr_dag(function, expr, compress),
  File "/root/ufl/ufl/corealg/map_dag.py", line 36, in map_expr_dag
    result, = map_expr_dags(function, [expression], compress=compress,
  File "/root/ufl/ufl/corealg/map_dag.py", line 99, in map_expr_dags
    r = handlers[v._ufl_typecode_](v, *[vcache[u] for u in v.ufl_operands])
  File "/root/ufl/ufl/algorithms/apply_algebra_lowering.py", line 168, in curl
    return as_vector((c(1, 2), c(2, 0), c(0, 1)))
  File "/root/ufl/ufl/algorithms/apply_algebra_lowering.py", line 161, in c
    return a[j].dx(i) - a[i].dx(j)
  File "/root/ufl/ufl/exproperators.py", line 498, in _dx
    d = Grad(d)
  File "/root/ufl/ufl/differentiation.py", line 140, in __new__
    dim = find_geometric_dimension(f)
  File "/root/ufl/ufl/domain.py", line 376, in find_geometric_dimension
    error("Cannot determine geometric dimension from expression.")
  File "/root/ufl/ufl/log.py", line 135, in error
    raise self._exception_type(self._format_raw(*message))
ufl.log.UFLException: Cannot determine geometric dimension from expression.
```
Maybe I just need to tweak something in the code, but I don't know what. If you can share your example with UPML maybe I can spot what I am missing.

Issue #2275, opened 2022-07-20
Discrete interpolation wrong for non-affine elements (using Jacobians in transformations)


jorgensd:
Inside the discrete interpolation operator, the derivatives (of cmap basis) at the first quadrature points is used to compute jacobians https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/discreteoperators.h#L200-L203

This should be changed in:
https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/discreteoperators.h#L274
to use the q-th entry of https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/discreteoperators.h#L200-L201

jorgensd:
Aldo affects non-matching map interpolation

PR #2278, opened 2022-07-21
Remove xtensor code from element code

garth-wells:
Simplifies code, uncovered some bugs, and is faster.

Some later optimisation can be performed now that code is clearer.

### DIFF
--- a/cpp/dolfinx/fem/CoordinateElement.cpp
+++ b/cpp/dolfinx/fem/CoordinateElement.cpp
@@ -59,9 +59,9 @@ ElementDofLayout CoordinateElement::create_dof_layout() const
                           _element->entity_closure_dofs(), {}, {});
 }
 //-----------------------------------------------------------------------------
-void CoordinateElement::pull_back_nonaffine_new(mdspan2_t X, cmdspan2_t x,
-                                                cmdspan2_t cell_geometry,
-                                                double tol, int maxit) const
+void CoordinateElement::pull_back_nonaffine(mdspan2_t X, cmdspan2_t x,
+                                            cmdspan2_t cell_geometry,
+                                            double tol, int maxit) const
 {
   // Number of points
   std::size_t num_points = x.extent(0);
--- a/cpp/dolfinx/fem/CoordinateElement.cpp
+++ b/cpp/dolfinx/fem/CoordinateElement.cpp
@@ -120,8 +120,8 @@ void CoordinateElement::pull_back_nonaffine_new(mdspan2_t X, cmdspan2_t x,
         for (std::size_t j = 0; j < basis.extent(2); ++j)
           dphi(i, j) = basis(i + 1, 0, j, 0);
 
-      compute_jacobian_new(dphi, cell_geometry, J);
-      compute_jacobian_inverse_new(J, K);
+      compute_jacobian(dphi, cell_geometry, J);
+      compute_jacobian_inverse(J, K);
 
       // Compute dX = K * (x_p - x_k)
       std::fill(dX.begin(), dX.end(), 0);
--- a/cpp/dolfinx/fem/CoordinateElement.cpp
+++ b/cpp/dolfinx/fem/CoordinateElement.cpp
@@ -74,32 +74,6 @@ void CoordinateElement::push_forward(
   math::dot(phi, cell_geometry, x);
 }
 //-----------------------------------------------------------------------------
-std::array<double, 3>
-CoordinateElement::x0(const xt::xtensor<double, 2>& cell_geometry)
-{
-  std::array<double, 3> x0 = {0, 0, 0};
-  for (std::size_t i = 0; i < cell_geometry.shape(1); ++i)
-    x0[i] += cell_geometry(0, i);
-  return x0;
-}
-//-----------------------------------------------------------------------------
-void CoordinateElement::pull_back_affine(xt::xtensor<double, 2>& X,
-                                         const xt::xtensor<double, 2>& K,
-                                         const std::array<double, 3>& x0,
-                                         const xt::xtensor<double, 2>& x)
-{
-  assert(X.shape(0) == x.shape(0));
-  assert(X.shape(1) == K.shape(0));
-  assert(x.shape(1) == K.shape(1));
-
-  // Calculate X for each point
-  X.fill(0.0);
-  for (std::size_t p = 0; p < x.shape(0); ++p)
-    for (std::size_t i = 0; i < K.shape(0); ++i)
-      for (std::size_t j = 0; j < K.shape(1); ++j)
-        X(p, i) += K(i, j) * (x(p, j) - x0[j]);
-}
-//-----------------------------------------------------------------------------
 void CoordinateElement::pull_back_nonaffine_new(mdspan2_t X, cmdspan2_t x,
                                                 cmdspan2_t cell_geometry,
                                                 double tol, int maxit) const
@@ -194,80 +168,6 @@ void CoordinateElement::pull_back_nonaffine_new(mdspan2_t X, cmdspan2_t x,
 
     std::copy(Xk_b.cbegin(), std::next(Xk_b.cbegin(), tdim),
               X.data_handle() + p * tdim);
-    if (k == maxit)
-    {
-      throw std::runtime_error(
-          "Newton method failed to converge for non-affine geometry");
-    }
-  }
-}
-//-----------------------------------------------------------------------------
-void CoordinateElement::pull_back_nonaffine(
-    xt::xtensor<double, 2>& X, const xt::xtensor<double, 2>& x,
-    const xt::xtensor<double, 2>& cell_geometry, double tol, int maxit) const
-{
-  // Number of points
-  std::size_t num_points = x.shape(0);
-  if (num_points == 0)
-    return;
-
-  const std::size_t tdim = mesh::cell_dim(this->cell_shape());
-  const std::size_t gdim = x.shape(1);
-  const std::size_t num_xnodes = cell_geometry.shape(0);
-  assert(cell_geometry.shape(1) == gdim);
-  assert(X.shape(0) == num_points);
-  assert(X.shape(1) == tdim);
-
-  xt::xtensor<double, 2> dphi({tdim, num_xnodes});
-  xt::xtensor<double, 2> Xk({1, tdim});
-  std::array<double, 3> xk = {0, 0, 0};
-  xt::xtensor<double, 1> dX = xt::empty<double>({tdim});
-  xt::xtensor<double, 2> J({gdim, tdim});
-  xt::xtensor<double, 2> K({tdim, gdim});
-  xt::xtensor<double, 4> basis(_element->tabulate_shape(1, 1));
-  for (std::size_t p = 0; p < num_points; ++p)
-  {
-    std::fill(Xk.begin(), Xk.end(), 0.0);
-    int k;
-    for (k = 0; k < maxit; ++k)
-    {
-      _element->tabulate(1, std::span(Xk.data(), Xk.size()), {1, tdim}, basis);
-
-      // x = cell_geometry * phi
-      auto phi = xt::view(basis, 0, 0, xt::all(), 0);
-      std::fill(xk.begin(), xk.end(), 0.0);
-      for (std::size_t i = 0; i < cell_geometry.shape(0); ++i)
-        for (std::size_t j = 0; j < cell_geometry.shape(1); ++j)
-          xk[j] += cell_geometry(i, j) * phi[i];
-
-      // Compute Jacobian, its inverse and determinant
-      std::fill(J.begin(), J.end(), 0.0);
-      dphi = xt::view(basis, xt::range(1, tdim + 1), 0, xt::all(), 0);
-      compute_jacobian(dphi, cell_geometry, J);
-      compute_jacobian_inverse(J, K);
-
-      // Compute dX = K * (x_p - x_k)
-      std::fill(dX.begin(), dX.end(), 0);
-      auto x_p = xt::row(x, p);
-      for (std::size_t i = 0; i < K.shape(0); ++i)
-        for (std::size_t j = 0; j < K.shape(1); ++j)
-          dX[i] += K(i, j) * (x_p[j] - xk[j]);
-
-      // Compute Xk += dX
-      std::transform(dX.cbegin(), dX.cend(), Xk.cbegin(), Xk.begin(),
-                     [](double a, double b) { return a + b; });
-
-      // Compute norm(dX)
-      if (auto dX_squared
-          = std::transform_reduce(dX.cbegin(), dX.cend(), 0.0, std::plus{},
-                                  [](auto v) { return v * v; });
-          std::sqrt(dX_squared) < tol)
-      {
-        break;
-      }
-    }
-    std::copy(Xk.cbegin(), std::next(Xk.cbegin(), tdim),
-              std::next(X.begin(), p * tdim));
     if (k == maxit)
     {
       throw std::runtime_error(
--- a/cpp/dolfinx/fem/CoordinateElement.cpp
+++ b/cpp/dolfinx/fem/CoordinateElement.cpp
@@ -9,9 +9,6 @@
 #include <cmath>
 #include <dolfinx/common/math.h>
 #include <dolfinx/mesh/cell_types.h>
-#include <xtensor/xadapt.hpp>
-#include <xtensor/xnoalias.hpp>
-#include <xtensor/xview.hpp>
 
 using namespace dolfinx;
 using namespace dolfinx::fem;
@@ -78,31 +75,25 @@ void CoordinateElement::pull_back_nonaffine_new(mdspan2_t X, cmdspan2_t x,
   assert(X.extent(0) == num_points);
   assert(X.extent(1) == tdim);
 
-  // xt::xtensor<double, 2> dphi({tdim, num_xnodes});
   std::vector<double> dphi_b(tdim * num_xnodes);
   mdspan2_t dphi(dphi_b.data(), tdim, num_xnodes);
 
-  // xt::xtensor<double, 2> Xk({1, tdim});
   std::vector<double> Xk_b(tdim);
   mdspan2_t Xk(Xk_b.data(), 1, tdim);
 
   std::array<double, 3> xk = {0, 0, 0};
 
-  // xt::xtensor<double, 1> dX = xt::empty<double>({tdim});
   std::vector<double> dX(tdim);
 
-  // xt::xtensor<double, 2> J({gdim, tdim});
   std::vector<double> J_b(gdim * tdim);
   mdspan2_t J(J_b.data(), gdim, tdim);
 
-  // xt::xtensor<double, 2> K({tdim, gdim});
   std::vector<double> K_b(tdim * gdim);
   mdspan2_t K(K_b.data(), tdim, gdim);
 
   namespace stdex = std::experimental;
   using mdspan4_t = stdex::mdspan<double, stdex::dextents<std::size_t, 4>>;
 
-  // xt::xtensor<double, 4> basis(_element->tabulate_shape(1, 1));
   const std::array<std::size_t, 4> bsize = _element->tabulate_shape(1, 1);
   std::vector<double> basis_b(
       std::reduce(bsize.begin(), bsize.end(), 1, std::multiplies{}));
@@ -118,7 +109,6 @@ void CoordinateElement::pull_back_nonaffine_new(mdspan2_t X, cmdspan2_t x,
       _element->tabulate(1, Xk_b, {1, tdim}, basis_b);
 
       // x = cell_geometry * phi
-      // auto phi = xt::view(basis, 0, 0, xt::all(), 0);
       std::fill(xk.begin(), xk.end(), 0.0);
       for (std::size_t i = 0; i < cell_geometry.extent(0); ++i)
         for (std::size_t j = 0; j < cell_geometry.extent(1); ++j)
@@ -126,7 +116,6 @@ void CoordinateElement::pull_back_nonaffine_new(mdspan2_t X, cmdspan2_t x,
 
       // Compute Jacobian, its inverse and determinant
       std::fill(J_b.begin(), J_b.end(), 0.0);
-      // dphi = xt::view(basis, xt::range(1, tdim + 1), 0, xt::all(), 0);
       for (std::size_t i = 0; i < tdim; ++i)
         for (std::size_t j = 0; j < basis.extent(2); ++j)
           dphi(i, j) = basis(i + 1, 0, j, 0);
--- a/cpp/dolfinx/fem/CoordinateElement.cpp
+++ b/cpp/dolfinx/fem/CoordinateElement.cpp
@@ -47,12 +47,12 @@ CoordinateElement::tabulate_shape(std::size_t nd, std::size_t num_points) const
   return _element->tabulate_shape(nd, num_points);
 }
 //-----------------------------------------------------------------------------
-void CoordinateElement::tabulate(int n, const xt::xtensor<double, 2>& X,
-                                 xt::xtensor<double, 4>& basis) const
+void CoordinateElement::tabulate(int nd, std::span<const double> X,
+                                 std::array<std::size_t, 2> shape,
+                                 std::span<double> basis) const
 {
   assert(_element);
-  _element->tabulate(n, std::span(X.data(), X.size()),
-                     std::array{X.shape(0), X.shape(1)}, basis);
+  _element->tabulate(nd, X, shape, basis);
 }
 //--------------------------------------------------------------------------------
 ElementDofLayout CoordinateElement::create_dof_layout() const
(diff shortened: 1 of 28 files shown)


cpp/dolfinx/fem/Function.h
```diff
@@ -291,32 +304,37 @@ class Function
...
-  /// @param[in,out] u The values at the points. Values are not computed
+  /// @param[out] u The values at the points. Values are not computed
   /// for points with a negative cell index. This argument must be
-  /// passed with the correct size.
-  void eval(const xt::xtensor<double, 2>& x,
-            const std::span<const std::int32_t>& cells,
-            xt::xtensor<T, 2>& u) const
+  /// passed with the correct size. Storage is row-major.
+  /// @param[in] ushape The shape of `u`.
+  void eval(std::span<const double> x, std::array<std::size_t, 2> xshape,
```
michalhabera, line 319: Could this be `mdspan` here? This C-like interface is too verbose.

garth-wells: > Could this be `mdspan` here? This C-like interface is too verbose.

I think, at least for now, it's better to avoid exposing `mdspan` in the public interface where possible. Also, in practice the 'c-like' interface is less verbose because we often have:
```cpp
auto [data, shape] = foo();
bar(data, shape);
```
`mdspan` doesn't help here because it can't be returned by `foo` because `mdspan` doesn't own the data. Similarly the 'c-like' interface is convenient in the pybind layer. In practice, it's rare to see any of this in user code anyway.

michalhabera: I agree that for python users this makes no difference, as all interface handling happens in our pybind11 cpp code. For the same reason I think that `mdspan` interface is as convenient as c-like interface, as you only need to construct and pass the mdspan in pybind11 cpp wrappers.

There are many places where
```cpp
std::span<...> x = mesh->geometry().x();
foo(x, {mesh->geometry().x().size() / 3, 3});
```
is really verbose and error prone. I thought that in these places where non-owning data object is returned we'd use `mdspan` instead of the above repeated access to the data to figure out its shape.

garth-wells: > I agree that for python users this makes no difference, as all interface handling happens in our pybind11 cpp code. For the same reason I think that `mdspan` interface is as convenient as c-like interface, as you only need to construct and pass the mdspan in pybind11 cpp wrappers.
> 
> There are many places where
> 
> ```c++
> std::span<...> x = mesh->geometry().x();
> foo(x, {mesh->geometry().x().size() / 3, 3});
> ```
> 
> is really verbose and error prone. I thought that in these places where non-owning data object is returned we'd use `mdspan` instead of the above repeated access to the data to figure out its shape.

This is unrelated to the current PR - methods that return a `span` have not been touched.  `xt::adapt` has just been replaced by `mdspan` for wrapping the data. It could be discussed as a future change. Using `mdspan` in these cases would be neater, at the expense of changing an STL type to a non-standard type. `mdspan` does bring some complication over `std::span` - memory is not guaranteed to be contiguous (strides are supported), storage layouts can differ, etc, so `foo.data_handle()` needs to be used with care.

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -72,97 +59,78 @@ ElementDofLayout CoordinateElement::create_dof_layout() const
...
-  for (std::size_t p = 0; p < x.shape(0); ++p)
-    for (std::size_t i = 0; i < K.shape(0); ++i)
-      for (std::size_t j = 0; j < K.shape(1); ++j)
-        X(p, i) += K(i, j) * (x(p, j) - x0[j]);
-}
-//-----------------------------------------------------------------------------
-void CoordinateElement::pull_back_nonaffine(
-    xt::xtensor<double, 2>& X, const xt::xtensor<double, 2>& x,
-    const xt::xtensor<double, 2>& cell_geometry, double tol, int maxit) const
+void CoordinateElement::pull_back_nonaffine_new(mdspan2_t X, cmdspan2_t x,
```
jorgensd, line 62:
```suggestion
void CoordinateElement::pull_back_nonaffine(mdspan2_t X, cmdspan2_t x,
```

cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -170,26 +177,47 @@ class CoordinateElement
...
   /// @param [in] maxit Maximum number of Newton iterations
-  /// @note If convergence is not achieved within maxit, the function throws a
-  /// run-time error.
-  void pull_back_nonaffine(xt::xtensor<double, 2>& X,
-                           const xt::xtensor<double, 2>& x,
-                           const xt::xtensor<double, 2>& cell_geometry,
-                           double tol = 1.0e-8, int maxit = 10) const;
+  /// @note If convergence is not achieved within maxit, the function
+  /// throws a runtime error.
+  void pull_back_nonaffine_new(mdspan2_t X, cmdspan2_t x,
```
jorgensd, line 218:
```suggestion
  void pull_back_nonaffine(mdspan2_t X, cmdspan2_t x,
```

cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -148,17 +156,16 @@ class CoordinateElement
   /// @param[in,out] x The physical coordinates of the reference points X
   /// @param[in] cell_geometry The cell node coordinates (physical)
   /// @param[in] phi Tabulated basis functions at reference points X
```
jorgensd, line 158: Do we want to add a comment here about the relation between dimensions of the different inputs?
i.e.
```
phi.extent(1) == cell_geometry.extent(0)
phi.extent(0) == x.extent(0)
x.extent(1) == cell_geometry.extent(1)
```
```

File: cpp/dolfinx/fem/CoordinateElement.cpp [hunk 5]
```diff
@@ -72,97 +59,78 @@ ElementDofLayout CoordinateElement::create_dof_layout() const
...
+  std::vector<double> dX(tdim);
+
+  std::vector<double> J_b(gdim * tdim);
+  mdspan2_t J(J_b.data(), gdim, tdim);
+
+  std::vector<double> K_b(tdim * gdim);
+  mdspan2_t K(K_b.data(), tdim, gdim);
+
+  namespace stdex = std::experimental;
+  using mdspan4_t = stdex::mdspan<double, stdex::dextents<std::size_t, 4>>;
```
jorgensd, line 95: Instead of redeclaring these span types all over the place, could we use those in 
https://github.com/FEniCS/basix/blob/3844e0f52e34106a9f23a2298a00224d816466e5/cpp/basix/finite-element.h#L25-L48
for all double spans?

cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -72,97 +59,78 @@ ElementDofLayout CoordinateElement::create_dof_layout() const
...
 
       // x = cell_geometry * phi
-      auto phi = xt::view(basis, 0, 0, xt::all(), 0);
       std::fill(xk.begin(), xk.end(), 0.0);
-      for (std::size_t i = 0; i < cell_geometry.shape(0); ++i)
-        for (std::size_t j = 0; j < cell_geometry.shape(1); ++j)
-          xk[j] += cell_geometry(i, j) * phi[i];
+      for (std::size_t i = 0; i < cell_geometry.extent(0); ++i)
+        for (std::size_t j = 0; j < cell_geometry.extent(1); ++j)
+          xk[j] += cell_geometry(i, j) * basis(0, 0, i, 0);
```
jorgensd, line 115: Could use mdspan and math::dot for this operation, but i guess declaring 2 new spans is as verbose as this.

cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -148,17 +156,16 @@ class CoordinateElement
...
-  static void push_forward(xt::xtensor<double, 2>& x,
-                           const xt::xtensor<double, 2>& cell_geometry,
-                           const xt::xtensor<double, 2>& phi);
+  template <typename U, typename V, typename W>
+  static void push_forward(U&& x, const V& cell_geometry, const W& phi)
+  {
+    for (std::size_t i = 0; i < x.extent(0); ++i)
+      for (std::size_t j = 0; j < x.extent(1); ++j)
+        x(i, j) = 0;
```
jorgensd, line 164: Do we need to do the double loop here? 
Doesn't mdspan work with 
```c++
std::fill(x.begin(), x.end(), 0);
```
?

There was a similar instance of this in `math.h`.

cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -170,26 +177,47 @@ class CoordinateElement
...
+  template <typename U, typename V, typename W>
+  static void pull_back_affine(U&& X, const V& K,
+                               const std::array<double, 3>& x0, const W& x)
+  {
+    assert(X.extent(0) == x.extent(0));
+    assert(X.extent(1) == K.extent(0));
+    assert(x.extent(1) == K.extent(1));
+    for (std::size_t i = 0; i < X.extent(0); ++i)
+      for (std::size_t j = 0; j < X.extent(1); ++j)
+        X(i, j) = 0;
```
jorgensd, line 189: Same comment as above.

cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -170,26 +177,47 @@ class CoordinateElement
...
+          X(p, i) += K(i, j) * (x(p, j) - x0[j]);
+  }
+
+  /// mdspan typedef
+  using mdspan2_t
+      = std::experimental::mdspan<double,
+                                  std::experimental::dextents<std::size_t, 2>>;
+  /// mdspan typedef
+  using cmdspan2_t
+      = std::experimental::mdspan<const double,
```
jorgensd, line 204: Same question regarding using these from `basix::impl`.

cpp/dolfinx/fem/Expression.h
```diff
@@ -238,7 +240,10 @@ class Expression
 
   /// @brief Evaluation points on the reference cell
   /// @return Evaluation points
-  const xt::xtensor<double, 2>& X() const { return _x_ref; }
+  std::pair<std::vector<double>, std::array<std::size_t, 2>> X() const
```
jorgensd, line 243: Maybe add a comment saying that the evaluation points is returned as a flattened and the shape is returned as well. Another question is do we want to return a reference to the points (an std::span) instead of a copy?

cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -427,7 +423,8 @@ bool FiniteElement::interpolation_ident() const noexcept
   return _element->interpolation_is_identity();
 }
 //-----------------------------------------------------------------------------
-xt::xtensor<double, 2> FiniteElement::interpolation_points() const
+std::pair<std::vector<double>, std::array<std::size_t, 2>>
```
jorgensd, line 426: Same question as above, should this rather be a pair of std::span and an std::array, to avoid copying?

cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -436,23 +433,22 @@ xt::xtensor<double, 2> FiniteElement::interpolation_points() const
         "this is a mixed element?");
   }
 
-  auto& [x, shape] = _element->points();
-  return xt::adapt(x, std::vector<std::size_t>{shape[0], shape[1]});
+  return _element->points();
 }
 //-----------------------------------------------------------------------------
-xt::xtensor<double, 2> FiniteElement::interpolation_operator() const
+std::pair<std::vector<double>, std::array<std::size_t, 2>>
```
jorgensd, line 439: Same as above, `std::pair<std::span<double>, std::array<std::size_t, 2>>`?

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -103,31 +104,34 @@ class FiniteElement
...
-  /// @param[in] X Two dimensional xtensor of shape [num_points,
-  /// geometric dimension] containing the points at the reference
-  /// element
+  /// @brief Evaluate derivatives of the basis functions up to given order
+  /// at points in the reference cell.
+  /// @param[in,out] values Array that will be filled with the tabulated
+  /// basis values. Must have shape `(num_derivatives, num_points,
+  /// num_dofs, reference_value_size)` (row-major storage)
+  /// @param[in] X The reference coordinates at which to evaluate the
+  /// basis functions. Shape is `(num_points, geometric dimension)`
```
jorgensd, line 113: Shouldn't this actually be the topological dimension of the underlying cell? As the geometric dimension of a triangular manifold is 3, while these are reference coordinates, which should be dim 2.

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -103,31 +104,34 @@ class FiniteElement
...
+  void tabulate(std::span<double> values, std::span<const double> X,
+                std::array<std::size_t, 2> shape, int order) const;
 
   /// Evaluate all derivatives of the basis functions up to given order
   /// at given points in reference cell
-  /// @param[in] X Two dimensional xtensor of shape [num_points,
-  /// geometric dimension] containing the points at the reference
-  /// element
+  /// @param[in] X The reference coordinates at which to evaluate the
+  /// basis functions. Shape is `(num_points, geometric dimension)`
```
jorgensd, line 124: Same comment as above

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -170,7 +174,8 @@ class FiniteElement
   /// nodal positions. For other elements the points will typically be
   /// the quadrature points used to evaluate moment degrees of freedom.
   /// @return Points on the reference cell. Shape is (num_points, tdim).
```
jorgensd, line 176: I guess `@return` docstring should be slightly modified.
Same remark regarding vector vs span in returntype

cpp/dolfinx/fem/Function.h
```diff
@@ -373,37 +398,53 @@ class Function
...
-    xt::xtensor<double, 2> coordinate_dofs
-        = xt::zeros<double>({num_dofs_g, gdim});
-    xt::xtensor<double, 2> xp = xt::zeros<double>({std::size_t(1), gdim});
+    namespace stdex = std::experimental;
+    using cmdspan4_t
+        = stdex::mdspan<const double, stdex::dextents<std::size_t, 4>>;
+    using mdspan2_t = stdex::mdspan<double, stdex::dextents<std::size_t, 2>>;
+    using mdspan3_t = stdex::mdspan<double, stdex::dextents<std::size_t, 3>>;
+
+    std::vector<double> coord_dofs_b(num_dofs_g * gdim);
```
jorgensd, line 407: General question, what does the `*_b` stand for when we create memory allocations for the spans?

cpp/dolfinx/fem/CoordinateElement.h
jorgensd: Nevermind, as we can't guarantee that the `mdspan` is contiguous in memory that doesn't work.

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -52,6 +48,13 @@ template <typename T, typename U>
 void discrete_gradient(const FunctionSpace& V0, const FunctionSpace& V1,
                        U&& mat_set)
 {
+  namespace stdex = std::experimental;
+  using cmdspan2_t
+      = stdex::mdspan<const double, stdex::dextents<std::size_t, 2>>;
+  using mdspan2_t = stdex::mdspan<double, stdex::dextents<std::size_t, 2>>;
+  using cmdspan4_t
+      = stdex::mdspan<const double, stdex::dextents<std::size_t, 4>>;
+
```
jorgensd, line 57: Same comment regarding using basix::impl

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -74,22 +77,21 @@ void discrete_gradient(const FunctionSpace& V0, const FunctionSpace& V1,
...
+  const auto [X, Xshape] = e1->interpolation_points();
 
   // Tabulate first order derivatives of Lagrange space at H(curl)
   // interpolation points
   const int ndofs0 = e0->space_dimension();
   const int tdim = mesh->topology().dim();
-  xt::xtensor<double, 4> phi0
-      = xt::empty<double>({tdim + 1, int(X.shape(0)), ndofs0, 1});
-  e0->tabulate(phi0, X, 1);
+  std::vector<double> phi0_b((tdim + 1) * Xshape[0] * ndofs0 * 1);
```
jorgensd, line 86:
```suggestion
  std::vector<double> phi0_b((tdim + 1) * Xshape[0] * ndofs0 * 1, 0);
```

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -107,21 +109,23 @@ void discrete_gradient(const FunctionSpace& V0, const FunctionSpace& V1,
   assert(dofmap1);
 
   // Build the element interpolation matrix
-  std::vector<T> A(e1->space_dimension() * ndofs0);
+  std::vector<T> Ab(e1->space_dimension() * ndofs0);
```
jorgensd, line 112: I think you have used `*_b` on most of the storages, so we should change this if we want to be consistent.

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -195,149 +200,185 @@ void interpolation_matrix(const FunctionSpace& V0, const FunctionSpace& V1,
...
   std::span<const double> x_g = mesh->geometry().x();
 
+  namespace stdex = std::experimental;
+  using cmdspan2_t
+      = stdex::mdspan<const double, stdex::dextents<std::size_t, 2>>;
+  using cmdspan3_t
+      = stdex::mdspan<const double, stdex::dextents<std::size_t, 3>>;
+  using cmdspan4_t
+      = stdex::mdspan<const double, stdex::dextents<std::size_t, 4>>;
+  using mdspan3_t = stdex::mdspan<double, stdex::dextents<std::size_t, 3>>;
```
jorgensd, line 210: `basix::impl`?

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -195,149 +200,185 @@ void interpolation_matrix(const FunctionSpace& V0, const FunctionSpace& V1,
...
+  std::vector<double> mapped_values_b(Xshape[0] * bs0 * dim0
+                                      * e1->value_size());
+  mdspan3_t mapped_values(mapped_values_b.data(), Xshape[0], bs0 * dim0,
+                          e1->value_size());
+
+  auto pull_back_fn1 = e1->basix_element().map_fn<u_t, U_t, K_t, J_t>();
+
+  std::vector<double> coordinate_dofs_b(num_dofs_g * 3);
+  stdex::mdspan<double, stdex::extents<std::size_t, stdex::dynamic_extent, 3>>
+      coordinate_dofs(coordinate_dofs_b.data(), num_dofs_g, 3);
```
jorgensd, line 276: Why do we want this to be static size 3, when we only ever use 
```c++
 auto coord_dofs = stdex::submdspan(coordinate_dofs, stdex::full_extent,
                                     std::pair(0, gdim))
```
?

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -195,149 +200,185 @@ void interpolation_matrix(const FunctionSpace& V0, const FunctionSpace& V1,
...
+      auto _J = stdex::submdspan(J, p, stdex::full_extent, stdex::full_extent);
+      cmap.compute_jacobian(dphi, coord_dofs, _J);
+      auto _K = stdex::submdspan(K, p, stdex::full_extent, stdex::full_extent);
+      cmap.compute_jacobian_inverse(_J, _K);
+      detJ[p] = cmap.compute_jacobian_determinant(_J, det_scratch);
     }
 
-    // Get evaluated basis on reference, apply DOF transformations, and
+    // Copy evaluated basis on reference, apply DOF transformations, and
     // push forward to physical element
```
jorgensd, line 313: Could we add a note here saying that we should not use `basis_reference0`, i.e. that this is a bug.

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -195,149 +200,185 @@ void interpolation_matrix(const FunctionSpace& V0, const FunctionSpace& V1,
...
-  cmap.tabulate(1, X, phi);
-  const xt::xtensor<double, 2> dphi
-      = xt::view(phi, xt::range(1, tdim + 1), 0, xt::all(), 0);
+  const auto [X, Xshape] = e1->interpolation_points();
+  std::array<std::size_t, 4> phi_shape = cmap.tabulate_shape(1, Xshape[0]);
+  std::vector<double> phi_b(
+      std::reduce(phi_shape.begin(), phi_shape.end(), 1, std::multiplies{}));
+  cmdspan4_t phi(phi_b.data(), phi_shape);
+  cmap.tabulate(1, X, Xshape, phi_b);
+  auto dphi
```
jorgensd, line 219: Could you add a comment here saying that this is a bug, and should be done at every quadrature point? 
This should now be a lot easier with subspans, i.e. the second dimension is `p` instead of 0

cpp/dolfinx/fem/interpolate.h
```diff
@@ -269,44 +300,44 @@ void interpolate_nonmatching_maps(Function<T>& u1, const Function<T>& u0,
...
-    for (std::size_t p = 0; p < X.shape(0); ++p)
+    std::fill(J_b.begin(), J_b.end(), 0);
+    for (std::size_t p = 0; p < Xshape[0]; ++p)
     {
-      auto _J = xt::view(J, p, xt::all(), xt::all());
-      cmap.compute_jacobian(dphi, coordinate_dofs, _J);
-      cmap.compute_jacobian_inverse(_J, xt::view(K, p, xt::all(), xt::all()));
-      detJ[p] = cmap.compute_jacobian_determinant(_J);
+      auto _J = stdex::submdspan(J, p, stdex::full_extent, stdex::full_extent);
+      cmap.compute_jacobian(dphi, coord_dofs, _J);
```
jorgensd, line 311: I think there is a similar bug here for non-affine elements, `dphi` refers to the derivatives at the first interpolation point, see line 240. This subspan should be moved into the loop. 

I guess we have too few tests on non-affine meshes.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -532,19 +569,33 @@ void interpolate(Function<T>& u, const xt::xarray<T>& f,
...
-    xt::xtensor<double, 3> dphi = xt::view(
-        cmap.tabulate(1, X), xt::range(1, tdim + 1), xt::all(), xt::all(), 0);
+    std::vector<double> J_b(Xshape[0] * gdim * tdim);
+    mdspan3_t J(J_b.data(), Xshape[0], gdim, tdim);
+    std::vector<double> K_b(Xshape[0] * tdim * gdim);
+    mdspan3_t K(K_b.data(), Xshape[0], tdim, gdim);
+    std::vector<double> detJ(Xshape[0]);
+    std::vector<double> det_scratch(2 * gdim * tdim);
+
+    std::vector<double> coord_dofs_b(num_dofs_g * 3);
```
jorgensd, line 579: Is there any point of assigning `3*num_dofs_g` instead of `gdim*num_dofs_g`, when we will never use the entry 
i>=gdim and results in strided memory access in line 625.

cpp/dolfinx/io/vtk_utils.cpp
```diff
@@ -79,33 +79,43 @@ tabulate_lagrange_dof_coordinates(const fem::FunctionSpace& V)
   const auto apply_dof_transformation
       = element->get_dof_transformation_function<double>();
 
+  namespace stdex = std::experimental;
+  using mdspan2_t = stdex::mdspan<double, stdex::dextents<std::size_t, 2>>;
+  using cmdspan4_t = stdex::mdspan<double, stdex::dextents<std::size_t, 4>>;
```
jorgensd, line 84: `basix::impl`?

cpp/dolfinx/io/vtk_utils.cpp
```diff
@@ -79,33 +79,43 @@ tabulate_lagrange_dof_coordinates(const fem::FunctionSpace& V)
...
   {
     // Extract cell geometry
     auto dofs_x = dofmap_x.links(c);
     for (std::size_t i = 0; i < dofs_x.size(); ++i)
-    {
-      std::copy_n(std::next(x_g.begin(), 3 * dofs_x[i]), gdim,
-                  std::next(coordinate_dofs.begin(), i * gdim));
-    }
+      for (std::size_t j = 0; j < gdim; ++j)
+        coordinate_dofs(i, j) = x_g[3 * dofs_x[i] + j];
```
jorgensd, line 113: As the `mdspan` in this case in contiguous (line 100-104) we could keep copy_n.

python/dolfinx/wrappers/fem.cpp
```diff
@@ -564,9 +554,9 @@ void declare_objects(py::module& m, const std::string& type)
           .def_property_readonly("X",
                                  [](const dolfinx::fem::Expression<T>& self)
                                  {
-                                   return py::array_t<double>(self.X().shape(),
-                                                              self.X().data(),
-                                                              py::cast(self));
+                                   auto [X, shape] = self.X();
```
jorgensd, line 557: if we use `property_readonly`, should we then move X, or should this just be a property?

python/dolfinx/wrappers/fem.cpp
```diff
@@ -1026,9 +1016,12 @@ void fem(py::module& m)
...
       .def_property_readonly("num_sub_elements",
                              &dolfinx::fem::FiniteElement::num_sub_elements)
-      .def_property_readonly(
-          "interpolation_points", [](const dolfinx::fem::FiniteElement& self)
-          { return xt_as_pyarray(self.interpolation_points()); })
+      .def_property_readonly("interpolation_points",
+                             [](const dolfinx::fem::FiniteElement& self)
+                             {
+                               auto [X, shape] = self.interpolation_points();
+                               return as_pyarray(std::move(X), shape);
```
jorgensd, line 1023: Same question as above.

cpp/dolfinx/fem/CoordinateElement.h
garth-wells: No - `mdspan` doesn't have `begin/end`. We could add an optimisation later that checks if `x` has contiguous storage, in which case we could use `std::fill`.

cpp/dolfinx/fem/Expression.h
garth-wells: I think a copy is better. These are the reference points, so it shouldn't be called inside a loop. Returning a span or reference puts constraints on how the data is stored internally, and since it's not performance critical we don't need to have this constraint.

cpp/dolfinx/fem/Function.h
garth-wells: 'buffer'

cpp/dolfinx/fem/CoordinateElement.cpp
garth-wells: The `impl` namespace implies private/internal. We shouldn't use `impl` from other libraries.

cpp/dolfinx/fem/interpolate.h
garth-wells: True, but since it's not a 'new' bug I'd rather fix it in a subsequent PR. It's much easier now to see what's happening to we should review all the relevant code. See also #2275.

cpp/dolfinx/fem/discreteoperators.h
garth-wells: Pattern is that `_b` is added only when the data is then wrapped using `mdspan`.

cpp/dolfinx/fem/discreteoperators.h
garth-wells: I don't see why zeroing is required (it's implicit anyway).

cpp/dolfinx/fem/interpolate.h
garth-wells: No point. Changed.

cpp/dolfinx/io/vtk_utils.cpp
garth-wells: I think it's clearer with an explicit loop.

python/dolfinx/wrappers/fem.cpp
garth-wells: Make into a function now since it returns a new copy.

jorgensd (approved):

PR #2282, opened 2022-07-22
X is a property, not a method

mscroggs:


jorgensd:
`X` is no longer a property, see:
https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/wrappers/fem.cpp#L545-L550
When going to the python layer, the data gets moved out of the C++ function, and is thus a copy, and is not modifiable by the user.

garth-wells:
As mentioned by @jorgensd, this is a conscious change. By copying the data we avoid tying together the function `X()` and the internal storage. Making `X` a property implies data sharing. There aren't performance implications since the `X` data is fixed for an element and the method can be called outside of any hot loops.

mscroggs:
I see. I've opened https://github.com/FEniCS/dolfinx/pull/2283 instead then, as the return type of X needs updating

PR #2287, opened 2022-07-24
Updates to demos to avoid segfaults

jorgensd:


.github/workflows/pyvista.yml
```diff
@@ -2,9 +2,9 @@ name: Test visualisation demos
 
 on:
   # Uncomment the below to trigger tests on push
-  # push:
-  #   branches:
-  #     - "**"
+  push:
+    branches:
+      - "**"
```
jorgensd, line 7:
```suggestion
  # push:
  #   branches:
  #     - "**"
```

garth-wells (commented):
I'm not crazy about adding more syntax to a (simple) demo. Is this just for running pyvista from Docker?

jorgensd:
> I'm not crazy about adding more syntax to a (simple) demo. Is this just for running pyvista from Docker?

These are so that we can run the codes through docker or any system that doesn’t have a frame buffer. We cannot test these otherwise. The code isn’t massively complicated by having a single if statement, and also instructs people on how to save figures.

PR #2288, opened 2022-07-24
Add ipygany and pythreejs to dolfinx/lab.

jorgensd:
The additional packages adds less than 26 MB to the Docker container
Added clearing of pip cache to further reduce image size.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -476,11 +476,18 @@ RUN apt-get -qq update && \
 
 # pyvista dependencies from pip. Only compatible with x86-64 (amd64).
 # matplotlib improves plotting quality with better color maps and properly rendering colorbars.
+# pythreejs and ipygany are the two backends of pyvista not requiring a framebuffer,
+# and is therefore preferred in notebooks
 RUN dpkgArch="$(dpkg --print-architecture)"; \
     case "$dpkgArch" in amd64) \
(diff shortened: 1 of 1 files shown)


jhale:
It's OK for me, but wait until the other Dockerfile changes have gone through before merging.

docker/Dockerfile
```diff
@@ -482,11 +482,18 @@ RUN apt-get -qq update && \
 # available on pypi for linux/arm64.
 # matplotlib improves plotting quality with better color maps and
 # properly rendering colorbars.
+# pythreejs and ipygany are the two backends of pyvista not requiring a framebuffer,
+# and is therefore preferred in notebooks
```
garth-wells, line 486:
```suggestion
# and are therefore preferred in notebooks
```

docker/Dockerfile
```diff
@@ -482,11 +482,18 @@ RUN apt-get -qq update && \
 # available on pypi for linux/arm64.
 # matplotlib improves plotting quality with better color maps and
 # properly rendering colorbars.
+# pythreejs and ipygany are the two backends of pyvista not requiring a framebuffer,
```
garth-wells, line 485:
```suggestion
# pythreejs and ipygany are two backends of pyvista not requiring a framebuffer,
```

garth-wells (changes requested):

PR #2292, opened 2022-07-26
Remove latest Docker tag, prefer nightly or stable

jhale:
This PR removes the use of the latest tag. The latest tag has no clear meaning. Instead, images will either be tagged nightly, stable or specifically versioned e.g. v0.4.1. All public facing documents will prioritise the use 
of the explicit stable tag. There will be further updates to be made in some of the other FEniCS components CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,7 +93,7 @@ unit-tests-python-mpi: &unit-tests-python-mpi
 jobs:
   build-real:
     docker:
-      - image: fenicsproject/test-env:latest-mpich
+      - image: fenicsproject/test-env:nightly-mpich
     environment:
       DEBIAN_FRONTEND: "noninteractive"
       PETSC_ARCH: "linux-gnu-real-32"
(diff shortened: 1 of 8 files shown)


jhale:
I will probably have to do this commit as two stages to stop the CI breaking overnight.

garth-wells (approved):

francesco-ballarin:
@jhale can you please login to docker hub and manually remove the latest docker tag, so that people realize that they must use the nightly tag? Related to this: what will happen if one uses `docker pull dolfinx/dolfinx` without any tag, and without `latest` being available? An error? `latest` will redirect to `nightly`?

jhale:
No problem, I should have posted it here too!

The `:latest` tag will be removed in 30 days. 

It is true we will no longer have a default tag `:latest`.  I'd rather not point `:latest` at `:nightly` because then we will start having end users using `:nightly` with all of the issues that brings with support. I would rather stick with the new explicit design. If not having a `:latest` tag causes lots of issues we can revisit the issue at a later date.

PR #2293, opened 2022-07-26
Remove some function callback from C++ to Python

garth-wells:


cpp/dolfinx/mesh/utils.cpp
```diff
@@ -25,6 +25,136 @@
...
+  auto c_to_v = topology.connectivity(tdim, 0);
+  assert(c_to_v);
+  xt::xtensor<double, 2> x_vertices({3, vertices.size()});
+  std::vector<std::int32_t> vertex_to_pos(v_to_c->num_nodes(), -1);
+  for (std::size_t i = 0; i < vertices.size(); ++i)
+  {
+    const std::int32_t v = vertices[i];
+
+    // Get first cell and find position
+    const int c = v_to_c->links(v)[0];
```
jorgensd, line 141: Use `.front()`?

PR #2300, opened 2022-07-28
Try a MacOS build

jhale:


.github/workflows/macos.yml
```diff
@@ -0,0 +1,107 @@
...
+
+    steps:
+      - name: Setup Python
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.10'
+
+      - name: Install Homebrew dependencies
+        run: |
+          brew install boost cmake hdf5-mpi ninja open-mpi pkg-config pugixml python@3.10 xtensor # FEniCS
```
garth-wells, line 30: It looks a bit weird installing Python with an 'action' and with Homebrew.

jhale: Apparently the action 'activates' the desired Homebrew Python, rather than installing it.

To be explicit for people who want to understand how FEniCS is built on MacOS, I left in dependencies that are already installed in the runner images. I'd prefer clean images but...

I can add comments to this effect.

PR #2301, opened 2022-07-28
Create Basix elements directly, rather than going via UFL

mscroggs:
Made functions in the Python interface use Basix's UFL wrapper rather than making UFL elements that FFCx will later convert to element from Basix's UFL wrapper.

~This depends on https://github.com/FEniCS/ufl/pull/117, https://github.com/FEniCS/basix/pull/575, and https://github.com/FEniCS/ffcx/pull/515.~ These have been merged now.


python/dolfinx/fem/function.py
```diff
@@ -17,6 +17,8 @@
 import numpy as np
 import numpy.typing as npt
 
+import basix
+import basix.ufl_wrapper
```
garth-wells, line 21: Is this import required?

python/dolfinx/io/gmshio.py
```diff
@@ -9,7 +9,10 @@
 import numpy as np
 import numpy.typing as npt
 
+import basix
+import basix.ufl_wrapper
```
garth-wells, line 13: Is this import required?

python/dolfinx/io/gmshio.py
```diff
@@ -40,8 +43,10 @@ def ufl_mesh(gmsh_cell: int, gdim: int) -> ufl.Mesh:
     """
     shape, degree = _gmsh_to_cells[gmsh_cell]
     cell = ufl.Cell(shape, geometric_dimension=gdim)
-    scalar_element = ufl.FiniteElement("Lagrange", cell, degree, variant="equispaced")
-    return ufl.Mesh(ufl.VectorElement(scalar_element))
+
+    element = basix.ufl_wrapper.create_vector_element(
+        "Lagrange", cell.cellname(), degree, basix.LagrangeVariant.equispaced, dim=gdim, gdim=gdim)
```
garth-wells, line 48: Is there a Basix `enum` for Lagrange?

python/dolfinx/fem/function.py
mscroggs: Yes. As UFL is an optional dependency of Basix, this is not imported as part of basix by default to avoid errors for any users without UFL installed

python/dolfinx/io/gmshio.py
mscroggs: Yes, although the `create_*_element` convenience functions can also accept strings which are then converted using functions like `basix.finite_element.string_to_family`.

In this case, I've changed `"Lagrange"` to `basix.ElementFamily.P`

`cell.cellname()` is also a string, but I think it's best to leave it as so as the alternative is something ugly like `attr(basix.CellType, cell.cellname())`

garth-wells (approved):

PR #2306, opened 2022-08-01
Generalise topology construction

garth-wells:
Generalise construction of topology, specifically with respect to ghost cells.

* allows the partitioner to return any set of destinations for each cell
* changes the way `Topology` is computed, without relying on a band of ghost cells to find the interface between processes
* stores the interface between processes as `interprocess_facets` in `Topology`
* removes some assumptions about symmetry in MPI comms
* corrects many errors in comments
* removes some dead code from Topology construction

This may not be a fully complete set of changes required. Additional changes for a subsequent PR:

* fully check over the assumptions in creating shared vertices
* remove GhostMode entirely, leave it up to the partitioner (provide user with built-in partitioners for no ghosting, or facet based) 
* provide a utility to compute boundary vertices of a topology


jpdean (approved):
Looks good to me!

PR #2310, opened 2022-08-02
replace 'jit_paramters' with 'jit_option', and 'form_compiler_parameters' with 'form_compiler_options'

### LINKED ISSUE #1290: Change form_compiler_parameters, jit_parameters to form_compiler_options, jit_options
After some discussion in #1280, it has been concluded that we should rename these variables in FFCx and dolfin-X

SarahRo:
This fixes #1290.
Depends on https://github.com/FEniCS/ffcx/pull/518


SarahRo:
> I see what you're getting at with the naming, but is it really a significant change? :)

It may not be very significant but it closes an open issue. The issue was based on a discussion in https://github.com/FEniCS/dolfinx/pull/1280 and was aiming to make the naming more consistent (e.g., with "petsc_options"). The latest changes I've added also remove the inconsistency of sometimes using "params" and sometimes "parameters".

PR #2314, opened 2022-08-08
Reusing mesh when saving to Fides

massimiliano-leoni:
Currently `FidesWriter` saves the mesh whenever the `write` method is called. This is natural for time-dependent problems with a moving mesh but it's a waste of disk space for problems with an unchanging mesh.

I asked on the Fides issue tracker [see https://gitlab.kitware.com/vtk/fides/-/issues/15 ] and I found out that if a new mesh is not saved then Fides will automagically reuse the previous one.
This pull request adds a flag to let the user reuse a previously-saved mesh and does not change the previous default behaviour.


garth-wells:
Does it work if the mesh is periodically re-written?

massimiliano-leoni:
> Does it work if the mesh is periodically re-written?
> 
> @massimiliano-leoni

From a first test it seems that it doesn't work. I think it either wants as many meshes as there are time steps or just one.

massimiliano-leoni:
> If that's the case, maybe the re-write/don't re-write the mesh setting should go in the file constructor?

Yes, that makes more sense then. I will apply the change shortly.

massimiliano-leoni:
@garth-wells it's done, let me know if you like the implementation.

massimiliano-leoni:
By the way, I checked if VTX can do the same but apparently it can't. Or rather, the file is created successfully but then Paraview can't read it properly and crashes.

cpp/dolfinx/io/ADIOS2Writers.cpp
```diff
@@ -702,16 +702,18 @@ void ADIOS2Writer::close()
...
 {
   assert(_io);
   assert(mesh);
   fides_initialize_mesh_attributes(*_io, *mesh);
 }
 //-----------------------------------------------------------------------------
 FidesWriter::FidesWriter(MPI_Comm comm, const std::filesystem::path& filename,
-                         const ADIOS2Writer::U& u)
-    : ADIOS2Writer(comm, filename, "Fides function writer", u)
+                         const ADIOS2Writer::U& u, const bool reuse_mesh)
```
garth-wells, line 714: I never find `bool` argument very expressive. Maybe an `enum` would be better?

cpp/dolfinx/io/ADIOS2Writers.cpp
```diff
@@ -786,10 +788,15 @@ void FidesWriter::write(double t)
...
 
-  fides_write_mesh(*_io, *_engine, *_mesh);
+  if (_reuse_mesh == false or first_write == true)
+  {
+    fides_write_mesh(*_io, *_engine, *_mesh);
+  }
   for (auto& v : _u)
     std::visit([&](const auto& u) { fides_write_data(*_io, *_engine, *u); }, v);
 
+  first_write = false;
```
garth-wells, line 798: We try to avoid objects having mutable state. Can we check in the Fides file whether or not a mesh has been written?

cpp/dolfinx/io/ADIOS2Writers.h
```diff
@@ -147,6 +149,13 @@ class FidesWriter : public ADIOS2Writer
   /// @brief Write data with a given time
   /// @param[in] t The time step
   void write(double t);
+
+private:
+   /// @brief If true, the mesh is saved only at the first write and then reused,
```
garth-wells, line 154: No need for doxygen docstrings on private data. Just use `//`.

garth-wells (changes requested):

cpp/dolfinx/io/ADIOS2Writers.cpp
massimiliano-leoni: It's a binary choice so a `bool` looks semantically appropriate. I can of course put an `enum` if you like it better, something like `mesh_reuse_policy`.

cpp/dolfinx/io/ADIOS2Writers.h
massimiliano-leoni: Done

cpp/dolfinx/io/ADIOS2Writers.cpp
massimiliano-leoni: Done. Now it checks if the connectivity exists.

cpp/dolfinx/io/ADIOS2Writers.cpp
massimiliano-leoni: I introduced an `enum` with the same design as in `XDMFFile`.

garth-wells (approved):

PR #2339, opened 2022-08-29
Add demo for solving axisymmetric Maxwell's equations - GSoC 2022

mikics:
Currently, there are no demos showing how to solve axisymmetric electromagnetic problems. This pull request adds a demo for showing how to solve a simple problem of such kind: a metallic sphere hit by a plane wave (with axisymmetric PML condition). By exploiting the cylindrical expansion for a plane wave, we can split the full 3D problems into few 2D problems, with an overall significantly lower computational cost.


jhale:
@mikics This is still marked as a draft, could you comment on what still needs doing?

mikics:
> @mikics This is still marked as a draft, could you comment on what still needs doing?

I need to change the post-processing of the single component of the electric field, and to fix a few typos in the formula I implemented. Will do it in the next days!

PR #2347, opened 2022-09-05
Fix permuting of Lagrange spaces inside mixed elements

### LINKED ISSUE #2343: Issue with DOLFINx solution when solving axisymmetric Maxwell's equations
Hi,

I am trying to solve the axisymmetric version of Maxwell's equations in DOLFINx for a simple problem: an electromagnetic wave scattered by a metallic sphere. I have already solved this problem in legacy DOLFIN, and I have tested the results against analytical ones, getting a very good agreement.

For this problem, I discretize the $\rho$ and $z$ components of the electric field with Nedelec elements, while the $\phi$ component is discretized with Lagrange elements.

The **issue** in DOLFINx is the following one: whenever I use `degree = 2` for the elements, I get quite good results, as confirmed by a comparison between numerical and analytical efficiency. However, whenever I try to use `degree = 3` or higher, the solution in DOLFINx is affected by artifacts, and the error between the numerical and the analytical efficiency gets really large. This issue does not affect legacy DOLFIN, where the solution looks clean and agrees with analytical calculation even up to `degree = 6`. The image here below shows a comparison between:

- $\operatorname{Re}(E_\phi)$ in legacy DOLFIN, for `degree = 3`
- $\operatorname{Re}(E_\phi)$ in DOLFINx, for `degree = 2`
- $\operatorname{Re}(E_\phi)$ in DOLFINx, for `degree = 3`, which has the **aforementiond artifacts**

comparison

[In this repository](https://github.com/mikics/dolfinx-axisymmetric-issue) I have uploaded two minimal working examples in legacy DOLFIN and DOLFINx, showing this issue.

mscroggs:
Adjusted how permutations inside mixed elements are handled, so that the dofmap for `MixedElement(Nedelec, Lagrange)` is created correctly.

Fixes #2343.

### DIFF
--- a/cpp/dolfinx/fem/FiniteElement.cpp
+++ b/cpp/dolfinx/fem/FiniteElement.cpp
@@ -147,14 +147,12 @@ FiniteElement::FiniteElement(const ufcx_finite_element& e)
   {
     ufcx_finite_element* ufcx_sub_element = e.sub_elements[i];
     _sub_elements.push_back(std::make_shared<FiniteElement>(*ufcx_sub_element));
-    if (_sub_elements[i]->needs_dof_permutations()
-        and !_needs_dof_transformations)
+    if (_sub_elements[i]->needs_dof_permutations())
     {
       _needs_dof_permutations = true;
     }
     if (_sub_elements[i]->needs_dof_transformations())
     {
-      _needs_dof_permutations = false;
       _needs_dof_transformations = true;
     }
   }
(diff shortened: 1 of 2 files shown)


garth-wells (approved):
Could a test be added?

IgorBaratta (approved):

mscroggs:
> Could a test be added?

Yes, I'm currently turning https://github.com/FEniCS/dolfinx/issues/2343#issuecomment-1237209908 into a test

mscroggs:
I've added a test, and ran locally with main to confirm that the test fails on the current main

PR #2353, opened 2022-09-07
Fix bug where not all required vertices are added to submesh vertex map

jpdean:
This PR fixes a bug where not all vertices in the submesh are added to the submesh vertex map on some processes. It arises because `entities` may or may not contain ghosts, but the submesh entity map will always contain ghost entities present in the submesh.

### DIFF
--- a/cpp/dolfinx/mesh/Mesh.cpp
+++ b/cpp/dolfinx/mesh/Mesh.cpp
@@ -195,9 +195,51 @@ mesh::create_submesh(const Mesh& mesh, int dim,
 {
   // -- Submesh topology
 
-  // Get the vertices in the submesh
+  // Get the entities in the submesh that are owned by this process
+  auto mesh_entity_index_map = mesh.topology().index_map(dim);
+  assert(mesh_entity_index_map);
+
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/mesh/Mesh.cpp
```diff
@@ -264,6 +235,40 @@ mesh::create_submesh(const Mesh& mesh, int dim,
...
+  // (instead of `entities`) to ensure vertices for ghost entities are
+  // included
+  std::vector<std::int32_t> submesh_vertices
+      = compute_incident_entities(mesh, submesh_to_mesh_entity_map, dim, 0);
+
+  // Get the vertices in the submesh owned by this process
+  auto mesh_vertex_index_map = mesh.topology().index_map(0);
+  assert(mesh_vertex_index_map);
+  std::vector<int32_t> submesh_owned_vertices
+      = dolfinx::common::compute_owned_indices(submesh_vertices,
```
garth-wells, line 248: `dolfinx::` can be removed.

garth-wells (changes requested):
Looks good - just one small suggestion.

garth-wells (approved):

PR #2356, opened 2022-09-08
Remove xtensor

### LINKED ISSUE #2350: Remove `xtensor`
`xtensor` has been largely removed, but remains in parts of the C++ user interface. While xtensor can be convenient in user-defined expressions for interpolation, domain marking, etc, it:
- Is a extra dependency providing limited benefit
- Does not fully support C++20 (basic functionality is fine, some more exotic use can segfault)
- Can introduce ABI portability issues when it combined with `xsimd`.

garth-wells:
This change removes xtensor. xtensor was used only in the C++ interface for mesh entity marking and user expressions. Other internal use was removed earlier.

Removing xtensor
- reduces the number of dependencies
- avoids ABI incompatibilities that can sneak in if `xsimd` is installed
- avoids potential C++20 compatibility issues (xtensor does not support C++20)

Closes #2350.

### DIFF
--- a/cpp/dolfinx/fem/DirichletBC.cpp
+++ b/cpp/dolfinx/fem/DirichletBC.cpp
@@ -22,6 +22,11 @@
 using namespace dolfinx;
 using namespace dolfinx::fem;
 
+namespace stdex = std::experimental;
+using cmdspan3x_t
+    = stdex::mdspan<const double,
+                    stdex::extents<std::size_t, 3, stdex::dynamic_extent>>;
+
 namespace
 {
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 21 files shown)


francesco-ballarin:
Is it correct to say that after this PR is merged none of the packages in the FEniCSx ecosystem will require `xtl` or `xtensor`?

If so, people like @drew-parsons and myself should make sure to remove such dependency in their packaging workflows.

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -451,7 +455,11 @@ std::array<std::vector<std::int32_t>, 2> fem::locate_dofs_topological(
 //-----------------------------------------------------------------------------
 std::vector<std::int32_t> fem::locate_dofs_geometrical(
     const FunctionSpace& V,
-    const std::function<xt::xtensor<bool, 1>(const xt::xtensor<double, 2>&)>&
+    const std::function<std::vector<std::int8_t>(
+        std::experimental::mdspan<
```
chrisrichardson, line 459: Is this the same as `cmdspan3x_t` ?

garth-wells: The problem is that older version of doxygen, i.e. what we have in the test image, seem to match strings rather than the underlying type when checking that the declaration in the header match what is in the implementation file.

garth-wells: doxygen checks fails when using the typedef.

garth-wells:
> Is it correct to say that after this PR is merged none of the packages in the FEniCSx ecosystem will require `xtl` or `xtensor`?

Yes.

PR #2361, opened 2022-09-13
Add LGPL license badge to readme

jorgensd:


garth-wells (commented):
This mixes status badges with a static badge.

I'm not sure this change adds much useful.

jhale:
People will only see this `README.md` via GitHub's rendering. I feel GitHub already does a nice job pulling out the `LICENSE` information and placing it in the right column of the page.

PR #2379, opened 2022-09-26
skip mypy checks for Python 3.10

mscroggs:


mscroggs:
Closing this as mypy checks are not run for different Python versions on CI

PR #2385, opened 2022-09-28
Check MPI error codes

### LINKED ISSUE #2058: Check MPI return codes
All MPI returned error codes should be checked.

IgorBaratta:
In the following code `MPI_Allreduce` fails silently (but an error code is actually returned).
```c++
#include <dolfinx.h>
#include <dolfinx/fem/petsc.h>
#include <iostream>

using namespace dolfinx;

int main(int argc, char* argv[])
{
  PetscInitialize(&argc, &argv, nullptr, nullptr);

  {
    MPI_Comm comm{MPI_COMM_WORLD};
    MPI_Comm_dup(MPI_COMM_WORLD, &comm);

    [[maybe_unused]] int rank = dolfinx::MPI::rank(comm);
    [[maybe_unused]] int ierr0 = MPI_Comm_free(&comm);
    // dolfinx::MPI::assert_and_throw(MPI_COMM_WORLD, ierr0);

    int size = 0;
    [[maybe_unused]] int ierr1
        = MPI_Allreduce(&rank, &size, 1, MPI_INT, MPI_SUM, comm);
    // dolfinx::MPI::assert_and_throw(MPI_COMM_WORLD, ierr1);

    assert(size == dolfinx::MPI::size(MPI_COMM_WORLD));
  }

  PetscFinalize();
  return 0;
}
```

This PR adds a function to check whether an error code returned by an MPI function is equal to MPI_SUCCESS. If the check fails then it prints a useful error message and aborts. 

Fixes #2058.

### DIFF
--- a/cpp/dolfinx/common/MPI.cpp
+++ b/cpp/dolfinx/common/MPI.cpp
@@ -15,11 +15,7 @@ dolfinx::MPI::Comm::Comm(MPI_Comm comm, bool duplicate)
   if (duplicate and comm != MPI_COMM_NULL)
   {
     int err = MPI_Comm_dup(comm, &_comm);
-    if (err != MPI_SUCCESS)
-    {
-      throw std::runtime_error(
-          "Duplication of MPI communicator failed (MPI_Comm_dup)");
-    }
+    dolfinx::MPI::check_error(comm, err);
   }
   else
     _comm = comm;
@@ -42,11 +38,7 @@ dolfinx::MPI::Comm::~Comm()
   if (_comm != MPI_COMM_NULL)
   {
     int err = MPI_Comm_free(&_comm);
-    if (err != MPI_SUCCESS)
-    {
-      std::cout << "Error when destroying communicator (MPI_Comm_free)."
-                << std::endl;
-    }
+    dolfinx::MPI::check_error(_comm, err);
   }
 }
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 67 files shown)


francesco-ballarin:
Related to #2058 ?

IgorBaratta:
> Related to #2058 ?

Yes, it is. If merged, I think it would close #2058.

cpp/dolfinx/common/MPI.cpp
```diff
@@ -79,14 +79,32 @@ int dolfinx::MPI::rank(const MPI_Comm comm)
...
 int dolfinx::MPI::size(const MPI_Comm comm)
 {
   int size;
-  MPI_Comm_size(comm, &size);
+  int ierr = MPI_Comm_size(comm, &size);
+  dolfinx::MPI::assert_and_throw(comm, ierr);
   return size;
 }
 //-----------------------------------------------------------------------------
+void dolfinx::MPI::assert_and_throw(MPI_Comm comm, int error_code)
```
garth-wells, line 90: I don't think the function name is quite right. `assert` might imply that the check is skipped in non-debug mode, and the function doesn't 'throw' an exception; it just aborts.

cpp/dolfinx/common/MPI.h
```diff
@@ -74,6 +74,13 @@ int rank(MPI_Comm comm);
 /// communicator
 int size(MPI_Comm comm);
 
+/// @brief Checks wether an error code returned by an MPI
+/// function is equal to MPI_SUCCESS. If the check fails then
+/// throw a runtime error.
```
garth-wells, line 79: 'runtime error' is a bit vague. Does it just 'abort'?

garth-wells (changes requested):

cpp/dolfinx/common/MPI.h
IgorBaratta: Yep, I need to review documentation and function name.
I was throwing a runtime error before, but it can cause a deadlock when `MPI_Abort` fails.
This can happen when the communicator is NULL (
reusing a communicator that has been freed) followed by a barrier with `MPI_COMM_WORLD`.
So I think a sensible solution is to forcibly abort the execution while still printing the error message.

garth-wells (approved):

PR #2388, opened 2022-09-29
Create Basix elements directly in high order mesh tests

mscroggs:
This is the easiest/best way to make these elements use equispaced points.

This and https://github.com/FEniCS/ffcx/pull/538 need to be merged together

### DIFF
--- a/cpp/demo/poisson/poisson.py
+++ b/cpp/demo/poisson/poisson.py
@@ -23,8 +23,7 @@ element = FiniteElement("Lagrange", triangle, 1)
 # (:math:`u` and :math:`v`) and the coefficient functions (:math:`f` and
 # :math:`g`)::
 
-coord_element = create_vector_element(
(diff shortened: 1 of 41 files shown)


cpp/demo/poisson/poisson.py
```diff
@@ -21,7 +23,8 @@
 # (:math:`u` and :math:`v`) and the coefficient functions (:math:`f` and
 # :math:`g`)::
 
-coord_element = VectorElement("Lagrange", triangle, 1)
+coord_element = create_vector_element(
+    "Lagrange", "triangle", 1, lagrange_variant=basix.LagrangeVariant.equispaced)
```
garth-wells, line 27: Could we have a default argument for the variant? Seems heavy to require a variant for a P1 element.

mscroggs: There was a mismatch between the defaults Basix and DOLFINx were using which was causing errors. I've fixed this and removed these variants for meshes with degree < 3

garth-wells (approved):

PR #2390, opened 2022-10-03
Add divergence conforming discontinuous Galerkin demo for the Navier-Stokes equations

jpdean:
This PR adds a divergence conforming discontinuous Galerkin demo for the Navier-Stokes equations.

It demonstrates:
- how to implement DG methods
- how to use predefined and custom restriction operators
- how upwinding can be implemented
- the use of H(div) elements in a non-conforming context

As far as I'm aware, none of the other demos demonstrate the above functionality.

I still need to so some tidying. Any feedback would be much appreciated.


jpdean:
If anyone feels this would be better kept outside the main repository, it could be added to the [FEniCSx tutorial](https://jorgensd.github.io/dolfinx-tutorial/) instead if @jorgensd is happy

python/demo/demo_navier_stokes.py
```diff
@@ -0,0 +1,419 @@
...
+#       format_version: '1.5'
+#       jupytext_version: 1.13.6
+# ---
+
+# # A divergence conforming discontinuous Galerkin method for the Navier-Stokes equations
+# This demo illustrates how to implement a divergence conforming
+# discontinuous Galerkin method for the Navier-Stokes equations in
+# FEniCSx. The method conserves mass exactly and uses upwinding. The
+# formulation is based on a combination of "A fully divergence-free
+# finite element method for magnetohydrodynamic equations" by Himpmair
```
nate-sime, line 16:
```suggestion
# finite element method for magnetohydrodynamic equations" by Hiptmair
```

python/demo/demo_navier_stokes.py
```diff
@@ -0,0 +1,419 @@
...
+# of the Navier-Stokes Equations" by Cockburn et al, and "On the Divergence
+# Constraint in Mixed Finite Element Methods for Incompressible Flows" by
+# John et al.
+
+# ## Governing equations
+# We consider the incompressible Navier-Stokes equations in a domain
+# $\Omega \subset \mathbb{R}^d$, $d \in \{2, 3\}$, and time interval
+# $(0, \infty)$, given by
+# $$
+#     \partial_t u - \nu \Delta u + (u \cdot \nabla)u + \nabla p = f
```
nate-sime, line 27: Does `\nu` actually get used in the discretisation?

python/demo/demo_navier_stokes.py
```diff
@@ -0,0 +1,419 @@
...
+# We define some simulation parameters
+
+n = 16
+num_time_steps = 25
+t_end = 10
+R_e = 25  # Reynolds Number
+k = 1  # Polynomial degree
+
+# Next, we create a mesh and the required functions spaces over
+# it. Since the velocity uses an $H(\textnormal{div}$-conforming function
```
nate-sime, line 210:
```suggestion
# it. Since the velocity uses an $H(\textnormal{div})$-conforming function
```

python/demo/demo_navier_stokes.py
```diff
@@ -0,0 +1,419 @@
...
+
+# Boundary conditions
+
+boundary_facets = mesh.locate_entities_boundary(
+    msh, msh.topology.dim - 1, boundary_marker)
+boundary_vel_dofs = fem.locate_dofs_topological(
+    V, msh.topology.dim - 1, boundary_facets)
+bc_u = fem.dirichletbc(u_D, boundary_vel_dofs)
+
+# The pressure is only determined up to a constant, so pin a single degree
```
nate-sime, line 275: Is this compatible with the constraint stated above that `\int_\Omega p \mathrm{d}\vec{x} = 0`?

python/demo/demo_navier_stokes.py
```diff
@@ -0,0 +1,419 @@
...
+u_file = io.VTXWriter(msh.comm, "u.bp", [u_vis._cpp_object])
+p_file = io.VTXWriter(msh.comm, "p.bp", [p_h._cpp_object])
+t = 0.0
+u_file.write(t)
+p_file.write(t)
+
+# Create function to store solution and previous time step
+
+u_n = fem.Function(V)
+u_n.x.array[:] = u_h.x.array
```
nate-sime, line 344:
```suggestion
u_n.x.array[:] = u_h.x.array_r
```

python/demo/demo_navier_stokes.py
jpdean: The discretisation uses the dimensionless form, where `\nu` is incorporated into the Reynolds number. I'll make this clearer

python/demo/demo_navier_stokes.py
jpdean: No, it's just a simple way to fix the pressure in the discrete problem, but wouldn't be mathematically valid in the continuous problem. I could subtract the average off the pressure, as I do when computing the error norm. Alternatively, I could change the solver setting and remove the nullspace without pinning a pressure dof.

python/demo/demo_navier_stokes.py
jpdean: `u_h.x` returns a dolfinx vector, not a PETSc vector

python/demo/demo_navier_stokes.py
```diff
@@ -0,0 +1,428 @@
...
+a_00 = (1 / Re) * (inner(grad(u), grad(v)) * dx
+                   - inner(avg(grad(u)), jump(v, n)) * dS
+                   - inner(jump(u, n), avg(grad(v))) * dS
+                   + alpha / avg(h) * inner(jump(u, n), jump(v, n)) * dS
+                   - inner(grad(u), outer(v, n)) * ds
+                   - inner(outer(u, n), grad(v)) * ds
+                   + alpha / h * inner(outer(u, n), outer(v, n)) * ds)
+a_01 = - inner(p, div(v)) * dx
+a_10 = - inner(div(u), q) * dx
+a_11 = fem.Constant(msh, 0.0) * inner(p, q) * dx
```
garth-wells, line 250: This is rather ugly. I see that it's required to support the Dirichlet condition to constrain the pressure.

Is there a better way? Maybe supporting 'diagonal' matrices from the DOLFINx side, where for block problems the shape can be deduced from the other blocks?

python/demo/demo_navier_stokes.py
```diff
@@ -275,7 +275,7 @@ def jump(phi, n):
     pressure_dof = [pressure_dofs[0]]
 else:
     pressure_dof = []
-bc_p = fem.dirichletbc(0.0, np.array(pressure_dof, dtype=np.int32), Q)
+bc_p = fem.dirichletbc(PETSc.ScalarType(0.0), np.array(pressure_dof, dtype=np.int32), Q)
```
garth-wells, line 278: This is ugly, especially since the demo doesn't support complex types (due to a bunch of UFL shortcomings). Why is it required for mypy?

jpdean: `float` isn't included in the typing list for `dirichletbc`. I'll revert the change to this PR and make a new PR against dolfinx with float added

python/demo/demo_navier_stokes.py
jpdean: In the case of this demo, I'd rather not have to apply  Dirichlet BC to the pressure space at all. MUMPS should be able to deal with the small nullspace, I think. I'll see what I can do to tidy it up.

Still, having the functionality you mention would be useful for other applications

PR #2413, opened 2022-10-21
Remove redefinition of `VERSION_INFO`

jpdean:
Installing the python layer with Spack fails with `error: "VERSION_INFO" redefined`. This PR removes the line where `VERSION_INFO` is redefined.

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -59,8 +59,6 @@ class CMakeBuild(build_ext):
 
         import pybind11
         env['pybind11_DIR'] = pybind11.get_cmake_dir()
-        env['CXXFLAGS'] = '{} -DVERSION_INFO=\\"{}\\"'.format(env.get('CXXFLAGS', ''),
-                                                              self.distribution.get_version())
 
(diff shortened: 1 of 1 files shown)


chrisrichardson (approved):

jhale:
We should look and see if it's possible to move to a combined pyproject.toml/setup.py configuration for setuptools:

https://setuptools.pypa.io/en/latest/userguide/ext_modules.html

All of the metadata would be in pyproject.toml file, just the cmake stuff would remain in setup.py.

jorgensd:
We would need to move the flake8 options to a `.flake8` file (which is fine by me). See:
https://github.com/PyCQA/flake8/issues/234

garth-wells:
I don't think this is a good 'fix'. My understanding is that it versions the binaries. I suspect the problem lies elsewhere in the Spack config.

chrisrichardson:
Ok, let's revisit before a release. It was not documented, and seemed to have no purpose.

PR #2245, opened 2022-06-16
Interpolation between different meshes

massimiliano-leoni:
This pull request supersedes #1598 .
This pull request implements the same algorithm using the MPI Neighbourhood communication paradigm instead of the MPI One-sided communication one. The paradigm switch is joint work with @IgorBaratta .

PR #2414, opened 2022-10-25
Add functionalities to cache data when interpolating on nonmatching meshes

massimiliano-leoni:
After merging #2245 dolfinx can interpolate discrete functions defined on different meshes. This is, however, quite costly.
This PR adds a class that performs this interpolation caching as much data as possible so that interpolating again between the same two functions, when their meshes haven't moved, is much cheaper.

Some timings for the same example as in the discussion of #2245 [mesh with 750'000 cells]
```
serial
Interpolating       7.311494
Interpolating again 0.572593

n = 6 processes
Interpolating       1.772976
Interpolating again 0.164332
```


jhale:
Would it be possible to implement this not using a class? For example, the interpolate method could return a map containing the cache which could then be passed as an optional argument. `std::optional` might help.

jhale:
Another (perhaps better option) would be to bring the expensive call to `determine_point_ownership` outside the function.

IgorBaratta:
> Another (perhaps better) option would be to bring the expensive call to `determine_point_ownership` outside the function.

There's also opportunity of improvement in eval, which is quite expensive (evaluating basis at interpolation points).
Maybe, for starters, we could use a more explicit way of caching data, with the functions that are already exist (in a demo for example).

massimiliano-leoni:
@jhale I guess it is possible to avoid using a class. Out of ~~ignorance~~ curiosity, what is the matter with using a class?

massimiliano-leoni:
@jhale I understand that it's a design decision, but I was wondering for my own knowledge what the reason behind the decision is!

garth-wells:
> @jhale I understand that it's a design decision, but I was wondering for my own knowledge what the reason behind the decision is!

Avoiding, where possible, objects with (mutable) state leads to simpler code (not need to check state), avoids inconsistent state, avoids hidden memory 'leaks' (not technically leaks, but objects that might hold a lot of memory and which isn't obvious to the user), and thread safety. 

This is why we prefer explicit approaches first. Sometimes it might then be appropriate to build a convenience class/function on top of the low-level implementation.

For this particular case, it seems simplest to me that the user calls `geometry::determine_point_ownership` and then uses the data. This give more flexibility, e.g. a case where the caller knows the cells and just needs to determine the ranks.

massimiliano-leoni:
@garth-wells Thanks for the explanation. I implemented it the way you and @jhale suggested. Is it better now?

cpp/dolfinx/fem/interpolate.h
```diff
@@ -835,103 +841,81 @@ void interpolate(Function<T>& u, std::span<const T> f,
...
-/// Interpolate from one finite element Function to another
-/// @param[in,out] u The function to interpolate into
-/// @param[in] v The function to be interpolated
-/// @param[in] cells List of cell indices to interpolate on
+/// Generate data needed to interpolate discrete functions across different
+/// meshes
+///
+/// @param[out] u The function to interpolate into
+/// @param[in] v The function to interpolate from
+/// @param[in] cells Indices of the cells in the mesh on which to
```
jhale, line 849: Specifically refer to whether these are cells associated with mesh on function u or v.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -835,103 +841,81 @@ void interpolate(Function<T>& u, std::span<const T> f,
...
 /// @param[in] v The function to be interpolated
 /// @param[in] cells List of cell indices to interpolate on
+/// @param[in] m2mInterpolationData Shared pointer to fem::M2MInterpolationData
+/// that provides cached data to speed up the procedure when interpolating the
+/// same functions multiple times (optional).
 template <typename T>
 void interpolate(Function<T>& u, const Function<T>& v,
-                 std::span<const std::int32_t> cells)
+                 std::span<const std::int32_t> cells,
+                 std::shared_ptr<M2MInterpolationData> m2mInterpolationData
```
jhale, line 917: I don't feel using `std::shared_ptr` is necessary. Pass by reference?

cpp/dolfinx/fem/interpolate.h
```diff
@@ -952,7 +936,70 @@ void interpolate(Function<T>& u, const Function<T>& v,
   {
     // Get mesh and check that functions share the same mesh
     if (auto mesh_v = v.function_space()->mesh(); mesh != mesh_v)
-      impl::interpolate_nonmatching_meshes(u, v, cells);
+    {
```
jhale, line 939: Can we keep the function `impl::interpolate_nonmatching_meshes` by redefining it to take the precomputed data?

cpp/dolfinx/fem/interpolate.h
```diff
@@ -835,103 +841,81 @@ void interpolate(Function<T>& u, std::span<const T> f,
...
+}
+
+/// Generate data needed to interpolate discrete functions across different
+/// meshes. Interpolate on all cells in the mesh.
+///
+/// @param[out] u The function to interpolate into
+/// @param[in] v The function to interpolate from
+/// @tparam Scalar type
+template <typename T>
+M2MInterpolationData generate_m2m_interpolation_data(const Function<T>& u,
```
jhale, line 891: The terminology is getting a bit confusing. We used `nonmatching_meshes`?

cpp/dolfinx/fem/interpolate.h
```diff
@@ -952,7 +936,70 @@ void interpolate(Function<T>& u, const Function<T>& v,
...
+      std::shared_ptr<const FiniteElement> element_u
+          = u.function_space()->element();
+      const std::size_t value_size = element_u->value_size();
+
+      std::tuple_element_t<0, M2MInterpolationData> dest_ranks;
+      std::tuple_element_t<1, M2MInterpolationData> src_ranks;
+      std::tuple_element_t<2, M2MInterpolationData> received_points;
+      std::tuple_element_t<3, M2MInterpolationData> evaluation_cells;
+
+      if (m2mInterpolationData == nullptr)
```
jhale, line 960: Best to make the expensive call mandatory, users then are 'forced' to understand that it is there.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -38,6 +38,12 @@ class Function;
 std::vector<double> interpolation_coords(const fem::FiniteElement& element,
                                          const mesh::Mesh& mesh,
                                          std::span<const std::int32_t> cells);
+
+/// Helper type for the data that can be cached to speed up repeated
+/// interpolation of discrete functions across different meshes
+using M2MInterpolationData = decltype(std::function{
```
jhale, line 44: @garth-wells for comments on this style, perhaps its more transparent to type the return type explicitly?

In any case, we normally use `blah_blah_t` for user-defined types.

cpp/dolfinx/fem/interpolate.h
jhale: Does the function actually use the Function u or v? If not, then it would be better to pass the FunctionSpaces.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -835,103 +841,81 @@ void interpolate(Function<T>& u, std::span<const T> f,
...
-      valuesT(j, i) = values(i, j);
-
-  // Call local interpolation operator
-  fem::interpolate<T>(u, std::span(valuesT_b.data(), valuesT_b.size()),
-                      {valuesT.extent(0), valuesT.extent(1)}, cells);
+  return dolfinx::geometry::determine_point_ownership(*mesh_v, x);
+}
+
+/// Generate data needed to interpolate discrete functions across different
+/// meshes. Interpolate on all cells in the mesh.
```
jhale, line 885: Maybe 'discrete functions defined on different meshes'.

jhale (changes requested):

jhale:
Pretty much I think, just a few style things I left in the review. Another thing we do is try to make expensive calls explicit, so I don't think we would include a code path where the data is not pre-computed in this low-level interface.

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: It doesn't use the functions directly but I think that passing the function makes "more sense" for the user as they are the objects he or she wants to work with. The function space is "an implementation detail".

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: I didn't mean to remove it, I just forgot to restore it :)

cpp/demo/interpolation_different_meshes/main.cpp
```diff
@@ -66,6 +66,13 @@ int main(int argc, char* argv[])
     // Interpolate from u_tet to u_hex
     u_hex->interpolate(*u_tet);
 
+    // Cache data explicitly and use it to interpolate quickly
+    u_hex->x()->set(0);
+    auto m2mInterpolationData
```
jhale, line 71: I should have been more specific; we don't use camel case for variable names e.g. `m2m_interpolation_data`.

jhale: We still have 'm2m' and 'nmm'. Could we standardise on 'non_matching_meshes' and 'nmm'?

cpp/dolfinx/fem/interpolate.h
```diff
@@ -38,6 +38,18 @@ class Function;
 std::vector<double> interpolation_coords(const fem::FiniteElement& element,
                                          const mesh::Mesh& mesh,
                                          std::span<const std::int32_t> cells);
+
+/// Helper type for the data that can be cached to speed up repeated
+/// interpolation of discrete functions on nonmatching meshes
+using NMMInterpolationData_t = decltype(std::function{
```
jhale, line 44: `nmm_interpolation_data_t`

cpp/dolfinx/fem/interpolate.h
```diff
@@ -835,103 +916,82 @@ void interpolate(Function<T>& u, std::span<const T> f,
...
+/// @param[in] v The function to interpolate from
+/// @param[in] cells Indices of the cells in the destination mesh on which to
+/// interpolate. Should be the same as the list used when calling
+/// fem::interpolation_coords.
+/// @tparam Scalar type
 template <typename T>
-void interpolate_nonmatching_meshes(Function<T>& u, const Function<T>& v,
-                                    std::span<const std::int32_t> cells)
+NMMInterpolationData_t generate_nonmatching_meshes_interpolation_data(
+    const Function<T>& u, const Function<T>& v,
```
jhale, line 930: I don't agree that it's an implementation detail.

1. It makes it clear to the user that nothing in the generated data is tied to the functions, only to the spaces.
2. There might be a circumstance when the user needs to make this data prior to creating a first function (I don't have a concrete example).

cpp/dolfinx/fem/interpolate.h
```diff
@@ -835,103 +916,82 @@ void interpolate(Function<T>& u, std::span<const T> f,
...
+/// @param[out] u The function to interpolate into
+/// @param[in] v The function to interpolate from
+/// @param[in] cells Indices of the cells in the destination mesh on which to
+/// interpolate. Should be the same as the list used when calling
+/// fem::interpolation_coords.
+/// @tparam Scalar type
 template <typename T>
-void interpolate_nonmatching_meshes(Function<T>& u, const Function<T>& v,
-                                    std::span<const std::int32_t> cells)
+NMMInterpolationData_t generate_nonmatching_meshes_interpolation_data(
```
jhale: We are using the verb create rather than generate.

jhale:
I think this is close. How does it all look from the Python interface?

cpp/dolfinx/fem/interpolate.h
massimiliano-leoni: Ok, I'll do it your way.

massimiliano-leoni:
> I think this is close. How does it all look from the Python interface?

I think I updated it accordingly. At least, the CI is satisfied. I suggest someone quickly goes through my additions because I'm not very familiar with pybind11 and writing Python wrappers so I might have overlooked something.

jhale (approved):

PR #2416, opened 2022-10-27
Attempt to use pyproject.toml

jhale:
This is work in progress and probably not ready for merge today for a few reasons.

1. pip has moved to a model of isolated builds. pip sets up a totally clean Python environment, installs the build dependencies, and then builds the wheel before installing it. Our cmake build asks this clean python environment for the location of the petsc4py and slepc4py headers, which then fails in this isolated environment. Adding e.g. petsc4py to the build environment specification works, but can lead to the runtime environment having a different ABI to the build time. pip/setuptools seems to have no concept of a build and run dependency.

3. pip and setuptools in the distributions have to be upgraded beyond any packaged version (setuptools>=61, pip 'newish' version).

4. `pip install --no-build-isolation .` works, but then it is (by design) entirely the users responsibility to make sure the build environment is prepared. This can be done using a `requirements.txt` file but it's not very pretty.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,7 @@ build-python-interface: &build-python-interface
   name: Build Python/pybind11 interface
   command: |
     cd python
-    pip3 -v install --global-option build --global-option --debug . --user
+    pip3 -v install --no-build-isolation --global-option build --global-option --debug . --user
 
 demos-python: &demos-python
   name: Run demos (Python, serial)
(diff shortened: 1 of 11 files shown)


jhale:
Status update: PEP Isolated builds don't really work for projects like DOLFINx which by definition require matching build and runtime dependencies. Should be kept open for the future.

jhale:
Superseded by #2707

PR #2422, opened 2022-10-30
Use basix data in get_sub_entities

tiagovla:
This PR should replace hard-coded data with basix data.


tiagovla:
Are those static variables essential? If yes, maybe `static graph::AdjacencyList` instead would be better. If not, perhaps a `graph::AdjacencyList` constructor taking a `range`?

garth-wells:
This looks neat, but a concern is that C++20 ranges support is patchy across compilers. This is why we haven't used ranges anywhere is the code base yet.

tiagovla:
> This looks neat, but a concern is that C++20 ranges support is patchy across compilers. This is why we haven't used ranges anywhere is the code base yet.

I noticed the tests failed and the feature I used will only be fixed in clang 16.0 https://godbolt.org/z/d5M6enh5a. I will change it and use `std::vector` instead.

Are the static variables necessary?

tiagovla:
I'm not sure why `dim0=2` is defined for `point` and `interval`. I left those there in case they are used somewhere else.

PR #2424, opened 2022-10-31
Fix `set_exterior_facet_domains` for arbitrary ghosting

jpdean:
Currently, `set_exterior_facet_domains` makes assumptions about the ghosting of the mesh. Looks like it was missed when https://github.com/FEniCS/dolfinx/pull/2337 was merged. This PR removes these assumptions. Unfortunately, it relies on an `std::lower_bound` in a loop, which is not ideal. However, I plan to replace this in https://github.com/FEniCS/dolfinx/pull/2269

### DIFF
--- a/cpp/dolfinx/fem/Form.h
+++ b/cpp/dolfinx/fem/Form.h
@@ -487,12 +487,12 @@ private:
         {
           // Ge the facet as a pair of (cell, local facet) pairs, one for each
           // cell
-          std::array<std::array<std::int32_t, 2>, 2> facets
+          auto [facet_0, facet_1]
               = get_cell_local_facet_pairs<2>(f, f_to_c->links(f), *c_to_f);
-          it->second.second.insert(it->second.second.end(), facets[0].cbegin(),
-                                   facets[0].cend());
-          it->second.second.insert(it->second.second.end(), facets[1].cbegin(),
-                                   facets[1].cend());
+          it->second.second.insert(it->second.second.end(), facet_0.cbegin(),
+                                   facet_0.cend());
+          it->second.second.insert(it->second.second.end(), facet_1.cbegin(),
+                                   facet_1.cend());
         }
       }
     }
--- a/cpp/dolfinx/fem/Form.h
+++ b/cpp/dolfinx/fem/Form.h
@@ -467,11 +467,10 @@ private:
   }
 
   // Set interior facet domains
-  template <typename iterator>
   static void set_interior_facet_domains(
       const mesh::Topology& topology,
       std::map<int, std::pair<kern, std::vector<std::int32_t>>>& integrals,
-      const iterator& tagged_facets_begin, const iterator& tagged_facets_end,
+      const std::span<const std::int32_t>& owned_tagged_facets,
       const std::vector<int>& tags)
   {
     int tdim = topology.dim();
@@ -479,19 +478,21 @@ private:
     assert(f_to_c);
     auto c_to_f = topology.connectivity(tdim, tdim - 1);
     assert(c_to_f);
-    for (auto f = tagged_facets_begin; f != tagged_facets_end; ++f)
+    for (int i = 0; i < owned_tagged_facets.size(); ++i)
     {
-      if (f_to_c->num_links(*f) == 2)
+      const std::int32_t f = owned_tagged_facets[i];
+      if (f_to_c->num_links(f) == 2)
       {
-        const std::size_t pos = std::distance(tagged_facets_begin, f);
-        if (auto it = integrals.find(tags[pos]); it != integrals.end())
+        if (auto it = integrals.find(tags[i]); it != integrals.end())
         {
-          const std::array<std::array<std::int32_t, 2>, 2> pairs
-              = get_cell_local_facet_pairs<2>(*f, f_to_c->links(*f), *c_to_f);
-          it->second.second.insert(it->second.second.end(), pairs[0].cbegin(),
-                                   pairs[0].cend());
-          it->second.second.insert(it->second.second.end(), pairs[1].cbegin(),
-                                   pairs[1].cend());
+          // Ge the facet as a pair of (cell, local facet) pairs, one for each
+          // cell
+          const std::array<std::array<std::int32_t, 2>, 2> facets
+              = get_cell_local_facet_pairs<2>(f, f_to_c->links(f), *c_to_f);
+          it->second.second.insert(it->second.second.end(), facets[0].cbegin(),
+                                   facets[0].cend());
+          it->second.second.insert(it->second.second.end(), facets[1].cbegin(),
+                                   facets[1].cend());
         }
       }
     }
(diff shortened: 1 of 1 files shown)


chrisrichardson (approved):
This fixes an issue, and is no worse than what went before, in terms of efficiency (replaces binary search with lower bound)...

cpp/dolfinx/fem/Form.h
```diff
@@ -422,46 +422,37 @@ class Form
...
-    const std::vector<std::int32_t> fwd_shared_facets
-        = topology.index_map(tdim)->overlapped()
-              ? std::vector<std::int32_t>()
-              : topology.index_map(tdim - 1)->shared_indices();
+    const std::vector<std::int32_t> boundary_facets
+        = mesh::exterior_facet_indices(topology);
 
+    // Create list of tagged boundary facets
+    std::vector<std::int32_t> tagged_ext_facets;
+    std::set_intersection(tagged_facets_begin, tagged_facets_end,
```
garth-wells, line 430: This requires that the input arrays are sorted. Is this guaranteed?

Any preconditions on the function input with regard to sorting should be documented.

cpp/dolfinx/fem/Form.h
```diff
@@ -422,46 +422,37 @@ class Form
...
-            const std::array<std::int32_t, 2> pair
-                = get_cell_local_facet_pairs<1>(*f, f_to_c->links(*f),
-                                                *c_to_f)[0];
-            it->second.second.insert(it->second.second.end(), pair.cbegin(),
-                                     pair.cend());
-          }
-        }
+        // There will only be one pair for an exterior facet integral
+        const std::array<std::int32_t, 2> pair
+            = get_cell_local_facet_pairs<1>(f, f_to_c->links(f), *c_to_f)[0];
```
garth-wells, line 453: Could `[0]` be `.front()`?

garth-wells (changes requested):

cpp/dolfinx/fem/Form.h
jpdean: Yes, the input arrays are already sorted. I'll add documentation

cpp/dolfinx/fem/Form.h
```diff
@@ -415,53 +415,50 @@ class Form
   }
 
   // Set exterior facet domains
+  // @param[in] topology The mesh topology
+  // @param[in] integrals The integrals to set exterior facet domains for
+  // @param[in] tagged_facets_begin The start of the list of facets
+  // NOTE: The list of tagged facets must be sorted
```
garth-wells, line 421: Use `@pre` (https://doxygen.nl/manual/commands.html#cmdpre)

cpp/dolfinx/fem/Form.h
```diff
@@ -415,53 +415,50 @@ class Form
...
-                = get_cell_local_facet_pairs<1>(*f, f_to_c->links(*f),
-                                                *c_to_f)[0];
-            it->second.second.insert(it->second.second.end(), pair.cbegin(),
-                                     pair.cend());
-          }
-        }
+        // There will only be one pair for an exterior facet integral
+        const std::array<std::int32_t, 2> pair
+            = get_cell_local_facet_pairs<1>(f, f_to_c->links(f), *c_to_f).front();
+        it->second.second.insert(it->second.second.end(), pair.cbegin(),
```
garth-wells, line 460: I find pairs of pairs really hard to read. Could an intermediate reference be introduced for readability?

garth-wells (changes requested):

cpp/dolfinx/fem/Form.h
garth-wells: Also, could replace the two iterators by a `std::span`?

cpp/dolfinx/fem/Form.h
```diff
@@ -415,53 +415,50 @@ class Form
...
-            // There will only be one pair for an exterior facet integral
-            const std::array<std::int32_t, 2> pair
-                = get_cell_local_facet_pairs<1>(*f, f_to_c->links(*f),
-                                                *c_to_f)[0];
-            it->second.second.insert(it->second.second.end(), pair.cbegin(),
-                                     pair.cend());
-          }
-        }
+        // There will only be one pair for an exterior facet integral
+        const std::array<std::int32_t, 2> pair
```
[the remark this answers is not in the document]
jpdean: I can make it not `const`. Out of interest, what harm is `const` doing?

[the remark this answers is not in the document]
garth-wells: Just taste/style. It doesn't do any harm, but is extra syntax (reduced readability) that doesn't give any significant benefit.

garth-wells (approved):

francesco-ballarin:
@jpdean is this supposed to affect serial runs too, or not? 

I have a few tutorials in multiphenicsx that worked before this PR was merged, and failed tonight, in some cases returning nan on solve, in other cases returning a value of `dolfinx.fem.assemble_scalar` which is not the expected one.

The failing tutorials are
```
tutorial_3a_advection_diffusion_reaction_neumann_control.ipynb
tutorial_4a_poisson_dirichlet_control.ipynb
tutorial_4b_poisson_neumann_control_boundary_observation.ipynb
tutorial_8_navier_stokes_neumann_control.ipynb
```
in
https://github.com/multiphenics/multiphenicsx/tree/main/tutorials/06_optimal_control
while
```
tutorial_3b_advection_diffusion_reaction_neumann_control.ipynb
tutorial_7a_stokes_dirichlet_control.ipynb
tutorial_7b_stokes_neumann_control.ipynb
```
are tutorials that use `ds` or `dS` too, but are not failing.

The only difference I can spot between the two groups is that in the failing ones I only define a custom `ds` measure and use `ufl.dx` for volume integration, while in the ones that do not fail I define either a custom `dx` or a custom `dS` measure too. Can this really be the relevant difference? Is there a unit test covering both cases?

chrisrichardson:
I think there may be a lack of suitable unit tests. The PR should have fixed a bug in correctly determining surface facets for `ds` integrals. Would it be possible for you to create a minimal failing example from one of your notebooks?

jpdean:
Oh no! Thanks for reporting this. I'll do some tests to see if I can figure out what is going wrong. As Chris mentioned, a minimum failing example would be really helpful.

garth-wells:
I don't think we should have the check on by default - that's what we have DEBUG mode for. Release mode checks should be O(1).

jorgensd:
I guess it should be added to the Python documentation, as most Python users will not dig into the C++ documentation.
https://docs.fenicsproject.org/dolfinx/v0.5.1/python/generated/dolfinx.mesh.html#dolfinx.mesh.meshtags

francesco-ballarin:
Thanks @chrisrichardson for sorting (pun intended ;) ) this out! I've pushed a fix to my repo.

I've gone through `git blame` of both `dolfinx` and `multiphenicsx` and apparently the timeline was the following:
* on March 30, 2020, `MeshTags` were introduced in `dolfinx`, without any sorting requirement
* on April 9, 2020, I updated my tutorials to use `MeshTags`
* on April 11, 2020, the sorting requirement was added in `dolfinx`, and somehow I missed or overlooked that ;)
This was clearly my fault, but I agree that it might be helpful to add the requirement to the python documentation too. Still, I am amazed that this mistake only came to bite me 2.5 years later! :)

PR #2425, opened 2022-11-01
Explicitly include loguru using cmake

tiagovla:
This PR removes `#include "loguru.cpp"` and sets compile definition flags using cmake.


jhale:
Main question: Does this improve compile time? At the moment we are using log guru in a kind of header-only mode, which is nice from a simplicity point of view.

Further comments:

- The logguru.hpp include can perhaps be moved to log.cpp
- The macro definitions should be removed from log.h

tiagovla:
> Main question: Does this improve compile time? At the moment we are using log guru in a kind of header-only mode, which is nice from a simplicity point of view.

I might be wrong, but compile time should be the same. The only difference is that before `loguru.cpp` was being included in `log.cpp` and compiled there. Now, it's compiled in its own translation unit.

> Further comments:
> 
> * The logguru.hpp include can perhaps be moved to log.cpp

I think INFO, WARNING, and other identifiers are defined in the header, so they should be included. 

> * The macro definitions should be removed from log.h

I tested and they are still needed before `loguru.hpp`. That wouldn't be the case if the header was included in the cmake interface. I guess that's probably the reason why some projects I've seen move third-party dependencies to their own folder.

jhale:
I just reviewed this again and doesn't seem to bring any major advantages over the existing approach. A major disadvantage is the duplication of the options between the two places (violates DRY principles). I'm going to close with comment to keep things tidy, but I'm still willing to be persuaded otherwise.

Issue #2132, opened 2022-04-16
Support GMSH/VTK prism/pyramid cells


tiagovla:
I noticed the cpp `io` interface does not yet support `GMSH/VTK` prism/pyramid cells which might be needed to support hybrid 3d meshes in the future. Could I make PR to contribute on that?

- [ ] Add logic to [cells.cpp](https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/io/cells.cpp)
- [ ] Add documentation to [cells.h](https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/io/cells.h)
- [ ] Add tests (maybe test fixtures)

jorgensd:
@tiagovla Feel free to add such a contribution to the code.
As far as I can tell by quickly looking at:
https://raw.githubusercontent.com/Kitware/vtk-examples/gh-pages/src/Testing/Baseline/Cxx/GeometricObjects/TestLinearCellDemo.png and  https://github.com/FEniCS/basix#prism
The map for a first order prism/wedge is the identity: [0,1,2,3,4,5]
while the map for a pyramid is : [0,1,2,3,2,4].
Feel free to try to add these for vtk and check the correctness by writing it to `pvd` (`VTKFile`).

garth-wells:
I'd suggest waiting on this until we have the low-level plumbing in place for mixed topologies.

PR #2427, opened 2022-11-02
Add pyramid and prism cells

tiagovla:
This PR should address #2132.

To do: 
- [ ] vtk/gmsh <-> dolfinx conversion
- [ ] tests


jhale:
Full marks for ASCII art skills!

tiagovla:
I couldn't find tests for the other cell types, so maybe those should be included for all cells in another PR.

PR #2428, opened 2022-11-02
Remove std::map from new non-matching interpolation feature

jhale:


garth-wells:
Open for a long time, so closing. Can be re-visited later.

PR #2429, opened 2022-11-02
Get cell dimension from basix

tiagovla:


jhale:
@garth-wells I'm inclined towards merging on the grounds that the Ciarlet Triple information should stored in basix.

PR #2430, opened 2022-11-02
Get number of entities of a cell from basix

tiagovla:


tiagovla:
[basix::cell::num_sub_entities](https://github.com/FEniCS/basix/blob/75e880990af503717f1b894f14376b57057721dd/cpp/basix/cell.cpp#L344-L349) should probably be optimized.

tiagovla:
[basix::cell::num_sub_entities](https://github.com/FEniCS/basix/blob/75e880990af503717f1b894f14376b57057721dd/cpp/basix/cell.cpp#L344-L349) should probably be optimized.

tiagovla:
> [basix::cell::num_sub_entities](https://github.com/FEniCS/basix/blob/75e880990af503717f1b894f14376b57057721dd/cpp/basix/cell.cpp#L344-L349) should probably be optimized.

This PR could be further simplified after https://github.com/FEniCS/basix/pull/616.

jhale:
@garth-wells We had a brief discussion about relying more heavily on basix definitions in DOLFINx. What's your thoughts on this? My view is that anything related to the Ciarlet Triple should be in basix.

PR #2431, opened 2022-11-02
Cmake-format support

tiagovla:


jorgensd:
Note that the demo cmake files are generated by https://github.com/FEniCS/dolfinx/blob/main/cpp/cmake/scripts/generate-cmakefiles.py
So these strings should be modified to adhere to cmake-format

jhale:
I quite like it. Would also be good to have this added as a non-blocking step in the CI.

garth-wells (dismissed):
See the comment in https://github.com/FEniCS/dolfinx/pull/2431#issuecomment-1301319271.

jhale:
Can you make sure that cmakelang gets put into the Dockerfile so it can be subsequently tidied up in the GitHub action after the merge (one day!).

jhale (approved):

PR #2457, opened 2022-11-20
Template `mesh::Geometry` over scalar type

garth-wells:
This is a first attempt at dealing with compatibility between mesh geometry scalar types and form type by templating `mesh::Geometry` over the scalar type. To get the performance benefits of reduced precision, we need the scalar types in a form to match, otherwise expensive casting is done behind the scenes. 

No guarantee that this is right approach, and if it is there is more work to handle the sharing of data. Putting it out there are this stage for comment. @IgorBaratta @jhale @chrisrichardson?


chrisrichardson:
I'm wondering if there is a way to do this without templating `Geometry` everywhere, if it is only really needed in the assembler.

garth-wells:
@chrisrichardson suggestions?

garth-wells:
@IgorBaratta?

garth-wells:
@jhale, @chrisrichardson, @IgorBaratta any further views? I'm coming around to this being the right approach.

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -23,10 +23,10 @@ namespace dolfinx::fem::impl
 
 /// Assemble functional over cells
 template <typename T>
-T assemble_cells(const mesh::Geometry& geometry,
-                 std::span<const std::int32_t> cells, FEkernel<T> auto fn,
-                 std::span<const T> constants, std::span<const T> coeffs,
-                 int cstride)
+T assemble_cells(
+    const mesh::Geometry<scalar_value_type_t<scalar_value_type_t<T>>>& geometry,
```
IgorBaratta, line 27:
```suggestion
    const mesh::Geometry<scalar_value_type_t<T>>& geometry,
```

Issue #2481, opened 2022-12-07
[BUG]: Infinite recursion when calling `dolfinx.fem.Constant.__float__()` 


mikics:
### How to reproduce the bug

When calling the `__float__` method of a `dolfinx.fem.Constant` object, an infinite recursion happens, since the `__float__` call eventually results in a `ufl.core.Terminal.evaluate()` call, which contains a reference to `__float__`, [see here](https://github.com/FEniCS/ufl/blob/4fd9a80f35ac1b24f7685fa08155ea02a20eea19/ufl/core/terminal.py#L52).

### Minimal Example (Python)

```Python
from mpi4py import MPI
from petsc4py import PETSc
from dolfinx import fem, mesh

domain = mesh.create_unit_square(
    MPI.COMM_WORLD, 1, 1, mesh.CellType.triangle)

m_constant = fem.Constant(domain, PETSc.RealType(0.0))
m_constant.__float__()
```

### Version

main branch

### DOLFINx git commit

3d4813a681b18ce1f6b594e3301ed08ccb1b1239

### Installation

Nightly ARM docker image (test-env)

PR #2482, opened 2022-12-12
Fix infinite recursion for `dolfinx.fem.Constant.__float__()` calls

mikics:
This PR may solve the infinite recursion related to [GH issue #2481](https://github.com/FEniCS/dolfinx/issues/2481) by adding `__int__`, `__float__` and `__complex__` methods directly to the `dolfinx.fem.Constant` class, instead of inheriting them from parents' classes.


jhale:
Thanks for this! A couple of small points

1. Can you add some test coverage? I suppose we never call these methods in our tests, so the issue was never picked up.
2. Is the `__int__` method necessary? Under what circumstances would it be called?

mikics:
Sure, I can add some tests!

For the `__int__` method: I checked my code, and indeed it is not needed to solve the bug. I will remove it.

mikics:
Tests added!

PR #2483, opened 2022-12-14
Downgrade PETSc version

jpdean:
Communicators don't seem to be getting cleaned up properly with PETSc 3.18.2, which causes the tests to fail simply by running them multiple times (see https://github.com/FEniCS/dolfinx/pull/2475). Downgrading to 3.17.5 seems to resolve this issue


jhale:
If you get this merged in before tonight then it will be re-built for tomorrow morning.

jorgensd:
> Communicators don't seem to be getting cleaned up properly with PETSc 3.18.2, which causes the tests to fail simply by running them multiple times (see #2475). Downgrading to 3.17.5 seems to resolve this issue

Is it worthwhile submitting an issue with PETSc on Gitlab.

In general the 3.18 release has other issues as well (https://gitlab.com/petsc/petsc/-/issues/1288)
which are not yet resolved: https://gitlab.com/petsc/petsc/-/merge_requests/5882

This causes issues for me when running CI on DOLFINx_mpc (https://github.com/jorgensd/dolfinx_mpc/actions/runs/3692863323/jobs/6252191437) as I want to test for both clang and gcc compiled versions of DOLFINx.

PR #2494, opened 2022-12-23
Fixes deprecation warning in random

jorgensd:
Linked to https://github.com/FEniCS/ufl/pull/139


garth-wells:
I think we should just remove the function in question. We can just make the timer name the name of the test function. No need for a random name.

PR #2500, opened 2023-01-06
Wrap C++ `Mesh` from Python (fixes support for Python 3.11)

### LINKED ISSUE #2423: Test on Python 3.11
Python 3.11 has been released: https://docs.python.org/3.11/whatsnew/3.11.html
We should probably add some test suite running over different versions of Python, and add it to the conda CI

garth-wells:
The difference between C++ and Python meshes is that Python meshes have `ufl_cell` and `ufl_domain` methods. The 'trick' used to handle mesh construction does not work with Python 3.11 (see https://github.com/FEniCS/dolfinx/issues/2423#issuecomment-1370847185).

This change wraps all `Mesh` instances in Python. This change is simple and fixes #2423, and will allow straightforward future support of different precision mesh geometry storage. It does, however, involve more hand-coding of methods and functions that operate of meshes. The design is similar to how `FunctionSpace` is handled on the Python side. The upside of the extra hand-coding is that docstring are easier to write.

### DIFF
--- a/python/demo/demo_lagrange_variants.py
+++ b/python/demo/demo_lagrange_variants.py
@@ -165,7 +165,7 @@ u_exact = saw_tooth(x[0])
 for variant in [basix.LagrangeVariant.equispaced, basix.LagrangeVariant.gll_warped]:
     element = basix.create_element(basix.ElementFamily.P, basix.CellType.interval, 10, variant)
     ufl_element = basix.ufl_wrapper.BasixElement(element)
-    V = fem.FunctionSpace(mesh, ufl_element)
+    V = fem.FunctionSpace(msh, ufl_element)
     uh = fem.Function(V)
     uh.interpolate(lambda x: saw_tooth(x[0]))
     if MPI.COMM_WORLD.size == 1:  # Skip this plotting in parallel
@@ -205,11 +205,11 @@ for variant in [basix.LagrangeVariant.equispaced, basix.LagrangeVariant.gll_warp
 for variant in [basix.LagrangeVariant.equispaced, basix.LagrangeVariant.gll_warped]:
     element = basix.create_element(basix.ElementFamily.P, basix.CellType.interval, 10, variant)
     ufl_element = basix.ufl_wrapper.BasixElement(element)
-    V = fem.FunctionSpace(mesh, ufl_element)
+    V = fem.FunctionSpace(msh, ufl_element)
     uh = fem.Function(V)
     uh.interpolate(lambda x: saw_tooth(x[0]))
     M = fem.form((u_exact - uh)**2 * dx)
-    error = mesh.comm.allreduce(fem.assemble_scalar(M), op=MPI.SUM)
+    error = msh.comm.allreduce(fem.assemble_scalar(M), op=MPI.SUM)
     print(f"Computed L2 interpolation error ({variant.name}):", error ** 0.5)
 # -
 
--- a/python/demo/demo_lagrange_variants.py
+++ b/python/demo/demo_lagrange_variants.py
@@ -163,15 +163,15 @@ def saw_tooth(x):
 # elements, and plot the finite element interpolation.
 
 # +
-mesh = mesh.create_unit_interval(MPI.COMM_WORLD, 10)
+msh = mesh.create_unit_interval(MPI.COMM_WORLD, 10)
 
-x = ufl.SpatialCoordinate(mesh)
+x = ufl.SpatialCoordinate(msh)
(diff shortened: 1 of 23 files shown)


python/demo/demo_lagrange_variants.py
```diff
@@ -163,15 +163,15 @@ def saw_tooth(x):
 # elements, and plot the finite element interpolation.
 
 # +
-mesh = mesh.create_unit_interval(MPI.COMM_WORLD, 10)
+msh = mesh.create_unit_interval(MPI.COMM_WORLD, 10)
```
jorgensd, line 166: I do personally not like the variable name `msh`, as it to me seems to like the variable to the GMSH MSH format (https://gmsh.info/doc/texinfo/gmsh.html#MSH-file-format). Could we instead use `domain`? Then usage in `ufl.Measure` would be, `ufl.Measure("dx", domain=domain, ...)`.

python/dolfinx/geometry.py
```diff
@@ -44,7 +43,15 @@ def __init__(self, mesh: Mesh, dim: int, entities=None, padding: float = 0.0):
         if entities is None:
             entities = range(0, map.size_local + map.num_ghosts)
 
-        super().__init__(mesh, dim, entities, padding)
+        super().__init__(mesh._mesh, dim, entities, padding)
+
+
+def compute_closest_entity(tree, midpoint_tree, mesh, points):
```
jorgensd, line 49: We should probably use typing and docstrings for these.

python/dolfinx/mesh.py
```diff
@@ -32,9 +30,16 @@
            "create_box", "create_unit_cube", "to_type", "to_string"]
 
 
-class Mesh(_cpp.mesh.Mesh):
-    def __init__(self, comm: _MPI.Comm, topology: _cpp.mesh.Topology,
-                 geometry: _cpp.mesh.Geometry, domain: ufl.Mesh):
+def compute_incident_entities(mesh, entities, d0: int, d1: int):
+    return _cpp.mesh.compute_incident_entities(mesh._mesh, entities, d0, d1)
```
jorgensd, line 34: Needs docs and type hints.

python/dolfinx/mesh.py
```diff
@@ -47,17 +52,31 @@ def __init__(self, comm: _MPI.Comm, topology: _cpp.mesh.Topology,
             Mesh objects are not generally created using this class directly.
 
         """
-        super().__init__(comm, topology, geometry)
+        # super().__init__(comm, topology, geometry)
+        self._mesh = mesh
```
francesco-ballarin, line 56: If I understand correctly, is this the C++ mesh object? If so, you may want to consider calling it `_cpp_object`, since  that convention is used several times elsewhere, e.g. in `python/dolfinx/fem/function.py` and `python/dolfinx/fem/dofmap.py`

garth-wells:
PR is at draft stage. No need for detailed comments yet, plenty to fix up first.

python/demo/demo_lagrange_variants.py
garth-wells: I think `msh` is the best compromise.

python/dolfinx/mesh.py
garth-wells: @francesco-ballarin yes, can do.

Caveat: I used `_cpp_object` at the very when moving DOFLIN-old to pybind11. Are we happy with the name?

francesco-ballarin: I personally am quite happy wit that name :)

garth-wells:
@jorgensd, @francesco-ballarin PR is ready now if you have more comments.

francesco-ballarin (approved):
Looks good to me!

python/test/unit/mesh/test_mesh.py
```diff
@@ -330,7 +330,7 @@ def test_facet_h(ct):
     N = 3
     mesh = create_unit_cube(MPI.COMM_WORLD, N, N, N, ct)
     left_facets = locate_entities_boundary(mesh, mesh.topology.dim - 1, lambda x: np.isclose(x[0], 0))
-    h = _cpp.mesh.h(mesh, mesh.topology.dim - 1, left_facets)
+    h = _cpp.mesh.h(mesh._cpp_object, mesh.topology.dim - 1, left_facets)
```
jorgensd, line 333: Maybe we should make a Pythonwrapper for cpp.mesh.h?

jorgensd (approved):
All Good. Only one note about an extra Python wrapper for dolfinx.cpp.mesh.h

PR #2501, opened 2023-01-13
Add `demo_axis` in `demos.rst`

mikics:
This PR adds the `demo_axis.py` from https://github.com/FEniCS/dolfinx/pull/2339 in the documentation.

### DIFF
--- a/python/doc/source/demos.rst
+++ b/python/doc/source/demos.rst
@@ -50,4 +50,5 @@ All demos
    demos/demo_scattering_boundary_conditions.md
    demos/demo_pml.md
    demos/demo_half_loaded_waveguide.md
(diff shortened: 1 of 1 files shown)


python/doc/source/demos.rst
```diff
@@ -50,4 +50,5 @@ All demos
    demos/demo_scattering_boundary_conditions.md
    demos/demo_pml.md
    demos/demo_half_loaded_waveguide.md
+   demos/demo_axis.py
```
IgorBaratta, line 53: isn't it `demo_axis.md`?

mikics: Oops, my bad! Will change it.

PR #2502, opened 2023-01-16
Add demo of mixed Poisson problem

jhale:
This PR adds a demo of the mixed Poisson problem, closely following the one from the old DOLFIN.

https://fenicsproject.org/olddocs/dolfin/1.5.0/python/demo/documented/mixed-poisson/python/documentation.html

The boundary condition on the H(div) space is a bit unsatisfactory and would require facet expressions for a proper 
solution on general meshes. The current approach is roughly how it is done in Firedrake.


garth-wells:
Be sure to check the rendered version - some of the markup doesn't look quite right.

PR #2503, opened 2023-01-18
Add exception subclass

### LINKED ISSUE #2398: Using exception subtypes
SonarCloud complains about our exception throwing, and I think it makes sense:

If you throw a general exception type, such as std::exception, std::logic_error or std::runtime_error, it forces consumers to catch all exceptions, including unknown exceptions they don’t necessarily know how to handle.

Instead, either throw a subtype that already exists ( for example in <stdexcept> ), or create your own type that derives from a standard one.

https://wiki.sei.cmu.edu/confluence/display/java/ERR07-J.+Do+not+throw+RuntimeException%2C+Exception%2C+or+Throwable

nils-imhoff:
"Resolves #2398  - Addresses SonarCloud complaint about exception throwing.
Instead of throwing general exception types (such as std::exception, std::logic_error, or std::runtime_error), this pull request modifies the code to throw a specific subtype that already exists, or a custom type that derives from a standard one. This allows for better handling of exceptions by consumers, as they will only catch exceptions they know how to handle."

### DIFF
--- a/python/doc/source/demos.rst
+++ b/python/doc/source/demos.rst
@@ -50,4 +50,5 @@ All demos
    demos/demo_scattering_boundary_conditions.md
    demos/demo_pml.md
    demos/demo_half_loaded_waveguide.md
+   demos/demo_axis.md
 


jhale:
I think this is a step in the right direction, I'm just wondering if it's necessary to go for quite so many exception classes. Is there anyway we could harmonise into a generic set of DOLFINx exception types that could be applied uniformly?

nils-imhoff:
> I think this is a step in the right direction, I'm just wondering if it's necessary to go for quite so many exception classes. Is there anyway we could harmonise into a generic set of DOLFINx exception types that could be applied uniformly?

Yes, this would be possible. Should I change it that way?

Issue #2312, opened 2022-08-03
Add a demo using an EnrichedElement


mscroggs:
Following https://github.com/FEniCS/basix/pull/574 and https://github.com/FEniCS/ffcx/pull/512, enriched elements will be supported. We should add a demo that uses these

juliusgh:
I have created a demo using the MINI element as an EnrichedElement in #2505. Is this close to what you had in mind?

mscroggs:
Perfect. I've attached this issue to that PR so this will be closed once its merged

PR #2505, opened 2023-01-18
Add demo of stokes with various stable pairs of finite elements

juliusgh:
There is already a demo of the Stokes equations for the lid driven cavity scenario with Taylor-Hood elements: https://docs.fenicsproject.org/dolfinx/main/python/demos/demo_stokes.html

Based on this, I have created a demo using various stable pairs of finite elements. So far these are:
1. $(\mathcal{P}_2, \mathcal{P}_1$): The [Taylor-Hood element](https://defelement.com/elements/taylor-hood.html) for $k=2$
2. $(\mathcal{P}_1 + \mathcal{B}_3, \mathcal{P}_1)$: The [MINI element](https://defelement.com/elements/mini.html)
3. $(\mathcal{P}_1^{\rm CR}, \mathcal{P}_0)$: The [non-conforming Crouzeix-Raviart element](https://defelement.com/elements/crouzeix-raviart.html)
4. $(\mathcal{P}_4, \mathcal{P}_3^{-1})$: The Scott-Vogelius element for $k=4$

With the MINI element, an enriched finite element is also used in a demo now. Such a demo was requested in https://github.com/FEniCS/dolfinx/issues/2312.
I think this demo would be helpful for beginners who want to try new finite elements.
Since this is my first contribution to dolfinx I would appreciate any feedback.


jhale:
Thanks! The Scott-Vogelius element is only (provably) int-sup stable for k = 4 ( https://doi.org/10.1051/m2an/1985190101111) so I would propose using a higher-order for the demo.

juliusgh:
Thank your for the feedback! I added the Scott-Vogelius element using the iterative procedure explained in the [FEniCS book](https://fenicsproject.org/pub/book/book/fenics-book-2011-06-14.pdf) on page 388. I will add proper documentation later.
I am not completely sure about the reconstruction of the pressure field. Is there a built-in `project` function in DOLFINx like in DOLFIN? I could not find it so far and used the code from https://github.com/michalhabera/dolfiny/blob/master/dolfiny/projection.py.
In the meantime, I also noticed that the Scott-Vogelius element is not stable for $k<4$, since the iteration does not converge in this case.

nate-sime:
Does this demo exhibit anything that's not already shown in other demos?

E.g. https://github.com/FEniCS/dolfinx/blob/main/python/demo/demo_lagrange_variants.py and https://github.com/FEniCS/dolfinx/blob/main/python/demo/demo_stokes.py?

nate-sime:
For every demo added, there's a burden of maintenance for the developers. This demo only shows element enrichment which has overloaded the `+` operator which at a glance is about 3 of 529 lines. I vote for incorporating this into one of the existing demos if it's _really_ required.

Otherwise I think the rest of the demo is an interesting report of various numerical methods which would be better hosted as part of a tutorial or personal portfolio.

garth-wells:
Remained open for some time, so closing.

PR #2508, opened 2023-01-20
Add biharmonic demo (Python and C++)

juliusgh:
This PR adds following demos that solve the biharmonic equation using a discontinuous Galerkin method:
- A Python demo for DOLFINx based on the [DOLFIN biharmonic Python demo](https://fenicsproject.org/olddocs/dolfin/latest/python/demos/biharmonic/demo_biharmonic.py.html)
- A C++ demo for DOLFINx based on the [DOLFIN biharmonic C++ demo](https://bitbucket.org/fenics-project/dolfin/src/master/demo/documented/biharmonic/cpp/main.cpp.rst)

As far as I know, DOLFINx does so far not include a demo using discontinuous Galerkin methods. In my opinion this would be helpful for new users of DOLFINx and users who want to migrate old code from legacy DOLFIN.


juliusgh:
The generated `CMakeLists.txt` file of the demo was ignored by `.gitignore`, but the build pipelines do not automatically generate the file using `python cmake/scripts/generate-cmakefiles`, causing the regression tests to fail. I have forced Git to add the `CMakeLists.txt` file. Is this a good practice or what would be the way to go here?

juliusgh:
Hi @jhale. Thank you for taking a look! In the meantime, there has been a small change in `fem::apply_lifting`. I have adjusted the demo and now the regression tests work again :)

PR #2509, opened 2023-01-20
Add ADIOS2 to macOS build

jhale:
Cannot use ADIOS2 from Homebrew.


garth-wells:
Why doesn't the Homebrew version work?

jhale:
> Why doesn't the Homebrew version work?

I had this (incorrect) idea it didn't work, but in fact homebrew adios2 works just fine.

PR #2510, opened 2023-01-20
add exception subclass 

nils-imhoff:
"Resolves https://github.com/FEniCS/dolfinx/issues/2398 - Addresses SonarCloud complaint about exception throwing.
Instead of throwing general exception types (such as std::exception, std::logic_error, or std::runtime_error), this pull request modifies the code to throw a custom DOLFINx exception type that derives from a standard one. This allows for better handling of exceptions by consumers, as they will only catch exceptions they know how to handle." It also addresses the [comment](https://github.com/FEniCS/dolfinx/pull/2503#issuecomment-1397048762) from @jhale


cpp/dolfinx/common/DolfinXException.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+#include <stdexcept>
+
+class DolfinXException : public std::runtime_error
```
jhale, line 5: I would propose a namespaced ``runtime_error`` exception, e.g.

``dolfinx::runtime_error``

I think this should be straightforward to do with search and replace.

cpp/dolfinx/common/DolfinXException.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
```
jhale, line 1: Rename the file `exception.h` for consistency with the standard library.

jhale:
My only more general question is how the new exception is transferred across to Python:

https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html

Do we need to implement a handler or will inheriting from std::runtime_error be enough?

garth-wells:
I think we need a careful discussion on what we want to achieve with our own exceptions, and how far we might want to go.

jhale:
The idea of this patch was to move us towards having that discussion.

My main criticism at moment we throw `runtime_error` everywhere, even if the error is potentially recoverable, e.g.:

https://github.com/FEniCS/dolfinx/pull/2510/files#diff-ef4a406ff8181d49ef090a0d232c037269dcd7e2e83760358ba4e091a529ad81R123

Having said that, I don't think we should go down the route of having many, many exceptions.

cpp/dolfinx/common/DolfinXException.h
nils-imhoff: I added the suggestion

cpp/dolfinx/common/DolfinXException.h
nils-imhoff: I added the suggestion

nils-imhoff:
> My only more general question is how the new exception is transferred across to Python:
> 
> https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html
> 
> Do we need to implement a handler or will inheriting from std::runtime_error be enough?

I think if you don't add a translation for the custom exception class dolfinx::runtime_error when calling C++ code from Python, the exception might not be handled properly by the Python interpreter.

So, I think it would be good.

jhale:
@garth-wells I would propose I rewrite this patch to simply change std::runtime_error to dolfinx::runtime_error so the error is properly namespaces. A more fine grained exception system would be another task.

garth-wells:
Lot's of conflicts, so closing. We can re-open a clean version.

PR #2513, opened 2023-01-23
Fix concept requirements

chrisrichardson:
Currently gcc-12 is showing warnings about "requires" statements, which this fixes.

### DIFF
--- a/cpp/dolfinx/fem/utils.h
+++ b/cpp/dolfinx/fem/utils.h
@@ -601,12 +601,13 @@ void pack(std::span<T> coeffs, std::int32_t cell, int bs, std::span<const T> v,
 /// @private
 /// @brief  Concepts for function that returns cell index
 template <typename F>
-concept FetchCells = requires(F&& f, std::span<const std::int32_t> v) {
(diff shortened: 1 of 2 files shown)


cpp/dolfinx/fem/utils.h
```diff
@@ -769,7 +770,8 @@ void pack_coefficients(const Form<T>& form, IntegralType integral_type, int id,
     {
     case IntegralType::cell:
     {
-      auto fetch_cell = [](auto& entity) { return entity.front(); };
+      auto fetch_cell
```
garth-wells, line 773: Are the changes removing `auto` intended?

chrisrichardson: It wasn't compiling. I think because `auto&` is wrong, but `auto` seems OK.

garth-wells: I'm not sure it's wrong.

Have you checked performance? This is a critical part of the code - is the pass by value an issue?

garth-wells: Did you test? It brings in a copy of the size at every call.

PR #1959, opened 2022-01-21
Add matrix-free demo (C++)

IgorBaratta:
Uses manufactured solution to check the error.

Issue #1776, opened 2021-10-25
Add matrix-free solver demos


garth-wells:
It would be helpful to add a matrix-free solver demo. Problem should include boundary conditions.

- [x] #1959 
- [ ] Python version

mapengfei-nwpu:
Will matrix-free methods preconditioned with multigrid method be considered in the future? Just like in Dealii https://dealii.org/developer/doxygen/deal.II/step_37.html

jhale:
Yes, putting in the basic infrastructure to do scalable geometric multigrid is on our roadmap.

garth-wells:
@IgorBaratta could a Python version be added?

IgorBaratta:
> @IgorBaratta could a Python version be added?

Yes, I can do that.
And once we have the interpolation operator (different meshes) we could also do matrix free geometric mutigrid.

PR #2517, opened 2023-01-30
Add Poisson matrix-free demo for Python

juliusgh:
This PR adds a Python version of the existing matrix-free solver demo in C++ #1959, which was requested in #1776.
The problem setting is the same as in the C++ demo. I was not sure what the best approach is, therefore the demo contains the following variants so far:
- Matrix-free CG solver using PETSc vectors (just as in the C++ demo)
- ~~Matrix-free CG solver using NumPy arrays~~
- Matrix-free CG solver using the built-in PETSc CG solver
- Matrix-free CG solver using a custom PETSc KSP solver

All of them reproduce the behavior of the C++ demo.

~~However, there is a problem with the MPI parallelization that I could not manage to solve yet~~. All variants also work in parallel like the C++ demo.


jhale:
Very nice to have these ideas ported to Python.

Is it possible to do an example using our built-in linear algebra objects?

Could you elaborate further on your point regarding MPI parallelisation?

juliusgh:
I also got the MPI parallelization to work. I had a wrong understanding of when exactly MPI communication is required. However, now the Python demo exactly reproduces the behavior of the C++ demo also in the parallel case.

python/demo/demo_poisson_matrix_free.py
```diff
@@ -0,0 +1,508 @@
...
+# #### 1. Implementation using PETSc vectors
+
+# To implement the matrix-free CG solver using *PETSc* vectors, we define the
+# function `action_A` with which the matrix-vector product $y = A x$
+# is computed.
+
+def action_A(x):
+    # Update coefficient ui of the linear form M
+    x.copy(ui.vector)
+    ui.vector.ghostUpdate(addv=PETSc.InsertMode.INSERT,
```
IgorBaratta, line 219: Can we use `dolfinx` vectors here?
Maybe it's too late for this change because it's already been merged.

garth-wells (commented):
With multi-part demos like this, I think it's useful to encapsulate the 'parts' in functions. With a 500 line flat script it can be hard to see what relates to what.

python/demo/demo_poisson_matrix_free.py
```diff
@@ -0,0 +1,508 @@
...
+                          mode=PETSc.ScatterMode.FORWARD)
+
+    # Compute action of A on x using the linear form M
+    y = fem.petsc.assemble_vector(fem.form(M))
```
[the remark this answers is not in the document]
juliusgh: Thanks for the feedback! It is sufficient to call `fem.form` only once. I have adapted the demo.

python/demo/demo_poisson_matrix_free.py
juliusgh: No, it's not too late :)
I replaced some `ghostUpdate` calls with `scatter_forward` where it is straight-forward and will try using `dolfinx` vectors everywhere later.

jhale:
I'm working on the native DOLFINx vector version on a separate branch, I will merge here once I'm done.

garth-wells:
@jhale any updates on the native DOLFINx implementation?

garth-wells:
No activity for some months, so closing.

This would be a good addition - happy to see PR re-opened once updated.

juliusgh:
@garth-wells @jhale If this PR would be beneficial with the PETSc implementation, I could adapt the existing version to the current main branch. But I don't know how to implement it using the native DOLFINx API.

IgorBaratta:
Hi @juliusgh , Appreciate your work, it's super useful. 
I'm planning to merge it into my branch and submit a new updated PR. 
Let me know if you have any objections.

juliusgh:
Hi @IgorBaratta. Thanks for the update, that sounds like a good plan! I'm happy if my work can be helpful.

PR #2520, opened 2023-02-02
Remove unnecessary call to NBX algorithm when creating sub index map

jpdean:


cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -481,15 +481,10 @@ IndexMap::create_submap(std::span<const std::int32_t> indices) const
...
-  std::sort(src.begin(), src.end());
-  src.erase(std::unique(src.begin(), src.end()), src.end());
-
-  // Determine destination ranks (ranks that ghost my indices), and sort
-  std::vector<int> dest
-      = dolfinx::MPI::compute_graph_edges_nbx(this->comm(), src);
-  std::sort(dest.begin(), dest.end());
+  // Get source ranks (ranks that own ghosts) and destination ranks
+  // (ranks that ghost my indices)
+  std::vector<int> src = this->src();
```
garth-wells, line 486: Could these be const references to avoid a copy?

jpdean: Whoops, yep!

garth-wells (approved):

PR #2526, opened 2023-02-10
Try BuildJet

jhale:
This replaces the AWS ARM runner with BuildJet instances. Advantages are infrastructure is managed elsewhere and parallel jobs are possible. Disadvantage is higher costs (~4x AWS). Would propose trying it for a month to get an idea of the cost impact.


jhale:
I'm going to merge this and try and get an idea of possible costs over a week or so.

PR #2547, opened 2023-02-23
Expose real version of inverse transform to python

jorgensd:
Needed for prototyping of checkpointing, ref: https://github.com/jorgensd/adios4dolfinx/pull/4


python/dolfinx/wrappers/fem.cpp
```diff
@@ -652,6 +652,16 @@ void fem(py::module& m)
                                           cell_permutation, dim);
           },
           py::arg("x"), py::arg("cell_permutation"), py::arg("dim"))
+      .def(
+          "apply_inverse_dof_transformation",
+          [](const dolfinx::fem::FiniteElement& self,
+             py::array_t<double, py::array::c_style>& x,
```
garth-wells, line 658: It's odd to pass by non-const reference here. It should be pass-by-value. Same for the wrapper above (I know the above hasn't been touched by the PR).

garth-wells (changes requested):

python/dolfinx/wrappers/fem.cpp
jorgensd: The whole point here is that we do not want to make copies of the input? We want to change the data in place, as this would be called inside hot loops (loop over every cell local to the process). Thus in C++ we send in a non-const span.

If we pass by value we copy the data, or am I missing something?

garth-wells: `py::array_t` doesn't necessarily hold the data. It should be holding a pointer. 

You might be able to pass by a const ref (depending how `mutable_data` works).

garth-wells (approved):

PR #2571, opened 2023-03-06
Attempt again at improved `setup.py` cmake build

garth-wells:


francesco-ballarin:
@garth-wells I tested this on my workflow, no issues there.

jhale:
Only cmake > 3.24 has the `--fresh` flag

https://cmake.org/cmake/help/latest/manual/cmake.1.html#cmdoption-cmake-fresh

Issue #2556, opened 2023-02-26
EM demos wrong because Gmsh-created meshes are by default not ghosted


garth-wells:
The EM demos (`demo_axis.py`, `demo_pml.py`, maybe more) compute wrong diagnostic quantities because some quantities involve interior facet integrals. Proper computation of interior facet integrals requires meshes to be ghosted by facet. But, `gmshio.model_to_mesh` by default does not use any ghosting.

We should really be able to tell whether or not the mesh has the required ghosting.

garth-wells:
@jorgensd thoughts?

jorgensd:
> @jorgensd thoughts?

I guess we should be able to add a check when generating the form that we have the appropriate ghosting (at least for dS integrals).

At least a quick fix for the demos is to set the Ghost mode in the partitioner input to `read_from_msh` and `model_to_mesh` https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/io/gmshio.py#L170 for the two demos

jorgensd:
@garth-wells  Added an error message that is hit if you run `demo_axis.py` pre this PR on two processes

PR #2574, opened 2023-03-07
Add ghost_mode shared facet to demos

jorgensd:
Resolves: https://github.com/FEniCS/dolfinx/issues/2556


garth-wells:
To fix the underlying bug I think we need a way to determine if we have the required ghost cells when executing a kernel over interior facets.

jorgensd:
@garth-wells Added an error message that is hit if you run demo_axis.py pre this PR on two processes

garth-wells:
Open for some time, so closing.

The runtime check added in this PR introduces non-trivial computation, so a different approach is needed.

PR #2610, opened 2023-04-02
Add default scalar type

garth-wells:
Avoids directly importing petsc4py just to get a scalar type.

### DIFF
--- a/python/demo/demo_biharmonic.py
+++ b/python/demo/demo_biharmonic.py
@@ -112,15 +112,14 @@
 
 # +
 import numpy as np
-
 import ufl
-from dolfinx import fem, io, mesh, plot
 from dolfinx.mesh import CellType, GhostMode
+from mpi4py import MPI
+from petsc4py.PETSc import ScalarType
 from ufl import (CellDiameter, FacetNormal, avg, div, dS, dx, grad, inner,
                  jump, pi, sin)
 
-from mpi4py import MPI
-from petsc4py.PETSc import ScalarType
+from dolfinx import fem, io, mesh, plot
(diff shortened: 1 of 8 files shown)


python/demo/demo_biharmonic.py
```diff
@@ -112,15 +112,14 @@
 
 # +
 import numpy as np
-
 import ufl
-from dolfinx import fem, io, mesh, plot
 from dolfinx.mesh import CellType, GhostMode
+from mpi4py import MPI
+from petsc4py.PETSc import ScalarType
```
francesco-ballarin, line 118: @garth-wells is there any reason why the new `default_scalar_type` is not used here?

python/demo/demo_elasticity.py
```diff
@@ -35,10 +33,11 @@
...
                           locate_entities_boundary)
-from ufl import dx, grad, inner
-
 from mpi4py import MPI
 from petsc4py import PETSc
+from ufl import dx, grad, inner
+
+from dolfinx import la
 
 dtype = PETSc.ScalarType
```
francesco-ballarin, line 42: and similarly here?

python/demo/demo_biharmonic.py
garth-wells: Logic was that these demos import `petsc4py` functionality via DOLFINx modules with petsc in the name, so it's not a problem to get the scalar type from `petsc4py`. I made the change in the main part of the code to avoid importing `petsc4py` at lower levels and where `petsc4py` is not used, e.g. in `forms.py`.

PR #2629, opened 2023-04-20
Change writeable flag after moving array from C++

### LINKED ISSUE #2625: [BUG]: `AttributeError: Can't set attribute` -- except I can
### How to reproduce the bug

I wanted to flip the sign of the values of a `MeshTags` object that I read from file.
In all my ignorance, I did `meshTags.values *= -1` but I got an error. The values still got changed though...

```
meshTags.values
Out[47]: array([-11, -11, -11, ..., -11, -11, -11], dtype=int32)

meshTags.values *= -1

  File "/tmp/ipykernel_94324/1239962826.py", line 1, in <module>
    meshTags.values *= -1

AttributeError: can't set attribute 'values'


meshTags.values
Out[49]: array([11, 11, 11, ..., 11, 11, 11], dtype=int32)
```

### Minimal Example (Python)

_No response_

### Output (Python)

_No response_

### Version

main branch

### DOLFINx git commit

_No response_

### Installation

I'm running in the latest docker image.

### Additional information

_No response_

jorgensd:
Resolves #2625.
Not the prettiest of fixes, but I cannot see a nice workaround for this.

### DIFF
--- a/.github/workflows/docker-dev-test-env.yml
+++ b/.github/workflows/docker-dev-test-env.yml
@@ -1,0 +1,167 @@
+---
+name: Docker build environment images
+
+# Builds build environment images (no FEniCS components). 
+#
+# Must be triggered manually via GitHub interface, possibly using specific git
+# refs and tag prefix for e.g. releases.
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag_prefix:
+        description: "tag prefix for docker images"
+        default: "current"
+        type: string
+        required: true
+
+      dockerfile:
+        description: "Dockerfile"
+        default: "docker/Dockerfile"
+        type: string
+        required: true
+
+jobs:
+  create_build_images:
+    name: Create build env images
+    if: ${{ github.repository == 'FEniCS/dolfinx' }}
+    strategy:
+      matrix:
+        variant: ["test-env", "dev-env"]
+        mpi: ["openmpi", "mpich"]
+        os: ["ubuntu-latest", "buildjet-4vcpu-ubuntu-2204-arm"]
+
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Set git ref
+        run: |
+          USER_INPUT=${{ github.event.inputs.dolfinx_ref }}
+          echo "DOLFINX_REF=${USER_INPUT:-main}" >> $GITHUB_ENV
+
+      - name: Set Dockerfile
+        run: |
+          USER_INPUT=${{ github.event.inputs.dockerfile }}
+          echo "DOCKERFILE=${USER_INPUT:-docker/Dockerfile}" >> $GITHUB_ENV
+
+      - name: Checkout DOLFINx
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ env.DOLFINX_REF }}
+
+      - name: Set root repository (docker.io/fenicsproject)
+        if: ${{ matrix.variant == 'test-env' }}
+        run: echo "DH_PREFIX=docker.io/fenicsproject" >> $GITHUB_ENV
+
+      - name: Set root repository (docker.io/dolfinx)
+        if: ${{ matrix.variant == 'dev-env' }}
+        run: echo "DH_PREFIX=docker.io/dolfinx" >> $GITHUB_ENV
+
+      - name: Set architecture tag (amd64)
+        if: ${{ matrix.os == 'ubuntu-latest' }}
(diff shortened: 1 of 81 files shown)


garth-wells:
See https://github.com/pybind/pybind11/pull/1466/.

jorgensd:
> See [pybind/pybind11#1466](https://github.com/pybind/pybind11/pull/1466).

I cannot get that to work with a `py::array_t` structure.
One could use the C++ equivalent of the current solution, suggested in:
https://github.com/pybind/pybind11/issues/481#issue-187301065
It seems like the buffer solution never made its way up to array_t:
https://github.com/pybind/pybind11/issues/338#issuecomment-263565163

jorgensd:
> > See [pybind/pybind11#1466](https://github.com/pybind/pybind11/pull/1466).
> 
> I cannot get that to work with a `py::array_t` structure. One could use the C++ equivalent of the current solution, suggested in: [pybind/pybind11#481 (comment)](https://github.com/pybind/pybind11/issues/481#issue-187301065) It seems like the buffer solution never made its way up to array_t: [pybind/pybind11#338 (comment)](https://github.com/pybind/pybind11/issues/338#issuecomment-263565163)

@michalhabera suggested the new approach, which is influenced by the C++ assign flag comment

michalhabera:
> This should be wrapped up in a helper function, otherwise we'll have low-level code everywhere.

@jorgensd I think Garth wrote wrappers/array.h for this purpose. My suggested approach might be still flawed wrt. lifetime management, `py::cast(self)` passed as a base also made lifetime of returned numpy array coupled with the MeshTags cpp object.

jorgensd:
> > This should be wrapped up in a helper function, otherwise we'll have low-level code everywhere.
> 
> @jorgensd I think Garth wrote wrappers/array.h for this purpose. My suggested approach might be still flawed wrt. lifetime management, `py::cast(self)` passed as a base also made lifetime of returned numpy array coupled with the MeshTags cpp object.

How about:
```c++
        py::array_t<T> array(self.values().size(), self.values().data(),
                                 py::cast(self));
            reinterpret_cast<py::detail::PyArray_Proxy*>(array.ptr())->flags
                &= ~py::detail::npy_api::NPY_ARRAY_WRITEABLE_;
            return array;
```
I guess we could add this to the `as_pyarray` function. Just need to deduce if the arguments should be readonly or not

jorgensd:
Not if you look at the C++ signature, which returns a const span https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/mesh/MeshTags.h#L98

garth-wells:
> Not if you look at the C++ signature, which returns a const span https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/mesh/MeshTags.h#L98

For the `indices` yes, but not for the `values`. The PR prevents users from changing the values.

Python doesn't generally have the concept of `const`, despite NumPy arrays having a 'writeable' flag. The philosophy is to trust users. What is important here is that that the `owndata` flag is correctly set since changing size would cause a corruption of the `MeshTags` object.

garth-wells:
> Not if you look at the C++ signature, which returns a const span https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/mesh/MeshTags.h#L98

Indeed, my mistake.

Still, I don't think it's unreasonable to be able change the values from Python, even if it risks breaking something else.

jorgensd:
Im in favor of avoiding alot of unexpected behavior with the interaction with dolfinx::fem::Form, which has major assumptions on the const-ness of integration entities specified in MeshTags. As I mentioned in an earlier commit, we could also force this purely on the Python side (inside the property functions).

jhale:
This should really be fixed at the pybind11 level.

Having said that, I fall on the side that the numpy array should be set as non-writeable, reflecting the underlying const specification in the C++ span. I don't have a strong view on whether it should be done at the C++ level, or in the Python wrapper.

jorgensd:
I agree with that, and I am happy with setting it in the pybinding, as done in this PR currently.

garth-wells:
I'm not too fussed either way. In any case, a helper function should be added in `array.h` rather than adding the low-level code to each wrapper function.

Some lifetime/reference counting tests are need too since we're not 100% sure what is happing behind the scenes with pybind11 and what this PR does.

garth-wells:
Note: `nanobind` handles `const` and NumPy array.

garth-wells:
This will be addressed when we switch to nanobind, see #2820.

Issue #2624, opened 2023-04-19
`AdjacencyList` offset array can overflow


chrisrichardson:
### How to reproduce the bug

Running a problem of sufficient size on one process (e.g. Q2 problem on mesh with 60 million dofs) can overwhelm the range of `AdjacencyList`.
* When running e.g. on a GPU system, we may have many degrees of freedom on one process.
* If the number of non-zeros in the local sparsity pattern exceeds 2^31, the AdjacencyList offsets used when building the sparsity pattern will overflow.

Possible solution: use 64-bit indexing for offsets in `AdjacencyList` or template it.

### Version

main branch

jorgensd:
Oh dear.
I would say change it to int64, as I think we have introduce too many templates already (and I do not see a large benefit of this being templated).

jorgensd:
> I'm going to think about other solutions: we only really need it in a few places (sparsity pattern and MatrixCSR) so I don't think we want to make it int64 everywhere...

Then I guess we want to template it, but maybe make some wrappers already at the C++ level to avoid everything being moved into header files.

chrisrichardson:
I think I'm going to close this. On reflection, it is only useful when we have >50M dofs, and the associated matrix will occupy around 64Gb of memory, which is close to the limit, at the moment. We should however, have some kind of test to make sure we are not overflowing when creating sparsity patterns or matrices.

garth-wells:
This should be fixed at some point. The simplest approach might be (1) avoid AdjacencyList when we don't really need it (see #2632), and then (2) store the offsets in a `std::vector<std::size_t>`.

PR #2632, opened 2023-04-20
Flatten dofmap storage

garth-wells:
This change stores dofmap data as flat vectors rather than AdjacencyLists, and uses `mdspan` for convenient access. This uses less memory that AdjacencyLists and data access is more performance friendly.

Assumption is that all elements in a DofMap have the same number io dofs. This is presently that case, and for more complex cases in the future, e.g. mixed cell types, each element type will have its own DofMap. This follows the design that has been started for mixed topology, where cells are stored by type, and operations like assembly will be per cell type/kernel.

It's useful to make this change now as unpicking the code will be a lot harder once mixed topology support is further developed. Moreover, removing necessary use of `AdjacencyList` makes considerations dealing with  #2624 simpler as the impact of any change is reduced.

### DIFF
--- a/cpp/dolfinx/fem/DofMap.cpp
+++ b/cpp/dolfinx/fem/DofMap.cpp
@@ -42,7 +42,10 @@ fem::DofMap build_collapsed_dofmap(const DofMap& dofmap_view,
   assert(cells);
 
   // Build set of dofs that are in the new dofmap (un-blocked)
-  std::vector<std::int32_t> dofs_view = dofmap_view.list().array();
+  auto dofs_view_md = dofmap_view.list();
+  std::vector<std::int32_t> dofs_view(dofs_view_md.data_handle(),
+                                      dofs_view_md.data_handle()
+                                          + dofs_view_md.size());
   dolfinx::radix_sort(std::span(dofs_view));
   dofs_view.erase(std::unique(dofs_view.begin(), dofs_view.end()),
                   dofs_view.end());
@@ -104,10 +107,13 @@ fem::DofMap build_collapsed_dofmap(const DofMap& dofmap_view,
   }
 
   // Map dofs to new collapsed indices for new dofmap
-  const std::vector<std::int32_t>& dof_array_view = dofmap_view.list().array();
+  // const std::vector<std::int32_t>& dof_array_view =
+  // dofmap_view.list().array();
+  auto dof_array_view = dofmap_view.list();
   std::vector<std::int32_t> dofmap;
   dofmap.reserve(dof_array_view.size());
-  std::transform(dof_array_view.begin(), dof_array_view.end(),
+  std::transform(dof_array_view.data_handle(),
+                 dof_array_view.data_handle() + dof_array_view.size(),
                  std::back_inserter(dofmap),
                  [&old_to_new](auto idx_old) { return old_to_new[idx_old]; });
 
@@ -130,20 +136,22 @@ fem::DofMap build_collapsed_dofmap(const DofMap& dofmap_view,
 } // namespace
 
 //-----------------------------------------------------------------------------
-graph::AdjacencyList<std::int32_t>
-fem::transpose_dofmap(const graph::AdjacencyList<std::int32_t>& dofmap,
-                      std::int32_t num_cells)
+graph::AdjacencyList<std::int32_t> fem::transpose_dofmap(
+    std::experimental::mdspan<const std::int32_t,
+                              std::experimental::dextents<std::size_t, 2>>
+        dofmap,
+    std::int32_t num_cells)
 {
   // Count number of cell contributions to each global index
   const std::int32_t max_index = *std::max_element(
-      dofmap.array().begin(),
-      std::next(dofmap.array().begin(), dofmap.offsets()[num_cells]));
+      dofmap.data_handle(),
+      std::next(dofmap.data_handle(), num_cells * dofmap.extent(1)));
 
   std::vector<int> num_local_contributions(max_index + 1, 0);
-  for (int c = 0; c < num_cells; ++c)
+  for (std::int32_t c = 0; c < num_cells; ++c)
   {
-    for (auto dof : dofmap.links(c))
-      num_local_contributions[dof]++;
+    for (std::size_t d = 0; d < dofmap.extent(1); ++d)
+      num_local_contributions[dofmap(c, d)]++;
   }
 
   // Compute offset for each global index
@@ -154,10 +162,10 @@ fem::transpose_dofmap(const graph::AdjacencyList<std::int32_t>& dofmap,
   std::vector<std::int32_t> data(index_offsets.back());
   std::vector<int> pos = index_offsets;
   int cell_offset = 0;
-  for (int c = 0; c < num_cells; ++c)
+  for (std::int32_t c = 0; c < num_cells; ++c)
   {
-    for (auto dof : dofmap.links(c))
-      data[pos[dof]++] = cell_offset++;
+    for (std::size_t d = 0; d < dofmap.extent(1); ++d)
+      data[pos[dofmap(c, d)]++] = cell_offset++;
   }
 
   // Sort the source indices for each global index
@@ -172,7 +180,6 @@ fem::transpose_dofmap(const graph::AdjacencyList<std::int32_t>& dofmap,
   return graph::AdjacencyList(std::move(data), std::move(index_offsets));
 }
 //-----------------------------------------------------------------------------
-/// Equality operator
 bool DofMap::operator==(const DofMap& map) const
 {
   return this->_index_map_bs == map._index_map_bs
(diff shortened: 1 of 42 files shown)


cpp/dolfinx/fem/DofMap.cpp
```diff
@@ -154,10 +161,11 @@ fem::transpose_dofmap(const graph::AdjacencyList<std::int32_t>& dofmap,
...
   std::vector<int> pos = index_offsets;
   int cell_offset = 0;
-  for (int c = 0; c < num_cells; ++c)
+  for (std::int32_t c = 0; c < num_cells; ++c)
   {
-    for (auto dof : dofmap.links(c))
-      data[pos[dof]++] = cell_offset++;
+    auto dofs = stdex::submdspan(dofmap, c, stdex::full_extent);
+    for (std::size_t d = 0; d < dofmap.extent(1); ++d)
+      data[pos[dofs[d]]++] = cell_offset++;
```
jorgensd, line 168: Similar comment as above, as we never use anything but `dofs[d]` (never `d` itself).

cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -231,20 +240,8 @@ build_basic_dofmap(
...
-  // Go through elements twice: regular cells followed by ghost cells.
-  for (std::size_t i = 0; i < 2 * nelem; ++i)
-  {
-    // Number of dofs per cell for this element layout
-    const int local_dim = element_dof_layouts[i % nelem].num_dofs();
-    for (int j = group_offsets[i]; j < group_offsets[i + 1]; ++j)
-      cell_ptr.push_back(cell_ptr.back() + local_dim);
-  }
-  std::vector<std::int32_t> dofs(cell_ptr.back());
+  std::vector<std::int32_t> dofs(num_cells * element_dof_layouts[0].num_dofs());
```
jorgensd, line 244: Is this correct?

Above (Line 177-187) we verify that the number of dofs in each layout is the same for d=0,..<tdim, but the number of dofs in the whole layout 0 can be different from the one in layout one, as the number of dofs owned by cell can differ.

cpp/dolfinx/io/vtk_utils.cpp
```diff
@@ -20,18 +20,21 @@ using namespace dolfinx;
...
+    std::experimental::mdspan<const std::int32_t,
+                              std::experimental::dextents<std::size_t, 2>>
+        dofmap_x,
+    mesh::CellType cell_type)
 {
   // Get DOLFINx to VTK permutation
   // FIXME: Use better way to get number of nodes
   // const std::size_t num_nodes = geometry.cmap().dim();
-  const std::size_t num_nodes = dofmap_x.links(0).size();
+  const std::size_t num_nodes = dofmap_x.extent(1);
```
jorgensd, line 32: I guess you now can remove lines 30-31, as this would be a correct way of getting the number of nodes.

cpp/dolfinx/refinement/plaza.h
```diff
@@ -193,7 +195,8 @@ face_long_edge(const mesh::Mesh<T>& mesh)
     assert(it1 != cell_vertices.end());
     const std::size_t local1 = std::distance(cell_vertices.begin(), it1);
 
-    auto x_dofs = x_dofmap.links(cells.front());
+    // auto x_dofs = x_dofmap.links(cells.front());
```
jorgensd, line 198: Remove commented code, remainder from debug?

python/dolfinx/wrappers/fem.cpp
```diff
@@ -775,8 +793,17 @@ void fem(py::module& m)
...
-      .def("list", &dolfinx::fem::DofMap::list,
-           py::return_value_policy::reference_internal);
+      .def(
+          "list",
+          [](const dolfinx::fem::DofMap& self)
+          {
+            auto dofs = self.list();
+            std::array<py::ssize_t, 2> shape
+                = {py::ssize_t(dofs.extent(0)), py::ssize_t(dofs.extent(1))};
+            return py::array_t<std::int32_t>(shape, dofs.data_handle(),
```
jorgensd, line 803: Shouldn't this be const, as discussed in https://github.com/FEniCS/dolfinx/pull/2629 ?

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -144,7 +144,15 @@ void declare_mesh(py::module& m, std::string type)
...
-      .def_property_readonly("dofmap", &dolfinx::mesh::Geometry<T>::dofmap)
+      .def_property_readonly(
+          "dofmap",
+          [](dolfinx::mesh::Geometry<T>& self)
+          {
+            auto dofs = self.dofmap();
+            std::array<std::size_t, 2> shape = {dofs.extent(0), dofs.extent(1)};
+            return py::array_t<std::int32_t>(shape, dofs.data_handle(),
+                                             py::cast(self));
+          })
```
jorgensd, line 155: Same comment as in fem.cpp

python/test/unit/fem/test_assembler.py
```diff
@@ -499,7 +499,7 @@ def monolithic():
     create_unit_square(MPI.COMM_WORLD, 12, 11, ghost_mode=GhostMode.shared_facet),
     create_unit_cube(MPI.COMM_WORLD, 3, 7, 3, ghost_mode=GhostMode.none),
     create_unit_cube(MPI.COMM_WORLD, 3, 7, 3, ghost_mode=GhostMode.shared_facet)])
-def test_assembly_solve_taylor_hood(mesh):
+def xtest_assembly_solve_taylor_hood(mesh):
```
jorgensd, line 502: why is this now xtest?

jorgensd (dismissed):
In general Im in favor of this change, as it would simplify libraries such as dolfinx_mpc and asimov_contact.
See comments for minor things.

cpp/dolfinx/fem/DofMap.h
```diff
@@ -85,12 +88,14 @@ class DofMap
...
-            std::convertible_to<graph::AdjacencyList<std::int32_t>> U>
+            std::convertible_to<std::vector<std::int32_t>> U>
   DofMap(E&& element, std::shared_ptr<const common::IndexMap> index_map,
          int index_map_bs, U&& dofmap, int bs)
       : index_map(index_map), _index_map_bs(index_map_bs),
         _element_dof_layout(std::forward<E>(element)),
-        _dofmap(std::forward<U>(dofmap)), _bs(bs)
+        _dofmap(std::forward<U>(dofmap)), _bs(bs),
+        _shape1(_element_dof_layout.num_dofs()
+                * _element_dof_layout.block_size() / _bs)
```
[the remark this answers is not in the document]
garth-wells: This isn't the index map block size. It handles cases like the velocity subspace of a Taylor-Hood element where the `_element_dof_layout. block_size == 2` (in 2D) and `_bs == 1`. The velocity is a 'blocked' element, but the dofmap has no block structure since it is a view into a P2 x P1 element.

cpp/dolfinx/fem/DofMap.cpp
```diff
@@ -122,28 +127,30 @@ fem::DofMap build_collapsed_dofmap(const DofMap& dofmap_view,
...
 
   std::vector<int> num_local_contributions(max_index + 1, 0);
-  for (int c = 0; c < num_cells; ++c)
+  for (std::int32_t c = 0; c < num_cells; ++c)
   {
-    for (auto dof : dofmap.links(c))
-      num_local_contributions[dof]++;
+    auto dofs = stdex::submdspan(dofmap, c, stdex::full_extent);
+    for (std::size_t d = 0; d < dofmap.extent(1); ++d)
+      num_local_contributions[dofs[d]]++;
```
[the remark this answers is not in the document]
garth-wells: `mdspan` doesn't have `begin`/`end` (because it could be strided).

cpp/dolfinx/fem/DofMap.cpp
garth-wells: `mdspan` doesn't have `begin`/`end` (because it could be strided).

cpp/dolfinx/fem/dofmapbuilder.cpp
garth-wells: None of the code for multiple layouts works yet.

python/dolfinx/wrappers/fem.cpp
garth-wells: Maybe, #2629 doesn't offer a solution yet.

jorgensd: What do you think that PR is missing?

francesco-ballarin:
@garth-wells I am reusing as much as I can from DOLFINx assembly, with a few minimal changes in matrix/vector allocation that I keep in sync with the upstream repo. The first place in which I swap out a standard dolfinx dofmap for a `graph::AdjacencyList<std::int32_t>` is
https://github.com/FEniCS/dolfinx/compare/main...francesco-ballarin:dolfinx:pr1
with the goal of allowing to use my own dofmap (which does not inherit from dolfinx one).

I guess I can still swap that out for a pair containing: as first item, the `std::vector` of the flattened dof map, and as second item a vector of offsets. 
In my implementation I basically have them already
https://github.com/multiphenics/multiphenicsx/blob/main/multiphenicsx/cpp/multiphenicsx/fem/DofMapRestriction.cpp#L258
and I just need to store them (rather than providing them to an adjacency list). Still, since I want my assembly functions to be working also with the standard dolfinx dofmap (rather than my own), I will try to work out a way on how to get that same pair from the new ``DofMap::list()`` method, where the first item of the pair should basically be the underlying `std::vector` which is stored as private member `DofMap::_dofmap`, and the second item of the pair a ``std::iota`` built from `DofMap::_shape1`, i.e. the second extent of  ``DofMap::list()``

python/dolfinx/wrappers/fem.cpp
garth-wells: > What do you think that PR is missing?

You changed it ;).

francesco-ballarin:
> I guess I can still swap that out for a pair containing: as first item, the std::vector of the flattened dof map, and as second item a vector of offsets.

I confirm I have proceeded in this way, and restored `multiphenicsx:main` compatibility with `dolfinx:main` :thumbsup:

PR #2639, opened 2023-04-25
Use container instead of allocator in `dolfinx::la::Vector`

chrisrichardson:


garth-wells:
Could we use type traits to support:
```cpp
la::Vector<double>;
la::Vector<std::vector<double>;
```
?

PR #2642, opened 2023-04-28
Split Docker builds

jhale:
Should be able to fix unreliable/broken image caching and reduce build times


garth-wells (approved):
We could later add workflow files to build the Redhat, oneAPI, etc, images too.

jhale:
Yes, that would be useful, clicking buttons is so much easier. I have to merge this to test it properly, so there will probably be another PR.

PR #2643, opened 2023-05-05
Remove "needs" from the first step of the end-user Docker images workflow, since now build images are prepared in a standalone workflow

francesco-ballarin:
Fixes #2642. Should resume end-user docker workflow, which has been broken for a week.

### DIFF
--- a/.github/workflows/docker-end-user.yml
+++ b/.github/workflows/docker-end-user.yml
@@ -17,7 +17,7 @@ on:
         default: "nightly"
         type: string
         required: true
-      
+
       baseimage:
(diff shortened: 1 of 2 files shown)


garth-wells (approved):

francesco-ballarin:
@jhale unfortunately that wasn't enough to restore a working end-user docker workflow. Failure is now

ERROR: failed to solve: rpc error: code = Unknown desc = failed to solve with frontend dockerfile.v0: failed to create LLB definition: docker.io/dolfinx/dev-env:current: not found

but I don't know how to fix that. Can you have a look?

PR #2650, opened 2023-05-08
Adjust API to be more consistent

chrisrichardson:
* use `squared_norm()` for both `Vector` and `MatrixCSR`
* Use `InsertMode` not `ScatterMode`
* Use `finalize` for both `MatrixCSR` and `SparsityPattern`
* `index_map()` not `map()` for `Vector` 

If any others seem obvious, we can add to this branch.

### DIFF
--- a/cpp/dolfinx/fem/dofmapbuilder.cpp
+++ b/cpp/dolfinx/fem/dofmapbuilder.cpp
@@ -547,8 +547,9 @@ std::pair<std::vector<std::int64_t>, std::vector<int>> get_global_indices(
 
       // Number and values to send and receive
       const int num_indices = global[d].size();
-      std::vector<int> size_recv(src.size());
-      size_recv.reserve(1);
+      std::vector<int> size_recv;
+      size_recv.reserve(1); // ensure data is not a nullptr
+      size_recv.resize(src.size());
       MPI_Neighbor_allgather(&num_indices, 1, MPI_INT, size_recv.data(), 1,
                              MPI_INT, comm[d]);
 
--- a/python/dolfinx/wrappers/la.cpp
+++ b/python/dolfinx/wrappers/la.cpp
@@ -26,8 +26,8 @@ namespace py = pybind11;
 
 namespace
 {
-// ScatterMode types
-enum class PyScatterMode
+// InsertMode types
+enum class PyInsertMode
 {
   add,
   insert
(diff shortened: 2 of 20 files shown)


cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -547,8 +547,9 @@ std::pair<std::vector<std::int64_t>, std::vector<int>> get_global_indices(
 
       // Number and values to send and receive
       const int num_indices = global[d].size();
-      std::vector<int> size_recv(src.size());
-      size_recv.reserve(1);
+      std::vector<int> size_recv;
+      size_recv.reserve(1); // ensure data is not a nullptr
```
garth-wells, line 551: The order of this line and the one below should be switched.

python/dolfinx/wrappers/la.cpp
```diff
@@ -26,8 +26,8 @@ namespace py = pybind11;
 
 namespace
 {
-// ScatterMode types
-enum class PyScatterMode
+// InsertMode types
+enum class PyInsertMode
```
garth-wells, line 30: I don't think this is clearer. What's being controlled is what happens during a scatter. Moreover, the previous naming had a consistency with PETSc, see https://petsc.org/release/manualpages/Vec/VecGhostUpdateBegin/.

The name change could be confused with switching between insert and set during assembly.

IgorBaratta:
> assemble not finalize for MatrixCSR (similar to PETSc and we also use assemble for SparsityPattern)

I still prefer `finalize` over `assemble`. Or maybe a more descriptive name like `scatter_values()`?

python/dolfinx/wrappers/la.cpp
chrisrichardson: I think it was inconsistent with PETSc before, and is more inline now: PETSc has a scattermode (forward or reverse).

garth-wells:
> > > assemble not finalize for MatrixCSR (similar to PETSc and we also use assemble for SparsityPattern)
> > 
> > 
> > I still prefer `finalize` over `assemble`. Or maybe a more descriptive name like `scatter_values()`?
> 
> Yes, I think so too. I can change it back, unless anyone has a better idea - I'd like to change `SparsityPattern` to also use `finalize()` in which case...

Not crazy about `finalize` for the matrix :). Sounds as if it can't be re-used. Happy with it for `SparsityPattern`.

chrisrichardson:
In which case, the other alternative is some variant of `scatter_foo*`. Basically this operation is the equivalent operation to `scatter_reverse` that we do for `Vector`, except for matrix rows.
Could also be `gather_rows()` ...

PR #2653, opened 2023-05-11
Add IndexMap imbalance "diagnostic" function

IgorBaratta:
This PR adds a new `imbalance()` function to the `IndexMap` class. 
The imbalance is defined as the maximum number of indices on any process divided by the average number of indices per process, and is computed separately for owned and ghost indices. 

Usage example:
```python3
import ufl
import dolfinx
from dolfinx import fem
import numpy as np
from mpi4py import MPI

comm = MPI.COMM_WORLD

# Create a mesh and function space.
mesh = dolfinx.mesh.create_unit_cube(comm, 10, 10, 10,
                                     cell_type=dolfinx.mesh.CellType.tetrahedron,
                                     ghost_mode=dolfinx.mesh.GhostMode.none)
V = fem.FunctionSpace(mesh, ("Lagrange", 5))
index_map = V.dofmap.index_map
tdim = mesh.topology.dim

dof_imbalance = index_map.imbalance
cell_imbalance = mesh.topology.index_map(tdim).imbalance

if (comm.rank == 0):
    print("dof imbalance", dof_imbalance)
    print("cell imbalance", cell_imbalance)
```
```
dof imbalance [1.0140896035461473, 1.3187465232709068]
cell imbalance [1.006, -1.0]
```

### DIFF
--- a/cpp/dolfinx/common/IndexMap.h
+++ b/cpp/dolfinx/common/IndexMap.h
@@ -240,7 +240,7 @@ public:
   ///
   /// @return An array containing the imbalance in owned indices
   /// (first element) and the imbalance in ghost indices (second element).
-  std::array<double, 2> imbalance() const noexcept;
+  std::array<double, 2> imbalance() const;
 
 private:
   // Range of indices (global) owned by this process
--- a/cpp/dolfinx/common/IndexMap.h
+++ b/cpp/dolfinx/common/IndexMap.h
@@ -228,6 +228,20 @@ public:
   /// false.
   bool overlapped() const noexcept;
 
+  /// @brief Returns the imbalance of the current IndexMap.
+  ///
+  /// The imbalance is a measure of load balancing across all processes, defined
+  /// as the maximum number of indices on any process divided by the average
+  /// number of indices per process. This function calculates the imbalance
+  /// separately for owned indices and ghost indices and returns them as a
+  /// std::array<double, 2>.
+  ///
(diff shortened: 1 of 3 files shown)


jorgensd (approved):

chrisrichardson:
So, perfect balance = 1? And it can never be less than 1. I am not sure about the -1, maybe this should also just be 1 - because zero everywhere is also perfectly balanced?

And then, why not just subtract the 1? In the example above, we are mostly interested that one process has 31.8% more ghost dofs than average... 0.318 would be the "imbalance", 0.0 perfect balance...

cpp/dolfinx/common/IndexMap.h
```diff
@@ -228,6 +228,22 @@ class IndexMap
...
 
+  /// @brief Returns the imbalance of the current IndexMap.
+  ///
+  /// The imbalance is a measure of load balancing across all processes, defined
+  /// as the maximum number of indices on any process divided by the average
+  /// number of indices per process. This function calculates the imbalance
+  /// separately for owned indices and ghost indices and returns them as a
+  /// std::array<double, 2>. If the total number of owned or ghost indices is
+  /// zero, the respective entry in the array is set to -1.
+  ///
```
chrisrichardson, line 239: It would be useful to say that "1.0" represents a "perfect" balance.

jhale:
This seems useful, and is non-trivial to implement. Thanks!

Issue #2628, opened 2023-04-20
Split pointers for diagonal and off-diagonal parts of MatrixCSR


chrisrichardson:
For better compatibility with 3rd party libraries, it would be better to have the diagonal and off-diagonal blocks of the sparse matrix in `MatrixCSR` stored separately. The data structures would be the same, just two instead of one, and the storage should not be any more.

garth-wells:
@chrisrichardson has this been completed?

chrisrichardson:
No. I am also not convinced it is always the right solution. Some solvers use a unified format, some split. I have got a working prototype (split format) so maybe we can work on having both available  via an option or subclassing.

chrisrichardson:
https://github.com/FEniCS/dolfinx/tree/chris/matrixcsr-bs-experiment

garth-wells:
Splitting the matrix would make wrapping sparse matrices locally using SciPy more difficult.

chrisrichardson:
I agree. I think we should consider supporting both formats internally.

jorgensd:
@chrisrichardson  is this now resolved with: https://github.com/FEniCS/dolfinx/blob/f6a4a5a3de45003a38d3b8d3c667a1d1cf14adf9/cpp/dolfinx/la/MatrixCSR.h#L343-L359 ?

chrisrichardson:
I am revisiting this whilst looking into more MatrixCSR operations (SpGEMM and transpose). I think splitting into diagonal and off-diagonal blocks *is* the right approach. Scipy is not used greatly in parallel and anyway, can be accommodated with two operations (diag and offdiag) separately.
Having separation will allow e.g. transposition of local diag block whilst using non-blocking MPI on offdiag...

PR #2654, opened 2023-05-14
Add Block support to MatrixCSR

chrisrichardson:
Adds block size support to `MatrixCSR<T>`.

* Block size of the matrix is defined by the block size of the `SparsityPattern` used to construct it.
* Internal storage within each block is row major.
* The option `BlockMode=expanded` allows the sparsity pattern to be expanded to create a matrix with `bs=1` but the appropriate sparsity to accept block entries.
* Three `mat_insert` routines are provided for "set" and "add":
    - insert data with the same block size as the matrix (data_bs = matrix_bs)
    - insert data with block size 1 (scalar) into a matrix with a block size > 1 (data_bs=1)
    - insert data with block size > 1 into a matrix with block size 1 (matrix_bs = 1).
* Insertion routines are templated for performance
* Python interface allows insertion with data block size `(1, 1)`, `(2, 2)` or `(3, 3)`.
* Example of non-square blocks with `cppimport` in test suite.
* Separation of implementation into `matrix_csr.impl.h`

Storage is currently in a single data structure, but this can be revised, either to use two arrays ("on-diagonal", "off-diagonal" blocks), (see Issue #2628) or to move ghost columns to the end of the existing structure. Neither should affect the user interface for assembly.


garth-wells (commented):
Quick first comment: the method `finalize` for matrices is misleading. It doesn't finalise anything. The sparse structure is set at creation, so it's not finalising the sparsity. Any the matrix entries can be changed after calling `finalize`.

PR #2660, opened 2023-05-24
Wrong function-space used in direct solver

jorgensd:
This function space is not the same as the collapsed one. Causes issue if you change the bc to be space-dependent


jpdean (approved):
Good spot!

PR #2661, opened 2023-05-24
Update demo_stokes.py

massimiliano-leoni:
Fix sign of pressure term


garth-wells:
The sign is 'correct' - it's typical to flip the sign of the pressure to make the operator symmetric.

massimiliano-leoni:
Ah ok, I wasn't aware, sorry about that! You might consider saving the pressure with the opposite sign then, but at this point it's more of a matter of taste.

PR #2668, opened 2023-06-07
Bump PETSc to 3.19

jhale:
So, will this cause issues in the CI with mixing clang/gcc with gcc built PETSc?


jorgensd:
> So, will this cause issues in the CI with mixing clang/gcc with gcc built PETSc?

The mixing issue was already introduced in 3.18, so i dont think this should break anything.

I think they made a way to work around it (currently we dont test with clang any longer), Ref
last comment of https://gitlab.com/petsc/petsc/-/issues/1288

I guess we should add this.

garth-wells (approved):

PR #2679, opened 2023-06-21
Remove faulty FidesWriter constructor

massimiliano-leoni:
With the newest version of ADIOS2 the engine format `"BPFile"`, which is currently the default choice for Fides and VTX writers, does not exist any more. Possible choices include `"BP4" `and `"BP5"`, see https://adios2.readthedocs.io/en/latest/engines/engines.html .

While testing this edit I realised that there was an almost empty constructor that was called when no engine was specified, effectively shadowing the default parameter.
This constructor lead to the creation of malformed files [crashed Paraview]. I think it was a vestigial piece of code so I removed it.


garth-wells:
I'm not keen on forcing non-default options for compatibility with other libraries. It's not sustainable longer term and we don't have any policy on when an option can be updated. I'd also argue that the issue should be fixed at the source, i.e. the libraries that read the ADIOS2 files should check the engine type.

I'm satisfied with the recent change that the user can set the engine.

massimiliano-leoni:
I don't feel very strongly about this topic but the fact remains that currently we have the constructor

https://github.com/FEniCS/dolfinx/blob/0367aa9164e0c5d6d2fe0639b0b6a4971573a9c0/cpp/dolfinx/io/ADIOS2Writers.h#L567-L574

that forces a default parameter that ADIOS2 doesn't like [and thus yields a broken output file]; then we have the constructor

https://github.com/FEniCS/dolfinx/blob/0367aa9164e0c5d6d2fe0639b0b6a4971573a9c0/cpp/dolfinx/io/ADIOS2Writers.h#L475-L481

that is working fine and finally the constructor

https://github.com/FEniCS/dolfinx/blob/0367aa9164e0c5d6d2fe0639b0b6a4971573a9c0/cpp/dolfinx/io/ADIOS2Writers.h#L448-L453

that constructs a writer for a mesh, not a list of functions, I know, but accepts a default parameter, which again has a deprecated value and is structurally more similar to my proposed solution.

Plus, there's the fact that the first constructor does nothing while the second one, which should be equivalent, has a function body that performs a lot of checks and whatnot. Is this correct at all?

This looks a bit messy to me...
Since my proposed solution of removing constructor 1 and having constructor 2 pass on its default parameter was rejected, how would you like to address this issue?

garth-wells:
@massimiliano-leoni what is the string for the ADIOS2 default engine? I thought it was `"BPFile"`?

massimiliano-leoni:
@garth-wells I think it's simply "BP5" now. I've used this successfully in my latest tests.

garth-wells:
> @garth-wells I think it's simply "BP5" now. I've used this successfully in my latest tests.

My understanding is that "BPFile" maps to whatever the default engine is for ADIOS2 [1], which can change between ADIOS2 releases. As you say latest version uses BP5 as the default.

I'm struggling to see the issue. We allow the user to specify the engine, and if they don't specify it we use the ADIOS2 default engine.

[1] https://adios2.readthedocs.io/en/latest/api_full/api_full.html#_CPPv4N6adios22IO9SetEngineEKNSt6stringE

massimiliano-leoni:
I have looked into this more carefully and I think I see the issue more clearly now. I suspect I was partially hallucinating because of issue #2702 .
Anyway, as you correctly say, @garth-wells , `"BPFile"` works fine as a default. The real issue is that when the user specifies no engine type, thus calling constructor 1 from my previous comment, the resulting file is ill-formed and Paraview cannot read it properly.
I will now revert the change of default parameter but I will keep the removal of the faulty constructor so that the discussion can proceed from there.

jorgensd:
Could we expose the engine to the Python interface ?

garth-wells:
Engine options exposed in Python in #2708.

I this this PR can be closed. One constructor uses the default engine, the other allows the user to set the engine. We have two constructors to void having two default parameters, which isn't nice because to set the second last option explicitly requires the last to be set explicitly too.

There is no issue with 'shadowing' because the types are different - one is a `std::string` and the other is a strongly typed enum.

garth-wells:
See #2709 and #2708.

massimiliano-leoni:
I just pulled from `main`
but the issue still remains:
when I create a file using the constructor that doesn't specify the engine

```
io::FidesWriter<double> writeruL(meshL->comm(), "uL.bp",
                                 io::adios2_writer::U<double>{uL},
                                 io::FidesMeshPolicy::reuse);
```

then the output file is ill-formed and Paraview crashes when reading it [or rather, doesn't recognise it as a Fides file].

garth-wells:
You need to specify an engine that Paraview supports. DOLFINx uses the **default** ADIOS2 engine by default. We can't control Paraview/VTK bugs.

It is not sustainable to add hacks for version-dependent issues in 3rd-party libraries.

massimiliano-leoni:
The way I see it, that cannot be the issue: that constructor passes `"BPFile"` as a default parameter for the engine. If I pass the same string explicitly, which is supposed to do the same thing, then the output file works correctly.

I think I am failing at pointing out what the problem is but I created another branch that fixes the issue with a smaller edit. See https://github.com/FEniCS/dolfinx/compare/main...massimiliano-leoni:dolfinx:mleoni/FixFidesConstructor .

I think it's clear that that `FidesWriter` constructor should be wrapping the other `FidesWriter` constructor, not the `ADIOS2Writer` constructor.

If you see my point I'll open another pull request :)

PR #2685, opened 2023-06-26
Add elasticity to 'types' demo

garth-wells:
Demonstrates use of `MatrixCSR` for systems with a block size.

Also adds a matrix type (block size) option to the matrix assembly function.

### DIFF
--- a/python/demo/demo_types.py
+++ b/python/demo/demo_types.py
@@ -47,6 +47,7 @@ facets = mesh.locate_entities_boundary(msh, dim=1,
 
 
 def poisson():
+    """Poisson problem solver"""
 
     V = fem.FunctionSpace(msh, ("Lagrange", 1))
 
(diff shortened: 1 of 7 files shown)


python/demo/demo_types.py
```diff
@@ -40,94 +40,185 @@
...
+        """Solve the variational problem"""
+
+        # Process forms. This will compile the forms for the requested type.
+        a0, L0 = fem.form(a, dtype=dtype), fem.form(L, dtype=dtype)
+
+        # Create a Dirichlet boundary condition
+        bc = fem.dirichletbc(np.zeros(2, dtype=dtype), dofs, V=V)
+
+        # Assemble forms (BSR matrix)
+        sp = fem.create_sparsity_pattern(a0)
```
chrisrichardson, line 172: Don't need sp

chrisrichardson (approved):

PR #2690, opened 2023-06-28
Constrain scalar types with concept

jhale:
This PR constrains T (which is usually used to denote a scalar related to the finite element coefficients) to be a real or complex floating point type. Let me know if this needs to be widened re: multi-precision support.

In terms of naming 'field' might be more mathematically correct than 'scalar'.

I wonder also if we should change all of the Mesh related Ts to Us throughout?


jhale:
Looking at the diff I've used dolfinx::scalar where instead I should have used std::floating_point so I will make these changes.

cpp/dolfinx/fem/Constant.h
```diff
@@ -10,12 +10,14 @@
 #include <span>
 #include <vector>
 
+#include "dolfinx/common/types.h"
```
garth-wells, line 13: Remove blank line above.

cpp/dolfinx/fem/Constant.h
```diff
@@ -10,12 +10,14 @@
...
+
 namespace dolfinx::fem
 {
 
 /// Constant value which can be attached to a Form. Constants may be
 /// scalar (rank 0), vector (rank 1), or tensor-valued.
-template <typename T>
+template <dolfinx::scalar T>
```
garth-wells, line 20: Add `/// @tparam T ...` to docstring.

garth-wells (commented):
Looks good. Tiny formatting comment.

garth-wells (approved):

PR #2695, opened 2023-06-29
Remove more petsc-specific tests

chrisrichardson:
* Remove some more tests that don't really need PETSc
* Adds support for `discrete_gradient` to `MatrixCSR` and tests it.


python/dolfinx/wrappers/assemble.cpp
```diff
@@ -37,6 +37,52 @@ namespace py = pybind11;
...
+        sp.finalize();
+
+        // Build operator
+        dolfinx::la::MatrixCSR<T> A(sp);
+        dolfinx::fem::discrete_gradient<T, U>(
+            *V0.mesh()->topology_mutable(), {*V0.element(), *V0.dofmap()},
+            {*V1.element(), *V1.dofmap()}, A.mat_set_values());
+        return A;
+      },
+      py::return_value_policy::take_ownership, py::arg("V0"), py::arg("V1"));
```
garth-wells, line 83: I don't think `py::return_value_policy::take_ownership` is required - pybind11 handles move/return-by-value fine by default.

python/test/unit/fem/test_assemble_cppimport.py
```diff
@@ -64,18 +67,18 @@ def compile_eigen_csr_assembler_module():
 assemble_csr(const dolfinx::fem::Form<T>
 & a,
              const std::vector<std::shared_ptr<const dolfinx::fem::DirichletBC<T>>>& bcs)
-{
+{{
```
garth-wells, line 70: Double braces required?

python/test/unit/fem/test_discrete_operators.py
```diff
@@ -69,9 +65,11 @@ def test_gradient_interpolation(cell_type, p, q):
     V = FunctionSpace(mesh, (family0, p))
     W = FunctionSpace(mesh, (family1, q))
     G = discrete_gradient(V._cpp_object, W._cpp_object)
-    G.assemble()
+    # NB do not 'finalize' G
```
garth-wells, line 68: Could you expand on this comment?

python/test/unit/fem/test_discrete_operators.py
```diff
@@ -80,113 +78,15 @@ def test_gradient_interpolation(cell_type, p, q):
...
-        el0 = s_el
-        el1 = v_el
-
-    V = FunctionSpace(mesh, el0)
-    W = FunctionSpace(mesh, el1)
-    G = interpolation_matrix(V._cpp_object, W._cpp_object)
-    G.assemble()
 
-    u = Function(V)
+    # Get the local part of G (no ghost rows)
```
garth-wells, line 82: Is this correct? It is different from https://github.com/FEniCS/dolfinx/blob/6034c266dea5185eca26e1cedd226de1f18adfdf/python/test/unit/fem/test_expression.py#L124.

python/test/unit/fem/test_petsc_discrete_operators.py
```diff
@@ -0,0 +1,192 @@
...
+from dolfinx.mesh import (CellType, GhostMode, create_mesh, create_unit_cube,
+                          create_unit_square)
+from mpi4py import MPI
+from petsc4py import PETSc
+
+from dolfinx import default_real_type
+
+
+@pytest.mark.skip_in_parallel
+@pytest.mark.parametrize("mesh", [create_unit_square(MPI.COMM_WORLD, 11, 6, ghost_mode=GhostMode.none),
```
garth-wells, line 24: Parameterise over `dtype`.

garth-wells (changes requested):
Moving in the right direction!

python/test/unit/fem/test_discrete_operators.py
chrisrichardson: I don't see much of a difference - except for explicitly telling scipy the shape - but it can infer that.

python/test/unit/fem/test_assemble_cppimport.py
chrisrichardson: Yes with "f"-string

chrisrichardson:
> Could we add the interpolation matrix as well?

This is harder because we have templated block size for MatrixCSR and it is difficult in Python. Future PR?

garth-wells (approved):

.github/workflows/oneapi.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
     strategy:
       matrix:
-        petsc_arch: [linux-gnu-real32-32, linux-gnu-real64-32, linux-gnu-complex128-32]
```
garth-wells, line 24: This needs to be changed back.

garth-wells (changes requested):

PR #2697, opened 2023-06-29
Make numerical tolerance dependent on input type

jorgensd:
With the introduction of reduced precision, a careful examination of tolerances (especially for geometry) has to be undertaken.

I've tried to adjust the tolerance based on the input floating type epsilon, taking into account that some of these tolerances are squared in tests (thus should be much larger than the precision of the value).


jorgensd:
Could make `eps` a constexpr once [P1383R0](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1383r0.pdf) goes into the C++ standard (currently only works with gcc)

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -226,10 +226,11 @@ void interpolation_matrix(const FunctionSpace<U>& V0,
   e0->tabulate(basis_derivatives_reference0_b, X, Xshape, 0);
 
   // Clamp values
+  constexpr U atol = 1.0e2 * std::numeric_limits<U>::epsilon();
   std::transform(basis_derivatives_reference0_b.begin(),
                  basis_derivatives_reference0_b.end(),
                  basis_derivatives_reference0_b.begin(),
-                 [atol = 1e-14](auto x)
+                 [atol = atol](auto x)
```
garth-wells, line 233:
```suggestion
                 [atol](auto x)
```

garth-wells (approved):

PR #2707, opened 2023-07-04
Use `pyproject.toml` for install of Python part

chrisrichardson:
* Probably need to use `--no-build-isolation` to make it work.


jhale:
I'm not saying this is impossible to get working, but perhaps it's not time yet. Various issues with RPATH and from a user's point of view the --no-build-isolation is really easy to forget and can lead to hard to debug problems.

chrisrichardson:
I agree. I started looking at it, because I wanted to use Debug mode in the CI - I think this is currently broken.
Maybe we can pass an environment variable or something?

garth-wells:
We need to a bit cautious - `scikit-build-core` doesn't have full support for editable installs. The cost of re-compiling the `pybind11` wrapper at each install when just editing the Python makes development very slow.

jhale:
Good news, they fixed that:

`pip install --no-build-isolation --config-settings=build-dir="build/" .`

garth-wells:
> Good news, they fixed that:
> 
> `pip install --no-build-isolation --config-settings=build-dir="build/" .`

Nice to hear, have you tested it?

jhale:
Yep. Let me know when I can do a merge from your branch onto a (branch of) this branch.

jhale:
> > Good news, they fixed that:
> > `pip install --no-build-isolation --config-settings=build-dir="build/" .`
> 
> Nice to hear, have you tested it?

Yes, works on macOS (fast build).

francesco-ballarin:
@jhale I believe that this part

```
      package_data={'dolfinx.wrappers': ['*.h'], 'dolfinx': ['py.typed'],
                    'dolfinx.fem': ['py.typed'], 'dolfinx.nls': ['py.typed']},
```

which was removed from `setup.py` does not have an equivalent yet in the current `pyproject.toml`.
For instance, the following addition to `pyproject.toml`

```
[tool.setuptools.package-data]
dolfinx = ["py.typed"]
```

should be enough to install the `py.typed` marker, so that it can be used by downstream projects while running `mypy`.

I guess a similar syntax can be used for `'dolfinx.wrappers': ['*.h']`.

For the other two `py.typed` markers, I don't understand why they are needed at all.

python/setup.cfg
garth-wells, line 1: Can we remove this file?

python/CMakeLists.txt
```diff
@@ -114,3 +108,7 @@ if(NOT MPI4PY_COMMAND_RESULT)
 else()
   message(FATAL_ERROR "mpi4py could not be found.")
 endif()
+
+set_target_properties(cpp PROPERTIES INSTALL_RPATH_USE_LINK_PATH TRUE)
```
garth-wells, line 112: This might break relocatable builds, e.g. wheels. See how it work in Basix.

python/CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
-cmake_minimum_required(VERSION 3.16)
+cmake_minimum_required(VERSION 3.18)
```
garth-wells, line 1: Is this version high enough for `scikit-build-core`?

python/setup.cfg
jhale: Maybe. I cannot even work out where we're using `pycodestyle`! It doesn't support `pyproject.toml`.

jhale:
@francesco-ballarin Initially I agreed with your comment, but then I realised we're no longer using setuptools at all.

Can you confirm that `py.typed` is missing?

python/CMakeLists.txt
jhale: Yes, scikit-build-core requires cmake 3.15 or later. I copied 3.18 from the C++ layer.

jhale:
@francesco-ballarin I have `py.typed` and all of the source files in my `site-packages/dolfinx` directory on Mac, so I think it's OK. Thanks for following up though, easy to miss.

python/CMakeLists.txt
jhale: I/we to look at this more closely.

Basix are relocatable, I agree. Not convinced about DOLFINx.

francesco-ballarin:
Thanks @jhale, I can confirm that `py.typed` and header files are indeed there, I didn't realize (even though it was clearly written in the name :D ) that such option was specific for `setuptools`.

The situation is somehow the other way around now: there are too many files, in the sense that probably we wouldn't want `dolfinx/wrappers/*.cpp` files to be installed.

garth-wells:
> Yep. Let me know when I can do a merge from your branch onto a (branch of) this branch.

Actually, I'm happy for this branch to go first.

python/CMakeLists.txt
garth-wells: > I/we to look at this more closely.
> 
> Basix are relocatable, I agree. Not convinced about DOLFINx.

Could it affect conda?

jhale: @minrk Do you have something to add here? I'm honestly not sure how this setting `INSTALL_RPATH_USE_LINK_PATH` affects downstream packaging.

python/README.rst
```diff
@@ -1,32 +1,21 @@
...
-     mkdir build-dir
-     cd build-dir
-     cmake -DPYBIND11_TEST=off -DCMAKE_INSTALL_PREFIX=/path/to/pybind11/install ..
-     make install
+2. Ensure the build time requirements are installed::
 
-   The FEniCS Docker images are configured with pybind11.
-
-2. Build and install the DOLFINx C++ library in the usual way.
+     pip -v -r build-requirements.txt
```
francesco-ballarin, line 13: is there a `install` missing?

python/pyproject.toml
```diff
@@ -0,0 +1,47 @@
+# The DOLFINx Python interface must be built without build isolation (PEP517)
+# due to its depedency on system built petsc4py and mpi4py.
+# pip install -r build_requirements.txt
+[build-system]
+requires = ["scikit-build-core[pyproject]", "pybind11>=2.9.1", "petsc4py", "mpi4py"]
```
jhale, line 5: Is this correct? I would say we don't require python basix to build dolfinx python wrappers.

francesco-ballarin: Not sure what the question exactly is, but I do not follow why `petsc4py` and `mpi4py` are both here and in `dependencies`.

We should probably be aware of https://github.com/pypa/pip/issues/9314
which then led to the introduction of `--check-build-dependencies` in https://github.com/pypa/pip/issues/9794 and https://github.com/pypa/pip/pull/10886: due to `--no-build-isolation` the addition of  `petsc4py` and `mpi4py` to `requires` will only matter for checking build dependencies, not installing them.

jhale: petsc4py and mpi4py are build time and runtime dependencies of the python interface, so are listed in both places. pybind11 is only a build time dependency.

We can leave petsc4py and mpi4py out but that gives a much more cryptic error message from cmake.

minrk: As a downstream maintainer, I'd appreciate it if build dependencies are explicit. In conda, we don't use pip isolated builds (because conda is responsible for the env), but build dependencies and runtime dependencies are separate environments, and it can be beneficial to avoid putting unused runtime dependencies in the build environment (saves a small amount of time, but can also save unnecessarily strict dependency pinning at runtime).

francesco-ballarin: From @jhale's answer I understood that `mpi4py` and `petsc4py` are explicitly put in the build dependencies because `cmake` is looking for them at build time, right? They are both obviously also runtime dependencies, since the library imports them.

garth-wells: `mpi4py` and `petsc4py` *are* build dependencies. Maybe not immediately obvious, but we need the `mpi4py` and `petsc4py` header files for the C++ <-> Python casters.

jhale: I have improved the comment to mention buildtime and runtime.

garth-wells:
@jhale do you think this is ready to go?

No doubt there will be some refinements once it gets more testing.

francesco-ballarin:
@jhale I am not sure about 6ec932dde5bc5dd765fc6e183bc733e4f9f57c89. That file is being used by people to do `pip install build-requirements.txt`: do we want to risk `petsc4py` (and its bundled version of `PETSc`) getting installed by `pypi`?

francesco-ballarin:
> @jhale I am not sure about [6ec932d](https://github.com/FEniCS/dolfinx/commit/6ec932dde5bc5dd765fc6e183bc733e4f9f57c89). That file is being used by people to do `pip install build-requirements.txt`: do we want to risk `petsc4py` (and its bundled version of `PETSc`) getting installed by `pypi`?

My point is: it is fine to have them in `pypropject.toml`, so that `--check-build-dependencies` complains, but maybe do not add them in `build-requirements.txt`.

jhale:
Let's get this and refine as developers start to use it. There will certainly be some unforseen side effects!

PR #2723, opened 2023-07-13
Get Debug mode in Python working again for CI

chrisrichardson:


python/setup.py
```diff
@@ -48,11 +48,13 @@ def build_extension(self, ext):
                        f'-DPython3_LIBRARIES={sysconfig.get_config_var("LIBDEST")}',
                        f'-DPython3_INCLUDE_DIRS={sysconfig.get_config_var("INCLUDEPY")}']
 
-        cfg = 'Debug' if self.debug else 'Release'
-        build_args = ['--config', cfg]
-        cmake_args += ['-DCMAKE_BUILD_TYPE=' + cfg]
-
         env = os.environ.copy()
+        build_args = []
+        if "CMAKE_BUILD_TYPE" not in env:
```
garth-wells, line 53: Is this still needed? Is there a way to pass `self.debug`?

chrisrichardson: You can still use it with `setup.py` but not using pip

garth-wells (approved):

PR #2724, opened 2023-07-13
Enable KaHIP in test builds

jhale:
KaHIP is by default off, so we're not building/testing against it.


garth-wells:
It's tested in the https://github.com/FEniCS/dolfinx/blob/main/docker/Dockerfile.oneapi image.

jhale:
Thanks! My mistake.

PR #2728, opened 2023-07-14
Perform checks on matrix insertion in `MatrixCSR`

chrisrichardson:
Currently, there is no runtime check that matrix entries are correctly added (exist in sparsity), except in Debug mode.
The overhead to doing so is negligible, so this PR adds that check.

### DIFF
--- a/python/test/unit/la/test_matrix_csr.py
+++ b/python/test/unit/la/test_matrix_csr.py
@@ -222,15 +222,16 @@ def test_bad_entry(dtype):
     sp = create_test_sparsity(6, 1)
     mat1 = matrix_csr(sp, dtype=dtype)
 
-    # Set block in bs=1 matrix
+    # Set block in bs=1 matrix (tests insert_blocked_csr)
     with pytest.raises(RuntimeError):
         mat1.set([1.0, 2.0, 3.0, 4.0], [0], [0], 2)
-    # Normal
+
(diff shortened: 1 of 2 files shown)


jhale:
Looks good. I'd like to see more cheap sanity checks added throughout DOLFINx.

python/test/unit/la/test_matrix_csr.py
```diff
@@ -215,3 +215,22 @@ def test_set_diagonal_distributed(dtype):
...
+
+@pytest.mark.parametrize('dtype', [np.float32, np.float64, np.complex64, np.complex128])
+def test_bad_entry(dtype):
+    sp = create_test_sparsity(6, 1)
+    mat1 = matrix_csr(sp, dtype=dtype)
+
+    # Set block in bs=1 matrix
+    with pytest.raises(RuntimeError):
+        mat1.set([1.0, 2.0, 3.0, 4.0], [0], [0], 2)
+    # Normal
```
garth-wells, line 228: What does this comment mean?

chrisrichardson: Data has block size but matrix does not (bs=1)

garth-wells: My comment was in invitation to improve the code comment :). `# Normal` carries no meaning.

chrisrichardson: Sure will do.

Issue #2363, opened 2022-09-13
`dolfinx.fem.Expression` does not support "pure UFL expressions"


nate-sime:
`dolfinx.fem.Expression`'s first argument is a "pure UFL expression" https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/fem/function.py#L86.

However `dolfinx.fem.Expression` cannot be instantiated with a pure UFL expression since a dolfinx mesh and its MPI communicator are assumed present https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/fem/function.py#L117.

This causes the following to fail:

```python
from mpi4py import MPI
import dolfinx
import ufl

mesh = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, 2, 2)
v = dolfinx.fem.Constant(mesh.ufl_cell(), 1.0)
v_f = dolfinx.fem.Function(dolfinx.fem.FunctionSpace(mesh, ("CG", 1)))
v_expr = dolfinx.fem.Expression(v, v_f.function_space.element.interpolation_points())
```

```bash
Traceback (most recent call last):
  File "$file", line 8, in <module>
    v_expr = dolfinx.fem.Expression(v, v_f.function_space.element.interpolation_points())
  File "$PYTHONPATH/dolfinx/fem/function.py", line 117, in __init__
    self._ufcx_expression, module, self._code = jit.ffcx_jit(mesh.comm, (ufl_expression, _X),
AttributeError: 'NoneType' object has no attribute 'comm'
```

This is also true for simple pure UFL forms such as `v = ufl.as_ufl(1)`.

These ufl forms are particularly useful when working with non-matching meshes to avoid re-evaluating/JITting forms for each mesh.

Should `dolfinx.fem.Expression` have an explicit mesh/MPI communicator argument which is checked against a (potentially existing) `ufl.Expr.ufl_cargo` mesh?

jorgensd:
A tangential issue is if one wants to have different expressions on different processes, say 
```python
if mesh.comm.rank == 0;
    expr = Expression(p, np.array([0.,0.])
else if mesh.comm.rank ==1:
    expr = Expression(p, np.array([1.,0.])
else:
    pass
```
will cause a deadlock when executed on more than two processes, due to 
https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/jit.py#L80
which expects the expression to be called on all processes.

PR #2730, opened 2023-07-18
Simplify `fem::Expression`

garth-wells:
C++ Expression no longer requires a mesh at construction, which supports fixing of #2363 and will avoid some unnecessary JIT. The mesh is now passed to the `Expression::eval` function.

Also remove `dict()` default parameters, use `None` instead.

### DIFF
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -464,8 +464,7 @@ class Function(ufl.Coefficient):
 
     def collapse(self) -> Function:
         u_collapsed = self._cpp_object.collapse()
-        V_collapsed = FunctionSpace(
-            self.function_space._mesh, self.ufl_element(), u_collapsed.function_space)
+        V_collapsed = FunctionSpace(self.function_space._mesh, self.ufl_element(), u_collapsed.function_space)
         return Function(V_collapsed, Vector(u_collapsed.x))
 
 
@@ -666,8 +665,7 @@ def VectorFunctionSpace(mesh: Mesh,
                         dim=None) -> FunctionSpace:
     """Create vector finite element (composition of scalar elements) function space."""
     if not _is_scalar(mesh, element):
-        raise ValueError(
-            "Cannot create vector element containing a non-scalar.")
+        raise ValueError("Cannot create vector element containing a non-scalar.")
 
     try:
         ufl_e = basix.ufl.element(element.family(), element.cell_type,         # type: ignore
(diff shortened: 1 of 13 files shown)


nate-sime (approved):
Looks great.

python/dolfinx/fem/function.py
```diff
@@ -131,51 +146,55 @@ def __init__(self, ufl_expression: ufl.core.expr.Expr, X: np.ndarray,
...
-            )._cpp_object
+            self._argument_function_space = arguments[0].ufl_function_space()._cpp_object
         else:
-            raise RuntimeError(
-                "Expressions with more that one Argument not allowed.")
+            raise RuntimeError("Expressions with more that one Argument not allowed.")
 
         def create_expression(dtype):
-            if dtype is np.float32:
+            if dtype == np.float32:
```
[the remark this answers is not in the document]
garth-wells: @nate-sime I don’t quite follow, could you expand?

python/dolfinx/fem/function.py
```diff
@@ -83,9 +84,9 @@ def __complex__(self):
 
 
 class Expression:
-    def __init__(self, ufl_expression: ufl.core.expr.Expr, X: np.ndarray,
-                 form_compiler_options: dict = {}, jit_options: dict = {},
-                 dtype=default_scalar_type):
+    def __init__(self, e: ufl.core.expr.Expr, X: np.ndarray,
+                 comm: typing.Optional[_MPI.Comm] = None, form_compiler_options: dict = {},
+                 jit_options: dict = {}, dtype=None):
```
jorgensd, line 89: We should really use `jit_options: Optional[dict] = None`, ref https://stackoverflow.com/questions/26320899/why-is-the-empty-dictionary-a-dangerous-default-value-in-python

python/dolfinx/fem/function.py
```diff
@@ -659,12 +676,12 @@ def VectorFunctionSpace(mesh: Mesh,
...
                                   element.degree(), element.lagrange_variant,  # type: ignore
                                   element.dpc_variant, element.discontinuous,  # type: ignore
-                                  shape=(mesh.geometry.dim, ) if dim is None else (dim, ),
+                                  shape=(mesh.geometry.dim,) if dim is None else (dim, ),
                                   gdim=mesh.geometry.dim, rank=1)
     except AttributeError:
         ed = ElementMetaData(*element)
         ufl_e = basix.ufl.element(ed.family, mesh.basix_cell(), ed.degree,
-                                  shape=(mesh.geometry.dim, ) if dim is None else (dim, ),
+                                  shape=(mesh.geometry.dim,) if dim is None else (dim, ),
```
jorgensd, line 684: Why this change in formatting?

garth-wells: autopep8 formatter change.

PR #2734, opened 2023-07-22
Refactor Gmsh demo

garth-wells:
Refactor Gmsh demo into functions. Makes dependencies clearer.

Also makes it easier to extract core Gmsh steps for unit tests that need higher-order meshes.

### DIFF
--- a/python/demo/demo_gmsh.py
+++ b/python/demo/demo_gmsh.py
@@ -10,79 +10,72 @@
 
 # # Mesh generation with Gmsh
 #
-# Copyright (C) 2020-2022 Garth N. Wells and Jørgen S. Dokken
+# Copyright (C) 2020-2023 Garth N. Wells and Jørgen S. Dokken
 #
-# This demo shows how to create meshes uses the Gmsh Python interface.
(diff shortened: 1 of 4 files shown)


python/demo/demo_gmsh.py
```diff
@@ -183,15 +142,96 @@
...
+# for visualisation, e.g. using ParaView.
+
+# +
 
-with XDMFFile(msh.comm, "out_gmsh/mesh.xdmf", "a") as file:
-    file.write_mesh(msh)
-    file.write_meshtags(mt, msh.geometry, geometry_xpath=f"/Xdmf/Domain/Grid[@Name='{msh.name}']/Geometry")
-    file.write_meshtags(ft, msh.geometry, geometry_xpath=f"/Xdmf/Domain/Grid[@Name='{msh.name}']/Geometry")
 
+def create_mesh(comm: MPI.Comm, model: gmsh.model, name: str, filename: str, mode: str):
```
jorgensd, line 157: Maybe use pathlib instead of strings?

garth-wells: I think for a demo `str` is fine - gives and indication without adding clutter.

PR #2735, opened 2023-07-22
Require interpolation before XDMF output

### LINKED ISSUE #593: Replace Function.compute_point_values
`Function.compute_point_values` is used in post-processing (`VTKFile.write(u)`), which interpolates the solution for each point in the mesh. I think this code could be removed, and we could do the following:
- For CG spaces: Picks values from the underlying vector and avoids computation of the basis functions.
- For non-CG spaces: Generate the appropriate CG space (depending on mesh order) and interpolate the solution to that space using the `FunctionSpace.interpolate` function instead of reinventing the wheel.

### LINKED ISSUE #1936: Require Function interpolation before file output
A long standing issue is the 'magic' behind the scenes if interpolation to another space (usually P1) before output to a file. This is not sufficiently explicit, can be unexpectedly expensive and give surprising visualisation results, especially with H(div/curl) spaces.

Now that we have fast interpolation between spaces, we should require interpolation to a space supported by the file backend before calling the file output function. 

See also #593.

- [x] VTK
- [x] VTX
- [x] Fides
- [ ] XMDF

### LINKED ISSUE #1958: Remove `Function::compute_point_values`
This function is slow and no longer required now that we have improved interpolation support. Remaining use is the in the IO code.

Was removed from the Python interface in #1955.

garth-wells:
XDMF output could perform silent interpolation. This change requires explicit interpolation into a suitable space, consistent with other backends.

- Fixes #1958, fixes #1936 and fixes #593. 
- Expanded XDMF unit tests
- Fixes bug in creation of a `fem::FiniteElement` from a Basix element for tensor elements, where value shape was incorrect.
- Add Lagrange precondition check to VTK file output


cpp/dolfinx/io/ADIOS2Writers.h
```diff
@@ -843,13 +843,13 @@ class VTXWriter : public ADIOS2Writer
   ///
   /// This format supports arbitrary degree meshes.
   ///
-  /// @param[in] comm The MPI communicator to open the file on
-  /// @param[in] filename Name of output file
-  /// @param[in] mesh The mesh to write
-  /// @param[in] engine ADIOS2 engine type
-  /// @note This format support arbitrary degree meshes
+  /// @param[in] comm MPI communicator to open the file on.
+  /// @param[in] filename Name of output file,
```
jorgensd, line 847: We don't usually use commas to separate Params.

cpp/dolfinx/io/xdmf_function.cpp
```diff
@@ -88,98 +56,153 @@ void xdmf_function::add_function(MPI_Comm comm, const fem::Function<T, U>& u,
   assert(u.function_space());
   auto mesh = u.function_space()->mesh();
   assert(mesh);
+  auto element = u.function_space()->element();
+  assert(element);
+
+  // FIXME: is the below check adequate for detecting a Lagrange
+  // element?
+  // Check that element is Lagrange
```
jorgensd, line 64: I don't think so. Maybe @mscroggs  has examples of elements with point evaluations that are not Lagrange?

cpp/dolfinx/io/xdmf_function.cpp
```diff
@@ -88,98 +56,153 @@ void xdmf_function::add_function(MPI_Comm comm, const fem::Function<T, U>& u,
...
-  auto map_c = mesh->topology()->index_map(mesh->topology()->dim());
-  assert(map_c);
+    // Check that dofmap layouts are equal and check Lagrange variants
+    if (dofmap->element_dof_layout() != cmap.create_dof_layout())
+    {
+      throw std::runtime_error("Function and Mesh dof layouts do not match. "
+                               "Maybe the Function needs to be interpolated?");
+    }
+    if (cmap.degree() > 2
+        and element->basix_element().lagrange_variant() != cmap.variant())
```
jorgensd, line 125: Do we really need the lagrange varaint check as for degree less than 3 the spacing shouldn't matter?

python/demo/demo_stokes.py
```diff
@@ -275,8 +275,11 @@ def nested_iterative_solver():
     # `scatter_forward`.
     with XDMFFile(MPI.COMM_WORLD, "out_stokes/velocity.xdmf", "w") as ufile_xdmf:
         u.x.scatter_forward()
+        P1 = element("Lagrange", msh.basix_cell(), 1, rank=1)
```
jorgensd, line 278: Instead of using XDMFFile, should we use `VTX`, as this clearly reduces the accuracy of the solution?

python/dolfinx/io/gmshio.py
```diff
@@ -167,9 +171,9 @@ def extract_geometry(model: gmsh.model, name: typing.Optional[str] = None) -> np
         return points[perm_sort]
 
     def model_to_mesh(model: gmsh.model, comm: _MPI.Comm, rank: int, gdim: int = 3,
-                      partitioner: typing.Callable[
-            [_MPI.Comm, int, int, AdjacencyList_int32], AdjacencyList_int32] =
-            create_cell_partitioner(GhostMode.none), dtype=default_real_type) -> typing.Tuple[
+                      partitioner: typing.Optional[typing.Callable[
+                          [_MPI.Comm, int, int, AdjacencyList_int32], AdjacencyList_int32]] = None,
+                      dtype=default_real_type) -> typing.Tuple[
```
jorgensd, line 176: Why this change of partitioner?

python/dolfinx/io/gmshio.py
```diff
@@ -319,4 +323,5 @@ def read_from_msh(filename: str, comm: _MPI.Comm, rank: int = 0, gdim: int = 3,
                       11: ("tetrahedron", 2), 12: ("hexahedron", 2),
                       15: ("point", 0), 21: ("triangle", 3),
                       26: ("interval", 3), 29: ("tetrahedron", 3),
-                      36: ("quadrilateral", 3)}
+                      36: ("quadrilateral", 3),
+                      92: ("hexahedron", 3)}
```
jorgensd, line 327: Is this tested? I don't think we have the dof permutations for it, ref https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/io/cells.cpp#L294-L303

jorgensd: Later in the PR I see you have added a fixme to support higher order hexes.

jorgensd (changes requested):

cpp/dolfinx/io/xdmf_function.cpp
garth-wells: Hence the first check above - `cmap.degree() > 2`.

python/demo/demo_stokes.py
garth-wells: Could be done in a later PR. This change doesn't change the precision of the output, just makes it obvious rather than hidden.

ADIOS2 is an optional dependency, so there is a question as to how much it should be used in demos.

python/dolfinx/io/gmshio.py
garth-wells: Consistent with change elsewhere, see #2727. It avoids calling a partitioner in serial case.

python/dolfinx/io/gmshio.py
garth-wells: It now throws and exception at a place that makes the source of the error/limitation clear.

python/demo/demo_stokes.py
garth-wells: I added comment to the XDMF function docstring pointing to `VTXWriter`.

jorgensd (approved):

PR #2738, opened 2023-08-03
Reuse mesh when saving to VTX

massimiliano-leoni:
Similarly to Fides, VTX too supports saving the mesh only once and reusing it for all time steps. The reason this didn't work out-of-the-box when we implemented the corresponding functionality for Fides was that, while the mesh was only saved once and the actual finite element functions were saved many times -- which is correct -- the functions `"vtkOriginalPointIds"` and `"vtkGhostType"` were saved only once too as they are saved when saving a mesh. For some reason, ADIOS2 needs them to be saved for each time step, probably because it needs all functions to be saved at all time steps. See https://gitlab.kitware.com/vtk/vtk/-/issues/19033 for more details.

This pull request follows in the footsteps of the one that implemented mesh reuse for Fides but also fixes this issue by caching the data for `"vtkOriginalPointIds"` and `"vtkGhostType"` when the mesh is written and saving additional copies of these functions at each time step.

The main change is in `VTXWriter::write` where the code now checks if the policy is to always write the mesh [or if it's the first write], in which case it writes the mesh before writing the functions. In case the policy is to reuse the mesh, it only writes the necessary data [as discussed above] and the functions.


nate-sime (approved):

jorgensd:
A few things that I think could be useful to consider:
1. Python interface (we want it to be possible to control it from there)
2. Single constructor. The engine (BP4) should be possible to control
3. Does VTX support having meshes on a sub-set of time steps? i.e. say we have 10 time steps, but the mesh is constant from step 0-3, step 4-8 and step 9. If it is possible to then just store 3 meshes, the policy should be input to the `write` function rather than the constructor.
4. Does VTX support the mesh topology changing? If not, can we just write the geometry and not topology to file for every rewrite?

massimiliano-leoni:
> A few things that I think could be useful to consider:
> 
>     1. Python interface (we want it to be possible to control it from there)

I think I added it. Can you check it?

>     2. Single constructor. The engine (BP4) should be possible to control

I'm following the style Garth recommended for the Fides version of this PR.

>     3. Does VTX support having meshes on a sub-set of time steps? i.e. say we have 10 time steps, but the mesh is constant from step 0-3, step 4-8 and step 9. If it is possible to then just store 3 meshes, the policy should be input to the `write` function rather than the constructor.
> 
>     4. Does VTX support the mesh topology changing? If not, can we just write the geometry and not topology to file for every rewrite?

I don't know the answer to either of these questions but they are a little bit out of the scope of this pull request.
Can we merge this now and look at additional features later? :)

cpp/dolfinx/io/ADIOS2Writers.h
```diff
@@ -960,6 +975,27 @@ class VTXWriter : public ADIOS2Writer
...
+  /// @param[in] mesh_policy Controls if the mesh is written to file at
+  /// the first time step only or is re-written (updated) at each time
+  /// step.
+  /// @note This format supports arbitrary degree meshes.
+  VTXWriter(MPI_Comm comm, const std::filesystem::path& filename,
+            const typename adios2_writer::U<T>& u,
+            VTXMeshPolicy mesh_policy = VTXMeshPolicy::update)
+      : VTXWriter(comm, filename, u, "BPFile", mesh_policy)
+  {
+  }
```
jorgensd, line 997: I think we can remove this, as it could rather be a default argument of the previous constructor.

jorgensd (approved):
The only thing I'm missing from this PR is a test in C++ or Python on usage, as adding it with no testing in the library is a way of introducing bugs.

cpp/dolfinx/io/ADIOS2Writers.h
massimiliano-leoni: As I mentioned, in the Fides PR I used the approach that you recommend and it was heavily opposed on the grounds that

> a constructor shall not have more than one default argument.

I eventually changed it back then and now I just reused the same "idiom".

massimiliano-leoni:
@jorgensd I added a unit test.

cpp/test/io.cpp
```diff
@@ -77,6 +77,39 @@ void test_fides_function()
   writer1.write(0.0);
 }
 
+template <std::floating_point T>
```
jhale, line 80: Just put the test in Python, then we have coverage of the code and its Python interface.

cpp/test/io.cpp
```diff
@@ -77,6 +77,39 @@ void test_fides_function()
...
+
+  std::filesystem::path f = "test_vtx_reuse_mesh" + std::to_string(sizeof(T))
+                            + ".bp";
+  io::VTXWriter<T> writer(mesh->comm(), f, {u, v}, "BPFile",
+                          io::VTXMeshPolicy::reuse);
+  writer.write(0);
+
+  std::fill(u->x()->mutable_array().begin(), u->x()->mutable_array().end(), 1);
+
+  writer.write(1);
```
jhale, line 110: I don't know what is possible here, but there is no 'assertion' of correct functioning in this test.

jhale (commented):
Some small comments then I think this can be merged.

cpp/dolfinx/io/ADIOS2Writers.h
```diff
@@ -839,9 +840,18 @@ void vtx_write_mesh_from_space(adios2::IO& io, adios2::Engine& engine,
...
 
   engine.PerformPuts();
+  return {std::move(x_id), std::move(x_ghost)};
 }
 } // namespace impl_vtx
 
+/// Mesh reuse policy
+enum class VTXMeshPolicy
+{
+  update, ///< Re-write the mesh to file upon every write of a fem::Function
```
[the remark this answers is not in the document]
massimiliano-leoni: I wouldn't know, I diligently copy-pasted them from the corresponding code for the Fides Writer.

cpp/test/io.cpp
jorgensd: We could use `adios2` directly to verify the data, which I guess we should do more often.
If @massimiliano-leoni ports the test, I can have a look at what can be done wrt. verification.

massimiliano-leoni: @jhale I was thinking the same thing while writing the test.
@jorgensd Do you mean porting to Python?

cpp/test/io.cpp
massimiliano-leoni: The choice I was given was but an illusion D:

cpp/test/io.cpp
jorgensd: Yes, If you port it to Python, I can see if I can cook up some adios2 reader verification code.

cpp/dolfinx/io/ADIOS2Writers.h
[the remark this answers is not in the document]
jhale: Probably clever Doxygen stuff

https://docs.fenicsproject.org/dolfinx/v0.7.3/cpp/io.html#_CPPv4I00EN7dolfinx2io11FidesWriterE11FidesWriterINSt12remove_cvrefIN1T12element_typeEE4type13geometry_type10value_typeEE8MPI_Comm1U1T

cpp/test/io.cpp
massimiliano-leoni: Done. Or at least I think -- I'm not familiar with the framework.

jorgensd (commented):
Added a test that checks if data is placed in appropriate steps with the `VTXMeshPolicy.update` and `VTXMeshPolicy.reuse`.
I guess we should add a note somewhere that `reuse` can only be opened in Paraview>=5.12

PR #2755, opened 2023-09-03
fix: Commented terms in hyperelasticity demo

### LINKED ISSUE #2584: Commented terms in hyperelasticity demo
In the hyperelasticity demo, there are terms that have been commented out:

https://github.com/FEniCS/dolfinx/blob/dfee3c4449625a1141650e3c678487627a6a01b9/cpp/demo/hyperelasticity/hyperelasticity.py#L32-L33

https://github.com/FEniCS/dolfinx/blob/dfee3c4449625a1141650e3c678487627a6a01b9/cpp/demo/hyperelasticity/hyperelasticity.py#L65

These should either be uncommented or removed

xaea-twlv:
Fixes #2584

### DIFF
--- a/cpp/demo/hyperelasticity/hyperelasticity.py
+++ b/cpp/demo/hyperelasticity/hyperelasticity.py
@@ -10,8 +10,8 @@
 # trial and test functions on this space::
 from basix.ufl import element
 from ufl import (Coefficient, FunctionSpace, Identity, Mesh, TestFunction,
-                 TrialFunction, derivative, det, diff, dx, grad, ln, tr,
-                 variable)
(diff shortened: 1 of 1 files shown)


garth-wells:
@xaea-twlv could you fix the test failures?

francesco-ballarin:
No further replies from the author in two months, and out of date with current `main`: closing. Feel free to open another PR in future if you want to have another go at this, just make sure to also change the c++ source code, as the current attempt fails with
```
9: terminate called after throwing an instance of 'std::runtime_error'
9:   what():  Form coefficient "B" not provided.
```
(and, similarly, also `T` should be provided).

Issue #2750, opened 2023-08-25
[BUG]: Symmetric spaces broken


jorgensd:
### How to reproduce the bug

Based on: https://fenicsproject.discourse.group/t/projection-of-function-to-mixed-variational-form-error-dolfinx/11878?u=dokken
but simplfied.
Removing symmetry=True resolves the error.

### Minimal Example (Python)

```Python
from mpi4py import MPI
import ufl
import dolfinx
mesh = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, 10, 10)
P2 = ufl.TensorElement("CG",mesh.ufl_cell(),1, symmetry=True) # strain
P1 = ufl.VectorElement("CG",mesh.ufl_cell(),1)                 # displacement

MX = ufl.MixedElement([P1,P2])
W = dolfinx.fem.FunctionSpace(mesh,MX)
W_eps, _ = W.sub(1).collapse()
u = dolfinx.fem.Function(W_eps)
p = dolfinx.fem.Function(W)
uk, _ = ufl.split(p)
v = ufl.TestFunction(W_eps)
L = ufl.inner(ufl.grad(uk), v) * ufl.dx
dolfinx.fem.form(L)
```

### Output (Python)

```bash
python3: can't open file '/root/shared/mwe123': [Errno 2] No such file or directory
root@20eb7ebf7e20:~/shared# python3 mwe123.py 
Traceback (most recent call last):
  File "/root/shared/mwe123.py", line 9, in <module>
    W = dolfinx.fem.FunctionSpace(mesh,MX)
  File "/usr/local/dolfinx-real/lib/python3.10/dist-packages/dolfinx/fem/function.py", line 519, in __init__
    (self._ufcx_element, self._ufcx_dofmap), module, code = jit.ffcx_jit(
  File "/usr/local/dolfinx-real/lib/python3.10/dist-packages/dolfinx/jit.py", line 55, in mpi_jit
    return local_jit(*args, **kwargs)
  File "/usr/local/dolfinx-real/lib/python3.10/dist-packages/dolfinx/jit.py", line 205, in ffcx_jit
    r = ffcx.codegeneration.jit.compile_elements([ufl_object], options=p_ffcx, **p_jit)
  File "/usr/local/lib/python3.10/dist-packages/ffcx/codegeneration/jit.py", line 117, in compile_elements
    ffcx.naming.compute_signature(elements, _compute_option_signature(p)
[...]

PR #2766, opened 2023-09-09
Add shape argument to `FunctionSpace` for creating blocked scalar element spaces

garth-wells:
The shape argument to `FunctionSpace` is for 'blocked' scalar elements. This removes the need for `VectorFunctionSpace` and `TensorFunctionSpace`. `TensorFunctionSpace` has been removed and `VectorFunctionSpace` has been deprecated.

PR also changes `FunctionSpace` to a function, and the class is renamed `FunctionSpaceBase`. This simplifies the code significantly. Eventually `FunctionSpace` can be changed to `functionspace`.

Many documentation additions and improvements.

### DIFF
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -479,7 +479,7 @@ class Function(ufl.Coefficient):
 
 
 class ElementMetaData(typing.NamedTuple):
-    """Data for representing a finite element.
+    """Data for representing a finite element
 
     :param family: Element type.
     :param degree: Polynomial degree of the element.
@@ -518,7 +518,7 @@ def FunctionSpace(mesh: Mesh,
         ufl_e = basix.ufl.element(e.family, mesh.basix_cell(), e.degree, shape=e.shape,
                                   gdim=mesh.ufl_cell().geometric_dimension())
     except TypeError:
-        ufl_e = element
+        ufl_e = element  # type: ignore
 
     # Check that element and mesh cell types match
     if ufl_e.cell() != mesh.ufl_domain().ufl_cell():
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -498,7 +498,7 @@ class ElementMetaData(typing.NamedTuple):
     symmetry: typing.Optional[bool] = None
 
 
-def FunctionSpace(mesh: Mesh,
+def functionspace(mesh: Mesh,
                   element: typing.Union[ufl.FiniteElementBase, ElementMetaData,
                                         typing.Tuple[str, int, typing.Tuple, bool]],
                   form_compiler_options: typing.Optional[dict[str, typing.Any]] = None,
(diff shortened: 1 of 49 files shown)


mscroggs (approved):

jorgensd:
So with this approach, you would have to use `basix.ufl.element` to encode symmetry to your space?

python/dolfinx/fem/function.py
```diff
@@ -473,79 +478,112 @@ def split(self) -> tuple[Function, ...]:
...
+
+    """
+
+    # Create UFL element
+    try:
+        e = ElementMetaData(*element)
+        ufl_e = basix.ufl.element(e.family, mesh.basix_cell(), e.degree, shape=e.shape,
+                                  gdim=mesh.ufl_cell().geometric_dimension())
+    except TypeError:
+        ufl_e = element  # type: ignore
```
jorgensd, line 525: The symmetry option doesn't seem to be passed down from the `ElementMetaData` to basix

python/dolfinx/fem/function.py
```diff
@@ -473,79 +478,112 @@ def split(self) -> tuple[Function, ...]:
...
+    try:
+        cppV = _cpp.fem.FunctionSpace_float64(mesh._cpp_object, cpp_element, cpp_dofmap)
+    except TypeError:
+        cppV = _cpp.fem.FunctionSpace_float32(mesh._cpp_object, cpp_element, cpp_dofmap)
+
+    return FunctionSpaceBase(mesh, ufl_e, cppV)
 
 
-class FunctionSpace(ufl.FunctionSpace):
+class FunctionSpaceBase(ufl.FunctionSpace):
```
jorgensd, line 562: I would rather like it if we kept the name `FunctionSpace` for the class, and went directly for `functionspace` as a constructor above. I don't think having a base-class that is never extended makes sense.

python/test/unit/fem/test_assemble_cppimport.py
```diff
@@ -8,21 +8,20 @@
 import pathlib
 
 import cppimport
+import dolfinx.pkgconfig
```
jorgensd, line 11: What happened to the import structure in this code?

python/dolfinx/fem/function.py
garth-wells: Fixed.

python/dolfinx/fem/function.py
garth-wells: Eventually I would like to have the function `functionspace` that returns a space, and rename the class `FunctionSpaceBase` -> `FunctionSpace`. But, this is an interface change which I think we might need to wait a release for.

I've now added `functionspace` and add a deprecation not to the function `FunctionSpace`.

jorgensd:
@mscroggs what is the status on symmetry, ref: https://github.com/FEniCS/dolfinx/issues/2750

python/test/unit/fem/test_assemble_cppimport.py
garth-wells: No idea!

Line is removed now.

PR #2771, opened 2023-09-14
Error when importing dolfinx.io.gmshio with no gmsh

jhale:
If gmsh is not installed there is no need to import this module.

In the old version, the import silently failed and functions related to gmsh were missing.

If importing gmsh fails, raise a new ModuleNotFoundError with a slightly better error message.


jorgensd:
Could we add the same thing for VTXWriter, FidesWriter (then it needs a build time requirement).

https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/io/utils.py#L38

jhale:
Would this require splitting out VTX and Fides (need adios2) and XDMF into separate files? At the moment they are all in `utils.py`.

jorgensd:
> Would this require splitting out VTX and Fides (need adios2) and XDMF into separate files? At the moment they are all in `utils.py`.

That would be the cleanest solution I guess

jhale:
I will split this into multiple parts as the other one will be an API change.

garth-wells:
From what I can see, only `doflinx.io.gmshio.read_from_msh` calls a GMsh function. Some other functions call Gmsh methods on a Gmsh object that the user passes in. Hence, type hints aside, only `read_from_msh` needs to import `gmsh`. How about we just put `import gmsh` inside `read_from_msh`, with a `try` / `except`.

jhale:
@garth-wells My understanding for this slightly odd design is its related to type checking. The if statements 'hide' the gmsh specific types in the function definitions from the type checker. I could be wrong though.

I am having knock on issues from the proposed change in our `common/test_public_api.py` test.

garth-wells:
> @garth-wells My understanding for this slightly odd design is its related to type checking. The if statements 'hide' the gmsh specific types in the function definitions from the type checker. I could be wrong though.

We can use strings for the types - no big deal if it's just in a small number of places. If `mypy` checking is the issue, we can just disable it for the very small number of affected functions.

garth-wells:
Superseded by #2775.

PR #2775, opened 2023-09-19
Re-design gmsh import in `io.gmshio`

garth-wells:
- Makes documentation generation independent on whether or not gmsh can be imported
- Adds errors message if function that requires gmsh is called and gmsh is not available
- Required a small number of type hints for gmsh types to be removed

Supersedes #2771.


jhale (approved):
Looks good. I would propose we merge this and then think about how to apply the principles to our other modules that use optional dependencies, e.g. adios2.

PR #2783, opened 2023-09-21
Add complex wrappers for equivalent floating type for exposed dof transformations

jorgensd:
Required for: https://github.com/jorgensd/adios4dolfinx/pull/32

### DIFF
--- a/python/dolfinx/wrappers/fem.cpp
+++ b/python/dolfinx/wrappers/fem.cpp
@@ -154,6 +154,36 @@ void declare_function_space(py::module& m, std::string type)
                   std::span(x.mutable_data(), x.size()), cell_permutation, dim);
             },
             py::arg("x"), py::arg("cell_permutation"), py::arg("dim"))
+        .def(
+            "apply_dof_transformation",
+            [](const dolfinx::fem::FiniteElement<T>& self,
+               py::array_t<std::complex<T>, py::array::c_style> x,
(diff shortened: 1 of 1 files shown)


garth-wells (approved):

garth-wells:
We probably want to add 'vectorised' versions of these functions if they're used for something other than small-scale testing.

jorgensd:
> We probably want to add 'vectorised' versions of these functions if they're used for something other than small-scale testing.

What do you imply by vectorized here?
You can use the `dim` argument to apply the transformation to loads of data, given that it is batched by the `permutation_index`. (https://github.com/FEniCS/dolfinx/blob/ca09b5d48de34cd0373d50ac0803696653483832/cpp/dolfinx/fem/FiniteElement.h#L527-L535). I'm not sure why we call it dim in the pybinding and block_size everywhere else.

jorgensd:
> We probably want to add 'vectorised' versions of these functions if they're used for something other than small-scale testing.

They are used in my checkpointing/input-output library, which is currently the only way of saving/loading `dolfinx.fem.Function` based on any finite element in DOLFINx, so I guess it is not only used in small-scale testing.

IgorBaratta (approved):

garth-wells:
> > We probably want to add 'vectorised' versions of these functions if they're used for something other than small-scale testing.
> 
> What do you imply by vectorized here? 

Process data for more than one cell at a time.

Issue #2774, opened 2023-09-19
[BUG]: Collision detection does not work below a certain tolerance


do-wen:
### How to reproduce the bug

I stumpled across this issue when using the interpolate function for non-matching meshes. When I have only a 'small' shift of coordinates of the two meshes, the interpolate function returns a zero. This behavior is clear for boundary values, but it also happens for values inside both domains. I could track this issue back to the geometry.compute_colliding_cells function.

In the minimal example I presented the issue:
The 'evaluate' function I define is based on this fenicsx tutorial: https://jsdokken.com/dolfinx-tutorial/chapter1/membrane_code.html#making-curve-plots-throughout-the-domain
I am creating two meshes and shift one of those meshes slightly by a random distribution (I added the seed for reproducability). I am then defining some function on mesh 1 and interpolate it to mesh 2. At one position in the function defined on mesh 2 a zero is interpolated, although it is inside both meshes. You can see this clearly in paraview, I also print the problematic DOF to the console. I also write the two functions to paraview where you can have a closed look. Making the eps parameters smaller results in more of these candidates.

Then I am using the 'evaluate' function to further track down the problem (I guess the interpolate function is using something similar). While the cell_candidates are still correct (the bounding box tree finds both elements which might be colliding with the problematic point), the compute_colliding_cells function returns an empty array of cells. I guess because it cannot decide which cell to choose, so it chooses none.

### Minimal Example (Python)

```Python
from dolfinx import io, fem, mesh, geometry
from mpi4py import MPI
import numpy as np

# Evaluate function values as implemented in
# https://jsdokken.com/dolfinx-tutorial/chapter1/membrane_code.html#making-curve-plots-throughout-the-domain
def evaluate_function(points, u):
    domain = u.function_space.mesh
    bb_tree = geometry.BoundingBoxTree(domain, domain.topology.dim)
    cells = []
    points_on_proc = []
    # Find cells whose bounding-box collide with the points
    cell_candidates = geometry.compute_collisions(bb_tree, points)
    print("Cell candidates (still correct):", cell_candidates.links(6166))
    # Choose one of the cells that contains the point
    colliding_cells = geometry.compute_colliding_cells(domain, cell_candidates, points)
    print("Colliding cells (wrong):", colliding_cells.links(6166))
    for i, point in enumerate(points):
        if len(colliding_cells.links(i))>0:
            points_on_proc.append(point)
            cells.append(colliding_cells.links(i)[0])
[...]

PR #2788, opened 2023-09-26
Make non-matching interpolation more robust and deterministic

jorgensd:
**Current approach**
1. Create process bounding box tree with padding (`bb_global`)
2. For all local interpolation points find intersection with `bb_global`
3. Send these points to those processes
4. Find cells on process whose bounding box intersect with these points
5. Find actual cell collisions ( `compute_first_colliding_cell` that has a fixed tolerance. This uses GJK that has another fixed tolerance).
6. Send these back to process owning points. This process chooses first process that has a collision (non-deterministic).
7. Send owner info to all relevant process to reduce the processes searching later on.

As we see in this approach step 5 is problematic, as they require tolerances (that currently aren't in sync, causing #2774 ), which also are tricky to tune when we get to variable precision of geometry.
Step 6 is also problematic as it depends on the MPI neighborhood comm layout.

**New approach**

1. Create process bounding box tree with padding (`bb_global`)
2. For all local interpolation points find intersection with `bb_global`
3. Send these points to those processes
4. Find cells on process whose bounding box intersect with these points
5. Find closest entity on process.
  a. Uses midpoint trees to get a good starting point in bounding box tree
  b. Finds closest entity within GJK distance (single tolerance)
7. Send these back to process owning points. This process chooses the entity with the smallest distance (this means that for points far outside the mesh we use extrapolation).
8. Send owner info to all relevant process to reduce the processes searching later on.

Currently the extrapolation is limited by the padding we give to the bounding box in
`determine_point_ownership`, but this could now trivially be extended as an input parameter, both to `determine_point_ownership` and `create_nonmatching_meshes_interpolation_data`.


jorgensd:
@massimiliano-leoni I would like you to review this as well, but cant add you as reviewer on github!

cpp/dolfinx/geometry/utils.h
```diff
@@ -791,20 +786,30 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
...
-      recv_sizes.data(), recv_offsets.data(),
-      dolfinx::MPI::mpi_type<std::int32_t>(), reverse_comm);
+      squared_distances.data(), send_sizes.data(), send_offsets.data(),
+      dolfinx::MPI::mpi_type<T>(), recv_distances.data(), recv_sizes.data(),
+      recv_offsets.data(), dolfinx::MPI::mpi_type<T>(), reverse_comm);
 
   std::vector<std::int32_t> point_owners(points.size() / 3, -1);
-  for (std::size_t i = 0; i < unpack_map.size(); i++)
+  std::vector<T> closest_distance(points.size() / 3, -1);
+  std::stringstream qq;
```
massimiliano-leoni, line 798: ?

jorgensd: Stringstream is remaining from debugging. Will remove

jorgensd, line 798:
```suggestion
```

cpp/dolfinx/geometry/utils.h
```diff
@@ -791,20 +786,29 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
...
-    // Only insert new owner if no owner has previously been found
-    if ((recv_ranks[i] >= 0) && (point_owners[pos] == -1))
-      point_owners[pos] = recv_ranks[i];
+    for (std::int32_t j = recv_offsets[i]; j < recv_offsets[i + 1]; j++)
+    {
+      const std::int32_t pos = unpack_map[j];
+      // If point has not been found yet distance is negative
+      // If new received distance smaller than current distance choose owner
+      if ((closest_distance[pos] < 0)
+          || (closest_distance[pos] > recv_distances[j]))
```
garth-wells, line 806:
```suggestion
      if ( auto d = closest_distance[pos]; d < 0 or d > recv_distances[j])
```

cpp/dolfinx/geometry/utils.h
```diff
@@ -753,18 +754,12 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
...
-  {
-    std::array<T, 3> point;
-    std::copy(std::next(received_points.begin(), p),
-              std::next(received_points.begin(), p + 3), point.begin());
-    const int colliding_cell
-        = geometry::compute_first_colliding_cell(mesh, bb, point);
-    cell_indicator[p / 3] = (colliding_cell >= 0) ? rank : -1;
-    colliding_cells[p / 3] = colliding_cell;
-  }
+  std::vector<std::int32_t> closest_cells = compute_closest_entity(
```
garth-wells, line 757: Should this be `const`?

cpp/dolfinx/geometry/utils.h
```diff
@@ -753,18 +754,12 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
...
+      bb, midpoint_tree, mesh,
+      std::span<const T>(received_points.data(), received_points.size()));
+  std::vector<T> squared_distances = squared_distance(
```
garth-wells, line 760: `const`?

cpp/dolfinx/geometry/utils.h
```diff
@@ -753,18 +754,12 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
 
   // Each process checks which points collides with a cell on the process
```
IgorBaratta, line 755: Do we need to change this comment?

IgorBaratta:
Do we need to change the documentation of the function?
It seems that the PR is changing its behavior.

cpp/dolfinx/geometry/utils.h
jorgensd, line 755:
```suggestion
  // Each process checks which local cell is closest and computes the squared distance to the cell
```

jorgensd:
Any further comments?

garth-wells:
> Any further comments?

Just hoping we can get the oneAPI CI working. It's LLVM-based, whereas other CI is gcc.

PR #2795, opened 2023-09-30
Try OneAPI on Buildjet

jhale:
This costs money so maybe only an interim solution.


garth-wells:
Working now on Actions again.

PR #2799, opened 2023-09-30
Disable mypy on petsc4py 3.20 types

garth-wells:
Somethig must have changed in PETSc.


francesco-ballarin:
For reference, this is the upstream issue that tracks this https://gitlab.com/petsc/petsc/-/issues/1367

It is a nightmare to add `# type: ignore` at every single PETSc line. In my CI I simply remove the `py.typed` file from `petsc4py` see
https://github.com/RBniCS/RBniCSx/blob/main/.github/workflows/ci.yml#L38
since the above issue establishes that `petsc4py` is not really typed. I do realize that it may not be the ideal solution, since each developer would have to remember to do that in their own `petsc4py` installation.

PR #2803, opened 2023-10-03
Let PETSc choose default LU solver within NewtonSolver

jhale:
LU is OK as a default but I would propose removing the MUMPS default, it's not been working consistently well for me recently.

This is more consistent with what was changed in LinearProblem (let PETSc handle defaults).


garth-wells (approved):

garth-wells:
Agree that it's better to go with defaults and specify the solver on a per-case basis if required, e.g. for the CH demo.

PR #2806, opened 2023-10-05
Switch default dev-env container pull to ghcr.io

jhale:
Faster and no rate limit.


jhale:
For the 0.7.0 images I am passing the ghcr image as a build time parameter.

PR #2809, opened 2023-10-06
Rename name -> __name__

mscroggs:
Following change in https://github.com/FEniCS/basix/pull/676


chrisrichardson:
I don't think this is necessary. It is true that `nanobind` does not provide `.name` by default, but it has been added in the wrapper with `.def_prop_ro` to the enums.
I think this is the "right" way, as Python native `enums` use `.name`.

mscroggs:
Some of the CI tests were failing, so this was a quick fix to get them passing again. Longer term, we should move to the better way of doing it you suggest

mscroggs:
I've undone this change in #2878. mypy was giving this error as the `name` property was missing from the pyi file in Basix

PR #2820, opened 2023-10-18
Switch to  nanobind

garth-wells:
This PR switches from `pybind11` to `nanobind` for generating the Python wrappers. `nanobind` is smaller than `pybind11`, the wrappers build faster and it has significantly improved support for wrapping multi-dimensional arrays, which we use heavily.

The `nanobind` docs are easier to follow on the low-level details, which makes understanding the memory management in the wrapper layer easier.

~~PR depends on the `nanobind` development version, which includes fixes for complex-valued arrays.~~ Fixed in `nanobind` 1.8.0.


chrisrichardson:
There is an issue with docstrings, and `__name__` which is not being set correctly. See: https://github.com/wjakob/nanobind/issues/349

This is affecting documentation.

python/CMakeLists.txt
```diff
@@ -1,43 +1,26 @@
...
-  CONFIG
-  HINTS
-  ${PYBIND11_CMAKE_DIR}
-  ${PYBIND11_DIR}
-  ${PYBIND11_ROOT}
-  $ENV{PYBIND11_DIR}
-  $ENV{PYBIND11_ROOT}
-)
+# Detect the installed nanobind package and import it into CMake
+execute_process(
```
jhale, line 19: nanobind only supports a python install?

.github/workflows/ccpp.yml
```diff
@@ -151,7 +151,7 @@ jobs:
           python3 -m pip -v install --check-build-dependencies --no-build-isolation python/
       - name: Build Python interface documentation
         run: |
-          python3 -m pip install sphinx==5.0.2 # TO REMOVE
+          python3 -m pip install sphinx
```
jhale, line 154: There is still a sphinx pinned at 5.0.2 in the `Dockerfile.test-env` which I think should be removed. Would recommend adding `# TO REMOVE` so this can be cleaned up later.

cpp/dolfinx/la/petsc.cpp
```diff
@@ -131,8 +131,8 @@ std::vector<IS> la::petsc::create_index_sets(
   std::int64_t offset = 0;
   for (auto& map : maps)
   {
-    const int bs = map.second;
-    const std::int32_t size
+    int bs = map.second;
+    std::int32_t size
```
jhale, line 135: Why remove `const`?

python/dolfinx/wrappers/graph.cpp
```diff
@@ -35,65 +37,78 @@ namespace dolfinx_wrappers
...
+                dolfinx::graph::regular_adjacency_list(std::move(data), 1));
+          },
+          nb::arg("adj").noconvert())
+      .def(
+          "__init__",
+          [](dolfinx::graph::AdjacencyList<T>* a,
+             nb::ndarray<const T, nb::ndim<2>, nb::c_contig> adj)
+          {
+            std::vector<T> data(adj.data(), adj.data() + adj.size());
+            new (a) dolfinx::graph::AdjacencyList<T>(
```
jhale, line 61: Is using the `new` keyword necessary? The old approach looked cleaner.

python/test/unit/mesh/test_higher_order_mesh.py
```diff
@@ -635,7 +634,7 @@ def test_gmsh_input_2d(order, cell_type):
         gmsh_cell_id = gmsh.model.mesh.getElementType("quadrangle", order)
     gmsh.finalize()
 
-    cells = cells[:, cell_perm_array(cell_type, cells.shape[1])]
+    cells = cells[:, cell_perm_array(cell_type, cells.shape[1])].copy()
```
jhale, line 637: Comment on why this copy is necessary would be good (I guess contiguity?).

python/test/unit/fem/test_function.py
```diff
@@ -173,7 +173,7 @@ def f(x):
     assert round(w.x.norm(la.Norm.l1) - 6 * num_vertices, 7) == 0
 
 
-@pytest.mark.parametrize("types", [
+@ pytest.mark.parametrize("types", [
     # (np.float32, "float"),  # Fails on Redhat CI, needs further investigation
```
jhale, line 177: We should file an issue on this, also noticed it.

python/CMakeLists.txt
garth-wells: See https://nanobind.readthedocs.io/en/latest/building.html#finding-nanobind.

.github/workflows/ccpp.yml
garth-wells: @chrisrichardson this change came from you - can you comment?

cpp/dolfinx/la/petsc.cpp
garth-wells: Syntactic noise in this case.

python/dolfinx/wrappers/graph.cpp
garth-wells: Yes. `nanobind` uses 'new placement' for construction.

python/test/unit/mesh/test_higher_order_mesh.py
garth-wells: I don't think we really need a comment on this in the test.

Yes, the issue is contiguous storage.

.github/workflows/ccpp.yml
chrisrichardson: I tried removing it, and it seems OK.

python/test/unit/mesh/test_higher_order_mesh.py
chrisrichardson: This was causing crashes without a copy. I think permuting the indices is just a "view" which does not work when accessing memory directly.

python/test/unit/fem/test_function.py
garth-wells: Re-enabled, works.

python/CMakeLists.txt
```diff
@@ -75,22 +58,15 @@ include(CheckCXXCompilerFlag)
...
 
 if(HAVE_PEDANTIC)
-  target_compile_options(cpp PRIVATE -Wall;-Werror;-pedantic)
-endif()
-
-# In Debug mode override pybind11 symbols visibility Symbols must be visible to
-# backtrace_symbols() to produce nice logs
-if(CMAKE_BUILD_TYPE STREQUAL "Debug")
-  target_compile_options(cpp PRIVATE "-fvisibility=default")
+  # target_compile_options(cpp PRIVATE -Wall;-Werror;-pedantic)
```
michalhabera, line 61: Leftover comment. I wonder how does symbols visibility (and readable backtrace) work with nanobind?

python/dolfinx/wrappers/geometry.cpp
```diff
@@ -12,274 +12,185 @@
...
       },
-      py::arg("p"), py::arg("q"));
+      //   nb::rv_policy::copy,
+      nb::arg("p"), nb::arg("q"));
 
   m.def(
       "squared_distance",
       [](const dolfinx::mesh::Mesh<T>& mesh, int dim,
-         std::vector<std::int32_t> indices, const py::array_t<T>& points)
+         const std::vector<std::int32_t>& indices,
```
michalhabera, line 171: Shoult this be rather nb::ndarray (to avoid the python list inefficiency)?

python/dolfinx/wrappers/graph.cpp
```diff
@@ -108,42 +123,43 @@ void graph(py::module& m)
...
-      {
-        return create_partitioner_py(dolfinx::graph::kahip::partitioner(
-            mode, seed, imbalance, suppress_output));
-      },
-      py::arg("mode") = 1, py::arg("seed") = 1, py::arg("imbalance") = 0.03,
-      py::arg("suppress_output") = true, "KaHIP graph partitioner");
-#endif
 
-  m.def("reorder_gps", &dolfinx::graph::reorder_gps, py::arg("graph"));
+  // #ifdef HAS_PTSCOTCH
```
michalhabera, line 127: Why are these commented?

python/CMakeLists.txt
garth-wells: I'll leave this in because we'd like to turn it on again. Need to figure out how to use `isystem` for the nanobind headers.

python/dolfinx/wrappers/geometry.cpp
garth-wells: Yes, and in quite a few mored places in the geometry interface. It wasn't done right before with pybind11, so this PR doesn't really change that.

We should check in nanobind is faster for `std::vector` than pybind11.

python/dolfinx/wrappers/graph.cpp
garth-wells: No idea. Will re-enable.

python/test/unit/fem/test_assemble_cppimport.py
```diff
@@ -10,8 +10,8 @@
 from mpi4py import MPI
 
 import cppimport
```
chrisrichardson, line 12: Should we just delete this test? It currently adds a dependency (test) on cppimport, even though we are skipping...

.github/workflows/oneapi.yml
```diff
@@ -99,14 +100,12 @@ jobs:
...
         run: |
           . /opt/intel/oneapi/setvars.sh
-          pytest -v -m mpi --num-proc=2 python/demo/test.py
+          pytest -m mpi --num-proc=2 python/demo/test.py
       - name: Run DOLFINx Python unit tests (serial)
         run: |
           . /opt/intel/oneapi/setvars.sh
-          cd python/test
-          pytest -n=auto --durations=50 unit/
+          pytest -vs -n=auto --durations=50 python/test/unit/fem
```
francesco-ballarin, line 107: Will limiting the tests to `fem` be reverted before merge?

(similarly for the mpi case below)

docker/Dockerfile.test-env
```diff
@@ -153,8 +153,8 @@ RUN if [ "$MPI" = "mpich" ]; then \
 # - Second set of packages are recommended and/or required to build
 #   documentation or run tests.
 RUN pip3 install --no-cache-dir --upgrade setuptools pip && \
-    pip3 install --no-cache-dir cffi mpi4py numba numpy==${NUMPY_VERSION} scikit-build-core[pyproject] && \
-    pip3 install --no-cache-dir breathe clang-format cppimport cmakelang flake8 isort jupytext matplotlib mypy myst-parser pybind11==${PYBIND11_VERSION} pytest pytest-xdist scipy sphinx==5.0.2 sphinx_rtd_theme types-setuptools
+    pip3 install  -no-cache-dir cffi mpi4py numba numpy==${NUMPY_VERSION} scikit-build-core[pyproject] && \
```
francesco-ballarin, line 156: Please add back the missing `-`.

python/CMakeLists.txt
```diff
@@ -1,43 +1,26 @@
...
 set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
 
-find_package(
-  Python3
-  COMPONENTS Interpreter Development.Module
-  REQUIRED
-)
+# set(CMAKE_BUILD_TYPE Debug)
```
francesco-ballarin, line 10: I guess this commented line can be removed, since the default `CMAKE_BUILD_TYPE` is set anyways two lines afterwards.

python/doc/source/conf.py
```diff
@@ -101,7 +102,7 @@
 # -- Options for HTMLHelp output ------------------------------------------
 
 # Output file base name for HTML help builder.
-htmlhelp_basename = 'DOLFINdoc'
+htmlhelp_basename = "DOLFINdoc"
```
francesco-ballarin, line 105: `DOLFINxdoc`?

python/dolfinx/wrappers/caster_petsc.h
```diff
@@ -1,30 +1,37 @@
...
 
-// pybind11 casters for PETSc/petsc4py objects
+// nanobind casters for PETSc/petsc4py objects
 
-namespace py = pybind11;
+namespace nb = nanobind;
 
 // Import petsc4py on demand
-#define VERIFY_PETSC4PY(func)                                                  \
+#define VERIFY_PETSC4PY0(func)                                                 \
```
francesco-ballarin, line 23: `VERIFY_PETSC4PY0` -> `VERIFY_PETSC4PY_BOOL_RETURN`? At first glance, the difference between the two was not evident to me.

python/dolfinx/wrappers/caster_petsc.h
```diff
@@ -1,30 +1,37 @@
...
   if (!func)                                                                   \
   {                                                                            \
     if (import_petsc4py() != 0)                                                \
-      throw std::runtime_error("Error when importing petsc4py");               \
+      return false;                                                            \
+  }
+
+#define VERIFY_PETSC4PY1(func)                                                 \
```
francesco-ballarin, line 30: `VERIFY_PETSC4PY1` -> `VERIFY_PETSC4PY_HANDLE_RETURN`? At first glance, the difference between the two was not evident to me.

python/dolfinx/wrappers/geometry.cpp
```diff
@@ -12,274 +12,185 @@
...
           [](const dolfinx::geometry::BoundingBoxTree<T>& self,
              const std::size_t i)
           {
             std::array<T, 6> bbox = self.get_bbox(i);
-            std::array<std::size_t, 2> shape{2, 3};
-            return py::array_t<T>(shape, bbox.data());
+            return dolfinx_wrappers::as_nbarray_copy(bbox, {2, 3});
           },
-          py::arg("i"))
+          nb::arg("i)"))
```
francesco-ballarin, line 57: The newly added parenthesis seems wrong.

python/test/unit/fem/test_assemble_cppimport.py
francesco-ballarin: I agree.

python/test/unit/la/test_casters.py
```diff
@@ -6,86 +6,89 @@
 #
 # SPDX-License-Identifier:    LGPL-3.0-or-later
 
-import pathlib
```
francesco-ballarin, line 9: If the comment by @chrisrichardson in `python/test/unit/fem/test_assemble_cppimport.py` is approved, this whole file could be deleted as well.

python/dolfinx/wrappers/geometry.cpp
michalhabera: I run a small benchmark. Nanobind is slightly (10%) faster for small arrays (<10^5), but for larger arrays this is simply the cost of `list(numpy_array)` that dominates, so no changes.

garth-wells:
> There is an issue with docstrings, and `__name__` which is not being set correctly. See: [wjakob/nanobind#349](https://github.com/wjakob/nanobind/issues/349)
> 
> This is affecting documentation.

Fixed in nanobind (https://github.com/wjakob/nanobind/commit/3c2b64a0ccde7ff09c28dfe20af0d5de96c0f364).

chrisrichardson (approved):

jhale (approved):

wjakob:
Now that this process has concluded, I was wondering if you have statistics to share? Do you observe changes in compilation time/binary size/runtime overheads for the python binding component?

chrisrichardson:
@wjakob - the binary size has decreased from 4052784 to 2157624 for the Python binding module, which is good. The compilation time also, subjectively feels faster - the LTO was always very slow for pybind.
I'm not sure we have timing statistics for the Python layer, as such - but we can perhaps have a look at it. We have recipes now for `py-fenics-dolfinx@0.7.1` and `py-fenics-dolfin@main` in https://github.com/FEniCS/spack/tree/latest which makes it relatively easy to compare the two.

garth-wells:
@wjakob very appealing is ndarray support in nanobind.

chrisrichardson:
I think we still have to think about the full potential of interoperation with other backends (not just numpy)

PR #2822, opened 2023-10-23
Improve documentation of `pack_coefficients`

jpdean:


cpp/dolfinx/fem/utils.h
```diff
@@ -1145,7 +1145,9 @@ Expression<T, U> create_expression(
 /// @brief Pack coefficients of a Form
 /// @param[in] form The Form
 /// @param[in] coeffs A map from a (integral_type, domain_id) pair to a
-/// (coeffs, cstride) pair
+/// (coeffs, cstride) pair, where coeffs is a storage container in which
```
garth-wells, line 1148: So `coeffs` represents a 2D array? What's the shape?

jpdean: Is it any better now?

PR #2826, opened 2023-10-25
Import mpi4py early for MPI initialisation

garth-wells:
Avoid later initialisation of MPI with possibly wrong thread support level.


jhale (approved):
Thanks Garth and isort!

PR #2834, opened 2023-10-28
[nanobind] Remove a few more occurences of pybind11

francesco-ballarin:


docker/Dockerfile.oneapi
```diff
@@ -13,7 +13,7 @@ ARG HDF5_PATCH=2
 ARG HDF5_FIX=
 ARG PETSC_VERSION=3.20.0
 ARG SLEPC_VERSION=3.20.0
-ARG PYBIND11_VERSION=2.11.1
+ARG NANOBIND_VERSION=1.7.0
```
francesco-ballarin, line 16: This, and similar ones in the remaining docker files, will need to be bumped upon the next nanobind release

python/dolfinx/wrappers/common.cpp
```diff
@@ -41,12 +41,6 @@ void common(nb::module_& m)
   m.attr("has_kahip") = dolfinx::has_kahip();
   m.attr("has_slepc") = dolfinx::has_slepc();
   m.attr("has_adios2") = dolfinx::has_adios2();
-
-#ifdef HAS_PYBIND11_SLEPC4PY
```
francesco-ballarin, line 45: I think this has been broken for a while, and is not used elsewhere. I tested on my local installation and it gave `False`, even though I surely have `slepc4py` installed.

python/README.rst
```diff
@@ -1,26 +1,18 @@
 DOLFINx Python interface
```
francesco-ballarin, line 1: The content of this file will be replaced with the one coming from #2707 anyways.

python/dolfinx/wrappers/common.cpp
francesco-ballarin: Furthermore, even on `main`, I can't find any trace of the macro `HAS_PYBIND11_SLEPC4PY` ever being defined.

francesco-ballarin:
`test_basix_element` in `test/unit/test/test_function_space.py` still fails, even if Basix and DOLFINx both using nanobind, with

```
    @pytest.mark.skipif(default_real_type != np.float64, reason="float32 not supported yet")
    def test_basix_element(V, W, Q, V2):
        for V_ in (V, W, V2):
>           e = V_.element.basix_element
E           TypeError: Unable to convert function return value to a Python type! The signature was
E               (self) -> basix::FiniteElement<double>

```

The note
```
NOTE: Test relies on Basix and DOLFINx both using pybind11
```
may need to be updated.

PR #2838, opened 2023-10-30
Add `flake8` and `isort` checks on python files in `cpp` folder

francesco-ballarin:
There are some python files in the `cpp` folder, but they were not passed through `flake8` or `isort` on CI. For the time being there is no point in running `mypy` on them because `ufl` is not typed.


cpp/.flake8
garth-wells, line 1: Could we move this file to the top level directory and remove `python/.flake8`?

.github/workflows/ccpp.yml
```diff
@@ -69,13 +69,21 @@ jobs:
         run: pip3 install matplotlib
       - name: Flake8 checks
         run: |
+          cd cpp/
```
garth-wells, line 72: Could we just do `python3 -m flake8 cpp/`?

cpp/.flake8
francesco-ballarin: Done

francesco-ballarin: `python/.flake8` is not there yet, it will be once #2707 is merged.

.github/workflows/ccpp.yml
francesco-ballarin: It depends: do you want `cpp/doc/source/conf.py` to be tested too?

garth-wells: I have a vague recollection that the order of the imports in `cpp/doc/source/conf.py` is important.

francesco-ballarin: I am not sure if that is a yes or a no to my question 😄 
* if you want `cpp/doc/source/conf.py` to be tested, then we could move to `python3 -m flake8 cpp/`, but then it stands to reason that `python/doc/source/conf.py` should be tested too, and thus we should try to move to `python3 -m flake8 python/` too.
* if you do not want `cpp/doc/source/conf.py` to be tested, then we cannot just do `python3 -m flake8 cpp/`.

garth-wells: Answer is do not use `isort` on `cpp/doc/source/conf.py` without testing.

garth-wells (approved):

PR #2844, opened 2023-11-01
Install python build requirements in docker end user images

francesco-ballarin:
Testing in https://github.com/FEniCS/dolfinx/actions/runs/6717503005

### DIFF
--- a/docker/Dockerfile.end-user
+++ b/docker/Dockerfile.end-user
@@ -90,7 +90,8 @@ ONBUILD RUN cd dolfinx && \
     PETSC_ARCH=linux-gnu-real64-32 cmake -G Ninja -DCMAKE_INSTALL_PREFIX=/usr/local/dolfinx-real -DCMAKE_BUILD_TYPE=${DOLFINX_CMAKE_BUILD_TYPE} -DCMAKE_CXX_FLAGS=${DOLFINX_CMAKE_CXX_FLAGS} ../cpp && \
     ninja install && \
     cd ../python && \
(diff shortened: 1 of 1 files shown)


francesco-ballarin:
I don't understand what is going on with the build at https://github.com/FEniCS/dolfinx/actions/runs/6717503005:
* the arm build https://github.com/FEniCS/dolfinx/actions/runs/6717503005/job/18255469197 failed, but no compilation errors are reported in the log. Maybe we run out of build minutes on `buildjet`?
* the amd64 build https://github.com/FEniCS/dolfinx/actions/runs/6717503005/job/18255469014 did test successfully the change, but the it was stopped when the other one failed.

----------

A different solution would be install build requirements in the test/dev image. This would allow to remove the line
```
python3 -m pip install -r python/build-requirements.txt # TO REMOVE
```
from `.github/workflows/ccpp.yml`. However, I am not convinced that it is a good idea because:
1. `.github/workflows/ccpp.yml` is the main CI workflow, which means that changes to `python/build-requirements.txt` would go untested
2. it would break the symmetry with `.github/workflows/oneapi.yml` and `.github/workflows/redhat.yml`

francesco-ballarin:
Opened https://github.com/FEniCS/dolfinx/issues/2846 for the workflow failure.

garth-wells:
We like to avoid `python3 -m pip install -r python/build-requirements.txt` and to install dependencies directly so we keep control over versioning. Otherwise we can be testing code changes and dependency changes at the same time. It's also nice if the Docker images are the same from a given Dockerfile (at least for the parts we care about) irrespective of the time at which the image is built.

The Redhat and OneAPI CI balance this out somewhat by installing more automatically. This gives a slightly different test coverage. These test are 'secondary' and the images are not intended to be user facing, so we control versions less tightly.

francesco-ballarin:
Closing this in a favor of a future one which addresses https://github.com/FEniCS/dolfinx/pull/2844#issuecomment-1789324585.

PR #2850, opened 2023-11-02
Fix bounding box leaf collision check

jorgensd:
Reported by @nate-sime 
The following MWE produces the wrong collision for the global_bb_tree
```python
import numpy as np
import dolfinx
from mpi4py import MPI

xp = np.array([[100.0, 100.0, 0.0]], dtype=np.double)

tree = dolfinx.geometry.bb_tree(mesh, mesh.topology.dim, padding=1e-4)
global_tree = tree.create_global_tree(mesh.comm)


def pprint(*msg):
    print(f"[{mesh.comm.rank}]: " + " ".join(map(str, msg)), flush=True)


pprint(dolfinx.common.git_commit_hash)
pprint(dolfinx.geometry.compute_collisions_points(tree, xp))
pprint(dolfinx.geometry.compute_collisions_points(global_tree, xp))

```
giving the following in serial:
```bash
[0]: f488f4ecc687d0065315bc9a74939634b18e11ff
[0]: <AdjacencyList> with 1 nodes
  0: []

[0]: <AdjacencyList> with 1 nodes
  0: [0 ]
```
and the correct result in parallel
```bash
[0]: f488f4ecc687d0065315bc9a74939634b18e11ff
[0]: <AdjacencyList> with 1 nodes
  0: []

[0]: <AdjacencyList> with 1 nodes
  0: []

[1]: f488f4ecc687d0065315bc9a74939634b18e11ff
[1]: <AdjacencyList> with 1 nodes
  0: []

[1]: <AdjacencyList> with 1 nodes
  0: []
```

This is due to a bug in `_compute_collisions_point` where we don't check if the leaf node actually collides with the point.
@massimiliano-leoni could you see if this impacts your mesh interpolation (in serial)? As the `global_bb_tree` is used it would add any interpolation point from the other mesh to the process, even if it doesn't collide.

### DIFF
--- a/cpp/dolfinx/geometry/utils.h
+++ b/cpp/dolfinx/geometry/utils.h
@@ -300,8 +300,7 @@ void _compute_collisions_point(const geometry::BoundingBoxTree<T>& tree,
   while (next != -1)
   {
     const std::array<int, 2> bbox = tree.bbox(next);
-    std::int32_t current_bbox = next;
-    if (is_leaf(bbox) and point_in_bbox(tree.get_bbox(current_bbox), p))
+    if (is_leaf(bbox) and point_in_bbox(tree.get_bbox(next), p))
     {
       // If box is a leaf node then add it to the list of colliding
       // entities
--- a/cpp/dolfinx/geometry/utils.h
+++ b/cpp/dolfinx/geometry/utils.h
@@ -303,7 +303,7 @@ void _compute_collisions_point(const geometry::BoundingBoxTree<T>& tree,
     const std::array<int, 2> bbox = tree.bbox(next);
     next = -1;
 
-    if (is_leaf(bbox))
+    if ((is_leaf(bbox)) && point_in_bbox(tree.get_bbox(next), p))
     {
       // If box is a leaf node then add it to the list of colliding entities
       entities.push_back(bbox[1]);
@@ -683,7 +683,16 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
   // Compute collisions:
   // For each point in `x` get the processes it should be sent to
   graph::AdjacencyList collisions = compute_collisions(global_bbtree, points);
-
+  for (std::size_t i = 0; i < collisions.num_nodes(); ++i)
+  {
+    std::cout << points[3 * i] << " " << points[3 * i + 1] << " "
+              << points[3 * i + 2] << ": ";
+    for (auto col : collisions.links(i))
+    {
+      std::cout << col << "  ";
+    }
+    std::cout << std::endl;
+  }
   // Get unique list of outgoing ranks
   std::vector<std::int32_t> out_ranks = collisions.array();
   std::sort(out_ranks.begin(), out_ranks.end());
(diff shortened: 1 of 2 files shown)


cpp/dolfinx/geometry/utils.h
```diff
@@ -303,7 +303,7 @@ void _compute_collisions_point(const geometry::BoundingBoxTree<T>& tree,
     const std::array<int, 2> bbox = tree.bbox(next);
     next = -1;
 
-    if (is_leaf(bbox))
+    if ((is_leaf(bbox)) && point_in_bbox(tree.get_bbox(next), p))
```
garth-wells, line 306: We and `and` rather than `&&`.

cpp/dolfinx/geometry/utils.h
```diff
@@ -683,7 +683,16 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
   // Compute collisions:
   // For each point in `x` get the processes it should be sent to
   graph::AdjacencyList collisions = compute_collisions(global_bbtree, points);
-
+  for (std::size_t i = 0; i < collisions.num_nodes(); ++i)
```
garth-wells, line 686: Remove debug output.

cpp/dolfinx/geometry/utils.h
```diff
@@ -711,7 +720,9 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
   for (std::size_t i = 0; i < points.size() / 3; ++i)
     for (auto p : collisions.links(i))
       send_sizes[rank_to_neighbor[p]] += 3;
-
+  for (auto rank : send_sizes)
```
garth-wells, line 723: Remove.

garth-wells (changes requested):
Could the PR have a more descriptive title?

And could a test be added?

cpp/dolfinx/geometry/utils.h
```diff
@@ -679,17 +679,19 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
   BoundingBoxTree bb(mesh, tdim, cells, padding);
   BoundingBoxTree midpoint_tree = create_midpoint_tree(mesh, tdim, cells);
   BoundingBoxTree global_bbtree = bb.create_global_tree(comm);
+  std::cout << "HERE!\n";
```
jorgensd, line 682:
```suggestion
```

cpp/dolfinx/geometry/utils.h
```diff
@@ -679,17 +679,19 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
...
-
   // Get unique list of outgoing ranks
   std::vector<std::int32_t> out_ranks = collisions.array();
   std::sort(out_ranks.begin(), out_ranks.end());
   out_ranks.erase(std::unique(out_ranks.begin(), out_ranks.end()),
                   out_ranks.end());
-
+  for (auto rank : out_ranks)
+    std::cout << rank << " ";
+  std::cout << "\n";
```
jorgensd, line 694:
```suggestion

```

garth-wells (approved):

IgorBaratta (approved):

nate-sime (approved):

jorgensd:
@garth-wells  Test added + minor simplification. Feel free to merge this once the CI passes.

PR #2854, opened 2023-11-02
Remove installation of build requirements from workflows, since they are already installed in the underlying docker images

francesco-ballarin:
Turns out that most of this was already available in the docker files, but we did not trigger test/dev env workflow.
See https://github.com/FEniCS/dolfinx/actions/runs/6729006694

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -86,7 +86,6 @@ jobs:
           python3 -m isort --check .
       - name: mypy checks
         run: |
-          python3 -m pip install mypy types-setuptools --upgrade # TO REMOVE
           cd python/
           mypy dolfinx
           mypy demo
@@ -118,7 +117,6 @@ jobs:
 
       - name: Build C++ interface documentation
         run: |
-          python3 -m pip install breathe # TO REMOVE
           export DOLFINX_VERSION=`cmake -L build | grep DOXYGEN_DOLFINX_VERSION | cut -f2 -d "="`
           echo $DOLFINX_VERSION
           cd cpp/doc
@@ -147,11 +145,9 @@ jobs:
 
       - name: Build Python interface
         run: |
-          python3 -m pip install -r python/build-requirements.txt # TO REMOVE
           python3 -m pip -v install --check-build-dependencies --no-build-isolation python/
       - name: Build Python interface documentation
         run: |
-          python3 -m pip install sphinx==5.0.2 # TO REMOVE
           cd python/doc
           make html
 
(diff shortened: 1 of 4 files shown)


francesco-ballarin:
I need to manually rerun https://github.com/FEniCS/dolfinx/actions/workflows/docker-oneapi-test-env.yml and https://github.com/FEniCS/dolfinx/actions/workflows/docker-redhat-test-env.yml for this to work, but I'll wait doing that until the manual rerun by @jhale is finished.

francesco-ballarin:
All base images (ubuntu CI, oneapi and redhat) have been regenerated, and now all workflow are passing.

.github/workflows/ccpp.yml
```diff
@@ -86,7 +86,6 @@ jobs:
           python3 -m isort --check .
       - name: mypy checks
         run: |
-          python3 -m pip install mypy types-setuptools --upgrade # TO REMOVE
```
francesco-ballarin, line 89: The current workflow upgrades `mypy` at every run, i.e. if the version of `mypy` on `ghcr.io/fenics/test-env:current-openmpi` is old it will get upgraded. 

From https://github.com/FEniCS/dolfinx/actions/workflows/docker-dev-test-env.yml I see that the  `docker-dev-test-env.yml` is manually triggered once a month or so, so I do not expect the version of `mypy` will ever be too old.

jhale: The workflows should not be using `pip` to install/upgrade dependencies or related tools. Everything should be in the appropriate `Dockerfile`.

Sometimes it's necessary to install something in a workflow to get the tests to pass and the PR through. The images should then be rebuilt and a cleanup PR created. 

It's good practice to add a note e.g. `# TO REMOVE` as it's easy to forget.

francesco-ballarin: The only remaining `pip install` in these three workflows is https://github.com/FEniCS/dolfinx/blame/ef7255b655e96c88ef9556efa982df1c00987cf7/.github/workflows/redhat.yml#L27
which was just added 3 days ago. Do you want me to get rid of that as well?

jhale: > The current workflow upgrades `mypy` at every run, i.e. if the version of `mypy` on `ghcr.io/fenics/test-env:current-openmpi` is old it will get upgraded. 
> 
> 
> 
> From https://github.com/FEniCS/dolfinx/actions/workflows/docker-dev-test-env.yml I see that the  `docker-dev-test-env.yml` workflow is manually triggered once a month or so, so I do not expect the version of `mypy` will ever be too old.

You're welcome to try. Would be interesting to see if redhat's pip is up to the task. If not we need to adjust the Dockerfile.

PR #2856, opened 2023-11-02
Remove unnecessary calls to `compute_graph_edges`

IgorBaratta:
The IndexMap has all the information required to compute shared_indices.
Also this PR simplifies the logic behing the computation of shared_indices().

### DIFF
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -8,6 +8,7 @@
 #include "sort.h"
 #include <algorithm>
 #include <functional>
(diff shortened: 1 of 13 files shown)


cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -8,6 +8,7 @@
 #include "sort.h"
 #include <algorithm>
 #include <functional>
+#include <iostream>
```
IgorBaratta, line 11:
```suggestion
```

garth-wells (approved):

PR #2857, opened 2023-11-02
Move upgrade of pip/setuptools from redhat workflow to redhat docker file, and remove installation of build requirements from workflows

francesco-ballarin:


garth-wells:
Let's wait on the nanobind PR - otherwise we continue with ping-pong of pip instal pybind11/nanobind.

garth-wells (approved):

Issue #2786, opened 2023-09-26
Add extrapolation to interpolation over non-matching meshes


jorgensd:
### Describe new/missing feature

Currently, we only evaluate interpolation points that fall inside the cell by a certain tolerance.
In many cases, when this distance is really small (and small GJK tolerance), we might not find a cell at all, even if we are very close to its boundary, ref: https://github.com/FEniCS/dolfinx/issues/2774

Using `compute_closest_entity` instead would give the algorithm all the distances, and it would choose the closest one as owner. 

The tolerance for considering a point owned could be an input to `dolfinx::geometry::determine_point_ownership`, which should be set to a large number if extrapolation is expected for all points, and a smaller value if one want more control over how far to extrapolate.

### Suggestion user interface

```python3
# For extrapolation
create_nonmatching_meshes_interpolation_data(geometry0, elemnt0, mesh1, cells, allowable_distance=std::numeric_limits<T>::max())

# For close control
create_nonmatching_meshes_interpolation_data(geometry0, elemnt0, mesh1, cells, allowable_distance=std::numeric_limits<T>::epsilon())

# For lenient control
create_nonmatching_meshes_interpolation_data(geometry0, elemnt0, mesh1, cells, allowable_distance=0.01)
```

Issue #2847, opened 2023-11-01
New non-matching meshes interpolation algorithm is extremely expensive


massimiliano-leoni:
### How to reproduce the bug

Commit c72113a98f00af53420a5fe034f7e258636d37d8 changed the non-matching-meshes interpolation algorithm to improve on robustness but this came at a then-undetected dramatic increase in computational cost.
I realised this when building interpolation data for two unstructured meshes, that I used many times in the past, went from taking 5.3 seconds to a baffling 734 seconds. Dare I say it, _preposterous_! D:

Early tests show that the overhead increases with the number of parallel processes but I haven't fully confirmed this.

I'm attaching a C++ file that allows to reproduce the issue using two overlapping cubes [remove the .txt extension].
To this end you are going to need
`basix` at revision `1117a8d98108e9ac156a6a7fab82bbf8800a4cbd`
`ufl` at revision `77ae57c4b9594851f6bff4a63b7eecb81477c486`
`ffcx` at revision `1d2723891bdc990334133bf78362cc601b784089`

and then you can use dolfinx's revisions c72113a98f00af53420a5fe034f7e258636d37d8 and a88a52e594d37b20da017ef6a5c8e0a945f747cc as test and control cases.

### Version

main branch

### Additional information

[interpolation.cpp.txt](https://github.com/FEniCS/dolfinx/files/13225909/interpolation.cpp.txt)

massimiliano-leoni:
@jorgensd I think this can be closed now.

PR #2858, opened 2023-11-02
Speed up non-matching interpolation data and add extrapolation parameter

jorgensd:
Following the regression documented in #2847 I propose that we make extrapolation distance an input parameter (padding), which is used to pad the bounding boxes used to find processes with potential collisions.

The new code will first do as the original interpolation data did:
- Given a set of interpolation points (locally to each process) create a padded bounding box on the mesh we want to interpolate to.
- Find what processes has the interpolation points, and distribute data so we can use GJK on those processes to find the first colliding cell.
- Send this cell back to the interpolation point owner

Then the strategy is:
- For points that have not gotten an owner assigned (and are within the padded bounding box),  compute the closest entity  and use this as owner.

@massimiliano-leoni  could you test this branch with your example, setting the padding first to 0, then to 1e-10, 1e-6, 1e-2 etc.?

For performance impact, see: #2862


cpp/dolfinx/geometry/utils.h
```diff
@@ -847,6 +937,7 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
   owned_recv_ranks.reserve(recv_offsets.back());
   std::vector<T> owned_recv_points;
   std::vector<std::int32_t> owned_recv_cells;
+
```
jorgensd, line 940:
```suggestion
```

jorgensd:
This PR would also resolve #2786

cpp/dolfinx/geometry/utils.h
```diff
@@ -758,15 +758,32 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
...
-  const std::vector<T> squared_distances = squared_distance(
-      mesh, tdim, closest_cells,
-      std::span<const T>(received_points.data(), received_points.size()));
+  std::vector<std::int32_t> cell_indicator(received_points.size() / 3);
+  std::vector<std::int32_t> closest_cells(received_points.size() / 3);
+  for (std::size_t p = 0; p < received_points.size(); p += 3)
+  {
+    std::array<T, 3> point;
+    std::copy_n(std::next(received_points.begin(), p), 3, point.begin());
+    const int colliding_cell = geometry::compute_first_colliding_cell(
```
garth-wells, line 781: This code isn't easy to read. Could it be simplified?

cpp/dolfinx/geometry/utils.h
```diff
@@ -793,26 +810,105 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
     std::swap(recv_offsets, send_offsets);
   }
 
+  std::vector<std::int32_t> recv_ranks(recv_offsets.back());
+  MPI_Neighbor_alltoallv(
+      cell_indicator.data(), send_sizes.data(), send_offsets.data(),
+      dolfinx::MPI::mpi_type<std::int32_t>(), recv_ranks.data(),
```
garth-wells, line 816: `MPI_INT32_T`

cpp/dolfinx/geometry/utils.h
```diff
@@ -793,26 +810,105 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
...
+      recv_sizes.data(), recv_offsets.data(),
+      dolfinx::MPI::mpi_type<std::int32_t>(), reverse_comm);
```
garth-wells, line 818: `MPI_INT32_T`

cpp/dolfinx/geometry/utils.h
```diff
@@ -793,26 +810,105 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
...
+
+  std::vector<std::int32_t> point_owners(points.size() / 3, -1);
+  for (std::size_t i = 0; i < unpack_map.size(); i++)
+  {
+    const std::int32_t pos = unpack_map[i];
+    // Only insert new owner if no owner has previously been found
+    if ((recv_ranks[i] >= 0) && (point_owners[pos] == -1))
```
garth-wells, line 825: No need for the extra parentheses.

cpp/dolfinx/geometry/utils.h
```diff
@@ -793,26 +810,105 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
...
+      point_owners[pos] = recv_ranks[i];
+  }
+
+  // Create extrapolation marker for those points already sent to other
+  // process
+  std::vector<std::uint8_t> send_extrapolate(recv_offsets.back());
+  for (std::int32_t i = 0; i < recv_offsets.back(); i++)
+  {
+    const std::int32_t pos = unpack_map[i];
+    send_extrapolate[i] = std::uint8_t(point_owners[pos] == -1);
```
garth-wells, line 835: Is explicit conversion required?

cpp/dolfinx/geometry/utils.h
```diff
@@ -793,26 +810,105 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points)
...
   for (std::size_t i = 0; i < out_ranks.size(); i++)
   {
     for (std::int32_t j = recv_offsets[i]; j < recv_offsets[i + 1]; j++)
     {
       const std::int32_t pos = unpack_map[j];
-      // If point has not been found yet distance is negative. If new
-      // received distance smaller than current distance choose owner.
-      if (auto d = closest_distance[pos]; d < 0 or d > recv_distances[j])
+      auto current_dist = recv_distances[j];
+      // Only update if closer than previous guess and was found
```
garth-wells, line 906: Remove 'Only"

garth-wells (dismissed):
I think use of the work 'extrapolation' can be tightened up, which will help understanding. 'Extrapolation' is used in the context of geometric operations, which doesn't seem right.

massimiliano-leoni:
I just tried the latest revision in this branch and I can confirm that the performance is again satisfying.
On the flip side, I tried setting the padding to `1` or `1e-1` just to see what happens and I noticed an explosion in RAM usage. In fact, I crashed my own machine :_D

jorgensd:
> I just tried the latest revision in this branch and I can confirm that the performance is again satisfying. On the flip side, I tried setting the padding to `1` or `1e-1` just to see what happens and I noticed an explosion in RAM usage. In fact, I crashed my own machine :_D

Well, that is expected, as you end up with massive padding (tons of points to loop through).
I don't have any good way of counteracting this behavior, and have therefore added warnings/notes that this value should be set on a problem basis, and should likely not be very large.

massimiliano-leoni:
Yes, I guess that's reasonable :)

cpp/dolfinx/geometry/utils.h
jorgensd: I've added comments to try to address this

jorgensd:
> I think use of the work 'extrapolation' can be tightened up, which will help understanding. 'Extrapolation' is used in the context of geometric operations, which doesn't seem right.

I've removed the usage of extrapolate in the documentation. (Except to explain that it could be used for this).

PR #2862, opened 2023-11-03
Improve collision detection

jorgensd:
Building on top of the fixes in #2858 I've made a more scalable version of the algorithm.

Full experiment available at: https://gist.github.com/jorgensd/55a2c06d512998687837cbaa00fec1c1

**Short story**
Instead of using the `compute_closest_entity_command`, that uses a midpoint tree to find the distances, instead use the cell candidates from the initial collision detection. This means that we are re-computing less collisions.

**Example**
Code available in the gist above. Idea is: 
Take to unit cubes of different resolution, one slightly shifted and interpolate data from one to the other. 
Try these with various parameters for extrapolation (0, 1e-6, 2e-3, 1e-2).
- If we chose 0 as extrapolation, we get accuracy issues due to the fact that GJK uses the squared norm as a tolerance.
- 1e-6 will avoid these issues, and adds minimal overhead to the selection of candidate cells, no extrapolation outside of mesh.
- 2e-3 is in this case enough to also extrapolate data to the shifted boundary
- 1e-2 is a very large parameter, including a vast set of cells, adding significant overhead.

# Compare with v0.6.0
Using tol=0, we compare with v0.6.0-r1 (24f86a9ce57df6978070dbee22b3eae8bb77235f)
```csv
size,max_data_time
1,6.533921017000466
2,3.99001837600008
4,2.3080832739997277
```
First column is number of processes, second is interpolation time (no separate creation of nms data)

# The other branch (that this is merging into) 
hash: b897f21eb67ac793d7149b43eb152e05dd814a6b

```csv
size,max_data_time,max_int_time, tol
1,6.055579162000868,0.9742533139997249,0
2,3.369704722001188,0.5460317219985882,0 
4,2.0919247600013477,0.2986966759999632,0
1,30.938958914999603,0.9725744930001383,0.01
2,22.509738571001435,0.56720664400018,0.01
4,17.364121013999466,0.32394687600026373,0.01
1,12.113241491000736,0.9981611050006904,0.002
2,8.768939632000183,0.569543411000268,0.002
4,6.953212500999143,0.2708331220001128,0.002
1,6.357045004000611,0.9862442329995247,1e-06
2,3.792262732000381,0.5485727609993774,1e-06
4,2.2686188660009066,0.2465637750010501,1e-06
```
We observe that with a small padding (tol) we have quite good and comparable results with the previous release.
However, with increasing tolerance/padding, we observe an explosion in runtime, which is not scalable when paralellizing.

# This PR
hash: 81d357874b207d201b204276b25176aab5f747a6
```csv
size,max_data_time,max_int_time, tol
1,5.426864094000848,0.9956041120003647,0,
2,3.037522261000049,0.5439577070010273,0
4,1.9478289220005536,0.24801453500003845,0
1,25.432961013999375,1.0255009730008169,0.01
2,13.682743048999328,0.561956891000591,0.01
4,8.967257972999505,0.26832217399896763,0.01
1,7.138527663000787,1.0103946380004345,0.002
2,4.0080116690005525,0.5580355759993836,0.002
4,2.531900571000733,0.2624842469995201,0.002
1,5.336107089999132,0.9821945519997826,1e-06
2,3.011473816999569,0.5505700230005459,1e-06
4,2.1531176140015305,0.25772153999969305,1e-06
```
We observe that we not get any penalty for a small tolerance/padding, and that for large padding the scaling is much better


jorgensd:
@massimiliano-leoni  I would like you to review this too;)

garth-wells (approved):

PR #2867, opened 2023-11-05
Re-enable basix element test

francesco-ballarin:
Was disabled when `basix` was using `nanobind` and `dolfinx` was using `pybind11` instead. It may have not been working when the `nanobind` PR was merged because the version of `nanobind` used by `basix` was slightly different then the one installed during CI for `dolfinx`, but that is sorted now that both `basix` and `dolfinx` use the same `nanobind` from the docker image.


garth-wells (approved):
Happy to re-enable, but it would be best if DOLFINx does not pretend to know anything about how the Basix Python interface is implemented.

PR #2870, opened 2023-11-06
Add test of quadrature element

mscroggs:
Add code from https://github.com/FEniCS/ffcx/pull/622 as a test


mscroggs:
Would be good to add more tests of quadrature elements to stop this from breaking in future. I've opened #2872, so we can post ideas for test there

garth-wells (approved):

jorgensd (approved):

PR #2876, opened 2023-11-07
Drop installation of `dolfinx.conf` file

francesco-ballarin:
Do we still need to install `dolfinx.conf`? In my opinion, the answer is no, because:
1. most of the current packaging platforms (docker, ubuntu apt packages, spack, conda) do not actually use it
2. for docker, the files `docker/dolfinx-real-mode` and `docker/dolfinx-complex-mode` already serve the same purpose
3. the vast majority of end users will use the python backend, and `PYTHONPATH` does not even get set in `dolfinx.conf`
4. a C++ end user is probably smart enough to know that they should set those paths. If they are not, this file would only partially help, since it does not necessarily set the paths for `basix` (and there isn't an equivalent file in basix that does that).


docker/Dockerfile.end-user
```diff
@@ -89,17 +89,19 @@ ONBUILD RUN cd dolfinx && \
...
+    . /usr/local/bin/dolfinx-real-mode && \
     cd ../python && \
     PETSC_ARCH=linux-gnu-real64-32 pip3 install -v --config-settings=cmake.define.-DCMAKE_CXX_FLAGS="${DOLFINX_CMAKE_CXX_FLAGS}" \
-      --config-settings=cmake.build-type="${DOLFINX_CMAKE_BUILD_TYPE}"  --no-build-isolation --check-build-dependencies \
+      --config-settings=cmake.build-type="${DOLFINX_CMAKE_BUILD_TYPE}" --no-build-isolation --check-build-dependencies \
       --target /usr/local/dolfinx-real/lib/python${PYTHON_VERSION}/dist-packages --no-dependencies --no-cache-dir . && \
     git clean -fdx && \
-    cd ../ && \
+
+ONBUILD RUN cd dolfinx && \
```
francesco-ballarin, line 99: Tested in https://github.com/FEniCS/dolfinx/actions/runs/6780380672.

I need to split into two `ONBUILD RUN` otherwise the `real` and `complex` environment variables get mixed up.

jhale:
A key design goal is keeping a usable C++ layer, no Python. Pkgconfig is a boring but important part of that.

1. Indeed, and that's ok. Different tools, different ways of doing things.
2. Keep in mind that those scripts also configure the Python environment which pkgconfig doesn't support because it's out of scope.
3. That's correct. dolfinx.conf is related only to the C++ library.
4. We should add pkgconfig to basix if it doesn't exist already.

garth-wells:
I use the `dolfinx.conf` file all the time because I never install into a system path.

francesco-ballarin:
OK! Feel free to close if this is not what we are looking for, I opened this to assess the feedback from you guys.

PR #2877, opened 2023-11-07
Change return argument of `IndexMap::create_submap` to use `std::shared_ptr`

francesco-ballarin:


francesco-ballarin:
I guess what confuses me is that:
1. `IndexMap` is not copyable or movable, so I don't quite understand how it can be moved into a `shared_ptr`,
2. every occurence in the library which does indeed move the object into a `shared_ptr`, does so in the same scope as the original object. I need to store the `shared_ptr` out of the scope of the original object, wouldn't the original object be destroyed as soon as it goes out of scope?

garth-wells:
> I guess what confuses me is that:
> 
> 1. `IndexMap` is not copyable or movable, so I don't quite understand how it can be moved into a `shared_ptr`,

Looks to me like it support move semantics, see https://github.com/FEniCS/dolfinx/blob/b36b23112e97ca5d2728c9a76efc83f1b531679c/cpp/dolfinx/common/IndexMap.h#L120 and https://github.com/FEniCS/dolfinx/blob/b36b23112e97ca5d2728c9a76efc83f1b531679c/cpp/dolfinx/common/IndexMap.h#L126.

Something like 
```cpp
auto [_map, pos] = foo.create_submap(....);
auto map = std::maked_shared<IndexMap>(std::move(_map)); 
```
should work.

francesco-ballarin:
Well, I must have misread `default` and `delete` :). Let me check about item 2, if `std::move` makes it survive out of the scope the object was originally defined, then I have all I need to use it in `multiphenicsx`, and I will close this.

garth-wells:
> Well, I must have misread `default` and `delete` :). Let me check about item 2, if `std::move` makes it survive out of the scope the object was originally defined, then I have all I need to use it in `multiphenicsx`, and I will close this.

They should be changed to `default`(!).

Issue #2872, opened 2023-11-06
Add tests of quadrature elements


mscroggs:
### Describe new/missing feature

Quadrature elements are not thoroughly tested. It would be good to add more tests. Open to suggestions of test that could be added.

A first test is added in #2870.

bleyerj:
Hi, on dolfinx:nightly, the following test works for scalar elements `shape=()` but fails for vector elements `shape=(1,)` or `shape=(2,)`
```python
def test_vector_element(shape):
    msh = mesh.create_unit_square(MPI.COMM_WORLD, 10, 10)

    dx_m = ufl.Measure(
        "dx",
        domain=msh,
        metadata={"quadrature_degree": 1, "quadrature_scheme": "default"},
    )

    Qe = quadrature_element(
        msh.topology.cell_name(), value_shape=shape, scheme="default", degree=1
    )
    Quad = fem.functionspace(msh, Qe)
    q_ = ufl.TestFunction(Quad)
    dq = ufl.TrialFunction(Quad)
    one = fem.Function(Quad)
    one.vector.set(1.0)
    mass_L_form = fem.form(ufl.inner(one, q_) * dx_m)
    mass_v = fem.assemble_vector(mass_L_form)
    mass_a_form = fem.form(ufl.inner(dq, q_) * dx_m)
    mass_A = fem.assemble_matrix(mass_a_form)
    assert np.isclose(sum(mass_v.array), 1.0)
    assert np.allclose(np.diag(mass_A.to_dense()), mass_v.array)
```
The error is:
```
    mass_L_form = fem.form(ufl.inner(one, q_) * dx_m)
  File "/usr/local/dolfinx-real/lib/python3.10/dist-packages/dolfinx/fem/forms.py", line 189, in form
    return _create_form(form)
  File "/usr/local/dolfinx-real/lib/python3.10/dist-packages/dolfinx/fem/forms.py", line 184, in _create_form
    return _form(form)
  File "/usr/local/dolfinx-real/lib/python3.10/dist-packages/dolfinx/fem/forms.py", line 141, in _form
    ufcx_form, module, code = jit.ffcx_jit(mesh.comm, form,
  File "/usr/local/dolfinx-real/lib/python3.10/dist-packages/dolfinx/jit.py", line 56, in mpi_jit
    return local_jit(*args, **kwargs)
  File "/usr/local/dolfinx-real/lib/python3.10/dist-packages/dolfinx/jit.py", line 204, in ffcx_jit
[...]

PR #2878, opened 2023-11-07
Implement interpolation for quadrature elements

### LINKED ISSUE #1546: Interpolation into Quadrature (non-basix) element
Elements which are not registered in basix, but still interpolatory, could prepare interpolation points and interpolation matrix during element construction in `FiniteElement.cpp`.
For Quadrature element this requires quadrature scheme provided from FFCx generated code, or better, a "variant".

mscroggs:
Requires https://github.com/FEniCS/ffcx/pull/634.

Resolves #1546.

Progress towards #2872


chrisrichardson:
I am a bit confused by the complexity of it - it seems to be adding more generated code (which we generally try to avoid) to pass data from basix via ffcx back to dolfinx. Is this avoidable?

mscroggs:
> I am a bit confused by the complexity of it - it seems to be adding more generated code (which we generally try to avoid) to pass data from basix via ffcx back to dolfinx. Is this avoidable?

Writing the quadrature rules into the generated code is definitely necessary.

We could avoid writing the map type into generated code if we make the assumption that elements with custom quadrature always use an identity pullback. This feels dangerous though.

The key issue is that quadrature elements are only defined in Python, so we need to write what we need into the generated code so that C++ can get hold of it

jorgensd:
> We could avoid writing the map type into generated code if we make the assumption that elements with custom quadrature always use an identity pullback. This feels dangerous though.

I'm not sure it is dangerous. QuadratureElements isn't really a finite element (as discussed earlier), and is a way of representing data at points. Having a pull back != identity on point data seems strange.

mscroggs:
> > We could avoid writing the map type into generated code if we make the assumption that elements with custom quadrature always use an identity pullback. This feels dangerous though.
> 
> I'm not sure it is dangerous. QuadratureElements isn't really a finite element (as discussed earlier), and is a way of representing data at points. Having a pull back != identity on point data seems strange.

The danger I'm worried about it that we're not checking if it's a quadrature element, we're checking if there's a custom quadrature rule. If at some point in the future we decided to implement (eg) a Nedelec-style element that has a custom quadrature rule (for whatever reason), this would lead to odd behaviour.

mscroggs:
> > > We could avoid writing the map type into generated code if we make the assumption that elements with custom quadrature always use an identity pullback. This feels dangerous though.
> > 
> > 
> > I'm not sure it is dangerous. QuadratureElements isn't really a finite element (as discussed earlier), and is a way of representing data at points. Having a pull back != identity on point data seems strange.
> 
> The danger I'm worried about it that we're not checking if it's a quadrature element, we're checking if there's a custom quadrature rule. If at some point in the future we decided to implement (eg) a Nedelec-style element that has a custom quadrature rule (for whatever reason), this would lead to odd behaviour.

I've changed it to check whether the element is a quadrature element directly, so no longer any need to worry about possible other future uses of custom quadrature

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -690,5 +690,9 @@ class FiniteElement
 
   // Basix Element (nullptr for mixed elements)
   std::unique_ptr<basix::FiniteElement<T>> _element;
+
+  // Quadrature points of a quadrature element (nullptr for non-quadrature
```
garth-wells, line 694: Not a pointer, so can't be `nullptr`.

jorgensd:
Could we merge this?

PR #2880, opened 2023-11-07
Add test of symmetric tensor elements

mscroggs:
Add a test to check that symmetric elements are correct.

Fixes https://github.com/FEniCS/dolfinx/issues/2750

### DIFF
--- a/python/test/unit/fem/test_symmetric_tensor.py
+++ b/python/test/unit/fem/test_symmetric_tensor.py
@@ -1,0 +1,21 @@
+from mpi4py import MPI
+import basix.ufl
+import ufl
+import dolfinx
+import numpy as np
+import pytest
+
+
+@pytest.mark.parametrize("degree", range(1, 4))
+def test_transpose(degree):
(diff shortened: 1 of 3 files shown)


python/test/unit/fem/test_symmetric_tensor.py
```diff
@@ -0,0 +1,23 @@
...
+
+import basix.ufl
+import dolfinx
+import ufl
+
+
+@pytest.mark.parametrize("degree", range(1, 4))
+def test_transpose(degree):
+    mesh = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, 10, 10)
+    for symmetry in [True, False]:
```
francesco-ballarin, line 14: You may want to move this to `pytest.mark.parametrize`, so that if only one of the two cases fails in future it will be more clear which one of the two cases is failing.

mscroggs: Good idea

garth-wells (approved):

PR #2885, opened 2023-11-10
Make topology const in `locate_dofs_topological`

IgorBaratta:
As for other functions in `dolfinx`, the user should call `create_connectivity` explicitly.
Locating dofs shouldn't modify the topology.


jorgensd (commented):
I've got mixed emotions regarding this. If one wants to call `locate_dofs_topological`, this is clearly needed, and it is not clear why we want to force users to call it for this function.

Also note that if we remove it, we need proper error handling (a simple assert isn't enough), as this is not checked in release mode. Then we would need a runtime error saying "Missing connectivity between entities of dim {dim} and {tdim}".

IgorBaratta:
I think you're right about the error handling.
But other functions already require the user to call `create_connectivity` and `create_entity`, for example `exterior_facet_indices`.

jorgensd:
> I think you're right about the error handling. But other functions already require the user to call `create_connectivity` and `create_entities`, for example `exterior_facet_indices`.

I know, and if we want to be consistent, we follow your code (and add an appropriate error message). 
I just feel like we should add to the doc-strings what connectives we  expect to ensure that it is easily available information, and that people don't have to run the code and get an error message to know what they need.

francesco-ballarin:
I for sure agree with all comments above, but I just want to point that these commits did _not_ change a single test or demo. It seems to me that, for all intents and purposes, in any of our actual use cases (C++ demos, C++ tests, python demos, python tests) those entities and connectives are indeed already computed in practice, so most of the users won't even notice this change.

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -32,13 +32,11 @@ namespace
...
-find_local_entity_index(mesh::Topology& topology,
+find_local_entity_index(const mesh::Topology& topology,
                         std::span<const std::int32_t> entities, int dim)
 {
   // Initialise entity-cell connectivity
   const int tdim = topology.dim();
-  topology.create_entities(tdim);
-  topology.create_connectivity(dim, tdim);
   auto e_to_c = topology.connectivity(dim, tdim);
   assert(e_to_c);
```
garth-wells, line 41: Change from assertion to runtime error (and below too).

garth-wells (changes requested):
This is good, but need to change assertions to exceptions.

On Python side we're ok with more implicit behaviour so any required initialisation can happen in the Python layer.

jorgensd (approved):

garth-wells (approved):

PR #2887, opened 2023-11-13
Export `CMAKE_PREFIX_PATH` in helper installation scripts

francesco-ballarin:
Mimicks the already existing export of `PKG_CONFIG_PATH` and helps downstream development in the nightly docker image.

This change is not required when building the docker nightly image itself, but I honestly do not understand why:
* before this change, https://github.com/FEniCS/dolfinx/blob/main/docker/Dockerfile.end-user#L102 did not set either `CMAKE_PREFIX_PATH` or `DOLFINX_DIR`, yet (as manually verified with a few runs) the complex installation gets correctly detected.
* https://github.com/FEniCS/dolfinx/blob/main/docker/Dockerfile.end-user#L87 does not even source `/usr/local/dolfinx-real/lib/dolfinx/dolfinx.conf`, yet the real installation gets correctly found, even if it is in a non standard path.

I also have verified that without `CMAKE_PREFIX_PATH` (and without `DOLFINX_DIR`) one could have had mixed configurations when using the nightly image, in which after sourcing `/usr/local/dolfinx-complex/lib/dolfinx/dolfinx.conf` the python part is imported from `/usr/local/dolfinx-complex`, while the C++ one from `/usr/local/dolfinx-real`.


jhale (approved):

jhale:
It isn't strictly necessary to be up to date with main to merge. The merge queue will handle merging main into your branch and running the test suite.

francesco-ballarin:
I'll wait to see if @garth-wells has any comment, and after that I'll merge to main.

python/CMakeLists.txt
```diff
@@ -29,15 +29,15 @@ execute_process(
   ERROR_VARIABLE BASIX_ERROR_OUT OUTPUT_STRIP_TRAILING_WHITESPACE
 )
 find_package(Basix REQUIRED CONFIG HINTS ${BASIX_PY_DIR})
+message(STATUS "Found Basix at ${BASIX_DIR}")
```
garth-wells, line 32: This needs an if/else, e.g.
```
if (Basix_FOUND)
  message(STATUS "Found Basix at ${BASIX_DIR}")
endif()
```

python/CMakeLists.txt
```diff
@@ -29,15 +29,15 @@ execute_process(
...
 find_package(DOLFINX REQUIRED CONFIG)
+message(STATUS "Found DOLFINx at ${DOLFINX_DIR}")
```
garth-wells, line 34: See above comment on if/else.

garth-wells (changes requested):

python/CMakeLists.txt
francesco-ballarin: Why is that? My understanding is that `find_package(... REQUIRED ...)` will stop with an error if the package is not found, and therefore the line with message will only get printed if the package was actually found.

garth-wells: I don't think CMake necessarily stops immediately. It's also a readability issue.

francesco-ballarin: OK, will do!

garth-wells (approved):

PR #2889, opened 2023-11-13
Enable partition of a `box mesh` on subset of processes

IgorBaratta:


garth-wells (commented):
Looks like this only works 'fully' for ParMETIS? Other partitioners partition using all ranks, even if graph is empty on some ranks.

IgorBaratta:
> Looks like this only works 'fully' for ParMETIS? Other partitioners partition using all ranks, even if graph is empty on some ranks.

I'll have a look at it.

jorgensd:
> Looks like this only works 'fully' for ParMETIS? Other partitioners partition using all ranks, even if graph is empty on some ranks.

Wasn't this one of the things that was working with KaHip way back when @IgorBaratta worked on it?

garth-wells:
> Before it was more explicit. Now, if the local graph doesn't contain any nodes, the process is not used for partition.

It's a work-around for ParMETIS, which fails if a rank has no part of the graph.

PR #2890, opened 2023-11-14
Fix `create_submesh` by generalising sub index map creation

jpdean:
This PR fixes a problem where sub-meshes can have vertices that are not connected to a cell on the owning process in parallel.

To illustrate, consider the following mesh, which is partitioned across two processes. Blue shading indicates owned entities. Red shading indicates ghost entities. Gray shading indicates entities that are not present on that process. The vertex numbering is local to each process.

mesh

Creating an L-shaped sub-mesh yields the following on main:

main

Note that there is a lone vertex that is not connected to a cell on the process that owns it. This is problematic because we usually perform operations cellwise, and having vertices that are not connected to a cell can yield unexpected results. For example, if a function were to be interpolated on the sub-mesh, any degrees-of-freedom associated with the lone vertex would not be assigned a value on process 0. On process 1, a value would be assigned, but this would be overwritten with a forwards scatter.

 This PR changes the submap owners of problematic vertices to avoid this issue, yielding:

pr

The problematic vertex is now owned by process 1 in the submap and has been removed from process 0.


cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,353 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
 
   return {std::move(src), std::move(dest)};
 }
+
+// A helper function that, given an index map and a set of local indices (can be
+// owned or ghost but must be unique and sorted), returns the owned, ghost and
+// ghost owners in the submap. All indices are local and with respect to the
+// original index map.
```
jorgensd, line 39:
```suggestion
/// Given a subset of indices in an index map, compute owned indices, ghost indices and owners of ghosts.
///
/// A helper function that, given an index map and a set of local indices (can be
/// owned or ghost but must be unique and sorted), returns the owned, ghost and
/// ghost owners in the submap. All indices are local and with respect to the
/// original index map.
/// @param[in] imap Index map.
/// @param[in] indices List of entity indices (indices local to the process).
/// @pre `indices` must be sorted and unique. 
```

jpdean:
I'm not really sure where the best place (or what the best name) is for the new function for creating sub index maps. I've put it in the index map class for the moment, but since it is really only needed for sub-mesh creation, maybe it's better in mesh? Any thoughts would be much appreciated.

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,353 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+  std::partial_sum(send_sizes.begin(), send_sizes.end(),
+                   std::next(send_disp.begin()));
+  std::partial_sum(recv_sizes.begin(), recv_sizes.end(),
+                   std::next(recv_disp.begin()));
+
+  // Send ghost indices to owner, and receive indices owned by this process
+  // that are ghosts on others
+  std::vector<std::int64_t> send_indices, recv_indices;
+  for (auto& d : send_data)
+    send_indices.insert(send_indices.end(), d.begin(), d.end());
```
jorgensd, line 112: You could make this communicator non-blocking, and pack the data while it is being sent. 
Then compute the `recv_disp` and `recv_indices` once packing of outgoing data is finished, see for instance: https://github.com/FEniCS/dolfinx/blob/6d0720e3dcdacdcd045afbb376c367a70c27fe42/cpp/dolfinx/geometry/utils.h#L733-L763

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,353 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+    if (idx != -1)
+    {
+      // TODO Choose new owner randomly for load balancing
+      const std::vector<std::int32_t>& possible_owners
+          = global_idx_to_possible_owner[idx];
+      // const int random_index = std::rand() % possible_owners.size();
+      // send_owners.push_back(possible_owners[random_index]);
+
+      // FIXME TEMPORARILY CHOOSE THE FIRST PROCESS, MAKE RANDOM
+      send_owners.push_back(possible_owners[0]);
```
jorgensd, line 179: I wouldn't make it completely random, as results would be nondeterministic. However, if you sort the owners for each index when you create `global_idx_to_possible_owner` we can get a consistent result

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,353 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+    }
+    else
+      send_owners.push_back(-1);
+  }
+
+  // Create neighbourhood comm (owner -> ghost)
+  MPI_Comm comm1;
```
jorgensd, line 186: A more explicit name would be good. What about `forward_comm` and `reverse_comm`?

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,353 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+    }
+  }
+  std::sort(submap_owned.begin(), submap_owned.end());
+
+  // TODO Move
+  return {submap_owned, submap_ghost, submap_ghost_owners};
+}
+
+// Helper function to compute the submap indices of ghosts.
+// NOTE: submap_owned must be sorted and contain no repeated indices
```
jorgensd, line 252: Same note as above, fix documentation style, and clarify what all inputs are.

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,353 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+{
+  // --- Step 1 ---: Send global ghost indices (w.r.t. original imap) to owning
+  // rank
+  std::vector<std::int64_t> recv_indices;
+  std::vector<std::size_t> ghost_perm;
+  std::vector<int> send_disp, recv_disp;
+  std::vector<std::int32_t> send_sizes, recv_sizes;
+  {
+    // Create neighbourhood comm (ghost -> owner)
+    MPI_Comm comm0;
```
jorgensd, line 270: Same comment regarding comm names

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,356 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+/// @param[in] indices List of entity indices (indices local to the process).
+/// @pre `indices` must be sorted and unique.
+/// @return The owned, ghost, and ghost owners in the submap. All indices are
+/// local and with respect to the original index map.
+std::tuple<std::vector<std::int32_t>, std::vector<std::int32_t>,
+           std::vector<std::int32_t>>
+compute_submap_indices(const dolfinx::common::IndexMap& imap,
+                       std::span<const std::int32_t> indices)
+{
+  // --- Step 1 ---: Send ghost indices in `indices` to their owners
```
garth-wells, line 49: For long functions with 'steps' it's useful to scope each step (by using a `{`, `}` pair).

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,357 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+    is_in_submap[v] = 1;
+
+  // Pack ghost indices that this process wants to include in the
+  // sub-map
+  std::vector<std::vector<std::int64_t>> send_data(src.size());
+  for (std::size_t i = 0; i < ghosts.size(); ++i)
+  {
+    auto it = std::lower_bound(src.begin(), src.end(), ghost_owners[i]);
+    assert(it != src.end() and *it == ghost_owners[i]);
+    int r = std::distance(src.begin(), it);
```
garth-wells, line 78:
```suggestion
    std::size_t r = std::distance(src.begin(), it);
```

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,357 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+                 std::back_inserter(send_sizes),
+                 [](auto& d) { return d.size(); });
+
+  // Send how many indices I ghost to each owner, and receive how many
+  // of my indices other ranks ghost
+  std::vector<std::int32_t> recv_sizes;
+  recv_sizes.resize(dest.size(), 0);
+  send_sizes.reserve(1);
+  recv_sizes.reserve(1);
+  ierr = MPI_Neighbor_alltoall(send_sizes.data(), 1, MPI_INT32_T,
```
garth-wells, line 98: This communication pattern is used lot - maybe put it in a function?

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,357 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+
+  // --- Step 2 ---: Create a map from the indices in `recv_indices` (i.e.
+  // indices owned by this process that are in `indices` on other processes) to
+  // their owner in the submap. This is required since not all indices in
+  // `recv_indices` will necessarily be in `indices` on this process, and thus
+  // other processes must own them in the submap.
+
+  // Create a map from (global) owned shared indices owned to processes that can
+  // own them.
+  // FIXME Avoid map
```
garth-wells, line 136: Resolve?

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,357 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+  std::map<std::int64_t, std::vector<std::int32_t>>
+      global_idx_to_possible_owner;
+  const std::array local_range = imap.local_range();
+  for (std::size_t i = 0; i < recv_disp.size() - 1; ++i)
+  {
+    for (int j = recv_disp[i]; j < recv_disp[i + 1]; ++j)
+    {
+      std::int64_t idx = recv_indices[j];
+      // Check that the index is in the submap
+      if (idx != -1)
```
garth-wells, line 149:
```suggestion
      if (std::int64_t idx = recv_indices[j]; idx != -1)
```

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,357 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+  // A map from (global) indices in `recv_indices` to a list of processes that
+  // can own the index in the submap.
+  std::map<std::int64_t, std::vector<std::int32_t>>
+      global_idx_to_possible_owner;
+  const std::array local_range = imap.local_range();
+  for (std::size_t i = 0; i < recv_disp.size() - 1; ++i)
+  {
+    for (int j = recv_disp[i]; j < recv_disp[i + 1]; ++j)
+    {
+      std::int64_t idx = recv_indices[j];
```
garth-wells, line 147:
```suggestion
```

cpp/dolfinx/common/IndexMap.cpp
garth-wells: I think it's ok - what will help most is scoping.

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,357 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+                                send_disp.data(), MPI_INT32_T, comm1);
+  dolfinx::MPI::check_error(comm, ierr);
+
+  // Free the communicator
+  ierr = MPI_Comm_free(&comm1);
+  dolfinx::MPI::check_error(comm, ierr);
+
+  // Local indices (w.r.t. original map) owned by this process in the submap
+  std::vector<std::int32_t> submap_owned;
+  // Local indices (w.r.t. original map) ghosted by this process in the submap
```
garth-wells, line 209: Having comments sandwiched without a line break isn't great for readability.

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,357 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+  for (std::size_t i = 0; i < send_indices.size(); ++i)
+  {
+    std::int32_t local_idx = send_indices_local[i];
+    std::int64_t global_idx = send_indices[i];
+    std::int32_t owner = recv_owners[i];
+
+    // Check if index is in the submap
+    if (global_idx >= 0)
+    {
+      if (owner == rank)
```
garth-wells, line 237:
```suggestion
      std::int32_t local_idx = send_indices_local[i];
      if (owner == rank)
```

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,357 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+  std::vector<std::int32_t> send_indices_local(send_indices.size());
+  imap.global_to_local(send_indices, send_indices_local);
+
+  // Loop over ghost indices (in the original map) and add to submap_owned
+  // if the owning process has decided this process to be the submap owner.
+  // Else, add the index and its (possibly new) owner to submap_ghost and
+  // submap_ghost_owners respectively.
+  for (std::size_t i = 0; i < send_indices.size(); ++i)
+  {
+    std::int32_t local_idx = send_indices_local[i];
```
garth-wells, line 230:
```suggestion
```

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,357 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+    std::int64_t global_idx = send_indices[i];
+    std::int32_t owner = recv_owners[i];
+
+    // Check if index is in the submap
+    if (global_idx >= 0)
```
garth-wells, line 235:
```suggestion
    if (std::int64_t global_idx = send_indices[i]; global_idx >= 0)
```

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,357 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+  imap.global_to_local(send_indices, send_indices_local);
+
+  // Loop over ghost indices (in the original map) and add to submap_owned
+  // if the owning process has decided this process to be the submap owner.
+  // Else, add the index and its (possibly new) owner to submap_ghost and
+  // submap_ghost_owners respectively.
+  for (std::size_t i = 0; i < send_indices.size(); ++i)
+  {
+    std::int32_t local_idx = send_indices_local[i];
+    std::int64_t global_idx = send_indices[i];
```
garth-wells, line 231:
```suggestion
```

cpp/dolfinx/common/IndexMap.cpp
garth-wells, line 237:
```suggestion
      if (std::int32_t owner = recv_owners[i]; owner == rank)
```

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,357 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+    std::int32_t owner = recv_owners[i];
```
garth-wells, line 232:
```suggestion
```

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,357 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+  std::sort(submap_owned.begin(), submap_owned.end());
+
+  return {std::move(submap_owned), std::move(submap_ghost),
+          std::move(submap_ghost_owners)};
+}
+
+/// Computes the global indices of ghosts in a submap.
+// NOTE: submap_owned must be sorted and contain no repeated indices
+std::vector<std::int64_t> compute_submap_ghost_indices(
+    const std::vector<int>& submap_src,
```
garth-wells, line 257: Use spans.

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,357 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+    const std::vector<int>& submap_dest,
+    const std::vector<std::int32_t>& submap_owned,
+    const std::vector<std::int64_t>& submap_ghosts_global,
+    const std::vector<std::int32_t>& submap_ghost_owners,
+    const int submap_offset,
+    const dolfinx::common::IndexMap& imap) // NOTE: This is the original imap!
+{
+  // --- Step 1 ---: Send global ghost indices (w.r.t. original imap) to owning
```
garth-wells, line 265: Scope the steps.

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -636,6 +987,58 @@ IndexMap::create_submap(std::span<const std::int32_t> indices) const
...
+                        MPI_SUM, _comm.comm());
+  dolfinx::MPI::check_error(_comm.comm(), ierr);
+
+  // Get submap source ranks
+  std::vector<int> submap_src(submap_ghost_owners.begin(),
+                              submap_ghost_owners.end());
+  std::sort(submap_src.begin(), submap_src.end());
+  submap_src.erase(std::unique(submap_src.begin(), submap_src.end()),
+                   submap_src.end());
+  submap_src.shrink_to_fit();
```
garth-wells, line 1012: Why is `shrink_to_fit` being called here?

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -636,6 +987,58 @@ IndexMap::create_submap(std::span<const std::int32_t> indices) const
...
+
+  // Compute submap destination ranks
+  // NOTE: The call to NBX can be avoided by a two-step communication process
```
garth-wells, line 1015: Will you implement this? @IgorBaratta has done some work to remove unnecessary NBX calls, so we don't want to add new unnecessary calls.

cpp/dolfinx/common/IndexMap.h
```diff
@@ -189,6 +189,21 @@ class IndexMap
   std::pair<IndexMap, std::vector<std::int32_t>>
   create_submap(std::span<const std::int32_t> indices) const;
 
+  /// @brief Create a new index map from a subset of indices in this index
```
garth-wells, line 192: This is hard to read. Maybe split into two sentences?

cpp/dolfinx/common/IndexMap.h
```diff
@@ -189,6 +189,21 @@ class IndexMap
...
+  /// This can be used when, for instance, creating a submesh to ensure
+  /// that all vertices have exactly one owner and are connected to at least
+  /// one cell on the owning process.
+  /// @param[in] indices Local indices to include in the new index map (owned
+  /// and ghost)
+  /// @pre `indices` must be sorted and contain no duplicates.
+  /// @return The (i) new index map and (ii) a map from local indices in the
+  /// submap to local indices in the original (this) map
+  std::pair<IndexMap, std::vector<std::int32_t>>
+  create_submap_conn(std::span<const std::int32_t> indices) const;
```
garth-wells, line 205: Improve the function name?

cpp/dolfinx/common/IndexMap.h
```diff
@@ -189,6 +189,21 @@ class IndexMap
...
+  ///
+  /// This can be used when, for instance, creating a submesh to ensure
+  /// that all vertices have exactly one owner and are connected to at least
+  /// one cell on the owning process.
+  /// @param[in] indices Local indices to include in the new index map (owned
+  /// and ghost)
+  /// @pre `indices` must be sorted and contain no duplicates.
+  /// @return The (i) new index map and (ii) a map from local indices in the
+  /// submap to local indices in the original (this) map
+  std::pair<IndexMap, std::vector<std::int32_t>>
```
garth-wells, line 204: Could this be a free function rather than a method? Free functions are generally preferred.

cpp/dolfinx/mesh/Geometry.h
```diff
@@ -328,22 +328,15 @@ create_subgeometry(const Topology& topology, const Geometry<T>& geometry,
   // Get the sub-geometry dofs owned by this process
   auto x_index_map = geometry.index_map();
   assert(x_index_map);
-  std::vector<std::int32_t> subx_to_x_dofmap
-      = common::compute_owned_indices(sub_x_dofs, *x_index_map);
+
   std::shared_ptr<common::IndexMap> sub_x_dof_index_map;
+  std::vector<std::int32_t> subx_to_x_dofmap;
   {
     std::pair<common::IndexMap, std::vector<int32_t>> map_data
```
garth-wells, line 335: Could `std::tie` make this code easier to read and avoid the `std::move` at the end of the scope?

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -1195,26 +1195,17 @@ mesh::create_subtopology(const Topology& topology, int dim,
...
-  std::sort(indices.begin(), indices.end());
-  std::vector<std::int32_t> subvertices0
-      = common::compute_owned_indices(indices, *map0);
 
   // Create map from the vertices in the sub-topology to the vertices in the
   // parent topology, and an index map
   std::shared_ptr<common::IndexMap> submap0;
+  std::vector<int32_t> subvertices0;
   {
     std::pair<common::IndexMap, std::vector<int32_t>> map_data
```
garth-wells, line 1204: See above comment.

garth-wells (changes requested):

cpp/dolfinx/common/IndexMap.cpp
jpdean: `submap_src` initially contains repeated indices which are then removed. It's capacity will then be larger than it's size, so I call `shrink_to_fit`. I can remove it if you prefer.

cpp/dolfinx/mesh/Geometry.h
jpdean: I don't think so, unless I'm missing something

francesco-ballarin (approved):

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,393 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+compute_submap_indices(const dolfinx::common::IndexMap& imap,
+                       std::span<const std::int32_t> indices,
+                       bool allow_owner_change)
+{
+  const MPI_Comm comm = imap.comm();
+  std::span<const std::int32_t> src = imap.src();
+  std::span<const std::int32_t> dest = imap.dest();
+
+  // Lookup array to determine if an index is in the sub-map
+  std::vector<int> is_in_submap(imap.size_local() + imap.num_ghosts(), 0);
```
garth-wells, line 174: Could this be a `std::uint8_t` rather than `int`?

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -32,6 +32,393 @@ std::array<std::vector<int>, 2> build_src_dest(MPI_Comm comm,
...
+    submap_dest.erase(std::unique(submap_dest.begin(), submap_dest.end()),
+                      submap_dest.end());
+    submap_dest.shrink_to_fit();
+
+    // Choose the submap owner for each index in `recv_indices`
+    std::vector<std::int32_t> send_owners;
+    send_owners.reserve(recv_indices.size());
+    for (auto idx : recv_indices)
+    {
+      // Check the index is in the submap, otherwise send -1
```
garth-wells, line 249: Comment doesn't quite make sense - need to be 'if X do Y, else do Z'.

garth-wells (approved):

Issue #2891, opened 2023-11-16
Regression in Python `assemble_vector` performance after nanobind migration


jorgensd:
### How to reproduce the bug

Using a quite simple linear form with a mixed element

### Minimal Example (Python)

```Python
from dolfinx import fem, mesh as dmesh
import basix.ufl
from mpi4py import MPI
import ufl
import time

degree = 1
mesh = dmesh.create_unit_square(MPI.COMM_WORLD, 1000, 1000)

cell = mesh.ufl_cell()
FE = basix.ufl.element("Lagrange", str(cell), degree)
ME = basix.ufl.mixed_element([FE, FE])
VQ = fem.functionspace(mesh, ME)

# Define test, trial and functions for previous timestep
Az, V = ufl.TrialFunctions(VQ)
vz, q = ufl.TestFunctions(VQ)
dx = ufl.Measure("dx", domain=mesh)
L = vz * dx
cpp_L = fem.form(L)
b = fem.create_vector(cpp_L)

start = time.perf_counter()
fem.assemble_vector(b.array, cpp_L)
end = time.perf_counter()
print(end - start)

```

### Output (Python)

```bash
ghcr.io/fenics/dolfinx/dolfinx:nightly gives: 1.091861698001594
ghcr.io/fenics/dolfinx/dolfinx:v0.7.2 gives: 0.03441590800139238
```

### Version

main branch

### DOLFINx git commit

a7fa17b9914564e4672ccb25a3fae7d9f6944aeb

### Installation

Using docker

jorgensd:
Even an issue without mixed elements:
```python
from dolfinx import fem, mesh as dmesh
import basix.ufl
from mpi4py import MPI
import ufl
import time

degree = 1
mesh = dmesh.create_unit_square(MPI.COMM_WORLD, 1000, 1000)

cell = mesh.ufl_cell()
FE = basix.ufl.element("Lagrange", str(cell), degree)
VQ = fem.functionspace(mesh, FE)

# Define test, trial and functions for previous timestep
Az = ufl.TrialFunction(VQ)
vz = ufl.TestFunction(VQ)
dx = ufl.Measure("dx", domain=mesh)
L = vz * dx
cpp_L = fem.form(L)
b = fem.create_vector(cpp_L)
start = time.perf_counter()
fem.assemble_vector(b.array, cpp_L)
end = time.perf_counter()
print(end - start)
```
Main: 1.004702399000962
0.7.2: 0.034690725999098504

[...]

Issue #2884, opened 2023-11-09
`nanobind` leaks memory in builtin mesh generation?


francesco-ballarin:
I have been struggling for the last couple of days in my `multiphenicsx` CI, where I have parallel unit tests that (on the same commit of `multiphenicsx`, and same `dolfinx/dolfinx:nightly` image) take either 10 minutes (as expected) or 3 hours.

I've simplified my CI and ran, out of all my tests, only this test
https://github.com/multiphenics/multiphenicsx/blob/artifact/tests/unit/fem/test_dofmap_restriction.py
which is heavily parametrized, for a total of more than 50 possible configurations.
I then ran only that test, either in serial or in parallel on three processors. I profiled a good CI run (which takes 3.670 s in parallel) and a bad CI run (which takes 122.788 seconds in parallel).

I attach the profiling results in [pstats.zip](https://github.com/FEniCS/dolfinx/files/13307629/pstats.zip): you have two cases (good vs bad), `cProfile` files for a serial vs parallel run, and in each case an auxiliary file `print.py` to print the following tables on process #0 of the parallel run.

For the good case (https://github.com/multiphenics/multiphenicsx/actions/runs/6810825566/job/18519911648), I get:
```
Thu Nov  9 12:00:15 2023    pstats-parallel-0/combined.prof

         321486 function calls (302748 primitive calls) in 3.670 seconds

   Ordered by: cumulative time
   List reduced from 584 to 30 due to restriction <30>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
       63    0.001    0.000    3.819    0.061 runner.py:111(pytest_runtest_protocol)
       63    0.006    0.000    3.798    0.060 runner.py:119(runtestprotocol)
      189    0.001    0.000    3.792    0.020 runner.py:219(call_and_report)
 1309/693    0.004    0.000    3.788    0.005 _hooks.py:479(__call__)
 1309/693    0.002    0.000    3.785    0.005 _manager.py:106(_hookexec)
 1309/693    0.020    0.000    3.784    0.005 _callers.py:27(_multicall)
      189    0.001    0.000    3.759    0.020 runner.py:247(call_runtest_hook)
      189    0.001    0.000    3.755    0.020 runner.py:318(from_call)
      189    0.000    0.000    3.753    0.020 runner.py:262(<lambda>)
       63    0.001    0.000    2.981    0.047 pytest_hooks_unit_tests.py:27(runtest_teardown)
[...]

PR #2894, opened 2023-11-17
Correct CMAKE_BUILD_TYPE in end user Docker images

jorgensd:
Related to the nightly timings #2891.
We still have a regression with the introduction of nanobind, but now we are down to a halving in performance.

### DIFF
--- a/.github/workflows/docker-end-user.yml
+++ b/.github/workflows/docker-end-user.yml
@@ -134,7 +134,6 @@ jobs:
         with:
           build-args: |
             DOLFINX_CMAKE_BUILD_TYPE=RelWithDebInfo
-            DOLFINX_CMAKE_CXX_FLAGS=-O2
             BASEIMAGE=${{ env.BASEIMAGE }}
           context: .
           file: dolfinx/${{ env.DOCKERFILE }}
@@ -147,7 +146,6 @@ jobs:
         with:
           build-args: |
             DOLFINX_CMAKE_BUILD_TYPE=RelWithDebInfo
-            DOLFINX_CMAKE_CXX_FLAGS=-O2
             BASEIMAGE=${{ env.BASEIMAGE }}
           context: .
           file: dolfinx/${{ env.DOCKERFILE }}
@@ -158,7 +156,6 @@ jobs:
         with:
           build-args: |
             DOLFINX_CMAKE_BUILD_TYPE=RelWithDebInfo
-            DOLFINX_CMAKE_CXX_FLAGS=-O2
             BASEIMAGE=${{ env.BASEIMAGE }}
           context: .
           file: dolfinx/${{ env.DOCKERFILE }}
@@ -171,7 +168,6 @@ jobs:
         with:
           build-args: |
             DOLFINX_CMAKE_BUILD_TYPE=RelWithDebInfo
-            DOLFINX_CMAKE_CXX_FLAGS=-O2
             BASEIMAGE=${{ env.BASEIMAGE }}
           context: .
           file: dolfinx/${{ env.DOCKERFILE }}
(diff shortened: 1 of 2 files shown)


jhale:
I would propose we remove `DOLFINX_CMAKE_CXX_FLAGS=-O2` as `RelWithDebInfo` is usually picked as `-O2 -g -DNDEBUG` anyway.

Perhaps we should go for `Release` on these images, that typically enables `-O3`. How useful really is the DebugInfo for users?

jhale:
We can discuss the nanobind performance issues separately.

jorgensd:
> I would propose we remove `DOLFINX_CMAKE_CXX_FLAGS=-O2` as `RelWithDebInfo` is usually picked as `-O2 -g -DNDEBUG` anyway.
> 
> Perhaps we should go for `Release` on these images, that typically enables `-O3`. How useful really is the DebugInfo for users?
> 
> Edit: I'm thinking the tagged versions e.g. `:v0.7.2` should be done with `Release`, and `:nightly` should have `RelWithDebInfo`. It's not urgent, let's switch to `RelWithDebInfo` as you suggest for now.

Agreed, this is a separate issue. It is mostly important to get some better performance on the nightly images.

michalhabera:
Remove CMAKE_CXX_FLAGS, it does not work anyway because `--config-settings=cmake.define.-DCMAKE_CXX_FLAGS` is wrong and should be `--config-settings=cmake.define.CMAKE_CXX_FLAGS`. Moreover, build type ovewrites these, so we'd need `--config-settings=cmake.define.CMAKE_CXX_FLAGS_RELEASE` for Release build etc.

francesco-ballarin:
Once we are done with the change proposed by @michalhabera , can @jorgensd manually trigger https://github.com/FEniCS/dolfinx/actions/workflows/docker-end-user.yml ?
After the nightly image is manually rebuilt, I'll trigger a handful of runs in the multiphenicsx repo to see if it is still affected by #2884

.github/workflows/docker-end-user.yml
```diff
@@ -146,7 +146,7 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           build-args: |
-            DOLFINX_CMAKE_BUILD_TYPE=RelWithDebug
+            DOLFINX_CMAKE_BUILD_TYPE=RelWithDebInfo
             DOLFINX_CMAKE_CXX_FLAGS=-O2
```
jorgensd, line 150:
```suggestion
```

.github/workflows/docker-end-user.yml
```diff
@@ -157,7 +157,7 @@ jobs:
...
```
jorgensd, line 161:
```suggestion
```

.github/workflows/docker-end-user.yml
```diff
@@ -170,7 +170,7 @@ jobs:
...
```
jorgensd, line 174:
```suggestion
```

jorgensd:
> Once we are done with the change proposed by @michalhabera , can @jorgensd manually trigger https://github.com/FEniCS/dolfinx/actions/workflows/docker-end-user.yml ? After the nightly image is manually rebuilt, I'll trigger a handful of runs in the multiphenicsx repo to see if it is still affected by #2884

I can also trigger my own pipelines, as I've got several tests timing out because of this

michalhabera:
@jorgensd Can you also remove `--config-settings=cmake.define.-DCMAKE_CXX_FLAGS`?

michalhabera (approved):

garth-wells:
Let's improve issue and PR names. This is a Docker-only change but the title is very broad.

jorgensd:
> Let's improve issue and PR names. This is a Docker-only change but the title is very broad.

The issue is still a regression issue due to transfering to nanobind. Happy to rename this PR though.

francesco-ballarin:
Can #2895 be merged _before_ this, and then merge the updated `main` into this branch? I'll test the combined effect of both.

garth-wells (approved):

PR #2895, opened 2023-11-17
Change nanobind build options to remove `-Os`

jorgensd:
With gcc, `-Os` (recommended by nanobind) can lead to significantly slower binaries than `-O2`. DOLFINx has considerable templated code, which gets compiled by the nanobind wrappers. This makes performance of the wrapper library sensitive to the compiler options.

This improves the run-time of #2891 to be better than in 0.7.1

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -41,6 +41,7 @@ endif()
 # target_link_libraries
 nanobind_add_module(
   cpp
+  NOMINSIZE
   MODULE
   dolfinx/wrappers/dolfinx.cpp
   dolfinx/wrappers/assemble.cpp
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -42,7 +42,6 @@ endif()
 nanobind_add_module(
   cpp
   NOMINSIZE
-  MODULE
   dolfinx/wrappers/dolfinx.cpp
   dolfinx/wrappers/assemble.cpp
   dolfinx/wrappers/common.cpp


IgorBaratta:
I confirm that the changes fix #2891 .

But:
> Size optimizations can be disabled by specifying the optional NOMINSIZE argument, though doing so is not recommended.

from  https://nanobind.readthedocs.io/en/latest/api_cmake.html

Is it then a bug in nanobind?

francesco-ballarin (approved):

chrisrichardson (commented):
Do we need `MODULE`?

chrisrichardson (approved):

chrisrichardson:
@IgorBaratta - not a bug in nanobind, but a feature of the compiler. Each compiler may treat `-Os` differently.

jorgensd:
> Do we need `MODULE`?

Reading the docs Igor linked to,
> [nanobind_add_module()](https://nanobind.readthedocs.io/en/latest/api_cmake.html#command:nanobind_add_module) performs the following steps to produce bindings.

> It creates a CMake library via add_library(target_name MODULE ...) and enables the use of C++17 features during compilation.

garth-wells:
With macOS/clang, the difference between `-Os` and `-O3` is negligible for the code in https://github.com/FEniCS/dolfinx/issues/2891#issuecomment-1814721725.

jorgensd:
> With macOS/clang, the difference between `-Os` and `-O3` is negligible for the code in [#2891 (comment)](https://github.com/FEniCS/dolfinx/issues/2891#issuecomment-1814721725).

I clearly has a large effect on linux with gcc.

chrisrichardson:
add_library needs MODULE - I don't think nanobind_add_module does.

jorgensd:
> add_library needs MODULE - I don't think nanobind_add_module does.

Feel free to test it. I'm currently rewriting dolfinx_mpc to support nanobind

garth-wells (approved):

Issue #2861, opened 2023-11-03
Reduce number of build threads in Docker end-user image builds


jhale:
### Describe new/missing feature

Using ninja on a 4-core 6GB arm64 instance leads to a crash in the pybind11/Python build. Increasing the instance size to 8-cores works.

Reducing the number of build threads used by ninja on the 4-core instance will probably fix the issue.

garth-wells:
`nanobind` will likely fix this.

jhale:
Indeed. That's why I hold off as it's easier to let ninja do its thing on build threads.

garth-wells:
Working after nanobind switch.

francesco-ballarin:
After #2894 #2895 #2896, this is happening again. I have temporarily increased back the number of cores to 8 in #2897, but that change needs to be reverted at some point to avoid doubling the charge on every nightly run.

garth-wells:
Let's not waste human time on this.

Issue #2846, opened 2023-11-01
`arm64` step of docker end user workflow is failing without a meaningful log


francesco-ballarin:
While testing #2844, both runs in https://github.com/FEniCS/dolfinx/actions/runs/6717503005/job/18255469197 and https://github.com/FEniCS/dolfinx/actions/runs/6717953807/job/18256739847 failed on `arm64` while building the `pybind11`.  In both cases, the output shows
```
#17 36.84   *** Building project with Ninja...
#17 41.33   [1/13] Building CXX object CMakeFiles/cpp.dir/dolfinx/wrappers/dolfinx.cpp.o
#17 43.42   [2/13] Building CXX object CMakeFiles/cpp.dir/dolfinx/wrappers/graph.cpp.o
#17 46.45   [3/13] Building CXX object CMakeFiles/cpp.dir/dolfinx/wrappers/common.cpp.o
#17 47.55   [4/13] Building CXX object CMakeFiles/cpp.dir/dolfinx/wrappers/geometry.cpp.o
#17 52.78   [5/13] Building CXX object CMakeFiles/cpp.dir/dolfinx/wrappers/log.cpp.o
```
and no furthe errors. In #2844 I imagined that we were out of build minutes, but after seeing the same failure twice at exactly the same point, my suspicion would be `/tmp` runs out of space. I don't have access to a `arm64` CPU so I cannot test the workflow locally.

jhale:
I am getting a crash at a different time. The instances have 64GB of SSD so /tmp is not an issue. Possibly a problem at BuildJet's end, will wait a day and see if it sorts itself out.

https://github.com/FEniCS/dolfinx/actions/runs/6732618747

jhale:
No issues building locally.

francesco-ballarin:
> I am getting a crash at a different time.

To me, 
https://github.com/FEniCS/dolfinx/actions/runs/6732618747/job/18299658286
looks like the same point as in the opening post
https://github.com/FEniCS/dolfinx/actions/runs/6717953807/job/18256739847

jhale:
It's an out-of-memory error. The old build system was using make, the new one ninja, which uses more build threads.

Fixed in #2860

jhale:
It's an out-of-memory error. The old build system was using make, the new one ninja, which uses more build threads.

Fixed in #2860.

PR #2897, opened 2023-11-19
Increase again number of cores on ARM buildjet runners while building Docker end-user image

francesco-ballarin:
We are now back to https://github.com/FEniCS/dolfinx/issues/2846, see log in https://github.com/FEniCS/dolfinx/actions/runs/6917539955/job/18818837676

I'll temporarily increase back to 8 cores, which was the temporary fix in https://github.com/FEniCS/dolfinx/pull/2860
and re-open https://github.com/FEniCS/dolfinx/issues/2861 as a reminder to revert this change.

After test in https://github.com/FEniCS/dolfinx/actions/runs/6919475270 runs successfully, I'll merge this

### DIFF
--- a/.github/workflows/docker-end-user.yml
+++ b/.github/workflows/docker-end-user.yml
@@ -64,7 +64,7 @@ jobs:
           - arch_tag: amd64
             os: ubuntu-latest
           - arch_tag: arm64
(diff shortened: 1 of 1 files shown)


jhale:
This can probably be fixed properly by passing a reduced number of ninja build threads to scikit-build-core via a docker build env.

PR #2902, opened 2023-11-21
Remove packages from pyproject project section

jhale:


python/pyproject.toml
```diff
@@ -22,7 +22,6 @@ authors = [
       { email = "fenics-steering-council@googlegroups.com" },
       { name = "FEniCS Steering Council" },
 ]
-packages = ["dolfinx"]
```
garth-wells, line 25: I thought we needed this so pip knows which files to install. Reason being the project name is 'fenics-dolfinx' rather than 'dolfinx'.

jorgensd: I think it is set further down:
```toml
[tool.scikit-build]
wheel.packages = ["dolfinx"]
```

jhale: I thought so too, but I found out using this tool that it's invalid:

https://github.com/abravalheri/validate-pyproject/

It seems like this is meant to be handled by the build tool (scikit-build for us).

jorgensd: See: https://scikit-build.readthedocs.io/en/latest/usage.html#setuptools-options

> packages: Explicitly list of all packages to include in the distribution. Setuptools will not recursively scan the source tree looking for any directory with an __init__.py file. To automatically generate the list of packages, see [Using find_package()](https://setuptools.readthedocs.io/en/latest/setuptools.html#using-find-packages).

Are you sure it should be set like:
```toml
[tool.setuptools]
packages = ["dolfinx"]
```

jorgensd: Nevermind, written in:
https://scikit-build-core.readthedocs.io/en/latest/configuration.html#customizing-the-built-wheel

garth-wells (approved):

PR #2903, opened 2023-11-21
Fix size of map from old to new dofs in `build_collapsed_dofmap`

jpdean:
The block size should be multiplied by rather than added.


garth-wells (approved):
Could a test be added that would have picked up this issue?

jpdean:
The code is here is a bit confusing, so I'm going to close this and tidy it in another PR

PR #2907, opened 2023-11-24
Add test of quadrature element

mscroggs:
See https://github.com/FEniCS/dolfinx/issues/2872#issuecomment-1825356038

This adds a test that was faling before https://github.com/FEniCS/ffcx/pull/641.

Progress towards #2872


python/dolfinx/wrappers/assemble.cpp
```diff
@@ -211,6 +211,48 @@ void declare_assembly_functions(nb::module_& m)
               mat_add, a, std::span(constants.data(), constants.size()),
               py_to_cpp_coeffs(coefficients), bcs);
         }
+        else if (data_bs[0] == 4)
```
garth-wells, line 214: `dolfinx::la::MatrixCSR` should really support runtime as well as compile-time block sizes. @chrisrichardson?

PR #2908, opened 2023-11-24
Unfix Sphinx version and remove Sphinx Makefile

jhale:
- Unpin sphinx.
- Removes trivial `Makefile`: Sphinx now has a proper entry point which means using `python -m sphinx` is more reliable, particularly within virtual environments.
- Update `python/doc/README.md` with one liner to build docs.


garth-wells (approved):

garth-wells (changes requested):
It would be good to keep the `-W` option.

PR #2909, opened 2023-11-26
Add `ruff` to base test environment

francesco-ballarin:
In preparation of a PR similar to https://github.com/FEniCS/basix/pull/748 for `dolfinx`.


francesco-ballarin:
I can do the same for all remaining docker files too, if the transition in https://github.com/FEniCS/basix/pull/748 is successful. I just wanted to open this before transition to `ruff` takes place in `dolfinx` as a reminder to avoid that the `ruff` PR plasters workflow files with `pip install`s.

PR #2912, opened 2023-11-28
Update GH bug template

garth-wells:
Add summary box and update versions


francesco-ballarin:
Maybe add a note to https://github.com/FEniCS/dolfinx/blob/main/RELEASE.md so that @jhale has a reminder of this while preparing the next release?

PR #2914, opened 2023-11-29
Fix misleading commit message when updating documentation repository

francesco-ballarin:
Noticed this while I was investigating whether to prepare an equivalent of https://github.com/FEniCS/basix/pull/751 for `dolfinx`. At first glance I couldn't understand why there wasn't a push of the python part of the documentation: there was, but it was stored alongside the cpp in a commit message starting with `C++`.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -210,5 +210,5 @@ jobs:
           git config --global user.email "fenics@github.com"
           git config --global user.name "FEniCS GitHub Actions"
           git add --all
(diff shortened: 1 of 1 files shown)


francesco-ballarin:
For reference, the reason why this doesn't need a change similar to the one in https://github.com/FEniCS/basix/pull/751 is because of/thanks to the use of `*` in

```
      - name: Copy documentation into repository
        ...
        run: |
          ...
          cp -r ../python/doc/build/html/* dolfinx/${{ env.VERSION_NAME }}/python
```

PR #2916, opened 2023-11-29
Encourage proper management of subspaces in user code

IgorBaratta:
We've observed a common pattern where subspaces are frequently extracted but not stored, leading to unnecessary overhead in smart pointer management and  repetitive creation and destruction of subspaces.

The idea of this PR is to encourage users to retain their subspaces, eg:
```python3
V0 = V.sub(0)
...
# reuse V0
```


cpp/dolfinx/fem/FunctionSpace.h
```diff
@@ -62,7 +61,11 @@ class FunctionSpace
...
 
-  /// Extract subspace for component
+  ///  @brief Extracts a subspace for a specific component. 
+  ///
+  /// @note For performance reasons, it's advisable to store and reuse the
+  /// returned subspace instead of calling this function repeatedly for the same
+  /// component.
   /// @param[in] component The subspace component
   /// @return The subspace
   std::shared_ptr<FunctionSpace<T>> sub(const std::vector<int>& component) const
```
garth-wells, line 71: I think it would be clearer if the return type was `FunctionSpace<T>` (not a shared_ptr) and we remove the cache.

IgorBaratta: I've updated the return type.

PR #2924, opened 2023-12-07
Add Basix dependency to `pyproject.toml`

garth-wells:


garth-wells:
> I think there was an argument in the past that Python dolfinx did not depend directly on the Python interface to Basix. This may no longer be true with the recent changes relayed to element construction.

In the past `basix` was not imported. `basix` is now imported.

Issue #2929, opened 2023-12-12
[BUG]: VTXWriter outputs erroneous data in parallel when outputting collapsed space


nate-sime:
### Summarize the issue

Using VTXWriter to output a DG function and one of its collapsed sub functions yields incorrect data in parallel.

### How to reproduce the bug

The following MWE and attached paraview rendering illustrates the issue. The image is produced after running with 5 MPI processes.

### Minimal Example (Python)

```Python
import numpy as np
from mpi4py import MPI
import dolfinx

mesh = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, 4, 4)

V = dolfinx.fem.VectorFunctionSpace(mesh, ("DG", 1), dim=2)
u_vec = dolfinx.fem.Function(V, name="u")

u_vec.interpolate(lambda x: np.stack((x[0], x[1])))
u_vec.x.scatter_forward()
u1 = u_vec.sub(0).collapse()
u1.x.scatter_forward()
u1.name = "u1"

# Fails in parallel with odd ghosting issues
with dolfinx.io.VTXWriter(
        mesh.comm, "solution.bp", [u_vec, u1], engine="BP4") as vtx:
    vtx.write(0.0)

# Works fine
with dolfinx.io.VTXWriter(
        mesh.comm, "solution_u1.bp", [u1], engine="BP4") as vtx:
    vtx.write(0.0)
```

### Version

0.7.2

### Installation

From source

### Additional information

![image](https://github.com/FEniCS/dolfinx/assets/34454754/558b3aef-45ba-4504-8d4b-da79e52b384d)

jorgensd:
Same issue on main with 3 processes and:
```python
import numpy as np
from mpi4py import MPI
import dolfinx

mesh = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, 2, 2)

V = dolfinx.fem.functionspace(mesh, ("DG", 1, (2, )))
u_vec = dolfinx.fem.Function(V, name="u")

u_vec.interpolate(lambda x: np.stack((x[0], x[1])))
u_vec.x.scatter_forward()
u1 = u_vec.sub(0).collapse()
u1.x.scatter_forward()
u1.name = "u1"

# Fails in parallel with odd ghosting issues
with dolfinx.io.VTXWriter(
        mesh.comm, "solution.bp", [u_vec, u1], engine="BP4") as vtx:
    vtx.write(0.0)

# Works fine
[...]

Issue #2881, opened 2023-11-07
`IndexMap::create_submap` reshuffles ghost nodes: is this expected?


francesco-ballarin:
Consider the following example in which I create a submap of a `IndexMap` which actually doesn't throw away any index.

```python3
import dolfinx.fem
import dolfinx.mesh
import mpi4py.MPI
import numpy as np

mesh = dolfinx.mesh.create_unit_square(mpi4py.MPI.COMM_WORLD, 4, 4)
V = dolfinx.fem.functionspace(mesh, ("Lagrange", "1"))
entities = np.arange(0, V.dofmap.index_map.size_local)
index_map_sub, ghost_map_sub = V.dofmap.index_map.create_submap(entities)
print(mpi4py.MPI.COMM_WORLD.rank, "Ghost map", ghost_map_sub)
print(mpi4py.MPI.COMM_WORLD.rank, "Original map ghosts", V.dofmap.index_map.ghosts)
print(mpi4py.MPI.COMM_WORLD.rank, "Sub map ghosts", index_map_sub.ghosts)
```

If I run it with 2 processes on my laptop I get
```
0 Ghost map [0 1 2 3 4 5 6]
0 Original map ghosts [19 21 22 24 17 18 23]
0 Sub map ghosts [19 21 22 24 17 18 23]
1 Ghost map [0 1 2 3]
1 Original map ghosts [7 0 2 5]
1 Sub map ghosts [7 0 2 5]
```
that is the ghosts in the original and new maps are ordered the same way.

If I run it with 3 processes on my laptop I get
```
0 Ghost map [ 0  1  2  4  7  9  3  5  6  8 10]
0 Original map ghosts [12 10  9 16  8 19 23 11 17  6 21]
0 Sub map ghosts [12 10  9  8 11  6 16 19 23 17 21]
1 Ghost map [0 1 3 9 2 4 5 6 7 8]
1 Original map ghosts [ 0  1 24  2 16 14 15 19 23  5]
1 Sub map ghosts [ 0  1  2  5 24 16 14 15 19 23]
2 Ghost map [0 1 6 8 2 3 4 5 7]
2 Original map ghosts [ 3  1  9 11  8 10  2  6  5]
2 Sub map ghosts [ 3  1  2  5  9 11  8 10  6]
```
i.e., some of the ghosts have been swapped.

Is this expected? I imagined that the submap in this case was supposed to be the identity.

[...]

PR #2930, opened 2023-12-13
Sort ghost indices in sub-dofmap based on original ordering

jorgensd:
Relates to: #2881, but now in the context of usage in #2929.
Currently, collapsing a dofmap of a blocked space states that is does not reorder the degrees of freedom.
However, the ghost dofs are re-ordered, as the new index map is created with `create_sub_map`.

This PR adds in a sorting of the ghost indices, based on the indices of the initial ghosts.
It also relates to #2919.

### DIFF
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -625,8 +625,23 @@ IndexMap::create_submap(std::span<const std::int32_t> indices) const
 
   if (_overlapping)
   {
-    return {IndexMap(_comm.comm(), local_size_new, ghosts, src_ranks),
-            std::move(new_to_old_ghost)};
+    // Sort ghosts according to orignal ordering
+    std::vector<std::int64_t> sorted_ghosts(ghosts.size());
+    std::vector<int> sorted_ranks(src_ranks.size());
+    std::vector<std::int32_t> ghost_order(new_to_old_ghost.size());
+
+    // Sort ghosts by their original ordering in _ghosts
+    std::vector<std::int32_t> perm(new_to_old_ghost.size());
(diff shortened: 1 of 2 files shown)


francesco-ballarin:
So basically now we are moving to the position I used to have in https://github.com/FEniCS/dolfinx/issues/2881#issuecomment-1799531959?

I am fine either way, I am just asking because if we do guarantee that ghosts are now sorted as in the original ordering, I will beef up again my asserts :stuck_out_tongue:

francesco-ballarin:
@jorgensd @jpdean need to decide which one you'd like to get merged first between #2890 and #2930, since they are changing the same functionality, and then update the other PR.

garth-wells:
There are a few things to work through first, see https://github.com/FEniCS/dolfinx/issues/2929#issuecomment-1860033785. 

We want to aim for imposing as few constraints as reasonably possible, so ideally we don't want to force the re-ordering. If the caller needs a certain order then it should be possible for the caller to do the re-ordering.

jorgensd:
> There are a few things to work through first, see [#2929 (comment)](https://github.com/FEniCS/dolfinx/issues/2929#issuecomment-1860033785).
> 
> We want to aim for imposing as few constraints as reasonably possible, so ideally we don't want to force the re-ordering. If the caller needs a certain order then it should be possible for the caller to do the re-ordering.

the points you have made in the referenced issue does not address the underlying issue, that if you create a sub index map of all indices, it ends up shuffling the ghosts.
The issue is that the shuffling of ghosts, when done on a collapsed map, completely changes the dofmap for the collapsed space of a blocked element, which does not make sense to me.
The re-ordering of a dofmap based on this index map is way more expensive than reordering prior to the «sub» index map construction.

garth-wells:
Superseded by #3091

PR #2931, opened 2023-12-14
Improve overview of online documentation with autosummary

jorgensd:
Expose autosummary for __all__ variables: https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html#confval-autosummary_ignore_module_all

### DIFF
--- a/python/doc/source/conf.py
+++ b/python/doc/source/conf.py
@@ -143,6 +143,7 @@ autodoc_default_options = {
     "undoc-members": True,
 }
 autosummary_generate = True
+autosummary_ignore_module_all = False
 autoclass_content = "both"
 
 napoleon_google_docstring = True


francesco-ballarin:
(Partially OT): except for maybe `dolfinx/__init__.py`, do we need `__all__` in the other submodules? It gets populated by default with everything that gets imported in the `__init__.py` file.

jorgensd:
> (Partially OT): except for maybe `dolfinx/__init__.py`, do we need `__all__` in the other submodules? It gets populated by default with everything that gets imported in the `__init__.py` file.

We need `__all__` for autosummary to work.

jorgensd:
There are also functions that we import that we do not want to expose (such as helper functions, external imports etc).

PR #2935, opened 2023-12-15
Improve `dolfinx.common` documentation

jorgensd:
Expose various functions through `__all__`.

Update docs for timings, and add more documentation.

Add documentation for user-facing IndexMap.

### DIFF
--- a/python/dolfinx/common.py
+++ b/python/dolfinx/common.py
@@ -5,28 +5,40 @@
 # SPDX-License-Identifier:    LGPL-3.0-or-later
 """General tools for timing and configuration."""
 
+from mpi4py import MPI as _MPI
+
 import functools
 import typing
 
 from dolfinx import cpp as _cpp
-from dolfinx.cpp.common import (IndexMap, git_commit_hash, has_adios2,  # noqa
-                                has_debug, has_kahip, has_parmetis)
+from dolfinx.cpp.common import has_adios2  # noqa
+from dolfinx.cpp.common import (IndexMap, git_commit_hash, has_debug,
+                                has_kahip, has_parmetis)
 
-__all__ = ["IndexMap", "Timer", "timed"]
+__all__ = ["IndexMap", "Timer", "timed", "list_timings", "timing", "TimingType", "Reduction"]
 
 TimingType = _cpp.common.TimingType
 Reduction = _cpp.common.Reduction
 
 
 def timing(task: str):
+    """Return timing (count, total wall time, total user time, total system time) for given task."""
     return _cpp.common.timing(task)
 
 
-def list_timings(comm, timing_types: list, reduction=Reduction.max):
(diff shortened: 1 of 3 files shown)


python/dolfinx/common.py
```diff
@@ -5,35 +5,47 @@
...
+    """Print out a summary of all Timer measurements.
+
+    One can specify the timing types as a subset of wall time, system time or user time.
+
+    Note:
+        When used in parallel, a reduction is applied across all processes. By default, the maximum time is shown.
+
+    Args:
+        comm: Communicator to reduce timings over
+        timing_types: List of timings to return
```
garth-wells, line 39: What's the type (e.g., list of enums?) and what are the possible values?

python/dolfinx/common.py
```diff
@@ -5,35 +5,47 @@
...
 
-def list_timings(comm, timing_types: list, reduction=Reduction.max):
-    """Print out a summary of all Timer measurements, with a choice of
-    wall time, system time or user time. When used in parallel, a
-    reduction is applied across all processes. By default, the maximum
-    time is shown."""
+def list_timings(comm: _MPI.Comm, timing_types, reduction=Reduction.max):
+    """Print out a summary of all Timer measurements.
+
+    One can specify the timing types as a subset of wall time, system time or user time.
```
garth-wells, line 32: I think this is redundant if `timing_types` is expanded on.

python/dolfinx/common.py
```diff
@@ -91,7 +102,17 @@ def elapsed(self):
 
 
 def timed(task: str):
-    """Decorator for timing functions."""
+    """Decorator for timing functions. The basic usage is::
+
+        @timed('Description of function')
```
garth-wells, line 107: Google doctring style uses `Examples:` section for code example usage, see https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html.

python/dolfinx/wrappers/common.cpp
```diff
@@ -145,16 +161,32 @@ void common(nb::module_& m)
             return std::pair(std::move(map),
                              dolfinx_wrappers::as_nbarray(std::move(ghosts)));
           },
-          nb::arg("entities"));
+          nb::arg("entities"),
+          nb::raw_doc("Create new index map from a subset of indices in this "
```
garth-wells, line 165: This is too fragile. The approach we've adopted (which is not fully complete) is to document from Python.

python/dolfinx/common.py
jorgensd: Could not use type hints due to nanobind missing stubs (so mypy failed).
Can add a reference to the cpp documentation

python/dolfinx/wrappers/common.cpp
jorgensd: So then we need to wrap every function we expose to the Python interface?

PR #2937, opened 2023-12-21
Add SECURITY.md

jhale:
Recommended by GitHub for meeting best Community Standards.


garth-wells:
I don't see the value in this for us, but do see future maintenance effort. DOLFINx isn't a system library; security concerns that this change might address seem vanishingly small to me. Or am I missing something?

PR #2950, opened 2024-01-02
Simpify mesh interfaces

### LINKED ISSUE #2949: Simplify mesh creation interface (remove AdjacencyList for cells, use flat or 2D array)

garth-wells:
Removes complication of supporting 'mixed-topology' meshes, which aren't (yet) supported and for which the appropriate interface is not yet known.

This PR removed complexity that is unlikely to serve a purpose.  

Fixes #2949.


chrisrichardson (commented):
I think too much has been removed. We will still need to have multiple cell types and multiple coordinate maps in topology and geometry. Can we just keep the lists, and use the first entry?

garth-wells:
> I think too much has been removed. We will still need to have multiple cell types and multiple coordinate maps in topology and geometry. Can we just keep the lists, and use the first entry?

We should add it back later but o it differently (bottom-up).

We should keep `Geometry::cmap`, etc, and later when we're ready add  `Geometry::cmaps`, while keeping `Geometry::cmap`. Code continue to use `Geometry::cmap` until it supports  `Geometry::cmaps`. The attempt we made introduced code litter all over the place with loads check with assertions that add nothing to the code.

garth-wells:
Closing in favour of an updated branch.

Issue #202, opened 2018-10-09
Add __citation__ PEP compatibility


jhale:
As discussed at NumFOCUS summit.

https://github.com/adrn/CitationPEP

chrisrichardson:
What should go in the citation? The FEniCS Book (which is gradually becoming obsolete)?
Or should we push to zenodo, get a doi, and use that as the citation? (which is a bit chicken and egg) - can we get a DOI from zenodo before packing up the repo?

jhale:
The FEniCS Book is obsolete, apart from describing the academic goals of the project. On the other hand, it does pickup citations for some people :)

I know that Figshare will issue DOIs prior to making an item public. I find Zenodo a bit complex, but I'm sure it will do the same.

wence-:
You can pre-reserve a zenodo DOI for an upload before finalising the contents.

PR #2958, opened 2024-01-04
add __citation__

mscroggs:
See #202


mscroggs:
Closing this as CITATION.cff is sufficient

PR #2963, opened 2024-01-05
Fix `IndexMap->global_to_local()`

### LINKED ISSUE #2962: `dolfinx::common::IndexMap->global_to_local` is broken
### Summarize the issue

The function does no longer return -1 for all indices not owned or ghosted on the current process.
Probably introduced in #2890.

### How to reproduce the bug

Minimal C++ example below (as the function is not exposed to Python).
On v0.7.3 one gets the following output by running it on two processes
```bash
0->0
1->1
2->2
3->3
4->-1
5->-1
6->5
7->4

Local range:4<8
0->5
1->4
2->-1
3->-1
4->0
5->1
6->2
7->3
```
which is expected.

### Minimal Example (Python)

```Python
#include <basix/e-lagrange.h>
#include <cmath>
#include <dolfinx.h>
#include <utility>
#include <vector>

using namespace dolfinx;
using T = double;
using U = typename dolfinx::scalar_value_type_t<T>;

int main(int argc, char* argv[])
{
  MPI_Init(&argc, &argv);

  {
    // Create mesh and function space
    auto part = mesh::create_cell_partitioner(mesh::GhostMode::shared_facet);
    auto mesh = std::make_shared<mesh::Mesh<U>>(
        mesh::create_rectangle<U>(MPI_COMM_WORLD, {{{0.0, 0.0}, {2.0, 1.0}}},
                                  {3, 1}, mesh::CellType::triangle, part));

    basix::FiniteElement element = basix::element::create_lagrange<double>(
        mesh::cell_type_to_basix_type(mesh->topology()->cell_types()[0]), 1,
        basix::element::lagrange_variant::equispaced, false);
    auto V = std::make_shared<fem::FunctionSpace<double>>(
        fem::create_functionspace(mesh, element, std::vector<std::size_t>{3}));
    auto imap = V->dofmap()->index_map;
    std::vector<std::int64_t> global_indices(imap->size_global());
    std::iota(global_indices.begin(), global_indices.end(), 0);
    std::vector<std::int32_t> local_indices(imap->size_global(), 0);
    imap->global_to_local(global_indices, local_indices);
    std::stringstream cc;
    auto range = imap->local_range();
    cc << "Local range:" << range[0] << "<" << range[1] << "\n";
    for (std::int64_t i = 0; i < imap->size_global(); ++i)
    {
      cc << global_indices[i] << "->" << local_indices[i] << "\n";
    }
    std::cout << cc.str() << "\n";
  }

  MPI_Finalize();

  return 0;
}
```

### Output (Python)

```bash
Local range:0<4
0->0
1->1
2->2
3->3
4->5
5->5
6->5
7->4

Local range:4<8
0->5
1->4
2->-1
3->-1
4->0
5->1
6->2
```

### Version

main branch

### DOLFINx git commit

_No response_

### Installation

_No response_

### Additional information

_No response_

jorgensd:
**Issue**
Current check for a global value coming into `global_to_local`  has several issues

1. If the global value is lower than the lowest existing ghost on process
2. If the value is within the range of the smallest and largest ghost value, but not equal

**Fixes**
- Check for equality if not at end of ghost array

Resolves #2962

### DIFF
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -865,19 +865,25 @@ void IndexMap::global_to_local(std::span<const std::int64_t> global,
   for (std::size_t i = 0; i < _ghosts.size(); ++i)
     global_to_local[i] = {_ghosts[i], i + local_size};
 
-  std::sort(global_to_local.begin(), global_to_local.end());
+  std::sort(global_to_local.begin(), global_to_local.end(),
+            [](auto a, auto b) { return a.first < b.first; });
   std::transform(global.begin(), global.end(), local.begin(),
                  [range = _local_range,
                   &global_to_local](std::int64_t index) -> std::int32_t
                  {
                    if (index >= range[0] and index < range[1])
                      return index - range[0];
+                   else if (index < global_to_local.front().first
+                            or index > global_to_local.back().first)
+                     return -1;
                    else
                    {
                      auto it = std::lower_bound(
                          global_to_local.begin(), global_to_local.end(), index,
                          [](auto a, auto b) { return a.first < b; });
-                     return it != global_to_local.end() ? it->second : -1;
+                     return (it != global_to_local.end() and it->first == index)
+                                ? it->second
+                                : -1;
                    }
                  });
 }
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -3,12 +3,12 @@
 // This file is part of DOLFINx (https://www.fenicsproject.org)
 //
 // SPDX-License-Identifier:    LGPL-3.0-or-later
-
 #include "IndexMap.h"
 #include "sort.h"
 #include <algorithm>
 #include <cstdint>
 #include <functional>
+#include <iostream>
 #include <numeric>
 #include <span>
 #include <utility>
@@ -872,9 +872,6 @@ void IndexMap::global_to_local(std::span<const std::int64_t> global,
                  {
                    if (index >= range[0] and index < range[1])
                      return index - range[0];
-                   else if (index < global_to_local.front().first
-                            or index > global_to_local.back().first)
-                     return -1;
                    else
                    {
                      auto it = std::lower_bound(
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -865,19 +865,25 @@ void IndexMap::global_to_local(std::span<const std::int64_t> global,
   for (std::size_t i = 0; i < _ghosts.size(); ++i)
     global_to_local[i] = {_ghosts[i], i + local_size};
 
-  std::sort(global_to_local.begin(), global_to_local.end());
+  std::sort(global_to_local.begin(), global_to_local.end(),
```
chrisrichardson, line 868: Why is this necessary? Comparison of `std::pair` will use first element anyway.

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -865,19 +865,25 @@ void IndexMap::global_to_local(std::span<const std::int64_t> global,
...
+            [](auto a, auto b) { return a.first < b.first; });
   std::transform(global.begin(), global.end(), local.begin(),
                  [range = _local_range,
                   &global_to_local](std::int64_t index) -> std::int32_t
                  {
                    if (index >= range[0] and index < range[1])
                      return index - range[0];
+                   else if (index < global_to_local.front().first
```
chrisrichardson, line 876: Is this needed? Seems like it will be caught below.

jorgensd: Where is the caught below?
And even if it is caught for an element outside of the range (on the upper bound) it requires looping through the whole array?

jorgensd: Removing as it doesn't work with this check in the corner case where a process has no ghosts.

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -3,12 +3,12 @@
 // This file is part of DOLFINx (https://www.fenicsproject.org)
 //
 // SPDX-License-Identifier:    LGPL-3.0-or-later
-
```
garth-wells, line 6: Add line break back.

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -3,12 +3,12 @@
...
 #include "IndexMap.h"
 #include "sort.h"
 #include <algorithm>
 #include <cstdint>
 #include <functional>
+#include <iostream>
```
garth-wells, line 11: Remove `iostream` include.

garth-wells (changes requested):
Looks good.

Would a test be hard to add?

chrisrichardson (approved):

jorgensd:
> Looks good.
> 
> Would a test be hard to add?

If we expose `global_to_local` to the Python layer the test is quite easy to make.
A bit more of a pain in the C++ interface

PR #2964, opened 2024-01-05
Update stokes demo tolerance

IgorBaratta:
Currently the tolerance is too tight, and even small changes in the kernel (for assembling a matrix or vector) could cause the demo to fail.

The current values for `norm_u_1`, `norm_u_0` are:
Max absolute difference: 1.52587891e-05
Max relative difference: 8.73011621e-07


IgorBaratta:
> What's the difference between `np.allclose` and `np.testing.assert_allclose`?

In `np.testing.assert_allclose` conflicting values are appended to the error message.
Eg:
https://github.com/FEniCS/dolfinx/actions/runs/7422791174/job/20198838085#step:22:128

IgorBaratta:
Example of error message with np.testing:

``` bash
AssertionError: 
Not equal to tolerance rtol=1e-15, atol=0

Mismatched elements: 1 / 1 (100%)
Max absolute difference: 7.8864737e-09
Max relative difference: 4.51213336e-10
 x: array(17.47837)
 y: array(17.47837)
 ```
whereas with `assert isclose` we get:
```bash
  File "/~/dolfinx/python/demo/demo_stokes.py", line 541, in <module>
    assert np.isclose(norm_u_1, norm_u_0, rtol=1e-15)
AssertionError
```

garth-wells (approved):

Issue #2946, opened 2023-12-31
[BUG]: test_timer.py test_context_manager_named sometimes fails in complex build


drew-parsons:
### Summarize the issue

Python unit test common/test_timer.py is sometimes failing test_context_manager_named in the complex number build with PETSc 3.19 on amd64.

Full test log at https://ci.debian.net/data/autopkgtest/testing/amd64/f/fenics-dolfinx/41398806/log.gz

The discrepancy is small, should the tolerance in the test be relaxed?

### How to reproduce the bug

In principle
`PETSC_DIR=/usr/lib/petscdir/petsc-complex SLEPC_DIR=/usr/lib/slepcdir/slepc-complex pytest-3 -v python/test/unit/common/test_timer.py`
though it does not fail on my own local machine.

### Output (Python)

```bash
2193s __________________________ test_context_manager_named __________________________
2193s 
2193s     def test_context_manager_named():
2193s         """Test that named Timer works as context manager"""
2193s         task = "test_context_manager_named_str"
2193s     
2193s         # Execute task in the context manager
2193s         t = common.Timer(task)
2193s         sleep(0.05)
2193s >       assert t.elapsed()[0] >= 0.05
2193s E       assert 0.04 >= 0.05
2193s 
2193s python/test/unit/common/test_timer.py:21: AssertionError
```

### Version

0.7.2

### DOLFINx git commit

debian build 1:0.7.3-2
(git commit 25db9a743ef95a78f686f05e38607aa18281b0d6)

### Installation

debian build 1:0.7.3-2 (debian unstable)

drew-parsons:
Given the nature of the test, simply testing timings, it doesn't really make sense that it's actually related to the complex build or to PETSc 3.19.  It must be a question of the precision of the values returned by `t.elapsed()`, or of the precision in the execution of `sleep(0.05)`, with some random variation in the results.

drew-parsons:
dolfinx's Timer is  boost::timer::cpu_timer.  The Boost version used here is 1.83.0.

drew-parsons:
Showing the randomness of the discrepancy, the test passed on a repeat run,
https://ci.debian.net/data/autopkgtest/testing/amd64/f/fenics-dolfinx/41401318/log.gz

drew-parsons:
For what it's worth, legacy dolfin is also affected, https://ci.debian.net/data/autopkgtest/unstable/amd64/d/dolfin/41436063/log.gz
[...]

PR #2966, opened 2024-01-05
Timer fix

jhale:
This makes the comparisons less silly but doesn't fix any issues in Boost or Python sleep.


jhale:
Related to issue #2946.

PR #2968, opened 2024-01-06
Improve FFI support for calling C functions from Python/Numba

garth-wells:
- Move Numba/ctypes/CFFI helpers into a utility module `dolfinx.utils` (is the name ok?).
- Simplify custom assembly and kernel tests.
- Parameterise custom assembly/kernel tests over scalar type.
- Add case using Numba -specific support for calling C functions.


python/demo/demo_static-condensation.py
```diff
@@ -54,15 +55,12 @@
 infile.close()
 
 # Stress (Se) and displacement (Ue) elements
-Se = element("DG", msh.basix_cell(), 1, shape=(2, 2), symmetry=True)
-Ue = element("Lagrange", msh.basix_cell(), 2, shape=(2,))
+Se = element("DG", msh.basix_cell(), 1, shape=(2, 2), symmetry=True, dtype=PETSc.RealType)  # type: ignore
+Ue = element("Lagrange", msh.basix_cell(), 2, shape=(2,), dtype=PETSc.RealType)  # type: ignore
```
jorgensd, line 59: Do we want to use `dolfinx.default_real_type` here instead of `PETSc.RealType`?

python/demo/demo_static-condensation.py
```diff
@@ -42,6 +42,7 @@
 from dolfinx.io import XDMFFile
 from dolfinx.jit import ffcx_jit
 from dolfinx.mesh import locate_entities_boundary, meshtags
+from dolfinx.utils import numba_ufcx_kernel_signature as ufcx_signature
```
jhale, line 45: I would propose we move this routine to FFCx so it sits with the C definition of the UFCx interface. I already started a branch so happy to do it?

python/demo/demo_static-condensation.py
jhale: This seems like a broader question; if a demo uses PETSc for the final solve, should we be recommending the use of PETSc types throughout?

python/demo/demo_static-condensation.py
```diff
@@ -99,45 +97,42 @@ def sigma_u(u):
...
-    numba.types.CPointer(numba.typeof(PETSc.RealType())),  # type: ignore
-    numba.types.CPointer(numba.types.int32),
-    numba.types.CPointer(numba.types.uint8))
 
+# Get local dofmap sizes for later local tensor tabulations
+Ssize = S.element.space_dimension
+Usize = U.element.space_dimension
+
+
+@numba.cfunc(ufcx_signature(PETSc.ScalarType, PETSc.RealType), nopython=True)  # type: ignore
```
jhale, line 132: Keep the `kernel` in `ufcx_kernel_signature`?

python/dolfinx/utils.py
```diff
@@ -0,0 +1,233 @@
...
+        MatSetValuesLocal = _lib_cffi.MatSetValuesLocal
+        """See PETSc `MatSetValuesLocal
+        <https://petsc.org/release/manualpages/Mat/MatSetValuesLocal>`_
+        documentation."""
+
+        MatSetValuesBlockedLocal = _lib_cffi.MatSetValuesBlockedLocal
+        """See PETSc `MatSetValuesBlockedLocal
+        <https://petsc.org/release/manualpages/Mat/MatSetValuesBlockedLocal>`_
+        documentation."""
+    except ImportError:
```
jhale, line 209: Re-raise exception? Otherwise no error and no failure.

python/dolfinx/utils.py
```diff
@@ -0,0 +1,233 @@
...
+        pass
+
+
+def numba_ufcx_kernel_signature(dtype: _npt.DTypeLike, xdtype: _npt.DTypeLike):
```
jhale, line 213: As above, would propose moving this to FFCx.

jhale (approved):
This is a big improvement, few minor comments.

jhale:
[jhale/improve-ffi](https://github.com/FEniCS/dolfinx/tree/jhale/improve-ffi) and https://github.com/FEniCS/ffcx/tree/jhale/low-level-kernel-smoothing

python/demo/demo_static-condensation.py
chrisrichardson: I think that makes sense. For the tests, I think we should ultimately aim to make them PETSc-free unless they are actually testing PETSc integration.

garth-wells: This demo uses PETSc solvers and matrices directly, so `dtype=PETSc.RealType` is appropriate in this case.

Ideally, in the end we can remove `dolfinx.default_real_type` and use `PETSc.RealType` where PETSc is explicitly being used.

python/demo/demo_static-condensation.py
jhale: It's fine, leave as is.

python/demo/demo_static-condensation.py
jhale: Makes sense.

jhale:
To be merged alongside https://github.com/FEniCS/ffcx/pull/646

DOLFINx requires post-merge CI branch cleanup.

PR #2960, opened 2024-01-04
Support creation of meshes from Basix elements from Python

garth-wells:
Until now, an `ufl.Mesh` was required to create a DOLFINx mesh. With this change a mesh can be created with a Basix element or a DOLFINx `CoordinateElement`.

Resolves #2943.

PR #2969, opened 2024-01-07
Further typing fixes

francesco-ballarin:
Fixes to #2960 in addition to those already in #2967.

Should fix the runtime error 
```
E       TypeError: Cannot interpret 'typing.Union[numpy.dtype[typing.Any], NoneType, type[typing.Any], numpy._typing._dtype_like._SupportsDType[numpy.dtype[typing.Any]], str, tuple[typing.Any, int], tuple[typing.Any, typing.Union[typing.SupportsIndex, collections.abc.Sequence[typing.SupportsIndex]]], list[typing.Any], numpy._typing._dtype_like._DTypeDict, tuple[typing.Any, typing.Any]]' as a data type
```
that appears when calling `dolfinx.io.gmshio.ufl_mesh` e.g. in ngsPETSc by @UZerbinati at
https://github.com/NGSolve/ngsPETSc/actions/runs/7437684405/job/20235586388
(or at least, make it into the more meaningful error "from now on you must also pass dtype")

### DIFF
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -260,7 +260,7 @@ class Function(ufl.Coefficient):
                               _cpp.fem.Function_float32, _cpp.fem.Function_float64]
 
     def __init__(self, V: FunctionSpace, x: typing.Optional[la.Vector] = None,
-                 name: typing.Optional[str] = None, dtype: npt.DTypeLike = None):
+                 name: typing.Optional[str] = None, dtype: typing.Optional[npt.DTypeLike] = None):
         """Initialize a finite element Function.
 
         Args:
--- a/python/dolfinx/io/gmshio.py
+++ b/python/dolfinx/io/gmshio.py
@@ -38,7 +38,7 @@ _gmsh_to_cells = {1: ("interval", 1), 2: ("triangle", 1),
                   92: ("hexahedron", 3)}
 
 
-def ufl_mesh(gmsh_cell: int, gdim: int, dtype=npt.DTypeLike) -> ufl.Mesh:
+def ufl_mesh(gmsh_cell: int, gdim: int, dtype: npt.DTypeLike) -> ufl.Mesh:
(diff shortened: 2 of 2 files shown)


garth-wells (approved):

python/dolfinx/fem/function.py
```diff
@@ -260,7 +260,7 @@ class Function(ufl.Coefficient):
                               _cpp.fem.Function_float32, _cpp.fem.Function_float64]
 
     def __init__(self, V: FunctionSpace, x: typing.Optional[la.Vector] = None,
-                 name: typing.Optional[str] = None, dtype: npt.DTypeLike = None):
+                 name: typing.Optional[str] = None, dtype: typing.Optional[npt.DTypeLike] = None):
```
francesco-ballarin, line 263: `npt.DTypeLike` apparently already includes a `typing.Union[...., None]`, so technically this may not be required. If we want to make use of this fact, I will revert this change, but for consistency I will also remove `typing.Optional` from the other occurrence of `dtype = None` in this file.

python/dolfinx/io/gmshio.py
```diff
@@ -60,7 +60,7 @@ def ufl_mesh(gmsh_cell: int, gdim: int, dtype=npt.DTypeLike) -> ufl.Mesh:
     cell = ufl.Cell(shape, geometric_dimension=gdim)
     element = basix.ufl.element(basix.ElementFamily.P, cell.cellname(), degree,
                                 basix.LagrangeVariant.equispaced, shape=(gdim,),
-                                gdim=gdim, dtype=dtype)
+                                gdim=gdim, dtype=dtype)  # type: ignore[arg-type]
```
francesco-ballarin, line 63: `basix.ufl.element` is decorated with `_functools.lru_cache()`, and there doesn't seem to be a way to make that decorator behave nicely with `mypy` for python < 3.11, see https://github.com/python/mypy/issues/5107, hence the ignore.

PR #2972, opened 2024-01-08
Add `IndexMap::global_to_local` to the Python interface

jorgensd:
Expose `dolfinx::common::IndexMap::global_to_local` to Python interface.
Add appropriate test.
Motivated by the bug found in #2962


python/dolfinx/wrappers/common.cpp
```diff
@@ -125,14 +125,23 @@ void common(nb::module_& m)
...
-            self.local_to_global(std::span(local.data(), local.size()), global);
-            return dolfinx_wrappers::as_nbarray(std::move(global));
+            self.local_to_global(std::span(local.data(), local.size()),
+                                 std::span(global.data(), global.size()));
           },
-          nb::arg("local"));
-
+          nb::arg("local"), nb::arg("global"))
+      .def(
+          "global_to_local",
```
garth-wells, line 136: Do we want to make this a pure function rather than in/out?

If we keep in/out, the array sizes should be checked.

jorgensd: The array sizes are checked in debug mode in C++. https://github.com/FEniCS/dolfinx/blob/90b47de6f8fbc27213be93de7c6c3b9838f5a9c4/cpp/dolfinx/common/IndexMap.cpp#L843
making them pure can make it quite expensive calling this for set of dofs

jorgensd: So I would prefer keeping spans. If we want to change the tests to be non-debug mode it is fine with me.

garth-wells: > The array sizes are checked in debug mode in C++.
> 
> https://github.com/FEniCS/dolfinx/blob/90b47de6f8fbc27213be93de7c6c3b9838f5a9c4/cpp/dolfinx/common/IndexMap.cpp#L843
> 
> 
> making them pure can make it quite expensive calling this for set of dofs

I'm not sure it the relative cost would be high from Python.

jorgensd: It would if you call it on a cell by cell basis, as you would have to create a new array per call.
I also don't see the reason of not exposing the efficient version of this (as this is how we do it in C++), and this is a function used by expert users that want to develop or test in python.

garth-wells: > It would if you call it on a cell by cell basis, as you would have to create a new array per call. I also don't see the reason of not exposing the efficient version of this (as this is how we do it in C++), and this is a function used by expert users that want to develop or test in python.

Cell-by-cell anything is slow in Python. Do you have a use case in mind? We make it an in/out function in C++ because it's used in a small number of places where it's performance critical. If not for this, we'd have made it a pure function on the C++ size. In/out isn't idiomatic Python, so w need a good reason to deviate from idiomatic Python.

chrisrichardson (approved):

jhale (approved):

PR #2977, opened 2024-01-09
Simplify access to wrapped C++ form classes

garth-wells:
Tidies up some Python custom assembler code.


python/dolfinx/fem/forms.py
```diff
@@ -81,6 +81,30 @@ def integral_types(self):
         """Integral types in the form"""
         return self._cpp_object.integral_types
 
+    @staticmethod
```
jhale, line 84: I think I can see why you did this, however, we're not using a lot of static class methods in DOLFINx so it's another 'convention' for users and developers to learn. 

Would it be better as a free standing function?

jorgensd: I would suggest something along the lines of:
```python
    @classmethod
    def from_dtype(cls, dtype, function_spaces, integrals, coefficients, constants, mesh):
        spaces = [V._cpp_object for V in function_spaces]
        if dtype == np.float32:
            cpp_cls =  _cpp.fem.Form_float32
        elif dtype == np.float64:
            cpp_cls = _cpp.fem.Form_float64
        elif dtype == np.complex64:
            cpp_cls =  _cpp.fem.Form_complex64
        elif dtype == np.complex128:
            cpp_cls = _cpp.fem.Form_complex128
        return cls(cpp_cls(function_spaces, integrals, coefficients, constants, mesh))   
```
as it would simplify 
```
formtype = Form.cpp_class(dtype)
a = Form(formtype([V._cpp_object, V._cpp_object], integrals, [], [], False, None))
```
to
```
a = Form.from_cpp(dtype, [V, V], integrals, [], [], False, None
```
This would also allow us to properly document the input arguments with type-hints and a docstring.

garth-wells: The above is a second step that can build on the first. This PR is deliberately lower level because it's used elsewhere in the library to simply get the C++ class for a given dtype.

garth-wells: > I think I can see why you did this, however, we're not using a lot of static class methods in DOLFINx so it's another 'convention' for users and developers to learn.
> 
> Would it be better as a free standing function?

There is nothing wrong with static methods, so we don't need to avoid them.

The issue is name spacing; a free function with longer name or a static function with a short name. My reasoning was that this function would be used by a small minority of users, so I lent towards not cluttering `dolfinx.fem` with another free function.

jorgensd: > The above is a second step that can build on the first. This PR is deliberately lower level because it's used elsewhere in the library to simply get the C++ class for a given dtype.

I do not see the point of just exposing the C++ form, as all examples directly wrap this into the Python form. Could you elaborate on when you would like just the C++ class without wrapping it as a Python form later on?

garth-wells: > > The above is a second step that can build on the first. This PR is deliberately lower level because it's used elsewhere in the library to simply get the C++ class for a given dtype.
> 
> I do not see the point of just exposing the C++ form, as all examples directly wrap this into the Python form. Could you elaborate on when you would like just the C++ class without wrapping it as a Python form later on?

Because it's used elsewhere in the library, not only in tests and demos. The proposed change affects the `Form` initialiser. The initialiser could be changed but I don't see a benefit.

PR #2979, opened 2024-01-10
Generalise topology to take multiple cell types

chrisrichardson:
Adds a new `create_topology()` method that takes a list of cell_types, cells, ghost_owners and original_cell_index, instead of single values for each.
The resulting topology now has `index_maps` and `connectivity` between each type, and the types of the mesh are stored in `entity_types` (instead of `cell_type`).
To access the extra tables of connectivity and index_maps, we now need to supply an index within each dimension: e.g. `connectivity((3, 0), (2, 1))` gets the connectivity from cell_type 0 to facet_type 1 (in a 3D mesh).
Some tests are added.

Behaviour for single cell_type mesh is unaffected.


cpp/dolfinx/mesh/Topology.h
```diff
@@ -70,6 +73,16 @@ class Topology
   /// @warning This is experimental and likely to change
   void set_index_map(int dim, std::shared_ptr<const common::IndexMap> map);
 
+  /// @todo Merge with set_connectivity
+  ///
+  /// @brief Set the IndexMap for dimension dim
+  /// @warning This is experimental and likely to change
+  /// @param dim Topological dimension
+  /// @param i Index of celltype within dimension `dim`.
```
garth-wells, line 81: Docstring for `i` is not easy to make sense of.

cpp/dolfinx/mesh/Topology.h
```diff
@@ -134,8 +177,8 @@ class Topology
   /// computed.
   const std::vector<std::int32_t>& interprocess_facets() const;
 
-  /// Original cell index
-  std::vector<std::int64_t> original_cell_index;
+  /// Original cell index for each cell type
+  std::vector<std::vector<std::int64_t>> original_cell_index;
```
garth-wells, line 181: How is it indexed?

garth-wells (changes requested):

cpp/dolfinx/mesh/Topology.h
```diff
@@ -113,7 +151,12 @@ class Topology
 
   /// @brief Cell type
   /// @return Cell type that the topology is for
-  CellType cell_type() const noexcept;
+  CellType cell_type() const;
+
+  /// @brief  Get the entity types in the topology for a given dimension
+  /// @param dim Topological dimension
+  /// @return List of entity types
+  std::span<const CellType> entity_types(int dim) const;
```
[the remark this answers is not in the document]
chrisrichardson: Yes. This works better with nanobind too.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -702,28 +712,107 @@ std::vector<std::int32_t> convert_to_local_indexing(
...
+    _entity_type_offsets[i] = i;
+
+  _entity_types = {CellType::point, CellType::interval};
+  if (tdim == 2)
+    _entity_types.push_back(cell_type);
+  else if (tdim == 3)
+  {
+    _entity_types.push_back(cell_facet_type(cell_type, 0));
+    _entity_types.push_back(cell_type);
+  }
```
jorgensd, line 733: Some documentation here would be good.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -702,28 +712,107 @@ std::vector<std::int32_t> convert_to_local_indexing(
...
+    if (tdim == 3)
+    {
+      // Find all facet types
+      std::vector<mesh::CellType> facet_types;
+      for (auto c : cell_types)
+      {
+        assert(cell_dim(c) == tdim);
+        std::int8_t nf = cell_num_entities(c, tdim - 1);
+        for (std::int8_t i = 0; i < nf; ++i)
+          facet_types.push_back(cell_facet_type(c, i));
```
jorgensd, line 758:
```suggestion
        for (std::int8_t i = 0; i < cell_num_entities(c, tdim - 1); ++i)
          facet_types.push_back(cell_facet_type(c, i));
```

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -702,28 +712,107 @@ std::vector<std::int32_t> convert_to_local_indexing(
...
+      }
+      std::sort(facet_types.begin(), facet_types.end());
+      facet_types.erase(std::unique(facet_types.begin(), facet_types.end()),
+                        facet_types.end());
+      _entity_types.insert(_entity_types.end(), facet_types.begin(),
+                           facet_types.end());
```
jorgensd, line 764:
```suggestion
      std::transform(facet_types.begin(), facet_types.end(),
                                  std::back_inserter(_entity_types));
```

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -702,28 +712,107 @@ std::vector<std::int32_t> convert_to_local_indexing(
...
+  // Create square list of lists
+  _connectivity.resize(conn_size);
+  for (auto& c : _connectivity)
+    c.resize(conn_size);
 }
 //-----------------------------------------------------------------------------
-int Topology::dim() const noexcept { return _connectivity.size() - 1; }
+int Topology::dim() const noexcept { return _entity_type_offsets.size() - 2; }
 //-----------------------------------------------------------------------------
 void Topology::set_index_map(int dim,
```
jorgensd, line 786: Do we need to change the documentation of this function? as it now only sets the index map for the first entity of the given dimension?

cpp/dolfinx/mesh/Topology.h
```diff
@@ -89,11 +115,40 @@ class Topology
   std::shared_ptr<const graph::AdjacencyList<std::int32_t>>
```
jorgensd, line 115: Probably needs a note for now that this should only be used for single cell type meshes?

cpp/dolfinx/mesh/Topology.h
```diff
@@ -113,7 +168,12 @@ class Topology
 
   /// @brief Cell type
   /// @return Cell type that the topology is for
-  CellType cell_type() const noexcept;
+  CellType cell_type() const;
+
+  /// @brief  Get the entity types in the topology for a given dimension
```
jorgensd, line 173:
```suggestion
  /// @brief Get the entity types in the topology for a given dimension
```

cpp/dolfinx/mesh/Topology.h
```diff
@@ -145,13 +205,25 @@ class Topology
...
+  std::vector<std::int8_t> _entity_type_offsets;
+
+  // Parallel layout of entities for each dimension and cell type
+  // flattened in the same layout as _entity_types above.
+  std::vector<std::shared_ptr<const common::IndexMap>> _index_map;
+
+  // Connectivity between entity dimensions and cell types, arranged as
+  // a 2D array. Indexing is 0=vertex, 1=edge, followed by [facet types], [cell
+  // types]. e.g. for a 3D mesh with tets, pyramid and hex: 0=vertex, 1=edge,
+  // 2=triangle, 3=quad, 4=tet, 5=pyr, 6=hex. Connectivity between different
```
jorgensd, line 225: The text here should probably not be hard-coded, and rather refer to the CellType enum that we use for sorting, as this might change.

python/test/unit/mesh/test_mixed_topology.py
```diff
@@ -0,0 +1,101 @@
...
+                               cells, orig_index, ghost_owners, boundary_vertices)
+
+    maps = topology.index_maps(topology.dim)
+    assert len(maps) == 2
+    # Two triangles and one quad
+    assert maps[0].size_local == 2
+    assert maps[1].size_local == 1
+
+    # Six vertices in map
+    map0 = topology.index_maps(0)
```
jorgensd, line 22: Should `index_maps` rather be a property that return a set?

jorgensd: It would then match `entity_types`.

python/test/unit/mesh/test_mixed_topology.py
```diff
@@ -0,0 +1,101 @@
...
+    ghost_owners = [[], []]
+    # All vertices are on boundary
+    boundary_vertices = [3 * rank + i for i in range(6)]
+
+    topology = create_topology(MPI.COMM_WORLD, [CellType.triangle, CellType.quadrilateral],
+                               cells, orig_index, ghost_owners, boundary_vertices)
+
+    assert topology.entity_types[2][0] == CellType.triangle
+    assert topology.entity_types[2][1] == CellType.quadrilateral
+    print(topology.connectivity((2, 0), (0, 0)))
```
jorgensd, line 91: Remove debug print

python/test/unit/mesh/test_mixed_topology.py
```diff
@@ -0,0 +1,101 @@
...
+
+    topology.create_entities(2)
+
+    print(topology.connectivity(2, 0))
+    print(topology.connectivity(3, 2))
+    print(topology.entity_types)
+
+
+def test_parallel_mixed_mesh():
+    rank = MPI.COMM_WORLD.Get_rank()
```
jorgensd, line 73: What is the difference between `Get_rank` and `rank` for an MPI communicator?

cpp/dolfinx/mesh/Topology.cpp
chrisrichardson: I might just redo this with `std::set`

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -702,28 +712,107 @@ std::vector<std::int32_t> convert_to_local_indexing(
...
+      _entity_types.insert(_entity_types.end(), facet_types.begin(),
+                           facet_types.end());
+
+      _entity_type_offsets.push_back(_entity_types.size());
+    }
+  }
+
+  // Cell Types
+  _entity_types.insert(_entity_types.end(), cell_types.begin(),
+                       cell_types.end());
```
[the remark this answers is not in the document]
chrisrichardson: not very relevant, since size is tiny here.

cpp/dolfinx/mesh/Topology.cpp
chrisrichardson: Probably should throw an error if there are multiple maps.

garth-wells: Topology code should be almost exclusively `std::int32_t` or `std::int64_t` - avoid `int` unless it's obvious that it's wide enough.

chrisrichardson: This one can be `std::int8_t`

cpp/dolfinx/mesh/Topology.h
chrisrichardson: I've added a check to throw an error if misused. Trying not to touch the existing interface otherwise.

cpp/dolfinx/mesh/Topology.h
chrisrichardson: clarified documentation here

python/test/unit/mesh/test_mixed_topology.py
chrisrichardson: It could be. Or we could use `ìndex_map(dim, index)`. I guess I was keeping it the same as the single-cell-type versions.

python/test/unit/mesh/test_mixed_topology.py
chrisrichardson: don't know

python/test/unit/mesh/test_mixed_topology.py
chrisrichardson: done

python/test/unit/mesh/test_mixed_topology.py
chrisrichardson: I'll keep it for now

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -289,22 +290,32 @@ determine_sharing_ranks(MPI_Comm comm, std::span<const std::int64_t> indices)
 /// 2. With undetermined ownership
 /// 3. Not owned by the caller
 std::array<std::vector<std::int64_t>, 2>
-vertex_ownership_groups(std::span<const std::int64_t> cells_owned,
-                        std::span<const std::int64_t> cells_ghost,
+vertex_ownership_groups(std::vector<std::span<const std::int64_t>> cells_owned,
+                        std::vector<std::span<const std::int64_t>> cells_ghost,
```
jorgensd, line 294: Should this be a `const std::vector<std::span<const std::int64_T>>&`, or is copying of spans very cheap?

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -702,28 +713,110 @@ std::vector<std::int32_t> convert_to_local_indexing(
...
-  // Do nothing
+  std::int8_t tdim = cell_dim(cell_type);
+
+  // Create all the entity types in mesh, one per dimension for a single cell
+  // type mesh.
+  _entity_type_offsets.resize(tdim + 2);
+  for (std::int8_t i = 0; i < tdim + 2; ++i)
+    _entity_type_offsets[i] = i;
+
+  _entity_types = {CellType::point, CellType::interval};
```
jorgensd, line 730: A bit off topic, but this excludes the option of creating point cloud meshes, as we assume that the topological dimension of the mesh is at least 1.

cpp/dolfinx/mesh/Topology.cpp
chrisrichardson: I am assuming that copying of spans is cheap. I could make it a reference, as you suggest - no harm. Each vector is going to be pretty small.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -702,28 +713,110 @@ std::vector<std::int32_t> convert_to_local_indexing(
...
 //-----------------------------------------------------------------------------
 void Topology::set_index_map(int dim,
                              std::shared_ptr<const common::IndexMap> map)
 {
-  assert(dim < (int)_index_map.size());
-  _index_map[dim] = map;
+  assert(dim < (int)_entity_type_offsets.size() - 1);
+  // Check there is only one index map in this dimension
+  if (_entity_type_offsets[dim + 1] - _entity_type_offsets[dim] != 1)
+    throw std::runtime_error("Cannot set IndexMap on mixed topology mesh");
```
jorgensd, line 793: Maybe refer to the other `Topology::set_index_map` function below?

cpp/dolfinx/mesh/Topology.cpp
chrisrichardson: That is a good point (ha ha). I will fix it. Not sure there are many applications though.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -810,17 +903,52 @@ void Topology::create_entity_permutations()
...
+}
+//-----------------------------------------------------------------------------
+void Topology::set_connectivity(
+    std::shared_ptr<graph::AdjacencyList<std::int32_t>> c,
+    std::pair<std::int8_t, std::int8_t> d0,
+    std::pair<std::int8_t, std::int8_t> d1)
+{
+  auto [dim0, i0] = d0;
+  auto [dim1, i1] = d1;
+  assert(dim0 < (int)_entity_type_offsets.size() - 1);
```
jorgensd, line 944: Change to int8_t cast here and below.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -943,21 +1097,27 @@ mesh::create_topology(MPI_Comm comm, std::span<const std::int64_t> cells,
...
   // Get global indices of ghost cells
-  std::span cell_idx(original_cell_index);
-  const std::vector cell_ghost_indices = graph::build::compute_ghost_indices(
-      comm, cell_idx.first(num_local_cells),
-      std::span(original_cell_index).last(ghost_owners.size()), ghost_owners);
-
-  // Create an index map for cells
-  auto index_map_c = std::make_shared<common::IndexMap>(
-      comm, num_local_cells, cell_ghost_indices, ghost_owners);
+  std::vector<std::vector<std::int64_t>> cell_ghost_indices;
```
jorgensd, line 1105: Reserve size prior to loop?

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -1056,8 +1223,10 @@ mesh::create_topology(MPI_Comm comm, std::span<const std::int64_t> cells,
       });
   std::sort(global_to_local_vertices.begin(), global_to_local_vertices.end());
 
-  std::vector<std::int32_t> _cells_local_idx
-      = convert_to_local_indexing(cells, global_to_local_vertices);
+  std::vector<std::vector<std::int32_t>> _cells_local_idx;
```
jorgensd, line 1231: Same comment regarding reserveration.

cpp/dolfinx/mesh/Topology.h
```diff
@@ -78,8 +97,15 @@ class Topology
...
+  /// @param dim Topological dimension
+  /// @warning Experimental
+  /// @return List of index maps, one for each cell type
+  std::vector<std::shared_ptr<const common::IndexMap>>
+  index_maps(std::int8_t dim) const;
+
   /// @brief Return connectivity from entities of dimension d0 to
-  /// entities of dimension d1.
+  /// entities of dimension d1. Simple version for compatibility, assumes only
+  /// one celltype per dimension.
```
jorgensd, line 108:
```suggestion
  /// one cell type per dimension.
```

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -1007,11 +1167,18 @@ mesh::create_topology(MPI_Comm comm, std::span<const std::int64_t> cells,
       // communicated via ghost cells. Note that the ghost cell owner
       // (who we get the vertex index from) is not necessarily the
       // vertex owner.
-      const std::vector<std::array<std::int64_t, 3>> recv_data
-          = exchange_ghost_indexing(*index_map_c, cells, num_cell_vertices,
-                                    owned_vertices.size(), global_offset_v,
-                                    global_to_local_vertices, ghost_vertices,
-                                    ghost_vertex_owners);
+      std::vector<std::array<std::int64_t, 3>> recv_data;
```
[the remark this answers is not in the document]
chrisrichardson: not really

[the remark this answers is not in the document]
jorgensd: Ok, then we keep it as is

PR #2982, opened 2024-01-11
Implement VTK I/O for arbitrary degree hex cells

mscroggs:


cpp/dolfinx/io/cells.cpp
```diff
@@ -23,32 +23,15 @@ int cell_degree(mesh::CellType type, int num_nodes)
...
-    case 45:
-      LOG(WARNING) << "8th order mesh is untested";
-      return 8;
-    case 55:
-      LOG(WARNING) << "9th order mesh is untested";
-      return 9;
-    default:
-      throw std::runtime_error("Unknown triangle layout. Number of nodes: "
-                               + std::to_string(num_nodes));
+      throw std::runtime_error("Triangle of order " + std::to_string(num_nodes)
```
jorgensd, line 30: Not sure this is a better error message, as people might try to use "non-Lagrangian" elements for their IO (Like elements with nodes only on edges and vertices.

jorgensd (approved):
Tested locally for a 10th order hex. Extended uint8->uint16 to avoid overflow for hex order>5

Issue #2942, opened 2023-12-29
Add C++ demo that uses a custom C/C++ kernel 


garth-wells:
We have Python examples, but nothing in C++. The interface might need to be enhanced to support this.

jorgensd:
@jhale can we close this issue now?

garth-wells:
Implemented in #2983.

PR #2983, opened 2024-01-11
Custom C/C++ kernel demo

jhale:
Work towards solving https://github.com/FEniCS/dolfinx/issues/2942.

First draft for discussion. P1 Mass matrix assemble.


jorgensd:
> Blank slate to work towards solving #2942.
> 
> First initial thought is (and this applies to the Python interface too); how can I make a function space with dofs assigned to vertices without codegen?

@jhale see; https://github.com/FEniCS/dolfinx/blob/main/cpp/demo/interpolation-io/main.cpp#L34-L42

chrisrichardson (approved):

jorgensd:
@jhale 
Is there a reason that our custom kernel has to use pre-defined finite element tabulation?
I think it would be really nice to show that you can make up your own quadrature scheme (or take one from a text-book), and use basix to do runtime tabulation.

jhale:
No, it's just a placeholder. I plan to add tabulation via Basix.

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+// .. code-block:: cpp
+
+int main(int argc, char* argv[])
+{
+  dolfinx::init_logging(argc, argv);
+  MPI_Init(&argc, &argv);
+  {
+    // Create mesh and function space
+    const auto part
```
garth-wells, line 37: This isn't required - can just let DOLFINx use the default mesh partitioner.

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+        = mesh::create_cell_partitioner(mesh::GhostMode::shared_facet);
+    auto mesh = std::make_shared<mesh::Mesh<U>>(
+        mesh::create_rectangle<U>(MPI_COMM_WORLD, {{{0.0, 0.0}, {1.0, 1.0}}},
+                                  {10, 10}, mesh::CellType::triangle, part));
+
+    // Create basix element for the field u. This will be used to construct
+    // basis functions inside the custom cell kernel.
+    const int order = 1;
```
garth-wells, line 45: Use `constexpr`.

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+    const basix::FiniteElement e = basix::create_element<T>(
```
garth-wells, line 46: No need for `const` - just extra syntactic clutter. (more below that can be removed)

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+        basix::element::family::P,
+        mesh::cell_type_to_basix_type(mesh::CellType::triangle), order,
+        basix::element::lagrange_variant::unset,
+        basix::element::dpc_variant::unset, false);
```
garth-wells, line 50: Maybe Basix should use an enum for the last arg - booleans are hard to understand.

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    const int max_degree = 2 * order;
```
garth-wells, line 52: `constexpr`

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+    const auto quadrature_type = basix::quadrature::get_default_rule(
+        basix::cell::type::triangle, max_degree);
+    const auto [points, weights] = basix::quadrature::make_quadrature<T>(
```
garth-wells, line 55: Template type is wrong - should come from real part of the scalar field.

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include <dolfinx.h>
+#include <dolfinx/la/MatrixCSR.h>
+#include <dolfinx/la/SparsityPattern.h>
+#include <utility>
+#include <vector>
+
+using namespace dolfinx;
+
+using T = double; // field scalar type
+using U = double; // geometry scalar type
```
garth-wells, line 23: This should probably be derived from `T` as the real part of `T`.

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
+// Custom cell kernel (C++)
+//
+// This demo shows how to define a custom cell kernel in C++ and have it
+// assembled into a dolfinx::la::MatrixCSR.
+//
+// .. code-block:: cpp
+
+#include <iostream>
```
garth-wells, line 8: Remove.

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+    const auto length = std::accumulate(
+        tabulate_shape.begin(), tabulate_shape.end(), 1, std::multiplies<>{});
+    std::vector<T> basis(length);
+    mdspan_t<T, 4> basis_span(basis.data(), tabulate_shape);
+
+    // Tabulate basis functions
+    e.tabulate(0, points_span, basis_span);
+
+    // Calculate mass matrix on reference cell
+    const auto e_dim = e.dim();
```
garth-wells, line 82: Use type here rather than `auto` for readability.

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+        for (std::size_t j = 0; j < A_hat_span.extent(1); ++j)
+        {
+          A_hat_span(i, j)
+              += weights[k] * basis_span(0, k, i, 0) * basis_span(0, k, j, 0);
+        }
+      }
+    }
+
+    // Define finite element mass kernel.
+    std::function<void(T*, const T*, const T*, const U*, const int*,
```
garth-wells, line 100: Use `auto` to make it an anonymous function, which will be inlined by the assembler.

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+                       const u_int8_t*)>
+        mass_cell_kernel
+        = [&A_hat_span](T* A_cell, const T*, const T*, const U* cdofs,
```
garth-wells, line 103: Not need for the reference on `A_hat_span` since it's effectively just a pointer.

(I would have expected the function to need `mutable` arg).

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+                        const int*, const u_int8_t*)
+    {
+      T detJ = abs((cdofs[0] - cdofs[3]) * (cdofs[7] - cdofs[4])
+                   - (cdofs[1] - cdofs[4]) * (cdofs[6] - cdofs[3]));
+      for (std::size_t i = 0; i < A_hat_span.extent(0); ++i)
+      {
+        for (std::size_t j = 0; j < A_hat_span.extent(1); ++j)
+        {
+          A_cell[i * A_hat_span.extent(0) + j] = detJ * A_hat_span(i, j);
```
garth-wells, line 112: Maybe lift ` A_hat_span.extent(0)` out of the loop.

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+        }
+      }
+    };
+    const std::map integrals{
```
garth-wells, line 116: This bit is ugly - can/should we make it simpler in DOLFINx?

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+        std::pair{fem::IntegralType::cell,
+                  std::vector{std::tuple{-1, mass_cell_kernel, cells}}}};
+
+    // Define form from integral
+    const auto a = std::make_shared<fem::Form<T>>(
+        fem::Form<T>({V, V}, integrals, {}, {}, false, mesh));
```
garth-wells, line 122: Not clear what `false` is - should we make it an enum?

garth-wells (changes requested):
Should we assembler for an FFCx version too to compare the result? This would also be super helpful for profiling the kernel call cost (since the pure C++ version can inline to call).

cpp/demo/custom_cell_kernel/main.cpp
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    // Define finite element mass kernel.
+    std::function<void(T*, const T*, const T*, const U*, const int*,
+                       const u_int8_t*)>
+        mass_cell_kernel
+        = [&A_hat_span](T* A_cell, const T*, const T*, const U* cdofs,
+                        const int*, const u_int8_t*)
+    {
+      T detJ = abs((cdofs[0] - cdofs[3]) * (cdofs[7] - cdofs[4])
```
chrisrichardson, line 106:
```suggestion
      U detJ = std::abs((cdofs[0] - cdofs[3]) * (cdofs[7] - cdofs[4])
```

cpp/demo/custom_cell_kernel/main.cpp
jhale: The Form constructor only accepts mass_cell_kernel if it's a std::function.

```
/usr/local/include/dolfinx/fem/Form.h:94:57: note:   no known conversion for argument 2 from 'const std::map<dolfinx::fem::IntegralType, std::vector<std::tuple<int, main(int, char**)::<lambda(T*, const T*, const T*, const U*, const int*, const u_int8_t*)>, std::vector<int, std::allocator<int> > >, std::allocator<std::tuple<int, main(int, char**)::<lambda(T*, const T*, const T*, const U*, const int*, const u_int8_t*)>, std::vector<int, std::allocator<int> > > > >, std::less<dolfinx::fem::IntegralType>, std::allocator<std::pair<const dolfinx::fem::IntegralType, std::vector<std::tuple<int, main(int, char**)::<lambda(T*, const T*, const T*, const U*, const int*, const u_int8_t*)>, std::vector<int, std::allocator<int> > >, std::allocator<std::tuple<int, main(int, char**)::<lambda(T*, const T*, const T*, const U*, const int*, const u_int8_t*)>, std::vector<int, std::allocator<int> > > > > > > >' to 'const std::map<dolfinx::fem::IntegralType, std::vector<std::tuple<int, std::function<void(double*, const double*, const double*, const double*, const int*, const unsigned char*)>, std::vector<int, std::allocator<int> > >, std::allocator<std::tuple<int, std::function<void(double*, const double*, const double*, const double*, const int*, const unsigned char*)>, std::vector<int, std::allocator<int> > > > >, std::less<dolfinx::fem::IntegralType>, std::allocator<std::pair<const dolfinx::fem::IntegralType, std::vector<std::tuple<int, std::function<void(double*, const double*, const double*, const double*, const int*, const unsigned char*)>, std::vector<int, std::allocator<int> > >, std::allocator<std::tuple<int, std::function<void(double*, const double*, const double*, const double*, const int*, const unsigned char*)>, std::vector<int, std::allocator<int> > > > > > > >&'
```

cpp/demo/custom_cell_kernel/main.cpp
jhale: It's how it is in Python, except of course C++ only has `{}` for initialising.

cpp/demo/custom_cell_kernel/main.cpp
jhale: @mscroggs What do you think?

jhale:
> Should we assembler for an FFCx version too to compare the result? This would also be super helpful for profiling the kernel call cost (since the pure C++ version can inline to call).

Let's discuss this on Slack in #development - I don't think this is currently possible as std::function is everywhere. If we can get that issue solved, yes, I'd be happy to produce a counterpart for benchmarking/comparison.

cpp/demo/custom_cell_kernel/main.cpp
garth-wells: It was designed to work, but untested. I'll look into it as soon as I can.

garth-wells: The issue is that the `Form` constructor requires a vector that holds a `std::function`. We could probably generalise it.

I've added to the demo a version that calls `assemble_cells` directly and passes an anonymous function for the element kernel. With this very light element kernel it's about 30% faster. Difference will be even greater for a RHS vector.

garth-wells (approved):

PR #2985, opened 2024-01-14
 Implement VTK I/O for arbitrary degree tetrahedron cells

mscroggs:
Add code for generating permutations from DOLFINx dof formatting to the recursive VTK tetrahedron dof ordering for higher order Lagrange cells.


jorgensd (approved):
Tested with Paraview from order 1-10 and order 12

IgorBaratta:
Don't we need some documentation or at least references to the formulas/drawings that you used?

PR #2989, opened 2024-01-15
Add transformation_type enum

mscroggs:
Hopefully makes code more readable than having multiple bools being passed in


cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -21,6 +21,15 @@ struct ufcx_finite_element;
 
 namespace dolfinx::fem
 {
+/// DOF transformation type
+enum class transformation_type
```
garth-wells, line 25: `fem::transformation_type` is a bit generic and could mean quite a few things.

Maybe put it inside `FiniteElement` and call it something like `doftransform`? Could leave it in the `fem` namespace, in which case the name needs to be more specific. The former is probably better since it's used only in `FiniteElement`.

garth-wells: @mscroggs ping

garth-wells (approved):

PR #2993, opened 2024-01-17
Stop cffi_utils failing on KeyError

mscroggs:
If PETSc's int or scalar type is not in `_CTYPES`, dolfinx currently throws a `KeyError` on import. This will prevent that, and errors will only occur if `cffi_utils` is used.


garth-wells (commented):
How are CI tests managing to pass without this change?

mscroggs:
> How are CI tests managing to pass without this change?

the images on CI have types that are in `_CTYPES` so it's fine. Somehow my local Docker image doesn't though

garth-wells:
> > How are CI tests managing to pass without this change?
> 
> the images on CI have types that are in `_CTYPES` so it's fine. Somehow my local Docker image doesn't though



> > How are CI tests managing to pass without this change?
> 
> the images on CI have types that are in `_CTYPES` so it's fine. Somehow my local Docker image doesn't though

I can’t see what other NumPy types petsc4py supports. Can you check what the petsc4py integer and scalar types are in your image?

garth-wells:
> > I can’t see what other NumPy types petsc4py supports. Can you check what the petsc4py integer and scalar types are in your image?
> 
> IntType is `np.longlong`, ScalarType is `np.float64`

Maybe just add `np.longlong` to the dictionary?

mscroggs:
> > > I can’t see what other NumPy types petsc4py supports. Can you check what the petsc4py integer and scalar types are in your image?
> > 
> > 
> > IntType is `np.longlong`, ScalarType is `np.float64`
> 
> Maybe just add `np.longlong` to the dictionary?

Done, but I think it's worth merging the rest of this in too so that if someone at any point uses petsc with a type we're yet to include, they only get errors when using `cffi_utils` and not when importing dolfinx

garth-wells (approved):

PR #2994, opened 2024-01-17
Topology computation for mixed topology

chrisrichardson:
Build the correct entity-entity connectivity maps for mixed topology.

Now that facets and cells can have multiple types, we need to adjust the computation algorithm.
All facets of the same type need to be computed at the same time, so all cells which contain them need to be worked on together.

This updates `compute_entities()` and `compute_connectivity()` to account for this.

Including some tests on a simple topology (2 tet + 1 prism + 1 hex)


python/test/unit/mesh/test_mixed_topology.py
```diff
@@ -86,3 +147,41 @@ def test_parallel_mixed_mesh():
...
     assert topology.index_maps(2)[0].size_global == size * 2
     assert topology.index_maps(2)[1].size_global == size
+
+    print(topology.index_maps(2)[0].size_global)
+
+
+def test_create_entities():
+    mesh = create_unit_cube(MPI.COMM_WORLD, 2, 2, 2, CellType.prism, ghost_mode=GhostMode.none)
+
+    print('ncells = ', mesh.topology.index_map(3).size_local)
```
chrisrichardson, line 157:
```suggestion
```

python/test/unit/mesh/test_mixed_topology.py
```diff
@@ -86,3 +147,40 @@ def test_parallel_mixed_mesh():
...
     assert topology.index_maps(2)[0].size_global == size * 2
     assert topology.index_maps(2)[1].size_global == size
+
+    print(topology.index_maps(2)[0].size_global)
+
+
+def test_create_entities():
+    mesh = create_unit_cube(MPI.COMM_WORLD, 2, 2, 2, CellType.prism, ghost_mode=GhostMode.none)
+
+
```
chrisrichardson, line 157:
```suggestion
```

python/test/unit/mesh/test_mixed_topology.py
```diff
@@ -50,18 +51,77 @@ def test_mixed_mesh_3d():
...
-
-def test_prism():
-    cells = [[0, 1, 2, 3, 4, 5]]
-    orig_index = [[0]]
-    ghost_owners = [[]]
-    boundary_vertices = []
-
-    topology = create_topology(MPI.COMM_SELF, [CellType.prism], cells, orig_index, ghost_owners, boundary_vertices)
-    assert len(topology.entity_types[2]) == 2
+    assert CellType.triangle == entity_types[2][1]
```
jpdean, line 54: What determines the order of `entity_types[d]`? Is it always quads then triangles? Or will this fail if the implementation changes? I couldn't find anything about the order in the docs

python/test/unit/mesh/test_mixed_topology.py
```diff
@@ -50,18 +51,77 @@ def test_mixed_mesh_3d():
...
+    assert CellType.quadrilateral == entity_types[2][0]
+
+    # Create triangle and quadrilateral facets
+    topology.create_entities(2)
+
+    # Tet -> quad
+    assert topology.connectivity((3, 0), (2, 0)) is None
```
jpdean, line 61: At the moment, it looks like to get the connectivity between particular entity types, the user needs to first run the code to determine what cell type corresponds to what index in `entity_types[d]`. Would it be better if the pair had a `CellType` rather than an entity type index?

chrisrichardson: That's true. I am not sure if it is better, but we could try it out. Maybe in another branch?

python/test/unit/mesh/test_mixed_topology.py
chrisrichardson: At the moment, the facet order is determined in the constructor of `Topology`, which effectively does a sort, so quad(-4) appears before triangle(3). This could be changed, and the test would then fail, yes. 
The cell order, however, is arbitrary, and depends on the user input. 
So, what is the best solution? Maybe I should change the test, so it doesn't assume anything about the order?

jpdean: I think it would probably be better to avoid making the assumption in the test

cpp/dolfinx/mesh/topologycomputation.cpp
```diff
@@ -478,92 +460,120 @@ get_local_indexing(MPI_Comm comm, const common::IndexMap& cell_map,
...
-  const std::size_t entity_list_shape0 = num_cells * num_entities_per_cell;
-  const std::size_t entity_list_shape1 = max_vertices_per_entity;
-  std::vector<std::int32_t> entity_list(entity_list_shape0 * entity_list_shape1,
-                                        -1);
-  for (std::size_t c = 0; c < num_cells; ++c)
+  for (std::size_t k = 0; k < cell_lists.size(); ++k)
   {
-    // Get vertices from cell
-    auto vertices = cells.links(c);
+    auto cell_type = std::get<0>(cell_lists[k]);
```
jpdean, line 510: I quite like structured binding in these cases i.e. `auto [cell_type, cells, cell_index_map] = cell_lists[k];`

cpp/dolfinx/mesh/topologycomputation.h
```diff
@@ -34,27 +34,30 @@ class Topology;
 /// @param[in] comm MPI Communicator
 /// @param[in] topology Mesh topology
 /// @param[in] dim The dimension of the entities to create
+/// @param[in] index Index of entity in dimension `dim`
```
jpdean, line 37: Maybe refer to `entity_types` for clarity?

jpdean (approved):
Looks good to me!

PR #2996, opened 2024-01-18
 Remove gdim input to UFL elements

mscroggs:
See https://github.com/FEniCS/ufl/pull/249

### DIFF
--- a/cpp/dolfinx/fem/FiniteElement.cpp
+++ b/cpp/dolfinx/fem/FiniteElement.cpp
@@ -123,7 +123,8 @@ _extract_sub_element(const FiniteElement<T>& finite_element,
 template <std::floating_point T>
 FiniteElement<T>::FiniteElement(const ufcx_finite_element& e)
     : _signature(e.signature), _space_dim(e.space_dimension),
-      _value_shape(e.value_shape, e.value_shape + e.value_rank),
+      _value_shape(e.reference_value_shape,
+                   e.reference_value_shape + e.reference_value_rank),
       _bs(e.block_size), _is_mixed(e.element_type == ufcx_mixed_element)
 {
   const ufcx_shape _shape = e.cell_shape;
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -340,7 +340,7 @@ class Function(ufl.Coefficient):
 
         # Allocate memory for return value if not provided
         if u is None:
-            value_size = ufl.product(self.ufl_element().value_shape)
+            value_size = ufl.product(self.ufl_element().value_shape(self._V.mesh.domain))
             u = np.empty((num_points, value_size), self.dtype)
 
         self._cpp_object.eval(_x, _cells, u)  # type: ignore
@@ -517,7 +517,7 @@ def functionspace(mesh: Mesh,
     try:
         e = ElementMetaData(*element)
         ufl_e = basix.ufl.element(e.family, mesh.basix_cell(), e.degree, shape=e.shape,
-                                  symmetry=e.symmetry, gdim=mesh.ufl_cell().geometric_dimension())
+                                  symmetry=e.symmetry)
     except TypeError:
         ufl_e = element  # type: ignore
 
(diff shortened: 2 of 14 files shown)


garth-wells (approved):

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -691,7 +693,11 @@ class FiniteElement
       _sub_elements;
 
   // Dimension of each value space
-  std::vector<std::size_t> _value_shape;
+  std::vector<std::size_t> _reference_value_shape;
+  std::vector<std::size_t> _value_shape_0;
+  std::vector<std::size_t> _value_shape_1;
+  std::vector<std::size_t> _value_shape_2;
+  std::vector<std::size_t> _value_shape_3;
```
chrisrichardson, line 700:
```suggestion
  std::array<std::vector<std::size_t>, 4> _value_shape;
```
I think this looks a lot better as an array (or vector) of vectors.

cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -123,7 +123,12 @@ _extract_sub_element(const FiniteElement<T>& finite_element,
 template <std::floating_point T>
 FiniteElement<T>::FiniteElement(const ufcx_finite_element& e)
     : _signature(e.signature), _space_dim(e.space_dimension),
-      _value_shape(e.value_shape, e.value_shape + e.value_rank),
+      _reference_value_shape(e.reference_value_shape,
+                             e.reference_value_shape + e.reference_value_rank),
+      _value_shape_0(e.value_shape[0], e.value_shape[0] + e.value_rank),
+      _value_shape_1(e.value_shape[1], e.value_shape[1] + e.value_rank),
+      _value_shape_2(e.value_shape[2], e.value_shape[2] + e.value_rank),
+      _value_shape_3(e.value_shape[3], e.value_shape[3] + e.value_rank),
```
chrisrichardson, line 131: I don't see why this is needed? Seems like redundant data.

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -117,7 +118,8 @@ class FiniteElement
 
   /// Shape of the value space. The rank is the size of the
   /// `value_shape`.
-  std::span<const std::size_t> value_shape() const noexcept;
+  /// @param[in] gdim The geometric dimension of the physical cell
+  std::span<const std::size_t> value_shape(int gdim) const noexcept;
```
chrisrichardson, line 122: What does it mean if `gdim` is less than the topological dimension of the cell?

jorgensd: It should not be possible

cpp/dolfinx/fem/FiniteElement.cpp
mscroggs: I've moved value_shape to function space so these are no longer needed

cpp/dolfinx/fem/FiniteElement.h
mscroggs: resolving as I've removed these variables now

cpp/dolfinx/fem/FiniteElement.h
mscroggs: I've removed this function so this is no longer a concern

garth-wells:
Looking at the direction of travel in this PR, could we remove block size from `FiniteElement` and move it to the `FunctionSpace`?

mscroggs:
> Looking at the direction of travel in this PR, could we remove block size from `FiniteElement` and move it to the `FunctionSpace`?

I think that would make sense

garth-wells:
> > Looking at the direction of travel in this PR, could we remove block size from `FiniteElement` and move it to the `FunctionSpace`?
> 
> I think that would make sense

I guess mixed elements are problematic. Maybe info on mixed elements should be held by the `FunctionSpace` too?

mscroggs:
> > > Looking at the direction of travel in this PR, could we remove block size from `FiniteElement` and move it to the `FunctionSpace`?
> > 
> > 
> > I think that would make sense
> 
> I guess mixed elements are problematic. Maybe info on mixed elements should be held by the `FunctionSpace` too?

Yes, was just thinking about this. I think we consider this PR now, then work on this afterwards. I think there a fair amount of refactoring in FFCx and possibly UFL to make this change.

PR #2998, opened 2024-01-18
Lighter switch to ruff

jhale:
Ruff linting without consistent formatting (e.g. black).


python/dolfinx/fem/assemble.py
```diff
@@ -213,7 +213,7 @@ def _assemble_vector_array(b: np.ndarray, L: Form, constants=None, coeffs=None):
 
 
 @functools.singledispatch
-def assemble_matrix(a: typing.Any, bcs: typing.Optional[typing.List[DirichletBC]] = None,
+def assemble_matrix(a: typing.Any, bcs: typing.Optional[list[DirichletBC]] = None,
```
francesco-ballarin, line 216: This change, and all similar ones, will break python `< 3.9`. I think that `ruff` uses by default whatever version of python is available in the docker image (`3.10`, if I remember correctly), but there is a configuration option to prevent it from doing that. However, I don't remember which of the rules is enforcing this change (`pyupgrade` maybe?)  to check the actual name of this option.

python/dolfinx/io/utils.py
```diff
@@ -18,8 +18,8 @@
 import basix.ufl
 import ufl
 from dolfinx import cpp as _cpp
-from dolfinx.cpp.io import perm_gmsh as cell_perm_gmsh  # noqa F401
-from dolfinx.cpp.io import perm_vtk as cell_perm_vtk  # noqa F401
+from dolfinx.cpp.io import perm_gmsh as cell_perm_gmsh  # F401
```
francesco-ballarin, line 21: Removing the `noqa` is odd: I guess it should have been either `noqa: F401`, or no comment at all

python/dolfinx/fem/assemble.py
jhale: We should discuss deprecating 3.8, only has ~6 months of support left.

python/pyproject.toml
```diff
@@ -57,9 +56,6 @@ wheel.license-files = ["../COPYING*"]
 [tool.pytest]
 junit_family = "xunit2"
 
-[tool.isort]
```
francesco-ballarin, line 60: If we can't use `isort` within `ruff`, I guess this should be added back.

python/test/unit/io/test_vtk.py
```diff
@@ -17,7 +17,7 @@
 from dolfinx import default_real_type
 from dolfinx.fem import Function, functionspace
 from dolfinx.io import VTKFile
-from dolfinx.io.utils import cell_perm_vtk  # noqa F401
```
francesco-ballarin, line 20: noqa here too.

python/dolfinx/fem/assemble.py
jhale: I checked this. The target version is inferred from pyproject.toml.

`ruff --show-settings | grep  target_version`

gives

```target_version: Py38,
target_version: Py38,```

File: python/demo/demo_cahn-hilliard.py [hunk 5]
```diff
@@ -194,7 +194,8 @@
 u.x.array[:] = 0.0
 
 # Interpolate initial condition
-u.sub(0).interpolate(lambda x: 0.63 + 0.02 * (0.5 - np.random.rand(x.shape[1])))
+rng = np.random.default_rng()
```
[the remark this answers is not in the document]
francesco-ballarin: I guess there is value in this comment, but how come it did not notice it before? The call to `np.random.rand` was already there.

chrisrichardson (approved):

python/demo/demo_cahn-hilliard.py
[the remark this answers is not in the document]
jorgensd: I think it is a new feature of Github, Lets add a seed?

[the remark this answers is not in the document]
jhale: I think GitHub can pickup on new issues picked up by SonarCloud and now put them back in to the review, which seems very useful.

PR #3001, opened 2024-01-19
Create 'solver' fixture for tests, and update `test_complex_assembler.py` to use it.

chrisrichardson:
Removes the need for PETSc in complex assembler test.


python/test/conftest.py
```diff
@@ -136,3 +139,64 @@ def tempdir(request):
...
+        rnorm = rnorm_new
+        if (rnorm / rnorm0 < rtol2):
+            return
+        p.array[:nr] = beta * p.array[:nr] + r
+
+    raise RuntimeError(f"Solver exceeded max iterations ({kmax}).")
+
+
+@pytest.fixture(scope="function")
+def solver():
```
garth-wells, line 189: Name is very generic - it will only work for SPD problems.

garth-wells, line 189: Name is very generic - parallel case only work for SPD problems.

chrisrichardson: Maybe I should always use CG, even in serial? At least it is consistent then... and I can change the name to `cg_solver`?

garth-wells: It's probably better to use the same solver in serial and parallel.

I suggest adding [BiCGSTAB](https://en.wikipedia.org/wiki/Biconjugate_gradient_stabilized_method) for non-symmetric and indefinite problems.

chrisrichardson: OK, I will change this to `cg_solver()` - maybe we can save BiCGStab for another PR.

python/test/conftest.py
```diff
@@ -136,3 +138,56 @@ def tempdir(request):
...
+       which can work in serial or parallel for testing use.
+       Not suitable for large problems."""
+
+    # Basic Conjugate Gradient solver
+    def _cg(comm, A, b, x):
+        kmax = 500
+        rtol2 = 10 * np.finfo(x.array.dtype).eps
+
+        def _global_dot(comm, v0, v1):
+            return comm.allreduce(np.dot(v0, v1), MPI.SUM)
```
garth-wells, line 155: Doesn't look like the correct inner product for the complex-valued case.

python/test/conftest.py
```diff
@@ -136,3 +138,56 @@ def tempdir(request):
...
+
+@pytest.fixture(scope="function")
+def cg_solver():
+    """Simple Conjugate Gradient solver for SPD problems,
+       which can work in serial or parallel for testing use.
+       Not suitable for large problems."""
+
+    # Basic Conjugate Gradient solver
+    def _cg(comm, A, b, x):
+        kmax = 500
```
garth-wells, line 151: `maxit` is more intuitive, and should probably be a parameter with a default value.

python/test/conftest.py
```diff
@@ -136,3 +138,56 @@ def tempdir(request):
...
+        rtol2 = 10 * np.finfo(x.array.dtype).eps
```
garth-wells, line 152: This is very tight. 

Should probably be a parameter with a default value.

python/test/unit/fem/test_complex_assembler.py
```diff
@@ -6,111 +6,109 @@
...
-    b.ghostUpdate(addv=PETSc.InsertMode.ADD, mode=PETSc.ScatterMode.REVERSE)
+    b.scatter_reverse(la.InsertMode.add)
 
-    # Create solver
-    solver = PETSc.KSP().create(mesh.comm)
-    x = A.createVecRight()
-    solver.setOperators(A)
-    solver.solve(b, x)
+    u = Function(V, dtype=complex_dtype)
+    cg_solver(MPI.COMM_WORLD, A, b, u.x)
```
garth-wells, line 105: comm should com from `V` or the mesh.

garth-wells (changes requested):

python/test/conftest.py
chrisrichardson: replacing with `vdot`

python/test/conftest.py
chrisrichardson: Note this is rtol^2, so about 1e-8 for double. I will make it into a parameter

mscroggs (approved):

jorgensd (approved):

PR #3002, opened 2024-01-20
Make isort check blocking

garth-wells:
Now that we set `isort` configuration options in `pyproject.toml` I think we can be stricter on import sorting.


francesco-ballarin:
I think @jhale is already planning to make it blocking in #2998. If we plan to merge #2998 somewhat soon, maybe this change could be deferred there, to avoid having to sort through merge conflicts between the two PRs.

francesco-ballarin:
#2998 has been added to the merge queue, and will also make isort blocking. Closing.

PR #3003, opened 2024-01-20
Re-name overloaded `create_form` function

garth-wells:
We have three `create_form` overloaded functions. Two take form coefficients and constants that are associated with a name (string). This allows coefficients to be set by the name used in a UFL file. One `create_form` takes coefficients and constants that are in the required order.

The problem with this overloading is when a form has no coefficients or constants and one pass empty initializer lists (`{}`), which is the simplest syntax, the compiler cannot resolve the overload. 

This change re-names the `create_form` function that requires data in the correct order to `create_form_factory`. There are no changes top C++ demos or tests, and no changes to Python.


chrisrichardson (approved):
This has long been a problem. Good to fix.

PR #3005, opened 2024-01-21
Update all notebooks to render correctly on web page and in notebooks.

jorgensd:
Reported at: https://fenicsproject.discourse.group/t/current-commet-may-lead-error-in-jupyter-notebook/13347?u=dokken

Especially multi-line mathematics with `&` for alignment needs to be wrapped with `\begin{align} .... \end{align}`

### DIFF
--- a/python/demo/demo_tnt-elements.py
+++ b/python/demo/demo_tnt-elements.py
@@ -67,11 +67,9 @@ wcoeffs = np.eye(8, 9)
 # $f_i\in\operatorname{span}\{q_1, q_2,\dots\}$ (as $\{f_0, f_1,\dots\}$ is a
 # basis of the polynomial space for our element). Using the properties of
 # orthonormal polynomials, we see that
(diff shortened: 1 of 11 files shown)


python/demo/demo_tnt-elements.py
```diff
@@ -67,11 +67,9 @@
 # $f_i\in\operatorname{span}\{q_1, q_2,\dots\}$ (as $\{f_0, f_1,\dots\}$ is a
 # basis of the polynomial space for our element). Using the properties of
 # orthonormal polynomials, we see that
-# \[f_i = \sum_j\left(\int_R f_iq_j\,\mathrm{d}\mathbf{x}\right)q_j,\]
```
francesco-ballarin, line 70: Not a big deal, but maybe this one and the next one should be `$$` if you still want them to show up in display style

francesco-ballarin (approved):

python/demo/demo_tnt-elements.py
jorgensd: I think the current wrapping is fine:)

PR #3007, opened 2024-01-21
Small tweaks to custom cell kernel demo

jhale:
- Fix typo and improve readability of timings.
- Generalise top description.

I noticed that explicit geometry typing was removed - I guess this was intentional to reduce complexity?

### DIFF
--- a/cpp/demo/custom_kernel/main.cpp
+++ b/cpp/demo/custom_kernel/main.cpp
@@ -77,10 +77,10 @@ double assemble_matrix0(std::shared_ptr<fem::FunctionSpace<T>> V, auto kernel,
                         std::span<const std::int32_t> cells)
 {
   // Kernel data (ID, kernel function, cell indices to execute over)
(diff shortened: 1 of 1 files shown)


garth-wells:
> I noticed that explicit geometry typing was removed - I guess this was intentional to reduce complexity?

The kernel wasn’t written for a complex-valued mass matrix (no conjugates), so having different scalar and geometry types didn’t add anything except extra complexity.

garth-wells (approved):

jhale:
> > I noticed that explicit geometry typing was removed - I guess this was intentional to reduce complexity?
> 
> The kernel wasn’t written for a complex-valued mass matrix (no conjugates), so having different scalar and geometry types didn’t add anything except extra complexity.

I agree with your argument.

My thinking was if someone wanted to use this as inspiration for a complex-valued kernel, it would be easier to copy and paste a skeleton if it had the explicit geometry type. In any case, I think it's relatively straightforward for a C++ user to work it out.

PR #3009, opened 2024-01-22
Fix interior facet permutations

anzil:
This PR is joint work with @michalhabera and supersedes #3000.

Non-symmetric assembled operators were observed for bilinear forms that combine e.g. Nedelec and Regge elements when interior facet integrals were present in the form.

This PR fixes post dof transformation during sub-block matrix/vector assembly for interior facet integrals.


jhale:
Thanks for finding and fixing this subtle bug! There is a patch coming in with the aim of removing PETSc solvers from the tests; can you wait for that to come in and then update?

jhale (changes requested):
Update tests to not use PETSc solves.

PR #3016, opened 2024-01-25
Add maps for mixed domain assembly

jpdean:
This PR adds the maps that allow mixed domain forms to be assembled. The maps take entities in the integration domain mesh (the mesh associated with the integration measure) to the corresponding entities in the meshes of other domains in the problem.

In this PR, I've only implemented mixed domain assembly for cell integrals without coefficients to keep it more manageable. If everyone is happy with this approach, I will implement other integral type, coefficient packing, and add a much more thorough set of tests. This will mean that `entity_maps` will appear in many places where the dofmap is used to access cell dofs (e.g. `assemble_scalar`, `assemble_vector`, `apply_lifting` and `pack_coefficients` etc.). If this is problematic, we can discuss some alternative ideas.

It would be great to hear everyones thoughts on this approach.

### DIFF
--- a/cpp/dolfinx/fem/assemble_matrix_impl.h
+++ b/cpp/dolfinx/fem/assemble_matrix_impl.h
@@ -68,6 +68,9 @@ void assemble_cells(la::MatSet<T> auto mat_set, mdspan2_t x_dofmap,
     // TODO Add assert?
     std::int32_t c_0 = cell_map0(c);
     std::int32_t c_1 = cell_map1(c);
+    // TODO Also check upper limit?
+    assert(c_0 >= 0);
+    assert(c_1 >= 0);
 
     // Get cell coordinates/geometry
     auto x_dofs
--- a/cpp/dolfinx/fem/assemble_matrix_impl.h
+++ b/cpp/dolfinx/fem/assemble_matrix_impl.h
@@ -41,7 +41,8 @@ void assemble_cells(la::MatSet<T> auto mat_set, mdspan2_t x_dofmap,
                     std::span<const std::int8_t> bc1, FEkernel<T> auto kernel,
                     std::span<const T> coeffs, int cstride,
                     std::span<const T> constants,
-                    std::span<const std::uint32_t> cell_info,
+                    std::span<const std::uint32_t> cell_info0,
+                    std::span<const std::uint32_t> cell_info1,
                     const std::function<std::int32_t(std::int32_t)>& cell_map0,
                     const std::function<std::int32_t(std::int32_t)>& cell_map1)
 {
@@ -83,8 +84,8 @@ void assemble_cells(la::MatSet<T> auto mat_set, mdspan2_t x_dofmap,
     kernel(Ae.data(), coeffs.data() + index * cstride, constants.data(),
            coordinate_dofs.data(), nullptr, nullptr);
 
-    pre_dof_transform(_Ae, cell_info, c, ndim1);
-    post_dof_transform(_Ae, cell_info, c, ndim0);
+    pre_dof_transform(_Ae, cell_info1, c_1, ndim1);
+    post_dof_transform(_Ae, cell_info0, c_0, ndim0);
 
     // Zero rows/columns for essential bcs
     auto dofs0 = std::span(dofmap0.data_handle() + c_0 * num_dofs0, num_dofs0);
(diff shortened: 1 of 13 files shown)


cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -41,7 +41,10 @@ void assemble_cells(la::MatSet<T> auto mat_set, mdspan2_t x_dofmap,
                     std::span<const std::int8_t> bc1, FEkernel<T> auto kernel,
                     std::span<const T> coeffs, int cstride,
                     std::span<const T> constants,
-                    std::span<const std::uint32_t> cell_info)
+                    std::span<const std::uint32_t> cell_info0,
+                    std::span<const std::uint32_t> cell_info1,
+                    const std::function<std::int32_t(std::int32_t)>& cell_map0,
+                    const std::function<std::int32_t(std::int32_t)>& cell_map1)
```
jorgensd, line 47: Wouldn't the interface be simpler if we pass in `std::span<const std::int32_t>` instead? We already store the span inside entity maps in `entity_maps`.
I get that this means that we increase the memory of normal integration kernels with 1xnum_cells (as we could pass in the same span in both test and trial space). However, that is a relatively low cost.

python/dolfinx/fem/forms.py
```diff
@@ -119,14 +119,16 @@ def form_cpp_class(dtype: npt.DTypeLike) -> typing.Union[_cpp.fem.Form_float32,
...
+         jit_options: typing.Optional[dict] = None,
+         entity_maps: dict[_cpp.mesh.Mesh, np.typing.NDArray[np.int32]] = {}):
     """Create a Form or an array of Forms.
 
     Args:
         form: A UFL form or list(s) of UFL forms.
         dtype: Scalar type to use for the compiled form.
         form_compiler_options: See :func:`ffcx_jit <dolfinx.jit.ffcx_jit>`
         jit_options: See :func:`ffcx_jit <dolfinx.jit.ffcx_jit>`.
+        entity_maps: The entity maps required to assemble the form.
```
jorgensd, line 131: I think we have to be a bit more expressive here:)

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -100,3 +99,69 @@ def test_submesh_facet_assembly(n, k, space, ghost_mode):
...
+    assert np.isclose(A_0.norm(), A.norm())
+
+    # Now assemble a mixed domain form using msh as integration domain
+    # Entity maps must take cells in msh (the integration domain mesh) to
+    # cells in smsh
+    cell_imap = msh.topology.index_map(tdim)
+    num_cells = cell_imap.size_local + cell_imap.num_ghosts
+    msh_to_smsh = np.full(num_cells, -1)
+    msh_to_smsh[smhs_to_msh] = np.arange(len(smhs_to_msh))
+    entity_maps = {smsh._cpp_object: np.array(msh_to_smsh, dtype=np.int32)}
```
jorgensd, line 162: Do we check what happens if we send in a -1 in the map, and we try to use it in the kernels?
This should probably raise an error saying invalid map (maybe an assert).
Otherwise it would be hard to debug custom maps.

jorgensd (changes requested):
In general very positive to this.
I added some comments regarding usage of map vs using a span.
What are your thoughts @garth-wells @IgorBaratta  @chrisrichardson ?

The PR also implies that we already have mixed assembly kernels in FFCx, which is great!
Do we have some good tests in FFCx for those?

cpp/dolfinx/fem/assemble_matrix_impl.h
chrisrichardson: I already raised this idea with @jpdean. There are three lists we would need: `cells` - to get the geometry, and `cells_0` and `cells_1` which are mapped. For normal kernels, these would all be the same span, so no memory overhead.

cpp/dolfinx/fem/sparsitybuild.cpp
```diff
@@ -17,12 +17,14 @@ using namespace dolfinx::fem;
 //-----------------------------------------------------------------------------
 void sparsitybuild::cells(
     la::SparsityPattern& pattern, std::span<const std::int32_t> cells,
-    std::array<std::reference_wrapper<const DofMap>, 2> dofmaps)
```
francesco-ballarin, line 20: To me, it feels odd that the same change is not done in `sparsitybuild::interior_facets`. When I had to customize the sparsity pattern builder in `multiphenicsx` I had to change both `sparsitybuild::interior_facets` and `sparsitybuild::cells`.

cpp/dolfinx/fem/assemble_matrix_impl.h
jpdean: I'm happy to change to this approach if it is preferred

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -13,8 +13,9 @@
 
 import ufl
 from dolfinx import default_scalar_type, fem, la
+from dolfinx.fem.petsc import assemble_matrix
```
francesco-ballarin, line 16: I think that @chrisrichardson was trying to untie these tests from PETSc. Discuss with him whether assembly in the builtin matrix format would be acceptable here, or not.

cpp/dolfinx/fem/sparsitybuild.cpp
jpdean: So far, I've only added changes for cell integrals in this PR. When everyone is happy with the approach, I'll add the other integral types; focusing on cell integrals only for the time being just makes dealing with suggested changes a bit more manageable.

cpp/dolfinx/fem/assemble_matrix_impl.h
jpdean: The main issue with passing spans is that we'd need an "identity map" span for single domain problems, and it seems silly to have to store this. I tried to get around it by passing and empty span in this case and creating lambda functions where needed, but this is messy and not necessarily a good idea.

One approach might be to pass `entity_maps` to `create_form_factory` only, use the maps there, and for each integral, store the mapped cells along with the active cells. This data could then be passed to the assembler as a span, and in the case of single domain problems, we could just pass the same span twice. It also means we have contiguous access in the assembler with no need to map there. However, it might use more memory in certain cases, and it will also make [this](https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/Form.h#L88-L94) (already pretty hefty) argument more complex.

jpdean:
Closing in favour of [this](https://github.com/FEniCS/dolfinx/pull/3056)

Issue #3011, opened 2024-01-23
[BUG]: cmake: explicitly requested optional features that aren't available don't stop build


minrk:
### Summarize the issue

It would be helpful for me maintaining dolfinx on conda-forge if when I explicitly request optional features (e.g. `-DDOLFINX_ENABLE_KAHIP=ON`) and have cmake fail if those features aren't found instead of proceeding as if it wasn't requested.

These are well reported for humans, but I'm working on CI builds that would benefit from stopping automatically as soon as it's known that they aren't going to do what was requested.

off-by-default kahip is not hard because if ENABLE_KAHIP is set it's necessarily been explicitly requested, but on-by-default optional features like parmetis would need 3-state ON/OFF/AUTO so that it could be made to fail if unavailable and requested.

### How to reproduce the bug

Run:

```
cmake -DDOLFINX_ENABLE_KAHIP=ON -B build cpp
```

in a place without kahip, and proceed to build instead of exiting with nonzero status.

### Version

0.7.2

### DOLFINx git commit

v0.7.3 actually, but 0.7.3 isn't in Version dropdown

### Installation

conda, but the issue is in the cmake build, not for installed dolfinx

### Additional information

Something wonky's been going on right now with kahip detection failing _sometimes_ with `illegal instruction`. It's obviously not dolfinx's problem to fix the underlying issue, but the builds don't find out that the requested configuration wasn't used until the build completes and the tests fail with `dolfinx.cpp.common.has_kahip()`, and I could save a lot of time if they stopped at the configure step.

jhale:
@minrk In short, explicitly enabling a feature at the command line will now make the CMake package search `REQUIRED`.

minrk:
Awesome, thank you!

PR #3020, opened 2024-01-30
Make options required if enabled.

jhale:
This is one way to address https://github.com/FEniCS/dolfinx/issues/3011

The user must explicitly enable optional dependencies.


jhale:
This retains the previous behaviour while allowing package maintainers to explicitly ensure optional dependencies are found. @garth-wells I didn't totally understand your point, perhaps I've missed something so ping me.

garth-wells:
> This retains the previous behaviour while allowing package maintainers to explicitly ensure optional dependencies are found. @garth-wells I didn't totally understand your point, perhaps I've missed something so ping me.

PR has changed between my comment and the above comment.

I'm not sure the proposed change is super maintainable. How about for each optional package  we add a `FOO_REQUIRED` variable. If the user explicitly sets `DOLFINX_ENABLE_FOO`, then `FOO_REQUIRED` is set to `true`. This keeps the logic close to the package-related code in the CMake file and lets CMake do the error handling.

garth-wells:
@jhale how this approach: https://github.com/FEniCS/dolfinx/tree/garth/cmake-config?

jhale:
Superseded by #3024

PR #3028, opened 2024-02-01
Add matrix-free solver python demo

IgorBaratta:


python/demo/demo_poisson_matrix_free.py
```diff
@@ -0,0 +1,292 @@
...
+# We begin by using {py:func}`create_rectangle
+# <dolfinx.mesh.create_rectangle>` to create a rectangular
+# {py:class}`Mesh <dolfinx.mesh.Mesh>` of the domain, and creating a
+# finite element {py:class}`FunctionSpace <dolfinx.fem.FunctionSpace>`
+# $V$ on the mesh.
+
+dtype = dolfinx.default_scalar_type
+real_type = np.real(dtype(0.0)).dtype
+
+# Create mesh with tensor product ordering
```
mscroggs, line 102: update this

chrisrichardson:
It turned out there were quite a lot of fixes needed to the documentation for this... but now I think is pretty good.

nate-sime:
Just a couple of typos and things that I've spotted. Will push soon.

python/demo/demo_poisson_matrix_free.py
```diff
@@ -0,0 +1,264 @@
...
+            x.scatter_forward()
+            return k
+        p.array[:] = beta * p.array + r
+
+    raise RuntimeError(f"Solver exceeded max iterations ({max_iter}).")
+
+
+# This matrix-free solver is now used to compute the finite element solution.
+# The finite element solution's approximation error as compared with the
+# exact solution is measured in the $L_2$-norm.
```
nate-sime, line 244: The solution selected here is actually exactly represented in the FE space chosen. What we're measuring is the iterative solver error as propagated to the FE solution in the L2 norm. If we chose `u_D` to be something not in $P^3$ then we'd have a combination of the two. Perhaps this should be distinguished?

python/demo/demo_poisson_matrix_free.py
```diff
@@ -0,0 +1,264 @@
...
+
+rtol = 1e-6
+u = fem.Function(V, dtype=dtype)
+iter_cg1 = cg(mesh.comm, action_A, u.x, b, max_iter=200, rtol=rtol)
+
+# Set BC values in the solution vector
+fem.set_bc(u.x.array, [bc], scale=1.0)
```
nate-sime, line 251: I think I'm missing something here. Why is the BC set as a post solve step? Shouldn't this be accounted for in the matrix-free solve?

python/demo/demo_poisson_matrix_free.py
IgorBaratta: Good point, I think we should use something not in $P^3$.

python/demo/demo_poisson_matrix_free.py
```diff
@@ -0,0 +1,264 @@
...
+# finite element {py:class}`FunctionSpace <dolfinx.fem.FunctionSpace>`
+# on the mesh.
+
+dtype = dolfinx.default_scalar_type
+real_type = np.real(dtype(0.0)).dtype
+comm = MPI.COMM_WORLD
+mesh = dolfinx.mesh.create_rectangle(comm, [[0.0, 0.0], [1.0, 1.0]], [10, 10], dtype=real_type)
+
+# Create function space
+degree = 3
```
IgorBaratta, line 99:
```suggestion
degree = 2
```

PR #3034, opened 2024-02-05
Re-enable a test

garth-wells:


python/test/unit/fem/test_dofmap.py
```diff
@@ -272,9 +272,9 @@ def test_higher_order_coordinate_map(points, celltype, order):
 
 
 @pytest.mark.skip_in_parallel
-@pytest.mark.parametrize("order", [1, 2, 3])
+@pytest.mark.parametrize("order", [1, 2])
```
jorgensd, line 275: Should we instead investigate why third order fails?

garth-wells: It's simply not programmed to cover the P=3 case.

PR #3037, opened 2024-02-05
Wrap discrete_gradient in Python

### LINKED ISSUE #3022: Wrap discrete gradients in Python
### Summarize the issue

Currently have to pass in the C++ function spaces and no doc-strings passed to python
https://github.com/FEniCS/dolfinx/blob/f0054ad8061cac4285e37354b6b48d0a0b3cd76d/python/test/unit/fem/test_discrete_operators.py#L33

### How to reproduce the bug

From test_discrete_operators.py

### Minimal Example (Python)

```Python
from dolfinx.cpp.fem import discrete_gradient
G = discrete_gradient(V._cpp_object, W._cpp_object)
```

### Output (Python)

_No response_

### Version

main branch

### DOLFINx git commit

_No response_

### Installation

_No response_

### Additional information

_No response_

mscroggs:
Resolves #3022


bleyerj:
The lines 
https://github.com/FEniCS/dolfinx/blob/26f2067755ec21afa3df40cf6bd9b2195fd3881a/python/dolfinx/fem/petsc.py#L785-L791
seem to suggest that this is made for Lagrange -> Nedelec only. Is this really the case?
I was at some point interested in having a discrete gradient operator from Lagrange to DG for instance, or a Quadrature space.

michalhabera:
> The lines
> 
> https://github.com/FEniCS/dolfinx/blob/26f2067755ec21afa3df40cf6bd9b2195fd3881a/python/dolfinx/fem/petsc.py#L785-L791
> 
> 
> seem to suggest that this is made for Lagrange -> Nedelec only. Is this really the case?
> I was at some point interested in having a discrete gradient operator from Lagrange to DG for instance, or a Quadrature space.

The code makes assumptions that target space maps using covariant Piola, which would fail if you use DG. Just note that you can always build these discrete operators using dolfinx.Expression that contains ufl.TrialFunction on the input space. Assembly of these expressions into PETSc/scipy matrix is not implemented yet, but you can easily write a  custom (Numba) loop that does that. I presented this in my FEniCS2021 talk, can find the snippets if you want.

python/dolfinx/fem/__init__.py
```diff
@@ -57,6 +59,22 @@ def create_sparsity_pattern(a: Form):
...
 
+def discrete_gradient(space0: FunctionSpace, space1: FunctionSpace) -> _MatrixCSR:
+    """Assemble a discrete gradient operator.
+
+    The discrete gradient operator A interpolates the gradient of
+    a Lagrange finite element function into a H(curl) space.
+
+    Args:
+        space0: Lagrange space to interpolate the gradient from
+        space1: H(curl) space to interpolate into
```
mscroggs, line 70: say that this function assumes a covariant piola map

python/dolfinx/fem/__init__.py
```diff
@@ -57,6 +59,22 @@ def create_sparsity_pattern(a: Form):
...
 
 
+def discrete_gradient(space0: FunctionSpace, space1: FunctionSpace) -> _MatrixCSR:
+    """Assemble a discrete gradient operator.
+
+    The discrete gradient operator A interpolates the gradient of
+    a Lagrange finite element function into a H(curl) space.
+
+    Args:
+        space0: Lagrange space to interpolate the gradient from
```
mscroggs, line 69:
```suggestion
        space0: H1 space to interpolate the gradient from
```

python/dolfinx/fem/petsc.py
```diff
@@ -952,3 +957,31 @@ def J(self, x: PETSc.Vec, A: PETSc.Mat) -> None:
...
+
+def discrete_gradient(space0: _FunctionSpace, space1: _FunctionSpace) -> PETSc.Mat:
+    """Assemble a discrete gradient operator.
+
+    The discrete gradient operator A interpolates the gradient of
+    a Lagrange finite element function into a H(curl) space.
+
+    Args:
+        space0: Lagrange space to interpolate the gradient from
+        space1: H(curl) space to interpolate into
```
mscroggs, line 970: same changes here

PR #3038, opened 2024-02-05
Wrap entities_to_geometry in Python

### LINKED ISSUE #3023: Wrap `entities_to_geometry` in Python interface
### Summarize the issue

Documentation not created, and C++ objects have to be sent through

### How to reproduce the bug

See test_mesh: https://github.com/FEniCS/dolfinx/blob/f0054ad8061cac4285e37354b6b48d0a0b3cd76d/python/test/unit/mesh/test_mesh.py#L64C1-L64C97

### Minimal Example (Python)

```Python
e_to_g = entities_to_geometry(mesh._cpp_object, entity_dim, np.array(entity_map), False)
```

### Output (Python)

_No response_

### Version

main branch

### DOLFINx git commit

_No response_

### Installation

_No response_

### Additional information

_No response_

mscroggs:
Resolves #3023

### DIFF
--- a/python/dolfinx/mesh.py
+++ b/python/dolfinx/mesh.py
@@ -645,3 +645,20 @@ def create_unit_cube(comm: _MPI.Comm, nx: int, ny: int, nz: int, cell_type=CellT
     """
     return create_box(comm, [np.array([0.0, 0.0, 0.0]), np.array([1.0, 1.0, 1.0])],
                       [nx, ny, nz], cell_type, dtype, ghost_mode, partitioner)
+
+
+def entities_to_geometry(
+    mesh: Mesh, dim: int, entities: np.NDArray[np.int32], orient: bool
+) -> np.NDArray[np.int32]:
+    """Determine the indices in the geometry data for each vertex of the given mesh entities.
+
+    Args:
+        mesh: The mesh
+        dim: Topological dimension of the entities of interest
+        entities: Entity indices (local) to compute the vertex geometry indices for
+        orient: If true, in 3D, reorients facets to have consistent normal direction
+
+    Returns:
+        Indices in the geometry array for the entity vertices
+    """
+    return _entities_to_geometry(mesh, dim, entities, orient)
--- a/python/test/unit/mesh/test_mesh.py
+++ b/python/test/unit/mesh/test_mesh.py
@@ -18,11 +18,11 @@ from basix.ufl import element
 from dolfinx import cpp as _cpp
 from dolfinx import graph
 from dolfinx import mesh as _mesh
-from dolfinx.cpp.mesh import create_cell_partitioner, entities_to_geometry, is_simplex
+from dolfinx.cpp.mesh import create_cell_partitioner, is_simplex
 from dolfinx.fem import assemble_scalar, coordinate_element, form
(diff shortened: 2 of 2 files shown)


jorgensd (approved):

chrisrichardson (approved):

garth-wells (approved):

python/dolfinx/mesh.py
```diff
@@ -645,3 +645,26 @@ def create_unit_cube(comm: _MPI.Comm, nx: int, ny: int, nz: int, cell_type=CellT
...
+    Warning:
+        This function should not be used unless there is no alternative.
+        It may be removed in the future.
+
+    Args:
+        mesh: The mesh.
+        dim: Topological dimension of the entities of interest.
+        entities: Entity indices (local to the process) to determine the
+            vertex geometry indices for.
+        orient: If ``True``, in 3D, re-orients facets to have consistent
```
garth-wells, line 664: What does this mean? (I know it straight from the C++ docs). @jorgensd?

jorgensd: This is a flag only used by BEMPP (at least at its introduction) to ensure consistent orientation of each facet (to make all cell normals point outwards).

@mscroggs might be able to comment on the necessity of this at the current stage of BEMPP

garth-wells: It it's still needed it needs an explanation of what is required for a consistent orientation, i.e. document how we related the normal vector to a facet to the vertex ordering. And does it apply for all facet shapes?

mscroggs: This is still used by Bempp-cl, so I'm keen to keep it for now. I think we've only tested it for triangle facets, and I'd be keen to work towards removing this once Bempp-rs can do coupling without needing this.

python/dolfinx/mesh.py
```diff
@@ -753,3 +754,26 @@ def create_unit_cube(
...
+        This function should not be used unless there is no alternative.
+        It may be removed in the future.
+
+    Args:
+        mesh: The mesh.
+        dim: Topological dimension of the entities of interest.
+        entities: Entity indices (local to the process) to determine the
+            vertex geometry indices for.
+        orient: If True, the triangular facets of a 3D mesh will be reordered
+            so that they have a consistent normal direction.
```
mscroggs, line 774: state that this normal is inferred from edge 2 x edge 1

mscroggs: And add a note that this argument will be removed in future

python/dolfinx/mesh.py
```diff
@@ -753,3 +754,26 @@ def create_unit_cube(
         ghost_mode,
         partitioner,
     )
+
+
+def entities_to_geometry(
+    mesh: Mesh, dim: int, entities: npt.NDArray[np.int32], orient: bool
```
mscroggs, line 760:
```suggestion
    mesh: Mesh, dim: int, entities: npt.NDArray[np.int32], orient: bool = False
```

python/dolfinx/mesh.py
mscroggs, line 774: document that this option may be removed in future

PR #3044, opened 2024-02-07
Ruff format mk2

jhale:
- Format, still some line too long errors?
- Modify CI.
- Remove flake8 and isort from image.
- Format C++ part.
- Fix remaining check issues.


cpp/doc/source/conf.py
```diff
@@ -17,25 +17,26 @@
...
-project = 'DOLFINx'
-copyright = '2022, FEniCS Project'
-author = 'FEniCS Project'
+project = "DOLFINx"
+copyright = "2022, FEniCS Project"
+author = "FEniCS Project"
 
 # The full version, including alpha/beta/rc tags
-release = '0.3.1'
+release = "0.3.1"
```
francesco-ballarin, line 25: This number seems a bit out of date :P

francesco-ballarin: (of course that's unrelated to these changes, but noticed it while scrolling)

python/doc/source/conf.py
```diff
@@ -6,11 +6,12 @@
 import os
 import sys
 
+import jupytext_process
```
francesco-ballarin, line 9: This is wrong. The reason the import of `jupytext_process` was on line 13 (original numbering) is that it is a file in the source directory, and therefore it needs lines 11

cpp/doc/source/conf.py
jhale: I think we should leave this PR as minimal as possible with respect to it's goal of introducing ruff formatting.

python/pyproject.toml
```diff
@@ -72,31 +72,17 @@ warn_unused_ignores = false
...
-      "STDLIB",
-      "MPI",
-      "THIRDPARTY",
-      "FIRSTPARTY",
-      "LOCALFOLDER",
-]
 
 [tool.ruff]
-line-length = 120
+line-length = 100
```
francesco-ballarin, line 77: Was this on purpose?

(asking because the very first bullet in the PR says about some lines too long errors)

python/test/unit/fem/test_custom_assembler.py
```diff
@@ -71,27 +73,36 @@ def get_matsetvalues_cffi_api():
...
+    petsc_dir = PETSc_get_config()["PETSC_DIR"]
     petsc_arch = petsc4py.lib.getPathArchPETSc()[1]
 
-    worker = os.getenv('PYTEST_XDIST_WORKER', None)
+    worker = os.getenv("PYTEST_XDIST_WORKER", None)
     module_name = f"_petsc_cffi_{worker}"
     if MPI.COMM_WORLD.Get_rank() == 0:
         ffibuilder = cffi.FFI()
-        ffibuilder.cdef("""typedef int... PetscInt;
+        ffibuilder.cdef(
```
francesco-ballarin, line 86: Am I reading this right: ruff also formatted the C code in the string, not only the python code around it?

That is not an issue here, but we may want to be aware of this when applying ruff format e.g. to ffcx, which I guess has more parts like this one.

python/pyproject.toml
jhale: Yes, 100 is considered the 'maximum' permissible length, see e.g. https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length

python/test/unit/fem/test_custom_assembler.py
chrisrichardson: In general though, I think it is bad that ruff edits the contents of triple quotes.

chrisrichardson:
Not really happy about the multi-line imports, formatting around the `:` slice operator, and reformatting of triple quoted text. But I guess we'll just have to live with it.

PR #3045, opened 2024-02-07
Simplify `Form` constructor by adding a `struct` for integral data

jpdean:
The [integrals](https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/Form.h#L88-L94) argument in the `Form` constructor isn't very easy to read. I've added a `struct` to simplify things (this will also help make the code more readable when mixed-domain support is added), and I've removed a `std::map`. The struct is templated over container type to avoid some copies, but maybe this isn't the best solution.


cpp/dolfinx/fem/Form.h
```diff
@@ -110,12 +137,14 @@ class Form
     if (!_mesh)
       throw std::runtime_error("No mesh could be associated with the Form.");
 
+    // TODO Check integrals is sorted by ID?
+
```
chrisrichardson, line 141:
```suggestion
 for (auto [k, v] : integrals)
    {
      if (!is_sorted(v.begin(), v.end(),
                     [](auto a, auto b) { return (a.id < b.id); }))
                throw std::runtime_error("Integral IDs not sorted");
    }
```

chrisrichardson (approved):

garth-wells:
I'm not 100% sure on the syntax suggestions with `auto`. It would though help with inlining kernels.

PR #3052, opened 2024-02-10
C++ demos document generation

ampdes:
An illustration for one of the demos: biharmonic equation


jhale:
I think we would like to keep the single source file per demo approach.

Jupytext, the tool we use to convert light formatted Python to rst, also supports converting light formatted C++ to rst, which can then be included in sphinx/breathe.

jhale:
Additionally, this will require converting the existing rst comments to myst flavoured markdown. There is an automated tool to do this.

https://docs.readthedocs.io/en/stable/guides/migrate-rest-myst.html

garth-wells:
Thanks for looking at this.

We aim with the doc system to be as 'standard' and simple as possible, especially at the lowest levels in the processing chain. We've seen (many times over now!) that we're not able to maintain bespoke doc processing.

Ideally on the C++ we can just run `doxygen` and not bring in CMake or any Python dependencies at this level, and keep all the Python-required handling in once place (`python/doc`).

jhale:
@garth-wells I'm having a chat with @ampdes in Slack #development, I think we've found a good solution; for the demos we use light formatted C++ + myst-flavoured markdown + jupytext + sphinx/breathe, just like for the Python demos. It's been very robust so far.

A link to the demos will appear here:

https://docs.fenicsproject.org/dolfinx/v0.7.3/cpp/

Everything else stays the same, including the standard doxygen build for API documentation and the CI.

ampdes:
This is not yet configured to run in github workflows.
The cmake still exists, since it was easier to build it using cmake.

cpp/demo/hyperelasticity/main.cpp
```diff
@@ -1,6 +1,19 @@
...
+// jupyter:
+//   jupytext:
+//     text_representation:
+//       extension: .cpp
+//       format_name: light
+//       format_version: '1.5'
+//       jupytext_version: 1.16.1
+//   kernelspec:
+//     display_name: C++17
+//     language: C++17
```
jorgensd, line 11: Can we do C++20 here? That is the minimal required standard we have for the actual code.

ampdes: Specifying kernel_spec is not necessary. This is only needed when this file is opened in a notebook environment, then kernel_spec will be used for a kernel with interactive compiler.
I was trying [xeus-cling](https://github.com/jupyter-xeus/xeus-cling) locally and this jupytext metadata is a remant from that explore. 

One could have just got hold of `main.cpp` file and run

```bash
~/.local/bin/jupytext --set-formats cpp,md main.cpp
```
to bind `.cpp` with a `.md` file, which results in

```
---
jupyter:
  jupytext:
    cell_metadata_filter: -all
    formats: cpp,md
    main_language: c++
    text_representation:
      extension: .md
      format_name: markdown
      format_version: '1.3'
      jupytext_version: 1.16.1
---
```
in `.md` file and 

```
// ---
// jupyter:
//   jupytext:
//     cell_metadata_filter: -all
//     formats: cpp:light,md
//     text_representation:
//       extension: .cpp
//       format_name: light
//       format_version: '1.5'
//       jupytext_version: 1.16.1
// ---
```
`.cpp` file.

https://jupytext.readthedocs.io/en/latest/advanced-options.html

I will simply remove all kernel_spec.

ampdes: Actually only

```bash
~/.local/bin/jupytext --set-formats cpp main.cpp
```
is needed for each `main.cpp` file to populate with jupytext header, which will then be used later to convert to `.md`

cpp/doc/source/jupytext_process.py
```diff
@@ -0,0 +1,64 @@
...
+                                                       "````" + python_myst_text + "````")
+
+            for cmake_file in demo_subdir.glob("CMakeLists.txt"):
+                shutil.copy(cmake_file, demo_doc_subdir)
+
+            myst_file = (demo_doc_dir / fname.name).with_suffix(".md")
+            with open(myst_file, "w") as fw:
+                fw.write(cpp_myst_text)
+
+            # There is a posibility to use jupyter-notebooks with C++/C kernels
```
jhale, line 59: I feel C++ and Notebooks is a bit of an edge use case.

However, free to export them, but don't link to them at this stage; someone could take a look and see if https://github.com/jupyter-xeus/xeus-cling works with the conda FEniCS packages.

cpp/demo/biharmonic/main.cpp
```diff
@@ -1,124 +1,160 @@
...
+// V = \left\{v \in H^{1}_{0}(\Omega)\,:\, v \in P_{k}(K) \
+// \forall \ K \in \mathcal{T} \right\}
+// $$
 //
 // and considering the boundary conditions
 //
-// .. math::
-//    u            &= 0 \quad {\rm on} \ \partial\Omega \\
-//    \nabla^{2} u &= 0 \quad {\rm on} \ \partial\Omega
+// $$
```
jhale, line 71: Would be better to use the amsmath environments rather than $$

https://myst-parser.readthedocs.io/en/v0.14.0/using/syntax-optional.html#syntax-amsmath

cpp/doc/Doxyfile
```diff
@@ -1010,7 +1009,11 @@ EXCLUDE                = ../dolfinx/io/pugixml.hpp \
                          ../dolfinx/io/pugixml.cpp \
                          ../dolfinx/io/pugiconfig.hpp \
                          ../dolfinx/common/loguru.hpp \
-                         ../dolfinx/common/loguru.cpp
+                         ../dolfinx/common/loguru.cpp \
+                        # TODO: resolve Doxygen parsing in:
+                        #  ../dolfinx/fem/interpolate.h \
```
jhale, line 1014: Please discuss in Slack if struggling.

cpp/demo/biharmonic/main.cpp
```diff
@@ -1,124 +1,160 @@
-// Biharmonic equation (C++)
-// =========================
+// ---
+// jupyter:
+//   jupytext:
+//     cell_metadata_filter: -all
```
jhale, line 4: What does this do?

jhale (commented):
Still a few more things to do; looking very good.

cpp/demo/hyperelasticity/main.cpp
ampdes: I was presuming that this jupytext header is needed for reading and then converting to `.md` files.
It turns out that all these specifications are only for connecting `.md` or `.cpp` files to notebook files.

All these are deleted.

I didn't yet try xeus-cling with dolfinx.

cpp/demo/biharmonic/main.cpp
ampdes: One could use both $$ and amsmath simultaneously.

cpp/demo/biharmonic/main.cpp
ampdes: It turns out it isnt needed unless for connecting the scripts to notebooks.

jhale:
This looks close to being ready to merge?

cpp/demo/hyperelasticity/main.cpp
jhale: Interesting, so perhaps we can also reduce or remove them in the .py files.

jhale:
I'm going to merge this, cleanup and tweaking can be done later.

garth-wells:
@ampdes very nice job.

PR #3054, opened 2024-02-14
Fix REQUIRED dependencies when using repeat CMake calls.

jhale:
This stores the `_REQUIRED_*` variable in the cache so that a second call to cmake gives the expected behaviour. Without this if e.g. SCOTCH is not found on the first run, it will be REQUIRED on the second, due to DOLFINX_REQUIRE_SCOTCH being ON in CMakeCache.txt.


chrisrichardson:
Thanks for looking at this - it has been annoying me too.

python/demo/demo_stokes.py
```diff
@@ -427,7 +427,7 @@ def block_direct_solver():
     # handle pressure nullspace
     pc = ksp.getPC()
     pc.setType("lu")
-    pc.setFactorSolverType("mumps")
```
chrisrichardson, line 430: Should this be in the same PR?

chrisrichardson (approved):

PR #3056, opened 2024-02-15
Add maps for mixed-domain assembly

jpdean:
This PR is a redesign of [this PR](https://github.com/FEniCS/dolfinx/pull/3016). It only focuses on cell integrals without coefficients; I'll add coefficient packing and the other integral types if everyone is happy with this approach.


python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -116,3 +118,71 @@ def test_submesh_facet_assembly(n, k, space, ghost_mode):
...
+    cell_imap = msh.topology.index_map(tdim)
+    num_cells = cell_imap.size_local + cell_imap.num_ghosts
+    msh_to_smsh = np.full(num_cells, -1)
+    msh_to_smsh[smhs_to_msh] = np.arange(len(smhs_to_msh))
+    entity_maps = {smsh._cpp_object: np.array(msh_to_smsh, dtype=np.int32)}
+
+    a_1 = fem.form(ufl_form(u, w, dx_msh(tag)), entity_maps=entity_maps)
+    A_1 = assemble_matrix(a_1)
+    A_1.assemble()
+    assert np.isclose(A_1.norm(), A.norm())
```
jorgensd, line 188: Should we add a test where we compute this by using a `MeshTag` to restrict the integration and compare norms? Currently this test would not pick up a consistent error in mixed-assembly

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -28,20 +28,20 @@ using mdspan2_t = MDSPAN_IMPL_STANDARD_NAMESPACE::mdspan<
...
-                    std::span<const std::int8_t> bc0,
-                    std::span<const std::int8_t> bc1, FEkernel<T> auto kernel,
-                    std::span<const T> coeffs, int cstride,
-                    std::span<const T> constants,
-                    std::span<const std::uint32_t> cell_info)
+void assemble_cells(
+    la::MatSet<T> auto mat_set, mdspan2_t x_dofmap,
+    std::span<const scalar_value_type_t<T>> x,
+    std::span<const std::int32_t> cells, std::span<const std::int32_t> cells_0,
+    std::span<const std::int32_t> cells_1,
```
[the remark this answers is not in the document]
jpdean: No need for any more cell arrays, regardless of how many domains the coefficients are defined over. Mixed-domain coefficients will be handled when they are packed. The approach will work for an arbitrary number of coefficients defined over an arbitrary number of domains

python/test/unit/fem/test_assemble_submesh.py
jpdean: Sorry, I'm not sure what you mean. This test uses a meshtag to compare to a single-domain problem. A better test would be great though; ideally it would be better to do a more thorough check than just comparing norms (comparing norms also won't work for more complex elements)

jpdean: When the other mixed-domain routines have been added, I can add some more comprehensive tests

cpp/dolfinx/fem/Form.h
```diff
@@ -139,16 +145,21 @@ class Form
...
   {
     // Extract _mesh from FunctionSpace, and check they are the same
     if (!_mesh and !V.empty())
       _mesh = V[0]->mesh();
     for (auto& space : V)
-      if (_mesh != space->mesh())
-        throw std::runtime_error("Incompatible mesh");
+      if (_mesh != space->mesh()
+          and entity_maps.find(space->mesh()) == entity_maps.end())
+        throw std::runtime_error(
```
garth-wells, line 161: Add braces.

cpp/dolfinx/fem/Form.h
```diff
@@ -277,6 +292,43 @@ class Form
       throw std::runtime_error("No mesh entities for requested domain index.");
   }
 
+  /// @brief Compute the list of entity indices for the ith integral (kernel)
+  /// of a given type. The indices are mapped from the integration domain mesh
```
garth-wells, line 296: 'of a given type' isn't clear. Does it mean same kernel, or same integration domain type?

cpp/dolfinx/fem/Form.h
```diff
@@ -277,6 +292,43 @@ class Form
...
+  /// to the mesh `mesh`.
+  ///
+  /// @param type Integral type
+  /// @param i Integral ID, i.e. (sub)domain index
+  /// @param mesh The mesh to map the entities to
+  /// @return List of active entities in `mesh` for the given integral
+  std::vector<std::int32_t>
+  domain(IntegralType type, int i,
+         std::shared_ptr<const mesh::Mesh<U>> mesh) const
```
garth-wells, line 305: This should be a `const mesh::Mesh<U>&`. Only pass `shared_ptr` to a class method that holds onto the pointer.

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -28,20 +28,20 @@ using mdspan2_t = MDSPAN_IMPL_STANDARD_NAMESPACE::mdspan<
...
-                    mdspan2_t dofmap1, int bs1,
-                    std::span<const std::int8_t> bc0,
-                    std::span<const std::int8_t> bc1, FEkernel<T> auto kernel,
-                    std::span<const T> coeffs, int cstride,
-                    std::span<const T> constants,
-                    std::span<const std::uint32_t> cell_info)
+void assemble_cells(
+    la::MatSet<T> auto mat_set, mdspan2_t x_dofmap,
+    std::span<const scalar_value_type_t<T>> x,
+    std::span<const std::int32_t> cells, std::span<const std::int32_t> cells_0,
```
garth-wells, line 36: `cells0`

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -56,9 +56,16 @@ void assemble_cells(la::MatSet<T> auto mat_set, mdspan2_t x_dofmap,
...
 
   // Iterate over active cells
+  assert(cells_0.size() == cells.size());
+  assert(cells_1.size() == cells.size());
   for (std::size_t index = 0; index < cells.size(); ++index)
   {
+    // Cell in integration domain mesh
     std::int32_t c = cells[index];
+    // Corresponding cell in test function mesh
+    std::int32_t c_0 = cells_0[index];
```
garth-wells, line 66: `c0`

cpp/dolfinx/fem/sparsitybuild.cpp
```diff
@@ -19,10 +19,19 @@ void sparsitybuild::cells(
     la::SparsityPattern& pattern, std::span<const std::int32_t> cells,
     std::array<std::reference_wrapper<const DofMap>, 2> dofmaps)
 {
+  sparsitybuild::cells(pattern, cells, cells, dofmaps);
+}
+//-----------------------------------------------------------------------------
+void sparsitybuild::cells(
+    la::SparsityPattern& pattern, std::span<const std::int32_t> cells_0,
```
garth-wells, line 26: `cells0`

cpp/dolfinx/fem/Form.h
```diff
@@ -139,16 +145,21 @@ class Form
        const std::vector<std::shared_ptr<const Constant<scalar_type>>>&
            constants,
        bool needs_facet_permutations,
-       std::shared_ptr<const mesh::Mesh<U>> mesh = nullptr)
+       std::shared_ptr<const mesh::Mesh<U>> mesh = nullptr,
+       const std::map<std::shared_ptr<const mesh::Mesh<U>>,
+                      std::span<const std::int32_t>>& entity_maps
```
garth-wells, line 150: We try top avoid default args when we can. Maybe add a second constructor?

garth-wells (changes requested):
First pass.

python/test/unit/fem/test_assemble_submesh.py
jorgensd: Oh, sorry, missed that you are not comparing A_1 to A_0, but both of them to `A`

garth-wells (approved):

PR #3066, opened 2024-02-24
Mark IndexMaps on communicators with size 1 as non-overlapped

garth-wells:
For Functions on an MPI communicator of size 1, this change reduced the number of communicators that are created.

### DIFF
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -788,7 +788,7 @@ IndexMap::IndexMap(MPI_Comm comm, std::int32_t local_size,
                    std::span<const int> owners)
     : _comm(comm, true), _ghosts(ghosts.begin(), ghosts.end()),
       _owners(owners.begin(), owners.end()), _src(src_dest[0]),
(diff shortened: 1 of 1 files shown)


chrisrichardson:
Can we get rid of `_overlapping` and just store the comm size (or ask MPI for it)? I don't think we ever use an `IndexMap` with no overlap, except when in serial. At least, the constructor `IndexMap(comm, local_size)` doesn't seem to be used internally.

jorgensd:
> Can we get rid of `_overlapping` and just store the comm size (or ask MPI for it)? I don't think we ever use an `IndexMap` with no overlap, except when in serial. At least, the constructor `IndexMap(comm, local_size)` doesn't seem to be used internally.

I use that indexmap ( `IndexMap(comm, local_size)`) in some of my codes. I guess I can adapt it to sending in no ghosts.

garth-wells:
> I use that indexmap ( `IndexMap(comm, local_size)`) in some of my codes. I guess I can adapt it to sending in no ghosts.

It’s used in a few tests too, so I think we should keep the constructor.

PR #3068, opened 2024-02-25
C++ demo documentation refinements

garth-wells:


cpp/demo/biharmonic/main.cpp
```diff
@@ -193,26 +183,24 @@ int main(int argc, char* argv[])
...
-//  created using the class {cpp:class}`DirichletBC`. A
-//  {cpp:class}`DirichletBC` takes two arguments: the value of the
-//  boundary condition, and the part of the boundary on which the
-//  condition applies. In our example, the value of the boundary
-//  condition (0.0) can represented using a {cpp:class}`Function`,
-//  and the Dirichlet boundary is defined by the indices of degrees
-//  of freedom to which the boundary condition applies. The
-//  definition of the Dirichlet boundary condition then looks as
-//  follows:
+    //  Now, the Dirichlet boundary condition ($u = 0$) can be
```
ampdes, line 192: I think when these spaces are introduced, then the comment appears inside code block. To make the text appear outside code block, the `//` should not be preceded by any spaces.

garth-wells: We use `clang-format`, which indents the comments.  This means comments in function are inline comments.

The same happens with Python docs. To work with code formatters, all comments within function bodies are inline.

ampdes: Demos could be an exception to the formatters/linters, which already is the case for cpp demos.

Lengthy comments are explanation which looks nicer if rendered outside code block in Sphinx documentation.

garth-wells: We've learnt from many past mistakes to avoid 'custom' solutions. We need to support `clang-format` on all C++ files.

I used to think comments looked better rendered, but it does break up the body of the function which reduces readability. Some demos could do with more structuring into functions (see #3076), with detailed docs for functions and lightweight inline comments.

PR #3079, opened 2024-02-28
Add mixed-domain matrix assembly for facet integrals (codimension 0)

jpdean:


cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -162,28 +162,41 @@ template <dolfinx::scalar T>
 void assemble_exterior_facets(
     la::MatSet<T> auto mat_set, mdspan2_t x_dofmap,
     std::span<const scalar_value_type_t<T>> x,
-    std::span<const std::int32_t> facets, fem::DofTransformKernel<T> auto P0,
-    mdspan2_t dofmap0, int bs0, fem::DofTransformKernel<T> auto P1T,
-    mdspan2_t dofmap1, int bs1, std::span<const std::int8_t> bc0,
+    std::span<const std::int32_t> facets,
+    std::tuple<mdspan2_t, int, std::span<const std::int32_t>> dofmap0,
```
jorgensd, line 166: Could we have a docstring for this internal function, as the dofmap0 contains alot more than the dofmap

jorgensd:
Should we also do interior facets while we are at it?

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -201,12 +242,12 @@ void assemble_exterior_facets(
     kernel(Ae.data(), coeffs.data() + index / 2 * cstride, constants.data(),
            coordinate_dofs.data(), &local_facet, nullptr);
 
-    P0(_Ae, cell_info, cell, ndim1);
-    P1T(_Ae, cell_info, cell, ndim0);
+    P0(_Ae, cell_info1, cell1, ndim1);
```
garth-wells, line 245: Are you sure things are the correct way around here?

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -154,19 +162,25 @@ def test_mixed_dom_codim_0(n, k, space, ghost_mode):
...
     # Define form to compare to and assemble
-    a = fem.form(ufl_form(u, v, dx_msh(tag)))
+    a = fem.form(ufl_form(u, v, dx_msh(tag), ds_msh(tag)))
     A = fem.assemble_matrix(a)
     A.scatter_reverse()
 
     # Assemble a mixed-domain form, taking smsh to be the integration domain
     # Entity maps must map cells in smsh (the integration domain mesh) to
     # cells in msh
+    facets_sm = locate_entities(smsh, fdim, lambda x: np.isclose(x[0], 0.0))
```
garth-wells, line 176: Could we use something more informative than `_sm`?

cpp/dolfinx/fem/assemble_matrix_impl.h
jpdean: Whoops, these aren't right. Well spotted!

garth-wells: Not that I'm 100% sure either . . .

jpdean: It was wrong before, the DOF transformation for the test element should be applied using the test mesh cell info, indexed using the test mesh cell

PR #3087, opened 2024-03-04
Add mixed-domain vector assembly

jpdean:
Same as [this](https://github.com/FEniCS/dolfinx/pull/3056) PR but for vector assembly


cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -194,36 +204,50 @@ void _lift_bc_exterior_facets(
     std::span<T> b, mdspan2_t x_dofmap,
     std::span<const scalar_value_type_t<T>> x, FEkernel<T> auto kernel,
     std::span<const std::int32_t> facets,
-    fem::DofTransformKernel<T> auto pre_dof_transform, mdspan2_t dofmap0,
-    int bs0, fem::DofTransformKernel<T> auto post_dof_transpose,
-    mdspan2_t dofmap1, int bs1, std::span<const T> constants,
+    std::tuple<mdspan2_t, int, std::span<const std::int32_t>> dofmap0,
```
jorgensd, line 207: I am still not comfortable with us using tuples for these input arguments. We do not use it for any other variables, even if they are tied together. @garth-wells  what do you think?
It doesn't hurt readability and documentation to have three input arguments per direction.

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -194,36 +204,50 @@ void _lift_bc_exterior_facets(
...
   std::vector<scalar_value_type_t<T>> coordinate_dofs(3 * x_dofmap.extent(1));
   std::vector<T> Ae, be;
   assert(facets.size() % 2 == 0);
+  assert(facets0.size() == facets.size());
+  assert(facets1.size() == facets.size());
   for (std::size_t index = 0; index < facets.size(); index += 2)
   {
+    // Cell in integration domain mesh
     std::int32_t cell = facets[index];
+    // Cell in test fuction mesh
```
jorgensd, line 232:
```suggestion
    // Cell in test function mesh
```

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -515,11 +558,11 @@ void assemble_cells(fem::DofTransformKernel<T> auto dof_transform,
     std::fill(be.begin(), be.end(), 0);
     kernel(be.data(), coeffs.data() + index * cstride, constants.data(),
            coordinate_dofs.data(), nullptr, nullptr);
-    dof_transform(_be, cell_info, c, 1);
+    dof_transform(_be, cell_info, c0, 1);
```
jorgensd, line 561: Do we have a test that checks that with using `c` in the transform, that the code goes wrong?

Also, what mesh is cell_info associated with? In the codes above we use `cell_info0`, `cell_info1` explicitly, and I guess cell info for a submesh can be different than the parent mesh.

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -582,11 +630,11 @@ void assemble_exterior_facets(fem::DofTransformKernel<T> auto dof_transform,
     fn(be.data(), coeffs.data() + index / 2 * cstride, constants.data(),
        coordinate_dofs.data(), &local_facet, nullptr);
 
-    dof_transform(_be, cell_info, cell, 1);
+    dof_transform(_be, cell_info, cell0, 1);
```
jorgensd, line 633: Same question regarding cell_info

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -667,8 +725,8 @@ void assemble_interior_facets(
     std::span<T> _be(be);
     std::span<T> sub_be = _be.subspan(bs * dmap0.size(), bs * dmap1.size());
 
-    dof_transform(be, cell_info, cells[0], 1);
-    dof_transform(sub_be, cell_info, cells[1], 1);
+    dof_transform(be, cell_info, cells0[0], 1);
+    dof_transform(sub_be, cell_info, cells0[1], 1);
```
jorgensd, line 729: Same comment

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -945,18 +1029,23 @@ void assemble_vector(
     std::span<const std::int32_t> cells = L.domain(IntegralType::cell, i);
     if (bs == 1)
     {
-      impl::assemble_cells<T, 1>(dof_transform, b, x_dofmap, x, cells, dofs, bs,
-                                 fn, constants, coeffs, cstride, cell_info);
+      impl::assemble_cells<T, 1>(
+          dof_transform, b, x_dofmap, x, cells,
+          {dofs, bs, L.domain(IntegralType::cell, i, *mesh0)}, fn, constants,
+          coeffs, cstride, cell_info0);
```
jorgensd, line 1035: Ok, so I see that you are sending in cell_info for the submesh here. Maybe we should change the name of the cell_info inside each of these functions to ensure that we understand what they are associated with? 

I think we need proper docstrings of all the assemblies (`assemble_cells`, `assemble_exterior_facets`, ..., `lift_bc_cells` etc) as they are getting more complicated with this kind of additional arguments and association with different meshes.

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -119,86 +119,138 @@ def test_submesh_facet_assembly(n, k, space, ghost_mode):
 
 
 @pytest.mark.parametrize("n", [4, 6])
-@pytest.mark.parametrize("k", [1, 4])
+@pytest.mark.parametrize("k", [1, 3])
```
jorgensd, line 122: What is the reason for this reduction?

python/test/unit/fem/test_assemble_submesh.py
jorgensd, line 1: I think it would be good to simplify `test_mixed_dom_codim_0` into multiple tests for the different domains.
The test now is very lengthy, and makes it hard to pinpoint issues for the user.
This could either be done by:
- Creating a fixture that creates the mesh and sub-mesh of codim 0, and use that in multiple tests.
- Keep a single test, but paramterize over the integration measure.

jorgensd (changes requested):

cpp/dolfinx/fem/assemble_vector_impl.h
jpdean: It's just to highlight which cells index which dofmap, but I agree, we should probably be consistent

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -480,27 +517,33 @@ void _lift_bc_interior_facets(
...
-                    std::span<const scalar_value_type_t<T>> x,
-                    std::span<const std::int32_t> cells, mdspan2_t dofmap,
-                    int bs, FEkernel<T> auto kernel,
-                    std::span<const T> constants, std::span<const T> coeffs,
-                    int cstride, std::span<const std::uint32_t> cell_info)
+void assemble_cells(
+    fem::DofTransformKernel<T> auto dof_transform, std::span<T> b,
+    mdspan2_t x_dofmap, std::span<const scalar_value_type_t<T>> x,
+    std::span<const std::int32_t> cells,
+    std::tuple<mdspan2_t, int, std::span<const std::int32_t>> dofmap,
```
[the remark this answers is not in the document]
jpdean: True. Since matrix assembly has already been merged using the same approach, maybe we could make an issue and address the best consistent approach in a future PR?

cpp/dolfinx/fem/assemble_vector_impl.h
jpdean: We don't have a test for that, but it doesn't make sense to index the cell info from the test function mesh with `c`. 

`cell_info` is associated with the rest function mesh, I'll rename it to `cell_info0` for clarity.

python/test/unit/fem/test_assemble_submesh.py
jpdean: No reason. I think third degree tests the same things as fourth but it a bit cheaper. I can put it back though if you prefer

jorgensd: Just worried that it was removing a failing test. no worries, don't need to add it back

cpp/dolfinx/fem/assemble_vector_impl.h
garth-wells: I think the change is in the right direction - it reigns in the number of arguments, grouping together related objects. Otherwise the list of arguments gets too long and this can become error prone.

We have adopted this approach in other places, e.g. to collect together the arguments that define a finite element when creating a function spaces (Python).

garth-wells (approved):

PR #3090, opened 2024-03-06
Let  `*_apply_dof_transformation` accept lists of cells in Python layer

### LINKED ISSUE #3089: Expose looped versions of dof_transformations

To use dof transformations efficiently in Python, it would be good to have an interface where we can send in a contiguous chunk of data, and a set of cell_perm infos, to for instance: `pre_apply_transpose_dof_transformation`, and loop over the data (for instance in nanobind), calling the appropriate function for sub-spans of the original input.

### Suggested user interface

```python3
dof_data = np.arange(num_cells*num_dofs_per_cell, dtype=np.int32)
cell_infos = np.arange(num_cells*num_dofs_per_cell, dtype=np.uint8)
element.pre_apply_transpose_dof_transformation(dof_data, cell_infos, 1)
```

jorgensd:
Makes this function useful from Python when you have a contiguous array relating to more than a single cell with a single permutation.
Resolves #3089.
Tested for multiple cells in: https://github.com/jorgensd/adios4dolfinx/actions/runs/8179344989/job/22365266424?pr=92
while the DOLFINx C++ interface now uses the one cell version (which throws errors in `test_assembly` if not done correctly).

### DIFF
--- a/cpp/dolfinx/fem/FiniteElement.h
+++ b/cpp/dolfinx/fem/FiniteElement.h
@@ -330,29 +330,24 @@ public:
       return [this](std::span<U> data, std::span<const std::uint32_t> cell_info,
                     std::int32_t cell, int block_size)
       {
-        pre_apply_inverse_transpose_dof_transformation(
-            data, cell_info.subspan(cell, 1), block_size);
+        pre_apply_inverse_transpose_dof_transformation(data, cell_info[cell],
+                                                       block_size);
       };
     case doftransform::transpose:
       return [this](std::span<U> data, std::span<const std::uint32_t> cell_info,
-                    std::int32_t cell, int block_size)
-      {
-        pre_apply_transpose_dof_transformation(data, cell_info.subspan(cell, 1),
+                    std::int32_t cell, int block_size) {
+        pre_apply_transpose_dof_transformation(data, cell_info[cell],
                                                block_size);
       };
     case doftransform::inverse:
       return [this](std::span<U> data, std::span<const std::uint32_t> cell_info,
-                    std::int32_t cell, int block_size)
-      {
-        pre_apply_inverse_dof_transformation(data, cell_info.subspan(cell, 1),
-                                             block_size);
+                    std::int32_t cell, int block_size) {
+        pre_apply_inverse_dof_transformation(data, cell_info[cell], block_size);
       };
     case doftransform::standard:
       return [this](std::span<U> data, std::span<const std::uint32_t> cell_info,
-                    std::int32_t cell, int block_size) {
-        pre_apply_dof_transformation(data, cell_info.subspan(cell, 1),
-                                     block_size);
-      };
+                    std::int32_t cell, int block_size)
+      { pre_apply_dof_transformation(data, cell_info[cell], block_size); };
     default:
       throw std::runtime_error("Unknown transformation type");
     }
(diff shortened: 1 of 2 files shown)


garth-wells (commented):
It would be better to fix this (1) at the 'top' (in the nanobind wrapper) or (2) at the 'bottom' (in Basix). This PR changes it in the 'middle'. The C++ code is more complicated with no benefit to the C++ layer.

jorgensd:
> It would be better to fix this (1) at the 'top' (in the nanobind wrapper) or (2) at the 'bottom' (in Basix). This PR changes it in the 'middle'. The C++ code is more complicated with no benefit to the C++ layer.

@mscroggs what do you think? Should we move this to basix?

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -468,7 +468,7 @@ class FiniteElement
   /// @param[in,out] data The data to be transformed. This data is flattened
   /// with row-major layout, shape=(num_dofs, block_size)
   /// @param[in] cell_permutation Permutation data for the cell
-  /// @param[in] block_size The block_size of the input data
+  /// @param[in] block_size The block_size of the input data per cell
```
jorgensd, line 471:
```suggestion
  /// @param[in] block_size The block_size of the input data
```

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -533,7 +533,7 @@ class FiniteElement
   ///
   /// @param[in,out] data The data to be transformed. This data is flattened
   /// with row-major layout, shape=(num_dofs, block_size)
-  /// @param[in] cell_permutation Permutation data for the cell
+  /// @param[in] cell_permutation Permutation data for the cells
```
jorgensd, line 536:
```suggestion
  /// @param[in] cell_permutation Permutation data for the cell
```

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -548,7 +548,7 @@ class FiniteElement
...
```
jorgensd, line 551:
```suggestion
  /// @param[in] cell_permutation Permutation data for the cell
```

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -565,7 +565,7 @@ class FiniteElement
   /// @param[in,out] data The data to be transformed. This data is flattened
   /// with row-major layout, shape=(num_dofs, block_size)
   /// @param[in] cell_permutation Permutation data for the cell
-  /// @param[in] block_size The block_size of the input data
+  /// @param[in] block_size The block_size of the input datas
```
jorgensd, line 568:
```suggestion
  /// @param[in] block_size The block_size of the input data
```

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -580,7 +580,7 @@ class FiniteElement
   ///
   /// @param[in,out] data The data to be transformed. This data is flattened
   /// with row-major layout, shape=(num_dofs, block_size)
-  /// @param[in] cell_permutation Permutation data for the cell
+  /// @param[in] cell_permutation Permutation data for the cells
```
jorgensd, line 583:
```suggestion
  /// @param[in] cell_permutation Permutation data for the cell
```

jorgensd:
> It would be better to fix this (1) at the 'top' (in the nanobind wrapper) or (2) at the 'bottom' (in Basix). This PR changes it in the 'middle'. The C++ code is more complicated with no benefit to the C++ layer.

I've moved the fix for this into the nanobind wrapper.
Currently running tests at: https://github.com/jorgensd/adios4dolfinx/actions/runs/8185238997/job/22409020117

Issue #3061, opened 2024-02-20
Reopening the discussion regarding `MPI_Comm_dup`


uvilla:
Hi there,

I know this issue has already been discussed in several threads, but I would like to understand better the rationale behind using `MPI_Comm_dup` when wrapping the MPI comm in dolfinx.

With a student of mine, we are finally (it was overdue) porting hIPPYlib (https://hippylib.github.io/) to `dolfinx`, and we are quickly facing the _Too many communicators_ issue.

Our code does not store any long list of `dlx.fem.Function` but we do create quite a lot of temporary `dlx.fem.Function` and `dlx.fem.form` instances in our code. All of these should be destroyed when the function returns, but it seems that is not the case for us.

While this may be a problem of the garbage collection in python, it is still an issue that affects the usability of the library.

So my questions are:
1. Is duplication of the MPI Comm truly needed?
2. Could an option of *not* to duplicate the communicator be exposed to the final user, without having to recompile the library?

_Originally posted by @uvilla in https://github.com/FEniCS/dolfinx/issues/2308#issuecomment-1954583103_

uvilla:
I tried to write a simple code to reproduce the issue.
Unfortunately, I did not observe the MPI Comm duplication issue, but I can see some memory leakage.

```
from mpi4py import MPI
import dolfinx as dlx
import dolfinx.fem.petsc
import ufl

def run(Vh):
    u = dlx.fem.Function(Vh)
    v = ufl.TestFunction(Vh)
    form = dlx.fem.form( ufl.inner(u,v)*ufl.dx )
    vec = dlx.fem.petsc.assemble_vector(form)    


if __name__=='__main__':
    comm = MPI.COMM_WORLD
    rank = comm.rank

    mesh = dlx.mesh.create_unit_square(MPI.COMM_WORLD, 8, 8, dlx.mesh.CellType.quadrilateral)
    V = dlx.fem.FunctionSpace(mesh, ("Lagrange", 1 ) )

    for i in range(2**32-1):
        if (i % 10 == 0) and (rank == 0): print(i)
     
```

I run this snippet using 8 processors. At the beginning the memory allocation was about ~300MB per process and after about 1 hour of running is ~3.5GB per process.

I'm running FeniCS through conda on an Intel MacOS laptop. (python-3.10, `fenics-libdolfinx 0.7.0 h2cc1089_100`

[...]

PR #3092, opened 2024-03-07
Let dlx.la.Vector manage the scope of the petsc4py wrapper (#1)

uvilla:
This PR allows for a `dlx.la.Vector` (rather than a `dlx.fem.Function`) to manage the memory and the scope of the `petsc4py.PETSc.vec` object that wraps the underlying data.

The implementation of the property `dlx.fem.Function.vector` was also modified to delegate the creation of the wrapper to the underlying `dlx.la.Vector`.

All changes are backward compatible.

See also Issue #3061
@jorgensd 
@garth-wells
@jhale

### DIFF
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -498,7 +498,8 @@ class Function(ufl.Coefficient):
         warnings.warn(
             "dlx.fem.Function.vector is deprecated.\n"
             "Please use dlx.fem.Function.x.petsc_vec "
-            "to access the underlying petsc4py wrapper"
+            "to access the underlying petsc4py wrapper",
+            DeprecationWarning
         )
         return self.x.petsc_vec
 
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -369,7 +369,7 @@ class Function(ufl.Coefficient):
         # Store DOLFINx FunctionSpace object
         self._V = V
 
-        # Store casting to la.Vector
+        # Store Python wrapper around the underlying Vector
         self._x = la.Vector(self._cpp_object.x)
 
     @property
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -7,8 +7,8 @@
 
 from __future__ import annotations
 
-import warnings
 import typing
+import warnings
 from functools import singledispatch
 
 import numpy as np
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -498,7 +498,7 @@ class Function(ufl.Coefficient):
         warnings.warn(
             "dlx.fem.Function.vector is deprecated.\n"
             "Please use dlx.fem.Function.x.petsc_vec "
-            "to access the underlining petsc4py wrapper"
+            "to access the underlying petsc4py wrapper"
         )
         return self.x.petsc_vec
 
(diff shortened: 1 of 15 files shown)


python/dolfinx/fem/function.py
```diff
@@ -373,9 +369,8 @@ def functiontype(dtype):
         # Store DOLFINx FunctionSpace object
         self._V = V
 
-    def __del__(self):
-        if self._petsc_x is not None:
-            self._petsc_x.destroy()
+        # Store casting to la.Vector
```
garth-wells, line 372:
```suggestion
        # Store Python wrapper around the underlying Vector
```

python/dolfinx/fem/function.py
```diff
@@ -485,7 +480,7 @@ def copy(self) -> Function:
     @property
     def x(self) -> la.Vector:
         """Vector holding the degrees-of-freedom."""
-        return la.Vector(self._cpp_object.x)  # type: ignore
+        return self._x  # type: ignore
```
garth-wells, line 483:
```suggestion
        return self._x
```
Note sure about this change - mypy will tell us if removing the 'ignore' is ok or nor.

python/dolfinx/fem/function.py
```diff
@@ -485,7 +480,7 @@ def copy(self) -> Function:
...
 
     @property
     def vector(self):
```
garth-wells, line 486: We should probably deprecate this method.

python/dolfinx/la.py
```diff
@@ -245,6 +250,18 @@ def array(self) -> np.ndarray:
         """Local representation of the vector."""
         return self._cpp_object.array
 
+    @property
+    def vector(self):
```
garth-wells, line 254: I think we need a more descriptive name. Maybe `petsc_vec`?

garth-wells (changes requested):
Looks good - a few small comments.

python/dolfinx/la.py
uvilla: I also prefer `petsc_vec`. I originally used `vector` since this was the method used in the `fem.Function` class.
I will make the update.

python/dolfinx/fem/function.py
uvilla: Yes, we can deprecate the method. However, I would like to get consensus on this, since it will require to update all the python examples using `Function.vector` to use `Function.x.petsc_vec`

jorgensd: I think we should put in a deprecation-warning, and remove it with the release of 0.9.x

uvilla: Sounds good. I'll do that and replace all occurrences of `u.vector` to `u.x.petsc_vec`.

uvilla: @garth-wells @jorgensd : This has been addressed. All occurrences of `u.vector` were replaced with  `u.x.petsc_vec` and a deprecation warning was added to `dlx.fem.Function.vector`.

python/dolfinx/la.py
uvilla: @garth-wells : This has been renamed as recommended.

uvilla:
@garth-wells @jhale @jorgensd : Requested changes have been addressed.
I know this is the 11-th hour, but it would be much appreciated if this PR can make the  0.8 release. 

We are about to opensource `hippylibx`---the porting  of [hippylib](https://hippylib.github.io/) to `dolfinx`---and this new wrapper would be extremely handy.

Thank you

Issue #3097, opened 2024-03-09
Submesh creation fails on 6 processes with ghostmode shared facet


jorgensd:
### Summarize the issue

Minimal example below. Runs with less than 6 processes, and if ghostmode is none

### How to reproduce the bug

Runing test below.

### Minimal Example (Python)

```Python
import dolfinx
import numpy as np
from mpi4py import MPI

mesh = dolfinx.mesh.create_unit_square(
    MPI.COMM_WORLD, 4, 2, ghost_mode=dolfinx.mesh.GhostMode.shared_facet)


# Create connectivities required for defining integration entities
tdim = mesh.topology.dim

# Get number of cells on process
cell_map = mesh.topology.index_map(tdim)
num_cells = cell_map.size_local + cell_map.num_ghosts

# Create markers for each size of the interface
cell_values = np.ones(num_cells, dtype=np.int32)
cell_values[dolfinx.mesh.locate_entities(
    mesh, tdim, lambda x: x[0] <= 0.5 + 1e-13)] = 2
ct = dolfinx.mesh.meshtags(mesh, tdim, np.arange(
    num_cells, dtype=np.int32), cell_values)

submesh, sub_cell_to_parent, sub_vertex_to_parent, _ = dolfinx.mesh.create_submesh(mesh, ct.dim, ct.find(2))
```

### Version

main branch

jpdean:
On my old branches, this runs fine. I'll have a look into it

jpdean:
This problem was caused when NBX was removed from `create_sub_index_map`. The submap destination ranks are now being computed incorrectly in certain corner cases. I'll make a PR with a fix soon.

PR #3099, opened 2024-03-11
Fix sub index map destination rank computation

jpdean:
The method for computing submap destination ranks in `main` does not work in general (see [this](https://github.com/FEniCS/dolfinx/issues/3097) issue). This is because, when a index changes owner in the submap and there are multiple possible owners, the new owning process does not know to include the other processes as destination ranks.

This PR fixes the issue by using NBX to compute the submap destination ranks from the submap source ranks and adds a test that fails on `main`.

In a future PR, I will add an extra communication step to replace NBX, which should be cheaper.

### DIFF
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -205,7 +205,7 @@ compute_submap_indices(const IndexMap& imap,
   // `recv_indices` will necessarily be in `indices` on this process, and thus
   // other processes must own them in the submap.
 
-  std::vector<int> recv_owners(send_disp.back()), submap_dest;
+  std::vector<int> recv_owners(send_disp.back());
   const int rank = dolfinx::MPI::rank(imap.comm());
   {
     // Flag to track if the owner of any indices have changed in the submap
@@ -233,11 +233,7 @@ compute_submap_indices(const IndexMap& imap,
           // this process remains its owner in the submap. Otherwise,
           // add the process that sent it to the list of possible owners.
           if (is_in_submap[idx_local])
-          {
             global_idx_to_possible_owner.push_back({idx, rank});
-            // Add the sending process as a destination rank in the submap
-            submap_dest.push_back(dest[i]);
-          }
           else
           {
             owners_changed = true;
@@ -252,10 +248,6 @@ compute_submap_indices(const IndexMap& imap,
 
     std::sort(global_idx_to_possible_owner.begin(),
               global_idx_to_possible_owner.end());
-    std::sort(submap_dest.begin(), submap_dest.end());
-    submap_dest.erase(std::unique(submap_dest.begin(), submap_dest.end()),
-                      submap_dest.end());
-    submap_dest.shrink_to_fit();
 
     // Choose the submap owner for each index in `recv_indices`
     std::vector<int> send_owners;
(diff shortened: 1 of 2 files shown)


python/test/unit/common/test_index_map.py
```diff
@@ -192,3 +192,45 @@ def test_create_submap_owner_change():
...
     assert np.array_equal(global_indices, np.arange(comm.rank * 2, comm.rank * 2 + 3))
+
+
+def test_sub_index_map_multiple_possible_owners():
+    """Check that creating a submap doesn't crash when an index need to change owner and
+    there are multiple possible new owners"""
+    comm = MPI.COMM_WORLD
+
+    if comm.size < 3:
+        return
```
jorgensd, line 203:
```suggestion
    if comm.size < 3:
        pytest.skip("Test only work with 3 or more processes")
```

jhale: We now have a weekly CI run on macOS that uses three. 

It might be worth enabling macOS CI on all PR commits as the 3 case has sometimes shown issues where 2 did not.

jhale: Does negate your point of course.

garth-wells: > How often do we run CI with more than 2 processes?

We run on CircleCI with 3 MPI ranks.

jorgensd: Another thing we could do is to use ipythonparallel to create specific parallel scenarios.
I use this in adios4dolfinx, example at:
https://github.com/jorgensd/adios4dolfinx/blob/main/tests/conftest.py#L12-L17
https://github.com/jorgensd/adios4dolfinx/blob/main/tests/test_original_checkpoint.py#L310-L314

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -377,6 +369,12 @@ compute_submap_indices(const IndexMap& imap,
     submap_ghost = std::move(submap_ghost1);
   }
 
+  // Compute submap destination ranks
+  // FIXME Remove call to NBX
```
garth-wells, line 373: How much work to fix this now? We need to guard against unnecessary NBX calls.

jpdean: It will need a bit of thought to do it in a tidy and efficient way and to check for corner cases. I'm probably not going to have time to look at it until next week. I think it's probably best to merge this fix in the meantime and make an issue to remove NBX, but it's up to you

garth-wells: Could you create an issue so it doesn't get forgotten?

jpdean: Done, see https://github.com/FEniCS/dolfinx/issues/3100

jorgensd: > How much work to fix this now? We need to guard against unnecessary NBX calls.

As far as I've seen, the NBX calls take very little time compared to almost any other function we have in DOLFINx, so I wouldn't be too worried. But in general I agree that if we can avoid it by using existing comms we should.

garth-wells (approved):

cpp/dolfinx/common/IndexMap.cpp
garth-wells: > > How much work to fix this now? We need to guard against unnecessary NBX calls.
> 
> As far as I've seen, the NBX calls take very little time compared to almost any other function we have in DOLFINx, so I wouldn't be too worried. But in general I agree that if we can avoid it by using existing comms we should.

It is a performance concern at scale with a large number of MPI ranks. NBX should be used only when there is no reasonable alternative.

garth-wells:
Closes #3097

PR #3106, opened 2024-03-17
Change `compute_graph_edges_nbx` to `compute_graph_edges`

chrisrichardson:
NBX doesn't seem to work well at large scale. This just makes a general interface, so we can try different implementations, currently using a compiler flag: `-DUSE_NBX` or `-DUSE_PCX` etc. 
Default is to use gather-scatter, which seems to be fastest on some machines.
If we can switch between implementations, we can benchmark more easily.


chrisrichardson:
Could also add `MPI_dist_graph_create` for comparison.

garth-wells:
> NBX doesn't seem to work well at large scale. 

Caveat is that this has been tested at scale on one machine only, and that machine is known to have serious MPI and networking issues.

chrisrichardson:
> Caveat is that this has been tested at scale on one machine only, and that machine is known to have serious MPI and networking issues.

Yes, maybe we need more data.

PR #3108, opened 2024-03-20
Refactor `la.Vector.norm`

### LINKED ISSUE #3103: Refactor Python wrapper method `la.Vector.norm` into function

I would like to move the Python wrapper method `la.Vector.norm`

https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/wrappers/la.cpp#L51-L55

to a function defined in the `declare_functions` function

https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/wrappers/la.cpp#L179

To better match the C++ structure from the Python interface the norm method

https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/la.py#L261-L270

would be refactored into a standalone function of the form

```
def norm(x: Vector, type: _cpp.la.Norm = _cpp.la.Norm.l2) -> np.floating:
    ...
```

This would further permit declaring `la.Vector`s containing `int32_t`, `int64_t` etc, which is very useful for, e.g., mesh associated data. E.g:

```Python
from mpi4py import MPI
import numpy as np
import dolfinx

def pprint(*msg):
    print(f"[{MPI.COMM_WORLD.rank}]: {', '.join(map(str, msg))}", flush=True)

Nx, Ny = 2, 2
mesh = dolfinx.mesh.create_rectangle(
    MPI.COMM_WORLD, [[0.0, 0.0], [1.0, 1.0]], [Nx, Ny],
    cell_type=dolfinx.mesh.CellType.quadrilateral)

im_v = mesh.topology.index_map(0)
v = dolfinx.cpp.la.Vector_int32(im_v, 1)
v.array[:im_v.size_local] = np.arange(*im_v.local_range)
pprint("before:", v.array)
v.scatter_forward()
pprint("after:", v.array)
```
yields
```bash
[0]: before:, [0 1 2 3 0 0 0 0 0]
[1]: before:, [4 5 6 7 8 0 0 0 0]
[1]: after:, [4 5 6 7 8 2 0 1 3]
[0]: after:, [0 1 2 3 5 7 4 6 8]
```

### Suggested user interface

_No response_

nate-sime:
Closes #3103.

Refactor `la.Vector.norm` into a function to reflect the C++ layer.
Add implementation of integer types for `la.Vector`.
Redesign the `la.Vector` tests to parametrize over `dtype` as well as `xfail` on unsupported (or not yet implemented) norm types.

### DIFF
--- a/cpp/dolfinx/la/Vector.h
+++ b/cpp/dolfinx/la/Vector.h
@@ -262,7 +262,8 @@ auto squared_norm(const V& a)
 /// Compute the norm of the vector
 /// @note Collective MPI operation
 /// @param x A vector
-/// @param type Norm type (supported types are \f$L^2\f$ and \f$L^\infty\f$)
+/// @param type Norm type (supported types are \f$L^1\f$, \f$L^2\f$
+/// and \f$L^\infty\f$)
 template <class V>
 auto norm(const V& x, Norm type = Norm::l2)
 {
--- a/cpp/dolfinx/la/Vector.h
+++ b/cpp/dolfinx/la/Vector.h
@@ -262,8 +262,7 @@ auto squared_norm(const V& a)
 /// Compute the norm of the vector
 /// @note Collective MPI operation
 /// @param x A vector
-/// @param type Norm type (supported types are \f$L^1\f$, \f$L^2\f$
-/// and \f$L^\infty\f$)
+/// @param type Norm type
 template <class V>
 auto norm(const V& x, Norm type = Norm::l2)
 {
--- a/python/dolfinx/wrappers/la.cpp
+++ b/python/dolfinx/wrappers/la.cpp
@@ -48,11 +48,6 @@ void declare_objects(nb::module_& m, const std::string& type)
       .def(nb::init<const dolfinx::la::Vector<T>&>(), nb::arg("vec"))
       .def_prop_ro("dtype", [](const dolfinx::la::Vector<T>&)
                    { return dolfinx_wrappers::numpy_dtype<T>(); })
-      .def(
-          "norm",
-          [](const dolfinx::la::Vector<T>& self, dolfinx::la::Norm type)
-          { return dolfinx::la::norm(self, type); },
-          "type"_a = dolfinx::la::Norm::l2)
       .def_prop_ro("index_map", &dolfinx::la::Vector<T>::index_map)
       .def_prop_ro("bs", &dolfinx::la::Vector<T>::bs)
       .def_prop_ro(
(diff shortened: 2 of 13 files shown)


jorgensd (approved):
Looks good to me, even if it means that we change the API:P

Only request is to add support for l1 and l-inf norms for integers, as they are supported. "l2" is clearly not supported.

I guess if @garth-wells or @IgorBaratta  or @chrisrichardson has any opinions on this please chip in:)

nate-sime:
Regarding supporting norms of vectors of integral types, I'm having a little trouble coming up with a neat way to deduce the underlying type of a vector of `complex` whilst also handling those integral types. The outline for this is clearly in `dolfinx::scalar_value_type_t`; however, this would either require another similar definition with carefully chosen name, or expansion of `concept scalar` which I don't believe fits in with the scientific computing philosophy. Perhaps this could be approached in a future pull request?

IgorBaratta (approved):

garth-wells (commented):
I have some comments - will post tomorrow.

cpp/dolfinx/la/Vector.h
```diff
@@ -262,7 +262,8 @@ auto squared_norm(const V& a)
 /// Compute the norm of the vector
 /// @note Collective MPI operation
 /// @param x A vector
-/// @param type Norm type (supported types are \f$L^2\f$ and \f$L^\infty\f$)
+/// @param type Norm type (supported types are \f$L^1\f$, \f$L^2\f$
```
garth-wells, line 265: Best to remove to norm types. The available types can be found via the link to `Norm`.

python/dolfinx/wrappers/la.cpp
```diff
@@ -178,6 +173,11 @@ void declare_objects(nb::module_& m, const std::string& type)
 template <typename T>
 void declare_functions(nb::module_& m)
 {
+  m.def(
+        "norm",
+        [](const dolfinx::la::Vector<T>& x, dolfinx::la::Norm type)
+        { return dolfinx::la::norm(x, type); },
+        "vector"_a, "type"_a = dolfinx::la::Norm::l2);
```
garth-wells, line 180: Remove the default value here - it's already set in the Python wrapper.

Best to avoid default values in the nanobind layer the reduce the magic.

python/dolfinx/wrappers/la.cpp
```diff
@@ -285,6 +285,11 @@ void la(nb::module_& m)
           nb::rv_policy::reference_internal);
 
   // Declare objects that are templated over type
+  declare_objects<std::int8_t>(m, "int8");
```
garth-wells, line 288: Do we really want all these types at this stage? There is a compile time cost.

Does the MPI scatter support each the types?

garth-wells (changes requested):
Looks good - just a micro change suggestion.

python/dolfinx/wrappers/la.cpp
nate-sime: Good point. Looks like the types I've added in this PR are supported: https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/common/MPI.h#L273-L304, which are used in `Scatterer`.

Perhaps I went overboard with useful types and I agree regarding compile times. Could we go with `int32`, `int64` and something to efficiently represent a true/false? I can discuss with @IgorBaratta whether that be `int8` or `bool`?

garth-wells: I would suggest `int8`. Seems that NumPy uses 1 byte for `bool`, so no space saving with `bool` over `int8`.

garth-wells (approved):

PR #3115, opened 2024-03-27
Replace NumPy's logical_and and logical_or

michalhabera:
Improves readability, especially for nested conditionals.

### DIFF
--- a/python/demo/demo_biharmonic.py
+++ b/python/demo/demo_biharmonic.py
@@ -155,9 +155,10 @@ V = fem.functionspace(msh, ("Lagrange", 2))
 facets = mesh.locate_entities_boundary(
     msh,
     dim=1,
-    marker=lambda x: np.logical_or.reduce(
-        (np.isclose(x[0], 0.0), np.isclose(x[0], 1.0), np.isclose(x[1], 0.0), np.isclose(x[1], 1.0))
-    ),
+    marker=lambda x: np.isclose(x[0], 0.0)
+    | np.isclose(x[0], 1.0)
+    | np.isclose(x[1], 0.0)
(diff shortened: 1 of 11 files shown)


garth-wells (approved):

jhale:
The new code is correct and easier to read. However I've seen users replacing logical operator with a bitwise operator as 'they do the same thing' before, with bad results!

nate-sime:
@jhale do you have an example of where you'd get an unexpected result? I've been using the style introduced by this PR with no issues.

jhale:
https://numpy.org/doc/1.26/user/numpy-for-matlab-users.html

Yes I've seen people attempt bitwise operations on integer arrays.

There are some additional pitfalls especially for users coming from MATLAB. My view is that np.logical_* is the correct expression of the programmer's intent in Python. It's unfortunate there isn't a operator version of the logical numpy ops.

I'm confident this change will cause issues/confusion for non-expert users.

jhale:
I'm ok for one or two of the demos which needs this most to use this style, but please stick a comment there warning that this is the bitwise op!

michalhabera:
The PR uses `|` and `&` exclusively on boolean array, where bitwise and logical are equivalent. It is also hard to do something dangerous in our lambda markers, as bitwise ops work only on bool and int. So it would be better to show specific example where these ops are dangerous in the context of geometrical markers.
But whatever the decision is, I'd prefer a consistent approach.

jhale:
Everyone in this thread understands that logical operations are equivalent with bitwise operations on arrays of bools.

But do users reading our introductory demos understand that? A Python callable being passed is already difficult conceptually for many.

logical_* is clear, explicit and always has correct meaning.

Issue #1253, opened 2020-12-13
Support Meshtag IO for more types (only `int32_t` is supported at present)


jorgensd:
The meshtag IO is limited to int32, while the meshtags support 
int8, int32, int64 and double. Is there a particular reason for this?

michalhabera:
I do not see any. Just `XDMFFile::write_meshtags` must be made templated and header-only then. With proper `write_meshtags_int32`, `write_meshtags_int8`, etc generated in pybind and automatically dispatched based on datatype in python public interface.

garth-wells:
Simplicity. Is there a compelling reason/need to support `int8_t` and `int64_t` in IO?

Is there a current use case for double?

jorgensd:
> Simplicity. Is there a compelling reason/need to support `int8_t` and `int64_t` in IO?
> 
> 
> 
> Is there a current use case for double? 

I was currently planning to use the double one as a workaround for visualizing DG 0 functions, as we are lacking this functionality.

garth-wells:
@jorgensd do we have this now?

jorgensd:
@garth-wells we do not

chrisrichardson:
Is this still needed? It would be fairly easy to add more types (which? just `double`?)

jorgensd:
We should at least support float/double, as some use-cases read material parameters from the mesh .

Ref:

https://fenicsproject.discourse.group/t/i-o-from-xdmf-hdf5-files-in-dolfin-x/3122/48?u=dokken

Where scalar data is actually a meshtag double object.

garth-wells:
We should add more, but need to clean up the HDF5 interface first to better handle different types.

garth-wells:
Closing since we will deprecate XMDF in favour of VTKHDF5, see #3674.

Can be worked into VTKHDF5 design.

PR #3119, opened 2024-04-02
Template `distribute_entity_data` to support multiple data-types

jorgensd:
First step towards resolving https://github.com/FEniCS/dolfinx/issues/1253
Second step would be to template `XDMFFile.read_meshtags`. This is fairly trivial, but I figured we should agree on the core changes before doing that change.

Currently tested with:
```python
from mpi4py import MPI
import numpy as np
import dolfinx

num_vertices = mesh.topology.index_map(0).size_local
entities = np.arange(num_vertices, dtype=np.int32)


def distribute_data(dtype):
    values = entities.astype(dtype).copy()
    if dtype == np.float64:
        values -= 0.6
    return dolfinx.io.distribute_entity_data(mesh, 0, entities.reshape(-1, 1).copy(), values[::-1].copy())


a = distribute_data(np.int32)

b = distribute_data(np.float64)
c = distribute_data(np.complex128)
print(a, b, c)
```
I've removed the unnecessary mesh geometry templating of this function, and made the Python-binding more low-level (improving the Python layer of `distribute_entity_data` to work with the new binding). The PR also has documentation improvements of this function.

This would enabling reading vertex, facet and cell data that is single or double precision, which is the case in many biomedical applications (where one for instance have voxel data, or material parameters).

The major difference in the code is that the communication of entity indices and values are split into two separate neighbourhood alltoallv calls. 

No major changes in the actual functional structure of the code.


jorgensd:
@massimiliano-leoni could you test this?:)

python/dolfinx/wrappers/io.cpp
```diff
@@ -349,7 +360,10 @@ void io(nb::module_& m)
   declare_vtx_writer<float>(m, "float32");
   declare_vtx_writer<double>(m, "float64");
 
-  declare_real_types<float>(m);
-  declare_real_types<double>(m);
+  declare_data_types<std::int32_t>(m);
+  // declare_data_types<float>(m);
```
jorgensd, line 364:
```suggestion
  declare_data_types<float>(m);
```

cpp/dolfinx/io/xdmf_utils.h
```diff
@@ -24,6 +26,13 @@
 #include <utility>
 #include <vector>
 
+namespace
```
garth-wells, line 29: Anonymous namespaces should never appear in header files - should only appear in `.cpp` files.

garth-wells (changes requested):
I would suggest leaving the code in the `.cpp` file and explicitly instantiating for the small number of types that we'll use (at this stage). It's preferable to keep code in the implementation file. This is what we do elsewhere for the mesh.

cpp/dolfinx/io/xdmf_utils.h
```diff
@@ -15,6 +15,8 @@
 #include <concepts>
 #include <dolfinx/common/MPI.h>
 #include <dolfinx/common/types.h>
+#include <dolfinx/fem/ElementDofLayout.h>
+#include <dolfinx/mesh/Topology.h>
```
jorgensd, line 19:
```suggestion
```

jorgensd:
@garth-wells any further changes you would like to see in this PR?

garth-wells (approved):

PR #3124, opened 2024-04-15
Skip test if adios2 python module built without MPI.

jhale:
It would also be good to not build against ADIOS2 if it wasn't built against MPI...


python/test/unit/io/test_adios2.py
```diff
@@ -291,6 +291,8 @@ def partitioner(comm, nparts, local_graph, num_ghost_nodes):
 def test_vtx_reuse_mesh(tempdir, dim, simplex, reuse):
     "Test reusage of mesh by VTXWriter"
     adios2 = pytest.importorskip("adios2")
+    if not adios2.is_built_with_mpi:
```
drew-parsons, line 294: Keep in mind `adios2.is_built_with_mpi` is new, i.e. not available in ADIOS2 2.8.1.  2.10.0 has it (released two weeks ago)

cpp/CMakeLists.txt
```diff
@@ -375,6 +375,13 @@ elseif(DOLFINX_ENABLE_ADIOS2)
   find_package(ADIOS2 2.8.1)
```
drew-parsons, line 375: might want to bump to ADIOS2 2.10.0 for `adios2.is_built_with_mpi` (see below)
I suggest not pushing this into dolfinx 0.8 since ADIOS2 2.10 has only just been released.

python/test/unit/io/test_adios2.py
drew-parsons: Alternatively could place the test is a try block and catch the `AttributeError` from adios2 2.8 or 2.9.  Perhaps that could be more helpful for the dolfinx 0.8 release

python/test/unit/io/test_adios2.py
```diff
@@ -270,6 +270,161 @@ def partitioner(comm, nparts, local_graph, num_ghost_nodes):
             dest = np.full(len(cells), comm.rank, dtype=np.int32)
             return adjacencylist(dest)
 
+<<<<<<< HEAD
```
drew-parsons, line 273: This line looks like a git conflict marker that shouldn't be here (needs reconciliation after merging from HEAD?)

drew-parsons (changes requested):

jorgensd:
@drew-parsons We have decided that this test will only run if the user supplies `adios2>=2.10` as this is not a requirement in the actual code base.

cpp/CMakeLists.txt
jorgensd: We only require this in one test in Python, so we can keep the lower limit (2.8.1) in C++.

jorgensd (approved):

drew-parsons (approved):
Makes sense to check `adios2.is_built_with_mpi` if the adios2 serial build can't be used.

python/test/unit/io/test_adios2.py
```diff
@@ -312,19 +312,17 @@ def test_vtx_reuse_mesh(self, tempdir, dim, simplex, reuse):
...
-        )  # For mesh variables the step count is 1 if reuse else number of writes
-
-        # backwards compatibility adios2 < 2.10.0
-        try:
-            adios_file = adios2.open(str(filename), "r", comm=mesh.comm, engine_type="BP4")
-        except AttributeError:
-            # adios2 >= v2.10.0
+        target_mesh = 1 if reuse else 3
+        # For mesh variables the step count is 1 if reuse else number of writes
+        if hasattr(adios2, "is_built_with_mpi"):
```
garth-wells, line 317: This is overcomplicated. Looking at the ADIOS2 code, the attribute `is_built_with_mpi` is always defined. If it's not in earlier versions, put ADIOS2 version check in a marker for the whole test.

garth-wells (changes requested):

python/test/unit/io/test_adios2.py
jorgensd: Version checks are not trivial without external libraries (i think), Ref https://til.simonwillison.net/python/comparing-version-numbers

drew-parsons: True.  I wasn't sure how easy it would be to set up which is why I didn't mention it.
mpi4py uses [versioned conditions](https://github.com/mpi4py/mpi4py/blob/fd1abfeb86305623e4381801f3a99b70f6505e51/test/test_spawn.py#L86) but that's implemented via [custom rules](https://github.com/mpi4py/mpi4py/blob/fd1abfeb86305623e4381801f3a99b70f6505e51/test/mpiunittest.py#L27). Similar h5py has [version conditions](https://github.com/h5py/h5py/blob/862eecefc72eb6941db3b370dff27a6e18bf6197/h5py/tests/test_file.py#L238), but on its own version provided as a tuple.

drew-parsons: That said, it could be simplified with 
```
@pytest.mark.skipif(not hasattr(adios2, "is_built_with_mpi"), "requires adios2>=2.10.0")
```

jorgensd: Thats not quite right as adios2 is only defined through importorskip

drew-parsons: importorskip would already skip the entire test file though.
Then too, importorskip has its own [version parsing](https://docs.pytest.org/en/6.2.x/skipping.html#skipping-on-a-missing-import-dependency)
```
adios2 = pytest.importorskip("adios2", minversion="2.10.0")
```

jorgensd:
@garth-wells are you happy with the simplification? I can't get it to auto-merge without "force-dismissing" your review, which seems rather rude;)

garth-wells (approved):

PR #3127, opened 2024-04-15
Split linting and build steps in CI

garth-wells:
This partially preparation for adding a build step without PETSc.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -66,8 +66,6 @@ jobs:
         run: |
           find . -type f \( -name "*.cmake" -o -name "*.cmake.in" -o -name "CMakeLists.txt" \) | xargs cmake-format --check
 
-
-
   build:
     if: "!(contains(github.event.head_commit.message, '[ci skip]') || contains(github.event.head_commit.message, '[skip ci]'))"
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ccpp.yml
```diff
@@ -93,6 +66,38 @@ jobs:
         run: |
           find . -type f \( -name "*.cmake" -o -name "*.cmake.in" -o -name "CMakeLists.txt" \) | xargs cmake-format --check
 
+  build:
+    if: "!(contains(github.event.head_commit.message, '[ci skip]') || contains(github.event.head_commit.message, '[skip ci]'))"
```
jorgensd, line 70: I do not think this is needed any longer, ref: https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/

.github/workflows/ccpp.yml
```diff
@@ -93,6 +66,38 @@ jobs:
...
+    runs-on: ubuntu-latest
+    container: ghcr.io/fenics/test-env:current-openmpi
```
jhale, line 72: If you make `build` require `lint` then we can save pointless CI starts across the matrix of petsc architectures.

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

PR #3128, opened 2024-04-15
Make ADIOS tests compatible with 2.9 and 2.10

jhale:
We know we are not compatible with 2.10. Will do a minor DOLFINx release once ADIOS 2.10 is widely available.

### DIFF
--- a/python/test/unit/io/test_adios2.py
+++ b/python/test/unit/io/test_adios2.py
@@ -317,6 +317,7 @@ def test_vtx_reuse_mesh(tempdir, dim, simplex, reuse):
     try:
         adios_file = adios2.open(str(filename), "r", comm=mesh.comm, engine_type="BP4")
     except AttributeError:
+        # adios2 >= v2.10.0
         adios = adios2.Adios(comm=mesh.comm)
         io = adios.declare_io("TestData")
         io.set_engine("BP4")
--- a/python/test/unit/io/test_adios2.py
+++ b/python/test/unit/io/test_adios2.py
@@ -313,7 +313,14 @@ def test_vtx_reuse_mesh(tempdir, dim, simplex, reuse):
         1 if reuse else 3
     )  # For mesh variables the step count is 1 if reuse else number of writes
 
-    adios_file = adios2.open(str(filename), "r", comm=mesh.comm, engine_type="BP4")
+    # backwards compatibility adios2 < 2.10.0
+    try:
+        adios_file = adios2.open(str(filename), "r", comm=mesh.comm, engine_type="BP4")
+    except AttributeError:
+        adios = adios2.Adios(comm=mesh.comm)
+        io = adios.declare_io("TestData")
+        adios_file = adios2.Stream(io, str(filename), "r", mesh.comm)
+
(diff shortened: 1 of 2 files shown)


jorgensd:
> We know we are not compatible with 2.10. Will do a minor DOLFINx release once ADIOS 2.10 is widely available.

I'll add a try except in the test, give me a few minutes to build adios 2.10 locally to make a fix

jorgensd:
@jhale I added a patch to the test. As the only thing that doesn't work with adios2 is the test, i think it is better to patch the test for backwards compatibility than restricting build versions

garth-wells:
Why not apply fixes to be compatible with 2.10? How much work would it be?

python/test/unit/io/test_adios2.py
```diff
@@ -313,7 +313,14 @@ def test_vtx_reuse_mesh(tempdir, dim, simplex, reuse):
...
     )  # For mesh variables the step count is 1 if reuse else number of writes
 
-    adios_file = adios2.open(str(filename), "r", comm=mesh.comm, engine_type="BP4")
+    # backwards compatibility adios2 < 2.10.0
+    try:
+        adios_file = adios2.open(str(filename), "r", comm=mesh.comm, engine_type="BP4")
+    except AttributeError:
+        adios = adios2.Adios(comm=mesh.comm)
+        io = adios.declare_io("TestData")
+        adios_file = adios2.Stream(io, str(filename), "r", mesh.comm)
```
francesco-ballarin, line 322: So with 2.10 one can't say anymore that `engine_type="BP4"`? Or is the default?

francesco-ballarin (approved):

jorgensd:
> Why not apply fixes to be compatible with 2.10? How much work would it be?

Ive changed the scope of the PR. We are now compatible with 3.10, given that try except clause in the adios tests (only Python interface changed from 2.9 to 2.10).

garth-wells (approved):

python/test/unit/io/test_adios2.py
```diff
@@ -313,7 +313,15 @@ def test_vtx_reuse_mesh(tempdir, dim, simplex, reuse):
         1 if reuse else 3
     )  # For mesh variables the step count is 1 if reuse else number of writes
 
-    adios_file = adios2.open(str(filename), "r", comm=mesh.comm, engine_type="BP4")
+    # backwards compatibility adios2 < 2.10.0
```
garth-wells, line 316: Could you make clear with a comment which part of the try/except clause is for v2.9 and which is for v2.10?

python/test/unit/io/test_adios2.py
```diff
@@ -313,7 +313,15 @@ def test_vtx_reuse_mesh(tempdir, dim, simplex, reuse):
...
+    try:
+        adios_file = adios2.open(str(filename), "r", comm=mesh.comm, engine_type="BP4")
+    except AttributeError:
+        adios = adios2.Adios(comm=mesh.comm)
```
jorgensd, line 320:
```suggestion
        # adios2 >= v2.10.0
        adios = adios2.Adios(comm=mesh.comm)
```

PR #3129, opened 2024-04-15
Remove catch2 from docker images

jorgensd:
As it is v 2.13.8 (https://packages.ubuntu.com/jammy/catch2), which is not compatible with our tests, and will mess up: https://github.com/FEniCS/dolfinx/blob/main/cpp/test/CMakeLists.txt#L20-L31.

### DIFF
--- a/docker/Dockerfile.oneapi
+++ b/docker/Dockerfile.oneapi
@@ -32,7 +32,6 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
     libpugixml-dev \
     ninja-build \
     pkg-config \
-    catch2 \
     git \
     wget \
     zlib1g-dev && \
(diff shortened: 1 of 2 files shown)


jhale:
It's not perfect but it looks 'OK' to me - The Catch2 check in CMakeLists.txt has a lower bound on version 3, so it will always go to git to get a new version when running in our containers. 

Hopefully this will fix itself on the new Ubuntu, as Garth says.

jorgensd:
> It's not perfect but it looks 'OK' to me - The Catch2 check in CMakeLists.txt has a lower bound on version 3, so it will always go to git to get a new version when running in our containers.
> 
> Hopefully this will fix itself on the new Ubuntu, as Garth says.

I had issues with this using our containers yesterday. However, as Garth says, with ubuntu 24.04 this is hopefully resolved

PR #3134, opened 2024-04-16
Don't run build steps if linting fails

jhale:
Also removes deprecated [skip ci] check.


jhale:
This could be further improved by shifting the lint step out to it's own action and making all other actions depend on it.

PR #3136, opened 2024-04-16
Reuse Basix element instead of recreating from generated code

mscroggs:
Depends on https://github.com/FEniCS/basix/pull/796.

Will allow us to do https://github.com/FEniCS/ffcx/pull/681


cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -359,12 +244,6 @@ std::string FiniteElement<T>::signature() const noexcept
 }
 //-----------------------------------------------------------------------------
 template <std::floating_point T>
-mesh::CellType FiniteElement<T>::cell_shape() const noexcept
-{
-  return _cell_shape;
-}
```
mscroggs, line 365: _cell_shape was left uninitialised for most elements. I'm unsure whether removing it or making sure it is set is a better plan

cpp/dolfinx/fem/utils.h
```diff
@@ -335,14 +358,17 @@ Form<T, U> create_form_factory(
...
+    /*
+        ufcx_finite_element* ufcx_element = ufcx_form.finite_elements[i];
+        assert(ufcx_element);
+        if (std::string(ufcx_element->signature)
+            != spaces[i]->element()->signature())
+        {
+          throw std::runtime_error(
+              "Cannot create form. Wrong type of function space for argument.");
+        }
+    */
```
mscroggs, line 371: This needs removing

python/dolfinx/fem/function.py
```diff
@@ -582,6 +582,38 @@ class ElementMetaData(typing.NamedTuple):
     symmetry: typing.Optional[bool] = None
 
 
+def _create_dolfinx_element(
+    comm: _MPI.Intracomm,
+    cell_type: _cpp.mesh.CellType,
+    ufl_e: ufl.FiniteElementBase,
+    dtype: np.dtype,
+) -> typing.Union[_cpp.fem.FiniteElement_float32, _cpp.fem.FiniteElement_float64]:
+    # TODO: remove this function or move it to element.py?
```
mscroggs, line 591: this needs doing, or TODO needs deleting

cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -121,173 +69,36 @@ _extract_sub_element(const FiniteElement<T>& finite_element,
 
 //-----------------------------------------------------------------------------
 template <std::floating_point T>
-FiniteElement<T>::FiniteElement(const ufcx_finite_element& e)
-    : _signature(e.signature), _space_dim(e.space_dimension),
-      _reference_value_shape(e.reference_value_shape,
-                             e.reference_value_shape + e.reference_value_rank),
-      _bs(e.block_size), _is_mixed(e.element_type == ufcx_mixed_element)
+FiniteElement<T>::FiniteElement(const mesh::CellType cell_type,
+                                const std::span<geometry_type> points,
```
garth-wells, line 73:
```suggestion
                                std::span<const geometry_type> points,
```

cpp/dolfinx/fem/FiniteElement.cpp
garth-wells: We can remove, and add it back later if required.

cpp/dolfinx/fem/utils.h
mscroggs: See #3169

cpp/dolfinx/fem/utils.h
```diff
@@ -239,11 +239,34 @@ la::SparsityPattern create_sparsity_pattern(const Form<T, U>& a)
...
+ElementDofLayout create_element_dof_layout(const fem::FiniteElement<T>& element,
                                            const std::vector<int>& parent_map
-                                           = {});
+                                           = {})
+{
+  // Create subdofmaps and compute offset
+  std::vector<int> offsets(1, 0);
+  std::vector<dolfinx::fem::ElementDofLayout> sub_doflayout;
+
+  for (int i = 0; i < element.num_sub_elements(); ++i)
```
garth-wells, line 252: This code is quite hard to read. Could it be simplified?

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -109,7 +114,19 @@ class FiniteElement
   int reference_value_size() const;
 
   /// The reference value shape
-  std::span<const std::size_t> reference_value_shape() const;
+  std::span<const std::size_t> reference_value_shape() const noexcept;
+
+  /// The local DOFs associated with each subentity of the cell
+  std::vector<std::vector<std::vector<int>>> entity_dofs() const
```
garth-wells, line 120: Move the implementation to the cpp file.

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -109,7 +114,19 @@ class FiniteElement
...
+  {
+    return _entity_dofs;
+  }
+
+  /// The local DOFs associated with the closure of each subentity of the cell
+  std::vector<std::vector<std::vector<int>>> entity_closure_dofs() const
```
garth-wells, line 126: Move the implementation to the cpp file.

garth-wells (changes requested):

cpp/dolfinx/fem/utils.h
mscroggs: I've added some comments to explain the conditionals. The other options is to change it to this, but I don't like the additional code duplication it adds:

```c++
if (element.block_size() > 1)
  { 
    // blocked element
    std::shared_ptr<const fem::FiniteElement<T>> sub_e
        = element.sub_elements()[0];
    for (int i = 0; i < element.num_sub_elements(); ++i)
    {
      offsets.push_back(offsets.back() + 1);
      std::vector<int> parent_map_sub(sub_e->space_dimension());
      for (std::size_t j = 0; j < parent_map_sub.size(); ++j)
        parent_map_sub[j] = offsets[i] + element.block_size() * j;
      sub_doflayout.push_back(
          dolfinx::fem::create_element_dof_layout(*sub_e, parent_map_sub));
    }
  } else {
    // Mixed element or element with no sub-elements
    for (int i = 0; i < element.num_sub_elements(); ++i)
    {
      std::shared_ptr<const fem::FiniteElement<T>> sub_e
          = element.sub_elements()[i];
      offsets.push_back(offsets.back() + sub_e->space_dimension());
      std::vector<int> parent_map_sub(sub_e->space_dimension());
      for (std::size_t j = 0; j < parent_map_sub.size(); ++j)
        parent_map_sub[j] = offsets[i] + j;
      sub_doflayout.push_back(
          dolfinx::fem::create_element_dof_layout(*sub_e, parent_map_sub));
    }
  }
```

garth-wells (approved):

PR #3138, opened 2024-04-17
Fix dtype assumption on Debian 32-bit architectures

jhale:
The default conversion between a list of Python ints and an ndarray (without dtype) on 32-bit architectures uses 32-bit integers. This patch makes the dtype explicitly `int64`.


jhale:
Doesn't work.

PR #3140, opened 2024-04-17
Remove linting from CircleCI - covered by Actions

garth-wells:


jhale:
I'm not sure about the circleci change - will circleci stop running if lint fails on GitHub actions?

garth-wells:
> I'm not sure about the circleci change - will circleci stop running if lint fails on GitHub actions?

No, but why should it? Also, CircleCI is run only on PRs, which having normally run through the Actions lint check before being run on CircleCI.

Testing the same thing redundantly in multiple places added complexity for no test coverage benefit.

PR #3149, opened 2024-04-19
Dof transform name updates

garth-wells:
Requires https://github.com/FEniCS/basix/pull/807

### DIFF
--- a/cpp/dolfinx/fem/FiniteElement.h
+++ b/cpp/dolfinx/fem/FiniteElement.h
@@ -433,22 +433,15 @@ public:
     case doftransform::inverse_transpose:
       return [this](std::span<U> data, std::span<const std::uint32_t> cell_info,
                     std::int32_t cell, int block_size)
-      {
-        post_apply_inverse_transpose_dof_transformation(data, cell_info[cell],
-                                                        block_size);
-      };
+      { Tt_inv_post_apply(data, cell_info[cell], block_size); };
     case doftransform::transpose:
       return [this](std::span<U> data, std::span<const std::uint32_t> cell_info,
-                    std::int32_t cell, int block_size) {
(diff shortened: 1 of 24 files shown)


cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -417,8 +425,7 @@ class FiniteElement
         const std::function<void(std::span<U>, std::span<const std::uint32_t>,
                                  std::int32_t, int)>
             sub_function
-            = _sub_elements[0]->template dof_transformation_function<U>(
-                ttype);
+            = _sub_elements[0]->template dof_transformation_fn<U>(ttype);
```
garth-wells, line 428: @mscroggs This looks fishy (before and after the change). These lines are the only 'left-apply' functions in a function that is otherwise all 'right-apply'. What do you think it should be?

mscroggs: Most likely this:

```suggestion
            = _sub_elements[0]->template dof_transformation_right_fn<U>(ttype);
```

garth-wells: Hmm, with this change:

https://github.com/FEniCS/dolfinx/compare/garth/transform-name-updates...garth/transform-test

tests fail:

https://github.com/FEniCS/dolfinx/actions/runs/8764139051

Any ideas?

mscroggs: I'll take a closer look on Monday. I guess there's a good reason for not having the right that I need to re-work out and add a comment explaining

garth-wells: There are some more transforms that look fishy and there be some cancellation by two incorrect transforms - I'll make a separate PR to highlight.

PR #3163, opened 2024-04-23
Try tests on GitHub Actions with three MPI ranks

jhale:
Can keep the others on two for coverage? To discuss.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -171,7 +171,7 @@ jobs:
       - name: Run C++ unit tests (MPI)
         run: |
           cd build/test
-          mpiexec -np 3 ctest -V --output-on-failure -R unittests
+          mpirun -n 3 --map-by :OVERSUBSCRIBE ctest -V --output-on-failure -R unittests
       - name: Build and run C++ regression tests (serial and MPI (np=2))
         run: |
           cmake -G Ninja -DCMAKE_BUILD_TYPE=Developer -B build/demo/ -S cpp/demo/
@@ -196,12 +196,12 @@ jobs:
       - name: Run demos (Python, serial)
         run: python3 -m pytest -n=2 -m serial --durations=10 python/demo/test.py
       - name: Run demos (Python, MPI (np=3))
-        run: python3 -m pytest -m mpi --num-proc=3 python/demo/test.py
+        run: mpirun -n 3 --map-by :OVERSUBSCRIBE python3 -m pytest python/demo/test.py
 
       - name: Run Python unit tests (serial)
         run: python3 -m pytest -n=auto --durations=50 python/test/unit/
       - name: Run Python unit tests (MPI, np=3)
-        run: mpirun -np 3 python3 -m pytest python/test/unit/
+        run: mpirun -n 3 --map-by :OVERSUBSCRIBE python3 -m pytest python/test/unit/
 
       - name: Upload C++ documentation artifact
         uses: actions/upload-artifact@v4
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -196,9 +196,9 @@ jobs:
           echo '{ "cffi_extra_compile_args": ["-g0", "-O0" ] }' > ~/.config/dolfinx/dolfinx_jit_options.json
 
       - name: Run demos (Python, serial)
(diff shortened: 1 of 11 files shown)


jhale:
```
--------------------------------------------------------------------------
There are not enough slots available in the system to satisfy the 3
slots that were requested by the application:

  ctest

Either request fewer procs for your application, or make more slots
available for use.

A "slot" is the PRRTE term for an allocatable unit where we can
launch a process.  The number of slots available are defined by the
environment in which PRRTE processes are run:

  1. Hostfile, via "slots=N" clauses (N defaults to number of
     processor cores if not provided)
  2. The --host command line parameter, via a ":N" suffix on the
     hostname (N defaults to 1 if not provided)
  3. Resource manager (e.g., SLURM, PBS/Torque, LSF, etc.)
  4. If none of a hostfile, the --host command line parameter, or an
     RM is present, PRRTE defaults to the number of processor cores

In all the above cases, if you want PRRTE to default to the number
of hardware threads instead of the number of processor cores, use the
--use-hwthread-cpus option.

Alternatively, you can use the --map-by :OVERSUBSCRIBE option to ignore the
number of available slots when deciding the number of processes to
launch.
--------------------------------------------------------------------------
```

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

Strange as the public runners all have at least 3 processors.

jorgensd:
> ```
> --------------------------------------------------------------------------
> There are not enough slots available in the system to satisfy the 3
> slots that were requested by the application:
> 
>   ctest
> 
> Either request fewer procs for your application, or make more slots
> available for use.
> 
> A "slot" is the PRRTE term for an allocatable unit where we can
> launch a process.  The number of slots available are defined by the
> environment in which PRRTE processes are run:
> 
>   1. Hostfile, via "slots=N" clauses (N defaults to number of
>      processor cores if not provided)
>   2. The --host command line parameter, via a ":N" suffix on the
>      hostname (N defaults to 1 if not provided)
>   3. Resource manager (e.g., SLURM, PBS/Torque, LSF, etc.)
>   4. If none of a hostfile, the --host command line parameter, or an
>      RM is present, PRRTE defaults to the number of processor cores
> 
> In all the above cases, if you want PRRTE to default to the number
> of hardware threads instead of the number of processor cores, use the
> --use-hwthread-cpus option.
> 
> Alternatively, you can use the --map-by :OVERSUBSCRIBE option to ignore the
> number of available slots when deciding the number of processes to
> launch.
> --------------------------------------------------------------------------
> ```
> 
> https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories
> 
> Strange as the public runners all have at least 3 processors.

Ubuntu runders only have two processors. Mac tynnere have more

jhale:
Not according to githubs own documentation?

jorgensd:
> Not according to githubs own documentation?

are we reading the same documentation?
IMG_8445

jhale:
No, there are two sections, one for public repositories and one for private. You are reading the latter.

jorgensd:
> No, there are two sections, one for public repositories and one for private. You are reading the latter.

Interesting! That is a new development? 

I had the same oversubscription issue when using openmpi with adios4dolfinx a while back. I haven't seen it with mpich before.

.github/workflows/ccpp.yml
```diff
@@ -107,7 +108,7 @@ jobs:
       - name: Run C++ unit tests (MPI)
         run: |
           cd build/test
-          mpiexec -np 2 ctest -V --output-on-failure -R unittests
+          mpirun -n 3 --map-by :OVERSUBSCRIBE ctest -V --output-on-failure -R unittests
```
garth-wells, line 111: Don't use MPI implementation specific args - we can't then switch MPI version. Use environment variables (we already use some OpenMPI env variables on the CI).

See https://github.com/open-mpi/ompi/issues/8955#issuecomment-846272943, https://github.com/open-mpi/ompi/issues/8955#issuecomment-1295456308 and https://github.com/mpi4py/mpi4py-testing/blob/1dcad02e04f8c65c3e31416dcb86a7b8c1affcb8/.github/workflows/openmpi.yml#L98.

garth-wells (approved):

PR #3174, opened 2024-04-26
cpp/test/mesh: remove unused variable cshape

minrk:
causing [compiler error](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=922042&view=logs&j=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&t=6d4b912b-175d-51da-0fd9-4d30fe1eb4e7&l=2252) with `-Werror=unused-variable`

PR #3175, opened 2024-04-26
TST: avoid unused variable in assert

minrk:
CHECK appears to be the way checks are done elsewhere, this was the only assert left in the file.

I assume assert is being optimized out, causing [compiler error](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=922042&view=logs&j=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&t=6d4b912b-175d-51da-0fd9-4d30fe1eb4e7&l=2252) with `-Werror=unused-variable`. cpp tests pass during conda build with this patch.

First try at #3174 was wrong, but it didn't let me reopen that PR after I figured it out because I updated the branch while it was closed.

### DIFF
--- a/cpp/test/mesh/distributed_mesh.cpp
+++ b/cpp/test/mesh/distributed_mesh.cpp
@@ -131,7 +131,7 @@ void test_distributed_mesh(mesh::CellPartitionFunction partitioner)
     io::XDMFFile infile(subset_comm, "mesh.xdmf", "r");
     auto [_cells, _cshape] = infile.read_topology_data("mesh");
     auto [_x, _xshape] = infile.read_geometry_data("mesh");
-    assert(_cshape[1] == cshape[1]);
+    CHECK(_cshape[1] == cshape[1]);
     x = std::move(std::get<std::vector<T>>(_x));
     cells = std::move(_cells);
   }


garth-wells (approved):

garth-wells:
Thanks - clearly wrong to be using `assert` here.

jhale:
@minrk could you help us get our compile flags tightened so this doesn't happen again?

I'm surprised `-Wall` doesn't include this one.

https://github.com/FEniCS/dolfinx/blob/main/cpp/test/CMakeLists.txt#L51

minrk:
conda-forge adds `-DNDEBUG` to CPPFLAGS, which strips asserts. It would be reasonable to _remove_ `-DNDEBUG` from the compile args in the test directory, I would imagine, which would also allow this to pass. I don't know the best cmake incantation to _remove_ arguments.

If you know you don't want asserts, you could add `-DNDEBUG` by default as well, but I'm not sure that's the right way to do it, because other asserts that aren't the only use of variables would just get silently skipped. Seems like a linter ought to be the level of avoiding asserts, but I don't know the state of linters for C++.

PR #3176, opened 2024-04-26
Do CI linting in Actions image

garth-wells:
Use an Github Actions image rather the the (very large) FEniCS Docker image.

Shaves a full minute off linting time.


jhale:
Looks good.

Very frustrating that pip doesn't have a 'dependencies only' option so this information can be read from `pyproject.toml`.

PR #3114, opened 2024-03-24
Interpolation to and from sub-meshes for co-dimension 0

jorgensd:
A feature that is quite neat for post-processing, as one often only wants to visualize on a sub mesh.
Can also be used to avoid mixed-assembly in cases of iterative schemes:
1. Solve on a parent mesh
2. Move solution to sub mesh
3. Solve equation on sub mesh
4. Move solution back and repeat.

There is a slight API change in the C++ interface, as expressions need the associated mesh to be supplied.
In Python, the API only changes if you want to map between meshes.

PR #3177, opened 2024-04-26
Make standalone function for interpolation over non-matching meshes

### LINKED ISSUE #3173: Interpolation to/from sub-meshes causes parallel deadlock
PR #3114 causes deadlock in parallel, see https://github.com/FEniCS/dolfinx/actions/runs/8838261829.

Also see https://github.com/FEniCS/dolfinx/pull/3114#discussion_r1580530520.

The logic in https://github.com/FEniCS/dolfinx/blob/59449a0bfc7902e95145a463c6353e039b3b694a/python/dolfinx/fem/function.py#L415 and in https://github.com/FEniCS/dolfinx/blob/59449a0bfc7902e95145a463c6353e039b3b694a/python/dolfinx/fem/function.py#L468 is muddled and conflates `None` and `len(foo) == 0`.

jorgensd:
Resolves #3173.
Simplifies logic of interpolate.
**API Change**
Non-matching meshes interpolation now has its own interface: `interpolate_nonmatching_meshes` in both C++ and Python.
It also clarifies that interpolation of `dolfinx::fem::Expression` is not supported on non-matching meshes.


garth-wells:
The logic behind interpolation looks too complicated and the code is hard to follow. I don't see why different cases, e.g. matching and non-matching meshes, are forced through the same interface.

jorgensd:
> The logic behind interpolation looks too complicated and the code is hard to follow. I don't see why different cases, e.g. matching and non-matching meshes, are forced through the same interface.

I've now split the interpolation of nonmatching meshes into a separate function.

cpp/dolfinx/fem/Function.h
```diff
@@ -153,39 +153,21 @@ class Function
...
-      const Function<value_type, geometry_type>& v,
-      std::span<const std::int32_t> cells,
-      std::span<const std::int32_t> cell_map,
-      const std::tuple<std::span<const std::int32_t>,
-                       std::span<const std::int32_t>,
-                       std::span<const geometry_type>,
-                       std::span<const std::int32_t>>& nmm_interpolation_data
-      = {})
+  /// @param[in] cell_map A map from cells in the mesh associated with \p this
+  /// function to cells in mesh associated with \p v
```
garth-wells, line 157: 'maps' are much clearer if an example is added, e.g.:

"For cell `i` in the mesh associated with `this`, `cell_map[i]` is the index of the same cell but in the mesh associated with `v`."

(I think this is what it means)

cpp/dolfinx/fem/Function.h
```diff
@@ -153,39 +153,21 @@ class Function
...
-      std::span<const std::int32_t> cell_map = std::span<const std::int32_t>()
-      = {},
-      const std::tuple<std::span<const std::int32_t>,
-                       std::span<const std::int32_t>,
-                       std::span<const geometry_type>,
-                       std::span<const std::int32_t>>& nmm_interpolation_data
-      = {})
+  /// @param[in] cell_map Map from cells in self to cell indices in \p v
+  void interpolate(const Function<value_type, geometry_type>& v,
+                   std::span<const std::int32_t> cell_map
```
garth-wells, line 169: The double `std::span<const std::int32_t>()` looks weird.

cpp/dolfinx/fem/Function.h
```diff
@@ -290,7 +272,7 @@ class Function
   /// with `u`.
   /// @param[in] cells The cells to interpolate on
   /// @param[in] expr_mesh The mesh to evaluate the expression on
-  /// @param[in] cell_map Map from `cells` to cells in expression
+  /// @param[in] cell_map Map from \p cells to cells in expression
```
garth-wells, line 275: I don't under this docstring.

cpp/dolfinx/fem/Function.h
```diff
@@ -346,19 +328,16 @@ class Function
     std::vector<std::int32_t> cells_expr;
     cells_expr.reserve(num_cells);
     // Get mesh and check if mesh is shared
-    if (auto mesh_v = _function_space->mesh();
-        expr_mesh.topology() == mesh_v->topology())
+    if (auto mesh_v = _function_space->mesh(); &expr_mesh == &(*mesh_v))
```
garth-wells, line 331: `&(*mesh_v)` looks complicated. Are you just trying to check that the two meshes point to the same thing?

cpp/dolfinx/fem/Function.h
```diff
@@ -400,15 +379,43 @@ class Function
...
-    if (cell_map.size() == 0)
-      interpolate(e, cells, expr_mesh, cell_map);
-    else
-      interpolate(e, cells, expr_mesh, std::span(cells.data(), cells.size()));
+  /// Interpolate a function defined on a non-matching mesh
+  /// @param[in] v The function to be interpolated
+  /// @param[in] padding Absolute padding for every bounding box on the mesh of
+  /// \p this to determine point ownership
+  /// @brief Evaluate the Function at points.
+  /// @note For repeated interpolation, it is adviced to precompute the
```
garth-wells, line 390: But this function doesn't accept any interpolation data. It's not clear what the caller is being advised to do.

cpp/dolfinx/fem/Function.h
```diff
@@ -400,15 +379,43 @@ class Function
...
+
+  /// Interpolate a function defined on a non-matching mesh
+  /// @param[in] v The function to be interpolated
+  /// @param cells THe cells in the mesh associated with @p u that will be
+  /// interpolated into
+  /// @param nmm_interpolation_data Data required for associating the
+  /// interpolation points of @p u with cells in @p v
+  /// @note For repeated interpolation, it is adviced to precompute the
+  /// non-matching interpolation data
+  void interpolate_nonmatching_meshes(
```
garth-wells, line 407: This can be called interpolate - the compiler can resolve based on the arguments.

cpp/dolfinx/fem/Function.h
```diff
@@ -400,15 +379,43 @@ class Function
...
+                                 T padding)
+  {
+    fem::interpolate_nonmatching_meshes(*this, v, padding);
+  }
+
+  /// Interpolate a function defined on a non-matching mesh
+  /// @param[in] v The function to be interpolated
+  /// @param cells THe cells in the mesh associated with @p u that will be
+  /// interpolated into
+  /// @param nmm_interpolation_data Data required for associating the
```
garth-wells, line 403: It's not clear what this is. Is it documented somewhere?

cpp/dolfinx/fem/Function.h
```diff
@@ -400,15 +379,43 @@ class Function
...
+  void
+  interpolate_nonmatching_meshes(const Function<value_type, geometry_type>& v,
+                                 T padding)
+  {
+    fem::interpolate_nonmatching_meshes(*this, v, padding);
+  }
+
+  /// Interpolate a function defined on a non-matching mesh
+  /// @param[in] v The function to be interpolated
+  /// @param cells THe cells in the mesh associated with @p u that will be
```
garth-wells, line 401:
```suggestion
  /// @param cells Cells in the mesh associated with `u`  that will be
```

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1102,12 +1042,12 @@ void interpolate(Function<T, U>& u, std::span<const T> f,
 /// which to interpolate. Should be the same as the list used when
 /// calling fem::interpolation_coords.
 /// @param[in] padding Absolute padding of bounding boxes of all
-/// entities on `mesh1`. This is used avoid floating point issues when
+/// entities on \p mesh1. This is used avoid floating point issues when
```
garth-wells, line 1045: The back ticks look tidier.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1164,25 +1104,122 @@ create_nonmatching_meshes_interpolation_data(const mesh::Mesh<T>& mesh0,
       mesh0.geometry(), element0, mesh1, cells, padding);
 }
 
+/// @brief  Interpolate a finite element function one one grid to a function
```
garth-wells, line 1107:
```suggestion
/// @brief Interpolate a finite element function on one mesh to a function
```

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1164,25 +1104,122 @@ create_nonmatching_meshes_interpolation_data(const mesh::Mesh<T>& mesh0,
...
+/// defined on another (non-matching) grid.
```
garth-wells, line 1108:
```suggestion
/// defined on a different (non-matching) mesh.
```

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1164,25 +1104,122 @@ create_nonmatching_meshes_interpolation_data(const mesh::Mesh<T>& mesh0,
...
+/// @tparam T The Function scalar type
+/// @tparam U The Mesh geometry scalar type
+/// @param u The function to interpolate into
+/// @param v The function to interpolate from
+/// @param padding Absolute padding of bounding boxes of all entities on the
```
garth-wells, line 1113: I don't follow the doc string.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1164,25 +1104,122 @@ create_nonmatching_meshes_interpolation_data(const mesh::Mesh<T>& mesh0,
...
+/// mesh of \p u
+/// @note For repeated interpolation, it is adviced to precompute the
+/// non-matching interpolation data with @ref
+/// create_nonmatching_meshes_interpolation_data
+template <dolfinx::scalar T, std::floating_point U>
+void interpolate_nonmatching_meshes(Function<T, U>& u, const Function<T, U>& v,
```
garth-wells, line 1119: Can this function be removed?

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1164,25 +1104,122 @@ create_nonmatching_meshes_interpolation_data(const mesh::Mesh<T>& mesh0,
...
+  std::vector<std::int32_t> cells(num_cells, 0);
+  std::iota(cells.begin(), cells.end(), 0);
+  const auto nmm_interpolation_data
+      = create_nonmatching_meshes_interpolation_data(
+          u->function_space()->mesh(), u->function_space()->element(),
+          v->function_space()->mesh(), cells, padding);
+  interpolate_nonmatching_meshes(u, v, cells, nmm_interpolation_data);
+}
+
+/// @brief  Interpolate a finite element function one one grid to a function
```
garth-wells, line 1137: Text needs grammar fix.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1164,25 +1104,122 @@ create_nonmatching_meshes_interpolation_data(const mesh::Mesh<T>& mesh0,
...
+/// defined on another (non-matching) grid.
+/// @tparam T The Function scalar type
+/// @tparam U The Mesh geometry scalar type
+/// @param u The function to interpolate into
+/// @param v The function to interpolate from
+/// @param cells THe cells in the mesh associated with @p u that will be
```
garth-wells, line 1143:
```suggestion
/// @param cells Cells in the mesh associated with `u` that will be
```

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1164,25 +1104,122 @@ create_nonmatching_meshes_interpolation_data(const mesh::Mesh<T>& mesh0,
...
+/// interpolated into
+/// @param nmm_interpolation_data Data required for associating the
```
garth-wells, line 1145: Needs some explanation - how is it computed?

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1164,25 +1104,122 @@ create_nonmatching_meshes_interpolation_data(const mesh::Mesh<T>& mesh0,
...
+/// interpolation points of @p u with cells in @p v
+template <dolfinx::scalar T, std::floating_point U>
+void interpolate_nonmatching_meshes(
+    Function<T, U>& u, const Function<T, U>& v,
+    std::span<const std::int32_t> cells,
+    const std::tuple<std::span<const std::int32_t>,
```
garth-wells, line 1151: This data structure should really be made a `struct` - would be much clearer and could be documented properly.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1164,25 +1104,122 @@ create_nonmatching_meshes_interpolation_data(const mesh::Mesh<T>& mesh0,
...
+
+  {
+    auto mesh_v = v.function_space()->mesh();
+    assert(mesh_v);
+    int result;
+    MPI_Comm_compare(comm, mesh_v->comm(), &result);
+    if (result == MPI_UNEQUAL)
+    {
+      throw std::runtime_error("Interpolation on different meshes is only "
+                               "supported with the same communicator.");
```
garth-wells, line 1167:
```suggestion
                               "supported on the same communicator.");
```

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1164,25 +1104,122 @@ create_nonmatching_meshes_interpolation_data(const mesh::Mesh<T>& mesh0,
...
+  fem::interpolate<T>(u, valuesT_b, {valuesT.extent(0), valuesT.extent(1)},
+                      cells);
+}
+
 /// @brief Interpolate from one finite element Function to another one.
 /// @param[out] u1 The function to interpolate into
 /// @param[in] u0 The function to be interpolated
 /// @param[in] cells1 List of cell indices associated with the mesh of `u1` that
 /// will be interpolated onto
 /// @param[in] cell_map Mapping of cells in mesh associated with `u1` to cells
```
garth-wells, line 1217: Add example.

python/dolfinx/fem/function.py
```diff
@@ -412,6 +412,54 @@ def eval(self, x: npt.ArrayLike, cells: npt.ArrayLike, u=None) -> np.ndarray:
             u = np.reshape(u, (-1,))
         return u
 
+    def interpolate_nonmatching_meshes(
```
garth-wells, line 415:
```suggestion
    def interpolate_nonmatching(
```

python/dolfinx/fem/function.py
```diff
@@ -412,6 +412,54 @@ def eval(self, x: npt.ArrayLike, cells: npt.ArrayLike, u=None) -> np.ndarray:
...
+        nmm_interpolation_data: typing.Optional[PointOwnershipData] = None,
+        padding: float = 0.0,
+    ) -> None:
+        """Interpolate a function defined on one mesh to a function defined on another mesh
+
+        Args:
+            u: The Function to interpolate.
+            cells: The cells to interpolate over. If `None` then all
+                cells are interpolated over.
+            nmm_interpolation_data: Data needed to interpolate functions defined on other meshes
```
garth-wells, line 428: Expand comment so user knows how (and why) to compute it.

python/dolfinx/fem/function.py
```diff
@@ -412,6 +412,54 @@ def eval(self, x: npt.ArrayLike, cells: npt.ArrayLike, u=None) -> np.ndarray:
...
         return u
 
+    def interpolate_nonmatching_meshes(
+        self,
+        u: Function,
+        cells: typing.Optional[np.ndarray] = None,
+        nmm_interpolation_data: typing.Optional[PointOwnershipData] = None,
+        padding: float = 0.0,
+    ) -> None:
+        """Interpolate a function defined on one mesh to a function defined on another mesh
```
garth-wells, line 422:
```suggestion
        """Interpolate a Function defined on one mesh to a function defined on a different mesh
```

python/dolfinx/fem/function.py
```diff
@@ -412,6 +412,54 @@ def eval(self, x: npt.ArrayLike, cells: npt.ArrayLike, u=None) -> np.ndarray:
...
+
+        Args:
+            u: The Function to interpolate.
+            cells: The cells to interpolate over. If `None` then all
+                cells are interpolated over.
+            nmm_interpolation_data: Data needed to interpolate functions defined on other meshes
+            padding: Absolute padding of bounding boxes for computing
```
garth-wells, line 429: Force user to create `nmm_interpolation_data`, then this isn't needed. 

We should avoid redundant arguments - if a user passes `nmm_interpolation_data` and `padding`, then `padding` has no effect. Need to keep the logic simple.

garth-wells (changes requested):
I think there is more that can be done to simplify and improve the logic.
- In C++ the number of functions can probably be reduced.
- The non-matching interpolation data structure should be a `struct`.
- The Python interface still has a function that tries to do too much. Interpolation for non-matching meshes is advanced functionality so it it ok for the user to be responsible for some of the steps, i.e. let the user create the required data structure.

cpp/dolfinx/fem/Function.h
jorgensd: Yes, that is what I am trying to do. As we do not have an equality operator for mesh, this was the best thing I could think of.

cpp/dolfinx/fem/Function.h
jorgensd: You suggested that I should add this back in: https://github.com/FEniCS/dolfinx/pull/3114#discussion_r1555708444

cpp/dolfinx/fem/Function.h
jorgensd, line 157:
```suggestion
  /// @param[in] cell_map For cell `i` in the mesh associated with \p this, 
  /// `cell_map[i]` is the index of the same cell, but in the mesh associated with `v`
```

cpp/dolfinx/fem/Function.h
jorgensd:
```suggestion
  /// @param[in] cell_map For cell `i` in the mesh associated with \p this, 
 /// `cell_map[i]` is the index of the same cell, but in \p expr_mesh
```

cpp/dolfinx/fem/interpolate.h
jorgensd, line 1045:
```suggestion
/// entities on `mesh`. This is used avoid floating point issues when
```

cpp/dolfinx/fem/Function.h
garth-wells: If both are shared pointers, `expr_mesh == mesh_v` should work.

cpp/dolfinx/fem/Function.h
garth-wells: The double equality in the comment is a typo.

cpp/dolfinx/fem/Function.h
jorgensd: They are not both shared pointers, as `expr_mesh` is a reference: `const dolfinx::mesh::Mesh<geometry_type>& expr_mesh,`

cpp/dolfinx/fem/Function.h
jorgensd: So you would like `std::span<const std::int32_t> cell_map = {}` ?

cpp/dolfinx/fem/Function.h
jorgensd: Added C++ struct

cpp/dolfinx/fem/Function.h
```diff
@@ -382,15 +360,14 @@ class Function
                      {value_size, num_cells * num_points}, cells);
   }
 
-  /// Interpolate an Expression (based on UFL) on all cells
+  /// @brief Interpolate an Expression (based on UFL) on all cells.
   /// @param[in] e The function to be interpolated
-  /// @param[in] expr_mesh Mesh the Expression `e` is defined on.
+  /// @param[in] expr_mesh Mesh the expression `e` is defined on.
   /// @param[in] cell_map Map from `cells` to cells in expression if
```
jorgensd, line 366:
```suggestion
  /// @param[in] cell_map Map from cells in the mesh of `this` to cells in expression if
```

garth-wells:
> @garth-wells anything that you want to further change here?

Yes, the docs need improvement. I haven't had time to do it yet.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1133,56 +1055,94 @@ create_nonmatching_meshes_interpolation_data(
...
+/// @brief Interpolate from one finite element fem::Function to another
+/// on the same mesh.
+///
+/// Functions may be defined on 'sub-meshes'.
+///
+/// @param[out] u1 Function to interpolate into.
+/// @param[in] u0 Function to b interpolated from.
+/// @param[in] cells1 Cell indices associated with the mesh of `u1` that
+/// will be interpolated onto.
+/// @param[in] cell_map For cell `i` in the mesh associated with `u1`,
```
garth-wells, line 1139: @jorgensd I think this would be a lot simpler if `cell_map` was `cells0`. For each cell in `cells1` (which are in `mesh1`, `cells0` is the corresponding cell in `mesh0`.

jorgensd: It would also mean that the handling of "no" cell-map (functions live on the same mesh) can be confused with having no cells on the current process that you want to interpolate into. It would mean that for every interpolation, a user would have to specify what cells they are interpolating from (major API breakage).

garth-wells: > I do not agree, as we use `entity_maps` in the form creation for mesh view assembly. Having multiple concepts floating around in the different parts of the SubMesh/Meshview code is confusing to the user.

And I don't agree. Functions should take the minimum reasonable data and not more. `cell_map` as it stands could be a large data structure that the user may wish to discard. Internally the function computes what I have suggested what should be passed in. The proposal therefore can save memory and increases re-use.

jorgensd: The abstraction is one level further out (in `Function.h`). I've moved the cell_map interface there.

garth-wells:
> @garth-wells I've now addressed all comments, any further?

I'm still going through it.

garth-wells (approved):

PR #3182, opened 2024-04-26
Support Python interface build and run without petsc4py

garth-wells:
This allows the Python interface to be built without petcc4py.

`pyproject.toml` doesn't handle optional build requirements, so comments have been added to explain why petsc4py isn't listed as build dependency.

As it stands, if the C++ interface is built with PETSc *and* petsc4py is found then petsc4py is enabled in the Python interface build. Otherwise petsc4py is not enabled in the build.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -83,6 +83,7 @@ execute_process(
 if(NOT PETSC4PY_COMMAND_RESULT)
   message(STATUS "Found petsc4py include directory at ${PETSC4PY_INCLUDE_DIR}")
   target_include_directories(cpp PRIVATE ${PETSC4PY_INCLUDE_DIR})
+  target_compile_definitions(cpp PRIVATE HAS_PETSC4PY)
 else()
   message(STATUS "petsc4py not found.")
 endif()
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -84,7 +84,7 @@ if(NOT PETSC4PY_COMMAND_RESULT)
   message(STATUS "Found petsc4py include directory at ${PETSC4PY_INCLUDE_DIR}")
   target_include_directories(cpp PRIVATE ${PETSC4PY_INCLUDE_DIR})
 else()
-  message(FATAL_ERROR "petsc4py could not be found.")
+  message(STATUS "petsc4py not found.")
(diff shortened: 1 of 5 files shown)


python/pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
 docs = ["markdown", "pyyaml", "sphinx", "sphinx_rtd_theme"]
 lint = ["ruff"]
 optional = ["numba"]
+petsc4py = ["petsc4py"]
```
jorgensd, line 39: Maybe we could just call this `petsc = ["petsc4py"]` for short?

garth-wells: `petsc` and `petssc4py` are different things.

.github/workflows/ccpp.yml
```diff
@@ -124,10 +124,12 @@ jobs:
           cd build/test
           mpirun -n 3 ctest -V --output-on-failure -R unittests
 
-      # - name: Build Python interface
-      #   run: python3 -m pip -v install --check-build-dependencies --no-build-isolation --config-settings=cmake.build-type="Debug" python/
-      # - name: Test Python import
-      #   run: python3 -c "import dolfinx; print(dolfinx.__version__)"
+      - name: Install Python build dependencies
+        run: pip install scikit-build-core[pyproject] mpi4py nanobind scikit-build-core[pyproject]
```
jhale, line 128: Repeated scikit-build-core

python/CMakeLists.txt
```diff
@@ -62,12 +62,6 @@ nanobind_add_module(
   dolfinx/wrappers/refinement.cpp
 )
 
-# Add strict compiler flags include(CheckCXXCompilerFlag)
-# check_cxx_compiler_flag("-Wall -Werror -pedantic" HAVE_PEDANTIC)
-
-# if(HAVE_PEDANTIC) # target_compile_options(cpp PRIVATE
-# -Wall;-Werror;-pedantic) endif()
```
jhale, line 69: Don't we want these on?

jhale (approved):

python/CMakeLists.txt
garth-wells: No, because they fail on nanobind.

PR #3184, opened 2024-04-28
Split PETSc and non-PETSc tests on CI

garth-wells:
This speeds up CI, and will avoid regressions when PETSc is not installed.


jorgensd:
We need to add some separate tests of the `petsc` interaction/integration with DOLFINx. Currently LinearProblem is untested, which is not great.

jhale (approved):

PR #3185, opened 2024-04-28
Run Actions with OpenMPI and MPICH, and remove CircleCI config

garth-wells:
CircleCI was used to run tests using MPICH. The PR brings OpenMPI and MPICH under Actions.

To simplify also, also splits documentation publishing into its own step/job.


jhale (approved):
Looks good. The doc publish step could potentially be done by packing up artefacts between steps (avoids a build) but I'm not sure it's worth the extra complexity.

PR #3186, opened 2024-04-29
Add check of element hash

### LINKED ISSUE #3169: Add check of element hash

After #3136, we reuse Basix elements rather than recreating them based on information from UFCx.

In https://github.com/FEniCS/ffcx/pull/681, a hash of the element is added to the generated code. We should add a check (here: https://github.com/FEniCS/dolfinx/blob/e343a76e1e6d2d1ca28a26811d5d4085180b7c01/cpp/dolfinx/fem/utils.h#L361) that confirms that the hash of the element being used matches that in the generated code.

### Suggested user interface

_No response_

mscroggs:
Resolves #3169.

Depends on https://github.com/FEniCS/ffcx/pull/688


python/test/unit/fem/test_forms.py
```diff
@@ -48,3 +52,88 @@ def test_extract_forms():
...
+        f = ftype(
+            module.ffi.cast("uintptr_t", module.ffi.addressof(ufcx_form)),
+            [incorrect_space._cpp_object, incorrect_space._cpp_object],
+            [],
+            [],
+            {dolfinx.cpp.fem.IntegralType.cell: []},
+            {},
+            mesh._cpp_object,
+        )
+        dolfinx.fem.Form(f, ufcx_form, code)
```
mscroggs, line 139: Checking all these cases is probably overkill. Maybe checking one is enough:

```suggestion
    with pytest.raises(RuntimeError):
        f = ftype(
            module.ffi.cast("uintptr_t", module.ffi.addressof(ufcx_form)),
            [incorrect_space._cpp_object, incorrect_space._cpp_object],
            [],
            [],
            {dolfinx.cpp.fem.IntegralType.cell: []},
            {},
            mesh._cpp_object,
        )
        dolfinx.fem.Form(f, ufcx_form, code)
```

garth-wells (approved):

PR #3191, opened 2024-05-02
Update basix interface

mscroggs:
see https://github.com/FEniCS/basix/pull/823


cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -8,7 +8,7 @@
 
 #include "traits.h"
 #include <array>
-#include <basix/finite-element.h>
+#include <basix/basix.h>
```
garth-wells, line 11: Good practice is to include only what is needed rather than a kitchen sink header.

PR #3198, opened 2024-05-05
Port DOLFINx to Windows

jhale:
This ports DOLFINx to Windows building against vcpkg using Microsoft Visual C/C++ compiler. It uses an overlay port for Intel MPI (MSMPI does not support neighbourhood collective operations from MPI3). Note that PETSc and petsc4py are not available on Windows.

The primary purpose of this port is to enable conda Windows packages, rather than users building DOLFINx on Windows from source themselves.

Main changes:

- Windows CI build of DOLFINx including C++ and Python unit tests and Python demos.
- Disables C99 `_Complex` when using Windows/MSVC.
- Allows for Basix to be found without asking Python module for hints `-DDOLFINX_BASIX_PYTHON=OFF`.
- Adds note about beta conda windows packages to README.
- Adds note about vcpkg windows build possibility to README.

Untested but potentially possible:

- Building against vcpkg on Linux and macOS.
- Building using a different compiler on Windows.

Outstanding points to tackle in other PRs:
- High runtime of Python unit tests - should be checked on Conda build. Probably slow JIT.
- Porting subset of C++ demos to work without PETSc.


jhale:
@minrk What's the status of Intel MPI in Conda? It seems to be the only MPI3 compliant build of MPI on Windows - MSMPI doesn't have neighbourhood operations, which we use extensively.

jorgensd:
> @minrk What's the status of Intel MPI in Conda? It seems to be the only MPI3 compliant build of MPI on Windows - MSMPI doesn't have neighbourhood operations, which we use extensively.

The support in MSMPI seems strange, as you can create distributed communicators, but not use alltoall on them: 
- https://learn.microsoft.com/en-us/message-passing-interface/mpi-dist-graph-create-function
- https://learn.microsoft.com/en-us/message-passing-interface/mpi-collective-functions

minrk:
I've never interacted with it personally so I don't know the details or pitfalls, but at least mpi4py is built with intel mpi on windows and linux. So it should be an option, at least.

jhale:
> > @minrk What's the status of Intel MPI in Conda? It seems to be the only MPI3 compliant build of MPI on Windows - MSMPI doesn't have neighbourhood operations, which we use extensively.
> 
> The support in MSMPI seems strange, as you can create distributed communicators, but not use alltoall on them:
> 
> * https://learn.microsoft.com/en-us/message-passing-interface/mpi-dist-graph-create-function
> * https://learn.microsoft.com/en-us/message-passing-interface/mpi-collective-functions

https://github.com/microsoft/Microsoft-MPI - This clears it up. MSMPI is based on an old MPICH and is 2.2 compliant, and implements some features of 3.1 (not the critical neighbourhood alltoall we need).

It seems that there is no open source and modern standards compliant MPI for Windows. We'll have to stick with Intel MPI for the proposed Conda builds.

cpp/dolfinx/CMakeLists.txt
```diff
@@ -139,13 +139,28 @@ endif()
...
+    EXPORT DOLFINXTargets
+    RUNTIME_DEPENDENCY_SET dependencies
+    RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR} COMPONENT RuntimeExecutables
+    LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT RuntimeLibraries
+    ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT Development
+  )
+
+  LIST(APPEND PRE_EXCLUDE_REGEXES "api-ms-.*")
+  LIST(APPEND PRE_EXCLUDE_REGEXES "ext-ms-.*")
+  install(RUNTIME_DEPENDENCY_SET dependencies DESTINATION ${CMAKE_INSTALL_BINDIR} PRE_EXCLUDE_REGEXES ${PRE_EXCLUDE_REGEXES} POST_EXCLUDE_REGEXES ".*system32/.*\\.dll" DIRECTORIES $<TARGET_FILE_DIR:Basix::basix>)
```
minrk, line 154: noting from https://github.com/FEniCS/basix/pull/822 that installing dependencies should at least be optional, and I would _guess_ opt-in is best, rather than opt-out like I did in https://github.com/FEniCS/basix/pull/822.

jhale: I think I'm going to change both to be opt-in.

.github/workflows/ccpp.yml
```diff
@@ -85,7 +85,16 @@ jobs:
...
         run: |
-          sudo apt-get install catch2 cmake g++ libboost-dev libboost-filesystem-dev libboost-timer-dev libhdf5-mpi-dev libparmetis-dev libpugixml-dev mpi-default-dev ninja-build pkg-config
+          sudo apt-get install catch2 cmake g++ libboost-dev libboost-filesystem-dev libboost-timer-dev libgflags-dev libhdf5-mpi-dev libparmetis-dev libpugixml-dev mpi-default-dev ninja-build pkg-config
+          git clone https://github.com/google/glog.git
+          cd glog
+          cmake -S . -B build -G "Unix Makefiles"
+          cmake --build build
+          sudo cmake --build build --target install
+          cd ..
+          sudo rm -rf glog
```
jorgensd, line 95: Cant we use: https://packages.ubuntu.com/search?keywords=google-glog here, i.e.
https://packages.ubuntu.com/noble/libgoogle-glog-dev

jhale: Comments on the glog switch on Chris's PR. I'm just doing some incremental merges as this cannot go in until glog is done (loguru is not working on Windows).

chrisrichardson: I'm trying out spdlog now, I think it is better.

jhale:
@minrk This is probably now ready to test on Windows. I took a quick look and it seemed that two important dependencies (HDF5 + Parmetis) could be missing Intel MPI builds on Windows before this can really be tested.

minrk:
With ffcx working on conda, I'm looking at dolfinx now. An impediment to getting this in conda-forge is that neither parmetis nor hdf5 + mpi are currently packaged on Windows. ptscotch is on Windows, though, but only built with msmpi and I seem to recall you needed intel-mpi and not msmpi.

Current conda-forge dependency status on Windows:

- adios2 - no mpi
- kahip - no
- hdf5 - no mpi
- parmetis - no
- ptscotch - msmpi
- petsc/slepc - lol, no

I can work on some of this, but it would help to know what to prioritize. Looking at this PR, it seems parmetis + hdf5 on impi is sufficient. Is that right? hdf5 without mpi is definitely required?

jhale:
Thanks for the summary. HDF5 and a partitioner are mandatory; Parmetis, and if Parmetis proves difficult, KaHiP instead. Don't worry about the others (ADIOS2 would be the next thing to include, but no worries for a first release).

minrk:
Yeah, I've been looking at the [vcpkg parmetis recipe](https://github.com/microsoft/vcpkg/blob/2fd62b5d878104f4092af80533923bfe2bba2ee0/ports/parmetis/portfile.cmake). Unfortunately, it requires quite a bit of patching to build on Windows, and is based on an unreleased version of parmetis, not 4.0.3, which was the last release quite a long time ago. But perhaps we could port those patches. I don't know what to do about being based on an unreleased version, though.

KaHIP also [doesn't support Windows](https://github.com/KaHIP/KaHIP/issues/131), and scotch technically doesn't either, but someone is maintaining patches that work for msmpi, at least.

So there are really no parallel partitioners that I know of that support Windows, but since there is already ptscotch with msmpi, I figured that might be the easiest one: https://github.com/conda-forge/scotch-feedstock/pull/83

jhale:
Ptscotch would be also ok. Note that we don't need Python KaHiP, just the C++ part.

.github/workflows/windows.yml
```diff
@@ -0,0 +1,144 @@
+name: Windows build
+
+on:
```
garth-wells, line 3: Should we have run on pull requests?

.github/workflows/windows.yml
```diff
@@ -0,0 +1,144 @@
...
+            core.exportVariable('ACTIONS_RUNTIME_TOKEN', process.env.ACTIONS_RUNTIME_TOKEN || '');
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+
+      - uses: actions/checkout@v4
+        with:
+          repository: "fenics/ufl"
```
garth-wells, line 34: Simple `pip install git+https://github.com/...` doesn't work?

cpp/.vcpkg-overlay/intel-mpi/portfile.cmake
```diff
@@ -0,0 +1,78 @@
+set(INTELMPI_VERSION "2021.12")
```
garth-wells, line 1: Could you add a comment to summarise what this file does? What are the prerequisites?

cpp/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 # ------------------------------------------------------------------------------
 # Top level CMakeLists.txt file for DOLFINx
-cmake_minimum_required(VERSION 3.19)
```
garth-wells, line 3: Let's just require 3.21 - different requirements adds complexity.

cpp/CMakeLists.txt
```diff
@@ -55,6 +65,9 @@ add_feature_info(
   BUILD_SHARED_LIBS BUILD_SHARED_LIBS "Build DOLFINx with shared libraries."
 )
 
+option(INSTALL_RUNTIME_DEPENDENCIES "Include runtime dependencies in install (Windows-only)" OFF)
```
garth-wells, line 68: What are the runtime dependencies that are installed?

cpp/CMakeLists.txt
```diff
@@ -72,6 +85,18 @@ add_feature_info(
...
+# Control Basix discovery
+option(
+  DOLFINX_BASIX_PYTHON
+  "Enable Basix discovery using Python. Disable if Basix should be found using CMake."
+  ON
+)
+add_feature_info(
+  DOLFINX_BASIX_PYTHON
+  DOLFINX_BASIX_PYTHON
+  "Enable Basix discovery using Python. Disable if Basix should be found using CMake."
```
garth-wells, line 97: I think this is a bit misleading - Basix is always 'found' by cmake. Python can be used to find where the Basix CMake file is.

cpp/CMakeLists.txt
```diff
@@ -81,7 +106,7 @@ option(
 add_feature_info(
   DOLFINX_UFCX_PYTHON
   DOLFINX_UFCX_PYTHON
-  "Enable UFCx discovery using Python. Disable if UFCx should be found using a CMake config file."
+  "Enable UFCx discovery using Python. Disable if UFCx should be found using CMake."
```
garth-wells, line 109: See comment on Basix.

cpp/CMakeLists.txt
```diff
@@ -257,10 +282,15 @@ if(HAVE_O2_OPTIMISATION)
   list(APPEND DOLFINX_CXX_DEVELOPER_FLAGS -O2)
 endif()
 
-# Turn off some checks in gcc12 and gcc13 due to false positives with the fmt library
-if(CMAKE_COMPILER_IS_GNUCXX AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "11.4"
-  AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS "14.0")
-  list(APPEND DOLFINX_CXX_DEVELOPER_FLAGS -Wno-array-bounds;-Wno-stringop-overflow)
+# Turn off some checks in gcc12 and gcc13 due to false positives with the fmt
+# library
+if(CMAKE_COMPILER_IS_GNUCXX
```
garth-wells, line 287: Does this include clang? clang sometimes identifies as GNU.

cpp/CMakeLists.txt
```diff
@@ -284,33 +314,41 @@ set_package_properties(
...
-  execute_process(
-    COMMAND
-      ${Python3_EXECUTABLE} -c
-      "import basix, os, sys; sys.stdout.write(os.path.dirname(basix.__file__))"
-    OUTPUT_VARIABLE BASIX_PY_DIR
-    RESULT_VARIABLE BASIX_PY_COMMAND_RESULT
-    ERROR_QUIET OUTPUT_STRIP_TRAILING_WHITESPACE
+# If requested (default), ask the Python interpreter for hints on where to find
+# Basix C++ library.
+if(DOLFINX_BASIX_PYTHON)
```
garth-wells, line 322: This is much better than before.

cpp/cmake/modules/FindUFCx.cmake
```diff
@@ -48,6 +54,8 @@ execute_process(
     "import ffcx.codegeneration, sys; sys.stdout.write(ffcx.codegeneration.get_include_path())"
   OUTPUT_VARIABLE UFCX_INCLUDE_DIR
 )
+# TODO: CMake 3.20 has more modern cmake_path.
```
garth-wells, line 57: Let's update to CMake 3.21 (see above comment).

cpp/demo/poisson_matrix_free/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ project(${PROJECT_NAME} LANGUAGES C CXX)
 
 # Set C++20 standard
 set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_C_STANDARD 17)
```
garth-wells, line 12: Any reason for C++17 here? Could cause problems when including header files.

I think the CMake files are generated, so changes should be make in the generating script.

cpp/demo/poisson_matrix_free/CMakeLists.txt
```diff
@@ -35,6 +36,9 @@ else()
     set(SCALAR_TYPE "--scalar_type=float32")
   endif()
 endif()
+
+set(SCALAR_TYPE "--scalar_type=float64")
```
garth-wells, line 40: See above comment on generating o CMake files.

cpp/test/CMakeLists.txt
```diff
@@ -44,12 +45,17 @@ add_executable(
...
+if (WIN32)
+  target_link_libraries(unittests PRIVATE bcrypt)
+endif()
 target_include_directories(
   unittests PRIVATE $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
 )
-target_compile_options(
-  unittests PRIVATE $<$<COMPILE_LANGUAGE:CXX>:-Wall -Werror -Wextra -pedantic>
-)
+#target_compile_options(
```
garth-wells, line 56: Uncomment or remove.

cpp/vcpkg.json
```diff
@@ -0,0 +1,22 @@
+{
```
garth-wells, line 1: Can a comment be added to summarise what it's form?

python/CMakeLists.txt
```diff
@@ -64,6 +55,11 @@ nanobind_add_module(
 
 # Add DOLFINx libraries
 target_link_libraries(cpp PRIVATE dolfinx)
+# UUID requires bcrypt to be linked on Windows, broken in vcpkg.
```
garth-wells, line 58: Would be nice if we could remove UUID.

python/dolfinx/__init__.py
```diff
@@ -6,9 +6,14 @@
 """Main module for DOLFINx"""
 
 # flake8: noqa
-
+import os
 import sys
 
+if sys.platform.startswith("win32"):
```
garth-wells, line 12: Hardwired paths don't look great.

python/test/unit/fem/test_element_integrals.py
```diff
@@ -440,7 +480,15 @@ def test_plus_minus_vector(cell_type, pm1, pm2, dtype):
 @pytest.mark.parametrize("pm1", ["+", "-"])
 @pytest.mark.parametrize("pm2", ["+", "-"])
 @parametrize_cell_types
-@pytest.mark.parametrize("dtype", [np.float32, np.float64, np.complex64, np.complex128])
+@pytest.mark.parametrize(
+    "dtype",
```
garth-wells, line 484: Could we move the list of types to a single place at the top of the file?

garth-wells (changes requested):
Nice work!

garth-wells:
> So there are really no parallel partitioners that I know of that support Windows, but since there is already ptscotch with msmpi, I figured that might be the easiest one: [conda-forge/scotch-feedstock#83](https://github.com/conda-forge/scotch-feedstock/pull/83)

What about METIS and Windows? We're not targeting HPC with Windows, so we could use METIS in DOLFINx if METIS supports Windows.

minrk:
metis is indeed packaged on Windows.

It looks like it might not be super hard. ptscotch builds with impi are done (https://github.com/conda-forge/scotch-feedstock/pull/83), and hdf5 may be working soon: https://github.com/conda-forge/hdf5-feedstock/pull/218

jhale:
Well, I'm impressed and a little surprised! Great work @minrk

.github/workflows/windows.yml
jhale: Not sure because it's > 45 mins per run (!). Nightly?

.github/workflows/windows.yml
jhale: Can change.

cpp/CMakeLists.txt
jhale: Breaks Rocky 9 as it only has CMake 3.19.

cpp/CMakeLists.txt
jhale: Will change.

cpp/CMakeLists.txt
jhale: Will change.

cpp/cmake/modules/FindUFCx.cmake
jhale: See comment on Rocky 9.

cpp/demo/poisson_matrix_free/CMakeLists.txt
jhale: Not the first to misread this - it's C17, not C++17.

cpp/demo/poisson_matrix_free/CMakeLists.txt
jhale: @chrisrichardson kindly offered to help port the C++ demos including rethinking this bit.

cpp/test/CMakeLists.txt
jhale: My view is we should only play with Developer flags on Unix. All others should be left default.

python/dolfinx/__init__.py
jhale: This has to be removed completely - on Windows the CI needs to manually inject these paths at install time (python on windows will not respect PATH for dlls)

python/test/unit/fem/test_element_integrals.py
jhale: Will change.

cpp/CMakeLists.txt
jhale: Don't know - @chrisrichardson ? This line probably needs checking out from main anyway as it's not chsnged by my branch.

cpp/vcpkg.json
jhale: No comments allowed in json but could put in readme.

minrk:
conda-forge update: hdf5 and ptscotch now have impi builds, so I’ve started [looking at dolfinx](https://github.com/conda-forge/fenics-dolfinx-feedstock/pull/78).

My first snag was that I couldn't get FindSCOTCH to work on Windows (the test wasn't linking MPI correctly, despite MPI being found just fine). But since scotch 7 itself uses cmake, switching from `find_package(SCOTCH)` to `find_package(SCOTCH CONFIG)` worked fine. I'm not sufficiently confident in CMake to make a suggestion for how best to handle the fact that regular `find_package(SCOTCH)` without custom FindSCOTCH may work and might be preferred, but only with SCOTCH >= 7. It seems reasonable to require cmake scotch on Windows and not try to figure out FindSCOTCH.

Right now, compilation of libdolfinx completes, but linking hdf5 fails with some missing symbols, which is weird because hdf5.lib is definitely linked, and `dumpbin /exports hdf5.lib` shows that all the missing symbols are indeed defined, and `hdf5.lib` is being linked.

minrk:
cpp builds, now figuring out why python can't find MPI_C with the same cmake args that work for cpp.

jhale:
I wanted to point out the Cmake option to disable Basix being hinted using Python, which is probably useful to make your libdolfinx build Python free.

We can try removing findscotch, config mode is also preferred.

garth-wells:
We could try first for config mode, then findscotch. PETSc isn't building PTSCOTCH using CMake, in which case a PTSCOTCH CMake config file is not installed.

minrk:
so something like putting `find_package(SCOTCH CONFIG QUIET)` in the beginning of `FindSCOTCH.cmake`, and proceed with the rest `if(NOT SCOTCH_FOUND)`?

garth-wells:
Put it in the top-level DOLFINx `CMakeLists.txt`. I can have a stab shortly, but not sure I'll be able to test it.

python/dolfinx/__init__.py
minrk: conda Windows builds are now working (Python build finishes, cpp tests pass), but Python is stopping here because these directories don't exist.

If these came from the env like `DOLFINX_ADD_DLL_DIRECTORY_PATH` and/or checked if they existed before adding them, it would be easier.

cpp/CMakeLists.txt
garth-wells: > Breaks Rocky 9 as it only has CMake 3.19.

Seems to me to have 3.26.

minrk:
C++ and Python builds are working in conda. Folks should be able to try them out with:

```
conda install -c minrk/label/fenics-windows -c conda-forge fenics-dolfinx=0.9.0.dev
```

I've no idea how it's going to do with finding compilers on a real user system as opposed to the conda-forge CI, but I would be a bit surprised if it works by default without being launched in a session with vcvarsall.bat or similar.

All C++ tests are passing, and almost all Python tests are passing. All [failing python tests](https://gist.github.com/minrk/7eaf82447ba270b38474ee646be91e8c) appear to be attributable to missing skips for the absence of adios2 and petsc.

I used [this branch](https://github.com/FEniCS/dolfinx/compare/jhale/windows-build...minrk:dolfinx:jhale/windows-build), which is just this PR plus #3241 and a commit to remove the `os.add_dll_directory` which prevents import when the hardcoded directories do not exist.

FWIW, tests are run with ufl 2024.1 because I haven't packaged 2024.2 dev, and everything passes. That suggests to me that the lower bound on ufl has been bumped prematurely.

minrk:
Forgot to mention, this warning came up:

```
libffcx_forms_1b9958e91fd54e9c55f60346c28378fb63cad26e.c(722): warning C4305: 'initializing': truncation from 'double' to 'const float'
```

104,127 times when running the tests. I'm not sure if that's something that should be fixed in ffcx.

compiler warnings summary from running the tests with `pytest -vs`:

```
# grep -E -o 'warning C.+' dolfinx-win64.txt | sort | uniq -c
1665 warning C4113: 'ufcx_tabulate_tensor_float32 (__cdecl **)' differs in parameter lists from 'void (__cdecl **)(float *,const float *,const float *,const float *,const int *,const uint8_t *)'
1665 warning C4113: 'ufcx_tabulate_tensor_float64 (__cdecl **)' differs in parameter lists from 'void (__cdecl **)(double *,const double *,const double *,const double *,const int *,const uint8_t *)'
 165 warning C4244: 'initializing': conversion from 'double' to 'float', possible loss of data
104127 warning C4305: 'initializing': truncation from 'double' to 'const float'
```

jhale:
Thanks for the detailed feedback! We will look at the points you've raised, tidy this up and come back for another conda build.

In terms of testing on a user system, how can we get your builds onto a Windows desktop with conda setup? We have a Windows desktop for testing.

minrk:
The command I gave above should be able to get it for anyone with Conda on windows:

    conda install -c minrk/label/fenics-windows -c conda-forge fenics-dolfinx=0.9.0.dev

cpp/.vcpkg-overlay/intel-mpi/portfile.cmake
jhale: Added README.md.

jhale:
Once this runs green it can be re-built on conda-forge. The basix branch should now be unnecessary, I removed the DLL stuff in favour of a proper solution.

cpp/demo/poisson_matrix_free/CMakeLists.txt
jhale: Will leave for another PR - not yet done on Unix.

cpp/CMakeLists.txt
jhale: This is deprecated in any case:

https://cmake.org/cmake/help/latest/variable/CMAKE_COMPILER_IS_GNUCXX.html

So will switch it out for something more modern.

.github/workflows/windows.yml
jhale: Let's see how many complaints we get...

Issue #3076, opened 2024-02-27
Restructure demos into more functions where appropriate


garth-wells:
Most demos (C++ and Python) have a flat structure. This is fine for simple problems, but I see time and time again that users model their more extensive solvers on the demos and create single function (C++) / flat script (Python) demos that can be many hundreds or thousands of lines.

It would be good to review the structure of demos and refactor with more functions where appropriate. Hopefully users will then model their own codes on the more structured demos.

jhale:
I've seen issues amongst my collaborators on how to properly structure a 'real' DOLFINx solver. The flat files are very appealing, but it runs out of steam quickly. These are the things I see in 'real' solvers:

- Parametric problems.
- Mesh generation, real geometry, lots of facet and domain tags.
- Significant number of parameters, in different subdomains.
- A time loop. Perhaps with changing timestep at some point.
- Partial IO, i.e. not every timestep.
- Checkpointing output (after summer GSoC project).
- Multiple stages, e.g. preprocessing, simulation, postprocessing (e.g. summary statistics).

Clearly there is not only one way to do this, but an opinionated demo would be good to get people on the write track.

jhale:
- Good ways to write input parameters as flat files and reading them in.

risa-hirsh:
Hi, I’m new to contributing and would like to help with this. Is this still something that's needed? If so, is there a specific demo you would recommend starting with?

PR #3202, opened 2024-05-09
Add pyamg demo

chrisrichardson:
A simple pyamg example.


python/pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
 [project.optional-dependencies]
 docs = ["markdown", "pyyaml", "sphinx", "sphinx_rtd_theme"]
 lint = ["ruff"]
-optional = ["numba"]
+optional = ["numba", "pyamg"]
```
garth-wells, line 38: `pyamg` shouldn't be added here, it's require for a demo but is not used anywhere in the core.

python/demo/demo_pyamg.py
```diff
@@ -0,0 +1,96 @@
...
+g = ufl.sin(5 * x[0])
+a = form(inner(grad(u), grad(v)) * dx, dtype=dtype)
+L = form(inner(f, v) * dx + inner(g, v) * ds, dtype=dtype)
+
+A0 = assemble_matrix(a, [bc])
+b = assemble_vector(L)
+apply_lifting(b.array, [a], bcs=[[bc]])
+set_bc(b.array, [bc])
+
+A = csr_matrix((A0.data, A0.indices, A0.indptr))
```
garth-wells, line 84: Use https://github.com/FEniCS/dolfinx/blob/1c3831c908d74c2ca29d94f90ce1cd28ee71c07b/python/dolfinx/la.py#L135

garth-wells (commented):
This demo is more-or-less a copy of the existing Poisson demo. Could we combine, or maybe we could expand the demo to do more, e.g. add elasticity and used smoothed aggregation AMG?

python/pyproject.toml
jorgensd, line 38:
```suggestion
optional = ["numba"]
demos = ["numba", "pyamg"]
```

chrisrichardson: I'm just removing it for now.

chrisrichardson:
> This demo is more-or-less a copy of the existing Poisson demo. Could we combine, or maybe we could expand the demo to do more, e.g. add elasticity and used smoothed aggregation AMG?

Yes, now included. Probably needs a little reorganisation - maybe two functions that are called from main?

python/pyproject.toml
jorgensd: Removing it means that it will never be executed on CI?

garth-wells:
> > This demo is more-or-less a copy of the existing Poisson demo. Could we combine, or maybe we could expand the demo to do more, e.g. add elasticity and used smoothed aggregation AMG?
> 
> Yes, now included. Probably needs a little reorganisation - maybe two functions that are called from main?

Re-factoring into functions would be good, see related #3076.

PR #3204, opened 2024-05-10
Fix RedHat tests

jhale:
Doesn't seem to have the --check-build-dependencies option with this pip.


garth-wells:
There is a bit more than this to do. I’ve got changes for this in another PR.

PR #3209, opened 2024-05-13
Orient mesh entities consistently and generalise `entities_to_geometry` 

jpdean:
In `main`, the orientation of mesh entities is defined by whichever cell appears last in the entity-to-cell connectivity. This assumption is relied on in a few places in the code (see e.g. [`entities_to_geometry`](https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/mesh/utils.h#L680)), despite the assumption not being well documented.

This PR ensures that mesh entities are created such that their local (to the process) orientation agrees with their global orientation, which is unique, can be computed locally, and is a more natural definition. It will also greatly simplify mixed-dimensional coupling.

This PR also generalises `entities_to_geometry` for high-order elements. In `main`, this function computes the geometric DOFs associates with the vertices of a particular entity, whereas in this PR, it computes the geometric DOFs associated with the closure of the entities. I've removed the option `orient`, which only worked for the facets of tets, but let me know if this is problematic. I've added an option `permute`; this permutes the DOFs so that they agree with the orientation of mesh entities, but it requires `create_entity_permutations` to be called. In many cases, you don't need to do this (e.g. when computing the diameter of an entity). Alternatively, this could be implemented as a separate function.

Thanks @jorgensd for the idea to deal with high-order geometry via closure DOFs and @mscroggs for help with permuting the DOFs.


cpp/dolfinx/fem/CoordinateElement.h
```diff
@@ -93,6 +93,19 @@ class CoordinateElement
   void tabulate(int nd, std::span<const T> X, std::array<std::size_t, 2> shape,
                 std::span<T> basis) const;
 
+  /// @brief Permute a list of DOF indices associated with the closure of a cell
+  /// sub-entity.
```
garth-wells, line 97: Could you add some more explanation to the docstring on what this function does?

Helpful to include a mathematical description too, e.g. https://github.com/FEniCS/dolfinx/blob/ab79530e1ae60a646002609b510e8392cc09e213/cpp/dolfinx/fem/FiniteElement.h#L663.

cpp/dolfinx/mesh/topologycomputation.cpp
```diff
@@ -531,10 +531,39 @@ compute_entities_by_key_matching(
 
         // Get entity vertices. Padded with -1 if fewer than
         // max_vertices_per_entity
+        // NOTE Entity vertices are reordered so that their local (to
+        // the process) orientation agrees with their global orientation
+        // FIXME This might be better below when the entity to vertex
```
garth-wells, line 536: I can't quite follow this comment.

cpp/dolfinx/mesh/topologycomputation.h
```diff
@@ -30,7 +30,8 @@ class Topology;
 
 /// @brief Compute mesh entities of given topological dimension by
 /// computing entity-to-vertex connectivity (dim, 0), and cell-to-entity
-/// connectivity (tdim, dim).
+/// connectivity (tdim, dim). Computed entities are oriented such that their
```
garth-wells, line 33:
```suggestion
/// connectivity `(tdim, dim)`. 
///
/// Computed entities are oriented such that their
```

cpp/dolfinx/mesh/topologycomputation.h
```diff
@@ -30,7 +30,8 @@ class Topology;
 
 /// @brief Compute mesh entities of given topological dimension by
 /// computing entity-to-vertex connectivity (dim, 0), and cell-to-entity
```
garth-wells, line 32:
```suggestion
/// computing entity-to-vertex connectivity `(dim, 0)`, and cell-to-entity
```

cpp/dolfinx/mesh/utils.h
```diff
@@ -610,45 +604,35 @@ std::vector<std::int32_t> locate_entities_boundary(const Mesh<T>& mesh, int dim,
   return entities;
 }
 
-/// @brief Determine the indices in the geometry data for each vertex of
-/// the given mesh entities.
-///
-/// @warning This function should not be used unless there is no
-/// alternative. It may be removed in the future.
+/// @brief Compute the geometry degrees of freedom associated with
+/// the closure of a given set of entities.
```
garth-wells, line 608:
```suggestion
/// the closure of a given set of cell entities.
```

cpp/dolfinx/mesh/utils.h
```diff
@@ -610,45 +604,35 @@ std::vector<std::int32_t> locate_entities_boundary(const Mesh<T>& mesh, int dim,
...
-/// @param[in] entities Entity indices (local) to compute the vertex
-/// geometry indices for.
-/// @param[in] orient If true, in 3D, reorients facets to have
-/// consistent normal direction.
-/// @return Indices in the geometry array for the entity vertices. The
-/// shape is `(num_entities, num_vertices_per_entity)` and the storage
-/// is row-major. The index `indices[i, j]` is the position in the
-/// geometry array of the `j`-th vertex of the `entity[i]`.
+/// @param[in] entities Entity indices (local to process).
+/// @param[in] permute Permute the DOFs such that they are consistent with the
```
garth-wells, line 613:
```suggestion
/// @param[in] permute If `true`, permute the DOFs such that they are consistent with the
```

cpp/dolfinx/mesh/utils.h
```diff
@@ -610,45 +604,35 @@ std::vector<std::int32_t> locate_entities_boundary(const Mesh<T>& mesh, int dim,
...
+/// orientation of `dim`-dimensional mesh entities. This requires
+/// `create_entity_permutations` to be called first.
+/// @return Indices in the geometry array associated with the closure of each
```
garth-wells, line 616: What is the 'geometry array'?

cpp/dolfinx/mesh/utils.h
```diff
@@ -657,78 +641,60 @@ entities_to_geometry(const Mesh<T>& mesh, int dim,
...
-        std::swap(geometry_idx[i * num_vertices + 1],
-                  geometry_idx[i * num_vertices + 2]);
-      }
+    // Get the local index of the entity
+    std::span<const std::int32_t> cell_entities = c_to_e->links(c);
+    auto it = std::find(cell_entities.begin(), cell_entities.end(), e);
+    assert(it != cell_entities.end());
+    std::size_t local_entity = std::distance(cell_entities.begin(), it);
+
+    std::vector<std::int32_t> closure_dofs(
```
garth-wells, line 677: Would it be better if this could be moved outside of the loop?

cpp/dolfinx/mesh/utils.h
```diff
@@ -945,6 +911,93 @@ create_mesh(MPI_Comm comm, std::span<const std::int64_t> cells,
   }
 }
 
+/// @brief Create a sub-geometry for a subset of entities.
```
garth-wells, line 914: Add some more description.

cpp/dolfinx/mesh/utils.h
```diff
@@ -945,6 +911,93 @@ create_mesh(MPI_Comm comm, std::span<const std::int64_t> cells,
...
+/// @param mesh The full mesh.
```
garth-wells, line 915:
```suggestion
/// @param[in] mesh The full mesh.
```

python/dolfinx/mesh.py
```diff
@@ -757,24 +757,19 @@ def create_unit_cube(
...
         mesh: The mesh.
         dim: Topological dimension of the entities of interest.
-        entities: Entity indices (local to the process) to determine the
-            vertex geometry indices for.
-        orient: If True, the triangular facets of a 3D mesh will be reordered
-            so that they have a consistent normal direction. This option is likely
-            to be removed in the future.
+        entities: Entity indices (local to the process).
+        permute: Permute the DOFs such that they are consistent with the
+                 orientation of `dim`-dimensional mesh entities. This requires
```
garth-wells, line 769: Indentation looks wrong.

python/dolfinx/plot.py
```diff
@@ -64,7 +64,7 @@ def vtk_mesh(msh: mesh.Mesh, dim: typing.Optional[int] = None, entities=None):
         vtk_topology = _cpp.io.extract_vtk_connectivity(msh.geometry.dofmap, cell_type)[entities]
         num_nodes_per_cell = vtk_topology.shape[1]
     else:
-        # NOTE: This linearizes higher order geometries
+        # NOTE: This linearizes higher order geometries FIXME no longer true
```
garth-wells, line 67: Fix comment.

garth-wells (changes requested):

python/dolfinx/plot.py
jpdean: I'm not sure what the best thing is to do here. `entities_to_geometry` used to linearise geometry. This is no longer the case, so I doubt this code will work now if high-order geometry is used.

There are two options:
1) We could just get the DOFs associated with vertices from the output of `entities_to_geometry`
2) We generalise this function to work with high-order geometry

Option 2 is obviously better, but would need more work. What are your thoughts @jorgensd?

garth-wells: @jpdean do you have a change coming on this?

jorgensd: I can have a look at this if Joe doesn't have time.

garth-wells: > I can have a look at this if Joe doesn't have time.

Great. Let's not leave this PR hanging around.

jpdean: Sorry, I haven't had chance to look at this yet!

jorgensd: Plotting logic simplified alot.

jorgensd:
Example from: http://jsdokken.com/FEniCS23-tutorial/src/mesh_generation.html
now slightly modified:
```python
from mpi4py import MPI
import dolfinx.mesh
import dolfinx.plot
import basix.ufl

import ufl
import numpy as np
import pyvista
pyvista.start_xvfb(wait=0.1)



def plot_mesh(mesh, dim):
    mesh.topology.create_connectivity(dim, mesh.topology.dim)
    mesh.topology.create_entity_permutations()
    vtk_mesh = dolfinx.plot.vtk_mesh(mesh, dim=dim)
    linear_grid = dolfinx.plot.vtk_mesh(V)
    lgrid = pyvista.UnstructuredGrid(*linear_grid)

    ugrid = pyvista.UnstructuredGrid(*vtk_mesh)
    tgrid = ugrid.tessellate()
    plotter = pyvista.Plotter()
    plotter.add_mesh(ugrid,style="points", color="b", point_size=10, render_points_as_spheres=True)
    plotter.add_mesh(tgrid, show_edges=False, opacity=0.5)
    plotter.add_mesh(lgrid, style="wireframe", color="b")

    plotter.show()
    plotter.screenshot(f"grid_{dim}.png")

nodes = np.array([[1., 0.], [2., 0.], [3., 2.],
                  [2.9, 1.3], [1.5, 1.5], [1.5, -0.2]], dtype=np.float64)
connectivity = np.array([[0, 1, 2, 3, 4, 5]], dtype=np.int64)
c_el = ufl.Mesh(basix.ufl.element("Lagrange", "triangle", 2, shape=(2, )))
domain = dolfinx.mesh.create_mesh(MPI.COMM_SELF, connectivity, nodes, c_el)
plot_mesh(domain, 1)
plot_mesh(domain, 2)
```
yields:
Facet plot:
grid_1
Cell plot:
grid_2

garth-wells (approved):

jorgensd (approved):

cpp/dolfinx/mesh/utils.h
```diff
@@ -945,6 +914,97 @@ create_mesh(MPI_Comm comm, std::span<const std::int64_t> cells,
...
+                 });
+
+  // Create sub-geometry coordinate element
+  CellType sub_coord_cell
+      = cell_entity_type(geometry.cmap().cell_shape(), dim, 0);
+  fem::CoordinateElement<T> sub_cmap(sub_coord_cell, geometry.cmap().degree(),
+                                     geometry.cmap().variant());
+
+  // Sub-geometry input_global_indices
+  // TODO: Check this
```
jorgensd, line 994: Can the Todo be removed?

jpdean: Yep, I'll do that now

garth-wells (approved):

Issue #3213, opened 2024-05-15
`dolfinx.fem.Form.integral_types` nanobinding is incorrect.


jorgensd:
### Summarize the issue

Throws an error when accessed in Python

### How to reproduce the bug

See following mwe

### Minimal Example (Python)

```Python
from mpi4py import MPI
import numpy as np
import dolfinx
import ufl
if MPI.COMM_WORLD.rank == 0:
    print(dolfinx.__version__, dolfinx.git_commit_hash)


domain = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, 5, 6, ghost_mode=dolfinx.mesh.GhostMode.shared_facet)
form = dolfinx.fem.form(1*ufl.dx(domain=domain))
print(form.integral_types)
```

### Output (Python)

```bash
0.9.0.0 dbbbf9167f662f4bcc2c45da5a5998433ae24863
Traceback (most recent call last):
  File "/root/shared/debug/mwe.py", line 11, in <module>
    print(form.integral_types)
          ^^^^^^^^^^^^^^^^^^^
  File "/root/shared/dolfinx/python/dolfinx/fem/forms.py", line 94, in integral_types
    return self._cpp_object.integral_types
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Unable to convert function return value to a Python type! The signature was
    (self) -> std::set<dolfinx::fem::IntegralType, std::less<dolfinx::fem::IntegralType>, std::allocator<dolfinx::fem::IntegralType> >
```

### Version

main branch

### DOLFINx git commit

dbbbf9167f662f4bcc2c45da5a5998433ae24863

### Installation

Using docker (test-env/current) with local installation

PR #3215, opened 2024-05-15
Simplify integration domain packing

### LINKED ISSUE #3214: interior_facet_assembly causes deadlock if there are no interior facets on the process
### Summarize the issue

Issue reported at: 
https://fenicsproject.discourse.group/t/scallar-assembly-of-a-internal-surface-integral-misbehaves-in-parallel/14655/3
The problem is that not all processes get to `create_entity_permutations`
https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/assemble_scalar_impl.h#L190-L192
which calls `mesh.topology.create_entities(d)` for d=0,....,mesh.topology.dim.

In some cases, `create_entities(d)` needs parallel communication, causing a deadlock.

We need a code re-design to avoid these deadlock situations, as we cannot have `create_entity_permutations` inside loops that needs to be true on all processes.

To fix this for assembly, I would change how we compute `integration_entities`, i.e,
https://github.com/FEniCS/dolfinx/blob/b16ef8abc20d70c28d8f1850fe42a0ee0bcf5106/cpp/dolfinx/fem/utils.cpp#L131-L151
as we currently send in a full meshtags object, which doesn't have the subdomain ids from the form in question
https://github.com/FEniCS/dolfinx/blob/b16ef8abc20d70c28d8f1850fe42a0ee0bcf5106/python/dolfinx/fem/forms.py#L214-L215
I think this should take the meshtag information, and the id's we are interested in extracting (currently we extract way more than we need). This would yield integrals on all processes (with 0 facets on some).

### How to reproduce the bug

Run following code with 1, 2 and 4 processes.
Works on 1 and 2 processes, deadlocks on 4.

### Minimal Example (Python)

```Python
from mpi4py import MPI
import numpy as np
import dolfinx
import ufl
if MPI.COMM_WORLD.rank == 0:
    print(dolfinx.__version__, dolfinx.git_commit_hash)


domain = dolfinx.mesh.create_unit_cube(MPI.COMM_WORLD, 5, 6,6, ghost_mode=dolfinx.mesh.GhostMode.shared_facet)

def marker(x, eps=1e-14):
    return np.logical_and(x[0]<=0.5+eps,np.isclose(x[1], 0.5))

domain.topology.create_connectivity(domain.topology.dim - 1, domain.topology.dim)
domain.topology.create_connectivity(domain.topology.dim, domain.topology.dim-1)

facet_map = domain.topology.index_map(domain.topology.dim - 1)
num_facets_on_proc = facet_map.size_local + facet_map.num_ghosts
facets = np.arange(num_facets_on_proc, dtype=np.int32)
facet_values = np.zeros(num_facets_on_proc, dtype=np.int32)
facet_values[dolfinx.mesh.locate_entities(domain, domain.topology.dim - 1, marker)] = 2
ft = dolfinx.mesh.meshtags(domain, domain.topology.dim - 1, facets, facet_values)
dS = ufl.Measure("dS", domain=domain, subdomain_data=ft, subdomain_id=2)
form = dolfinx.fem.form(1*dS)


print(dolfinx.fem.assemble_scalar(form))
```

### Output (Python)

```bash
0.9.0.0 dbbbf9167f662f4bcc2c45da5a5998433ae24863
```

### Version

main branch

### DOLFINx git commit

dbbbf9167f662f4bcc2c45da5a5998433ae24863

### Installation

_No response_

### Additional information

_No response_

jorgensd:
Resolve #3214 and #3213.

This PR changes the packing such that we only pack integral entities for those subdomain id's that exist in the integrals.
For processors with no entities for a given tag, we return the tuple (id, []) indicating that there is an integral of this type on the process, but that it has no entities.

This simplifies the packing of integration entities massively, as we only pack those specified by the form data, not for every value in the meshtags. It also returns a cleaner data-structure that can be used directly by users.

### DIFF
--- a/cpp/dolfinx/fem/utils.h
+++ b/cpp/dolfinx/fem/utils.h
@@ -422,6 +422,7 @@ Form<T, U> create_form_factory(
       kern_t k = nullptr;
       if constexpr (std::is_same_v<T, float>)
         k = integral->tabulate_tensor_float32;
+#ifndef DOLFINX_NO_STDC_COMPLEX_KERNELS
       else if constexpr (std::is_same_v<T, std::complex<float>>)
       {
         k = reinterpret_cast<void (*)(
@@ -429,8 +430,10 @@ Form<T, U> create_form_factory(
             const typename scalar_value_type<T>::value_type*, const int*,
             const unsigned char*)>(integral->tabulate_tensor_complex64);
       }
+#endif // DOLFINX_NO_STDC_COMPLEX_KERNELS
       else if constexpr (std::is_same_v<T, double>)
         k = integral->tabulate_tensor_float64;
+#ifndef DOLFINX_NO_STDC_COMPLEX_KERNELS
       else if constexpr (std::is_same_v<T, std::complex<double>>)
       {
         k = reinterpret_cast<void (*)(
@@ -438,6 +441,8 @@ Form<T, U> create_form_factory(
             const typename scalar_value_type<T>::value_type*, const int*,
             const unsigned char*)>(integral->tabulate_tensor_complex128);
       }
+#endif // DOLFINX_NO_STDC_COMPLEX_KERNELS
+
       if (!k)
       {
         throw std::runtime_error(
@@ -486,6 +491,7 @@ Form<T, U> create_form_factory(
       kern_t k = nullptr;
       if constexpr (std::is_same_v<T, float>)
         k = integral->tabulate_tensor_float32;
+#ifndef DOLFINX_NO_STDC_COMPLEX_KERNELS
       else if constexpr (std::is_same_v<T, std::complex<float>>)
       {
         k = reinterpret_cast<void (*)(
@@ -493,8 +499,10 @@ Form<T, U> create_form_factory(
             const typename scalar_value_type<T>::value_type*, const int*,
             const unsigned char*)>(integral->tabulate_tensor_complex64);
       }
+#endif // DOLFINX_NO_STDC_COMPLEX_KERNELS
       else if constexpr (std::is_same_v<T, double>)
         k = integral->tabulate_tensor_float64;
+#ifndef DOLFINX_NO_STDC_COMPLEX_KERNELS
       else if constexpr (std::is_same_v<T, std::complex<double>>)
       {
         k = reinterpret_cast<void (*)(
@@ -502,6 +510,7 @@ Form<T, U> create_form_factory(
             const typename scalar_value_type<T>::value_type*, const int*,
             const unsigned char*)>(integral->tabulate_tensor_complex128);
       }
+#endif // DOLFINX_NO_STDC_COMPLEX_KERNELS
       assert(k);
 
       // Build list of entities to assembler over
@@ -557,6 +566,7 @@ Form<T, U> create_form_factory(
       kern_t k = nullptr;
       if constexpr (std::is_same_v<T, float>)
         k = integral->tabulate_tensor_float32;
+#ifndef DOLFINX_NO_STDC_COMPLEX_KERNELS
       else if constexpr (std::is_same_v<T, std::complex<float>>)
       {
         k = reinterpret_cast<void (*)(
@@ -564,8 +574,10 @@ Form<T, U> create_form_factory(
             const typename scalar_value_type<T>::value_type*, const int*,
             const unsigned char*)>(integral->tabulate_tensor_complex64);
       }
+#endif // DOLFINX_NO_STDC_COMPLEX_KERNELS
       else if constexpr (std::is_same_v<T, double>)
         k = integral->tabulate_tensor_float64;
+#ifndef DOLFINX_NO_STDC_COMPLEX_KERNELS
       else if constexpr (std::is_same_v<T, std::complex<double>>)
       {
         k = reinterpret_cast<void (*)(
@@ -573,6 +585,7 @@ Form<T, U> create_form_factory(
             const typename scalar_value_type<T>::value_type*, const int*,
             const unsigned char*)>(integral->tabulate_tensor_complex128);
       }
+#endif // DOLFINX_NO_STDC_COMPLEX_KERNELS
       assert(k);
 
       // Build list of entities to assembler over
@@ -665,6 +678,7 @@ Form<T, U> create_form(
   }
 
   // Place constants in appropriate order
+
   std::vector<std::shared_ptr<const Constant<T>>> const_map;
   for (const std::string& name : get_constant_names(ufcx_form))
   {
@@ -1075,6 +1089,7 @@ Expression<T, U> create_expression(
       tabulate_tensor = nullptr;
   if constexpr (std::is_same_v<T, float>)
     tabulate_tensor = e.tabulate_tensor_float32;
+#ifndef DOLFINX_NO_STDC_COMPLEX_KERNELS
   else if constexpr (std::is_same_v<T, std::complex<float>>)
   {
     tabulate_tensor = reinterpret_cast<void (*)(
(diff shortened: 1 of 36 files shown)


python/dolfinx/fem/forms.py
```diff
@@ -199,29 +200,52 @@ def _form(form):
...
                         subdomain._cpp_object.topology.create_connectivity(tdim - 1, tdim)
                         subdomain._cpp_object.topology.create_connectivity(tdim, tdim - 1)
-                    domains = _cpp.fem.compute_integration_domains(
-                        integral_type, subdomain._cpp_object
-                    )
+                    # Compute integration domains only for each subdomain id in the integrals
+                    # If a process has no integral entities, insert an empty array
+                    for id in subdomain_ids:
+                        try:
+                            integration_entities = _cpp.fem.compute_integration_domains(
```
jpdean, line 227: It seems a bit strange to loop over the IDs in Python and call `compute_integration_entities` on each when `compute_integration_domains` loops over IDs anyway. Maybe the fix should happen inside `compute_integration_domains` instead so that it returns `[..., (id, []), ....]` when there are no entities to integrate over for that ID? That way, we can just call it once and not have to pass `subdomain.values[subdomain.values == id]`, which is just a (possibly quite large) array filled with the integral ID. If there's a reason we can't do this, then maybe we should change `compute_integration_domains` so that it only does one ID at a time, but I think the first option is better.

python/dolfinx/fem/forms.py
```diff
@@ -199,29 +200,52 @@ def _form(form):
         ]
         constants = [c._cpp_object for c in form.constants()]
 
+        # Make map from integral_type to the meshtag indices of interest
+        subdomain_ids = {type: [] for type in sd.get(domain).keys()}
+        for integral in form.integrals():
+            if integral.subdomain_data() is not None:
+                subdomain_ids[integral.integral_type()].append(integral.subdomain_id())
+        subdomain_ids = {key: list(chain(values)) for key, values in subdomain_ids.items()}
```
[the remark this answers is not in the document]
jorgensd: We can get multiple subdomain_ids as tuples, or a s a single integer when we extract the subdomain_id. Thus we need to flatten this list of lists/list of tuples/list of floats.

python/dolfinx/fem/forms.py
jorgensd: `compute_integration_domains` only loops over ids present on the current process. It has no notion of what is on the other processes, or what is in the integral form. We cannot loop over every possible id (as it is an int 32).
Preferably, I would change `compute_integration_entities` to only take in the id of the entities, i.e.
```c++
/// @param[in] integral_type Integral type
/// @param[in] topology Mesh topology
/// @param[in] entities List of tagged mesh entities
/// @param[in] dim Topological dimension of tagged entities
/// @return List of `(integral id, entities)` pairs
/// @pre The topological dimension of the integral entity type and the
/// topological dimension of mesh tag data must be equal.
/// @pre For facet integrals, the topology facet-to-cell and
/// cell-to-facet connectivity must be computed before calling this
/// function.
std::vector<std::int32_t>
compute_integration_domains(IntegralType integral_type,
                            const mesh::Topology& topology,
                            std::span<const std::int32_t> entities, int dim);
```
then the user can associate what-ever value with the integration entities.

jorgensd:
@jpdean Thanks for your comments. I've addressed them by simplifying `compute_integration_entities`.

jorgensd:
Is there anything specific holding this PR back? 
it simplifies code and resolves multiple issues.

garth-wells:
> Is there anything specific holding this PR back? it simplifies code and resolves multiple issues.

Wait for @jpdean to return and add approval.

python/dolfinx/fem/forms.py
```diff
@@ -199,29 +200,49 @@ def _form(form):
         ]
         constants = [c._cpp_object for c in form.constants()]
 
+        # Make map from integral_type to subdomain id
+        subdomain_ids = {type: [] for type in sd.get(domain).keys()}
+        for integral in form.integrals():
+            if integral.subdomain_data() is not None:
+                subdomain_ids[integral.integral_type()].append(integral.subdomain_id())
+        subdomain_ids = {key: list(chain(values)) for key, values in subdomain_ids.items()}
```
jpdean, line 208: Could `key` and `values` be renamed to something a bit more informative?

cpp/dolfinx/fem/utils.cpp
```diff
@@ -215,42 +206,6 @@ fem::compute_integration_domains(fem::IntegralType integral_type,
           "Cannot compute integration domains. Integral type not supported.");
     }
   }
-
-  // Build permutation that sorts by meshtag value
```
jpdean, line 219: It doesn't look like any sort is happening in `forms.py`, which is fine as long as the integration data isn't assumed to be sorted anywhere. Are we sure that there are no assumptions that integration data should be sorted by meshtag values?

jpdean (commented):
I think letting `compute_integration_domains` just deal with turning a list of mesh entities into data needed for the assember (instead of worrying about meshtag values) is a good idea. It fits the DOLFINx design philosophy better than the current function. One downside is that it basically does nothing for cells now, but that's not a big issue. It might make user C++ code a bit more complex though, because they'll have to loop over integration domains and call this function manually. Overall, I prefer it to the current approach.

cpp/dolfinx/fem/utils.cpp
jorgensd: I’ve never seen such an assumption in the code. All tests pass, so i dont think it matters

cpp/dolfinx/fem/utils.h
```diff
@@ -99,18 +99,16 @@ get_cell_facet_pairs(std::int32_t f, std::span<const std::int32_t> cells,
 /// @param[in] topology Mesh topology
 /// @param[in] entities List of tagged mesh entities
 /// @param[in] dim Topological dimension of tagged entities
-/// @param[in] values Value associated with each entity
-/// @return List of `(integral id, entities)` pairs
+/// @return List of integration entities
 /// @pre The topological dimension of the integral entity type and the
 /// topological dimension of mesh tag data must be equal.
```
jorgensd, line 104:
```suggestion
```

jorgensd:
> It might make user C++ code a bit more complex though, because they'll have to loop over integration domains and call this function manually. Overall, I prefer the approach in this PR to the one in main.

As I've seen no C++ example where this function is actually used, I don't think we should worry to much.
The current code does way less work for problems where the `meshtags` are specified for all entities.
For instance, imagine that you have a facet marker where each facet has a unique number (local to process). The function on main will pack integration data for every facet on the process, while the new function will pack data only for those present in `subdomain_ids`.

cpp/dolfinx/fem/utils.cpp
jorgensd: I've added in a sort for good measure

jpdean (approved):
Looks good to me! One thing I'd like to do is simplify this logic if possible: https://github.com/FEniCS/dolfinx/blob/159cb1c845f53d67b30bd6fa130534ad82171845/python/dolfinx/fem/forms.py#L203-L242
It was already complicated in main, and this PR adds more complexity. Happy to look at this in a separate PR though; maybe we should make an issue?

jorgensd:
> Looks good to me! One thing I'd like to do is simplify this logic if possible:
> 
> https://github.com/FEniCS/dolfinx/blob/159cb1c845f53d67b30bd6fa130534ad82171845/python/dolfinx/fem/forms.py#L203-L242
> 
> It was already complicated in main, and this PR adds more complexity. Happy to look at this in a separate PR though; maybe we should make an issue?

We would have to change how ufl exposes integrals to touch upon this (I think).
It should be a separate issue.

Issue #3195, opened 2024-05-04
Change C++ logging library to `spdlog` or `glog`


garth-wells:
Some time ago we adopted [loguru](https://github.com/emilk/loguru) for C++ logging. At the time we considered several options (see https://github.com/FEniCS/dolfinx/issues/3). The history is a bit hard to follow, but we did have a PR for using [spdlog](https://github.com/gabime/spdlog) (https://github.com/FEniCS/dolfinx/pull/311), but eventually adopted loguru (https://github.com/FEniCS/dolfinx/pull/417).

A consideration at the time was a reluctance to introduce a substantial dependency that was not widely packaged. loguru is light and we included the source in the DOFLINx tree.

In the meantime, 
- loguru is not actively maintained. The code has many pragmas to ignore compiler warnings, we're having to edit the code to add more as compilers add stricter tests.
- spdlog and [glog](https://github.com/google/glog) are actively maintained and widely used.
- spdlog and glog are available in conda, brew, Debian/Ubuntu, Spack, vcpkg , and other package managers

Should we switch from loguru to spdlog or glog? 

@jhale, @chrisrichardson

chrisrichardson:
glog is also worth considering. It would be less disruptive as the syntax is the same as what we have in loguru.
loguru works well, so I'd like to confirm that the same or better features exist before switching.

jhale:
I support the change but am also interested in glog.

Both glog and spdlog are available in vcpkg, homebrew, Debian, RedHat, conda (feedstock) and Spack so no issues there.

glog (unlike spdlog) has a Windows CI. It is also a drop in replacement for loguru.

We are not logging heavily in tight loops so performance differences are not hugely important for us.

@chrisrichardson and I have nearly finished the Windows build and have identified loguru as producing strange linking (!) errors when compiling the Python headers, so we're keen to make a switch to something more modern.

garth-wells:
I've updated the title to add glog to spdlog.

chrisrichardson:
https://github.com/FEniCS/dolfinx/tree/chris/glog - seems an easy drop-in.

chrisrichardson:
On further testing, I think spdlog is a better choice:
https://github.com/FEniCS/dolfinx/compare/chris/spdlog

PR #3216, opened 2024-05-15
Switch to spdlog

chrisrichardson:
* Remove `loguru` and replace with `spdlog`


jhale:
This looks good to me apart from having to turn off some warnings. Would it be possible to only turn them off only on GCC? Then we'll still might get some coverage on LLVM/clang toolchains like oneAPI, or people developing on macOS.

chrisrichardson:
Yes I think so.

garth-wells:
> This looks good to me apart from having to turn off some warnings. Would it be possible to only turn them off only on GCC (I was told by Chris this only comes up on GCC)? Then we'll still might get some coverage on LLVM/clang toolchains like oneAPI, or people developing on macOS.

I don't need to disable the warnings with gcc-14.

garth-wells (approved):
Will need to add spdlog to Dockerfiles, Spack, Debian and conda.

chrisrichardson:
I already added to Docker files, but will need rebuilding.

jhale:
Looks good to me, we can deal with packaging afterwards.

garth-wells:
Resolves #3195

PR #3219, opened 2024-05-17
Start cleanup of Python dependencies.

jhale:


python/pyproject.toml
```diff
@@ -33,20 +33,29 @@ dependencies = [
...
-      "mypy",
-      "pytest-xdist",
-      "types-setuptools",
-      "fenics-dolfinx[build]",
+      "fenics-dolfinx[demos]",
       "fenics-dolfinx[docs]",
       "fenics-dolfinx[lint]",
-      "fenics-dolfinx[optional]",
       "fenics-dolfinx[test]",
+      "fenics-dolfinx[optional]",
```
jorgensd, line 58: I think this could be simplified to
```toml
fenics-dolfinx[demos,docs,lint,test,optional]
```

garth-wells:
I think it's better to do this in two steps, (1) simplify Dockerfiles and do 'pip install foo' in the CI yaml, then (2) handle the pyproject.yaml file.

garth-wells:
Closing because Dockerfile changes have been made in other PRs. Next step will be installing dependencies via pip automatically. This will test that dependencies are correct in the `pyproject.yaml` file.

Issue #3208, opened 2024-05-13
[BUG]: `create_entity_permutations` occasionally segfaults for prism meshes


jpdean:
### Summarize the issue

Calling `create_entity_permutations` occasionally segfaults for prism meshes. For some reason, I can only reproduce this on a clean Ubuntu install, it seems to run without failing on the docker images and on my local install.

### How to reproduce the bug

Run the minimal example below on a minimal Ubuntu install as described in the build and test workflow [here](https://github.com/FEniCS/dolfinx/blob/main/.github/workflows/ccpp.yml#L84).

### Minimal Example (Python)

```Python
from mpi4py import MPI
from dolfinx import mesh

n = 3
msh = mesh.create_unit_cube(MPI.COMM_WORLD, n, n, n, cell_type=mesh.CellType.prism)
msh.topology.create_entity_permutations()
```

### Version

main branch

### Installation

Follow the build and run workflow file [here](https://github.com/FEniCS/dolfinx/blob/main/.github/workflows/ccpp.yml#L84) using a plain Ubuntu image.

chrisrichardson:
I can reproduce this bug.
I think it stems from `compute_face_permutations` in `permutationcomputation.cpp` where there are separate computations for simplices and 'tp' elements. However, I don't think `prism` is being properly treated.

chrisrichardson:
Yes, confirmed: using `auto f_to_v = topology.connectivity(2, 0);` is not enough for prism cells, now they contain two types of facet, we need to use `auto f0_to_v = topology.connectivity({2, 0}, {0, 0});` and `auto f1_to_v = topology.connectivity({2, 1}, {0, 0});`
Probably this function (`compute_face_permutations`) needs general refactoring.

chrisrichardson:
Fixed by  #3223

PR #3223, opened 2024-05-18
Permutation computation refactoring

chrisrichardson:
Generalise face permutation computation to get ready for mixed topology meshes.
* reduce duplicated code in face permutation computation for tet, hex
* adds computation for prism, (should work for pyramid too)

### DIFF
--- a/cpp/dolfinx/mesh/permutationcomputation.cpp
+++ b/cpp/dolfinx/mesh/permutationcomputation.cpp
@@ -10,6 +10,8 @@
 #include <algorithm>
 #include <bitset>
 #include <dolfinx/common/IndexMap.h>
+#include <dolfinx/common/Timer.h>
+#include <dolfinx/common/log.h>
 #include <dolfinx/graph/AdjacencyList.h>
 
 namespace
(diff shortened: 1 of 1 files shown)


chrisrichardson:
Would be good to add a test - if anyone can think of a good one @mscroggs @jpdean ?

cpp/dolfinx/mesh/permutationcomputation.cpp
```diff
@@ -21,205 +23,185 @@ using namespace dolfinx;
...
+  // lowest numbered vertex
+  std::int32_t pre = e_vertices[prev[min_v]];
+
+  // post is the (local) number of the next vertex anticlockwise
+  // from the lowest numbered vertex
+  std::int32_t post = e_vertices[prev[3 - min_v]];
+
+  // If min_v is 2 or 3, swap. Why?
+  if (min_v == 2 or min_v == 3)
+    min_v = 5 - min_v;
```
mscroggs, line 89: To answer the why: min_v wants to be the number of vertices anticlockwise away the minimum numbered vertex is. If the vertices were numbered:

```
3 -- 2
|    |
0 -- 1
```

We wouldn't need to do the swap, but as we use the numbering:

```
2 -- 3
|    |
0 -- 1
```

we therefore need the swap

chrisrichardson: Thanks @mscroggs . Maybe you can add something in-line to document this.

chrisrichardson:
Fixes issue #3208

PR #3224, opened 2024-05-19
Add support for mixed-domain problems of codimension 1

jpdean:
This PR adds support for solving mixed-domain problems of codimension 1. This functionality can be used to, for example, implement a range of coupling schemes, Lagrange multiplier problems, and hybridised discontinuous Galerkin methods.

I've added a simple HDG demo for Poisson's equations, but in a future PR I'll change this to an HDG Stokes or Navier-Stokes solver with some nice stability and conservation properties. 

This PR requires https://github.com/FEniCS/ffcx/pull/675.


cpp/dolfinx/fem/Form.h
```diff
@@ -359,8 +359,42 @@ class Form
...
       case IntegralType::exterior_facet:
-        // Exterior and interior facets are treated the same
-        [[fallthrough]];
+      {
+        // Get the codimension of the mesh
+        const int tdim = _mesh->topology()->dim();
+        const int codim = tdim - mesh.topology()->dim();
+        assert(codim >= 0);
+
+        if (codim == 0)
```
garth-wells, line 368: Should ` mapped_entities.reserve(...)` be moved inside the case? Looks like the reserved length should be doubled for exterior facets.

jpdean: No; `mapped_entities.reserve(entities.size())` is correct because `entities` contains (flattened) `(cell, local_facet)` pairs for exterior facet integrals.

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -163,6 +163,7 @@ void assemble_cells(
 /// matrix.
 /// @param x_dofmap Dofmap for the mesh geometry.
 /// @param x Mesh geometry (coordinates).
+/// @param num_cell_facets Number of cell facets
```
jorgensd, line 166: Should we call this `num_facets_per_cell`, this would be more consistent with what we use elsewhere:
https://github.com/search?q=repo%3AFEniCS%2Fdolfinx%20_per_&type=code

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -534,51 +541,49 @@ void assemble_matrix(
...
+        a.domain(IntegralType::exterior_facet, i),
         {dofs0, bs0, a.domain(IntegralType::exterior_facet, i, *mesh0)}, P0,
         {dofs1, bs1, a.domain(IntegralType::exterior_facet, i, *mesh1)}, P1T,
-        bc0, bc1, fn, coeffs, cstride, constants, cell_info0, cell_info1);
+        bc0, bc1, fn, coeffs, cstride, constants, cell_info0, cell_info1,
+        get_perm);
   }
 
-  if (a.num_integrals(IntegralType::interior_facet) > 0)
+  for (int i : a.integral_ids(IntegralType::interior_facet))
```
jorgensd, line 573: Happy to remove the if statement (as it is not good on multiple processes, ref: https://github.com/FEniCS/dolfinx/pull/3215).

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -163,6 +163,7 @@ void assemble_cells(
 /// matrix.
 /// @param x_dofmap Dofmap for the mesh geometry.
 /// @param x Mesh geometry (coordinates).
+/// @param num_facets_per_cell Number of cell facets
```
garth-wells, line 166:
```suggestion
/// @param[in] num_facets_per_cell Number facets per cell.
```

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -185,10 +186,11 @@ void assemble_cells(
 /// mesh
 /// @param cell_info1 The cell permutation information for the trial function
 /// mesh
+/// @param get_perm Function to get the facet permutations
```
garth-wells, line 189: Make the docstring clearer. 

I'm not sure we want to pass a `std::function` that does very little work - there can be an overhead.

garth-wells (changes requested):
A few comments after a quick first pass.

garth-wells (approved):

jorgensd:
I don't think this should be merged prior to: #3260 as it should simplify some conditionals in the coefficient packing/extraction. What do you think @jpdean ?

cpp/dolfinx/fem/utils.h
```diff
@@ -997,8 +997,19 @@ void pack_coefficients(const Form<T, U>& form, IntegralType integral_type,
...
         assert(mesh);
+
+        // Other integrals in the form might have coefficients defined over
+        // entities of codim > 0, which don't make sense for cell integrals, so
+        // don't pack them.
+        const int codim
+            = form.mesh()->topology()->dim() - mesh->topology()->dim();
+        assert(codim >= 0);
+        if (codim != 0)
+          continue;
```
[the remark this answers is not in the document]
jorgensd: The behavior here is corrected in #3260

Issue #3212, opened 2024-05-14
Remove Fides file output


garth-wells:
Fides doesn't add anything over other formats (less, in fact), is more code to maintain and I doubt it is used.

Should we remove it?

@jorgensd

jorgensd:
I've not used it for a very long time, let's keep it until we have a PR for VTKHDF, and then remove it (and let other users chime in).

francesco-ballarin:
I am in favor of removing fides file output as soon as possible. Ping @massimiliano-leoni who may have an opinion on this topic.

mleoni-pf:
Thanks for pinging me!

I am mainly using VTX and Fides nowadays, the former for high-order elements and the latter for first-order elements. I could admittedly use VTX for everything, Fides is much faster to read in Paraview but that's not really our concern. Does VTX support all function types and elements that Fides supports?

jorgensd:
> Thanks for pinging me!
> 
> I am mainly using VTX and Fides nowadays, the former for high-order elements and the latter for first-order elements. I could admittedly use VTX for everything, Fides is much faster to read in Paraview but that's not really our concern. Does VTX support all function types and elements that Fides supports?

I think so yes. Fides is limited to first order, so it doesn’t really yield much.

mleoni-pf:
In that case I will just use VTX in the future.

PR #2848, opened 2023-11-01
Update computation of number of components for arbitrary value shape in `xdmf_function`

IgorBaratta:
Fixes #2843

PR #3227, opened 2024-05-22
Generalize all output formats for general length vectors and tensors

jorgensd:
Reported at: https://fenicsproject.discourse.group/t/saving-vector-function-with-length-greater-than-the-geometrical-dimension/14752
and follows 
https://github.com/FEniCS/dolfinx/pull/3095
https://github.com/FEniCS/dolfinx/pull/2848
in supporting arbitrary tensors.

Note that Fides does not support tensors that does not have the shape 3**rank, so I've added a runtime error for it. It is also discussed if we want to remove Fides https://github.com/FEniCS/dolfinx/issues/3212 due to its limited capabilities, so I don't think it is worth to pursue upstream.


francesco-ballarin (approved):

cpp/dolfinx/io/ADIOS2Writers.h
```diff
@@ -249,8 +249,17 @@ std::vector<T> pack_function_data(const fem::Function<T, U>& u)
   std::uint32_t num_vertices
       = vertex_map->size_local() + vertex_map->num_ghosts();
 
-  int rank = V->value_shape().size();
-  std::uint32_t num_components = std::pow(3, rank);
+  auto value_shape = u.function_space()->value_shape();
```
garth-wells, line 252: Best to avoid auto unless the return type is obvious or very complicated.

cpp/dolfinx/io/VTKFile.cpp
```diff
@@ -110,12 +110,12 @@ void add_pvtu_mesh(pugi::xml_node& node)
 
 /// Add float data to a pugixml node
 /// @param[in] name The name of the data array
-/// @param[in] rank The rank of the field, e.g. 1 for a vector and 2 for
-/// a tensor
+/// @param[in] num_components Number of components in the field. If scalar this
```
garth-wells, line 113: Setting the number of components to zero doesn't make sense.

`values` holds the length of the data.

cpp/dolfinx/io/VTKFile.cpp
```diff
@@ -477,8 +477,16 @@ void write_function(
     auto V = _u.get().function_space();
     auto e = V->element();
     assert(e);
-    int rank = V->value_shape().size();
-    std::int32_t num_comp = std::pow(3, rank);
+
+    // Pad to 3D if vector/tensor is product of dimensions is smaller than
+    // 3**rank to ensure that we can visualize them correctly in Paraview
+    auto value_shape = V->value_shape();
```
garth-wells, line 483: See previous comment on `auto`.

cpp/dolfinx/io/xdmf_function.cpp
```diff
@@ -75,13 +75,14 @@ void xdmf_function::add_function(MPI_Comm comm, const fem::Function<T, U>& u,
   assert(dofmap);
   const int bs = dofmap->bs();
 
-  std::span<const std::size_t> value_shape = u.function_space()->value_shape();
+  // Pad to 3D if vector/tensor is product of dimensions is smaller than 3**rank
+  // to ensure that we can visualize them correctly in Paraview
+  auto value_shape = u.function_space()->value_shape();
```
garth-wells, line 80: See above.

garth-wells (dismissed):

cpp/dolfinx/io/VTKFile.cpp
jorgensd: So what do you want to send in here then? We need an indicator for how many components a vector/tensor has. If scalar we do not want to add `NumberOfComponents`.

garth-wells: Something that makes sense, and a clear docstring. Passing 0 for the number of components when the number of components is 1 doesn't make sense - the docstring is clearly wrong.

jorgensd: Rewritten to use `value_shape`, in the same way we use `value_shape` for anything else in dolfinx, i.e. an empty span indicates scalar value.

PR #3229, opened 2024-05-24
Simplify Dockerfile packages

garth-wells:
Move non-core packages to CI yaml.


docker/Dockerfile.test-env
```diff
@@ -144,10 +145,14 @@ RUN if [ "$MPI" = "mpich" ]; then \
     ldconfig && \
     rm -rf /tmp/*
 
+# Create virtual env
```
francesco-ballarin, line 148: To avoid having to pass `--break-system-packages`, an even simpler trick that doesn't require creating a venv is to `rm /usr/lib/python3.12/EXTERNALLY-MANAGED`

PR #3232, opened 2024-05-28
Docker simplify (phase 2)

garth-wells:
Use a virtual env in the Dockerfile for installing dependencies. This is nicer than overwriting system Python packages when installing/updating. Will also allow advanced users to test Python package versions more freely.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -133,9 +133,9 @@ jobs:
       - name: Install Python demo/test dependencies
         run: pip install matplotlib pyamg pytest pytest-xdist scipy
       - name: Run demos (Python, serial)
-        run: python3 -m pytest -n auto -m serial --durations=10 python/demo/test.py
+        run: python -m pytest -n auto -m serial --durations=10 python/demo/test.py
       - name: Run demos (Python, MPI (np=3))
-        run: python3 -m pytest -m mpi --num-proc=3 python/demo/test.py
+        run: python -m pytest -m mpi --num-proc=3 python/demo/test.py
       - name: Run unit tests
         run: python -m pytest -n auto -m "not petsc4py and not adios2" python/test/unit
       - name: Run unit tests (MPI, np=3)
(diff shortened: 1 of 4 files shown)


jorgensd:
I think this is a rather intrusive change, as all CIs must change from `python3` to `python` for running code. Can't we use an alias so that we keep some backwards compatibility? I am not saying we should use venv, but is there a point of having another `python3` in that environment, as it is bound to cause confusion.

All my CIs will now fail and need adjustment, and those that are customized to test against both stable and nightly can no longer be merged into a build matrix without extra arguments.

garth-wells:
Why would a change `python3` -> `python` be required?

jorgensd:
> Why would a change `python3` -> `python` be required?

I inferred that from the changes in the CI, where you have replaced `python3 -m pip` with `pip` and `python3` with `python`. Please let me know if I am mistaken, or if there is another motivation for this change.

francesco-ballarin:
Will the script at https://github.com/FEniCS/dolfinx/blob/main/docker/dolfinx-real-mode (and similarly `dolfinx-complex-mode`) need adjustament?

I guess I am a bit confused of what having a `venv` will mean now for real vs complex build.

garth-wells:
> > Why would a change `python3` -> `python` be required?
> 
> I inferred that from the changes in the CI, where you have replaced `python3 -m pip` with `pip` and `python3` with `python`. Please let me know if I am mistaken, or if there is another motivation for this change.

There should be no change required. 

`python3 -m pip` is visual noise when `pip` is enough.

jorgensd:
> > > Why would a change `python3` -> `python` be required?
> > 
> > 
> > I inferred that from the changes in the CI, where you have replaced `python3 -m pip` with `pip` and `python3` with `python`. Please let me know if I am mistaken, or if there is another motivation for this change.
> 
> There should be no change required.
> 
> `python3 -m pip` is visual noise when `pip` is enough.

I dont necessarily agree, as by running pip you do not know what Python installation you are using.

Good to know that no changes are required. Is there a specific reason for the build docker and test Ci failing

PR #3234, opened 2024-05-29
Support mixed topology for dual graph

chrisrichardson:
Build dual graph for meshes with mixed cells of same dimension.

### DIFF
--- a/cpp/dolfinx/mesh/graphbuild.cpp
+++ b/cpp/dolfinx/mesh/graphbuild.cpp
@@ -485,6 +485,161 @@ mesh::build_local_dual_graph(CellType celltype,
           std::move(local_cells)};
 }
 //-----------------------------------------------------------------------------
+std::tuple<graph::AdjacencyList<std::int32_t>, std::vector<std::int64_t>,
+           std::size_t, std::vector<std::int32_t>>
+mesh::build_local_dual_graph(std::vector<CellType> celltype,
+                             std::vector<std::span<const std::int64_t>> cells)
+{
+  spdlog::info("Build local part of mesh dual graph (mixed)");
+  common::Timer timer("Compute local part of mesh dual graph (mixed)");
+
+  if (cells.empty())
+  {
+    // Empty mesh on this process
+    return {graph::AdjacencyList<std::int32_t>(0), std::vector<std::int64_t>(),
+            0, std::vector<std::int32_t>()};
+  }
+
+  assert(cells.size() == celltype.size());
+
+  // Create indexing offset for each cell type
+  // and determine max number of vertices per facet
+  std::vector<std::int32_t> cell_offsets = {0};
+  int max_vertices_per_facet = 0;
+  int facet_count = 0;
+  int tdim = mesh::cell_dim(celltype[0]);
+  for (std::size_t j = 0; j < cells.size(); ++j)
+  {
+    assert(tdim == mesh::cell_dim(celltype[j]));
+    int num_cell_vertices = mesh::cell_num_entities(celltype[j], 0);
+    std::int32_t num_cells = cells[j].size() / num_cell_vertices;
+    cell_offsets.push_back(cell_offsets.back() + num_cells);
+    facet_count += mesh::cell_num_entities(celltype[j], tdim - 1) * num_cells;
+
+    graph::AdjacencyList<int> cell_facets
+        = mesh::get_entity_vertices(celltype[j], tdim - 1);
+
+    // Determine maximum number of vertices for facet
+    for (int i = 0; i < cell_facets.num_nodes(); ++i)
+    {
+      max_vertices_per_facet
+          = std::max(max_vertices_per_facet, cell_facets.num_links(i));
+    }
+  }
+
+  const int shape1 = max_vertices_per_facet + 1;
+  std::vector<std::int64_t> facets;
+  facets.reserve(facet_count * shape1);
+
+  for (std::size_t j = 0; j < cells.size(); ++j)
+  {
+    // Build a list of facets, defined by sorted vertices, with the connected
+    // cell index after the vertices
+    int num_cell_vertices = mesh::cell_num_entities(celltype[j], 0);
+    std::int32_t num_cells = cells[j].size() / num_cell_vertices;
+    graph::AdjacencyList<int> cell_facets
+        = mesh::get_entity_vertices(celltype[j], tdim - 1);
+
+    for (std::int32_t c = 0; c < num_cells; ++c)
+    {
+      // Loop over cell facets
+      auto v = cells[j].subspan(num_cell_vertices * c, num_cell_vertices);
+      for (int f = 0; f < cell_facets.num_nodes(); ++f)
+      {
+        auto facet_vertices = cell_facets.links(f);
+        std::transform(facet_vertices.begin(), facet_vertices.end(),
+                       std::back_inserter(facets),
+                       [v](auto idx) { return v[idx]; });
+        std::sort(std::prev(facets.end(), facet_vertices.size()), facets.end());
+        facets.insert(facets.end(),
+                      max_vertices_per_facet - facet_vertices.size(), -1);
+        facets.push_back(c + cell_offsets[j]);
+      }
+    }
+  }
+
+  // Sort facets by vertex key
+  std::vector<std::size_t> perm(facets.size() / shape1, 0);
+  std::iota(perm.begin(), perm.end(), 0);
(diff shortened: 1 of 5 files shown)


cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -364,42 +363,61 @@ mesh::build_local_dual_graph(CellType celltype,
...
+  assert(cells.size() == celltypes.size());
 
-  // Determine maximum number of vertices for facet
+  // Create indexing offset for each cell type
+  // and determine max number of vertices per facet
+  std::vector<std::int32_t> cell_offsets = {0};
   int max_vertices_per_facet = 0;
-  for (int i = 0; i < cell_facets.num_nodes(); ++i)
+  int facet_count = 0;
+  int tdim = mesh::cell_dim(celltypes[0]);
```
jorgensd, line 373:
```suggestion
  int tdim = mesh::cell_dim(celltypes.front());
```

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -486,16 +504,15 @@ mesh::build_local_dual_graph(CellType celltype,
 }
 //-----------------------------------------------------------------------------
 graph::AdjacencyList<std::int64_t>
-mesh::build_dual_graph(MPI_Comm comm, CellType celltype,
-                       const graph::AdjacencyList<std::int64_t>& cells)
+mesh::build_dual_graph(MPI_Comm comm, const std::vector<CellType>& celltypes,
```
jorgensd, line 507:
```suggestion
mesh::build_dual_graph(MPI_Comm comm, std::span<const CellType> celltypes,
```
Wouldn't a span be more natural in this context, as we try avoiding using std::vector in the code-base in general? 
I get that you can't do it for `cells` as it is closer to a list of adjacency lists.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -364,42 +364,61 @@ mesh::build_local_dual_graph(CellType celltype,
...
   }
 
-  int tdim = mesh::cell_dim(celltype);
-  int num_cell_vertices = mesh::cell_num_entities(celltype, 0);
-  const std::int32_t num_cells = cells.size() / num_cell_vertices;
-
-  // Note: only meshes with a single cell type are supported
-  const graph::AdjacencyList<int> cell_facets
-      = mesh::get_entity_vertices(celltype, tdim - 1);
+  assert(cells.size() == celltypes.size());
```
garth-wells, line 367: I think this could be an exception rather than an assert.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -364,42 +364,61 @@ mesh::build_local_dual_graph(CellType celltype,
...
 
-  // Determine maximum number of vertices for facet
+  // Create indexing offset for each cell type
+  // and determine max number of vertices per facet
+  std::vector<std::int32_t> cell_offsets = {0};
```
garth-wells, line 371: Add `cell_offsets.reserve(cells.size() + 1);`

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -364,42 +364,61 @@ mesh::build_local_dual_graph(CellType celltype,
...
   {
-    max_vertices_per_facet
-        = std::max(max_vertices_per_facet, cell_facets.num_links(i));
+    assert(tdim == mesh::cell_dim(celltypes[j]));
+    int num_cell_vertices = mesh::cell_num_entities(celltypes[j], 0);
+    std::int32_t num_cells = cells[j].size() / num_cell_vertices;
+    cell_offsets.push_back(cell_offsets.back() + num_cells);
+    facet_count += mesh::cell_num_entities(celltypes[j], tdim - 1) * num_cells;
+
+    graph::AdjacencyList<int> cell_facets
```
garth-wells, line 383: Is `int` the appropriate type? Or is it `std::Int32_t`? (same in practice, but latter 'says' more).

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -364,42 +364,61 @@ mesh::build_local_dual_graph(CellType celltype,
...
+        = mesh::get_entity_vertices(celltypes[j], tdim - 1);
+
+    // Determine maximum number of vertices for facet
+    for (int i = 0; i < cell_facets.num_nodes(); ++i)
```
garth-wells, line 387: Use a fixed-width type (other places too).

cpp/dolfinx/mesh/graphbuild.h
```diff
@@ -23,26 +23,29 @@ namespace dolfinx::mesh
...
 /// 4. The attached cell (local index) to each returned facet in (2).
 ///
 /// Each row of the returned data (2) contains `[v0, ... v_(n-1), x, ..,
 /// x]`, where `v_i` is a vertex global index, `x` is a padding value
 /// (all padding values will be equal).
 ///
-/// @note The return data will likely change once we support mixed
-/// topology meshes.
+/// @note The cell indexing stacks the cells from each cell type consecutively,
+/// i.e. if there are n cells of type 0 and m cells of type 1, then cells are
```
garth-wells, line 43:
```suggestion
/// i.e. if there are `n` cells of type `0` and `m` cells of type `1`, then cells are
```

cpp/dolfinx/mesh/graphbuild.h
```diff
@@ -23,26 +23,29 @@ namespace dolfinx::mesh
...
+/// numbered 0..(n-1) and n..(n+m-1) respectively in the dual graph.
```
garth-wells, line 44:
```suggestion
/// numbered `0..(n-1)` and `n..(n+m-1)`, respectively, in the dual graph.
```

cpp/dolfinx/mesh/graphbuild.h
```diff
@@ -23,26 +23,29 @@ namespace dolfinx::mesh
...
 /// 3. The number of columns for the facet data array (2).
 /// 4. The attached cell (local index) to each returned facet in (2).
 ///
 /// Each row of the returned data (2) contains `[v0, ... v_(n-1), x, ..,
 /// x]`, where `v_i` is a vertex global index, `x` is a padding value
 /// (all padding values will be equal).
 ///
-/// @note The return data will likely change once we support mixed
-/// topology meshes.
+/// @note The cell indexing stacks the cells from each cell type consecutively,
```
garth-wells, line 42: I don't follow the comment - how does it map to the return arguments? A short example would help.

cpp/dolfinx/mesh/graphbuild.h
```diff
@@ -52,12 +55,12 @@ build_local_dual_graph(CellType celltype, std::span<const std::int64_t> cells);
 /// @note Collective function
 ///
 /// @param[in] comm The MPI communicator
-/// @param[in] celltype The cell type
-/// @param[in] cells Collection of cells, defined by the cell vertices
-/// from which to build the dual graph
+/// @param[in] celltypes List of cell types
```
garth-wells, line 58: Precondition is `celltypes.size() == cells.size()`?

Docstring doesn't explain relationship between `celltypes` and cells`.

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -457,10 +457,28 @@ void mesh(nb::module_& m)
...
+                                               {cells.array()});
+      },
       nb::arg("comm"), nb::arg("cell_type"), nb::arg("cells"),
       "Build dual graph for cells");
 
+  m.def(
+      "build_dual_graph",
+      [](const MPICommWrapper comm,
+         std::vector<dolfinx::mesh::CellType>& cell_types,
+         const std::vector<std::vector<std::int64_t>>& cells)
```
garth-wells, line 472: This is a nasty argument type for Python/nanobind. Probably should be a `std::vector` of  `nb::ndarray` objects.

garth-wells (dismissed):

cpp/dolfinx/mesh/graphbuild.cpp
chrisrichardson: Not really important, since `cells.size()` is tiny.

python/dolfinx/wrappers/mesh.cpp
chrisrichardson: OK, something like that. I've updated it.

PR #3240, opened 2024-05-31
Distribute fixed width data

chrisrichardson:
Topology data is currently distributed as an `AdjacencyList`. This PR reduces memory usage and improves efficiency of mesh creation in parallel, by sending fixed-width data for topology (which is always the case).
Preparation for sending multiple cell types in a future PR.


cpp/dolfinx/graph/partition.h
```diff
@@ -69,6 +69,25 @@ std::tuple<graph::AdjacencyList<std::int64_t>, std::vector<int>,
...
+/// @brief Distribute fixed-width nodes to destination ranks.
+///
+/// The global index of each node is assumed to be the local index plus
+/// the offset for this rank.
+///
+/// @param[in] comm MPI Communicator
+/// @param[in] list The list to distribute, as a flattened 2D row major array
+/// @param[in] width The width of the array
+/// @param[in] destinations Destination ranks for the ith node in the
+/// adjacency list. The first rank is the 'owner' of the node.
```
chrisrichardson, line 81:
```suggestion
///  list. The first rank is the 'owner' of the node.
```

cpp/dolfinx/graph/partition.h
```diff
@@ -69,6 +69,25 @@ std::tuple<graph::AdjacencyList<std::int64_t>, std::vector<int>,
 distribute(MPI_Comm comm, const graph::AdjacencyList<std::int64_t>& list,
            const graph::AdjacencyList<std::int32_t>& destinations);
 
+/// @brief Distribute fixed-width nodes to destination ranks.
```
garth-wells, line 72: What is a 'fixed-width node`?

cpp/dolfinx/graph/partition.h
```diff
@@ -69,6 +69,25 @@ std::tuple<graph::AdjacencyList<std::int64_t>, std::vector<int>,
...
+///
+/// The global index of each node is assumed to be the local index plus
+/// the offset for this rank.
+///
+/// @param[in] comm MPI Communicator
+/// @param[in] list The list to distribute, as a flattened 2D row major array
```
garth-wells, line 78: List of what? No need for the doc string to repeat the type, needs to explain what the argument is.

cpp/dolfinx/graph/partition.h
```diff
@@ -69,6 +69,25 @@ std::tuple<graph::AdjacencyList<std::int64_t>, std::vector<int>,
...
+/// @param[in] width The width of the array
```
garth-wells, line 79: Use fewer word and more examples! Does list have shape (`list.size()/width, width)?

`width` is not a widely recognised term. Should it be `shape1`?

garth-wells (dismissed):

chrisrichardson:
Thanks for comments @garth-wells - sometimes writing the documentation is the hardest part.

PR #3241, opened 2024-05-31
cmake: find scotch with cmake config first

minrk:
- setting explicit SCOTCH_DIR still has highest priority for user override
- next, try finding scotch with cmake config (will usually work with scotch >=7, I think)
- if scotch cmake config is found _and_ defines the target SCOTCH::ptscotch, populate scotch libraries/includes from cmake targets
- otherwise, proceed as before with FindSCOTCH

Intended effect:

- no change if SCOTCH_DIR is specified
- no change if scotch cmake config is not present

WIth this, scotch 7 installed via conda is found correctly by default via cmake at least on my mac and Windows tests


garth-wells:
#3244 should achieve some thing, but is simpler.

garth-wells:
Closing in favour of #3244

PR #3244, opened 2024-06-01
Enable SCOTCH detection via CMake `CONFIG` mode.

garth-wells:
Search first for SCOTCH in `CONFIG` mode (look for a SCOTCH-provided config file), and if not found search using `MODULE` mode (using the DOLFINx provided `FindSCOTCH.cmake`).


minrk:
Makes sense! I think the only case #3241 handles that this doesn't is SCOTCH_DIR having highest priority when `find_package(SCOTCH CONFIG)` succeeds _and would produce a different result_. I'm not sure that's worth handling, but seems worth mentioning. With this, I think there is no configuration or override available to fix things if SCOTCHConfig doesn't produce the right links.

minrk:
Just to confirm, I updated the conda-forge Windows build with this PR and it works fine.

PR #3245, opened 2024-06-03
Add dependabot for github actions

jorgensd:
@finsberg taught me this today.
It is a dependabot that checks if any of our github actions are out of date.
See for instance: https://github.com/jorgensd/cbcbeatx/pull/12 for an example of this in action


jhale:
Thanks, should save on some tedious work.

PR #3246, opened 2024-06-04
Make compile time options compile time constants

schnellerhase:
Resolves [#3](https://github.com/schnellerhase/dolfinx/issues/3)


garth-wells:
Looks good - could you add `has_ptscotch`? (The compiler definition is `HAS_PTSCOTCH`.)

Seems we had a bug before with the missing definition. SCOTCH was required, but now it's optional, something must have got lost in that change.

schnellerhase:
I am unsure about what to do with the `std::string`'s. The tested compilers do not seem to have implemented compile time strings yet, so we need to treat them differently.

I tested `inline`, instead of the `consteval` which works. Besides that I tried the use of `std::string_view` which I am unsure is a legal use of such to be honest. Do you have any thoughts on that?

jhale:
We tend not to manually inline things - let the compiler use its cost model (these are almost certainly already inlined already).

schnellerhase:
We need either an `inline` or a `consteval` to allow for the now multiple definitions in the different compile units. The real in lining of the compiler I do not want to manipulate.

schnellerhase:
> Compile time strings still looks complex in C++20 - just keep those routines as they are.

In the sense of the current state, or returning them to the previous compiled setting of before?

jhale:
Can you keep constexpr in the header and keep the string ones in the .cpp?

jhale:
You should generally avoid using force pushes when working on shared projects.

Could you also add a new wrapper for DOLFINX_STDC_NO_COMPLEX_KERNELS? e.g. has_complex_kernels?

schnellerhase:
Alright, will refrain from it, just worked with projects that had a linear history before.

I can't seem to find the option `DOLFINX_STDC_NO_COMPLEX_KERNELS`, can you point me to its definition?

cpp/dolfinx/common/defines.h
```diff
@@ -23,24 +23,83 @@ std::string git_commit_hash();
...
+consteval bool has_ptscotch()
+{
+#ifdef HAS_PTSCOTCH
+  return true;
+#else
+  return false;
+#endif
+}
+
+/// Return true if DOLFINx has complex kernels enabled
```
garth-wells, line 95: @jhale can you suggest wording for this docstring? The issue is subtle, so some background would be helpful. E.g., which C compilers don't support complex numbers?

cpp/dolfinx/common/defines.h
```diff
@@ -23,24 +23,83 @@ std::string git_commit_hash();
...
+consteval bool has_complex_ufcx_kernels()
+{
+#ifdef DOLFINX_STDC_NO_COMPLEX_KERNELS
```
garth-wells, line 98: Is the condition back-to-front?

schnellerhase: Indeed :)

cpp/dolfinx/common/defines.h
jhale, line 95: Docstring suggestion

Return true if DOLFINx supports UFCx kernels with arguments of type C99 _Complex. When DOLFINx was built with MSVC this returns false.

This does not preclude using DOLFINx with kernels accepting std::complex.

jhale:
You need to merge origin/main

garth-wells (approved):

jhale:
@schnellerhase could you make sure these are allwrapped into Python in another PR? Also good practice. Thanks.

schnellerhase:
> @schnellerhase could you make sure these are allwrapped into Python in another PR? Also good practice. Thanks.

Got lost somehow, sorry - now done in a PR.

PR #3257, opened 2024-06-06
Allow specifying attribute name when reading meshtag

mleoni-pf:
If no `attribute_name` is provided, read the first attribute. Otherwise, read the attribute with the provided name or throw an error if it couldn't be found.


python/dolfinx/io/utils.py
```diff
@@ -276,10 +276,13 @@ def read_mesh(
         )
         return Mesh(msh, domain)
 
-    def read_meshtags(self, mesh, name, attribute_name="", xpath="/Xdmf/Domain"):
-        mt = super().read_meshtags(mesh._cpp_object, name, attribute_name, xpath)
+    def read_meshtags_by_name(self, mesh, name, attribute_name, xpath="/Xdmf/Domain"):
+        mt = super().read_meshtags_by_name(mesh._cpp_object, name, attribute_name, xpath)
         return MeshTags(mt)
 
+    def read_meshtags(self, mesh, name, xpath="/Xdmf/Domain"):
```
garth-wells, line 283: Needs docs.

python/dolfinx/io/utils.py
```diff
@@ -276,10 +276,13 @@ def read_mesh(
         )
         return Mesh(msh, domain)
 
-    def read_meshtags(self, mesh, name, attribute_name="", xpath="/Xdmf/Domain"):
-        mt = super().read_meshtags(mesh._cpp_object, name, attribute_name, xpath)
+    def read_meshtags_by_name(self, mesh, name, attribute_name, xpath="/Xdmf/Domain"):
```
garth-wells, line 279: Would it be better as one function and using default arg `attribute_name=None`?

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -321,9 +321,10 @@ template void XDMFFile::write_meshtags(const mesh::MeshTags<std::int32_t>&,
                                        std::string, std::string);
 /// @endcond
 //-----------------------------------------------------------------------------
-mesh::MeshTags<std::int32_t> XDMFFile::read_meshtags_by_name(
-    const mesh::Mesh<double>& mesh, const std::string name,
-    const std::string attribute_name, const std::string xpath)
+mesh::MeshTags<std::int32_t>
+XDMFFile::read_meshtags_by_name(const mesh::Mesh<double>& mesh,
```
garth-wells, line 325: Let's change `name` -> `label` in the function name. 'name' is ambiguous.

garth-wells (changes requested):

python/dolfinx/io/utils.py
mleoni-pf: I remember that you don't like having two default arguments in one function but I'm guessing that's only for C++, right?

garth-wells: Not good for C++, fine with Python if it's simple (which this is) and doesn't introduce complicated conditional handling.

python/dolfinx/io/utils.py
mleoni-pf: Done

python/dolfinx/io/utils.py
mleoni-pf: Done

cpp/dolfinx/io/XDMFFile.h
```diff
@@ -164,6 +164,16 @@ class XDMFFile
                       const mesh::Geometry<T>& x, std::string geometry_xpath,
                       std::string xpath = "/Xdmf/Domain");
 
+  /// Read MeshTags by name
+  /// @param[in] mesh The Mesh that the data is defined on
+  /// @param[in] name
```
garth-wells, line 169: What is `name`? Can you add a doctstring?

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -336,8 +337,29 @@ XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
...
+      pugi::xml_attribute hint;
+      pugi::xml_attribute name = attribute_node.attribute("Name", hint);
+      if (std::string(name.value()) == attribute_label)
+      {
+        values_data_node = attribute_node.child("DataItem");
+        found = true;
+        break;
+      }
+    }
+    if (not found)
```
garth-wells, line 357:
```suggestion
    if (!found)
```

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -336,8 +337,29 @@ XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
 
   const auto [entities, eshape] = read_topology_data(name, xpath);
 
-  pugi::xml_node values_data_node
-      = grid_node.child("Attribute").child("DataItem");
+  pugi::xml_node attribute_node = grid_node.child("Attribute");
+  pugi::xml_node values_data_node = attribute_node.child("DataItem");
+  if (not attribute_label.empty())
```
garth-wells, line 342:
```suggestion
  if (!attribute_label.empty())
```
C++ not Python!

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -336,8 +337,29 @@ XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
...
+  {
+    bool found = false;
+    for (; attribute_node;
```
garth-wells, line 345: This looks weird and is hard to read. Should it be `while` loop?

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -336,8 +337,29 @@ XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
...
+         attribute_node = attribute_node.next_sibling("Attribute"))
+    {
+      pugi::xml_attribute hint;
+      pugi::xml_attribute name = attribute_node.attribute("Name", hint);
+      if (std::string(name.value()) == attribute_label)
```
garth-wells, line 350: Creating the string is a bit clumsy, maybe use https://en.cppreference.com/w/cpp/string/char_traits/compare.

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -336,8 +337,29 @@ XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
...
-      = grid_node.child("Attribute").child("DataItem");
+  pugi::xml_node attribute_node = grid_node.child("Attribute");
+  pugi::xml_node values_data_node = attribute_node.child("DataItem");
+  if (not attribute_label.empty())
+  {
+    bool found = false;
+    for (; attribute_node;
+         attribute_node = attribute_node.next_sibling("Attribute"))
+    {
+      pugi::xml_attribute hint;
```
garth-wells, line 348: What is this for?

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -376,6 +398,13 @@ XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
   return meshtags;
 }
 //-----------------------------------------------------------------------------
+mesh::MeshTags<std::int32_t>
+XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
+                        std::string xpath)
+{
+  return read_meshtags_by_label(mesh, name, "", xpath);
```
garth-wells, line 405:
```suggestion
  return read_meshtags_by_label(mesh, name, std:::string(), xpath);
```

python/dolfinx/wrappers/io.cpp
```diff
@@ -308,15 +310,18 @@ void io(nb::module_& m)
            nb::arg("name") = "mesh", nb::arg("xpath") = "/Xdmf/Domain")
       .def("read_cell_type", &dolfinx::io::XDMFFile::read_cell_type,
            nb::arg("name") = "mesh", nb::arg("xpath") = "/Xdmf/Domain")
-      .def("read_meshtags", &dolfinx::io::XDMFFile::read_meshtags,
-           nb::arg("mesh"), nb::arg("name"), nb::arg("xpath") = "/Xdmf/Domain")
+      .def("read_meshtags", &dolfinx::io::XDMFFile::read_meshtags_by_label,
+           nb::arg("mesh"), nb::arg("name"), nb::arg("attribute_label") = "",
```
garth-wells, line 314:
```suggestion
           nb::arg("mesh"), nb::arg("name"), nb::arg("attribute_label"),
```

python/dolfinx/wrappers/io.cpp
```diff
@@ -308,15 +310,18 @@ void io(nb::module_& m)
...
+           nb::arg("xpath") = "/Xdmf/Domain")
```
garth-wells, line 315:
```suggestion
           nb::arg("xpath"))
```

garth-wells (changes requested):
The code is `XDMFFile.cpp` looks to me more complicated than it needs to be. I've left some suggestions as a first pass. I can look again once it's been simplified some more.

cpp/dolfinx/io/XDMFFile.h
mleoni-pf: Done, but for the record this was already undocumented.

cpp/dolfinx/io/XDMFFile.cpp
mleoni-pf: Done

cpp/dolfinx/io/XDMFFile.cpp
mleoni-pf: Done. Although to be fair that's valid C++ :stuck_out_tongue:

cpp/dolfinx/io/XDMFFile.cpp
mleoni-pf: Done

cpp/dolfinx/io/XDMFFile.cpp
mleoni-pf: Using the function you suggested requires providing a number of characters to compare, so it would have to be something like

```c++
if (auto len = std::max(std::strlen(name.value()), attribute_label.size());
    std::char_traits<char>::compare(name.value(), attribute_label.c_str(),
                                    len) == 0)
```
right?

Doesn't look less clumsy to me.

cpp/dolfinx/io/XDMFFile.cpp
mleoni-pf: It's something that `pugixml` requires, like a hint to try speed-up the search of a node.

python/dolfinx/wrappers/io.cpp
mleoni-pf: Done

python/dolfinx/wrappers/io.cpp
mleoni-pf: Done

cpp/test/mesh/Domain.xdmf
```diff
@@ -0,0 +1,26 @@
+<Xdmf Version="3.0">
```
garth-wells, line 1: Can we avoid adding data files? Can it be created by the test rather than added to the repo?

mleoni-pf: I can convert the file to ascii, add it to the source code and have the code write it to `/dev/shm`, then read it back. Would that be better?

mleoni-pf: On second though, that's really just the same thing as adding a file with the ascii content, which is what I do now XD
Please advise!

mleoni-pf: Ok, never mind... I turned on my brain and found a solution.

cpp/dolfinx/io/XDMFFile.h
```diff
@@ -164,9 +164,21 @@ class XDMFFile
                       const mesh::Geometry<T>& x, std::string geometry_xpath,
                       std::string xpath = "/Xdmf/Domain");
 
+  /// Read MeshTags by name
```
garth-wells, line 167: This is a bit confusing - the docstring says "Read MeshTags by name" but the function name is "read_meshtags_by_label". Is it by name or by label?

cpp/dolfinx/io/XDMFFile.cpp
mleoni-pf: Done!

cpp/dolfinx/io/XDMFFile.cpp
mleoni-pf: I don't think it can because `hint` is also an output parameter of the function.

cpp/dolfinx/io/XDMFFile.h
mleoni-pf: The issue comes from the following fact: mesh tags are identified by "a word", so it is natural that this word be called the "name" of the mesh tag. The problem is that there is already another `name` parameter which refers to the name of the mesh node in the same file. I picked the word "label" to distinguish the two internally, in the code implementation, but a user should not be concerned with this implementation detail: a user saved a mesh tag giving it a "name" and wants to read it back by the same "name".
The fact remains that the user still needs to call a function whose name includes the word "label", not "name", which is indeed a bit confusing. I will change everything to be called more consistently, the code shouldn't become too unintelligible on the development side.

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -336,8 +337,28 @@ XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
...
   const auto [entities, eshape] = read_topology_data(name, xpath);
 
-  pugi::xml_node values_data_node
-      = grid_node.child("Attribute").child("DataItem");
+  pugi::xml_node attribute_node = grid_node.child("Attribute");
+  pugi::xml_node values_data_node = attribute_node.child("DataItem");
+  if (!attribute_label.empty())
+  {
+    bool found = false;
+    while (!found and attribute_node)
```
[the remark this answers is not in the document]
mleoni-pf: I remember looking at this. The problem was, if I recall correctly, that I am searching for an attribute by the name of `"Name"` and the "basic functionality" functions do not allow to search for a name. If I wanted to loop using `next_attribute` I would then need to check if the attribute's name is `"Name"` and, if not, `continue`.

I will add documentation but the code doesn't look to exotic to me [because I wrote it, of course XD].

cpp/dolfinx/io/XDMFFile.h
mleoni-pf: By the way, re-reading this conversation I noticed that in the first review, four months ago, you asked me to change the method name to avoid the word "name" in favour of "label". I had forgotten about that and, based on your comment, I reverted the change. Which one should I keep? To a user, a mesh tag's "label" means presumably very little...

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -334,8 +334,9 @@ template void XDMFFile::write_meshtags(const mesh::MeshTags<std::int32_t>&,
 /// @endcond
 //-----------------------------------------------------------------------------
 mesh::MeshTags<std::int32_t>
-XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
-                        std::string xpath)
+XDMFFile::read_meshtags_by_name(const mesh::Mesh<double>& mesh,
```
jorgensd, line 337:
```suggestion
XDMFFile::read_meshtags_by_label(const mesh::Mesh<double>& mesh,
```

cpp/dolfinx/io/XDMFFile.h
```diff
@@ -164,9 +164,21 @@ class XDMFFile
...
+  /// @param[in] mesh The Mesh that the data is defined on
```
jorgensd, line 168:
```suggestion
  /// @param[in] mesh The Mesh that the input data is defined on
```

cpp/dolfinx/io/XDMFFile.h
```diff
@@ -164,9 +164,21 @@ class XDMFFile
...
+  /// @param[in] name Name of the grid node in the xml file. E.g. "Material" in
```
jorgensd, line 169:
```suggestion
  /// @param[in] name Name of the grid node in the xml-scheme of the XDMF-file. E.g. "Material" in
```

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -349,38 +349,22 @@ XDMFFile::read_meshtags_by_name(const mesh::Mesh<double>& mesh,
...
-        values_data_node = attribute_node.child("DataItem");
-        found = true;
-      }
-      attribute_node = attribute_node.next_sibling("Attribute");
-    }
-
-    // If the search ended after testing all attributes but without finding
-    // a match, throw an error.
-    if (!found)
+    if (!attribute_node)
```
jorgensd, line 361: So much cleaner @mleoni-pf 👯

jorgensd:
@garth-wells 
@mleoni-pf and I sat down and looked through the pugixml docs today.
The new solution is much cleaner and much closer to what we use in general in DOLFINx (i.e. lambda function for search predicate).
This is a very neat function to have, as it will allow storing multiple data arrays within the same grid (for instance material parameters.

cpp/dolfinx/io/XDMFFile.cpp
jorgensd, line 337: Do we really want a separate function for this in C++? Couldn't we just put the attribute name as a fourth argument (optional) argument.

I don't see the added value of a constructor for handling one optional value to have a backwards compatible C++ interface. @garth-wells

garth-wells: Agree, one function would be better. Would is make sense to use `std::optional` for `attribute_name`?

Related to this, we should not pass empty strings as 'null' - use `std::optional`.

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -350,6 +351,21 @@ XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
 
   pugi::xml_node values_data_node
       = grid_node.child("Attribute").child("DataItem");
+  if (!attribute_name.empty())
```
garth-wells, line 354: Use `std::optional`, in which case this would become `if (attribute_name)`.

cpp/dolfinx/io/XDMFFile.cpp
garth-wells: Looks good. Could probably be simplified by removing the occurrences of `&` and `const`. Make the`pugi::xml_node` const if you want `auto` to inherit const.

cpp/test/mesh/read_named_meshtags.cpp
```diff
@@ -0,0 +1,96 @@
...
+  auto mesh = std::make_shared<mesh::Mesh<double>>(
+      mesh::create_rectangle(MPI_COMM_WORLD, {{{0.0, 0.0}, {1.0, 1.0}}}, {3, 3},
+                             mesh::CellType::triangle, part));
+
+  const std::int32_t n_cells = mesh->topology()->index_map(2)->size_local();
+  std::vector<std::int32_t> indices(n_cells);
+  std::iota(std::begin(indices), std::end(indices), 0);
+
+  std::vector<std::int32_t> domain_values(n_cells);
+  std::ranges::fill(domain_values, domain_value);
```
garth-wells, line 45: Just use `std::vector<std::int32_t> domain_values(n_cells, domain_value);`

cpp/test/mesh/read_named_meshtags.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+  std::vector<std::int32_t> material_values(n_cells);
```
garth-wells, line 47: See above comment.

python/dolfinx/io/utils.py
```diff
@@ -276,8 +276,15 @@ def read_mesh(
         )
         return Mesh(msh, domain)
 
-    def read_meshtags(self, mesh, name, xpath="/Xdmf/Domain"):
-        mt = super().read_meshtags(mesh._cpp_object, name, xpath)
+    def read_meshtags(self, mesh, name, attribute_label=None, xpath="/Xdmf/Domain"):
```
garth-wells, line 279: Follow the Google docs style with `Arguments`, etc. There are quite a few examples in the code.

garth-wells (changes requested):

garth-wells:
> @garth-wells @mleoni-pf and I sat down and looked through the pugixml docs today. The new solution is much cleaner and much closer to what we use in general in DOLFINx (i.e. lambda function for search predicate). This is a very neat function to have, as it will allow storing multiple data arrays within the same grid (for instance material parameters.

Nice - looking good.

cpp/dolfinx/io/XDMFFile.cpp
mleoni-pf: Done.

python/dolfinx/io/utils.py
```diff
@@ -205,14 +205,21 @@ def read_mesh(
         )
         return Mesh(msh, domain)
 
-    def read_meshtags(self, mesh, name, attribute_label="", xpath="/Xdmf/Domain"):
-        """Read meshtags with a specific name as specified in the xdmf file. If
-        `attribute_label` is empty, read the first attribute in the file.
-        If `attribute_label` is not empty but no attributes have the provided name,
-        throw an error. If multiple attributes have the provided name, read the first
-        one found."""
+    def read_meshtags(self, mesh, name, attribute_name="", xpath="/Xdmf/Domain"):
```
jorgensd, line 208:
```suggestion
    def read_meshtags(self, mesh: dolfinx.mesh.Mesh, name: str, attribute_name: str="", xpath: str="/Xdmf/Domain"):
```

jorgensd (approved):
LGTM!

garth-wells (changes requested):
I've pushed some fixes to the code. Last things that is missing is a Python test. @massimiliano-leoni can you add a Python test?

mleoni-pf:
@garth-wells Python test added!

garth-wells (approved):

PR #3259, opened 2024-06-07
Add ARM VTK to Docker Notebook image.

jhale:


jhale:
This is available at `docker.io/dolfinx/lab:test-vtk-arm64-arm64` and works OK for me.

PR #3261, opened 2024-06-07
Prepare for release of numpy 2.0 by using ruff ruleset

jorgensd:
Ref:  https://numpy.org/devdocs/numpy_2_0_migration_guide.html#ruff-plugin

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -8,7 +8,7 @@
 requires = [
       "scikit-build-core[pyproject]>=0.5",
       "nanobind>=2.0.0",
-      "mpi4py"
+      "mpi4py",
 ]
 build-backend = "scikit_build_core.build"
 
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -8,7 +8,7 @@
 requires = [
       "scikit-build-core[pyproject]>=0.5",
       "nanobind>=2.0.0",
-      "mpi4py",
+      "mpi4py"
 ]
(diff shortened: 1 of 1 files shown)


python/pyproject.toml
```diff
@@ -8,7 +8,7 @@
 requires = [
       "scikit-build-core[pyproject]>=0.5",
       "nanobind>=2.0.0",
-      "mpi4py",
+      "mpi4py"
```
jorgensd, line 11:
```suggestion
      "mpi4py",
```

python/pyproject.toml
```diff
@@ -34,7 +34,7 @@ dependencies = [
 
 [project.optional-dependencies]
 docs = ["markdown", "pyyaml", "sphinx", "sphinx_rtd_theme"]
-lint = ["ruff"]
+lint = ["ruff>=0.2.0"]
```
garth-wells, line 37: Is this pin required for ruff to run?

jorgensd: According to the release notes I've linked above, yes:
> Ruff plugin
>
> Many of the changes covered in the 2.0 release notes and in this migration guide can be automatically adapted > in downstream code with a dedicated [Ruff](https://docs.astral.sh/ruff/) rule, namely rule [NPY201](https://docs.astral.sh/ruff/rules/numpy2-deprecation/).
> You should install ruff>=0.2.0 and add the NPY201 rule to your pyproject.toml:

garth-wells (approved):

PR #3271, opened 2024-06-15
Create mixed topology mesh

chrisrichardson:
Add a new experimental `create_mesh()` interface, which accepts mixed topology input, calls the partitioner and creates a mesh, in the same way as for single cell type.


cpp/dolfinx/mesh/utils.h
```diff
@@ -888,6 +888,234 @@ Mesh<typename std::remove_reference_t<typename U::value_type>> create_mesh(
...
+/// @param[in] commt Communicator that the topology data (`cells`) is
+/// distributed on. This should be `MPI_COMM_NULL` for ranks that should
+/// not participate in computing the topology partitioning.
+/// @param[in] cells Cells on the calling process. Each cell is defined by its
+/// 'nodes' (using global indices) following the Basix ordering. For lowest
+/// order cells this will be just the cell vertices. For higher-order cells,
+/// other cells 'nodes' will be included. See dolfinx::io::cells for examples of
+/// the Basix ordering.
+/// @param[in] elements Coordinate elements for the cells.
+/// @param[in] commg
```
jorgensd, line 913: Add documentation for `commg`.

cpp/dolfinx/mesh/utils.h
```diff
@@ -888,6 +888,234 @@ Mesh<typename std::remove_reference_t<typename U::value_type>> create_mesh(
...
+/// the process offset  on`comm`, The offset  is the sum of `x` rows on
+/// all processed with a lower rank than the caller.
+/// @param[in] xshape Shape of the `x` data.
+/// @param[in] partitioner Graph partitioner that computes the owning
+/// rank for each cell. If not callable, cells are not redistributed.
+/// @return A mesh distributed on the communicator `comm`.
+template <typename U>
+Mesh<typename std::remove_reference_t<typename U::value_type>> create_mesh(
+    MPI_Comm comm, MPI_Comm commt,
+    std::vector<std::span<const std::int64_t>>& cells,
```
garth-wells, line 925: This should be pass by copy, or a const reference.

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -279,6 +279,50 @@ void declare_mesh(nb::module_& m, std::string type)
       },
       nb::arg("comm"), nb::arg("p"), nb::arg("n"), nb::arg("celltype"),
       nb::arg("partitioner").none());
+
+  m.def("create_mesh",
+        [](MPICommWrapper comm,
+           const std::vector<nb::ndarray<const std::int64_t, nb::ndim<1>,
+                                         nb::c_contig>>& cells_nb,
+           std::vector<dolfinx::fem::CoordinateElement<T>>& elements,
```
garth-wells, line 287: `const` ref?

cpp/dolfinx/mesh/utils.h
```diff
@@ -888,6 +888,234 @@ Mesh<typename std::remove_reference_t<typename U::value_type>> create_mesh(
...
+/// re-distribution of cells is done.
+///
+/// @note This is an experimental specialised version of `create_mesh` for mixed
+/// topology meshes, and does not include cell reordering.
+///
+/// @param[in] comm Communicator to build the mesh on.
+/// @param[in] commt Communicator that the topology data (`cells`) is
+/// distributed on. This should be `MPI_COMM_NULL` for ranks that should
+/// not participate in computing the topology partitioning.
+/// @param[in] cells Cells on the calling process. Each cell is defined by its
```
garth-wells, line 907: Needs improved docs - the type is now 'array of arrays'. Docs need to explain what's what.

cpp/dolfinx/mesh/utils.h
```diff
@@ -747,7 +747,7 @@ compute_incident_entities(const Topology& topology,
 /// 'nodes' will be included. See dolfinx::io::cells for examples of the
 /// Basix ordering.
 /// @param[in] element Coordinate element for the cells.
-/// @param[in] commg
+/// @param[in] commg Communicator for geometry
```
garth-wells, line 750: This docstring says very little. We need to aim to be really clear. It is the communicator that the coordinate data is distributed across. Is it `MPI_COMM_NULL` for processes without geometry data?

cpp/dolfinx/mesh/utils.h
```diff
@@ -888,6 +888,235 @@ Mesh<typename std::remove_reference_t<typename U::value_type>> create_mesh(
...
+/// @param[in] commt Communicator that the topology data (`cells`) is
+/// distributed on. This should be `MPI_COMM_NULL` for ranks that should
+/// not participate in computing the topology partitioning.
+/// @param[in] cells Cells on the calling process, as a list of lists,
+/// one list for each cell type. The cells are defined by their
+/// 'nodes' (using global indices) following the Basix ordering, and
+/// concatenated to form a flattened list. For lowest order cells this will be
+/// just the cell vertices. For higher-order cells, other cells 'nodes' will be
+/// included. See dolfinx::io::cells for examples of the Basix ordering.
+/// @param[in] elements Coordinate elements for the cells.
```
jorgensd, line 913: Should this list be consistent across all processes? I.e. Even if a process only has triangular elements, and another quadrilaterals, both should send in (triangle, quad) coordinate element to ensure that everything is consistent at the point of code generation?

this would mean that some of the lists in cells could be empty (But that shouldn’t be a problem?)

chrisrichardson: I can add a comment about that.

cpp/dolfinx/mesh/utils.h
chrisrichardson: Not sure, I didn't add it.

jorgensd:
I can't get this branch to work with mixing P1 and P2 triangles in the same mesh (even if they are completely disconnected.

Here is an image of my minimal mesh:
image
If i change the order of the second triangle to 1, the code runs smoothly.
I do not have minimal working example. But I've attached the mesh I've used here, and the code I've used for reading the mesh is available at:
https://gist.github.com/jorgensd/fcbcee61f773e8889eccacf2f75158a2
(needs `mesh_converter` installed: https://github.com/jorgensd/mesh_converter/
[test.vtkhdf.zip](https://github.com/user-attachments/files/15898734/test.vtkhdf.zip)

jorgensd:
Error message is:
```bash
python3: /root/shared/dolfinx/cpp/dolfinx/fem/dofmapbuilder.cpp:299: std::tuple<std::vector<{anonymous}::dofmap_t, std::allocator<{anonymous}::dofmap_t> >, std::vector<long int, std::allocator<long int> >, std::vector<std::pair<signed char, int>, std::allocator<std::pair<signed char, int> > >, std::vector<std::shared_ptr<const dolfinx::common::IndexMap>, std::allocator<std::shared_ptr<const dolfinx::common::IndexMap> > >, long int> {anonymous}::build_basic_dofmaps(const dolfinx::mesh::Topology&, const std::vector<dolfinx::fem::ElementDofLayout>&): Assertion `(int)num_entity_dofs == num_entity_dofs_et[k]' failed.
Aborted (core dumped)
```

jorgensd:
Another thing I realized when looking through this is that we need to think about how to wrap this as a Python object, as the C++ object is not really usable in Python.

chrisrichardson:
It will not currently work with mixed order meshes - the geometry dofmap would have hanging nodes... and I guess this causes a problem, even if the meshes are disconnected. I am sure this can be fixed in a future PR (e.g. by using DG geometry).

chrisrichardson:
> Another thing I realized when looking through this is that we need to think about how to wrap this as a Python object, as the C++ object is not really usable in Python.

It's too experimental at the moment. Accessing through `dolfinx.cpp` should be fine for developers. We still have other issues to resolve: e.g. FunctionSpace and assembly... https://github.com/orgs/FEniCS/projects/7

jorgensd:
> It will not currently work with mixed order meshes - the geometry dofmap would have hanging nodes... and I guess this causes a problem, even if the meshes are disconnected. I am sure this can be fixed in a future PR (e.g. by using DG geometry).

I don't understand why the dofmap would have hanging nodes? Shouldn't there be a dofmap per coordinate element type? Hanging nodes would only happen if a higher order cell shared two vertices with a lower order (linear) cell?

chrisrichardson:
The dofmap construction needs the same number of dofs en each entity type (edges, facets etc.) or it will currently fail, even if the mesh is disjoint. Whilst each element has its own dofmap, they have to match if they are sharing degrees of freedom between cells. I think we can fix it - either by using DG geometry, or by updating the dofmap constructor. It's a future challenge.

jorgensd:
> The dofmap construction needs the same number of dofs en each entity type (edges, facets etc.) or it will currently fail, even if the mesh is disjoint. Whilst each element has its own dofmap, they have to match if they are sharing degrees of freedom between cells. I think we can fix it - either by using DG geometry, or by updating the dofmap constructor. It's a future challenge.

So that means that we are limited to linear geometries for now?
As if we did some second order tetrahedra, prism and hexahedra the sub entities would have different dofs on them (i.e. a facet of a tetrahedra would have less dofs than a facet of a hexahedra?

Now it only works since the only place with dofs associated with them are the vertices, and each vertex has one dof regardless of the shape?

chrisrichardson:
> So that means that we are limited to linear geometries for now?

No, just compatible ones. If everything is 3rd order, in 3D for example, the dofs will agree on edges and facets (2 dofs on each edge, 1 dof on a triangle, 4 on a quad). This would actually be a good one to test, since the permutations have not been tested yet for mixed topology.

PR #3281, opened 2024-06-25
Improve check for facet->cell connectivity

chrisrichardson:
Currently `exterior_facet_indices()` only checks that the facet index map exists, which is insufficient.

* add a check for facet->cell connectivity instead

### DIFF
--- a/cpp/dolfinx/mesh/utils.cpp
+++ b/cpp/dolfinx/mesh/utils.cpp
@@ -60,9 +60,10 @@ std::vector<std::int32_t> mesh::exterior_facet_indices(const Topology& topology)
   const int tdim = topology.dim();
   auto f_to_c = topology.connectivity(tdim - 1, tdim);
   if (!f_to_c)
+  {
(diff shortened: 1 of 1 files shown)


jpdean (approved):

cpp/dolfinx/mesh/utils.cpp
```diff
@@ -58,14 +58,14 @@ mesh::extract_topology(CellType cell_type, const fem::ElementDofLayout& layout,
 std::vector<std::int32_t> mesh::exterior_facet_indices(const Topology& topology)
 {
   const int tdim = topology.dim();
-  auto facet_map = topology.index_map(tdim - 1);
-  if (!facet_map)
-    throw std::runtime_error("Facets have not been computed.");
+  auto f_to_c = topology.connectivity(tdim - 1, tdim);
+  if (!f_to_c)
+    throw std::runtime_error(
```
garth-wells, line 63: Add braces for multi-line statement.

PR #3283, opened 2024-06-25
Favor `exterior_facet_indices` over `locate_entities_boundary` for retrieving complete boundary

schnellerhase:
This allows for the usage of `locate_entities_boundary` without providing a 'all true' marker to identify all boundary facets. This increases readability of the use cases, which before had to manually encode this all true function, done in different ways, and refrains from evaluating any marker in these cases altogether. 

Fixes https://github.com/schnellerhase/dolfinx/issues/16

### DIFF
--- a/python/demo/demo_biharmonic.py
+++ b/python/demo/demo_biharmonic.py
@@ -126,8 +126,6 @@ else:
 from mpi4py import MPI
 
 # +
-import numpy as np
-
 import dolfinx
 import ufl
 from dolfinx import fem, io, mesh, plot
@@ -165,14 +163,7 @@ V = fem.functionspace(msh, ("Lagrange", 2))
 # function that returns `True` for points `x` on the boundary and
 # `False` otherwise.
 
-facets = mesh.locate_entities_boundary(
-    msh,
-    dim=1,
-    marker=lambda x: np.isclose(x[0], 0.0)
-    | np.isclose(x[0], 1.0)
-    | np.isclose(x[1], 0.0)
-    | np.isclose(x[1], 1.0),
-)
+facets = mesh.locate_entities_boundary(msh, dim=1)
 
 # We now find the degrees-of-freedom that are associated with the
 # boundary facets using {py:func}`locate_dofs_topological
--- a/python/demo/demo_biharmonic.py
+++ b/python/demo/demo_biharmonic.py
@@ -163,7 +163,8 @@ V = fem.functionspace(msh, ("Lagrange", 2))
 # function that returns `True` for points `x` on the boundary and
 # `False` otherwise.
 
-facets = mesh.exterior_facet_indices(msh)
+msh.topology.create_connectivity(1, 2)
+facets = mesh.exterior_facet_indices(msh.topology)
 
 # We now find the degrees-of-freedom that are associated with the
 # boundary facets using {py:func}`locate_dofs_topological
--- a/python/demo/demo_biharmonic.py
+++ b/python/demo/demo_biharmonic.py
@@ -163,7 +163,7 @@ V = fem.functionspace(msh, ("Lagrange", 2))
 # function that returns `True` for points `x` on the boundary and
 # `False` otherwise.
 
-facets = mesh.locate_entities_boundary(msh, dim=1)
+facets = mesh.exterior_facet_indices(msh)
 
 # We now find the degrees-of-freedom that are associated with the
 # boundary facets using {py:func}`locate_dofs_topological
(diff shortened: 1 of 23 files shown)


jorgensd:
We already have `dolfinx.mesh.exterior_facet_indices` that covers this case (https://docs.fenicsproject.org/dolfinx/v0.8.0/cpp/mesh.html#_CPPv4N7dolfinx4mesh22exterior_facet_indicesERK8Topology)
Is there a reason for adding a specialisation for the `locate_entities_boundary` function?

schnellerhase:
Right, that is also the one used in `locate_entities_boundary` before - and thus we could return there already. 

Wasn't really aware of that functionality to be honest - but passing that through the same interface seems like a good way to go nevertheless. So the difference between a specially marked boundary and the default whole boundary is just the marker.

schnellerhase:
I see, maybe we switch to the explicit usage of `exterior_facet_indices` then in those instances of the here changed code? If they had been used in these demos, I probably would also have not thought about extending the interface in any way.

jorgensd:
> I see, maybe we switch to the explicit usage of `exterior_facet_indices` then in those instances of the here changed code? If they had been used in these demos, I probably would also have not thought about extending the interface in any way.

I would be happy with that, we should keep one demo to illustrate how to use numpy binary operators, but the rest can go I guess

schnellerhase:
On a second though the `exterior_facet_indices` only allows access to the `top.dim - 1` entities on the boundary, not entities of general dimension - which the special case handling allows for. But I assume this is rather nitpicking to rectify my changes.

jorgensd:
> On a second thought the `exterior_facet_indices` only allows access to the `top.dim - 1` entities on the boundary, not entities of general dimension - which the special case handling allows for. But I assume this is rather nitpicking to rectify my changes.

I am happy to hear what others think. 

However, one can use exterior_facet indices + compute_incident_entities to get all entities of lower dimension associated with those facets.

schnellerhase:
So the optional is of the table, I changed to the usage of `exterior_facet_indices` where possible. This makes an explicit computation of the connectivity necessary before hand.

python/demo/demo_biharmonic.py
```diff
@@ -165,14 +163,8 @@
...
 
-facets = mesh.locate_entities_boundary(
-    msh,
-    dim=1,
-    marker=lambda x: np.isclose(x[0], 0.0)
-    | np.isclose(x[0], 1.0)
-    | np.isclose(x[1], 0.0)
-    | np.isclose(x[1], 1.0),
-)
+msh.topology.create_connectivity(1, 2)
```
jorgensd, line 166:
```suggestion
msh.topology.create_connectivity(msh.topology.dim-1, msh.topology.dim)
```

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -301,14 +302,8 @@ def test_mixed_dom_codim_1(n, k):
...
     tdim = msh.topology.dim
-    boundary_facets = locate_entities_boundary(
-        msh,
-        tdim - 1,
-        lambda x: np.isclose(x[0], 0.0)
-        | np.isclose(x[0], 1.0)
-        | np.isclose(x[1], 0.0)
-        | np.isclose(x[1], 1.0),
-    )
+    msh.topology.create_connectivity(1, 2)
```
jorgensd, line 305: Same as above, use msh.topology.dim to ensure extendibility 3D

python/test/unit/fem/test_bcs.py
```diff
@@ -119,9 +119,8 @@ def test_constant_bc_constructions():
     V2 = functionspace(msh, ("Lagrange", 1, (gdim, gdim)))
 
     tdim = msh.topology.dim
-    boundary_facets = locate_entities_boundary(
-        msh, tdim - 1, lambda x: np.ones(x.shape[1], dtype=bool)
-    )
+    msh.topology.create_connectivity(1, 2)
```
jorgensd, line 122: same as above.

python/test/unit/fem/test_fem_pipeline.py
```diff
@@ -250,9 +249,8 @@ def test_petsc_curl_curl_eigenvalue(family, order):
     a = inner(ufl.curl(u), ufl.curl(v)) * dx
     b = inner(u, v) * dx
 
-    boundary_facets = locate_entities_boundary(
-        mesh, mesh.topology.dim - 1, lambda x: np.full(x.shape[1], True, dtype=bool)
-    )
+    mesh.topology.create_connectivity(1, 2)
```
jorgensd, line 252: Same

jorgensd (approved):

python/demo/demo_biharmonic.py
```diff
@@ -165,14 +163,9 @@
 # function that returns `True` for points `x` on the boundary and
```
francesco-ballarin, line 163: @schnellerhase can you prepare a follow up PR in which you also update this comment? If you look two lines above this one, the former `locate_entities_boundary` is still mentioned.

PR #3284, opened 2024-06-25
Pyamg test Add xfail for now

chrisrichardson:


jorgensd:
Ref: https://github.com/pyamg/pyamg/issues/414

PR #3291, opened 2024-07-04
Write and read mesh using ADIOS2

ampdes:
This PR introduces module for native checkpointing using ADIOS2.

For the python APIs, we cannot use the ADIOS2 python bindings which are based on pybinds, hence a wrapper to ADIOS2 is created.

The ADIOS2 BeginStep() and EndStep() collectives are designed per write function. However, for the reader, we have that the BeginStep is external to function to first query the types and then call the reader with the correct type.

This PR has mesh writer/reader.

For time dependent mesh, the topology is written once and and only geometry is written with time dependency.
Therefore for reading, we have two separate functions `read_mesh` which reads the topology and geometry and constructs a mesh and then `update_mesh` which takes this mesh and updates the associated geometry at a later time.

TODO:
- [x] Add time dependent mesh write
- [x] Add time dependent mesh read
- [x] Add unit tests for time dependent mesh


cpp/dolfinx/io/ADIOS2_utils.cpp
```diff
@@ -0,0 +1,50 @@
+// Copyright (C) 2021-2023 Jørgen S. Dokken and Garth N. Wells
```
jorgensd, line 1: @ampdes You should add your own name here and update copyright year:)

cpp/dolfinx/io/ADIOS2_utils.h
```diff
@@ -0,0 +1,69 @@
+// Copyright (C) 2021-2023 Jørgen S. Dokken and Garth N. Wells
```
jorgensd, line 1: Same here:)

ampdes:
A preliminary review of the structure/naming in the slack channel will be appropriate first. 
I will update this then.

ampdes:
@jhale @jorgensd
Will you kindly review this?

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,151 @@
...
+{
+std::map<basix::element::lagrange_variant, std::string> lagrange_variants{
+    {basix::element::lagrange_variant::unset, "unset"},
+    {basix::element::lagrange_variant::equispaced, "equispaced"},
+    {basix::element::lagrange_variant::gll_warped, "gll_warped"},
+};
+
+template <std::floating_point T>
+void _write(ADIOS2Engine& adios2engine,
+            std::shared_ptr<dolfinx::mesh::Mesh<T>> mesh)
```
jhale, line 31: This implies that this function might share ownership of the mesh.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,151 @@
...
+{
+
+  auto io = adios2engine.io();
+  auto writer = adios2engine.engine();
+
+  const dolfinx::mesh::Geometry<T>& geometry = mesh->geometry();
+  auto topology = mesh->topology();
```
jhale, line 38: Over use of auto.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+  std::int16_t mesh_dim = geometry.dim();
```
jhale, line 40: Is this performing a cast? Check definition.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,151 @@
...
+  std::uint32_t num_dofs_per_cell
+      = geom_layout.num_entity_closure_dofs(mesh_dim);
+
+  const std::vector<int64_t> input_global_indices
+      = geometry.input_global_indices();
+  const std::span<const int64_t> input_global_indices_span(
+      input_global_indices.begin(), num_nodes_local);
+  const std::span<const T> mesh_x = geometry.x();
+
+  auto connectivity = topology->connectivity(mesh_dim, 0);
```
jhale, line 64: Overuse of auto

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,151 @@
...
+  auto topology_array = connectivity->array();
```
jhale, line 65: Don't use types in variable names, clear from right hand side.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,151 @@
...
+  const std::span<const int32_t> topology_array_span(
+      topology_array.begin(), topology_offsets[num_cells_local]);
+  std::vector<std::int64_t> topology_array_global(
+      topology_offsets[num_cells_local]);
+
+  std::iota(topology_array_global.begin(), topology_array_global.end(), 0);
+
+  imap->local_to_global(topology_array_span, topology_array_global);
+
+  for (std::size_t i = 0; i < num_cells_local + 1; ++i)
```
jhale, line 77: Possible with standard library iterators?

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,151 @@
...
+void dolfinx::io::checkpointing::write(
+    ADIOS2Engine& adios2engine,
+    std::shared_ptr<dolfinx::mesh::Mesh<float>> mesh)
+{
+
+  _write(adios2engine, mesh);
+}
+
+//-----------------------------------------------------------------------------
+void dolfinx::io::checkpointing::write(
```
jhale, line 143: Why not template directly on type?

cpp/dolfinx/io/checkpointing.h
```diff
@@ -0,0 +1,39 @@
...
+#include "ADIOS2_utils.h"
+#include <adios2.h>
+#include <basix/finite-element.h>
+#include <dolfinx/mesh/Mesh.h>
+#include <mpi.h>
+
+/// @file checkpointing.h
+/// @brief ADIOS2 based checkpointing
+
+namespace dolfinx::io::checkpointing
```
jhale, line 20: Will need to discuss checkpointing terminology. Native IO could be a better description.

cpp/dolfinx/io/ADIOS2_utils.cpp
```diff
@@ -0,0 +1,36 @@
+// Copyright (C) 2024 Abdullah Mujahid, Jørgen S. Dokken and Garth N. Wells
```
jhale, line 1: Please remove Garth unless I missed something

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,151 @@
...
+#include <mpi.h>
+
+using namespace dolfinx;
+using namespace dolfinx::io;
+
+/// @file checkpointing.h
+/// @brief ADIOS2 based checkpointing
+namespace
+{
+std::map<basix::element::lagrange_variant, std::string> lagrange_variants{
```
jhale, line 23: Is there a reason not to place this in Basix?

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,151 @@
...
+    std::shared_ptr<dolfinx::mesh::Mesh<float>> mesh)
+{
+
+  _write(adios2engine, mesh);
+}
+
+//-----------------------------------------------------------------------------
+void dolfinx::io::checkpointing::write(
+    ADIOS2Engine& adios2engine,
+    std::shared_ptr<dolfinx::mesh::Mesh<double>> mesh)
```
jhale, line 145: Not sure about use of shared pointer here.

cpp/dolfinx/io/ADIOS2_utils.h
```diff
@@ -0,0 +1,69 @@
...
+
+  /// @brief  Get the IO object
+  std::shared_ptr<adios2::IO> io() { return _io; }
+
+  /// @brief  Close the Engine object
+  std::shared_ptr<adios2::Engine> engine() { return _engine; }
+
+protected:
+  std::shared_ptr<adios2::ADIOS> _adios;
+  std::shared_ptr<adios2::IO> _io;
```
jhale, line 63: Shared with who?

cpp/dolfinx/io/ADIOS2_utils.h
```diff
@@ -0,0 +1,69 @@
...
+  // Copy assignment
+  ADIOS2Engine& operator=(const ADIOS2Engine&) = delete;
+
+  /// @brief  Close the file
+  void close();
+
+  /// @brief  Get the IO object
+  std::shared_ptr<adios2::IO> io() { return _io; }
+
+  /// @brief  Close the Engine object
```
jhale, line 58: Typo

cpp/dolfinx/io/ADIOS2_utils.h
```diff
@@ -0,0 +1,69 @@
...
+#include <mpi.h>
+
+/// @file ADIOS2_utils.h
+/// @brief Utils for ADIOS2
+
+namespace dolfinx::io
+{
+
+/// ADIOS2-based writers/readers
+class ADIOS2Engine
```
jhale, line 23: I'd rather set this up myself, or at most have a factory function.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,151 @@
...
+  std::uint32_t num_nodes_local = imap->size_local();
+  std::uint64_t offset = imap->local_range()[0];
+
+  const std::shared_ptr<const dolfinx::common::IndexMap> topo_imap
+      = topology->index_map(mesh_dim);
+  std::uint64_t num_cells_global = topo_imap->size_global();
+  std::uint32_t num_cells_local = topo_imap->size_local();
+  std::uint64_t cell_offset = topo_imap->local_range()[0];
+
+  auto cmap = mesh->geometry().cmap();
```
jhale, line 53: Do we have a clear idea of how this will extend to the mixed topology mesh case?

cpp/dolfinx/io/ADIOS2_utils.h
```diff
@@ -0,0 +1,69 @@
...
+{
+public:
+  /// @brief Create an ADIOS2-based engine writer/reader
+  /// @param[in] comm The MPI communicator
+  /// @param[in] filename Name of output file
+  /// @param[in] tag The ADIOS2 object name
+  /// @param[in] engine ADIOS2 engine type. See
+  /// https://adios2.readthedocs.io/en/latest/engines/engines.html.
+  /// @param[in] mode ADIOS2 mode, default is Write or Read
+  ADIOS2Engine(MPI_Comm comm, const std::filesystem::path& filename,
```
jhale, line 33: Pass by reference needed for filesystem?

cpp/dolfinx/io/ADIOS2_utils.h
```diff
@@ -0,0 +1,69 @@
...
+               std::string tag, std::string engine = "BP5",
+               const adios2::Mode mode = adios2::Mode::Write);
```
jhale, line 35: Implicit write seems odd; no default?

jhale (changes requested):
Overall I question the need for an Engine class for a library with a first class C++ interface.

HDF5 and PETSc are wrapped because they have C interfaces with manual memory management.

Additionally if DOLFINx C++ consumed native ADIOS2 objects it may be able to consume Python wrapped ADIOS2 objects.

Seeing a Python interface and the reader would be useful in terms of review.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+  imap->local_to_global(topology_array_span, topology_array_global);
+
+  for (std::size_t i = 0; i < num_cells_local + 1; ++i)
+    topology_offsets[i] += cell_offset * num_dofs_per_cell;
+
+  const std::span<const int32_t> topology_offsets_span(topology_offsets.begin(),
+                                                       num_cells_local + 1);
+
+  io->DefineAttribute<std::string>("name", mesh->name);
```
jhale, line 83: Need to add metadata e.g. DOLFINx version, DOLFINx git hash etc.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,151 @@
...
+#ifdef HAS_ADIOS2
+
+#include "checkpointing.h"
+#include "ADIOS2_utils.h"
+#include <adios2.h>
+#include <basix/finite-element.h>
+#include <dolfinx/mesh/Mesh.h>
+#include <mpi.h>
+
+using namespace dolfinx;
```
jhale, line 16: Your using fully typed names in a most places so not sure this is necessary.

cpp/dolfinx/io/checkpointing.cpp
ampdes: It was suggested to put this map where ever it is needed instead of placing it in Basix.
See https://github.com/FEniCS/basix/pull/839#issuecomment-2209534926

cpp/dolfinx/io/ADIOS2_utils.cpp
ampdes: This file is adapted (almost verbatim) from the ADIOS2Writers.*
Hence the Copyright header was extended.

cpp/dolfinx/io/checkpointing.cpp
ampdes: Yes, tried this. Done.

cpp/dolfinx/io/checkpointing.cpp
ampdes: meaning use `array`, `offsets` instead of `topology_array`, `topology_offsets`?

cpp/dolfinx/io/checkpointing.cpp
ampdes: Replaced the uses of auto with explicit types.

cpp/dolfinx/io/checkpointing.cpp
ampdes: Yes it was, fixed.

cpp/dolfinx/io/checkpointing.cpp
ampdes: Fixed.

cpp/dolfinx/io/ADIOS2_utils.h
ampdes: Fixed.

cpp/dolfinx/io/checkpointing.cpp
ampdes: @chrisrichardson can comment on the data structure of how mixed topology is implemented.

Similar to the example from meshtags, offsets and global size can be computed as

```
// Compute the global offset for owned (local) vertex indices
  std::uint64_t local_start = 0;
  {
    MPI_Exscan(&num_saved_tag_entities, &local_start, 1, MPI_UINT64_T, MPI_SUM,
               mesh->comm());
  }

  std::uint64_t num_tag_entities_global = 0;
  MPI_Allreduce(&num_saved_tag_entities, &num_tag_entities_global, 1,
                MPI_UINT64_T, MPI_SUM, mesh->comm());
```

cpp/dolfinx/io/checkpointing.cpp
ampdes: removed `dolfinx::`

cpp/dolfinx/io/checkpointing.cpp
ampdes: How to do this?

I tried

```
#include <dolfinx/common/version.h.in>

  std::cout << "version" << DOLFINX_VERSION_STRING << std::endl;

  std::cout << "DOLFINX_GIT_COMMIT_HASH" << DOLFINX_VERSION_GIT << std::endl;
```

which prints

```
version@DOLFINX_VERSION@
DOLFINX_GIT_COMMIT_HASH@GIT_COMMIT_HASH@
version@DOLFINX_VERSION@
DOLFINX_GIT_COMMIT_HASH@GIT_COMMIT_HASH@

```

cpp/dolfinx/io/ADIOS2_utils.h
ampdes: Default `mode` is `adios2::Mode::Write`

For a reader,  we need to specify `adios2::Mode::Read` explicitly.

Did you mean this or did I not understand your comment?

cpp/dolfinx/io/checkpointing.cpp
ampdes: Fixed. Using `*mesh`

cpp/dolfinx/io/ADIOS2_utils.h
```diff
@@ -0,0 +1,69 @@
...
+  void close();
+
+  /// @brief  Get the IO object
+  std::shared_ptr<adios2::IO> io() { return _io; }
+
+  /// @brief  Close the Engine object
+  std::shared_ptr<adios2::Engine> engine() { return _engine; }
+
+protected:
+  std::shared_ptr<adios2::ADIOS> _adios;
```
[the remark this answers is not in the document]
ampdes: Wrapping ADIOS2 is dropped.
Now the write function needs IO and Engine objects created by the user directly.

cpp/demo/checkpointing/main.cpp
```diff
@@ -0,0 +1,41 @@
...
+  MPI_Init(&argc, &argv);
+
+  // Create mesh and function space
+  auto part = mesh::create_cell_partitioner(mesh::GhostMode::shared_facet);
+  auto mesh = std::make_shared<mesh::Mesh<float>>(mesh::create_rectangle<float>(
+      MPI_COMM_WORLD, {{{0.0, 0.0}, {1.0, 1.0}}}, {4, 4},
+      mesh::CellType::quadrilateral, part));
+
+  // Set up ADIOS2 IO and Engine
+  adios2::ADIOS adios(mesh->comm());
```
jhale, line 30: Thanks. I think this is much better. It's not consistent with the Fides writer interface but don't worry about that for now.

jhale:
Next thing to do would be to tidy up and try this out from Python using the ADIOS2 Python bindings. Please note the recent changes in adios 2.10 to the Python bindings.

python/demo/demo_checkpointing.py
```diff
@@ -0,0 +1,60 @@
...
+filename = "mesh.bp"
+engine_type = "BP5"
+adios = adios2.bindings.ADIOS(msh.comm)
+adios2_io = adios.DeclareIO("mesh-write")
+adios2_io.SetEngine(engine_type)
+adios2_engine = adios2_io.Open(filename, adios2.bindings.Mode.Write)
+# -
+
+# +
+io.write_mesh(adios2_io, adios2_engine, msh)
```
jhale, line 58: Does this work?

ampdes: Oh no!, It doesn't, I forgot to delete remove that commit.

cpp/dolfinx/io/checkpointing.cpp
jhale: These are replaced with the correct strings only after `cmake install` and the correct include is then `#include <dolfinx/common/version.h>`

python/demo/demo_checkpointing.py
jhale: There is some discussion here but it all looks a bit hacky to me.

https://github.com/wjakob/nanobind/discussions/83

cpp/demo/checkpointing/main.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+// # Checkpointing
+//
+
+#include <adios2.h>
+#include <dolfinx.h>
+#include <dolfinx/io/checkpointing.h>
+#include <mpi.h>
+
+using namespace dolfinx;
```
jhale, line 15: We avoid `using ...`, only leads to problems, please remove.

jhale: My bad, I misread - it's OK in user code, just avoid it in library.

cpp/dolfinx/io/checkpointing.h
```diff
@@ -0,0 +1,33 @@
...
+{
+
+/// @brief Write mesh to a file.
+///
+/// @param[in] io ADIOS2 IO
+/// @param[in] engine ADIOS2 Engine
+/// @param[in] mesh Mesh of type float or double to write to the file
+template <std::floating_point T>
+void write_mesh(adios2::IO io, adios2::Engine engine,
```
jhale, line 28: These should all be passed by reference, not value (heavy objects).

cpp/dolfinx/io/ADIOS2_utils.cpp
ampdes: Done.

cpp/dolfinx/io/ADIOS2_utils.h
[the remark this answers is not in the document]
ampdes: ADIOS2 wrapping is introduced back so that python bindings can be written easily.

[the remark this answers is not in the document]
jhale: Looks good, but there should be no need to use the wrapper from the C++ interface.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+    adios2::Variable<T> var_x = io->InquireVariable<T>("x");
+
+    adios2::Variable<std::int64_t> var_topology_array
+        = io->InquireVariable<std::int64_t>("topology_array");
+
+    adios2::Variable<std::int32_t> var_topology_offsets
+        = io->InquireVariable<std::int32_t>("topology_offsets");
+
+    if (var_input_global_indices)
```
jhale, line 260: If the variables don't exist you cannot return meaningful output. Raise an exception immediately after inquiring.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+  {
+    for (std::uint32_t j = 0; j < (std::uint32_t)dim; ++j)
+      x_reduced[i * dim + j] = x[i * 3 + j];
+  }
+
+  fem::CoordinateElement<T> element
+      = fem::CoordinateElement<T>(cell_type, degree, lagrange_variant);
+
+  std::array<std::size_t, 2> xshape = {num_vertices_local, (std::uint32_t)dim};
+  auto part = mesh::create_cell_partitioner(mesh::GhostMode::shared_facet);
```
jhale, line 305: Ghosting cannot be fixed - pass as argument.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+/// @endcond
+
+//-----------------------------------------------------------------------------
+std::string query_type(ADIOS2Wrapper& ADIOS2)
+{
+  auto io = ADIOS2.io();
+  auto engine = ADIOS2.engine();
+  engine->BeginStep();
+  std::string floating_point = io->VariableType("x");
+  engine->EndStep();
```
jhale, line 329: Could this be refactored to start the step, query the dtype, dispatch, read the mesh, then end the step?

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+  return floating_point;
+}
+
+//-----------------------------------------------------------------------------
+std::variant<dolfinx::mesh::Mesh<float>, dolfinx::mesh::Mesh<double>>
+read_mesh_variant(ADIOS2Wrapper& _ADIOS2, ADIOS2Wrapper& ADIOS2, MPI_Comm comm)
```
jhale, line 336: This should be read_mesh and the main method should be hidden inside an impl namespace.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+      x_reduced[i * dim + j] = x[i * 3 + j];
+  }
+
+  fem::CoordinateElement<T> element
+      = fem::CoordinateElement<T>(cell_type, degree, lagrange_variant);
+
+  std::array<std::size_t, 2> xshape = {num_vertices_local, (std::uint32_t)dim};
+  auto part = mesh::create_cell_partitioner(mesh::GhostMode::shared_facet);
+
+  mesh::Mesh<T> mesh = mesh::create_mesh(comm, comm, array, element, comm,
```
jhale, line 307: As this will be a low level function for constructing arrays of data it should just return the low level arrays, the higher level read_mesh can do the actual creation of the mesh.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+template dolfinx::mesh::Mesh<float> read_mesh<float>(ADIOS2Wrapper& ADIOS2,
+                                                     MPI_Comm comm);
+
+template dolfinx::mesh::Mesh<double> read_mesh<double>(ADIOS2Wrapper& ADIOS2,
+                                                       MPI_Comm comm);
+
+/// @endcond
+
+//-----------------------------------------------------------------------------
+std::string query_type(ADIOS2Wrapper& ADIOS2)
```
jhale, line 323: This should not be part of the public API.

python/demo/demo_checkpointing.py
```diff
@@ -0,0 +1,84 @@
...
+msh = mesh.create_rectangle(
+    comm=MPI.COMM_WORLD,
+    points=((0.0, 0.0), (1.0, 1.0)),
+    n=(4, 4),
+    cell_type=mesh.CellType.triangle,
+)
+# -
+
+# +
+filename = "mesh.bp"
```
jhale, line 53: Unnecessary strings, pass directly.

python/dolfinx/io/utils.py
```diff
@@ -186,6 +201,29 @@ def write(self, t: float):
...
+        """Write mesh to a file using ADIOS2"""
+        dtype = mesh.geometry.x.dtype  # type: ignore
+        if np.issubdtype(dtype, np.float32):
+            _writer = _cpp.io.write_mesh_float32
+        elif np.issubdtype(dtype, np.float64):
+            _writer = _cpp.io.write_mesh_float64
+
+        return _writer(ADIOS2, mesh._cpp_object)
+
+    def read_mesh(_ADIOS2: ADIOS2, ADIOS2: ADIOS2, comm: _MPI.Comm) -> Mesh:
```
jhale, line 221: Not clear how or why MPI comm can be different to the one used to construct the Adios2 reader/writer.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+    engine->Get(var_num_vertices, num_vertices_global);
+    engine->Get(var_num_cells, num_cells_global);
+    engine->Get(var_num_dofs_per_cell, num_dofs_per_cell);
+  }
+
+  // Compute local sizes, offsets
+  std::array<std::int64_t, 2> _local_range
+      = dolfinx::MPI::local_range(rank, num_vertices_global, size);
+
+  std::array<std::uint64_t, 2> local_range{(std::uint64_t)_local_range[0],
```
jhale, line 232: What is happening here if the size of the MPI communicator has changed?

cpp/demo/checkpointing/main.cpp
```diff
@@ -0,0 +1,123 @@
...
+  io->DefineAttribute<std::string>("version", DOLFINX_VERSION_STRING);
+  io->DefineAttribute<std::string>("git_hash", DOLFINX_VERSION_GIT);
+
+  // io::checkpointing::write_mesh(io, engine, *mesh);
+  io::checkpointing::write_mesh(adios, *mesh);
+
+  adios.close();
+
+  // ----------------------------------------------------------------------
+  auto adios_query
```
jhale, line 49: This step should not be part of the public API.

cpp/dolfinx/io/ADIOS2_utils.cpp
```diff
@@ -0,0 +1,45 @@
...
+std::map<std::string, adios2::Mode> string_to_mode{
+    {"write", adios2::Mode::Write},
+    {"read", adios2::Mode::Read},
+};
+} // namespace
+
+using namespace dolfinx::io;
+
+//-----------------------------------------------------------------------------
+ADIOS2Wrapper::ADIOS2Wrapper(MPI_Comm comm, std::string filename,
```
jhale, line 24: This class should additionally support passing a path to an Adios2 configuration file. This will allow Python users to setup the Adios2 with all of the supported options. Not ideal but it will work.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+read_mesh_variant(ADIOS2Wrapper& _ADIOS2, ADIOS2Wrapper& ADIOS2, MPI_Comm comm)
+{
+  std::string floating_point = query_type(_ADIOS2);
+
+  if (floating_point == "float")
+  {
+    dolfinx::mesh::Mesh<float> mesh = read_mesh<float>(ADIOS2, comm);
+    return mesh;
+  }
+  else // floating_point == "double"
```
jhale, line 345: Else implying equality is a bug waiting to be triggered.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+  {
+    dolfinx::mesh::Mesh<double> mesh = read_mesh<double>(ADIOS2, comm);
+    return mesh;
+  }
```
jhale, line 349: Throw an exception if dtype is not supported.

cpp/demo/checkpointing/main.cpp
```diff
@@ -0,0 +1,123 @@
...
+  // since, in that case BeginStep and EndStep will be called twice
+  // on a dataset written with a single Step
+  auto mesh_read_variant = io::checkpointing::read_mesh_variant(
+      adios_query, adios_read, mesh->comm());
+
+  // We cannot resolve easily the variant
+  // Hence the user can query the type and
+  // call the correct read_mesh
+  auto mesh_read = std::get<0>(mesh_read_variant);
+  // auto mesh_read = std::get<mesh::Mesh<float>>(mesh_read_variant);
```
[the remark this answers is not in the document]
ampdes: Both are fine.

Index fetching via `variant.index()` was tried which fetches the index which is of type `constexpr std::size_t` which can be used only if we have constexpr for the Mesh as well.

cpp/demo/checkpointing/main.cpp
```diff
@@ -0,0 +1,123 @@
...
+  auto adios_query
+      = ADIOS2Wrapper(mesh->comm(), "mesh.bp", "mesh-read", "BP5", "read");
+
+  auto adios_read
+      = ADIOS2Wrapper(mesh->comm(), "mesh.bp", "mesh-read", "BP5", "read");
+
+  // Can't use the same engine to query as well as read
+  // since, in that case BeginStep and EndStep will be called twice
+  // on a dataset written with a single Step
+  auto mesh_read_variant = io::checkpointing::read_mesh_variant(
```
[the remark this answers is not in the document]
ampdes: We need `constexpr` friendly Mesh to do so.

cpp/dolfinx/io/checkpointing.cpp
ampdes: Determining the sizes for each processor in a balanced manner. 
The existing function is based on `int64_t` and a type cast is don't here. 

The existing function `local_range` can be templated.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+/// @file checkpointing.h
+/// @brief ADIOS2 based checkpointing
+namespace
+{
+std::map<basix::element::lagrange_variant, std::string> variant_to_string{
+    {basix::element::lagrange_variant::unset, "unset"},
+    {basix::element::lagrange_variant::equispaced, "equispaced"},
+    {basix::element::lagrange_variant::gll_warped, "gll_warped"},
+};
```
jorgensd, line 25:
```suggestion
std::map<basix::element::lagrange_variant, std::string> variant_to_string{
    {basix::element::lagrange_variant::unset, "unset"},
    {basix::element::lagrange_variant::equispaced, "equispaced"},
    {basix::element::lagrange_variant::gll_warped, "gll_warped"},
};
```
There should potentially be a wrapper for this in Basix. @mscroggs  what do you think?

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+    num_cells_global = topo_imap->size_global();
+    num_cells_local = topo_imap->size_local();
+    cell_offset = topo_imap->local_range()[0];
+  }
+
+  // Coordinate element information
+  {
+    const fem::CoordinateElement<T>& cmap = mesh.geometry().cmap();
+    fem::ElementDofLayout geom_layout = cmap.create_dof_layout();
+    num_dofs_per_cell = geom_layout.num_entity_closure_dofs(dim);
```
jorgensd, line 76:
```suggestion
    num_dofs_per_cell = geom_layout.num_entity_closure_dofs(topology()->dim());
```
This should be the topological dimension.
Imagine a 2D manifold embedded in 3D space. The num entity closure dofs is still of dimension 2 (associated with the 2D cell), not 3.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+  std::uint32_t num_vertices_local, num_cells_local, num_dofs_per_cell, degree;
+  std::string cell_type, lagrange_variant;
+  std::vector<std::int64_t> array_global;
+  std::vector<std::int32_t> offsets_global;
+
+  std::shared_ptr<const common::IndexMap> geom_imap;
+
+  // Vertices information
+  {
+    geom_imap = mesh.geometry().index_map();
```
jorgensd, line 57:
```suggestion
    geom_imap = geometry.index_map();
```

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+    const std::shared_ptr<const common::IndexMap> topo_imap
+        = topology->index_map(dim);
+    num_cells_global = topo_imap->size_global();
+    num_cells_local = topo_imap->size_local();
+    cell_offset = topo_imap->local_range()[0];
+  }
+
+  // Coordinate element information
+  {
+    const fem::CoordinateElement<T>& cmap = mesh.geometry().cmap();
```
jorgensd, line 74:
```suggestion
    const fem::CoordinateElement<T>& cmap = geometry.cmap();
```

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+  std::vector<std::int64_t> array_global;
+  std::vector<std::int32_t> offsets_global;
+
+  std::shared_ptr<const common::IndexMap> geom_imap;
+
+  // Vertices information
+  {
+    geom_imap = mesh.geometry().index_map();
+    num_vertices_global = geom_imap->size_global();
+    num_vertices_local = geom_imap->size_local();
```
jorgensd, line 59:
```suggestion
    num_vertices_global = geom_imap->size_global();
    num_vertices_local = geom_imap->size_local();
```
These are not vertices for higher order cells, but nodes.

cpp/dolfinx/io/checkpointing.cpp
mscroggs: Think it used to exist but we removed it... or maybe it only existed in the Python wrapper

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+    num_dofs_per_cell = geom_layout.num_entity_closure_dofs(dim);
+    cell_type = mesh::to_string(cmap.cell_shape());
+    degree = cmap.degree();
+    lagrange_variant = variant_to_string[cmap.variant()];
+  }
+
+  const std::vector<int64_t> input_global_indices
+      = geometry.input_global_indices();
+  const std::span<const int64_t> input_global_indices_span(
+      input_global_indices.begin(), num_vertices_local);
```
jorgensd, line 85: Do we need this span, or could we just use `input_global_indices.data()` in Line 151?

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+    const fem::CoordinateElement<T>& cmap = mesh.geometry().cmap();
+    fem::ElementDofLayout geom_layout = cmap.create_dof_layout();
+    num_dofs_per_cell = geom_layout.num_entity_closure_dofs(dim);
+    cell_type = mesh::to_string(cmap.cell_shape());
+    degree = cmap.degree();
+    lagrange_variant = variant_to_string[cmap.variant()];
+  }
+
+  const std::vector<int64_t> input_global_indices
+      = geometry.input_global_indices();
```
jorgensd, line 83:
```suggestion
  const std::vector<int64_t>& input_global_indices
      = geometry.input_global_indices();
```
This could be const right?

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+  const std::span<const int64_t> input_global_indices_span(
+      input_global_indices.begin(), num_vertices_local);
+  const std::span<const T> mesh_x = geometry.x();
+
+  // Connectivity
+  {
+    std::shared_ptr<const graph::AdjacencyList<std::int32_t>> connectivity
+        = topology->connectivity(dim, 0);
```
jorgensd, line 91: This `dim` should also be tdim.

cpp/dolfinx/io/checkpointing.cpp
```diff
@@ -0,0 +1,354 @@
...
+    array_global.resize(offsets[num_cells_local]);
+    offsets_global.resize(num_cells_local + 1);
+
+    std::iota(array_global.begin(), array_global.end(), 0);
+
+    geom_imap->local_to_global(array_span, array_global);
+
+    for (std::size_t i = 0; i < num_cells_local + 1; ++i)
+      offsets_global[i] = offsets[i] + cell_offset * num_dofs_per_cell;
+  }
```
jorgensd, line 107: I do not think the connectivity here is quite right.
You would have to map between the `mesh->topology()` and `mesh->geometry()` indices to make this correct.
This is for instance what is done in `dolfinx::mesh::entities_to_geometry`: https://github.com/FEniCS/dolfinx/blob/8e4122b5b28a8997b94148797125becf5ba704cb/cpp/dolfinx/mesh/utils.h#L634
which  you in turn can use `geom_imap->local_to_global` on. 

I don't even think we need this, as you could simply use the `geometry.dofmap()`  (https://github.com/FEniCS/dolfinx/blob/8e4122b5b28a8997b94148797125becf5ba704cb/cpp/dolfinx/mesh/Geometry.h#L121-L134) to get the local indices per cell and then in turn call `geom_imap->local_to_global(...)` on the array created from this mdspan.

cpp/dolfinx/io/ADIOS2_utils.cpp
ampdes: Done.

cpp/dolfinx/io/checkpointing.cpp
ampdes: A new function is added similar to local_range to compute offset and local size directly.

cpp/dolfinx/io/checkpointing.cpp
ampdes: Thanks. Updated to using `geometry.dofmap()`.

cpp/dolfinx/io/checkpointing.cpp
ampdes: This is deleted and not needed.

python/dolfinx/io/utils.py
ampdes: ADIOS2 doesn't have a public API to fetch its `MPI_Comm`.
So we have to pass the `comm` separately.

cpp/dolfinx/io/checkpointing.cpp
ampdes: Now there are functions to read_geometry and read_topology and read_mesh calls them and constructs the mesh.

read_mesh_variant is now in a impl namespace and it is the one exposed in python.

cpp/demo/checkpointing/main.cpp
```diff
@@ -40,6 +40,13 @@ int main(int argc, char* argv[])
     adios2::Engine engine = io.Open("mesh.bp", adios2::Mode::Append);
 
     io::native::write_mesh(io, engine, *mesh);
+    std::span<float> x = mesh->geometry().x();
+    for (std::size_t i = 0; i < x.size(); ++i)
+    {
+      x[i] *= 4;
+    }
```
jorgensd, line 47:
```suggestion
    for (std::size_t i = 0; i < x.size(); ++i)
    {
      x[i] *= 4;
    }
```
We could use `std::for_each` to illustrate good coding practices, see:
https://godbolt.org/z/97osavhb6

PR #3292, opened 2024-07-04
Use std::ranges

schnellerhase:
After passing in basix https://github.com/FEniCS/basix/pull/837 we are ready to make use of `std::ranges` in dolfinx.

### DIFF
--- a/cpp/demo/hyperelasticity/main.cpp
+++ b/cpp/demo/hyperelasticity/main.cpp
@@ -81,7 +81,7 @@ public:
     {
       // Assemble b and update ghosts
       std::span<T> b(_b.mutable_array());
-      std::fill(b.begin(), b.end(), 0.0);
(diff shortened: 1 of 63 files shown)


schnellerhase:
There are a few edge cases left out that need more detailed consideration. These are mainly related to use cases of

- `partial_sum`
- `transform`
- `reduce`
- `iota`
- `accumulate`
- `transform_reduce`

jhale:
We can do the trickier ones one PR at a time. Overall looks very nice.

schnellerhase:
Alright, just the bug I introduced left then :laughing:

schnellerhase:
Okay pretty sure I messed up the `std::ranges::unique` usage.

schnellerhase:
Okay not sure whats breaking here, but too many changes to make debugging feasible. Will split this one up in smaller PR's.

PR #3293, opened 2024-07-04
Replace `std::sort` with `std::ranges::sort`

schnellerhase:
And `std::is_sorted` replaced by `std::ranges::is_sorted`


chrisrichardson (approved):
This is all C++20 compliant, so should be no problem.

schnellerhase:
All of the related PR's are c++20 compliant, we haven't upgraded the c++ standard in dolfinx.

PR #3295, opened 2024-07-04
Replace `std::lower/uper_bound` with `std::ranges::lower/uper_bound`

schnellerhase:


schnellerhase:
Missing a couple `std::lower_bound` cases which are all related to more complicated comparison lambdas - need to investigate.

PR #3297, opened 2024-07-04
Replace `std::transform` with `std::ranges::transform`

schnellerhase:


garth-wells:
Pure functions are preferable whenever reasonable. Any reason for not using transform from ranges?

schnellerhase:
`ranges::transform` works analogously - usage of `for_each` improves readability and decreases code complexity, i.e. it is easy to miss that the `transform` version performs an in place operation.

schnellerhase:
Tried with godbolt (https://godbolt.org/z/5e871aWKG), seems like for optimization `-01` and higher at least the scaling operation produces the same assembly even, not sure how this extends to the more complicated cases - but this is the setting where would hope that pure functions are beneficial, right?

jhale:
Another big difference in favour `transform` algorithms is that it doesn't demand the operation takes place in order. This is potentially good for performance and/or parallelisation.

I take your point on safety but I think everyone who working at this level is clear that `transform` works as an in place operation.

jhale:
Could you change the title? We use the PR titles to create Changelogs.

schnellerhase:
Since this is now just the upgrade for `transform()`, I also added the remaining cases.

garth-wells:
> Tried with godbolt (https://godbolt.org/z/5e871aWKG), seems like for optimization `-01` and higher at least the scaling operation produces the same assembly even, not sure how this extends to the more complicated cases - but this is the setting where we would hope that pure functions are beneficial, right?

The preference for pure functions is that the logic is simpler and the design models mathematical function. We pass non-const references only when performance is critical.

cpp/dolfinx/mesh/utils.h
```diff
@@ -1251,9 +1245,9 @@ create_subgeometry(const Mesh<T>& mesh, int dim,
   const std::vector<std::int64_t>& igi = geometry.input_global_indices();
   std::vector<std::int64_t> sub_igi;
   sub_igi.reserve(subx_to_x_dofmap.size());
-  std::transform(subx_to_x_dofmap.begin(), subx_to_x_dofmap.end(),
-                 std::back_inserter(sub_igi),
-                 [&igi](std::int32_t sub_x_dof) { return igi[sub_x_dof]; });
+  std::ranges::transform(subx_to_x_dofmap, std::back_inserter(sub_igi),
+                         [&igi](std::int32_t sub_x_dof)
```
garth-wells, line 1249:
```suggestion
                         [&igi](auto sub_x_dof)
```

garth-wells (changes requested):

garth-wells (approved):

PR #3300, opened 2024-07-04
Replace `std::min/max` with `std::ranges::min/max`

schnellerhase:


jhale:
These all look fine to me. @garth-wells ?

Issue #3301, opened 2024-07-05
[BUG]: Can't (easily) assemble a block vector without providing a left-hand side in Python


jpdean:
### Summarize the issue

Calling `assemble_vector_block` in Python fails with `RuntimeError: Unable to extract a mesh.` if any rows of the bilinear forms provided are all `None`. This happens even when no boundary conditions need applying.

This problem occurs because `assemble_vector_block` calls `apply_lifting` on each row, which gets the mesh from the form (see [here](https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/assembler.h#L158-L168)). Hence, if there forms are all `nullptr`, it can't extract a mesh.

There are a couple of solutions to this:

1. If there is nothing to lift, `apply_lifting` should probably just return without doing anything. This would be done at the C++ level.
2. Don't call `apply_lifting` when there is nothing to lift. This would mean adding a check in Python to `assemble_vector_block`.

I think I prefer the first option. Regardless, I think it would be good to separate the lifting and bc application out of `assemble_vector_block` so we'd have `assemble_vector_block`, `apply_lifting_block`, and `set_bc_block`. This is consistent with non-blocked problems and gives more control to the user.

### How to reproduce the bug

Run the minimal example below.

### Minimal Example (Python)

```Python
from mpi4py import MPI
from dolfinx import mesh, fem
import ufl
import dolfinx.fem.petsc

comm = MPI.COMM_WORLD
msh = mesh.create_unit_square(comm, 2, 2)

V = fem.functionspace(msh, ("Lagrange", 1))
W = fem.functionspace(msh, ("Lagrange", 2))

v = ufl.TestFunction(V)
w = ufl.TestFunction(W)

L = fem.form([v * ufl.dx, w * ufl.dx])

b = dolfinx.fem.petsc.assemble_vector_block(L, [[None, None], [None, None]])
```

### Version

main branch

jorgensd:
Does this also happen if you try to call `dolfinx.fem.apply_lifting` without a list of `a`'s ?

jpdean:
You can't call `apply_lifting` without providing `a`, it's a required argument.

PR #3302, opened 2024-07-05
Only try to lift boundary conditions if forms are provided

jpdean:
Resolves https://github.com/FEniCS/dolfinx/issues/3301

### DIFF
--- a/cpp/dolfinx/fem/assembler.h
+++ b/cpp/dolfinx/fem/assembler.h
@@ -11,6 +11,7 @@
 #include "assemble_vector_impl.h"
 #include "traits.h"
 #include "utils.h"
+#include <algorithm>
 #include <cstdint>
 #include <dolfinx/common/types.h>
 #include <memory>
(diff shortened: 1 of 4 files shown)


nate-sime:
Should we add a test for this?

cpp/dolfinx/fem/assembler.h
```diff
@@ -155,6 +155,10 @@ void apply_lifting(
         bcs1,
     const std::vector<std::span<const T>>& x0, T scale)
 {
+  // If all forms are null, there is nothing to do
+  if (std::all_of(a.begin(), a.end(), [](auto ptr) { return ptr == nullptr; }))
```
jhale, line 159: You can use std::ranges::all_of here.

python/test/unit/fem/test_assembler.py
```diff
@@ -1139,6 +1139,17 @@ def block():
...
 
+    def test_block_null_lifting(self):
+        from dolfinx.fem.petsc import assemble_vector_block
+
+        comm = MPI.COMM_WORLD
+        msh = create_unit_square(comm, 2, 2)
+        V = functionspace(msh, ("Lagrange", 1))
+        W = functionspace(msh, ("Lagrange", 2))
+        v, w = ufl.TestFunction(V), ufl.TestFunction(W)
+        L = form([v * ufl.dx, w * ufl.dx])
```
jorgensd, line 1150:
```suggestion
        L = form([ufl.conj(v) * ufl.dx, ufl.conj(w) * ufl.dx])
```
To support complex assembly

PR #3304, opened 2024-07-09
Reorganize sparsity pattern creation

schnellerhase:
1. Create `create_sparsity_pattern.h/cpp` - as a central location for the name giving functionality
2. Transfer implementation details from `sparsitybuild.h/.cpp` to `impl` namespace inside `create_sparsity_pattern.h/.cpp`
3. Move `create_sparsity()` from `python/dolfinx/wrappers/assemble.cpp` to `create_sparsity_pattern.h` to allow for general usage. 

It should be discussed before merging this if the design choices of the free function handling are appropriate. If so maybe also make this part of the dolfinx c++ style guide then. Currently this has been:
- free functions that rectify own source file usage (otherwise stick to `utils.h`) go into header file of the functions name
- implementation details are to be put into an additional `impl` but are contained in the same files - be aware that this is a shared namespace, so choose naming schemes that do not pollute the namespace.
- no fancy dispatching in use to select the free function of choice.


garth-wells:
It's not clear to me what this PR adds.

On free functions, we balance keeping the number of files manageable whilst avoiding files becoming too long. We don't do one header file per free function. We try to keep file names short.

The code is `sparsitybuild.h` isn't implementation code. I can be used in user solvers.

schnellerhase:
This PR moves the anonymous `create_sparsity` functionality in `assemble.cpp` to a publicly usable location. And the reorganization was more a by-product of moving this functionality to the right place and to organize it within the context of its related functionality.

jhale:
This looks OK but can't we just keep the naming `sparsitybuild.h`? I see no reason to rename it.

Regarding `impl` we do use e.g. `impl_sparsitybuild` in places and this seems like it will avoid namespace clashes better.

garth-wells:
I'm going to close this PR because there is a bit too much going on. Some changes may be useful, but there are some issues, e.g. `impl::` appearing in a demo.

PR #3311, opened 2024-07-15
Fix sphinx

schnellerhase:
The recent new release of sphinx causes some new warnings. Not sure how to fix those for now, but a downgrade of the sphinx version fixes the ci.


garth-wells:
Fixed in #3302.

PR #3313, opened 2024-07-16
Introduce `sort_unique`

schnellerhase:
While working on https://github.com/FEniCS/dolfinx/pull/3293, noticed that a lot of the use cases of `sort` are coupled with a following `unique` call to produce a sorted and unique range.

Introduces `common::sort_unique` which sorts and uniquifies a range, given a custom sort routine. To allow for the passing of `radix_sort` its interface is made `std::ranges` compatible, i.e. functionality is exposed by `opeartor()` of a struct.

### DIFF
--- a/cpp/dolfinx/common/sort.h
+++ b/cpp/dolfinx/common/sort.h
@@ -10,6 +10,7 @@
 #include <algorithm>
 #include <bitset>
 #include <cstdint>
+#include <iterator>
 #include <numeric>
 #include <span>
 #include <type_traits>
@@ -23,9 +24,10 @@ namespace dolfinx
 /// @tparam T Integral type
 /// @tparam BITS The number of bits to sort at a time.
 /// @param[in, out] array The array to sort.
-template <typename T, int BITS = 8>
-void radix_sort(std::span<T> array)
+template <std::ranges::random_access_range R, int BITS = 8>
+void radix_sort(R&& array)
 {
+  using T = std::iter_value_t<R>;
   static_assert(std::is_integral<T>(), "This function only sorts integers.");
 
   if (array.size() <= 1)
--- a/cpp/dolfinx/common/sort.h
+++ b/cpp/dolfinx/common/sort.h
@@ -8,6 +8,7 @@
 
 #include "Timer.h"
 #include <algorithm>
+#include <bits/ranges_algo.h>
 #include <bitset>
 #include <cstdint>
 #include <iterator>
@@ -24,68 +25,74 @@ namespace dolfinx
 /// @tparam T Integral type
 /// @tparam BITS The number of bits to sort at a time.
 /// @param[in, out] array The array to sort.
-template <std::ranges::random_access_range R, int BITS = 8>
-void radix_sort(R&& array)
+struct __radix_sort
 {
-  using T = std::iter_value_t<R>;
-  static_assert(std::is_integral<T>(), "This function only sorts integers.");
-
-  if (array.size() <= 1)
-    return;
-
-  T max_value = *std::ranges::max_element(array);
-
-  // Sort N bits at a time
-  constexpr int bucket_size = 1 << BITS;
-  T mask = (T(1) << BITS) - 1;
-
-  // Compute number of iterations, most significant digit (N bits) of
-  // maxvalue
-  int its = 0;
-  while (max_value)
+  template <std::ranges::random_access_range R, int BITS = 8>
+  constexpr void operator()(R&& array) const
   {
-    max_value >>= BITS;
-    its++;
-  }
+    using T = std::iter_value_t<R>;
+    static_assert(std::is_integral<T>(), "This function only sorts integers.");
 
-  // Adjacency list arrays for computing insertion position
-  std::array<std::int32_t, bucket_size> counter;
-  std::array<std::int32_t, bucket_size + 1> offset;
+    if (array.size() <= 1)
+      return;
 
-  std::int32_t mask_offset = 0;
-  std::vector<T> buffer(array.size());
-  std::span<T> current_perm = array;
-  std::span<T> next_perm = buffer;
-  for (int i = 0; i < its; i++)
-  {
-    // Zero counter array
-    std::ranges::fill(counter, 0);
+    T max_value = *std::ranges::max_element(array);
 
-    // Count number of elements per bucket
-    for (T c : current_perm)
-      counter[(c & mask) >> mask_offset]++;
+    // Sort N bits at a time
+    constexpr int bucket_size = 1 << BITS;
+    T mask = (T(1) << BITS) - 1;
 
-    // Prefix sum to get the inserting position
-    offset[0] = 0;
-    std::partial_sum(counter.begin(), counter.end(), std::next(offset.begin()));
-    for (T c : current_perm)
+    // Compute number of iterations, most significant digit (N bits) of
+    // maxvalue
+    int its = 0;
+    while (max_value)
     {
-      std::int32_t bucket = (c & mask) >> mask_offset;
-      std::int32_t new_pos = offset[bucket + 1] - counter[bucket];
-      next_perm[new_pos] = c;
-      counter[bucket]--;
+      max_value >>= BITS;
+      its++;
     }
 
-    mask = mask << BITS;
-    mask_offset += BITS;
+    // Adjacency list arrays for computing insertion position
+    std::array<std::int32_t, bucket_size> counter;
+    std::array<std::int32_t, bucket_size + 1> offset;
 
-    std::swap(current_perm, next_perm);
+    std::int32_t mask_offset = 0;
+    std::vector<T> buffer(array.size());
+    std::span<T> current_perm = array;
+    std::span<T> next_perm = buffer;
+    for (int i = 0; i < its; i++)
+    {
+      // Zero counter array
+      std::ranges::fill(counter, 0);
+
+      // Count number of elements per bucket
+      for (T c : current_perm)
+        counter[(c & mask) >> mask_offset]++;
+
+      // Prefix sum to get the inserting position
+      offset[0] = 0;
+      std::partial_sum(counter.begin(), counter.end(),
+                       std::next(offset.begin()));
+      for (T c : current_perm)
+      {
+        std::int32_t bucket = (c & mask) >> mask_offset;
+        std::int32_t new_pos = offset[bucket + 1] - counter[bucket];
+        next_perm[new_pos] = c;
+        counter[bucket]--;
+      }
+
+      mask = mask << BITS;
+      mask_offset += BITS;
+
+      std::swap(current_perm, next_perm);
+    }
+
+    // Copy data back to array
+    if (its % 2 != 0)
+      std::ranges::copy(buffer, array.begin());
   }
+};
 
-  // Copy data back to array
-  if (its % 2 != 0)
-    std::ranges::copy(buffer, array.begin());
-}
+inline constexpr __radix_sort radix_sort{};
 
 /// Returns the indices that would sort (lexicographic) a vector of
 /// bitsets.
--- a/cpp/dolfinx/common/sort.h
+++ b/cpp/dolfinx/common/sort.h
@@ -20,13 +20,14 @@
 namespace dolfinx
 {
 
-/// Sort a vector of integers with radix sorting algorithm. The bucket
-/// size is determined by the number of bits to sort at a time (2^BITS).
-/// @tparam T Integral type
-/// @tparam BITS The number of bits to sort at a time.
-/// @param[in, out] array The array to sort.
+/// std::ranges compatible struct for radix sort.
 struct __radix_sort
 {
+  /// Sort a vector of integers with radix sorting algorithm. The bucket
+  /// size is determined by the number of bits to sort at a time (2^BITS).
+  /// @tparam T Integral type
+  /// @tparam BITS The number of bits to sort at a time.
+  /// @param[in, out] array The array to sort.
   template <std::ranges::random_access_range R, int BITS = 8>
   constexpr void operator()(R&& array) const
   {
@@ -92,6 +93,7 @@ struct __radix_sort
   }
 };
 
+/// Radix sort.
 inline constexpr __radix_sort radix_sort{};
 
 /// Returns the indices that would sort (lexicographic) a vector of
--- a/cpp/dolfinx/fem/DirichletBC.cpp
+++ b/cpp/dolfinx/fem/DirichletBC.cpp
@@ -11,6 +11,7 @@
 #include <array>
 #include <dolfinx/common/IndexMap.h>
 #include <dolfinx/common/sort.h>
+#include <dolfinx/common/utils.h>
 #include <dolfinx/mesh/Mesh.h>
 #include <dolfinx/mesh/Topology.h>
 #include <dolfinx/mesh/cell_types.h>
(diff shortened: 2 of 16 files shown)


jhale:
I like the ranges port of the radix sort in principle, although not reviewed.

Not sure about wrapping sort/unique/raise - yes it's a bit wordy, but it's canonical STL (we didn't pick C++ for conciseness ;)).

garth-wells:
Likewise, like the look of the radix sort but prefer to not wrap STL functionality.

schnellerhase:
Ok, so I'll remove the `sort_unique`. I'll open another PR with the `std::ranges`` update for the `std::unique`, left this out previously to try this first.

cpp/dolfinx/common/sort.h
```diff
@@ -23,144 +27,81 @@ namespace dolfinx
...
+    if (its % 2 != 0)
+      std::ranges::copy(buffer, range.begin());
   }
+};
 
-  if (its % 2 == 1)
-    std::ranges::copy(perm2, perm.begin());
-}
+/// Radix sort
+inline constexpr __radix_sort radix_sort{};
```
jhale, line 104: Is this `inline` necessary?

cpp/dolfinx/common/sort.h
```diff
@@ -23,144 +27,81 @@ namespace dolfinx
...
+    using I = std::iter_value_t<R>;
 
-  if (array.size() <= 1)
-    return;
+    // value type (if no projection is provided it holds I == T)
+    using T = std::remove_cvref_t<std::result_of_t<P(I)>>;
 
-  const auto [min, max] = std::ranges::minmax_element(array);
-  T range = *max - *min + 1;
+    static_assert(std::is_integral_v<T>, "This function only sorts integers.");
```
jhale, line 42: Can this be done by applying a concept/restriction to `R` or `P`?

cpp/dolfinx/common/sort.h
```diff
@@ -23,144 +27,81 @@ namespace dolfinx
 /// @tparam T Integral type
```
jhale, line 27: Documentation no longer correct.

cpp/dolfinx/common/sort.h
```diff
@@ -23,144 +27,81 @@ namespace dolfinx
...
 /// @tparam BITS The number of bits to sort at a time.
 /// @param[in, out] array The array to sort.
-template <typename T, int BITS = 8>
-void radix_sort(std::span<T> array)
+struct __radix_sort
```
jhale, line 30: Is it necessary to pack the method into a struct?

jhale:
Very nice work, a few comments that are possibly more out of ignorance.

cpp/dolfinx/common/sort.h
schnellerhase: This follows the style of the STL regarding the handling of functors and enables to use the `radix_sort` in any other STL algorithm as an inpute function.

The new design with theses structs follows the realization of lambdas, and removes the necessity of handling function pointers as template types. This reduces the amount of necessary type dispatching drastically if I remember correctly.

What we could do is move the implementation of the `radix_sort` into an anonymous namespace to separate it from its wrapping for external use.

cpp/dolfinx/common/sort.h
schnellerhase: I believe so - every time we include the `sort.h` we will have a different definition of `radix_sort` so we must allow for multiple definitions.

cpp/dolfinx/common/sort.h
jhale: Thanks for educating me!

cpp/dolfinx/common/sort.h
```diff
@@ -6,161 +6,105 @@
...
-      counter[bucket]++;
+      max_value >>= BITS;
+      its++;
     }
 
-    // Prefix sum to get the inserting position
-    offset[0] = 0;
-    std::partial_sum(counter.begin(), counter.end(), std::next(offset.begin()));
+    // Adjacency list arrays for computing insertion position
+    std::array<std::int32_t, bucket_size> counter;
```
jhale, line 66: Are these`std::int32_t` always OK? Seems strange for a generic algorithm.

cpp/test/common/sort.cpp
```diff
@@ -26,12 +27,23 @@ TEMPLATE_TEST_CASE("Test radix sort", "[vector][template]", std::int32_t,
...
-  dolfinx::radix_sort(std::span(vec));
+  dolfinx::radix_sort(vec);
 
   // Check if vector is sorted
   REQUIRE(std::ranges::is_sorted(vec));
 }
 
+TEST_CASE("Test radix sort (projection)", "[radix]")
+{
+  std::vector<int> vec = {3, 6, 2, 1, 5, 4, 0};
```
jhale, line 38: Switch to a modern int type with specific lengths (and perhaps parametrise over two or three types?).

cpp/dolfinx/common/sort.h
```diff
@@ -6,161 +6,105 @@
...
+      for (T c : current_perm)
+        counter[(proj(c) & mask) >> mask_offset]++;
+
+      // Prefix sum to get the inserting position
+      offset[0] = 0;
+      std::partial_sum(counter.begin(), counter.end(),
+                       std::next(offset.begin()));
+      for (T c : current_perm)
+      {
+        std::int32_t bucket = (proj(c) & mask) >> mask_offset;
```
jhale, line 88: Correct?

cpp/dolfinx/common/sort.h
```diff
@@ -6,161 +6,105 @@
...
+        std::int32_t new_pos = offset[bucket + 1] - counter[bucket];
```
jhale, line 89: Correct?

cpp/dolfinx/common/sort.h
schnellerhase: No :)

IgorBaratta:
Have you tested the radix algorithm for performance regression after these changes? 
It's used in performance-critical parts where we would otherwise use std::sort.

jhale:
Any comments @garth-wells?

schnellerhase:
No haven't done any such thing, how can I run/trigger them properly?

jhale:
@IgorBaratta had some benchmark code for his original PR.

https://github.com/FEniCS/dolfinx/pull/1622

schnellerhase:
Is the code that produced these plots still around?

IgorBaratta:
> Is the code that produced these plots still around?

The first test is just the creation of a cube with increasing number of cells.
You can use `dolfinx::list_timings` to query timing of create entities.

jhale:
Don't worry about boost:: and std:: etc. this was discussed before - just test against the radix in `main`.

schnellerhase:
What compile flags do I need to pass for a benchmark build? I'm assuming `CMAKE_BUILD_TYPE=Release` and what for the pip build?

jhale:
> What compile flags do I need to pass for a benchmark build? I'm assuming `CMAKE_BUILD_TYPE=Release` and what for the pip build?

I'd recommend doing the benchmark purely from C++. Release will give you -O3 which is fine.

cpp/dolfinx/common/sort.h
```diff
@@ -6,161 +6,105 @@
...
-  std::span<T> next_perm = buffer;
-  for (int i = 0; i < its; i++)
+  /// Sort a range with radix sorting algorithm. The bucket
+  /// size is determined by the number of bits to sort at a time (2^BITS).
+  /// @tparam R Type of range to be sorted.
+  /// @tparam P Projection type to be applied on range elements to produce a
+  /// sorting index.
+  /// @tparam BITS The number of bits to sort at a time.
+  /// @param[in, out] range The range to sort.
+  /// @param[in] P Element projection.
```
garth-wells, line 34: Could some more explanation be added? Would be nice to support with an example.

cpp/dolfinx/common/sort.h
```diff
@@ -6,161 +6,105 @@
...
-
-  std::vector<std::int32_t> perm2(perm.size());
-  std::span<std::int32_t> current_perm = perm;
-  std::span<std::int32_t> next_perm = perm2;
-  for (int i = 0; i < its; i++)
-  {
-    // Zero counter
-    std::ranges::fill(counter, 0);
+    // Sort N bits at a time
+    constexpr int bucket_size = 1 << BITS;
```
garth-wells, line 53: Is `int` the appropriate type?

cpp/dolfinx/common/sort.h
```diff
@@ -6,161 +6,105 @@
...
+    T mask = (T(1) << BITS) - 1;
 
-    // Count number of elements per bucket
-    for (auto cp : current_perm)
+    // Compute number of iterations, most significant digit (N bits) of
+    // maxvalue
+    int its = 0;
```
garth-wells, line 58: Is the the appropriate integer type?

cpp/dolfinx/common/sort.h
```diff
@@ -6,161 +6,105 @@
...
-      T value = array[cp] - *min;
-      std::int32_t bucket = (value & mask) >> mask_offset;
-      std::int32_t pos = offset[bucket + 1] - counter[bucket];
-      next_perm[pos] = cp;
-      counter[bucket]--;
+      // Zero counter array
+      std::ranges::fill(counter, 0);
+
+      // Count number of elements per bucket
+      for (const auto& c : current_perm)
```
garth-wells, line 79: Could this just be a copy rather than a reference?

cpp/dolfinx/common/sort.h
```diff
@@ -6,161 +6,105 @@
...
+        counter[(proj(c) & mask) >> mask_offset]++;
+
+      // Prefix sum to get the inserting position
+      offset[0] = 0;
+      std::partial_sum(counter.begin(), counter.end(),
+                       std::next(offset.begin()));
+      for (const auto& c : current_perm)
```
garth-wells, line 86: See above comment on reference vs copy.

cpp/dolfinx/common/sort.h
```diff
@@ -189,7 +133,8 @@ std::vector<std::int32_t> sort_by_perm(std::span<const T> x, std::size_t shape1)
     int col = shape1 - 1 - i;
     for (std::size_t j = 0; j < shape0; ++j)
       column[j] = x[j * shape1 + col];
-    argsort_radix<T, BITS>(column, perm);
+
+    radix_sort(perm, [&](auto index) { return column[index]; });
```
garth-wells, line 137:
```suggestion
    radix_sort(perm, [&column](auto index) { return column[index]; });
```

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -358,7 +358,8 @@ std::array<std::vector<std::int32_t>, 2> fem::locate_dofs_topological(
   // Remove duplicates
   std::vector<std::int32_t> perm(bc_dofs[0].size());
   std::iota(perm.begin(), perm.end(), 0);
-  dolfinx::argsort_radix<std::int32_t>(bc_dofs[0], perm);
+  dolfinx::radix_sort(perm, [&](auto index) { return bc_dofs[0][index]; });
```
garth-wells, line 361:
```suggestion
  dolfinx::radix_sort(perm, [&dofs = bc_dofs[0]](auto index) { return dofs[index]; });
```

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -413,7 +414,9 @@ std::array<std::vector<std::int32_t>, 2> fem::locate_dofs_topological(
     // Remove duplicates and sort
     perm.resize(sorted_bc_dofs[0].size());
     std::iota(perm.begin(), perm.end(), 0);
-    dolfinx::argsort_radix<std::int32_t>(sorted_bc_dofs[0], perm);
+    dolfinx::radix_sort(perm,
+                        [&](auto index) { return sorted_bc_dofs[0][index]; });
```
garth-wells, line 418:
```suggestion
                        [&dofs = sorted_bc_dofs[0]](auto index) { return dofs[index]; });
```

cpp/dolfinx/refinement/utils.cpp
```diff
@@ -257,7 +257,8 @@ std::array<std::vector<std::int32_t>, 2> refinement::transfer_facet_meshtag(
   // Sort values into order, based on facet indices
   std::vector<std::int32_t> sort_order(tag_values.size());
   std::iota(sort_order.begin(), sort_order.end(), 0);
-  dolfinx::argsort_radix<std::int32_t>(facet_indices, sort_order);
+  dolfinx::radix_sort(sort_order,
+                      [&](auto index) { return facet_indices[index]; });
```
garth-wells, line 261:
```suggestion
                      [&facet_indices](auto index) { return facet_indices[index]; });
```

cpp/dolfinx/refinement/utils.cpp
```diff
@@ -351,7 +352,8 @@ refinement::transfer_cell_meshtag(const mesh::MeshTags<std::int32_t>& tags0,
   // Sort values into order, based on cell indices
   std::vector<std::int32_t> sort_order(tag_values.size());
   std::iota(sort_order.begin(), sort_order.end(), 0);
-  dolfinx::argsort_radix<std::int32_t>(cell_indices, sort_order);
+  dolfinx::radix_sort(sort_order,
+                      [&](auto index) { return cell_indices[index]; });
```
garth-wells, line 356:
```suggestion
                      [&cell_indices](auto index) { return cell_indices[index]; });
```

garth-wells (changes requested):
Looks nice - hard to check in detail but test should detect any issues.

Would be good to verify performance.

cpp/dolfinx/common/sort.h
schnellerhase: Before these changes it was a copy. Since the value type `T` however now is not restricted to trivially copyable types any longer, we should stick to `const auto&` I think.

schnellerhase:
Results of [benchmark](https://github.com/schnellerhase/dolfinx/pull/24)

on `main`
```bash
2024-07-23T12:00:43+02:00
Running ./sort
Run on (16 X 4768 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 512 KiB (x8)
  L3 Unified 16384 KiB (x1)
Load Average: 3.53, 2.55, 1.80
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
-------------------------------------------------------------------------
Benchmark                               Time             CPU   Iterations
-------------------------------------------------------------------------
BM_create_box/10/iterations:5        11.9 ms         9.99 ms            5
BM_create_box/50/iterations:5        1159 ms         1159 ms            5
BM_create_box/100/iterations:5      10438 ms        10437 ms            5
```

and on `sort_unique`
```bash
2024-07-23T12:04:02+02:00
Running ./sort
Run on (16 X 4768 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x8)
  L1 Instruction 32 KiB (x8)
  L2 Unified 512 KiB (x8)
  L3 Unified 16384 KiB (x1)
Load Average: 2.21, 2.22, 1.81
***WARNING*** CPU scaling is enabled, the benchmark real time measurements may be noisy and will incur extra overhead.
-------------------------------------------------------------------------
Benchmark                               Time             CPU   Iterations
-------------------------------------------------------------------------
BM_create_box/10/iterations:5        5.87 ms         5.85 ms            5
BM_create_box/50/iterations:5        1110 ms         1110 ms            5
BM_create_box/100/iterations:5      10147 ms        10147 ms            5
```

schnellerhase:
So it seems that we even achieved quite a nice speed up throughout. But maybe somebody with access to a more stable server for benchmarking could maybe rerun this with a couple large values of `n` as well? This was only run on an average laptop so far, might not show the full picture.

jhale:
This is good enough evidence for me @schnellerhase - we're just looking for a large difference.

cpp/dolfinx/common/sort.h
schnellerhase: Implemented that. No more explicit int usage present.

PR #3314, opened 2024-07-16
One dimensional refinement routine

schnellerhase:
Implements a (topological) one dimensional mesh refinement routine.

For now this is not passed through a combined `refine(...)` interface with the 2D and 3D `plaza` refinement routines, but rather provides a standalone one dimensional case.

I'm not too sure about really having tested everything that should be tested, especially the construction of good test cases for different ghost modes is still open, but the construction of these cases is very tedious.


cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,235 @@
...
+std::tuple<mesh::Mesh<T>, std::vector<std::int32_t>>
+refine_interval(const mesh::Mesh<T>& mesh,
+                std::optional<std::span<const std::int32_t>> edges,
+                bool redistribute)
+{
+
+  if (mesh.topology()->cell_type() != mesh::CellType::interval)
+    throw std::runtime_error("Cell type not supported");
+
+  if (!mesh.topology()->index_map(1))
```
chrisrichardson, line 200: I think this can just be an `assert` as it must exist, unless there is an internal error.

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,235 @@
...
+  {
+    // Check if mesh has ghost cells on any rank
+    // FIXME: this is not a robust test. Should be user option.
+    const int num_ghost_cells = mesh.topology()->index_map(1)->num_ghosts();
+    int max_ghost_cells = 0;
+    MPI_Allreduce(&num_ghost_cells, &max_ghost_cells, 1, MPI_INT, MPI_MAX,
+                  mesh.comm());
+
+    // Build mesh
+    const auto ghost_mode = max_ghost_cells == 0
```
chrisrichardson, line 225: This is copying an existing design flaw. I think it would be better to make `ghost_mode` a function parameter, possibly with a default value of none.

chrisrichardson:
I'd rather use `cell` than `edge` because in 1D, entities of tdim 1 are cells, and it is easier to think about them as such. 

The MPI update then is really just to check if any ghost cells are marked. I'm a bit unsure if this is important or not.

schnellerhase:
Yes that wasn't clear to me either, I guess it comes down to us deciding whether we want to allow/support markings of ghost edges on a process or not.  If we can avoid another communication step that's always preferable.

The ownership of the edge is explicitly tied to a process. So, restricting the refinement marking to this process, seems like a reasonable thing to do.

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,224 @@
...
+namespace dolfinx::refinement
+{
+
+namespace impl
+{
+
+/// Refine with markers returning new mesh data.
+///
+/// @param[in] mesh Input mesh to be refined
+/// @param[in] edges Indices of the edges that are marked for refinement
```
chrisrichardson, line 33: Use 'cells' not 'edges' throughout.

schnellerhase:
```
 /__w/dolfinx/dolfinx/cpp/test/refinement/interval.cpp:286: FAILED:
  CHECK( v_to_e->links((center_index + 1) % 3)[0] != v_to_e->links((center_index + 1) % 3)[1] )
with expansion:
  1 != 1
```
That's a new one :D

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,224 @@
...
+std::tuple<graph::AdjacencyList<std::int64_t>, std::vector<T>,
+           std::array<std::size_t, 2>, std::vector<std::int32_t>>
+compute_interval_refinement(const mesh::Mesh<T>& mesh,
+                            std::optional<std::span<const std::int32_t>> cells)
+{
+  auto topology = mesh.topology();
+  assert(topology);
+  assert(topology->dim() == 1);
+
+  auto map_e = topology->index_map(1);
```
chrisrichardson, line 47:
```suggestion
  auto map_c = topology->index_map(1);
```

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,224 @@
...
+  assert(map_e);
```
chrisrichardson, line 48:
```suggestion
  assert(map_c);
```

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,224 @@
...
+
+  // TODO: creation of sharing ranks in external function? Also same code in use
+  // for plaza
+  // Get sharing ranks for each edge
+  graph::AdjacencyList<int> edge_ranks = map_e->index_to_dest_ranks();
```
chrisrichardson, line 53:
```suggestion
  graph::AdjacencyList<int> cell_ranks = map_c->index_to_dest_ranks();
```

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,224 @@
...
+
+  // Create unique list of ranks that share cells (owners of ghosts plus
+  // ranks that ghost owned indices)
+  std::vector<int> ranks = edge_ranks.array();
```
chrisrichardson, line 57:
```suggestion
  std::vector<int> ranks = cell_ranks.array();
```

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,224 @@
...
+  std::ranges::sort(ranks);
+  auto to_remove = std::ranges::unique(ranks);
+  ranks.erase(to_remove.begin(), to_remove.end());
+
+  // Convert edge_ranks from global rank to to neighbourhood ranks
```
chrisrichardson, line 62:
```suggestion
  // Convert cell_ranks from global rank to to neighbourhood ranks
```

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,224 @@
...
+  std::ranges::transform(edge_ranks.array(), edge_ranks.array().begin(),
```
chrisrichardson, line 63:
```suggestion
  std::ranges::transform(cell_ranks.array(), cell_ranks.array().begin(),
```

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,224 @@
...
+  // TODO: vector of bools? -> make of use std specialization for type bool
+  std::vector<std::int8_t> refinement_marker(
+      map_e->size_local() + map_e->num_ghosts(), !cells.has_value());
+
+  // mark edges for refinement
+  std::vector<std::vector<std::int32_t>> marked_for_update(ranks.size());
+  if (cells.has_value())
+  {
+    std::ranges::for_each(cells.value(),
+                          [&](auto edge)
```
chrisrichardson, line 82:
```suggestion
                          [&](auto cell)
```

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,224 @@
...
+                          {
+                            if (!refinement_marker[edge])
+                            {
+                              refinement_marker[edge] = true;
```
chrisrichardson, line 86:
```suggestion
                              refinement_marker[cell] = true;
```

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,224 @@
...
+                              for (int rank : edge_ranks.links(edge))
```
chrisrichardson, line 87:
```suggestion
                              for (int rank : cell_ranks.links(edge))
```

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,224 @@
...
+                                marked_for_update[rank].push_back(edge);
```
chrisrichardson, line 88:
```suggestion
                                marked_for_update[rank].push_back(cell);
```

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,224 @@
...
+  MPI_Comm neighbor_comm;
+  MPI_Dist_graph_create_adjacent(
+      mesh.comm(), ranks.size(), ranks.data(), MPI_UNWEIGHTED, ranks.size(),
+      ranks.data(), MPI_UNWEIGHTED, MPI_INFO_NULL, false, &neighbor_comm);
+
+  // Communicate ghost edges that might have been marked. This is not necessary
+  // for a uniform refinement.
+  if (cells.has_value())
+    update_logical_edgefunction(neighbor_comm, marked_for_update,
+                                refinement_marker, *map_e);
```
chrisrichardson, line 103:
```suggestion
                                refinement_marker, *map_c);
```

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,224 @@
...
+
+  // Construct the new vertices
+  const auto [new_vertex_map, new_vertex_coords, xshape]
+      = create_new_vertices(neighbor_comm, edge_ranks, mesh, refinement_marker);
```
chrisrichardson, line 107:
```suggestion
      = create_new_vertices(neighbor_comm, cell_ranks, mesh, refinement_marker);
```

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,224 @@
...
+  MPI_Comm_free(&neighbor_comm);
+
+  auto e_to_v = mesh.topology()->connectivity(1, 0);
```
chrisrichardson, line 110:
```suggestion
  auto c_to_v = mesh.topology()->connectivity(1, 0);
```

chrisrichardson:
I added a ton of suggested changes there - but basically: almost everywhere there is edge, it should be cell. I know it is the same thing, but it makes it clearer what we are doing.

python/test/unit/refinement/interval.py
```diff
@@ -0,0 +1,64 @@
...
+@pytest.mark.parametrize(
+    "ghost_mode", [mesh.GhostMode.none, mesh.GhostMode.shared_vertex, mesh.GhostMode.shared_facet]
+)
+@pytest.mark.parametrize(
+    "ghost_mode_refined",
+    [mesh.GhostMode.none, mesh.GhostMode.shared_vertex, mesh.GhostMode.shared_facet],
+)
+@pytest.mark.parametrize("redistribute", [True, False])
+def test_refine_interval_adaptive(ghost_mode, redistribute, ghost_mode_refined):
+    msh = mesh.create_interval(MPI.COMM_WORLD, 100, [0, 1], ghost_mode=ghost_mode)
```
jhale, line 48: Put the number of divisions as parameter of the test and run across a couple of values.

python/test/unit/refinement/interval.py
```diff
@@ -0,0 +1,64 @@
...
+    msh_refined, edges = mesh.refine_interval(
```
jhale, line 49: Could the current end-user refine routine be adjusted to dispatch to the right routine based on the topology of the mesh?

python/test/unit/refinement/interval.py
```diff
@@ -0,0 +1,64 @@
...
+        msh, np.arange(10), redistribute=redistribute, ghost_mode=ghost_mode_refined
```
jhale, line 50: I would also parameterise across a few different cell sets.

jhale, line 50: Local indices are usually typed as np.int32.

python/test/unit/refinement/interval.py
jhale: I guess this function is a bit richer with ghost mode being adjustable?

python/test/unit/refinement/interval.py
```diff
@@ -0,0 +1,64 @@
...
+# SPDX-License-Identifier:    LGPL-3.0-or-later
+
+from mpi4py import MPI
+
+import numpy as np
+import pytest
+
+from dolfinx import mesh
+
+
```
jhale, line 14: General comment: Add some tests checking the parent->child hierarchy is correct.

python/test/unit/refinement/interval.py
schnellerhase: Very possible, maybe do this in a follow up PR since it needs interface changes in plaza?

python/test/unit/refinement/interval.py
schnellerhase: In the cpp tests, where we have control over all related indices, they are present. Not to sure how to facilitate this in these highly parameterized setting.

schnellerhase:
In multiple runs of the pipelines we have observed the failure of one specific test in parallel, see: https://github.com/FEniCS/dolfinx/actions/runs/9973815307/job/27560013456?pr=3314 

Note: the test is also run in the non-petsc github pipeline, which passes without problems, so it must be related to a different environment, or build setup?

I can not reproduce this locally and the CI sometimes passes and sometimes fails. This needs to be investigated further.

schnellerhase:
Thanks to @chrisrichardson the bug is fixed. Was caused by an out of bound access of a vector in the test case, and we seem to have been highly unlucky to getting the correct value most of the time but not always...

jhale:
We should have one build setup with valgrind etc. that would pickup these common issues faster.

schnellerhase:
Yes sounds good - and maybe make `-fsanitize` then part of the build as well. On that note, I'd also be a big fan to see some form of static analysis, maybe `clang-tidy`, being run as part of the CI.

jhale:
We do have sonar cloud but it won't run on 'foreign' PRs.

I'd be open to experimenting with clang-tidy on another PR.

python/test/unit/refinement/interval.py
jhale: Yes I think that's best.

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,225 @@
...
+                         [&ranks](auto r)
+                         {
+                           auto it = std::lower_bound(ranks.begin(),
+                                                      ranks.end(), r);
+                           assert(it != ranks.end() and *it == r);
+                           return std::distance(ranks.begin(), it);
+                         });
+
+  // create refinement flag for cells
+  // TODO: vector of bools? -> make of use std specialization for type bool
```
jhale, line 73: This is best avoided - specialisation of vector for type bool is considered a mistake in the C++ standard.

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,225 @@
...
+  std::ranges::transform(cell_ranks.array(), cell_ranks.array().begin(),
+                         [&ranks](auto r)
+                         {
+                           auto it = std::lower_bound(ranks.begin(),
+                                                      ranks.end(), r);
+                           assert(it != ranks.end() and *it == r);
+                           return std::distance(ranks.begin(), it);
+                         });
+
+  // create refinement flag for cells
```
jhale, line 72: Start all comments with caps.

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,225 @@
...
+
+  graph::AdjacencyList cell_adj(std::move(cell_topology), std::move(offsets));
+
+  return {std::move(cell_adj), std::move(new_vertex_coords), xshape,
+          std::move(parent_cell)};
+}
+
+} // namespace impl
+
+/// Refines a (topologically) one dimensional mesh by splitting cells, i.e.
```
jhale, line 179: Don't understand the 'i.e. edges' comment.

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,225 @@
...
+/// edges.
+///
+/// @param[in] mesh Mesh to be refined
+/// @param[in] cells Optional indices of the cells that should be split by this
```
jhale, line 183: split -> refined

python/dolfinx/mesh.py
```diff
@@ -333,6 +334,37 @@ def refine(
...
+    mesh: Mesh,
+    cells: typing.Optional[np.ndarray] = None,
+    redistribute: bool = True,
+    ghost_mode: GhostMode = GhostMode.none,
+) -> tuple[Mesh, npt.NDArray[np.int32]]:
+    """Refine a (topologically) one dimensional mesh.
+
+    Args:
+        mesh: Mesh to refine
+        cells: Indices of cells, i.e. edges, to split druing refinement. If ``None``, mesh
```
jhale, line 347: Same comment on i.e. edges

python/dolfinx/mesh.py
```diff
@@ -333,6 +334,37 @@ def refine(
...
+            refinement is uniform.
+        redistribute: Refined mesh is re-partitioned if ``True``.
+        ghost_mode: ghost mode of the refined mesh
```
jhale, line 350: What is the use case for having a different ghost mode on the refined mesh?

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,225 @@
...
+/// across processes.
+/// @param[in] ghost_mode Ghost mode of the refined mesh, default is ghost mode
+/// none
+///
+/// @return Refined mesh, and list of parent cells - for every new cell index
+/// this contains the associated cell index of the pre-refinement mesh.
+template <std::floating_point T>
+std::tuple<mesh::Mesh<T>, std::vector<std::int32_t>>
+refine_interval(const mesh::Mesh<T>& mesh,
+                std::optional<std::span<const std::int32_t>> cells,
```
jhale, line 196: We should look at where we can use `std::optional` elsewhere.

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,225 @@
...
+/// @param[in] mesh Mesh to be refined
+/// @param[in] cells Optional indices of the cells that should be split by this
+/// refinement. If not provided, all cells are considered marked for refinement,
+/// i.e. a uniform refinement is performed.
+/// @param[in] redistribute Option to enable redistribution of the refined mesh
+/// across processes.
+/// @param[in] ghost_mode Ghost mode of the refined mesh, default is ghost mode
+/// none
+///
+/// @return Refined mesh, and list of parent cells - for every new cell index
```
jhale, line 191: and an array mapping the child cell index of the refined mesh to its parent cell index in the unrefined mesh.

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,225 @@
...
+
+  std::vector<std::int32_t> parent_cell;
+  parent_cell.reserve(refined_cell_count);
+
+  for (std::int32_t cell = 0; cell < map_c->size_local(); ++cell)
+  {
+    const auto& vertices = c_to_v->links(cell);
+    assert(vertices.size() == 2);
+
+    // we consider a (previous) cell, i.e. an edge of (global) vertices
```
jhale, line 140: confusing edge terminology

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,225 @@
...
+      cell_topology.insert(cell_topology.end(), {a, b});
+      parent_cell.push_back(cell);
+    }
+  }
+
+  assert(cell_topology.size() == refined_cell_count * 2);
+  assert(parent_cell.size() == refined_cell_count);
+
+  std::vector<std::int32_t> offsets(refined_cell_count + 1);
+  std::ranges::generate(offsets, [i = 0]() mutable { return 2 * i++; });
```
jhale, line 169: Looks complex. Wouldn't iota work ok?

schnellerhase: Unfortunately `iota` only allows for step size one.

cpp/dolfinx/refinement/interval.h
schnellerhase: One obvious one would be the plaza algorithm - this implementation was introduced after noticing the redundant implementation for both cases.

python/dolfinx/mesh.py
schnellerhase: Thats a good question. Would require that the refined mesh has different forms being evaluated on it, which do not require the same ghost mode (or the other way around) - not sure of a concrete example though. Maybe when coupling a surrogate model into an adaptively refined approach?

cpp/dolfinx/refinement/interval.h
jhale: Ah, it's always great to see the stuff that C++ has had to wait until C++23 to have...

cpp/dolfinx/refinement/interval.h
jhale: Understood. Let's look at the issue of a unified interface for refinement in another PR.

python/dolfinx/mesh.py
```diff
@@ -333,6 +334,37 @@ def refine(
     return Mesh(mesh1, mesh._ufl_domain)
 
 
+def refine_interval(
+    mesh: Mesh,
+    cells: typing.Optional[np.ndarray] = None,
+    redistribute: bool = True,
+    ghost_mode: GhostMode = GhostMode.none,
```
jhale, line 341: The default should be `none` - all our meshes are by default constructed with `shared_facet`.

cpp/dolfinx/refinement/interval.h
```diff
@@ -0,0 +1,225 @@
...
+///
+/// @return Refined mesh, and list of parent cells and an array mapping the
+/// child cell index of the refined mesh to its parent cell index in the
+/// unrefined mesh.
+template <std::floating_point T>
+std::tuple<mesh::Mesh<T>, std::vector<std::int32_t>>
+refine_interval(const mesh::Mesh<T>& mesh,
+                std::optional<std::span<const std::int32_t>> cells,
+                bool redistribute,
+                mesh::GhostMode ghost_mode = mesh::GhostMode::none)
```
jhale, line 198: Same comment, our default is `shared_facet` (currently the most general case).

python/dolfinx/mesh.py
schnellerhase: That's a typo? Will switch to `shared_facet`.

jhale: Indeed!

cpp/test/refinement/interval.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+  // complete refinement
+  {
+    auto [refined_mesh, parent_edges]
+        = refinement::refine_interval(mesh, std::nullopt, false);
+
+    // Check geometry
+    {
+
+      auto x = refined_mesh.geometry().x();
```
jhale, line 242: https://github.com/catchorg/Catch2/blob/devel/docs/comparing-floating-point-numbers.md

Never compare using equality of floating point types.

cpp/test/refinement/interval.cpp
```diff
@@ -0,0 +1,290 @@
...
+    // Check topology
+    {
+      auto topology = refined_mesh.topology_mutable();
+      CHECK(topology->dim() == 1);
+
+      topology->create_connectivity(0, 1);
+      auto v_to_e = topology->connectivity(0, 1);
+
+      // find the center index, i.e. the one with two outgoing edges
+      int center_index = v_to_e->num_links(0) == 2   ? 0
```
jhale, line 271: `std::int32_t`?

cpp/test/refinement/interval.cpp
```diff
@@ -0,0 +1,290 @@
...
+    CHECK(parent_edge[1] == 1);
+    CHECK(parent_edge[2] == 1);
+  }
+}
+
+TEST_CASE("Interval Refinement (parallel)", "refinement,interval,paralle")
+{
+  /**
+  Produces an interval with communicator size intervals. Every process is
+  assigned one intervall and we refine uniformly.
```
jhale, line 190: typo

cpp/test/refinement/interval.cpp
```diff
@@ -0,0 +1,290 @@
...
+        = [](MPI_Comm /* comm */, int /* nparts */,
+             const std::vector<mesh::CellType>& /* cell_types */,
+             const std::vector<std::span<const std::int64_t>>& /* cells */)
+        -> graph::AdjacencyList<std::int32_t>
+    {
+      return graph::AdjacencyList<std::int32_t>(
+          dolfinx::MPI::size(MPI_COMM_WORLD));
+    };
+
+    auto commt = rank == 0 ? MPI_COMM_SELF : MPI_COMM_NULL;
```
jhale, line 226: Avoid auto for simple types.

cpp/test/refinement/interval.cpp
```diff
@@ -0,0 +1,290 @@
...
+
+      std::ranges::sort(x);
+
+      CHECK(x[0] == static_cast<double>(rank) / comm_size);
+      CHECK(x[1]
+            == static_cast<double>(rank) / comm_size + 1. / (2 * comm_size));
+      CHECK(x[2]
+            == static_cast<double>(rank) / comm_size + 2. / (2 * comm_size));
+
+      CHECK(x[3] == rank + 1);
```
jhale, line 253: `rank` `comm_size` etc. not consistently cast to doubles - perhaps do it upfront to improve readability? Also mixed use of floating point literals and integer literals.

cpp/test/refinement/interval.cpp
```diff
@@ -0,0 +1,290 @@
...
+    std::vector<std::int64_t> cells;
+    fem::CoordinateElement<double> element(mesh::CellType::interval, 1);
+    if (rank == 0)
+    {
+      for (std::int64_t i = 0; i < comm_size + 1; i++)
+        x.insert(x.end(), {static_cast<double>(i) / comm_size,
+                           static_cast<double>(i) + 1, 2. * i + comm_size});
+      for (std::int64_t i = 0; i < 2 * comm_size; i++)
+      {
+        auto div = std::div(i, static_cast<std::int64_t>(2));
```
jhale, line 211: avoid auto for simple types.

schnellerhase: that would be a `ldiv_t` that really doesn't seem like a simple type. Maybe switch to `auto [quot, rem] = ...` instead?

cpp/test/refinement/interval.cpp
schnellerhase: Changed to `size_t` as `num_links` expect that.

PR #3315, opened 2024-07-17
Replace `std::unique` with `std::ranges::unique`

schnellerhase:


jhale:
Approved, direct translation of existing classic STL iterator algorithm to range.

PR #3317, opened 2024-07-17
nls ksp/pc choice

xisalz:
Hello
If the original intention was to force the NewtonSolver ksp/pc object to be set as a direct solver and only give the user the ability to tune the LU solver options with command line arguments, this PR is not going in the right direction. 

Anyway user, after NewtonSolver instantiation, can grab ksp pointer and reset the way he want  ksp/pc by either using KSPSetType, PCSetType, PetscOptionsSetValue and/or KSPSetFromOptions. For a newbie like me, it takes some time to figure out why adding :
`-nls_solve_ksp_type cg -nls_solve_pc_type hypre -nls_solve_pc_hypre_type boomeramg -nls_solve_pc_hypre_boomeramg_max_levels 15 `
options to my program did not work, and how to change my code so that this setting is taken into account.

This PR replace the hard-coded option setting by a call to KSPSetType, PCSetType so that KSPSetFromOptions woks. Now to avoid "polluting"  NewtonSolver class by direct use of KSPSetType, PCSetType,... some extra warper for these functions could have been implemented...but I prefer to add a function to KrylovSolver class ...

Regards

P.S. Won't be available before the 19th of August for any exchange ... if any ...

### DIFF
--- a/cpp/dolfinx/la/petsc.cpp
+++ b/cpp/dolfinx/la/petsc.cpp
@@ -674,6 +674,23 @@ petsc::KrylovSolver& petsc::KrylovSolver::operator=(KrylovSolver&& solver)
   return *this;
 }
 //-----------------------------------------------------------------------------
+void petsc::KrylovSolver::set_as_direct_solver() 
+{ 
+  assert(_ksp);
+  // remove iterative resolution: only use preconditioner pass
+  PetscErrorCode ierr = KSPSetType(_ksp, KSPPREONLY);
+  CHECK_ERROR("KSPSetType");
+
+  PC pc;
+  ierr = KSPGetPC(_ksp, &pc);
+  CHECK_ERROR("KSPGetPC");
+
+  // set preconditioner as a direct solver using LU factorization
+  ierr = PCSetType(pc, PCLU);
+  CHECK_ERROR("PCSetType");
+   
+}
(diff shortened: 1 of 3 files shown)


nate-sime:
I'm not sure what the motivation is here. The user can [access](https://github.com/FEniCS/dolfinx/blob/830a9e9c0e48815919dcfc3f7e6384feb1b93fae/cpp/dolfinx/nls/NewtonSolver.h#L81-L87) the Krylov solver to customise as they see fit, and further call [KSPSetFromOptions](https://petsc.org/main/manualpages/KSP/KSPSetFromOptions/) at will.

It's usually not a good idea to hard-code solver options as they're problem dependent.

jhale:
We've had a painful time dealing with overriding PETSc's defaults which is why we've made the decision to not do so. If there is an issue with clarity then it should be improved by improving the documentation - which does look poor at the moment.

https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/nls/petsc.py#L29
https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/nls/NewtonSolver.h

For example, our LinearProblem class has an example of how to set the underlying options:

https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/fem/petsc.py#L747

Issue #3312, opened 2024-07-15
Enable macOS CI on PRs (PETSc-less)


garth-wells:
Now that PETSc is optional, we can reasonably run the macOS CI on pull requests (without PETSc).

PR #3318, opened 2024-07-18
Mac os CI pipeline

schnellerhase:
Adds a new mac OS CI pipeline, that should be fast enough to run with every commit/PR. 

- The CI run both a `petsc`-free and a build with `petsc` using the `brew` package
- Currently we can not verify the python package in the `petsc` build, as I have no idea how to make `petsc4py` work properly with the brew package (Note: same holds true for the debian packages and thus restricts the running with debian petsc install)

- [x] There are two failing python tests in the `petsc` free run, not sure what's going on there, and whats causing this. Needs to be fixed before merge.

Fixes https://github.com/FEniCS/dolfinx/issues/3312


jhale:
Looks good. I think this could be an adjustment of the existing macOS pipeline?

schnellerhase:
I couldn't figure out how to maintain both the scheduling of the other macos pipeline and fit the new jobs into the same file, if we can make that work, we can move them into the same file.

schnellerhase:
The failing tests are fixed now. Not sure why this is not necessary on other platforms, but registering the complex types fixes it, not sure if that is done in the right spot of the test though.

python/dolfinx/utils.py
```diff
@@ -238,5 +245,7 @@ def set_vals(A: int,
         """See PETSc `MatSetValuesBlockedLocal
         <https://petsc.org/release/manualpages/Mat/MatSetValuesBlockedLocal>`_
         documentation."""
-    except (ImportError, KeyError):
-        pass
+    except ImportError:
```
jhale, line 248: We have our own logging library for these warnings.

garth-wells:
Rather than using Homebrew, how about using conda to install dependencies? The conda PETSc package has more options enabled, e.g. PETSc is configure with parallel LU solver. And petsc4py is available.

See #3319.

jhale:
I agree with @garth-wells , homebrew's petsc is not really useful for our users so it seems a waste to test against it. Although good to see it's now being built against HDF5 MPI!

I would propose:
- conda petsc (on commits)
- no petsc (on commits)
- homebrew + petsc build (weekly)

schnellerhase:
I can not get the basix build to pass, pretty sure its something to do with the availability of a C++ standard library, however I'm not really sure how to make that available in the conda setting. See [herer](https://github.com/schnellerhase/basix/pull/3) for the basix standalone failure.

schnellerhase:
The error I get is
```
  Undefined symbols for architecture arm64:
    "std::exception::what() const", referenced from:
```

Never worked on ARM before - do you guys know what else I need to change to make that work?

schnellerhase:
Okay I guess we have no real fix for this problem for now, so should we return to the case of the brew install and get this merged for now to have a macos pipeline running at all?

Also we could in continuation of https://github.com/FEniCS/dolfinx/pull/3335 make use of the sup optimal setting of being able to make petsc run and not petsc4py by compiling a the cpp library with petsc support and the python one without support of petsc4py and then run the tests. This is so far not tested in any CI.

schnellerhase:
The `cffi` fixes here we should maybe extract and merge independent of the macos pipeline. However just noted they are broken, got lost in some merge conflict.

jhale:
> The `cffi` fixes here we should maybe extract and merge independent of the macos pipeline. However just noted they are broken, got lost in some merge conflict.

They are at acd43c75e30246dc464df77b1bdc0bff5faecd47 - could you make a new PR with just this?

PR #3319, opened 2024-07-21
Restore oneAPI CI

garth-wells:
Due to broken Intel Python changes, run oneAPI CI now in a conda environment
- Custom Docker image no longer required
- Adds some missing includes


jhale:
Looks good.

My only comment is that the GitHub images are a bit 'kitchen sink' and we do not have control over them. See the recent issue with Windows where we could not roll back the broken MSVC version (and still cannot). So it can still be a good idea to use an image, even if it is a minimal one that we do not build ourselves.

garth-wells:
> My only comment is that the GitHub images are a bit 'kitchen sink' and we do not have control over them. See the recent issue with Windows where we could not roll back the broken MSVC version (and still cannot). So it can still be a good idea to use an image, even if it is a minimal one that we do not build ourselves.

Indeed, this is why is runs within a `ubuntu:24.04` Docker container, see https://github.com/FEniCS/dolfinx/blob/0c539488143658de83fb0f1e899ab8f02049c993/.github/workflows/oneapi.yml#L20. Moreover, this way we can be sure that there is gcc, etc, available.

jhale:
> > My only comment is that the GitHub images are a bit 'kitchen sink' and we do not have control over them. See the recent issue with Windows where we could not roll back the broken MSVC version (and still cannot). So it can still be a good idea to use an image, even if it is a minimal one that we do not build ourselves.
> 
> Indeed, this is why is runs within a `ubuntu:24.04` Docker container, see
> 
> https://github.com/FEniCS/dolfinx/blob/0c539488143658de83fb0f1e899ab8f02049c993/.github/workflows/oneapi.yml#L20
> 
> . Moreover, this way we can be sure that there is gcc, etc, available.

Perfect I missed that on my review. Thanks!

PR #3321, opened 2024-07-23
Try running windows action with defender disabled

jhale:
Rumours around GitHub suggest this might be why Windows is slow...


jhale:
Rumours unfounded in evidence.

PR #3322, opened 2024-07-24
Unify refine interface 

schnellerhase:
This PR contains several changes to the refinement routines to facilitate a neater, simpler and combined API to the user.

1. The interface of the `plaza` routines now uses `std::optional`'s as input arguments and thus allows for a single code path. (Previously the cases of uniform and adaptive refinement were treated independently)
2. The plaza refinement options was centralized to the (more general) `refinement` namespace and now is also used for configuring the interval refinement.
3. The return values of parent facets and cells are now `std::optional` as well, to no longer need to rely on an ambiguous empty array check. 
4. The python export and interface with the new optional arguments is greatly simplified, no longer do we need to export multiple versions of the refinement routines, and no more dispatching in the python interface is necessary. 
5. Refinement testing has been extended and simplified in some cases.

### DIFF
--- a/cpp/dolfinx/refinement/CMakeLists.txt
+++ b/cpp/dolfinx/refinement/CMakeLists.txt
@@ -4,6 +4,7 @@ set(HEADERS_refinement
     ${CMAKE_CURRENT_SOURCE_DIR}/plaza.h
     ${CMAKE_CURRENT_SOURCE_DIR}/refine.h
     ${CMAKE_CURRENT_SOURCE_DIR}/utils.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/option.h
     PARENT_SCOPE
 )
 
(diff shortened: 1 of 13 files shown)


jhale:
"Should probably return a shared_ptr to the new mesh, for consistent interface usage with the external API" - This isn't necessary, as the caller takes ownership.

Could you try `std::optional` on the return types from `refine`? This would avoid using empty arrays as sentinels for "not computed".

schnellerhase:
Optional return value seems to work as well, in python this just becomes a `None` value if nothing is set :) - will change the facets as well to this behavior.

schnellerhase:
Requires prior merge of https://github.com/FEniCS/dolfinx/pull/3331

garth-wells:
@schnellerhase could you resolve the merge conflict?

PR #3328, opened 2024-07-31
`Timer` interface improvement

schnellerhase:
Timer interface simplified in both cpp and python part, making use of `std::optional`.


schnellerhase:
If the optional handling is the path of choice, reconsider https://github.com/FEniCS/dolfinx/pull/3283

cpp/dolfinx/common/Timer.cpp
```diff
@@ -4,23 +4,23 @@
 //
 // SPDX-License-Identifier:    LGPL-3.0-or-later
 
-#include "Timer.h"
+#include <optional>
+#include <stdexcept>
+
```
garth-wells, line 9: Remove space.

cpp/dolfinx/common/Timer.cpp
```diff
@@ -4,23 +4,23 @@
...
 #include "TimeLogger.h"
-#include <stdexcept>
+#include "Timer.h"
 
 using namespace dolfinx;
 using namespace dolfinx::common;
 
 //-----------------------------------------------------------------------------
-Timer::Timer() : Timer::Timer("")
+Timer::Timer(std::optional<std::string> task) : _task("")
```
garth-wells, line 18: Is `_task("")` required?

cpp/dolfinx/common/Timer.cpp
```diff
@@ -4,23 +4,23 @@
...
 {
-  // Do nothing
-}
-//-----------------------------------------------------------------------------
-Timer::Timer(const std::string& task) : _task(task)
-{
-  // Do nothing
+  if (task)
+  {
```
garth-wells, line 21: Remove braces (house style).

cpp/dolfinx/common/Timer.cpp
```diff
@@ -4,23 +4,23 @@
...
+    _task = std::move(task.value());
```
garth-wells, line 22: No need to move here - simple copy-assign is fine.

Should `_task` have type `std::optional<std::string> task`? Might be neater.

cpp/dolfinx/common/Timer.h
```diff
@@ -30,11 +31,9 @@ namespace dolfinx::common
 class Timer
 {
 public:
-  /// Create timer without logging
-  Timer();
-
-  /// Create timer with logging
-  Timer(const std::string& task);
+  /// Create timer
+  /// if a task name is provided this enables logging, otherwise without logging
```
garth-wells, line 35: Add line break (this affects Doxygen formatting). Capitalise and punctuate. Sentence could be made clearer.

garth-wells (changes requested):

schnellerhase:
This means that the `optional` interface is the path of choice?

schnellerhase:
List of further optionals to change:

- [ ] `set_bc`: make $x_0$ optional
- [x] `refine`
- [ ] (not strictly optional) `locate_dofs_topological`

cpp/dolfinx/common/Timer.cpp
```diff
@@ -7,21 +7,14 @@
...
-Timer::Timer() : Timer::Timer("")
-{
-  // Do nothing
-}
-//-----------------------------------------------------------------------------
-Timer::Timer(const std::string& task) : _task(task)
-{
-  // Do nothing
-}
+Timer::Timer(std::optional<std::string> task) : _task(std::move(task)) {}
```
garth-wells, line 17: Just make a copy rather than move - no performance need for the extra complexity.

python/dolfinx/wrappers/common.cpp
```diff
@@ -152,8 +153,7 @@ void common(nb::module_& m)
           nb::arg("global"));
   // dolfinx::common::Timer
   nb::class_<dolfinx::common::Timer>(m, "Timer", "Timer class")
-      .def(nb::init<>())
-      .def(nb::init<std::string>(), nb::arg("task"))
+      .def(nb::init<std::string>(), nb::arg("task") = nb::none())
```
garth-wells: Let's remove the default arg in the wrapper layer. Since we wrap the interface in the Python layer, best to handle the default arg there. Defaults in too many places can lead to bugs.

garth-wells (changes requested):
Looking good, almost there.

garth-wells:
> This means that the `optional` interface is the path of choice?

Yes, when it makes logical sense, i.e. when `none` is appropriate. We shouldn't use it when a default value makes sense, e.g. a solver tolerance.

garth-wells (approved):

schnellerhase:
I'll file an issue for the remaining cases, then we can merge this one already.

python/dolfinx/wrappers/common.cpp
schnellerhase: This does not work, faced the same problem within #3322 - I'm not sure what causes this or how it may be fixed, but somehow only when the `nb::none()` is provided it properly picks up on the python `None` to cpp `std::nullopt` conversion.

garth-wells: Take a look at https://nanobind.readthedocs.io/en/latest/functions.html#none-arguments.

python/dolfinx/wrappers/common.cpp
```diff
@@ -152,8 +153,7 @@ void common(nb::module_& m)
           nb::arg("global"));
   // dolfinx::common::Timer
   nb::class_<dolfinx::common::Timer>(m, "Timer", "Timer class")
-      .def(nb::init<>())
-      .def(nb::init<std::string>(), nb::arg("task"))
+      .def(nb::init<std::optional<std::string>>(), nb::arg("task") = nb::none())
```
garth-wells, line 156:
```suggestion
      .def(nb::init<std::optional<std::string>>(), nb::arg("task").none())
```

PR #3329, opened 2024-08-01
Clarify memory management of returned petsc4py objects.

jhale:
Explicit instructions for handling petsc4py objects memory management.


jorgensd:
Some context: https://gitlab.com/petsc/petsc/-/issues/1309

jhale:
These additions are correct at least according to my understanding of the new petsc4py memory management model.

python/dolfinx/fem/petsc.py
```diff
@@ -111,6 +111,10 @@ def create_vector(L: Form) -> PETSc.Vec:
 def create_vector_block(L: list[Form]) -> PETSc.Vec:
     """Create a PETSc vector (blocked) that is compatible with a list of linear forms.
 
+    Note:
```
garth-wells, line 114: Suggested text:
```
Note:
    Due to subtle issues in the interaction between petsc4py memory management 
    and the Python garbage collector, it is recommended that the method ``PETSc.Vec.destroy()`` 
    is called on the returned object once the object is no longer required. Note that 
    ``PETSc.Vec.destroy()`` is collective over the object's MPI communicator.  
```

garth-wells (changes requested):
I've added suggest text that is technically accurate.

garth-wells (approved):

PR #3336, opened 2024-08-06
Ensure features are found/not found in our comprehensive builds

jhale:
I noticed that KAHIP wasn't being found. This PR enables explicit dependency checks on our kitchen sink build to catch silent mistakes. KAHIP issue still needs fixing.

Also includes #3335.


garth-wells:
This PR mixes some unrelated changes - can it be simplified to just include the CMake CI changes?

jhale:
Yes - I will tidy and merge #3335 first and then rebase this on top.

PR #3338, opened 2024-08-07
Sparsity pattern insert

schnellerhase:
Introduces a new sparsity pattern insert functionality for a single entry - which is the method of choice used in multiple locations of the code base.

Additionally the documentation of the `insert(span<>, span<>)` call on a  sparsity pattern  is extended to make clear how the inputs work together, and now hints that the interpretation of a pairwise operation is not correct to the caller.

### DIFF
--- a/cpp/dolfinx/la/SparsityPattern.cpp
+++ b/cpp/dolfinx/la/SparsityPattern.cpp
@@ -143,16 +143,20 @@ SparsityPattern::SparsityPattern(
 void SparsityPattern::insert(std::int32_t row, std::int32_t col)
 {
   if (!_offsets.empty())
+  {
     throw std::runtime_error(
         "Cannot insert into sparsity pattern. It has already been finalized");
+  }
 
   assert(_index_maps[0]);
   const std::int32_t max_row
       = _index_maps[0]->size_local() + _index_maps[0]->num_ghosts() - 1;
 
   if (row > max_row or row < 0)
+  {
     throw std::runtime_error(
         "Cannot insert rows that do not exist in the IndexMap.");
+  }
 
   _row_cache[row].push_back(col);
 }
@@ -161,8 +165,10 @@ void SparsityPattern::insert(std::span<const std::int32_t> rows,
                              std::span<const std::int32_t> cols)
 {
   if (!_offsets.empty())
+  {
     throw std::runtime_error(
         "Cannot insert into sparsity pattern. It has already been finalized");
+  }
 
   assert(_index_maps[0]);
   const std::int32_t max_row
@@ -181,8 +187,10 @@ void SparsityPattern::insert(std::span<const std::int32_t> rows,
 void SparsityPattern::insert_diagonal(std::span<const std::int32_t> rows)
 {
   if (!_offsets.empty())
+  {
     throw std::runtime_error(
         "Cannot insert into sparsity pattern. It has already been finalized");
+  }
 
   assert(_index_maps[0]);
   const std::int32_t max_row
--- a/cpp/dolfinx/la/SparsityPattern.cpp
+++ b/cpp/dolfinx/la/SparsityPattern.cpp
@@ -140,6 +140,23 @@ SparsityPattern::SparsityPattern(
   }
 }
 //-----------------------------------------------------------------------------
+void SparsityPattern::insert(std::int32_t row, std::int32_t col)
+{
+  if (!_offsets.empty())
+    throw std::runtime_error(
+        "Cannot insert into sparsity pattern. It has already been finalized");
+
+  assert(_index_maps[0]);
+  const std::int32_t max_row
+      = _index_maps[0]->size_local() + _index_maps[0]->num_ghosts() - 1;
+
+  if (row > max_row or row < 0)
+    throw std::runtime_error(
+        "Cannot insert rows that do not exist in the IndexMap.");
+
+  _row_cache[row].push_back(col);
+}
+//-----------------------------------------------------------------------------
 void SparsityPattern::insert(std::span<const std::int32_t> rows,
                              std::span<const std::int32_t> cols)
 {
--- a/cpp/dolfinx/la/SparsityPattern.cpp
+++ b/cpp/dolfinx/la/SparsityPattern.cpp
@@ -177,9 +177,10 @@ void SparsityPattern::insert(std::span<const std::int32_t> rows,
   for (std::int32_t row : rows)
   {
     if (row > max_row or row < 0)
+    {
       throw std::runtime_error(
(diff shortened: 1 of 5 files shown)


cpp/dolfinx/la/SparsityPattern.h
```diff
@@ -67,6 +67,19 @@ class SparsityPattern
...
+  /// indices.
+  ///
+  /// This inserts for every row in rows the columns (listed in cols) to the
+  /// sparsity pattern. This allows for easy setting of dense sub-blocks, for
+  /// example sp.insert({0, 1, 2}, {1, 2, 3}) creates a 3x3 block in the
+  /// sparsity pattern with top left corner at (0, 1).
+  ///
+  /// Note: Especially this operation does not perform a pairwise insertion to
+  /// the sparsity pattern, i.e. the above call does not introduce the entries
+  /// at locations (0, 1), (1, 2) and (2, 3)!
```
IgorBaratta, line 82: This is a bit confusing to me, because it does create these entries. Maybe you could say that rows and columns could have different sizes, so the 1-to-1 map could not be implied.

schnellerhase: How about: "..., i.e. the above call creates more then (9 in total) entries and not only entries at the locations (0,1), (1,2) and (2,3)!"

cpp/dolfinx/la/SparsityPattern.cpp
```diff
@@ -140,14 +140,29 @@ SparsityPattern::SparsityPattern(
   }
 }
 //-----------------------------------------------------------------------------
+void SparsityPattern::insert(std::int32_t row, std::int32_t col)
```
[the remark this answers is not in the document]
schnellerhase: It it, but the interface of the caller becomes cleaner. Instead of needing to call 
```
sp.insert(std::vector<std::int32_t>(row), std::vector<std::int32_t>(col));
```
 we can now simply call
```
sp.insert(row, col);
```
same holds for the python interface with `numpy` arrays.

Also, in combination with the span based version it makes clearer how the sparsity pattern is to be used, i.e. loop over entries and dense matrices and do not pre-compute the entries and then pass it in bulk. Sparsity pattern already performs the caching and associated memory management.

cpp/dolfinx/la/SparsityPattern.h
schnellerhase: Added a combination of both. I think we want to say $A[row[i], col[j]]$ is being set though.

cpp/dolfinx/la/SparsityPattern.cpp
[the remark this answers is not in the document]
jhale: Well, it's clear from the improved docstring. I don't see the point of this new method, the clarity of overloading functions is debatable.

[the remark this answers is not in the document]
jhale: To add, the interface is vectorised from the Python end to encourage users to pass in multiple entries, avoiding the overhead of the function call.

cpp/dolfinx/la/SparsityPattern.h
```diff
@@ -67,6 +67,22 @@ class SparsityPattern
...
+  ///
+  /// This routine inserts non-zero locations at the outer product of rows and
+  /// cols to the sparsity pattern, i.e. adds the matrix entries at
+  ///   A[row[i], col[j]] for all i, j
+  /// Enabling easy setting of dense sub-blocks, for example
+  ///   sp.insert({0, 1, 2}, {1, 2, 3})
+  /// creates a 3x3 block in the sparsity pattern with top left corner at (0,
+  /// 1).
+  ///
+  /// Note: Especially this operation does not perform a pairwise insertion to
```
jhale, line 83: Remove, no need to repeat yourself.

cpp/dolfinx/la/SparsityPattern.h
```diff
@@ -67,6 +67,22 @@ class SparsityPattern
...
   /// indices.
+  void insert(std::int32_t row, std::int32_t col);
+
+  /// @brief Insert non-zero locations using local (process-wise)
+  /// indices.
+  ///
+  /// This routine inserts non-zero locations at the outer product of rows and
+  /// cols to the sparsity pattern, i.e. adds the matrix entries at
+  ///   A[row[i], col[j]] for all i, j
+  /// Enabling easy setting of dense sub-blocks, for example
```
jhale, line 78: Not sure about the word dense. What if I pass e.g. `1, 3, 4` `1, 3, 4`.

cpp/dolfinx/la/SparsityPattern.h
```diff
@@ -67,6 +67,22 @@ class SparsityPattern
 
   /// @brief Insert non-zero locations using local (process-wise)
   /// indices.
+  void insert(std::int32_t row, std::int32_t col);
+
+  /// @brief Insert non-zero locations using local (process-wise)
+  /// indices.
+  ///
+  /// This routine inserts non-zero locations at the outer product of rows and
+  /// cols to the sparsity pattern, i.e. adds the matrix entries at
```
jhale, line 76: to -> into

cpp/dolfinx/la/SparsityPattern.h
schnellerhase: The introduced entries then form a dense 3x3 entries again, not locally dense in the sparse matrix, but a dense matrix worth of entries - not possible to only specify a sub set of these with this function.

cpp/dolfinx/la/SparsityPattern.cpp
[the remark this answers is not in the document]
schnellerhase: How/where is the interface vectorized? I can not find that.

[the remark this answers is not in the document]
schnellerhase: @chrisrichardson mentioned the other day, we might instead of the integer version consider to create a non owning `const std::vector<std::int32_t>&` version, this should allow for the interface to be `insert({row}, {col})`, i.e. it allows for conversion from an initializer list.

schnellerhase:
Same interface present for `MatrixCSR.set`.

schnellerhase:
A related problem of range based interface is present in the global to local mappings of the `IndexMap`, converting a single value is not straight forward, see [here](https://github.com/FEniCS/dolfinx/pull/3354/files#diff-20fded61aaf999f0eb5245b32d0fb7597464c8437101d413f64b3650d1029da7R235-R240)

jpdean (approved):
Wrapping single values in `std::vector` doesn't bother me too much, but I like the improvements to the docs; it's now much clearer what the function is doing!

PR #3339, opened 2024-08-07
Fix ubuntu CI

schnellerhase:
Runs a prior `apt-get update` call before running the installs. This is also the recommended setup by github https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/customizing-github-hosted-runners.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -85,6 +85,7 @@ jobs:
     steps:
       - name: Install dependencies
         run: |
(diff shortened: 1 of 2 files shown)


schnellerhase:
Rocky pipeline also broken? Can we retry with a fresh build of the docker image?

jhale:
Rocky build triggered.

schnellerhase:
No luck, problem persists.

Issue #3332, opened 2024-08-05
[BUG]: gmshio causes kernel crash if entities are tagged several times


pierricmora:
### Summarize the issue

It seems to me that gmshio requires that all facets are assigned into a single physical group. When facets get assigned several tags, I get a kernel crash in a jupyter lab.

I tracked the line causing kernel crash until [create_mesh](https://github.com/FEniCS/dolfinx/blob/8e4122b5b28a8997b94148797125becf5ba704cb/python/dolfinx/io/gmshio.py#L294) in `model_to_mesh`.

### How to reproduce the bug

The following 2D example represents an ellipse inclusion inside a rectangular matrix. Another smaller rectangle represents a region where I would like to assign a source term, without being linked to different physical properties. The source region overlaps the inclusion; therefore the natural way to define tags is:

- (run example with mode=1) 1 for the matrix (in & out the source), 2 for the inclusion (in & out the source), and 3 for the source (matrix & inclusion). -> kernel crash

As for now, I can bypass the bug by working with more cumbersome tags:

- (run example with mode=2) 10 for the matrix inside the source, 11 for the matrix outside the source, 20 and 21 for the inclusion on & out the source.

### Minimal Example (Python)

```Python
from mpi4py import MPI
from dolfinx.io import gmshio
import gmsh

# inspired from https://github.com/FEniCS/dolfinx/blob/5d15d263f20a9d8ada051dd2ece0f35894d38d4a/python/demo/demo_gmsh.py
def gmsh_rect_ellipse(mode: int=1, show: bool=False) -> gmsh.model:
    """Create a Gmsh model of an ellipse inclusion inside a rectangle,
    with another smaller rectangle representing a source term

    Args:
        show: if True: pop the Gmsh viewer

    Returns:
        Gmsh model
    """
    model = gmsh.model()

    name = "rectangle-ellipse"
    model.add(name)
    model.setCurrent(name)

    # Phase 1; "the matrix"
    x, y, dx, dy = 0, 0, 2, 1
    rec1 = gmsh.model.occ.addRectangle(x, y, 0, dx, dy)

    # Phase 2; "the inclusion"
    x, y, big_radius, small_radius = 1, 0.5, 0.5, 0.25
    ell1 = gmsh.model.occ.addEllipse(x, y, 0, big_radius, small_radius)
    cl_ell1 = gmsh.model.occ.addCurveLoop([ell1])
[...]

PR #3342, opened 2024-08-08
add checks of cell tag consistency in gmshio

pierricmora:
Relates to issue #3332 

Add checks to the gmshio interface to ensure that all cells are tagged once, to avoid undesired behaviour or crashes.

Compared to the discussion in issue #3332, the impact on the code is slightly more as not only `model_to_mesh` but also `extract_topology_and_markers` have been modified. This is because storing `entity_tags` in `extract_topology_and_markers` enables to call the relatively expensive `np.unique` on a smaller array. Then, the overcost due to these checks is below 5%, compared to 15% - 10% reported before.

### DIFF
--- a/python/dolfinx/io/gmshio.py
+++ b/python/dolfinx/io/gmshio.py
@@ -152,6 +152,9 @@ def extract_topology_and_markers(model, name: typing.Optional[str] = None):
             # Create marker array of length of number of tagged cells
             marker = np.full_like(entity_tags[0], tag)
 
+            # Keep track of entity tags to check tag consistency
+            entity_tags = entity_tags[0]
+
             # Group element topology and markers of the same entity type
             entity_type = entity_types[0]
             if entity_type in topologies.keys():
@@ -161,8 +164,14 @@ def extract_topology_and_markers(model, name: typing.Optional[str] = None):
                 topologies[entity_type]["cell_data"] = np.hstack(
                     [topologies[entity_type]["cell_data"], marker]
                 )
+                topologies[entity_type]["entity_tags"] = np.hstack(
+                    [topologies[entity_type]["entity_tags"], entity_tags]
+                )
             else:
-                topologies[entity_type] = {"topology": topology, "cell_data": marker}
+                topologies[entity_type] = {"topology": topology,
+                                           "cell_data": marker,
+                                           "entity_tags": entity_tags,
+                                          }
 
     return topologies
 
@@ -256,6 +265,20 @@ def model_to_mesh(
 
         # Sort elements by ascending dimension
         perm_sort = np.argsort(cell_dimensions)
+
+        # Check that all cells are tagged once
+        _d = model.getDimension()
+        assert _d in topologies.keys(), 'All cells are expected to be tagged once; none found'
+        _elementTypes, _elementTags, _nodeTags = model.mesh.getElements(dim=_d, tag=-1)
+        # assert only one type of elements
+        # assert len(_elementTypes) == 1  # NOTE: already checked in extract_topology_and_markers
+        nbcells = len(_elementTags[0])
+        nbcells_tagged = len(topologies[_d]["entity_tags"])
+        assert nbcells == nbcells_tagged, \
+            f'All cells are expected to be tagged once; found: {nbcells_tagged}, expected: {nbcells}'
+        nbcells_tagged_once = len(np.unique(topologies[_d]['entity_tags']))
+        assert nbcells_tagged == nbcells_tagged_once, \
+            'All cells are expected to be tagged once; found duplicates'
 
         # Broadcast cell type data and geometric dimension
         cell_id = cell_information[perm_sort[-1]]["id"]


python/dolfinx/io/gmshio.py
```diff
@@ -257,6 +266,20 @@ def model_to_mesh(
         # Sort elements by ascending dimension
         perm_sort = np.argsort(cell_dimensions)
 
+        # Check that all cells are tagged once
+        _d = model.getDimension()
+        assert _d in topologies.keys(), 'All cells are expected to be tagged once; none found'
```
jhale, line 271: As these are user-facing errors they should raise Exceptions of the appropriate type directly, probably RuntimeError, rather than using assert.

pierricmora: Done.

Suggestion: could the first RuntimeError (no cell is tagged) be converted into a RuntimeWarning by defaulting a uniform tag for all cells?

jorgensd: Currently most warnings get swallowed by FFCx (https://github.com/FEniCS/ffcx/issues/712) so I am not sure this is a good idea

pierricmora:
I corrected two formatting errors but I still get Lint failed. This time the error is not very verbose... Any idea? When I do ruff check and ruff format in my computer I get no error.

python/dolfinx/io/gmshio.py
```diff
@@ -257,6 +267,27 @@ def model_to_mesh(
         # Sort elements by ascending dimension
         perm_sort = np.argsort(cell_dimensions)
 
+        # Check that all cells are tagged once
+        _d = model.getDimension()
```
pierricmora, line 271: Could be used to remove `gdim` from parameters. What about this simplification? @jorgensd 

```python
if comm.rank == rank:
    gdim = model.getDimension()
    gdim = comm.bcast(gdim, root=rank)

# ...

else:
    gdim = comm.bcast(None, root=rank)
```

jorgensd: Not sure that is safe, as you might want: A flat manifold in some plane, i.e.
```bash

In [1]: import gmsh

In [2]: gmsh.initialize()

In [3]: idx = gmsh.model.occ.add_rectangle(0,0,0.1,1,1)

In [4]: gmsh.model.occ.synchronize()

In [5]: gmsh.model.add_physical_group(2,[idx])
Out[5]: 1

In [6]: gmsh.model.mesh.generate(2)
Info    : Meshing 1D...
Info    : [  0%] Meshing curve 1 (Line)
Info    : [ 30%] Meshing curve 2 (Line)
Info    : [ 60%] Meshing curve 3 (Line)
Info    : [ 80%] Meshing curve 4 (Line)
Info    : Done meshing 1D (Wall 0.000247945s, CPU 0.000433s)
Info    : Meshing 2D...
Info    : Meshing surface 1 (Plane, Frontal-Delaunay)
Info    : Done meshing 2D (Wall 0.00361243s, CPU 0.003612s)
Info    : 98 nodes 198 elements

In [7]: gmsh.model.getDimension()
Out[7]: 2

In [8]: gmsh.model.mesh.generate(3)
Info    : Meshing 3D...
Info    : Done meshing 3D (Wall 2.2143e-05s, CPU 2.1e-05s)
Info    : 98 nodes 198 elements

In [9]: gmsh.model.getDimension()
Out[9]: 2
```

ref:
```bash
In [10]: gmsh.model.occ.add_rectangle?
Signature: gmsh.model.occ.add_rectangle(x, y, z, dx, dy, tag=-1, roundedRadius=0.0)
Docstring:
gmsh.model.occ.addRectangle(x, y, z, dx, dy, tag=-1, roundedRadius=0.)

Add a rectangle in the OpenCASCADE CAD representation, with lower left
corner at (`x', `y', `z') and upper right corner at (`x' + `dx', `y' +
`dy', `z'). If `tag' is positive, set the tag explicitly; otherwise a new
tag is selected automatically. Round the corners if `roundedRadius' is
nonzero. Return the tag of the rectangle.

Return an integer.

Types:
- `x': double
- `y': double
- `z': double
- `dx': double
- `dy': double
- `tag': integer
- `roundedRadius': double
```

pierricmora: Ok, good point. I'm not sure to understand in details what should be the desired behaviour here -- I guess you would expect 3 as output, and another method that does not exist but which could be named `gmsh.model.getTopologicalDimension()` to return 2, am I right?
I think I am done with my code. Could you have a review?

python/dolfinx/io/gmshio.py
```diff
@@ -258,6 +268,27 @@ def model_to_mesh(
...
+        _elementTypes, _elementTags, _nodeTags = model.mesh.getElements(dim=_d, tag=-1)
+        # assert only one type of elements
```
garth-wells, line 274: This looks like 1/2 code, 1/2 comment. Rephrase as a comment.

python/dolfinx/io/gmshio.py
```diff
@@ -258,6 +268,27 @@ def model_to_mesh(
...
+        # assert len(_elementTypes) == 1  # NOTE: already checked in extract_topology_and_markers
```
garth-wells, line 275: Delete commented code,

python/dolfinx/io/gmshio.py
```diff
@@ -258,6 +268,27 @@ def model_to_mesh(
...
+        _elementType_dim = _elementTypes[0]
+        if _elementType_dim not in topologies.keys():
+            raise RuntimeError("All cells are expected to be tagged once; none found")
+        nbcells = len(_elementTags[0])
```
garth-wells, line 279: House style would be `num_cells`

python/dolfinx/io/gmshio.py
```diff
@@ -258,6 +268,27 @@ def model_to_mesh(
...
+        nbcells_tagged = len(topologies[_elementType_dim]["entity_tags"])
+        if nbcells != nbcells_tagged:
+            e = (
```
garth-wells, line 282: Do we need to introduce `e`? Can the string go straight into `RuntimeError`?

python/dolfinx/io/gmshio.py
```diff
@@ -258,6 +268,27 @@ def model_to_mesh(
...
+                "All cells are expected to be tagged once;"
+                f"found: {nbcells_tagged}, expected: {nbcells}"
+            )
+            raise RuntimeError(e)
+        nbcells_tagged_once = len(np.unique(topologies[_elementType_dim]["entity_tags"]))
```
garth-wells, line 287: See comment on house style.

python/dolfinx/io/gmshio.py
```diff
@@ -258,6 +268,27 @@ def model_to_mesh(
...
+        if nbcells_tagged != nbcells_tagged_once:
+            e = "All cells are expected to be tagged once; found duplicates"
```
garth-wells, line 289: See above comment on introducing `e`.

garth-wells (changes requested):

python/dolfinx/io/gmshio.py
pierricmora: Done

python/dolfinx/io/gmshio.py
pierricmora: Done

python/dolfinx/io/gmshio.py
pierricmora: Done

python/dolfinx/io/gmshio.py
pierricmora: Done

python/dolfinx/io/gmshio.py
pierricmora: Comment deleted

garth-wells:
@jorgensd is this ready?

garth-wells (approved):

Issue #3340, opened 2024-08-07
Unpin cffi from version < 1.17


schnellerhase:
The release of `cffi` version `v1.17.0` broke the red hat CI pipeline. 

An explicit downgrade as hot fix was introduced in https://github.com/FEniCS/dolfinx/pull/3339. This should be reverted, to using the latest version of `cffi` in its runs, to fix this issue.

Interestingly this seems to be a platform dependent problem as the other CI runs did not reproduce the behavior.

garth-wells:
Unpin once https://github.com/python-cffi/cffi/pull/111 is in cffi release.

francesco-ballarin:
Just a heads up that on my Debian testing system `cffi<1.17` failed to compile from source with `python 3.13`, while `cffi==1.17` works fine (and is already packaged). The pin should ideally be gone before our next release if at all possible (meaning, if `cffi` does have a new release in the meantime).

garth-wells:
We should unpin cffi now, and raise an exception if a user tries to use it with a complex type (and advise them to rollback cffi).

PR #3343, opened 2024-08-09
General tidy and restrict cffi < 1.17 due to static condensation demo issue

jhale:
Not totally clear why but there is an issue with the static condensation demo, numba and a new version of cffi. @michalhabera you can reproduce this by branching and removing the `cffi<1.17` restriction from `pyproject.toml`.

This PR also tidies up all of the CI pipelines; I would like to ask again, please do not put `pip install x y z` into pipelines. Please use the optional dependencies feature within `pyproject.toml`.

I have also split apart cffi registering Numba's complex types and creating cffi PETSc wrappers - you can now use cffi with numba complex numbers without having PETSc installed. This was causing a test failure.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -187,14 +187,6 @@ jobs:
           cmake --build build
           cmake --install build
 
-      - name: Build C++ interface documentation
-        run: |
-          export DOLFINX_VERSION=`cmake -L build | grep DOXYGEN_DOLFINX_VERSION | cut -f2 -d "="`
-          echo $DOLFINX_VERSION
-          cd cpp/doc
-          doxygen Doxyfile
-          make html
-
       - name: Build C++ unit tests
         run: |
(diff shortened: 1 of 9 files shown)


jhale:
See also:

https://github.com/FEniCS/dolfinx/issues/3340

michalhabera:
I can reproduce. The issue seems to come from some changes in CFFI, so the global FFI builder object, i.e. `cffi.FFI()` does not recognize `float _Complex` and `double _Complex` types, call to `ffi.typeof("double _Complex")` gives
```
  File "/dolfinx-env/lib/python3.12/site-packages/cffi/model.py", line 58, in get_cached_btype
    BType = ffi._cached_btypes[self]
            ~~~~~~~~~~~~~~~~~~^^^^^^
KeyError: <_cffi_double_complex_t>
```

What seems to work is if we call `typeof("double _Complex")` on the FFI of the out-of-line compiled module. I'll prepare a PR.

michalhabera:
Fix will come with https://github.com/python-cffi/cffi/pull/111, I'd suggest we wait with pinned cffi==1.16.0 in the meantime.

PR #3344, opened 2024-08-12
Improve python exposure of `determine_point_ownership`

ordinary-slim:
Python wrapper around nanobind exposure of `determine_point_ownership` and extra optional arguments `cells` so that the search is performed only for `cells`.


python/dolfinx/geometry.py
```diff
@@ -270,3 +270,32 @@ def compute_distance_gjk(
...
+            If ``None`` then all cells are considered.
+        padding: Padding added to the bounding boxes.
+
+    Returns:
+        PointOwnershipData
+
+    """
+    if cells is None:
+        map = mesh.topology.index_map(mesh.topology.dim)
+        cells = np.arange(map.size_local + map.num_ghosts, dtype=np.int32)
```
ordinary-slim, line 300: It should be only `map.size_local` here but I don't know how to modify a pull request.

jorgensd (changes requested):
Looks good to me. Could you add a test for this?:)

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,32 @@ def compute_distance_gjk(
...
     """
     return _cpp.geometry.compute_distance_gjk(p, q)
+
+def determine_point_ownership(
+    mesh: Mesh,
+    points: npt.NDArray[np.floating], 
+    cells: typing.Optional[np.ndarray] = None,
+    padding: float = 0.0
+) -> PointOwnershipData:
+    """Build PointOwnershipData for a mesh-points pair.
```
jhale, line 280: The types of the return arguments are already documented, just say point ownership data.

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,32 @@ def compute_distance_gjk(
...
+
+    Args:
+        mesh: The mesh
+        points: The points
+        cells: Cells in mesh who potentially contain the points.
+            If ``None`` then all cells are considered.
+        padding: Padding added to the bounding boxes.
+
+    Returns:
+        PointOwnershipData
```
jhale, line 295: The types of the return arguments are already documented via type system, just say point ownership data.

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,32 @@ def compute_distance_gjk(
...
     return _cpp.geometry.compute_distance_gjk(p, q)
+
+def determine_point_ownership(
+    mesh: Mesh,
+    points: npt.NDArray[np.floating], 
+    cells: typing.Optional[np.ndarray] = None,
+    padding: float = 0.0
+) -> PointOwnershipData:
+    """Build PointOwnershipData for a mesh-points pair.
+    First, potential collisions are found
```
jhale, line 281: The docstring on the C++ function seems more precise, adapt that?

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,32 @@ def compute_distance_gjk(
 
     """
     return _cpp.geometry.compute_distance_gjk(p, q)
+
+def determine_point_ownership(
+    mesh: Mesh,
+    points: npt.NDArray[np.floating], 
+    cells: typing.Optional[np.ndarray] = None,
+    padding: float = 0.0
```
jhale, line 278: The C++ method does not have a default, make consistent.

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,32 @@ def compute_distance_gjk(
 
     """
     return _cpp.geometry.compute_distance_gjk(p, q)
+
+def determine_point_ownership(
+    mesh: Mesh,
+    points: npt.NDArray[np.floating], 
+    cells: typing.Optional[np.ndarray] = None,
```
jhale, line 277: Use npt not np for typing.

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,32 @@ def compute_distance_gjk(
...
+        padding: Padding added to the bounding boxes.
+
+    Returns:
+        PointOwnershipData
+
+    """
+    if cells is None:
+        map = mesh.topology.index_map(mesh.topology.dim)
+        cells = np.arange(map.size_local, dtype=np.int32)
+    return PointOwnershipData(_cpp.geometry.determine_point_ownership(mesh._cpp_object, points, cells, padding))
```
jhale, line 301: Not sure about exposing this struct through to Python. What about e.g. data classes @jorgensd ?

jorgensd: This is a very neat function for alot of parallel cases (including point sources).
I am not against exposing it as I rely on it for many of my test-cases.

ordinary-slim:
> Looks good to me. Could you add a test for this?:)

I added a test comparing bounding box tree results to determine_point_ownership. Another possibility would be to look at post files and hardcode rank-cell data into a test.

python/dolfinx/geometry.py
ordinary-slim: I noticed np was used in a few other annotations, I added a commit changing them to npt

python/dolfinx/geometry.py
ordinary-slim: Added default padding = 0.0 to c++ method

python/dolfinx/geometry.py
ordinary-slim: Apart from the first two paragraphs, it is now the c++ docstring. I also tried to update a bit the c++ docstring.

python/dolfinx/wrappers/geometry.cpp
```diff
@@ -180,13 +180,18 @@ void declare_bbtree(nb::module_& m, std::string type)
...
+           const T padding)
         {
           const std::size_t p_s0 = points.ndim() == 1 ? 1 : points.shape(0);
           std::span<const T> _p(points.data(), 3 * p_s0);
           return dolfinx::geometry::determine_point_ownership<T>(mesh, _p,
+                                                                 std::span(cells.data(), cells.size()),
                                                                  padding);
-        });
+        },
+      nb::arg("mesh"), nb::arg("points"), nb::arg("cells"), nb::arg("padding") = 0.0,
```
jorgensd, line 193: I guess what @jhale meant is that the C++ function in `utils.h` doesn't have a default, and that should be made consistent.

python/test/unit/geometry/test_bounding_box_tree.py
```diff
@@ -496,3 +497,82 @@ def test_surface_bbtree_collision(dtype):
...
+    mesh = create_unit_square(MPI.COMM_WORLD,
+                              num_cells_side, num_cells_side,
+                              CellType.quadrilateral, dtype=dtype)
+    cell_map = mesh.topology.index_map(tdim)
+
+    def test_po_against_bb_tree(mesh,points,cells):
+        cells = np.array([cell for cell in cells if cell < cell_map.size_local])
+        num_points = points.shape[0]
+        po = determine_point_ownership(mesh,points,cells)
+        tree = bb_tree(mesh, mesh.topology.dim, cells)
```
jorgensd, line 520: Run `ruff format` and `ruff check --fix` on the Python files to assure appropriate formatting.

python/test/unit/geometry/test_bounding_box_tree.py
```diff
@@ -496,3 +497,82 @@ def test_surface_bbtree_collision(dtype):
...
     assert len(collisions) == 1
+
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_determine_point_ownership(dtype):
+    """Find point owners (ranks and cells) using bounding box trees + global communication
+    and compare to point ownership data results."""
+    comm = MPI.COMM_WORLD
+    rank = comm.Get_rank()
+
+    tdim = 2
```
jorgensd, line 508: Rather get `tdim` from `mesh.topology.dim` after mesh creation, in case we ever decide to change this to a 3D problem.

python/test/unit/geometry/test_bounding_box_tree.py
```diff
@@ -496,3 +497,82 @@ def test_surface_bbtree_collision(dtype):
...
+    num_cells_side = 4
+    h = dtype(1.0 / num_cells_side)
+    mesh = create_unit_square(MPI.COMM_WORLD,
+                              num_cells_side, num_cells_side,
+                              CellType.quadrilateral, dtype=dtype)
+    cell_map = mesh.topology.index_map(tdim)
+
+    def test_po_against_bb_tree(mesh,points,cells):
+        cells = np.array([cell for cell in cells if cell < cell_map.size_local])
```
jorgensd, line 517:
```suggestion
        cells = np.array([cell for cell in cells if cell < cell_map.size_local], dtype=np.int32)
```

python/test/unit/geometry/test_bounding_box_tree.py
```diff
@@ -496,3 +497,82 @@ def test_surface_bbtree_collision(dtype):
...
+        num_points = points.shape[0]
+        po = determine_point_ownership(mesh,points,cells)
+        tree = bb_tree(mesh, mesh.topology.dim, cells)
+        assert(mesh.topology.cell_type in [CellType.quadrilateral,CellType.hexahedron])
```
jorgensd, line 521: Is this assertion needed? I get that the test only makes sense if the grid is axis aligned with bounding boxes, but I would rather have a comment regarding this at mesh creation.

python/test/unit/geometry/test_bounding_box_tree.py
```diff
@@ -496,3 +497,82 @@ def test_surface_bbtree_collision(dtype):
...
+        # Since box elements are considered, bounding box collisions correspond
+        # to the actual collisions
+        tree_col = compute_collisions_points(tree, points)
+        local_owners = np.zeros(2*num_points,dtype=np.int32) # rankwise data before Allreduce
```
jorgensd, line 525:
```suggestion
        local_owners = np.empty(2*num_points,dtype=np.int32) # rankwise data before Allreduce
```

python/test/unit/geometry/test_bounding_box_tree.py
```diff
@@ -496,3 +497,82 @@ def test_surface_bbtree_collision(dtype):
...
+        owners = local_owners.copy()
```
jorgensd, line 526:
```suggestion
        owners = np.empty_like(local_owners)
```

python/test/unit/geometry/test_bounding_box_tree.py
```diff
@@ -496,3 +497,82 @@ def test_surface_bbtree_collision(dtype):
...
+
+        # For the points owned by the current rank, compare owner cells
+        search_indices_dest_points = list(range(len(points)))
+        for ipoint in range(dest_points.shape[0]):
+            dest_point = dest_points[ipoint,:]
+            # Find corresponding index in `points`
+            jpoint = -1
+            for jpoint in search_indices_dest_points:
+                point = points[jpoint,:]
+                if (np.isclose(np.linalg.norm(dest_point-point),0.0)):
```
jorgensd, line 559: Here we probably need a atol dependning on the floating point tolerance used for the mesh geometry, see for instance: https://github.com/FEniCS/dolfinx/blob/06e90b8506d967f1836ffac7f289f49cb0fd17ec/python/test/unit/fem/test_symmetry.py#L42

python/test/unit/geometry/test_bounding_box_tree.py
```diff
@@ -496,3 +497,82 @@ def test_surface_bbtree_collision(dtype):
...
+        dest_points = po.dest_points()
+        # Since all processors ask about the same points,
+        # dest_points contains repeated data:
+        # dest_points = [p0, p1, p2, p3, ..., pN] * comm.size()
+        ipoint = 1
+        duplicate_start_index = -1
+        while (ipoint < dest_points.shape[0]) and not(duplicate_start_index >= 0):
+            if ((dest_points[ipoint,:] == dest_points[0,:]).all()):
+                duplicate_start_index = ipoint
+            ipoint += 1
```
jorgensd, line 547: This seems over-engineered and is comparing floating point values with equality rather than `np.allclose`.
As you know that all points are the same on every process, `duplicate_start_index` will always be `num_points`?

python/test/unit/geometry/test_bounding_box_tree.py
```diff
@@ -496,3 +497,82 @@ def test_surface_bbtree_collision(dtype):
...
+        # to the actual collisions
+        tree_col = compute_collisions_points(tree, points)
+        local_owners = np.zeros(2*num_points,dtype=np.int32) # rankwise data before Allreduce
+        owners = local_owners.copy()
+        for ipoint in range(num_points):
+            cell = tree_col.links(ipoint)
+            if len(cell) > 0:
+                local_owners[2*ipoint] = rank
+                local_owners[2*ipoint+1] = cell[0]
+        comm.Allreduce(local_owners, owners, op=MPI.SUM)
```
jorgensd, line 532:
```suggestion
        comm.Allreduce(local_owners, owners, op=MPI.SUM)
```
The current test has some unwritten assumptions.

1. All points in the input `points` are the same on all processes and in the same order
2. An input point can only be found on one process (i.e.) it is in the interior of a single cell and not on a facet/edge/vertex

If we want to keep these assumptions, then this should be specified when constructing `local_owners`, as it then make sense to use `np.zeros`, as the sum will not accumulate ranks or cell indices.
However,  shouldn't `local_owners[2*i+1]` use global cell index rather than a local index?

python/test/unit/geometry/test_bounding_box_tree.py
jorgensd: Ignore this comment for now, see comments further down instead.

python/test/unit/geometry/test_bounding_box_tree.py
ordinary-slim: No, `duplicate_start_index` is the number of points found on a processor. it is smaller than `num_points`. I realize I didn't name this variable correctly.

Example: For a two processor run, both processors ask about the same points eg. `[p0, p1, p2, p3]`. Then if two points are found on each processor, `po.dest_points() = [p0, p1, p0, p1]` for one rank and `po.dest_points() = [p2, p3, p2, p3]` for the other. The duplication stems from both processors asking about the same points.

I will replace the `==` by an `np.allclose()`

jorgensd: What about distributing the points at creation, instead of sending in multiple points. As you have all the points computed, you can get a local_range by doing something similar to: https://github.com/FEniCS/dolfinx/blob/06e90b8506d967f1836ffac7f289f49cb0fd17ec/cpp/dolfinx/common/MPI.h#L84-L107 when sending in the points. A Python implementation of this function can be lifted from: https://github.com/jorgensd/adios4dolfinx/blob/8329afed02f5f76bf6cd8d33535437acaf29aca4/src/adios4dolfinx/utils.py#L95-L113

i.e. before this function call
```python
start, stop = compute_local_range(mesh.comm, all_midpoints.shape[0])
test_po_against_bb_tree(mesh, all_midpoints[start:stop])
```

python/test/unit/geometry/test_bounding_box_tree.py
ordinary-slim: Yes, 1 and 2 are indeed assumed. I will add a comment.
Since PointOwnershipData stores local cell indices, we use local indices rather than global indices for local_owners[2*i+1].

python/test/unit/geometry/test_bounding_box_tree.py
ordinary-slim: The test also runs in 3D in the next iteration.

ordinary-slim:
Sorry, I made a mess with my branches in my fork and closed the pull request by mistake. Is it possible to reopen it pointing to [](https://github.com/ordinary-slim/dolfinx/tree/main) ?

jorgensd:
> Sorry, I made a mess with my branches in my fork and closed the pull request by mistake. Is it possible to reopen it pointing to https://github.com/ordinary-slim/dolfinx/tree/main ?
> 
> @jorgensd @jhale

I think you have to reopen it from your repository. See: https://gist.github.com/Farhan-Haseeb/fd4cbb476de4e2931c322d7bacb75e5f

jorgensd:
Reopen

ordinary-slim:
> > Sorry, I made a mess with my branches in my fork and closed the pull request by mistake. Is it possible to reopen it pointing to https://github.com/ordinary-slim/dolfinx/tree/main ?
> > @jorgensd @jhale
> 
> I think you have to reopen it from your repository. See: https://gist.github.com/Farhan-Haseeb/fd4cbb476de4e2931c322d7bacb75e5f

I can't see the PR from my repository. I deleted the original branch of the PR and it automatically closed it. The reopen button is greyed out:
Screenshot
 
The link you provided says:

> If you changed the branch name or deleted the branch, the reopen pull request button will be disabled so first you need to click the restore branch button in the right section of your commit ( change branch name / delete branch commit) inside your pull request then the reopen button will be enabled

But I can't find this `restore branch` button. Could you look for it on your side? Else I think I have to open a new PR. Sorry for the inconvenience!

jorgensd:
I cannot reopen. Please open a new PR:)

ordinary-slim:
Ok, I had to recover the deleted branch with the same name then I could reopen. Sorry for the inconvenience!

python/dolfinx/fem/assemble.py
```diff
@@ -183,7 +184,7 @@ def _assemble_vector_form(L: Form, constants=None, coeffs=None) -> la.Vector:
 
 
 @assemble_vector.register(np.ndarray)
-def _assemble_vector_array(b: np.ndarray, L: Form, constants=None, coeffs=None):
+def _assemble_vector_array(b: npt.NDArray, L: Form, constants=None, coeffs=None):
```
jorgensd, line 187:
```suggestion
def _assemble_vector_array(b: npt.NDArray, L: Form, constants=None, coeffs=None):
```
I think we should revert this one.

Rationale:
`npt.NDArray` is neat when you know the `dtype` of your array (for instance int32, int64, float64, float32 etc as you can then use `npt.NDArray[np.int32]`.

Howver, for the assembly operators, `b` can take multiple dtypes, making it more natural to let it be kept as a generic `np.ndarray` with no datatype specification.

python/dolfinx/fem/assemble.py
```diff
@@ -297,10 +298,10 @@ def _assemble_matrix_csr(
 
 
 def apply_lifting(
-    b: np.ndarray,
+    b: npt.NDArray,
```
jorgensd, line 301: Same comment as above for `b` and `x0`.

python/dolfinx/fem/bcs.py
```diff
@@ -134,8 +135,8 @@ def function_space(self) -> dolfinx.fem.FunctionSpace:
 
 
 def dirichletbc(
-    value: typing.Union[Function, Constant, np.ndarray],
-    dofs: numpy.typing.NDArray[np.int32],
+    value: typing.Union[Function, Constant, npt.NDArray],
```
jorgensd, line 138: Same comment as above regarding `value`.

python/dolfinx/fem/forms.py
```diff
@@ -102,7 +102,7 @@ def integral_types(self):
 
 def get_integration_domains(
     integral_type: IntegralType,
-    subdomain: typing.Optional[typing.Union[MeshTags, list[tuple[int, np.ndarray]]]],
+    subdomain: typing.Optional[typing.Union[MeshTags, list[tuple[int, npt.NDArray]]]],
```
jorgensd, line 105:
```suggestion
    subdomain: typing.Optional[typing.Union[MeshTags, list[tuple[int, npt.NDArray[np.int32]]]]],
```

python/dolfinx/fem/function.py
```diff
@@ -36,7 +36,7 @@ class Constant(ufl.Constant):
     ]
 
     def __init__(
-        self, domain, c: typing.Union[np.ndarray, typing.Sequence, np.floating, np.complexfloating]
+        self, domain, c: typing.Union[npt.NDArray, typing.Sequence, np.floating, np.complexfloating]
```
jorgensd, line 39: Same comment as above.

python/dolfinx/fem/function.py
```diff
@@ -90,7 +90,7 @@ class Expression:
     def __init__(
         self,
         e: ufl.core.expr.Expr,
-        X: np.ndarray,
+        X: npt.NDArray,
```
jorgensd, line 93:
```suggestion
        X: typing.Union[npt.NDArray[np.float32], npt.NDArray[np.float64]],
```

python/dolfinx/fem/function.py
```diff
@@ -196,7 +196,7 @@ def _create_expression(dtype):
     def eval(
         self,
         mesh: Mesh,
-        entities: np.ndarray,
+        entities: npt.NDArray,
```
jorgensd, line 199:
```suggestion
        entities: npt.NDArray[np.int32],
```

python/dolfinx/fem/function.py
```diff
@@ -412,8 +412,8 @@ def interpolate_nonmatching(
     def interpolate(
         self,
         u0: typing.Union[typing.Callable, Expression, Function],
-        cells0: typing.Optional[np.ndarray] = None,
-        cells1: typing.Optional[np.ndarray] = None,
+        cells0: typing.Optional[npt.NDArray] = None,
+        cells1: typing.Optional[npt.NDArray] = None,
```
jorgensd, line 416:
```suggestion
        cells0: typing.Optional[npt.NDArray[np.int32]] = None,
        cells1: typing.Optional[npt.NDArray[np.int32]] = None,
```

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,49 @@ def compute_distance_gjk(
...
+    mesh: Mesh,
+    points: npt.NDArray[np.floating],
+    cells: typing.Optional[npt.NDArray[np.int32]] = None,
+    padding: float = 0.0,
+) -> PointOwnershipData:
+    """Build point ownership data for a mesh-points pair.
+
+    First, potential collisions are found
+    by building BoundingBoxTrees for the points
+    and the mesh cells and finding potential cell-point pairs.
```
jorgensd, line 285:
```suggestion
    First, potential collisions are by computing intersections between the bounding boxes of the cells and the set of points.
```

python/dolfinx/fem/assemble.py
ordinary-slim: Should I also revert `set_bc` in the same file?

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,49 @@ def compute_distance_gjk(
...
+
+    Returns:
+        Point ownership data
+
+    Note:
+        `dest_owner` is sorted
+
+        Returns -1 if no colliding process is found
+
+        Only looks through cells owned by the process
```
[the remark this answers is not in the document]
ordinary-slim: I changed "looks through" by "returns". I will also change this on the c++ side since I copied it from there.

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,49 @@ def compute_distance_gjk(
...
+
+        A large padding value can increase the runtime of the function by
+        orders of magnitude, because for non-colliding cells
+        one has to determine the closest cell among all processes with an
+        intersecting bounding box, which is an expensive operation to perform.
```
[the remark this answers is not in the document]
ordinary-slim: Done. btw could you have a look at the docstring in c++? There is `to do` from last year that wants to improve the docstring.

python/dolfinx/fem/assemble.py
jorgensd: Let's try to minimize changes in this PR. Let's only revert those stated, and keep others (even if inconsistent, and open another PR to change them later on).

cpp/dolfinx/geometry/utils.h
```diff
@@ -663,40 +663,37 @@ graph::AdjacencyList<std::int32_t> compute_colliding_cells(
...
+/// @param[in] cells Cells to check for ownership
 /// @param[in] padding Amount of absolute padding of bounding boxes of the mesh.
 /// Each bounding box of the mesh is padded with this amount, to increase
 /// the number of candidates, avoiding rounding errors in determining the owner
 /// of a point if the point is on the surface of a cell in the mesh.
-/// @return Tuple `(src_owner, dest_owner, dest_points, dest_cells)`,
+/// @return Point ownership data with fields src_owner, dest_owner, dest_points, dest_cells,
 /// where src_owner is a list of ranks corresponding to the input
 /// points. dest_owner is a list of ranks corresponding to dest_points,
 /// the points that this process owns. dest_cells contains the
```
jorgensd, line 674:
```suggestion
/// the points that this process owns. `dest_cells` contains the
```

cpp/dolfinx/geometry/utils.h
```diff
@@ -663,40 +663,37 @@ graph::AdjacencyList<std::int32_t> compute_colliding_cells(
 /// @param[in] mesh The mesh
 /// @param[in] points Points to check for collision (`shape=(num_points,
 /// 3)`). Storage is row-major.
+/// @param[in] cells Cells to check for ownership
 /// @param[in] padding Amount of absolute padding of bounding boxes of the mesh.
 /// Each bounding box of the mesh is padded with this amount, to increase
 /// the number of candidates, avoiding rounding errors in determining the owner
 /// of a point if the point is on the surface of a cell in the mesh.
-/// @return Tuple `(src_owner, dest_owner, dest_points, dest_cells)`,
+/// @return Point ownership data with fields src_owner, dest_owner, dest_points, dest_cells,
```
jorgensd, line 671:
```suggestion
/// @return Point ownership data with fields `src_owner`, `dest_owner`, `dest_points`, `dest_cells`,
```

cpp/dolfinx/geometry/utils.h
```diff
@@ -663,40 +663,37 @@ graph::AdjacencyList<std::int32_t> compute_colliding_cells(
...
 /// where src_owner is a list of ranks corresponding to the input
```
jorgensd, line 672:
```suggestion
/// where `src_owner` is a list of ranks corresponding to the input
```

cpp/dolfinx/geometry/utils.h
```diff
@@ -663,40 +663,37 @@ graph::AdjacencyList<std::int32_t> compute_colliding_cells(
...
 /// points. dest_owner is a list of ranks corresponding to dest_points,
```
jorgensd, line 673:
```suggestion
/// points. `dest_owner` is a list of ranks corresponding to `dest_points`,
```

cpp/dolfinx/geometry/utils.h
jorgensd, line 674:
```suggestion
/// the points that this process owns. `dest_cells` contains the
```

cpp/dolfinx/geometry/utils.h
```diff
@@ -663,40 +663,37 @@ graph::AdjacencyList<std::int32_t> compute_colliding_cells(
...
 /// the points that this process owns. dest_cells contains the
 /// corresponding cell for each entry in dest_points.
 ///
 /// @note `dest_owner` is sorted
-/// @note Returns -1 if no colliding process is found
+/// @note `src_owner` is -1 if no colliding process is found
 /// @note dest_points is flattened row-major, shape `(dest_owner.size(),
 /// 3)`
-/// @note Only looks through cells owned by the process
+/// @note Only returns cells owned by the process
```
jorgensd, line 681: Do you by this mean that `dest_cells` are only local (which they are).
I think the note can be removed, as in theory, you could now send in any set of cells (owned or ghosted) in the `cells` array.

ordinary-slim: Yes that is what was meant. I will remove the note.

python/dolfinx/fem/forms.py
```diff
@@ -190,7 +190,7 @@ def form(
     dtype: npt.DTypeLike = default_scalar_type,
     form_compiler_options: typing.Optional[dict] = None,
     jit_options: typing.Optional[dict] = None,
-    entity_maps: dict[Mesh, np.typing.NDArray[np.int32]] = {},
+    entity_maps: dict[Mesh, npt.NDArray[np.int32]] = {},
```
jorgensd, line 193: Note to self, we should revise this, as it is not safe to have default dictionaries.

jorgensd (approved):
I am happy with these changes.
What do you think @jhale ?:)

python/dolfinx/wrappers/geometry.cpp
garth-wells: We avoid default args in the nanobind layer.

jorgensd, line 193:
```suggestion
      nb::arg("mesh"), nb::arg("points"), nb::arg("cells"), nb::arg("padding"),
```

python/dolfinx/fem/forms.py
```diff
@@ -190,7 +190,7 @@ def form(
     dtype: npt.DTypeLike = default_scalar_type,
     form_compiler_options: typing.Optional[dict] = None,
     jit_options: typing.Optional[dict] = None,
-    entity_maps: dict[Mesh, npt.NDArray[np.int32]] = {},
+    entity_maps: typing.Optional[dict[Mesh, np.typing.NDArray[np.int32]]] = None,
```
ordinary-slim, line 193: I think we should use npt over np.typing

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,46 @@ def compute_distance_gjk(
...
+        padding: Amount of absolute padding of bounding boxes of the mesh.
+        Each bounding box of the mesh is padded with this amount, to increase
+        the number of candidates, avoiding rounding errors in determining the owner
+        of a point if the point is on the surface of a cell in the mesh.
+
+    Returns:
+        Point ownership data
+
+    Note:
+        `dest_owner` is sorted
```
garth-wells, line 301: Python docstrings use reStructuredText (not Markdown), so need "double" backticks for monospaces.

jorgensd, line 301:
```suggestion
        ``dest_owner`` is sorted
```

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,46 @@ def compute_distance_gjk(
...
+
+        `src_owner` is -1 if no colliding process is found
```
jorgensd, line 303:
```suggestion
        ``src_owner`` is -1 if no colliding process is found
```

python/test/unit/geometry/test_bounding_box_tree.py
```diff
@@ -496,3 +498,159 @@ def test_surface_bbtree_collision(dtype):
...
+    and compare to point ownership data results."""
+    comm = MPI.COMM_WORLD
+    rank = comm.Get_rank()
+
+    tdim = dim
+    num_cells_side = 4
+    h = dtype(1.0 / num_cells_side)
+    if tdim == 2:
+        mesh = create_unit_square(
+            MPI.COMM_WORLD, num_cells_side, num_cells_side, CellType.quadrilateral, dtype=dtype
```
jhale, line 516: Can it be. parameterised across the mesh? (triangle, quad, tet, hex)?

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,46 @@ def compute_distance_gjk(
...
+
+    Note:
+        ``dest_owner`` is sorted
+
+        ``src_owner`` is -1 if no colliding process is found
+
+        A large padding value will increase the run-time of the code by orders
+        of magnitude. General advice is to use a padding on the scale of the
+        cell size.
+
```
jhale, line 308: Remove new line?

cpp/dolfinx/geometry/utils.h
```diff
@@ -663,40 +663,36 @@ graph::AdjacencyList<std::int32_t> compute_colliding_cells(
...
+/// @param[in] cells Cells to check for ownership
 /// @param[in] padding Amount of absolute padding of bounding boxes of the mesh.
 /// Each bounding box of the mesh is padded with this amount, to increase
 /// the number of candidates, avoiding rounding errors in determining the owner
 /// of a point if the point is on the surface of a cell in the mesh.
-/// @return Tuple `(src_owner, dest_owner, dest_points, dest_cells)`,
-/// where src_owner is a list of ranks corresponding to the input
-/// points. dest_owner is a list of ranks corresponding to dest_points,
-/// the points that this process owns. dest_cells contains the
+/// @return Point ownership data with fields `src_owner`, `dest_owner`, `dest_points`, `dest_cells`,
```
jhale, line 671: I think this documentation should be moved out to the struct itself.

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,46 @@ def compute_distance_gjk(
 
     """
     return _cpp.geometry.compute_distance_gjk(p, q)
+
+
+def determine_point_ownership(
+    mesh: Mesh,
+    points: npt.NDArray[np.floating],
+    cells: typing.Optional[npt.NDArray[np.int32]] = None,
+    padding: float = 0.0,
```
jhale, line 279: As this argument is critical to the behaviour of the function it is best to leave it as a non-default. Then user's need to engage with it, rather than not realise it exists.

cpp/dolfinx/geometry/utils.h
```diff
@@ -663,40 +663,36 @@ graph::AdjacencyList<std::int32_t> compute_colliding_cells(
...
 /// @note A large padding value can increase the runtime of the function by
 /// orders of magnitude, because for non-colliding cells
 /// one has to determine the closest cell among all processes with an
 /// intersecting bounding box, which is an expensive operation to perform.
 template <std::floating_point T>
 PointOwnershipData<T> determine_point_ownership(const mesh::Mesh<T>& mesh,
                                                 std::span<const T> points,
-                                                T padding)
+                                                std::span<const std::int32_t> cells,
+                                                T padding = 0.0)
```
jhale, line 689: As this argument is critical to the behaviour of the function it is best to leave it as a non-default. Then user's need to engage with it, rather than not realise it exists.

cpp/dolfinx/geometry/utils.h
```diff
@@ -663,40 +663,36 @@ graph::AdjacencyList<std::int32_t> compute_colliding_cells(
...
-/// @note Only looks through cells owned by the process
 /// @note A large padding value can increase the runtime of the function by
 /// orders of magnitude, because for non-colliding cells
 /// one has to determine the closest cell among all processes with an
 /// intersecting bounding box, which is an expensive operation to perform.
 template <std::floating_point T>
 PointOwnershipData<T> determine_point_ownership(const mesh::Mesh<T>& mesh,
                                                 std::span<const T> points,
-                                                T padding)
+                                                std::span<const std::int32_t> cells,
```
jhale, line 688: This argument could be converted to `std::optional` - can be done on later patch.

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,46 @@ def compute_distance_gjk(
...
+        points: Points to check for collision (``shape=(num_points, gdim)``)
+        cells: Cells to check for ownership
+            If ``None`` then all cells are considered.
+        padding: Amount of absolute padding of bounding boxes of the mesh.
+        Each bounding box of the mesh is padded with this amount, to increase
+        the number of candidates, avoiding rounding errors in determining the owner
+        of a point if the point is on the surface of a cell in the mesh.
+
+    Returns:
+        Point ownership data
```
jhale, line 298: Are the fields of PointOwnershipData documented in the Python wrapped class?

cpp/dolfinx/geometry/utils.h
```diff
@@ -663,40 +663,36 @@ graph::AdjacencyList<std::int32_t> compute_colliding_cells(
 /// @param[in] mesh The mesh
 /// @param[in] points Points to check for collision (`shape=(num_points,
 /// 3)`). Storage is row-major.
+/// @param[in] cells Cells to check for ownership
```
jhale, line 666: What is the sentinel value? i.e. What to pass if we want to check all cells? (`None` in Python).

Again, this could be converted to std::optional.

jhale (changes requested):

python/test/unit/geometry/test_bounding_box_tree.py
ordinary-slim: No, as the test is now it needs the elements to coincide with their bounding boxes. I'll commit a version that covers triangles and tet elements.

cpp/dolfinx/geometry/utils.h
ordinary-slim: `padding` is forwarded to `BoundingBoxTree` constructor where it is also default.

python/dolfinx/geometry.py
ordinary-slim: https://github.com/FEniCS/dolfinx/blob/6fda65d339a32cea2972ae89e9dfbde5339d6320/python/dolfinx/geometry.py#L47-L61

The doc is the same as on the c++ side.

cpp/dolfinx/geometry/utils.h
ordinary-slim: I thought it was intentional to have optional arguments on the python side that are not optional on the c++ side. I copied this from an interpolate function:

https://github.com/FEniCS/dolfinx/blob/58f13df3668d03d4ba8d79ebb931e36880672950/python/dolfinx/fem/function.py#L412-L417
https://github.com/FEniCS/dolfinx/blob/58f13df3668d03d4ba8d79ebb931e36880672950/python/dolfinx/wrappers/fem.cpp#L416
https://github.com/FEniCS/dolfinx/blob/58f13df3668d03d4ba8d79ebb931e36880672950/cpp/dolfinx/fem/Function.h#L271-L273

For this interpolate function, `cells0` is optional on the Python side but not on the c++ side. Moreover `cells1` is optional in Python, non-optional in the wrapper and optional on the c++ side.

I'll commit a version where `cells` is optional. I won't touch other places of the code.

python/dolfinx/geometry.py
jhale: @jorgensd @garth-wells  What do you think about this `padding` argument? My view is that as it's critical to the operation of the bounding box and we cannot provide a sensible default, it's better not to have it hidden as a default parameter.

jorgensd: I think it is a crucial parameter. We have already rewritten the API for non-matching interpolation since 0.7, so I think it is fine to make it required.

jhale: @ordinary-slim Please make it non-default argument in both bounding box tree and here.

cpp/dolfinx/geometry/utils.h
jhale: @ordinary-slim Please make it non-default argument in both bounding box tree and here.

python/dolfinx/geometry.py
jhale: Looks good, thanks for pointing it out to me.

The C++ documentation for the return argument describing PointOwnershipData looks over-specified given the struct is now properly documented.

cpp/dolfinx/geometry/utils.h
jhale: I see this more of a hangover from not using std::optional in C++.

In any case I think we will do the full conversion to std::optional for `v0.10.0`.

python/test/unit/geometry/test_bounding_box_tree.py
jhale: That would be great, thank you.

python/dolfinx/geometry.py
ordinary-slim: Removed most of it.

ordinary-slim:
Bumping this @jhale

cpp/dolfinx/geometry/utils.h
```diff
@@ -663,40 +663,40 @@ graph::AdjacencyList<std::int32_t> compute_colliding_cells(
...
 
+  const int tdim = mesh.topology()->dim();
+
+  std::vector<std::int32_t> local_cells;
+  if (cells.empty()) {
+    auto cell_map = mesh.topology()->index_map(tdim);
+    local_cells.resize(cell_map->size_local());
+    std::iota(local_cells.begin(), local_cells.end(), 0);
+    cells = std::span<const std::int32_t>(local_cells.data(), local_cells.size());
+  }
```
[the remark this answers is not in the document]
jhale: Is this still an issue?

cpp/dolfinx/geometry/BoundingBoxTree.h
```diff
@@ -266,7 +266,7 @@ class BoundingBoxTree
   /// build the bounding box tree for
   /// @param[in] padding Value to pad (extend) the the bounding box of
   /// each entity by.
-  BoundingBoxTree(const mesh::Mesh<T>& mesh, int tdim, T padding = 0)
```
jhale, line 269: Could you switch this literal 0 to 0.0 so it's 100% clear T is a floating point type.

jhale:
I'd like to get this merged - can you address the two final small comments?

cpp/dolfinx/geometry/utils.h
```diff
@@ -663,41 +663,41 @@ graph::AdjacencyList<std::int32_t> compute_colliding_cells(
...
 /// @note A large padding value can increase the runtime of the function by
 /// orders of magnitude, because for non-colliding cells
 /// one has to determine the closest cell among all processes with an
 /// intersecting bounding box, which is an expensive operation to perform.
 template <std::floating_point T>
 PointOwnershipData<T> determine_point_ownership(const mesh::Mesh<T>& mesh,
                                                 std::span<const T> points,
-                                                T padding)
+                                                T padding,
+                                                std::span<const std::int32_t> cells = {})
```
garth-wells, line 685: Use `std::optional` rather than an empty span.

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,45 @@ def compute_distance_gjk(
...
+    padding: float,
+    cells: typing.Optional[npt.NDArray[np.int32]] = None,
+) -> PointOwnershipData:
+    """Build point ownership data for a mesh-points pair.
+
+    First, potential collisions are found by computing intersections
+    between the bounding boxes of the cells and the set of points.
+    Then, actual containment pairs are determined using the GJK algorithm.
+
+    Args:
```
garth-wells, line 287: Formatting needs fixing - second and subsequent lines of docstring for an argument must be indented.

python/dolfinx/geometry.py
```diff
@@ -270,3 +270,45 @@ def compute_distance_gjk(
...
+    Note:
+        ``dest_owner`` is sorted
+
+        ``src_owner`` is -1 if no colliding process is found
+
+        A large padding value will increase the run-time of the code by orders
+        of magnitude. General advice is to use a padding on the scale of the
+        cell size.
+    """
+    if cells is None:
```
garth-wells, line 309: If you switch the C++ interface to `std::optional`, this `if` could be removed.

python/dolfinx/wrappers/geometry.cpp
```diff
@@ -182,13 +181,18 @@ void declare_bbtree(nb::module_& m, std::string type)
       nb::arg("mesh"), nb::arg("dim"), nb::arg("indices"), nb::arg("points"));
   m.def("determine_point_ownership",
         [](const dolfinx::mesh::Mesh<T>& mesh,
-           nb::ndarray<const T, nb::c_contig> points, const T padding)
+           nb::ndarray<const T, nb::c_contig> points,
+           nb::ndarray<const std::int32_t, nb::ndim<1>, nb::c_contig> cells,
```
garth-wells, line 185: Use `std::optional` for `cells`.

garth-wells (changes requested):
Be sure to run `clang-format` on C++ code and use `ruff` to format Python code.

cpp/dolfinx/geometry/BoundingBoxTree.h
ordinary-slim: That line was removed in a previous commit in order to make padding a mandatory argument.

cpp/dolfinx/geometry/utils.h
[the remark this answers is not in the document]
ordinary-slim: It's a nonissue, minor detail about code elegance maybe.

jhale:
I will try to update and merge this, looks pretty solid.

ordinary-slim:
> I will try to update and merge this, looks pretty solid.

Thanks!

I am under the impression that the test is too convoluted / large hence the slow progress on this PR. Let me know if you have an idea for a different test. On the other hand, I believe it is a good test in the sense that it computes the desired output in a different fashion and compares it to the actual output of the function.

jorgensd:
Superseeded by #3927

PR #3346, opened 2024-08-14
Fix block vector assembly for mixed-domain problems

jpdean:
In a mixed-domain problem, the measures in each block might differ. Currently, this causes an error because a single mesh cannot be extracted for the form.

This PR addresses the issue and adds a test that fails on main.

### DIFF
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -190,7 +190,7 @@ def form(
     dtype: npt.DTypeLike = default_scalar_type,
     form_compiler_options: typing.Optional[dict] = None,
     jit_options: typing.Optional[dict] = None,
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/fem/assembler.h
```diff
@@ -161,31 +161,7 @@ void apply_lifting(
...
-                           mesh->geometry().x(), constants, coeffs, bcs1, x0,
-                           scale);
-  }
-  else
-  {
-    auto x = mesh->geometry().x();
-    std::vector<scalar_value_type_t<T>> _x(x.begin(), x.end());
-    impl::apply_lifting<T>(b, a, mesh->geometry().dofmap(), _x, constants,
-                           coeffs, bcs1, x0, scale);
-  }
```
jorgensd, line 188: Great to get rid of this!

jorgensd (approved):

PR #3347, opened 2024-08-15
Remove pytest markers

schnellerhase:
The `pytest` markers `petsc4py` and `adios2` are removed in this PR. They are no longer necessary to filter the test cases with the compile time flags set in `dolfinx.common.has_petsc4py()` and `dolfinx.common.has_adios2()`. 

This comes with a couple benefits:

- User must not filter the test cases manually to get a passable configuration giving their specific build. (Run `pytest` on unit test should yield a passing run, no matter the build/install configuration)
- The explicit control over the `petsc4py`/`adios2` test cases is not necessary even worse, it's error prone: if we need to take into account the availability of the external modules, this should be part of the test case, for example with `pytest.importorskip` not controlled by a new not related marker.
- The petsc ci previously only ran the `petsc4py` and `adios2` test cases, not sure if this was intended, but it's not the expected behavior from the CI's test stage.


schnellerhase:
Requires prior merge of https://github.com/FEniCS/dolfinx/pull/3335

francesco-ballarin:
> The petsc ci previously only ran the petsc4py and adios2 test cases, not sure if this was intended, but it's not the expected behavior from the CI's test stage

Please ask for @garth-wells' opinion, as I am quite sure that he introduced this on purpose in #3184, one of the reasons being to reduce CI time.

garth-wells:
Yes, the design is deliberate. Tests that require an optional dependency should have a marker. We don't want to redundantly re-run all non-PETSc tests when running `x` different PETSc configurations.

schnellerhase:
I see, what about:
> User must not filter the test cases manually to get a passable configuration giving their specific build. (Run pytest on unit test should yield a passing run, no matter the build/install configuration)

Should we add the `skipif`'s to address this without removing the marker?

PR #3354, opened 2024-08-19
Extend `to_dense` for parallel use cases

schnellerhase:
The (debug) functionality `to_dense` of the CSR matrix is extended to support the usage in parallel scenarios.

Before it returned the local-local block of the matrix (which in sequential is the whole matrix), i.e. the interaction between local rows and local cols. 

In parallel runs, this drops quite a lot of information, especially we loose all interactions across processes. This is now changed, to `to_dense` returning the full/global dense rows that are owned by the process. This means the previous behavior is still reproducible by filtering for the columns of choice, but if needed the full parallel block is now available in a dense format.

This will be needed to allow for a PR regarding the transfer operators of a multigrid implementation - see [here](https://github.com/schnellerhase/dolfinx/pull/25/files#diff-e28601892fce9bf896b8e8cc69f6fcf8ef04c4fa0b08b9c11934ea5c72af4d85) for an example of this use case.

### DIFF
--- a/cpp/dolfinx/la/MatrixCSR.h
+++ b/cpp/dolfinx/la/MatrixCSR.h
@@ -630,15 +630,17 @@ std::vector<typename MatrixCSR<U, V, W, X>::value_type>
 MatrixCSR<U, V, W, X>::to_dense() const
 {
   const std::size_t nrows = num_all_rows();
-  const std::size_t ncols
-      = _index_maps[1]->size_local() + _index_maps[1]->num_ghosts();
+  const std::size_t ncols = _index_maps[1]->size_global();
   std::vector<value_type> A(nrows * ncols * _bs[0] * _bs[1], 0.0);
   for (std::size_t r = 0; r < nrows; ++r)
     for (std::int32_t j = _row_ptr[r]; j < _row_ptr[r + 1]; ++j)
       for (int i0 = 0; i0 < _bs[0]; ++i0)
         for (int i1 = 0; i1 < _bs[1]; ++i1)
         {
-          A[(r * _bs[1] + i0) * ncols * _bs[0] + _cols[j] * _bs[1] + i1]
+          std::vector<std::int32_t> local_col {_cols[j]};
+          std::vector<std::int64_t> global_col{0};
+          _index_maps[1]->local_to_global(local_col, global_col);
+          A[(r * _bs[1] + i0) * ncols * _bs[0] + global_col[0] * _bs[1] + i1]
               = _data[j * _bs[0] * _bs[1] + i0 * _bs[1] + i1];
         }
 
--- a/cpp/dolfinx/la/MatrixCSR.h
+++ b/cpp/dolfinx/la/MatrixCSR.h
@@ -637,8 +637,8 @@ MatrixCSR<U, V, W, X>::to_dense() const
       for (int i0 = 0; i0 < _bs[0]; ++i0)
         for (int i1 = 0; i1 < _bs[1]; ++i1)
         {
-          std::vector<std::int32_t> local_col {_cols[j]};
-          std::vector<std::int64_t> global_col{0};
+          std::array<std::int32_t, 1> local_col {_cols[j]};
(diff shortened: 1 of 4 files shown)


cpp/test/matrix.cpp
```diff
@@ -215,23 +217,36 @@ void test_matrix()
...
-      Aref(Aref_data.data(), 8, 8);
-  Aref(0, 0) = 1;
-  Aref(4, 5) = 2.3;
+      T, MDSPAN_IMPL_STANDARD_NAMESPACE::extents<
+             std::size_t, 8, MDSPAN_IMPL_STANDARD_NAMESPACE::dynamic_extent>>
+      Aref(Aref_data.data(), 8, A.index_map(1)->size_global());
+
+  auto to_global_col = [&](auto col)
+  {
+    std::vector<std::int64_t> tmp(1);
```
jorgensd, line 237:
```suggestion
    std::array<std::int64_t,1> tmp;
```

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -630,15 +630,17 @@ std::vector<typename MatrixCSR<U, V, W, X>::value_type>
...
   std::vector<value_type> A(nrows * ncols * _bs[0] * _bs[1], 0.0);
   for (std::size_t r = 0; r < nrows; ++r)
     for (std::int32_t j = _row_ptr[r]; j < _row_ptr[r + 1]; ++j)
       for (int i0 = 0; i0 < _bs[0]; ++i0)
         for (int i1 = 0; i1 < _bs[1]; ++i1)
         {
-          A[(r * _bs[1] + i0) * ncols * _bs[0] + _cols[j] * _bs[1] + i1]
+          std::vector<std::int32_t> local_col {_cols[j]};
+          std::vector<std::int64_t> global_col{0};
+          _index_maps[1]->local_to_global(local_col, global_col);
```
jorgensd, line 642: Couldn't we move this call outside the loop, apply `local_to_global` on all of `_cols` at the same time?
It would avoid having `std::vector<T>/std::array<T,1>` inside the loop (and avoid calling: https://github.com/FEniCS/dolfinx/blob/ce09b5e8b56cbd69bcf5f17afb4c2b7217b21706/cpp/dolfinx/common/IndexMap.cpp#L883) multiple times).

schnellerhase: That would work. But doesn't it reduce readability of the test case quite drastically? We would loose the clear correlation of for example here we look at the local index 4 but in terms of global indices.

jorgensd: As long as we document it I think it is fine to add an initial comment stating:
```c++
// Transform all local column indices to global indices prior to looping through all rows
std::vector<std::int64_t> global_columns (_cols.size());
_index_maps[1]->local_to_global(_cols, global_columns);
```
and then in turn use `global_columns[j]` inside the loops.

Lets at least move the transformation out to line 637, as it is now called for every block size, which is unnecessary. And use `std::array<T, 1>` for the allocation.

schnellerhase: How about we use an `std::map` if we precompute? That way we keep the info in the line of the check?

jorgensd: I think it becomes a rather expensive (map lookup) and not necessary.
This isn't the most performance critical part of the code, but we should avoid potentially expensive patterns when necessary.

I'll let some of the other developers chime in on this as well, maybe @chrisrichardson or @garth-wells has some other opinions than I.

schnellerhase:
Need to include this in another PR, so I'll squash

cpp/dolfinx/la/MatrixCSR.h
chrisrichardson: Let's bear in mind this is really for debugging, so performance is not too important.

garth-wells (approved):

Issue #1741, opened 2021-10-05
Add function that provides a summary fo mesh memory usage


garth-wells:
Break down by different components, e.g. different entities, connectivities, etc.

PR #3356, opened 2024-08-20
Add rectangle mesh generation testing

schnellerhase:
This introduces explicit (index based) testing of the mesh generation for 2d rectangles. For now only in sequential.

These data structures are needed for more complex test cases (multigrid), following in other PRs. 

Also this gives a good baseline/examples for the fixing of https://github.com/FEniCS/dolfinx/issues/1741

### DIFF
--- a/cpp/test/mesh/rectangle.cpp
+++ b/cpp/test/mesh/rectangle.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2024 Paul Kühner
+// Copyright (C) 2024 Paul T. Kühner
 //
 // This file is part of DOLFINX (https://www.fenicsproject.org)
 //
--- a/cpp/test/mesh/rectangle.cpp
+++ b/cpp/test/mesh/rectangle.cpp
@@ -1,0 +1,211 @@
+// Copyright (C) 2024 Paul Kühner
+//
+// This file is part of DOLFINX (https://www.fenicsproject.org)
+//
+// SPDX-License-Identifier:    LGPL-3.0-or-later
+
+#include <mpi.h>
+
+#include <catch2/catch_template_test_macros.hpp>
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/matchers/catch_matchers.hpp>
+#include <catch2/matchers/catch_matchers_floating_point.hpp>
+#include <catch2/matchers/catch_matchers_range_equals.hpp>
+#include <catch2/matchers/catch_matchers_vector.hpp>
+
+#include <dolfinx/common/MPI.h>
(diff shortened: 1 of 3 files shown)


jpdean:
`Geometry::input_global_indices` might be needed here to avoid hardcoding values

schnellerhase:
> Geometry::input_global_indices might be needed here to avoid hardcoding values

Thanks for pointing me to this function. I think to make properly use of this and to simplify the tests in this and the related PR's we probably want to split the testing of the generation (everything prior to call to `create_mesh`) and the mesh generation testing (after redistributing, creating ghost nodes). So we need an interface change. How about something like `impl::create_interval_cells -> {x, cells}` (and equally for the other mesh types - this should allow for a shared code path for the mesh generation in `generation.h` all together) that is then called within `create_interval` to facilitate this?

What remains an open problem with this restructuring would be the question of how to check the mesh parallelization effectively (controlling in test cases which process owns which node and where ghost nodes are generated).

garth-wells:
Is this ready, or does it need changes based on @jpdean's comment?

schnellerhase:
I'd say its good to go, and we can adapt later on. Addressing the before mentioned issue will require more changes, than only the tests, so maybe a good idea to separate it from this.

cpp/test/mesh/rectangle.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+#include "util.h"
+
+using namespace dolfinx;
+using namespace Catch::Matchers;
+
+TEMPLATE_TEST_CASE("Rectangle quadrilateral mesh",
+                   "[mesh][rectangle][quadrilateral]", float, double)
+{
+  using T = TestType;
```
garth-wells, line 32: Where is `TestType` defined?

garth-wells (commented):
I'm wondering does this add anything over what we can test (more easily) from Python?

cpp/test/mesh/rectangle.cpp
schnellerhase: Is part of the catch macro `TEMPLATE_TEST_CASE`. Here, its either `float` or `double`.

schnellerhase:
In principle these tests could be moved to python, I think. Nevertheless I would push for keeping them here in the cpp part. We are strictly unit testing here and thus should avoid as many additional layers as possible.

For example the related tests with parallel cases (in other PRs) currently are not transferable to python due to `https://github.com/FEniCS/dolfinx/pull/3362` not working in python as noted in https://github.com/FEniCS/dolfinx/pull/3362.

schnellerhase:
Superseded by #3359

PR #3359, opened 2024-08-21
Add interval mesh generation testing

schnellerhase:
This introduces explicit (index based) testing of the mesh generation for 1d interval meshes.

These data structures are needed for more complex test cases (multigrid), following in other PRs.

Also this gives a good baseline/examples for the fixing of https://github.com/FEniCS/dolfinx/issues/1741


schnellerhase:
macos produces another node labeling....needs fixing.

garth-wells (approved):

schnellerhase:
What do you think of the interface change in the generation @garth-wells? Also this allows for the the vertex position tests as demonstrated with for the interval cases (both sequential and parallel) with `input_global_indices`.

I'm not yet sure how we extend this to the connectivity, though.

schnellerhase:
I think the state of [8d91259](https://github.com/FEniCS/dolfinx/pull/3359/commits/8d91259484a72d7819de0cc13ca53333833f0ea0) resolved the rebase problems.

PR #3260, opened 2024-06-07
Use `enabled_coefficients` to restrict packing of coefficients

jorgensd:
Pack only entities used in integral type:
MWE:
```python
from mpi4py import MPI
import dolfinx
import ufl

mesh = dolfinx.mesh.create_unit_cube(MPI.COMM_WORLD, 50, 50, 50)
V = dolfinx.fem.functionspace(mesh, ("Lagrange", 3))
N = 50
us = [dolfinx.fem.Function(V) for _ in range(N)]
for i,u in enumerate(us):
    u.x.array[:] = i + 1
    u.name=f"u_{i}"

volume_form = us[0] * ufl.dx
surface_form = sum(us[i] * ufl.ds for i in range(N))

combined_form = volume_form + surface_form
compiled_form = dolfinx.fem.form(combined_form)

import time
start = time.perf_counter()
coeffs = dolfinx.fem.assemble.pack_coefficients(compiled_form._cpp_object)
end =time.perf_counter()
print(f"{dolfinx.common.git_commit_hash=} {end-start=:.5e}")
print(coeffs[(dolfinx.fem.IntegralType.cell, -1)])
```
On the main branch, all coefficients are packed for the cell integral, even if only the first coefficient is used in that integral.
Runtime on this branch:
```bash
dolfinx.common.git_commit_hash='002e1fa9360995f9c71414a683e7ce618d6101e9' end-start=1.70250e+00
```
on main
```bash
dolfinx.common.git_commit_hash='64e35310085683f4f61804e70b8b58b9bb8653ae' end-start=2.89335e+00
```
It would also resolve #3256 and simplify some logic in: #3224

PR #3361, opened 2024-08-21
Skip negative cell indices when packing coefficients

jorgensd:
Continuation of https://github.com/FEniCS/dolfinx/pull/3260.

If one has multiple disjoint sub-meshes in a variational form, then restricting packing by `enabled_coefficients` per integral type is not sufficient.

Example follows:
Divide an interval mesh in three disjoint cell sets (left, center, right).
If we want to create a coupling from the right submesh to the parent mesh at the interface between center and right),
we can add an integral `(u_parent("+")-u_parent("-"))*v_right("+")*dS(interface_rc)` to our variational form.
Similarly we could do the same for the left integral: `(u_parent("+")-u_parent("-"))*v_left("+")*dS(interface_lc)`.

However, now both coefficient v_right and v_left will be packed for all facets marked with either interface_rc or interface_lc, which leads to undefined behavior (segfaults).

The following PR adds a check after fetching the cell during packing, and only packs coefficients if the cell index is positive.

Issue reported by @RemDelaporteMathurin when trying to adapt my example: https://gist.github.com/jorgensd/8a5c32f491195e838f5863ca88b27bce#file-problem-py to multiple surfaces.

### DIFF
--- a/cpp/dolfinx/fem/utils.h
+++ b/cpp/dolfinx/fem/utils.h
@@ -914,6 +914,8 @@ void pack_coefficient_entity(std::span<T> c, int cstride,
     {
       auto entity = entities.subspan(e, estride);
       std::int32_t cell = fetch_cells(entity);
+      if (cell < 0)
+        continue;
       auto cell_coeff = c.subspan((e / estride) * cstride + offset, space_dim);
       pack<T, 1>(cell_coeff, cell, bs, v, cell_info, dofmap, transformation);
     }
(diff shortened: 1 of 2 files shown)


RemDelaporteMathurin (approved):
Tested it locally with [this script](https://gist.github.com/jorgensd/29ca28ce55c4da3dba0e7031993f997e) and everything runs fine. The SEGV I was previously having doesn't appear anymore. 

Thanks @jorgensd ! 🎉

jorgensd:
@jpdean do you have time to look at this?

garth-wells (approved):
@jorgensd Is there a change we could make to avoid this check? I haven't looked at the details for a long time . . .

Issue #3358, opened 2024-08-21
Platform stable graph partitioner


schnellerhase:
### Describe new/missing feature

During the design of parallel test cases for PR (TODO) it was observed that the graph partitioners (at least scotch and parmetis, did not test kahip) produce unstable results across different platforms. 

This is a problem when it comes to reproducibility of results, especially for test cases. Tests running on a different platform will create differently parallelized meshes and thus mesh structures. Additionally for a user it might be hard to understand why a platform change (mind you with same parallelization settings) yields different, not strictly, comparable results - this should be at least be possible to enforce.

To fix this one would either need to figure out how to make these behave equally - tried that myself without luck - or introduce some kind of nice interface to facilitate this, i.e. a partitioner dummy that reproduces a given (previously computed) partitioning.

In the before mentioned PR this is facilitated by just mimicking the behavior of `graph::scotch::partitioner(...)` explicitly and switching the result depending on the current communicator size.

garth-wells:
I don't think this is worth the effort, and we should really design tests that are independent of the partitioning (and (re-)ordering too).

jorgensd:
I think reading in the partitioning information from an existing mesh would resolve this for most users (https://scientificcomputing.github.io/io4dolfinx/docs/api.html#io4dolfinx.read_mesh) where you can set `read_from_partition=True` if partitioning information is stored.

MariusCausemann:
I have had a case where differences in the partitioning had a big impact on the conditioning of the system, causing non-deterministic solver failures. For debugging purposes, I think it would be very useful to be able to reproduce the same partitioning deterministically. @jorgensd suggestion sounds like a good way to resolve this!

PR #3362, opened 2024-08-21
 Remove refine option `redistribute` and `ghost_mode` in favor of an explicit partitioner

schnellerhase:
The refine call previously hid from the user the control of the repartitioner in use for the refined mesh. It allowed for some configuration through `redistribute` and `ghost_mode` which in combination resulted in a selection of specific partitioning routine. 

Now the control over which partitioner is used is given to the caller. Previous behavior of `redistribute=False` is maintained as default with the new (refinement specific) partitioning routine `maintain_coarse_partitioner`. 

TODO:
- [ ] `create_cell_partitioner` not working in python, allowing no usage in the tests of the refinement routines
- [ ] export messy and very repetetive when it comes to handling the callbacks of partitioner - not restricted to these new changes -> move to general location and introduce shared naming


cpp/dolfinx/refinement/option.h
```diff
@@ -0,0 +1,50 @@
...
+#include <cstdint>
+#include <type_traits>
+
+namespace dolfinx::refinement
+{
+
+/// @brief Options for data to compute during mesh refinement.
+enum class Option : std::uint8_t
+{
+    none = 0b00,
```
chrisrichardson, line 18: Add comments (doxygen compatible) for each option

schnellerhase: Addressed in #3322 with 135ef1f60bf37edb26291ac03db3830ec1e0b00d. Which needs to merged prior to this one.

garth-wells:
@schnellerhase could you resolve the merge conflicts?

schnellerhase:
I guess the TODO's could also me retro fitted in a follow up PR.

cpp/dolfinx/refinement/refine.h
```diff
@@ -19,44 +20,29 @@
...
-template <std::floating_point T>
-mesh::Mesh<T>
-create_refined_mesh(const mesh::Mesh<T>& mesh,
-                    const graph::AdjacencyList<std::int64_t>& cell_adj,
-                    std::span<const T> new_vertex_coords,
-                    std::array<std::size_t, 2> xshape, bool redistribute,
-                    mesh::GhostMode ghost_mode)
+
+/// TODO: Document function
+inline graph::AdjacencyList<std::int32_t> maintain_coarse_partitioner(
```
garth-wells, line 25: The implementation should be moved into `refine.cpp`.

cpp/dolfinx/refinement/refine.h
```diff
@@ -19,44 +20,28 @@
...
+/// refined mesh. Meaning, process local data is just maintained and no
+/// redistribution happens.
+///
+/// @param[in] comm MPI Communicator
+/// @param[in] nparts Number of partitions (input has no effect)
+/// @param[in] cell_types Cell types in the mesh
+/// @param[in] cells Lists of cells of each cell type.
+/// @return Destination ranks for each cell on this process
+graph::AdjacencyList<std::int32_t>
+maintain_coarse_partitioner(MPI_Comm comm, int nparts,
```
garth-wells, line 33: I think this name could be improved - maybe 'identity_partitioner'?

schnellerhase: Open for new names, I agree it is not nicely named. 

I wasn't entirely sure if this partitioner routine makes any sense outside of the refinement context (which the name kind of hints). https://github.com/FEniCS/dolfinx/issues/3358 might be a related to this.

garth-wells: I don't think we need it - we can pass `nullptr` as the partitioning function, in which case `mesh::create_mesh` doesn't do any redistribution. Or am I missing something?

jorgensd:
@schnellerhase Imagine you have a ghosted mesh. Then the `maintain_coarse_partitioner` is not doing the right thing, as it removes any ghost cells. See for instance: https://github.com/jorgensd/adios4dolfinx/blob/main/src/adios4dolfinx/checkpointing.py#L656-L674,  https://github.com/jorgensd/adios4dolfinx/blob/main/src/adios4dolfinx/checkpointing.py#L485-L497 on how to maintain a given partition.

Secondly, could you elaborate on why `create_cell_partitioner` doesn't work from Python? As I use it in: https://jsdokken.com/dolfinx_docs/meshes.html#mesh-partitioning

> I guess the TODO's could also be retro fitted in a follow up PR, do not strictly break functionality as is.

Could you elaborate on why it doesn't break functionality? Previously tests with `redistribute=True` passed, while now all tests retain an initial (unghosted) partitioning.

schnellerhase:
Had a thought about it and you are completely right @jorgensd, the `maintain_coarse_partitioner` does not work correctly for the ghosted parts.

schnellerhase:
What we want to do is, recreate a mesh with the same ghost mode type, but no redistribution of the mesh nodes, those are already on the correct processes once we call `create_mesh` (refine produces the refinement only on the owned parts of the mesh).

schnellerhase:
That means we need a partitioner that only performs the  `compute_destination_ranks` if ghosting is enabled, otherwise just the identity, right?

schnellerhase:
Regarding the problems with the `python` interface to `create_cell_partitioner`, the following fails for me, which should be a shortened version of the tutorial code:

```python
from mpi4py import MPI

import numpy as np

import dolfinx
import ufl
from basix.ufl import element
from dolfinx.graph import partitioner_scotch

quadrilaterals = np.array([], dtype=np.int64)
quad_points = np.array([[0, 0], [0.3, 0]], dtype=np.float64)

ufl_quad = ufl.Mesh(element("Lagrange", "quadrilateral", 1, shape=(2,)))


partitioner = dolfinx.mesh.create_cell_partitioner(partitioner_scotch())
quad_mesh = dolfinx.mesh.create_mesh(
    MPI.COMM_WORLD, quadrilaterals, quad_points, ufl_quad, partitioner=partitioner
)
```

fails with 

```bash
TypeError: create_mesh(): incompatible function arguments. The following argument types are supported:
    1. create_mesh(arg0: MPICommWrapper, arg1: collections.abc.Sequence[ndarray[dtype=int64, writable=False, shape=(*), order='C']], arg2: collections.abc.Sequence[dolfinx.cpp.fem.CoordinateElement_float32], arg3: ndarray[dtype=float32, writable=False, order='C'], arg4: collections.abc.Callable[[MPICommWrapper, int, collections.abc.Sequence[dolfinx.cpp.mesh.CellType], collections.abc.Sequence[numpy.ndarray[dtype=int64, writable=False, ]]], dolfinx.cpp.graph.AdjacencyList_int32], /) -> dolfinx.cpp.mesh.Mesh_float32
    2. create_mesh(comm: MPICommWrapper, cells: ndarray[dtype=int64, writable=False, shape=(*, *), order='C'], element: dolfinx.cpp.fem.CoordinateElement_float32, x: ndarray[dtype=float32, writable=False, order='C'], partitioner: collections.abc.Callable[[MPICommWrapper, int, collections.abc.Sequence[dolfinx.cpp.mesh.CellType], collections.abc.Sequence[numpy.ndarray[dtype=int64, writable=False, ]]], dolfinx.cpp.graph.AdjacencyList_int32] | None) -> dolfinx.cpp.mesh.Mesh_float32
    3. create_mesh(arg0: MPICommWrapper, arg1: collections.abc.Sequence[ndarray[dtype=int64, writable=False, shape=(*), order='C']], arg2: collections.abc.Sequence[dolfinx.cpp.fem.CoordinateElement_float64], arg3: ndarray[dtype=float64, writable=False, order='C'], arg4: collections.abc.Callable[[MPICommWrapper, int, collections.abc.Sequence[dolfinx.cpp.mesh.CellType], collections.abc.Sequence[numpy.ndarray[dtype=int64, writable=False, ]]], dolfinx.cpp.graph.AdjacencyList_int32], /) -> dolfinx.cpp.mesh.Mesh_float64
    4. create_mesh(comm: MPICommWrapper, cells: ndarray[dtype=int64, writable=False, shape=(*, *), order='C'], element: dolfinx.cpp.fem.CoordinateElement_float64, x: ndarray[dtype=float64, writable=False, order='C'], partitioner: collections.abc.Callable[[MPICommWrapper, int, collections.abc.Sequence[dolfinx.cpp.mesh.CellType], collections.abc.Sequence[numpy.ndarray[dtype=int64, writable=False, ]]], dolfinx.cpp.graph.AdjacencyList_int32] | None) -> dolfinx.cpp.mesh.Mesh_float64

Invoked with types: Intracomm, ndarray, dolfinx.cpp.fem.CoordinateElement_float64, ndarray, nanobind.nb_func
```

garth-wells:
Superseded by #3444

PR #3363, opened 2024-08-22
Introduction of transfer opeartor assembly for geometry multigrid opeartions

schnellerhase:
TODO

- [ ] Check assumptions on space, for example $P^1$.
- [ ] finalize python interface, probably needs some more wrapping for the matrix assembly.
- [ ] extend python testing to other dimensions


schnellerhase:
Failing tests best fixed after https://github.com/FEniCS/dolfinx/issues/3443.

jhale:
I've taken a look at this and it's not clear what spaces/elements are supported (and in turn, meshes with which topological dimension). Can you clarify that?

schnellerhase:
Currently only $P^1$, scalar valued, but on any topological dimension. The extension to vector valued should require no big change. But other elements or higher degrees will require more work. Especially currently I take the connectivity information for the operators from the mesh, instead of working with the dof maps. So I guess as long as Dofmap is the identity this approach should apply for now.

jhale:
I think support for block size should be relatively straightforward to add.

The P1 limitation is more pressing - am I right in saying at the moment the code says mesh vertices == p1 dofs? So to generalise it you would need to look at the entities/dofs across all topological dimensions?

schnellerhase:
Also the implementation of the current inclusion map here, is very much brute force - checking all points and requiring an all to all communication step. However I couldn't think about a better general solution. However the data structure I believe is very well suited for the transfer operation as it makes clear how one mesh is to be recovered in another. So, better it would be if one can produce such an inclusion mapping as a by product, as for example of the refinement operation.

PR #3367, opened 2024-08-22
Fix refine constructor (currently overwriting ufl_cargo)

jorgensd:
Create new ufl mesh as ufl mesh carries ufl_cargo which references the e latest mesh added through the mesh constructor.

This might be relevant for you as well @schnellerhase.

### DIFF
--- a/python/dolfinx/mesh.py
+++ b/python/dolfinx/mesh.py
@@ -331,7 +331,8 @@ def refine(
         mesh1 = _cpp.refinement.refine(mesh._cpp_object, redistribute)
     else:
         mesh1 = _cpp.refinement.refine(mesh._cpp_object, edges, redistribute)
-    return Mesh(mesh1, ufl.Mesh(mesh._ufl_domain.ufl_coordinate_element()))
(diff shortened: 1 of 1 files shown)


schnellerhase:
Thanks for the ping, looks like an interesting bug. But it only occurs in the python module, right? In that case I should be fine :sweat_smile: - I haven't really used the refinement calls from python.

schnellerhase:
Can we add a check to the `Mesh` constructor to defend against it in general? Doesn't seem like an absurd use case of the Mesh constructor that caused that bug.

Maybe a check for `_ufl_domain._ufl_cargo is None` before overwriting? Not sure if we can assume this to be `None` beforehand though.

jorgensd:
> Can we add a check to the `Mesh` constructor to defend against it in general? Doesn't seem like an absurd use case of the Mesh constructor that caused that bug.
> 
> Maybe a check for `_ufl_domain._ufl_cargo is None` before overwriting? Not sure if we can assume this to be `None` in general though.

Not quite sure we can do that. The cargo can in theory be set through the `ufl.Mesh`-constructor: https://github.com/FEniCS/ufl/blob/004a678320e5c65015d56071b968ee08314aecc1/ufl/domain.py#L65-L73

PR #3107, opened 2024-03-18
Support DG-0 functions in VTXWriter

jorgensd:
Resolve #2720. Works with Paraview 5.12.0.RC3.

I've decided to keep it such that only single function-space functions should be saved in a single bp file.
This could of course be changed now, to store a mixture of DG-0 and a single other element function.

This is up for discussion.

We could also discuss how to implement the VTX mesh sharing policy for DG-0 functions.

Another point we now could address is meshtags (at least celltags) for VTXWriter.

@massimiliano-leoni do you mind testing this as well?:)

PR #3375, opened 2024-09-02
Support list of DG-0 + other space in VTXWriter

ordinary-slim:
Continuation of #3107 so that we're able to output both DG0 + another space in the same bp file. The previous PR by @jorgensd says:

> I've decided to keep it such that only single function-space functions should be saved in a single bp file.
> This could of course be changed now, to store a mixture of DG-0 and a single other element function.
> 
> This is up for discussion.

Since VTX is the recommended writer, I think it's necessary to support this feature. This code doesn't modify previous behavior regarding the written post files.

MWE
```python3
from dolfinx import fem, mesh, io
from mpi4py import MPI

cells_per_side = 4

domain  = mesh.create_unit_square(MPI.COMM_WORLD,
                                  cells_per_side,
                                  cells_per_side,
                                  )

cg1 = fem.functionspace(domain, ("Lagrange", 1),)
dg0 = fem.functionspace(domain, ("Discontinuous Lagrange", 0),)

u_cg1 = fem.Function(cg1, name="u_cg1")
u_dg0 = fem.Function(dg0, name="u_dg0")

expression =  lambda x : x[0]*x[1] 
u_cg1.interpolate(expression)
u_dg0.interpolate(expression)

writer = io.VTXWriter(domain.comm,
                      "result_folder/post.bp",
                      output=[u_cg1,u_dg0])
writer.write(0.0)
writer.close()
```


jorgensd:
Currently DG-0 support in VTX is limited by: https://github.com/ornladios/ADIOS2/issues/4179

ordinary-slim:
> Currently DG-0 support in VTX is limited by: [ornladios/ADIOS2#4179](https://github.com/ornladios/ADIOS2/issues/4179)

Thank you for the information, a bit of a bummer. This PR could still be useful for steady data.

ordinary-slim:
What is the recommended way to output both CG1 and DG0 time dependent data?

ordinary-slim:
> @ordinary-slim do you mind testing if time-dependent data is rendering correctly in paraview 6.0?

I've been using this branch without issues with Paraview 5.13. I'll edit this message when I get around trying Paraview 6.0 (soonish).

cpp/dolfinx/io/ADIOS2Writers.h
```diff
@@ -472,6 +478,24 @@ class VTXWriter : public ADIOS2Writer
...
+            if (!impl::is_cellwise(*V->element()))
+            {
+              V0 = V;
+              has_V0_changed = true;
+            }
+          },
+          v);
+      if (has_V0_changed)
+        break;
+    }
```
jorgensd, line 500: @ordinary-slim could you elaborate on this logic?
Here you look through all input functions, and replace the first function space with any (the last) function space if it is not a "constant" space. Is this so that you can mix the non-constant and constant spaces?

ordinary-slim: This V0 is used to check that all functions have the same dofmap. In the vtk scheme, we can provide both point data and cell data. Since the DG0 functions have this separate path, we only have to verify that the remaining non-DG0 functions use the same dofmap.

cpp/dolfinx/io/ADIOS2Writers.h
```diff
@@ -585,17 +606,9 @@ class VTXWriter : public ADIOS2Writer
     _engine->template Put<double>(var_step, t);
 
     // If we have no functions or DG functions write the mesh to file
-    if (_is_piecewise_constant or _u.empty())
+    if (_has_piecewise_constant or _u.empty())
```
jorgensd: I think this logic is wrong.
It should be: If we have a piecewise constant and no other functions write mesh to file.

jorgensd: Actually, it should now check if it has any non-DG functions.
I guess the easiest way to do that is by using
`    auto [names, dg0_names] = impl_vtx::extract_function_names<T>(u);`
and check that `names.size() == 0`.

jorgensd (approved):
I've made a slight change to ensure that this is working as expected when mixing different function spaces.
Other than that, I think this is good to go. @jhale, shall we merge?:)

PR #3378, opened 2024-09-06
Reduce complexity of `create_submap_indices`

jorgensd:
Current code in  `communicate_ghosts_to_owners`  pads `send_data` such that we always send `ghosts.size()` amount of data to all owners, irrespective of if `include_ghosts` has any `True` values in it. This is later discarded in an if test (removed in this PR).

The current assumption with `-1` entries was also not documented, and when used in `compute_submap_ghost_indices` all `include_ghosts` where set to 1, which would ensure one get no -1 back.

The current PR removes this padding, which simplifies the logic of `create_submap_indices` and possibly undefined behavior of `global_to_local` when `recv_indices<0`

### DIFF
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -216,7 +216,7 @@ compute_submap_indices(const IndexMap& imap,
       for (int j = recv_disp[i]; j < recv_disp[i + 1]; ++j)
       {
         // Compute the local index
-        std::int32_t idx = recv_indices[j];
+        std::int64_t idx = recv_indices[j];
         assert(idx >= 0);
         std::int32_t idx_local = idx - local_range[0];
         assert(idx_local >= 0);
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -220,25 +215,22 @@ compute_submap_indices(const IndexMap& imap,
     {
       for (int j = recv_disp[i]; j < recv_disp[i + 1]; ++j)
       {
-        // Check if the index is included in the submap by process dest[i].
-        // If so, it must be assigned a submap owner
-        if (std::int64_t idx = recv_indices[j]; idx != -1)
+        // Compute the local index
+        std::int32_t idx = recv_indices[j];
```
garth-wells, line 219: Integer type here looks too narrow.

jorgensd, line 219:
```suggestion
        std::int64_t idx = recv_indices[j];
```

jpdean (approved):
Nice, looks good to me!

PR #3379, opened 2024-09-06
Try Windows test without pytest-xdist

jhale:


jhale:
This didn't work. The C++ library is probably throwing an assert as pointed out here https://github.com/FEniCS/dolfinx/issues/3333- but it does not lead to meaningful information printed in the output of `py.test`.

PR #3380, opened 2024-09-08
Enforce linting in CI for C++ and CMake files

schnellerhase:
Changes to CI:
- `ruff` combined to a single call
- `clang-format` combined to a single call and no longer allowed to fail
- `cmake-format` no longer allowed to fail

Additionally this contains the necessary format changes that were previously missed due to allowed failure.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -38,18 +38,13 @@ jobs:
         uses: actions/setup-python@v5
         with:
           python-version: '3.12'
+
       - name: Install linting tools
         run: pip install clang-format cmake-format[YAML] mypy ruff
-      - name: ruff .py files in C++ code
-        run: |
-          cd cpp/
-          ruff check .
-          ruff format --check .
-      - name: ruff Python interface checks
-        run: |
-          cd python/
-          ruff check .
-          ruff format --check .
+
+      - name: Run ruff
+        run: ruff check python/ cpp/
+
       - name: mypy checks
         run: |
           cd python/
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -53,14 +53,10 @@ jobs:
           mypy test
 
       - name: Run clang-format
-        run: |
-          clang-format --version
-          find cpp/ python/dolfinx/wrappers/ -type f \( -name "*.cpp" -o -name "*.h" \) | xargs clang-format --dry-run --Werror
+        run: find cpp/ python/dolfinx/wrappers/ -type f \( -name "*.cpp" -o -name "*.h" \) | xargs clang-format --dry-run --Werror
 
-      - name: cmake-format (non-blocking)
-        continue-on-error: true
-        run: |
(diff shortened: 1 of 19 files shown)


.github/workflows/ccpp.yml
```diff
@@ -38,40 +38,25 @@ jobs:
...
-          ruff format --check .
-      - name: ruff Python interface checks
-        run: |
-          cd python/
-          ruff check .
-          ruff format --check .
-      - name: mypy checks
+
+      - name: Run ruff
+        run: ruff check python/ cpp/
```
jhale, line 46: The previous setup was deliberate. It checked formatting, and additionally the split allows the user to quickly see in the CI log which part of the Python interface is failing.

.github/workflows/ccpp.yml
```diff
@@ -38,40 +38,25 @@ jobs:
...
-        run: |
-          cd python/dolfinx/wrappers
-          clang-format --version
-          find . -type f \( -name "*.cpp" -o -name "*.h" \) | xargs clang-format --dry-run --Werror
-      - name: cmake-format (non-blocking)
-        continue-on-error: true
-        run: |
-          find . -type f \( -name "*.cmake" -o -name "*.cmake.in" -o -name "CMakeLists.txt" \) | xargs cmake-format --check
+
+      - name: Run clang-format
```
jhale, line 55: Non-blocking was intentional.

.github/workflows/ccpp.yml
```diff
@@ -38,40 +38,25 @@ jobs:
...
+        run: find cpp/ python/dolfinx/wrappers/ -type f \( -name "*.cpp" -o -name "*.h" \) | xargs clang-format --dry-run --Werror
+
+      - name: Run cmake-format
```
jhale, line 58: Non-blocking was intentional. Both make-format and clang-format are not very stable and behaviour differs widely between versions. This makes it difficult for developers to pass these checks. We periodically do a PR which runs these utilities using the versions in the container images.

.github/workflows/ccpp.yml
```diff
@@ -38,40 +38,25 @@ jobs:
...
-          cd python/dolfinx/wrappers
-          clang-format --version
-          find . -type f \( -name "*.cpp" -o -name "*.h" \) | xargs clang-format --dry-run --Werror
-      - name: cmake-format (non-blocking)
-        continue-on-error: true
-        run: |
-          find . -type f \( -name "*.cmake" -o -name "*.cmake.in" -o -name "CMakeLists.txt" \) | xargs cmake-format --check
+
+      - name: Run clang-format
+        run: find cpp/ python/dolfinx/wrappers/ -type f \( -name "*.cpp" -o -name "*.h" \) | xargs clang-format --dry-run --Werror
```
jhale, line 56: Split into two steps is intended so it is clear which part of the library is the issue.

jhale (changes requested):

schnellerhase:
In the case of all this being done by intent, I will close the PR.

PR #3384, opened 2024-09-11
Add read option of prims/wedges through xdmffile

### LINKED ISSUE #3351: [BUG]: Reading XMDF with prism elements
### Summarize the issue

The mesh export and import using XDMF file format does not appear to work for meshes with prismatic elements. It works for all other elements.

### How to reproduce the bug

I prepared a small script to illustrate the error

### Minimal Example (Python)

```Python
import numpy as np

# FeniCSx
from dolfinx import mesh, io
from mpi4py import MPI

# Define mesh parameters
comm = MPI.COMM_WORLD
nel = 10                                         # number of elements per domain

Dhi = mesh.create_box(comm, [np.array([0., 0., 0.]), np.array([1., 1., 1.])],
                     [nel,nel,nel], cell_type=mesh.CellType.prism)
with io.XDMFFile(comm, "box-pri6.xdmf", "w") as xdmf:
   xdmf.write_mesh(Dhi)
   
with io.XDMFFile(comm, "box-pri6.xdmf", "r") as xdmf:
    Dhf = xdmf.read_mesh()
```

### Output (Python)

```bash
RuntimeError: Cannot recognise cell type. Unknown value: wedge
```

### Version

0.8.0

### DOLFINx git commit

_No response_

### Installation

I used a Docker image on a Windows 11

### Additional information

_No response_

jorgensd:
Resolve #3351

### DIFF
--- a/cpp/dolfinx/io/xdmf_utils.cpp
+++ b/cpp/dolfinx/io/xdmf_utils.cpp
@@ -57,6 +57,7 @@ xdmf_utils::get_cell_type(const pugi::xml_node& topology_node)
          {"quadrilateral_9", {"quadrilateral", 2}},
          {"quadrilateral_16", {"quadrilateral", 3}},
          {"hexahedron", {"hexahedron", 1}},
+         {"wedge", {"prism", 1}},
          {"hexahedron_27", {"hexahedron", 2}}};
 
   // Convert XDMF cell type string to DOLFINx cell type string


cpp/dolfinx/io/xdmf_utils.cpp
```diff
@@ -57,6 +57,7 @@ xdmf_utils::get_cell_type(const pugi::xml_node& topology_node)
          {"quadrilateral_9", {"quadrilateral", 2}},
          {"quadrilateral_16", {"quadrilateral", 3}},
          {"hexahedron", {"hexahedron", 1}},
+         {"wedge", {"prism", 1}},
```
garth-wells, line 60: When did 'prism' enter DOLFINx? No a good name since more than one 2D cell shape is a prism, see https://en.wikipedia.org/wiki/Prism_(geometry).

garth-wells (approved):

PR #3391, opened 2024-09-12
Fix element name in mixed Poisson demo

jpdean:
The mixed Poisson demo says it uses the Brezzi-Douglas-Fortin-Marini element in the docs, but it's actually using the BDMCF element, which is (as far as I'm aware) just a Brezzi-Douglas-Marini element defined on a quad (the CF stands for cubical face, not Fortin). This PR fixes the name.

@mscroggs please correct me if this is wrong.


jhale:
I think this was my mistake, thanks Joe!

PR #3392, opened 2024-09-12
Remove NBX call from sub-IndexMap creation

### LINKED ISSUE #3100: NBX should be removed from sub index map creation
A call to NBX was added in [this](https://github.com/FEniCS/dolfinx/pull/3099) PR to fix a bug in sub map creation, however, it should be possible to compute the sub map destination ranks without it by adding an extra communication step

jorgensd:
Resolve #3100.

The code gets a bit complex, so I will walk you through the idea here.

1. All processes send indices that are ghosts on their process (and that they need in the sub-index map) to the owning process.
2. On that process, we check if the process also needs this index. If it does, we:
   a. Assign ownership of this index to existing owner, which means that we know that all indices that sent this to the process, is in the `dest_ranks` of the sub-map.
3. If the process does not need the index, we re-assign ownership to the lowest rank that requires this index.
   b. This means that we need to send all the other processes that require that index to the new owner, as they are new dest ranks.

The new code does this by using the (owner->ghost) communciator that already exists in the code, and sends only the unique set of new dest ranks to the new owner.
There we accumulate the dests from 2a. and 3b. into one unique sorted dest_rank list.


garth-wells:
@jpdean could you cast your eye over this?

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -199,8 +200,11 @@ compute_submap_indices(const IndexMap& imap,
   // their owner in the submap. This is required since not all indices in
   // `recv_indices` will necessarily be in `indices` on this process, and thus
   // other processes must own them in the submap.
-
+  // If ownership of received index doesn't change, then this process has the
+  // receiving rank as a destination
   std::vector<int> recv_owners(send_disp.back());
+  std::vector<int> submap_dest;
+  submap_dest.reserve(dest.size() + 1);
```
jpdean, line 207: Does this `reserve` make sense? How much can we really say about the size of `submap_dest` at this point? It will depend on the indices to include and on how the owners change. Though maybe reserving something in the right ballpark is better for performance...

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -241,19 +248,60 @@ compute_submap_indices(const IndexMap& imap,
     std::ranges::sort(global_idx_to_possible_owner);
 
     // Choose the submap owner for each index in `recv_indices`
+    // and pack destination ranks for each process that has received new indices
```
jpdean, line 251: So the idea here is that this process (let's call it p) may have received an index (lets call it `i`) that it hasn't been included in the submap, and thus we need to pick the submap owner and tell it what its destinations are for this index. The destination ranks are just the other processes that sent index `i` to this process (`p`). If I've interpreted this correctly, can we maybe improve this comment? I think `received new indices` especially could be made more descriptive

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -241,19 +248,60 @@ compute_submap_indices(const IndexMap& imap,
...
+
+        // NOTE: Could choose new owner in a way that is is better for
+        // load balancing, though the impact is probably only very small
+        auto it = std::ranges::lower_bound(global_idx_to_possible_owner, idx,
+                                           std::ranges::less(),
+                                           [](auto& e) { return e.first; });
+        assert(it != global_idx_to_possible_owner.end() and it->first == idx);
+        send_owners.push_back(it->second);
 
+        // If rank that sent this ghost is the owner, send all other ranks
```
jpdean, line 274:
```suggestion
        // If rank that sent this ghost is the submap owner, send all other ranks
```

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -241,19 +248,60 @@ compute_submap_indices(const IndexMap& imap,
...
+        if (it->second == dest[i])
+        {
+          // Find upper limit of recv index and pack all dest ranks for this
+          // process
+          auto it_upper = std::ranges::upper_bound(
+              it, global_idx_to_possible_owner.end(), idx, std::ranges::less(),
+              [](auto& e) { return e.first; });
+          std::transform(it + 1, it_upper,
```
jpdean, line 282: Is `it + 1` here so that we don't send the submap owner's rank as a destination rank to itself? If so, we should add a comment, otherwise it might break if the above code that chooses the owner is modified (e.g. for load balancing purposes)

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -241,19 +248,60 @@ compute_submap_indices(const IndexMap& imap,
     std::ranges::sort(global_idx_to_possible_owner);
 
     // Choose the submap owner for each index in `recv_indices`
+    // and pack destination ranks for each process that has received new indices
     std::vector<int> send_owners;
     send_owners.reserve(recv_indices.size());
-    for (auto idx : recv_indices)
+    std::vector<int> new_owner_dest_ranks;
+    new_owner_dest_ranks.reserve(recv_indices.size());
```
jpdean, line 255: Again, not sure if this reserve makes sense

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -241,19 +248,60 @@ compute_submap_indices(const IndexMap& imap,
...
+          // process
+          auto it_upper = std::ranges::upper_bound(
+              it, global_idx_to_possible_owner.end(), idx, std::ranges::less(),
+              [](auto& e) { return e.first; });
+          std::transform(it + 1, it_upper,
+                         std::back_inserter(new_owner_dest_ranks),
+                         [](auto& e) { return e.second; });
+        }
+      }
+      // Remove duplicate new dest ranks from recv process
```
jpdean, line 287: Sorry, can you explain why there would be duplicates? Since `new_owner_dest_ranks` is created from `global_idx_to_possible_owner`, do we not already have a unique set of ranks for each index?

jpdean (approved):
Nice, it looks good to me! Just a few comments. Also, maybe it would be clearer if things like `new_owner` were renamed to `submap_owner` etc?

cpp/dolfinx/common/IndexMap.cpp
jorgensd: You know that you will have somewhere around the number of destination ranks the original map had. I.e. if you had no dest ranks, you will like not have any new ones now. One reason for reserving is so that the buffer is initialized for MPI communication later on, as modern MPI is not fond of uninitialized buffers.

cpp/dolfinx/common/IndexMap.cpp
jorgensd: Yes, that is the idea.
As we at ownership determination have all ranks that wanted `i`, we know that they will have it in their map, and thus the new owner has these as `dest` ranks.

cpp/dolfinx/common/IndexMap.cpp
jorgensd: You need a minimal reserve for mpi communication.
Are you talking about the new owner dest ranks reservation?
We could reduce this to `.reserve(1)`, which would be the minimal case, where each re-assigned index is only required by the new owning process (no real determination of ownership).

jorgensd, line 255:
```suggestion
    new_owner_dest_ranks.reserve(1);
```

cpp/dolfinx/common/IndexMap.cpp
jorgensd: Yes, that is the reasoning. Feel free to add a suitable comment. (as it is in quite close proximity to the above, i.e. uses the `it` from the lower_bound, I feel it is quite clear that they relate).

cpp/dolfinx/common/IndexMap.cpp
jorgensd: We only need the unique set of ranks for all newly assigned indices to the process.
I.e. if a process has taken over index i from process (N, M) and index j from (M, N, Q), what we want to send to this process is (M,N, Q), not (N,M,M,N,Q), as one would then have to do a sort and remove on all received indices.

cpp/dolfinx/common/IndexMap.cpp
jorgensd, line 251:
```suggestion
    // and pack destination ranks for each process that has received new indices.
    // During ownership determination, we know what other processes
    // requires this index, and add them to the destination set.
```

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -241,19 +248,60 @@ compute_submap_indices(const IndexMap& imap,
...
+                                           std::ranges::less(),
+                                           [](auto& e) { return e.first; });
+        assert(it != global_idx_to_possible_owner.end() and it->first == idx);
+        send_owners.push_back(it->second);
 
+        // If rank that sent this ghost is the submap owner, send all other ranks
+        if (it->second == dest[i])
+        {
+          // Find upper limit of recv index and pack all dest ranks for this
+          // process
```
jorgensd, line 278:
```suggestion
          // Find upper limit of recv index and pack all ranks from
          // ownership determination (except new owner) as  dest ranks.
```

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -241,19 +248,60 @@ compute_submap_indices(const IndexMap& imap,
...
+          auto it_upper = std::ranges::upper_bound(
+              it, global_idx_to_possible_owner.end(), idx, std::ranges::less(),
+              [](auto& e) { return e.first; });
+          std::transform(it + 1, it_upper,
+                         std::back_inserter(new_owner_dest_ranks),
+                         [](auto& e) { return e.second; });
+        }
+      }
+      // Remove duplicate new dest ranks from recv process
```
jorgensd, line 287:
```suggestion
      // Remove duplicate new dest ranks from recv process
      // The new owning process can have taken ownership of multiple
      // indices from the same rank.
```

jorgensd:
@jpdean what do you think about the revised comments? (in suggestions).

garth-wells:
> Nice, it looks good to me! Just a few comments. Also, maybe it would be clearer if things like `new_owner` were renamed to `submap_owner` etc?

@jorgensd thoughts?

jorgensd:
> > Nice, it looks good to me! Just a few comments. Also, maybe it would be clearer if things like `new_owner` were renamed to `submap_owner` etc?
> 
> @jorgensd thoughts?

I made the name `new_owner` as that list only accounts for indices that have re-assigned ownership.
It does not account for dofs that: Already was owned by the processes, and was required by the owning process in the new map.

garth-wells (approved):

PR #3397, opened 2024-09-13
Add missing docs regarding diagonal

jorgensd:
Also add some notes about when diagonal is set.


python/dolfinx/fem/assemble.py
```diff
@@ -245,6 +246,9 @@ def assemble_matrix(
     Note:
         The returned matrix is not finalised, i.e. ghost values are not
         accumulated.
+
+    Note:
+        Diagonal entries are only inserted for matrices with the same trial and test space.
```
jhale, line 251: Notes are more asides. As this is a critical point I would add it to the argument as 'for constrained degrees of freedom belonging to the same trial and test space.'

jhale (changes requested):

python/dolfinx/fem/petsc.py
```diff
@@ -431,6 +431,9 @@ def assemble_matrix(
         The returned matrix is not 'assembled', i.e. ghost contributions
         have not been communicated.
 
+    Note:
```
jhale, line 434: Same comment.

jhale (dismissed):

python/dolfinx/fem/petsc.py
```diff
@@ -431,11 +431,14 @@ def assemble_matrix(
...
+        Diagonal entries are only inserted for matrices with the same trial and test space.
+
```
jorgensd, line 436:
```suggestion
```

garth-wells (approved):

garth-wells:
@jhale could you approve this PR if you're happy with it?

PR #3398, opened 2024-09-13
Strict mypy checking

### LINKED ISSUE #2297: Progressively turn on stricter mypy checking
We currently run mypy in a permissive mode.

https://github.com/FEniCS/dolfinx/blob/main/python/setup.cfg#L8

Long term we should aim to turn most or all of these checks to `True`.

jorgensd:
Try removing all restrictions except missing import types. Would resolve #2297 if passing

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -69,15 +69,6 @@ markers = [
 
 
 [tool.mypy]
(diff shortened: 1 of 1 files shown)


schnellerhase:
I've got some related attempts at https://github.com/schnellerhase/dolfinx/pull/28/ - if one was to retry maybe a good starting point.

PR #3399, opened 2024-09-13
Save FFI module in fem form.

bpachev:
This is a relatively minor change to make the ffi module returned by ffcx available in the Form class, in addition to the ufcx_form Python object and the code. The reason for this PR is to support my add-on package providing GPU accelerated assembly on top of dolfinx (https://github.com/bpachev/cuda-dolfinx). I need a way to pass the ufcx_form pointer to C, similar to what is done in the instantiation of the cpp Form class, e.g:

```
        f = ftype(
            module.ffi.cast("uintptr_t", module.ffi.addressof(ufcx_form)),
            V,
            coeffs,
            constants,
            subdomains,
            _entity_maps,
            mesh,
        )
```

The module variable is discarded after fe.form returns. I'm aware that it is possible to find the module by replicating the cache search logic within ffcx, but feel like saving module in Form is much simpler and stabler. I apologize in advance if this change isn't needed. If there's a simpler solution, I'm all ears :).


python/dolfinx/fem/forms.py
```diff
@@ -47,6 +47,7 @@ def __init__(
         ],
         ufcx_form=None,
         code: typing.Optional[str] = None,
+        module=None,
```
garth-wells, line 50: Add a type hint for `module`.

python/dolfinx/fem/forms.py
```diff
@@ -75,6 +78,11 @@ def code(self) -> typing.Union[str, None]:
         """C code strings"""
         return self._code
 
+    @property
+    def module(self):
```
garth-wells, line 82: Add return type hint.

garth-wells (dismissed):
Looks good - just needs some type hints.

python/dolfinx/fem/forms.py
bpachev: Done

python/dolfinx/fem/forms.py
bpachev: Done

jhale:
Can you ruff check and ruff format this? Then we will merge.

python/dolfinx/fem/forms.py
```diff
@@ -47,6 +48,7 @@ def __init__(
         ],
         ufcx_form=None,
         code: typing.Optional[str] = None,
+        module: typing.Optional[types.ModuleType] = None,
```
jhale, line 51: Is types.ModuleType correct? Not what I expected.

bpachev: The module object is created in ffcx using importlib.util.module_from_spec. According to
https://docs.python.org/3/library/types.html#types.ModuleType, types.ModuleType appears to be the correct type for Python modules.

bpachev:
Fixed the linting issue and addressed the comment about the type.

PR #3400, opened 2024-09-14
Expose geometry constructor in Python

### LINKED ISSUE #2978: C++ Geometry object not constructible from Python
Hello everyone!

I was wondering why the Geometry_float32/float64 objects are no longer accessible from Python.
Specifically, I'm trying to write code to refine a P2 mesh down to an equivalent P1 mesh for drawing and other operations.

This used to work in previous versions, but now seems not to be possible. It seems that a constructor for GF32/64 is not exposed to the Python side.

I'd appreciate your help very much, or if there is a better way to accomplish this, I'm also open to that (although I would like the flexibility since I have a few more uses in mind)

Thank you!

jorgensd:
Resolve #2978.

Illustrative use-case:
```python
from mpi4py import MPI
from dolfinx import cpp, mesh

msh = mesh.create_unit_square(MPI.COMM_WORLD, 10, 10, cpp.mesh.CellType.triangle)

dofmap = msh.geometry.dofmap.reshape(-1).copy()
igi = msh.geometry.input_global_indices.copy()
nodes =  msh.geometry.x.reshape(-1).copy()
new_geom = cpp.mesh.Geometry_float64(msh.geometry.index_map(), 
                                     dofmap, msh.geometry.cmap,nodes, 2, igi)

```

### DIFF
--- a/python/dolfinx/wrappers/mesh.cpp
+++ b/python/dolfinx/wrappers/mesh.cpp
@@ -209,8 +209,7 @@ void declare_mesh(nb::module_& m, std::string type)
                 nb::handle());
           },
           nb::rv_policy::reference_internal, nb::arg("i"),
-          "Get the geometry dofmap associated with coordinate element i "
-          "(mixed "
+          "Get the geometry dofmap associated with coordinate element i (mixed "
           "topology)")
       .def("index_map", &dolfinx::mesh::Geometry<T>::index_map)
       .def_prop_ro(
--- a/python/dolfinx/wrappers/mesh.cpp
+++ b/python/dolfinx/wrappers/mesh.cpp
@@ -171,20 +171,24 @@ void declare_mesh(nb::module_& m, std::string type)
           "__init__",
           [](dolfinx::mesh::Geometry<T>* self,
              std::shared_ptr<const dolfinx::common::IndexMap> index_map,
-             nb::ndarray<std::int32_t, nb::ndim<2>, nb::c_contig> dofmap,
+             nb::ndarray<const std::int32_t, nb::ndim<2>, nb::c_contig> dofmap,
              const dolfinx::fem::CoordinateElement<T>& element,
-             nb::ndarray<T, nb::ndim<2>, nb::c_contig> x,
-             nb::ndarray<std::int64_t, nb::ndim<1>, nb::c_contig>
+             nb::ndarray<const T, nb::ndim<2>> x,
+             nb::ndarray<const std::int64_t, nb::ndim<1>, nb::c_contig>
                  input_global_indices)
           {
-            // Pad geometry to be 3D
-            std::size_t shape0 = x.shape(0);
             int shape1 = x.shape(1);
-            std::vector<T> x_vec(3 * shape0, 0);
-            for (std::size_t i = 0; i < shape0; ++i)
+            std::vector<T> x_vec;
+            if (shape1 == 3 and x.stride(0) == 3 and x.stride(1) == 1)
+              x_vec.assign(x.data(), x.data() + x.size());
+            else
             {
-              std::copy_n(std::next(x.data(), shape1 * i), shape1,
-                          std::next(x_vec.begin(), 3 * i));
+              // Pad geometry to be 3D
+              x_vec.assign(3 * x.shape(0), 0);
+              auto _x = x.view();
+              for (std::size_t i = 0; i < x.shape(0); ++i)
+                for (int j = 0; j < shape1; ++j)
+                  x_vec[3 * i + j] = _x(i, j);
             }
 
             new (self) dolfinx::mesh::Geometry<T>(
--- a/python/dolfinx/wrappers/mesh.cpp
+++ b/python/dolfinx/wrappers/mesh.cpp
@@ -171,21 +171,19 @@ void declare_mesh(nb::module_& m, std::string type)
           "__init__",
           [](dolfinx::mesh::Geometry<T>* self,
              std::shared_ptr<const dolfinx::common::IndexMap> index_map,
-             nb::ndarray<std::int32_t, nb::ndim<1>, nb::c_contig> dofmap,
+             nb::ndarray<std::int32_t, nb::ndim<2>, nb::c_contig> dofmap,
              const dolfinx::fem::CoordinateElement<T>& element,
-             nb::ndarray<T, nb::ndim<1>, nb::c_contig> x, int dim,
+             nb::ndarray<T, nb::ndim<2>, nb::c_contig> x, int dim,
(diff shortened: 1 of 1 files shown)


python/dolfinx/wrappers/mesh.cpp
```diff
@@ -167,6 +167,28 @@ void declare_mesh(nb::module_& m, std::string type)
   std::string pyclass_geometry_name = std::string("Geometry_") + type;
   nb::class_<dolfinx::mesh::Geometry<T>>(m, pyclass_geometry_name.c_str(),
                                          "Geometry object")
+      .def(
+          "__init__",
+          [](dolfinx::mesh::Geometry<T>* self,
+             std::shared_ptr<const dolfinx::common::IndexMap> index_map,
+             nb::ndarray<std::int32_t, nb::ndim<1>, nb::c_contig> dofmap,
```
garth-wells, line 174: I think this would make more sense from Python as a 2D array.

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -167,6 +167,28 @@ void declare_mesh(nb::module_& m, std::string type)
...
+             const dolfinx::fem::CoordinateElement<T>& element,
+             nb::ndarray<T, nb::ndim<1>, nb::c_contig> x, int dim,
```
garth-wells, line 176: Make `x` 2D and remove `dim`.

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -167,6 +167,28 @@ void declare_mesh(nb::module_& m, std::string type)
...
+                 input_global_indices)
+          {
+            std::vector<std::int32_t> dofmap_vec(dofmap.data(),
+                                                 dofmap.data() + dofmap.size());
+            std::vector<T> x_vec(x.data(), x.data() + x.size());
+            std::vector<std::int64_t> igi_vec(
+                input_global_indices.data(),
+                input_global_indices.data() + input_global_indices.size());
+            new (self) dolfinx::mesh::Geometry<T>(
+                std::move(index_map), std::move(dofmap_vec), element,
```
garth-wells, line 187: No need to move a shared pointer.

garth-wells: Would be simpler to do 
```cpp
dolfinx::mesh::Geometry<T>(index_map, 
  std::vector(dofmap.data(), dofmap.data() + dofmap.size()), 
  element,   
  std::vector(x.data(), x.data() + x.size()),
  x.shape(1), 
  std::vector(input_global_indices.data(), input_global_indices.data() + input_global_indices.size())
  );
```

garth-wells (changes requested):

python/dolfinx/wrappers/mesh.cpp
jorgensd: I don't think we can do that, as we send in 3D coordinates, even when we want to use 1D or 2D meshes.

python/dolfinx/wrappers/mesh.cpp
jorgensd: I don't agree, as we would have to pad it in the nanobind-wrapper, as we have forced input as three-dimensional in the Geometry constructor in C++.

garth-wells: I don't follow. It's naturally a 2D array. See https://github.com/FEniCS/dolfinx/blob/7bb01adaa736a325ca048e37aec0b51cf9ff1f2e/cpp/dolfinx/mesh/Geometry.h#L121.

Why would it need padding?

jorgensd:
Updated example:
```python
from mpi4py import MPI
from dolfinx import cpp, mesh

msh = mesh.create_unit_square(MPI.COMM_WORLD, 10, 10, cpp.mesh.CellType.triangle)

dofmap = msh.geometry.dofmap.copy()
igi = msh.geometry.input_global_indices.copy()
nodes =  msh.geometry.x.copy()
new_geom = cpp.mesh.Geometry_float64(msh.geometry.index_map(), 
                                     dofmap, msh.geometry.cmap,nodes, 2, igi)
```

python/dolfinx/wrappers/mesh.cpp
jorgensd: Sorry, wrote this in the wrong place, meant for the points. I've now modified the code to take in 2D arrays for dofmap and points, but kept the `dim` argument.

python/dolfinx/wrappers/mesh.cpp
garth-wells: Send in 1D, 2D or 3D coords. Simpler and no significant performance issue since the data gets copied anyway.

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -167,6 +167,26 @@ void declare_mesh(nb::module_& m, std::string type)
   std::string pyclass_geometry_name = std::string("Geometry_") + type;
   nb::class_<dolfinx::mesh::Geometry<T>>(m, pyclass_geometry_name.c_str(),
                                          "Geometry object")
+      .def(
+          "__init__",
+          [](dolfinx::mesh::Geometry<T>* self,
+             std::shared_ptr<const dolfinx::common::IndexMap> index_map,
+             nb::ndarray<std::int32_t, nb::ndim<2>, nb::c_contig> dofmap,
+             const dolfinx::fem::CoordinateElement<T>& element,
+             nb::ndarray<T, nb::ndim<2>, nb::c_contig> x, int dim,
```
garth-wells, line 176: `x` should have shape `(num_points, dim)` and `dim` removed, or `x` should be constrained to have shape `(num_points, 3)`. At the moment it's neither. The former is simpler.

jhale:
Looks good. Would be good to put Python wrapper over the typed C++ methods.

jorgensd:
New API:
```python
from mpi4py import MPI
from dolfinx import cpp, mesh

msh = mesh.create_unit_square(MPI.COMM_WORLD, 10, 10, cpp.mesh.CellType.triangle)

dofmap = msh.geometry.dofmap.copy()
igi = msh.geometry.input_global_indices.copy()
nodes =  msh.geometry.x[:, :msh.geometry.dim]
new_geom = cpp.mesh.Geometry_float64(msh.geometry.index_map(), 
                                     dofmap, msh.geometry.cmap,nodes, igi)

import numpy as np
np.testing.assert_allclose(msh.geometry.x, new_geom.x)
```

jorgensd:
> Looks good. Would be good to put Python wrapper over the typed C++ methods.

I will add that for geometry in another PR, as it will touch the Mesh-class.

garth-wells (approved):

PR #3405, opened 2024-09-15
Fix nanobinding for interprocess facets

jorgensd:
Duplicate binding returning the wrong dtype and a `list` is now changed to one returning a `numpy`-array.


jhale:
Nasty bug, thanks for fixing.

jhale (approved):

PR #3403, opened 2024-09-15
Add Python-interface for geometry constructor and geometry class

jorgensd:
Continuation of #3400.

Add `Geometry` Python class and appropriate constructor (`create_geometry`).
Expose some bindings and wrap return types (`Geometry` and `CoordinateElement`) when returned from Python objects.

PR #3406, opened 2024-09-15
Create topology Python wrapper

jorgensd:
Wrap Topology as a Python class.

Follows along lines #3403


python/dolfinx/mesh.py
```diff
@@ -67,10 +66,140 @@
 ]
 
 
+class Topology:
+    """Topology for for a :class:`dolfinx.mesh.Mesh`"""
```
RemDelaporteMathurin, line 70:
```suggestion
    """Topology for a :class:`dolfinx.mesh.Mesh`"""
```
```suggestion
    """Topology for for a :class:`dolfinx.mesh.Mesh`"""
```
```suggestion
    """Topology for for a :class:`dolfinx.mesh.Mesh`"""
```

python/dolfinx/mesh.py
```diff
@@ -67,10 +66,140 @@
...
+    def dim(self) -> int:
+        """Return the topological dimension of the mesh."""
+        return self._cpp_object.dim
+
+    @property
+    def entity_types(self) -> list[list[CellType]]:
+        """Get the entity types in the topology for all topological dimensions."""
+        return self._cpp_object.entity_types
+
+    def get_cell_permutation_info(self) -> npt.NDArray[np.uint32]:
```
jhale, line 142: The content of each np.uint32 being interpreted as a bitfield should probably be mentioned.

I can only find https://numpy.org/doc/stable/reference/generated/numpy.unpackbits.html in numpy which is for uint8 only.

python/dolfinx/mesh.py
```diff
@@ -67,10 +66,140 @@
...
+        """Returns the permutation information"""
+        return self._cpp_object.get_cell_permutation_info()
+
+    def get_facet_permutations(self) -> npt.NDArray[np.uint8]:
```
jhale, line 146: The content of each np.uint8 being interpreted as a bitfield should probably be mentioned along with a pointer to `np.unpackbits`.

python/dolfinx/mesh.py
```diff
@@ -67,10 +66,140 @@
...
+        """Get the permutation number to apply to a facet."""
```
jhale, line 147: a facet? Isn't this all of the facets?

python/test/unit/mesh/test_mixed_topology.py
```diff
@@ -242,7 +242,7 @@ def test_create_entities():
 
     quad_v = mesh.topology.connectivity((2, qi), (0, 0))
     tri_v = mesh.topology.connectivity((2, ti), (0, 0))
-    ims = mesh.topology.index_maps(2)
+    ims = mesh.topology._cpp_object.index_maps(2)
```
jhale, line 245: Should this go through the new or extended version of the new interface?

python/dolfinx/mesh.py
```diff
@@ -67,10 +66,140 @@
...
+        Args: The C++ topology object
+        Note:
+            topology objects should not usually be created using this
+            initializer directly.
+        """
+        self._cpp_object = topology
+
+    def cell_name(self) -> str:
+        """String representation of the cell-type of the topology"""
+        return self._cpp_object.cell_name()
```
jhale, line 85: Is this necessary? Can't it come from the `CellType` itself?

python/dolfinx/mesh.py
```diff
@@ -67,10 +66,140 @@
...
+
+    @property
+    def comm(self):
+        return self._cpp_object.comm
+
+    def create_connectivity(self, d0: int, d1: int):
+        """Create connectivity between given pair of dimensions, ``d0`` and  ``d1``.
+
+        Args:
+            d0: Dimension of entity one is mapping from
```
jhale, line 112: plural entities?

python/dolfinx/mesh.py
```diff
@@ -67,10 +66,140 @@
...
+            d1: Dimension of entity one is mapping to
```
jhale, line 113: plural entities?

python/dolfinx/mesh.py
```diff
@@ -67,10 +66,140 @@
...
+        self._cpp_object.create_connectivity(d0, d1)
+
+    def create_entities(self, dim: int) -> int:
+        """Create entities of given topological dimension.
+
+        Args:
+            dim: Topological dimension
+
+        Returns:
+            Number of newly created entities, returns -1 if entities already existed
```
jhale, line 124: `-1` as sentinel doesn't seem very Pythonic - `None`?

python/dolfinx/mesh.py
```diff
@@ -67,10 +66,140 @@
...
+        """
+        return self._cpp_object.create_entities(dim)
+
+    def create_entity_permutations(self):
```
jhale, line 128: Possible to type return argument?

python/dolfinx/mesh.py
```diff
@@ -67,10 +66,140 @@
...
+
+        Args:
+            d0: Dimension of entity one is mapping from
+            d1: Dimension of entity one is mapping to
+        """
+        if (conn := self._cpp_object.connectivity(d0, d1)) is not None:
+            return conn
+        else:
+            raise RuntimeError(
+                f"Connectiivty between dimension {d0} and {d1} has not been computed."
```
jorgensd, line 101:
```suggestion
                f"Connectivity between dimension {d0} and {d1} has not been computed."
```

python/dolfinx/mesh.py
jorgensd, line 113:
```suggestion
            d0: Dimension of entities one is mapping from
            d1: Dimension of entities one is mapping to
```

python/dolfinx/mesh.py
jorgensd: I think it is overkill to check for `-1` and convert, as the return-type is rarely (if ever) used by anyone.

I didn't even know about it until I wrapped it up

python/dolfinx/mesh.py
jorgensd: No, there is no return type in C++: https://github.com/FEniCS/dolfinx/blob/202b446cd4a79a1545bfd750e24fa54b6f5f6876/cpp/dolfinx/mesh/Topology.h#L190

python/dolfinx/mesh.py
jorgensd: Yes, there doesn't seem to be a way of unpacking nicely:
https://stackoverflow.com/questions/18296035/how-to-extract-the-bits-of-larger-numeric-numpy-data-types

python/dolfinx/mesh.py
jorgensd: Do we really need to mention this?  Will anyone ever unpack the permutations (except maybe @mscroggs?)

jhale: If it's in the interface, I'd like to know how to use it.

python/dolfinx/mesh.py
```diff
@@ -65,10 +64,141 @@
...
+class Topology:
+    """Topology for a :class:`dolfinx.mesh.Mesh`"""
+
+    _cpp_object: _cpp.mesh.Topology
+
+    def __init__(self, topology: _cpp.mesh.Topology):
+        """Initialize a topology from a C++ topology.
+        Args: The C++ topology object
+        Note:
+            topology objects should not usually be created using this
```
jhale, line 76: Also leave an explicit pointer to the constructor here.

python/dolfinx/mesh.py
jhale: OK. I think we will switch to using std::optional instead of sentinel values - we can handle this then.

jhale (approved):

python/dolfinx/mesh.py
jorgensd: We only have a C++ binding for a constructor atm, so leaving that here for now.

python/dolfinx/mesh.py
jorgensd: Ive made a relevant issue #3407 and tried to amend this docstring. let me knowif it is sufficient.

python/dolfinx/mesh.py
```diff
@@ -65,10 +64,156 @@
...
+        Args:
+            dim: Topological dimension
+
+        Returns:
+            Index map for the entities of dimension `dim`.
+        """
+        if (imap := self._cpp_object.index_map(dim)) is not None:
+            return imap
+        else:
+            raise RuntimeError(f"Entities of dimension {dim} has not been computed")
```
jhale, line 176: Instruct user how to fix the issue.

python/dolfinx/mesh.py
```diff
@@ -65,10 +64,156 @@
...
+    @property
+    def entity_types(self) -> list[list[CellType]]:
+        """Get the entity types in the topology for all topological dimensions."""
+        return self._cpp_object.entity_types
+
+    def get_cell_permutation_info(self) -> npt.NDArray[np.uint32]:
+        """Returns the permutation information
+
+        The returned data is used for packing coefficients and assembling of tensors.
+        Each integer encodes the number of reflections and permutations for each sub-entity
```
jhale, line 145: The bits of each integer

python/dolfinx/mesh.py
```diff
@@ -65,10 +64,156 @@
...
+        of the cell to be able to map it to the reference element.
+        """
+        return self._cpp_object.get_cell_permutation_info()
+
+    def get_facet_permutations(self) -> npt.NDArray[np.uint8]:
+        """Get the permutation number to apply to facets.
+
+        An entry describes the number of reflections and rotations that has to
```
jhale, line 153: The bits of each integer

jhale (changes requested):

python/dolfinx/mesh.py
```diff
@@ -65,10 +64,159 @@
...
+
+    def cell_name(self) -> str:
+        """String representation of the cell-type of the topology"""
+        return to_string(self._cpp_object.cell_type)
+
+    def connectivity(self, d0: int, d1: int) -> _cpp.graph.AdjacencyList_int32:
+        """Return connectivity from entities of dimension ``d0`` to entities of dimension ``d1``.
+
+        Note:
+            Assumes only one entity type per dimension.
```
garth-wells, line 89: I don't understand this note.

jorgensd: This is taken directly from the C++ docs:
https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/mesh/Topology.h#L106-L130
and I interpret this as:
This connectivity does not take into account prism meshes or mixed meshes, as you would have multiple cell types per dimension. @chrisrichardson Feel free to correct me if I am wrong.

jhale (approved):

python/dolfinx/mesh.py
```diff
@@ -65,10 +64,159 @@
...
+        return self._cpp_object.get_facet_permutations()
+
+    def index_map(self, dim: int) -> _cpp.common.IndexMap:
+        """Get the IndexMap that described the parallel distribution of the mesh entities.
+
+        Args:
+            dim: Topological dimension
+
+        Returns:
+            Index map for the entities of dimension `dim`.
```
garth-wells, line 171:
```suggestion
            Index map for the entities of dimension ``dim``.
```

python/dolfinx/mesh.py
jhale: From a user's point of view the important point would be to mention that this connectivity is not correct for meshes containing prisms and/or mixed meshes.

garth-wells: Let's just remove the comment then. At this stage, anyone using mixed cell types is on their own . . .

python/dolfinx/mesh.py
```diff
@@ -68,6 +67,154 @@
...
+
+    def cell_name(self) -> str:
+        """String representation of the cell-type of the topology"""
+        return to_string(self._cpp_object.cell_type)
+
+    def connectivity(self, d0: int, d1: int) -> _cpp.graph.AdjacencyList_int32:
+        """Return connectivity from entities of dimension ``d0`` to entities of dimension ``d1``.
+
+        Note:
+            Assumes only one entity type per dimension.
```
jorgensd, line 92:
```suggestion
```

garth-wells (approved):

PR #3415, opened 2024-09-17
Fix cffi_utils for abscence of petsc4py

schnellerhase:
Make `cffi` type registry also work when `pets4py` is not available.

### DIFF
--- a/python/dolfinx/utils.py
+++ b/python/dolfinx/utils.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 import ctypes as _ctypes
 import os
 import pathlib
+import warnings
 
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


python/dolfinx/utils.py
```diff
@@ -195,25 +196,26 @@ def set_vals(A: int,
...
-        _ffi = _cffi.FFI()
+        # Register complex types
         _cffi_support.register_type(_ffi.typeof("float _Complex"), _numba.types.complex64)
         _cffi_support.register_type(_ffi.typeof("double _Complex"), _numba.types.complex128)
-    except (ImportError, KeyError):
+
+    except KeyError:
         pass
+    except ImportError:
+        warnings.warn("Could not import numba, so complex types are not registered!", ImportWarning)
```
jhale, line 214: Could not import numba, so cffi/numba complex types were not registered.

python/dolfinx/utils.py
```diff
@@ -248,5 +250,9 @@ def set_vals(A: int,
         """See PETSc `MatSetValuesBlockedLocal
         <https://petsc.org/release/manualpages/Mat/MatSetValuesBlockedLocal>`_
         documentation."""
-    except (ImportError, KeyError):
+    except KeyError:
         pass
+    except ImportError:
+        warnings.warn(
+            "Could not import petsc4py, so numba petsc overloads are not available!", ImportWarning
```
jhale, line 257: Could not import petsc4py, so cffi/PETSc ABI mode interface was not created.

PR #3419, opened 2024-09-18
Document `compute_midpoints` and remove outdated fixme.

jorgensd:


python/dolfinx/mesh.py
```diff
@@ -439,6 +439,16 @@ def compute_incident_entities(
 
 
 def compute_midpoints(mesh: Mesh, dim: int, entities: npt.NDArray[np.int32]):
+    """Compute the midpoints of a set of mesh entities
```
garth-wells, line 442:
```suggestion
    """Compute the midpoints of a set of mesh entities.
```

garth-wells (approved):

PR #3420, opened 2024-09-19
Try with MPI windows python tests

jhale:


jhale:
Still doesn't work @schnellerhase :( Thanks for the suggestion though!

schnellerhase:
Hmm, no luck. 

Did you try something like `python test.py` as well already? I think like this pytest gets still invoked. I only saw the asserts with my problems if I called the python file by hand, and through a package.

jhale:
I don't think so - the call `python -m pytest` is what does the magic to actually execute the tests - without it it will just execute the outer part of the file (imports, usually).

PR #3432, opened 2024-09-24
Fix mesh constructor again

jorgensd:
Fixing the same issue as: https://github.com/FEniCS/dolfinx/pull/3367/ again


garth-wells:
Can we add a test that would pick this up?

jorgensd:
> Can we add a test that would pick this up?

Added

garth-wells (approved):

Issue #3413, opened 2024-09-17
Replace sentinel arguments and return values with std::optional


jhale:
### Describe new/missing feature

`std::optional` provides a modern way to express passing or returning value or not computed.

We are currently using many sentinel function arguments and return values including `-1`, `nullptr` and empty strings that require the user to read the docstring of the function to understand the argument or return value. These can be replaced with `std::optional` which is very explicit. An additional advantage is that nanobind has support for `std::optional` which results in consistent behaviour between the Python (`typing.Optional`) and C++ interfaces and a reduction in Python wrapper code complexity.

A first step would be identifying all `typing.Optional` arguments in Python which typically leads to a sentinel in C++.

Examples of use are here:

https://github.com/FEniCS/dolfinx/pull/3328
https://github.com/FEniCS/dolfinx/pull/3322

schnellerhase:
- `set_bc`: make $x_0$ optional (this will require some more changes, but should be able to remove a quite drastic code duplication problem)
- `locate_entities_boundary` and `exterior_facet_indices` combinable - see https://github.com/FEniCS/dolfinx/pull/3283
- (`locate_dofs_topological`)

PR #3434, opened 2024-09-24
Use `std::optional` to reduce code duplication in boundary condition setting

garth-wells:
See #3413.

Also:
- Simplifies syntax, especially from C++.
- Now clearer what the Dirichet BC 'set' function does.
- Improves documentation.
- Changes the confusingly named 'scale' argument to 'alpha` in the function that modifies an array due to boundary conditions.


cpp/demo/hyperelasticity/main.cpp
```diff
@@ -86,8 +86,8 @@ class HyperElasticProblem
     return [&](const Vec x, Vec)
     {
       // Assemble b and update ghosts
-      std::span<T> b(_b.mutable_array());
-      std::ranges::fill(b, 0.0);
+      std::span b(_b.mutable_array());
+      std::ranges::fill(b, 0);
```
jhale, line 90: Tangential, but do we have a set style on this? Personally I prefer using floating point literals even if a cast from an integer literal is possible. We are also using `T(1)` below.

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -469,18 +470,29 @@ class DirichletBC
...
-  /// degrees-of-freedom that are owned by the calling rank, the length
-  /// of the array @p x should be equal to the number of dofs owned by
-  /// this rank.
-  /// @param[in] scale The scaling value to apply
-  void set(std::span<T> x, T scale = 1) const
+  /// Entries in the array `x` that are constrained by a Dirichlet
+  /// boundary conditions are set to `alpha * (x_bc - x0)`, where `x_bc`
+  /// is the boundary condition value.
+  ///
+  /// For elements with point-wise evaluated degrees-of-freedom, e.g.
```
jhale, line 480: I made this more precise and then more direct about what the user should be aware of.

`x_bc` is the value of the boundary condition interpolated into the finite element space via the degrees of freedom. Note that for many finite elements, the degrees of freedom are not straightforward point evaluations, as for the Lagrange element.

python/dolfinx/fem/bcs.py
```diff
@@ -132,7 +132,23 @@ def function_space(self) -> dolfinx.fem.FunctionSpace:
...
+        self, b: npt.NDArray, x0: typing.Optional[npt.NDArray[np.int32]] = None, alpha: float = 1
+    ) -> None:
+        """Insert boundary condition values into vector.
+
+        Only local (owned) entries are set, hence communication after
+        calling this function is not required unless ghost entries need to
+        be updated to the boundary condition value.
+
+        Args:
+            b:
```
jhale, line 145: Incomplete.

python/dolfinx/fem/bcs.py
```diff
@@ -132,7 +132,23 @@ def function_space(self) -> dolfinx.fem.FunctionSpace:
         """The function space on which the boundary condition is defined"""
         return self._cpp_object.function_space
 
-    def dof_indices(self) -> tuple[np.ndarray, int]:
+    def set(
```
jhale, line 135: Would be good to have the same docstring with stuff about x_0 etc. here.

jhale (changes requested):
Thanks for doing this! `std::optional` is great for us, we should carefully roll it out everywhere.

python/dolfinx/fem/assemble.py
```diff
@@ -344,23 +345,27 @@ def apply_lifting(
...
-    _cpp.fem.apply_lifting(b, _a, constants, coeffs, _bcs, x0, scale)
+    _cpp.fem.apply_lifting(b, _a, constants, coeffs, _bcs, x0, alpha)
 
 
 def set_bc(
     b: np.ndarray,
     bcs: list[DirichletBC],
     x0: typing.Optional[np.ndarray] = None,
-    scale: float = 1.0,
+    alpha: float = 1,
```
jorgensd, line 355: If the function is deprecated, is there a reason for redefining the keyword making it non-backwards compatible?

garth-wells: Should be changed back to `scale`.

cpp/demo/hyperelasticity/main.cpp
garth-wells: I think our house (`0.0`) style pre-dates supporting different types. I prefer
```cpp
std::ranges::fill(b, 0);
```
to
```cpp
std::ranges::fill(b, T(0));
```
because it's less syntax. I think
```cpp
std::ranges::fill(b, 0.0);
```
is possibly misleading since `T` is not necessarily `double`.

python/demo/demo_poisson_matrix_free.py
```diff
@@ -188,7 +188,7 @@ def action_A(x, y):
     y.scatter_reverse(la.InsertMode.add)
 
     # Set BC dofs to zero
-    fem.set_bc(y.array, [bc], scale=0.0)
+    bc.set(y.array, alpha=0.0)
```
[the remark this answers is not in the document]
garth-wells: We already have a function that can doe this more 'directly' (https://github.com/FEniCS/dolfinx/blob/bc88ca0336b8636024cbedb164c9a32c49e341ca/cpp/dolfinx/fem/DirichletBC.h#L557), but it had quite a bit of duplicated code.

python/dolfinx/fem/bcs.py
garth-wells: Completed.

python/dolfinx/fem/bcs.py
garth-wells: Done.

jhale (approved):

python/demo/demo_poisson_matrix_free.py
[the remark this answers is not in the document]
garth-wells: I've added an optimisation in `DirichletBC::set` for when `alpha == 0`.

PR #3439, opened 2024-09-26
Check for interprocess facets when assembling interior facet integrals

jorgensd:
A long-standing issue in DOLFINx is that we do not warn the user that using `GhostMode.none` with interior facet integrals can result in undefined behavior.
This includes:
https://github.com/FEniCS/dolfinx/issues/2556
and an attempt of addressing this was done in: https://github.com/FEniCS/dolfinx/pull/2574, but was rejected due to its non-trivial computations.

We do have `interprocess_facets`, which is a bi-product of `create_entities` which we can use to check that we have sufficient information when packing integration entities: https://github.com/FEniCS/dolfinx/blob/d5f56d66dcc561ad017b0d3fde42e59f209f82f5/cpp/dolfinx/mesh/Topology.h#L196

`interprocess_facets` are facets owned by the current process that is shared with other processes.


chrisrichardson:
https://github.com/FEniCS/dolfinx/blob/292c6234ea58c56f0efa19fc804623503cb3a3f5/cpp/dolfinx/mesh/topologycomputation.cpp#L98

Looks like it also needs updating docs

chrisrichardson (approved):

garth-wells:
Not a change is this PR, but what are "interprocess_facets"? The `mesh::Topology::interprocess_facets` is unclear.

garth-wells:
> Not a change is this PR, but what are "interprocess_facets"? The `mesh::Topology::interprocess_facets` docstring is unclear.

I've improved the docstring.

garth-wells (approved):
Nice.

PR #3442, opened 2024-09-28
Fix `create_cell_partitioner` interaction with `create_mesh`

schnellerhase:


schnellerhase:
Signature of the nanobind function looks good, still not able to pass it to `create_mesh`.

schnellerhase:
Caused by wrong dimension in cell array for the requested `create_mesh` overload. Problem with the `std::vector` version of the `create_mesh`? (In principle it should have worked with that one as well.

schnellerhase:
Calling `create_mesh` with
```python
cells = np.array([[]], dtype=np.int64)
cells = [np.array([], dtype=np.int64)]
```
both works, but 
```
cells = []
```
fails, with the weird output since in general it allows for a sequence input (and does not yield the more conclusive dimesion missmatch of numpy arrays).

schnellerhase:
This still allows for short notation calls with `cells=[[]]`.

schnellerhase:
This breaks the interaction with the mixed meshes however, which require different length arrays in the cell list. Guess that is also why the tutorial @jorgensd worked back then.

schnellerhase:
Doesn't really help with the problem.

Issue #3443, opened 2024-09-28
Fix refined meshes with ghosts


garth-wells:
At the moment, the handling of refinement of ghosted meshes is a mess. It should be easy to optionally include ghost cells in a refined mesh **without re-partitioning** by using `compute_destination_ranks(...)` in `graph/partitioners.cpp`.

This is compounded by `transfer_facet_meshtag(...)` and `transfer_cells_meshtag(...)` no supporting ghosted meshes, or re-partitioned refined meshes. Both should be relatively easy to support. This leads to unexpected 'user experiences'.

jhale:
Is this fixed by #3661?

PR #3444, opened 2024-09-28
Pass graph partitioner to `mesh::refine`

garth-wells:
Initial work by @schnellerhase. See #3443 for issue that this PR highlights (but doesn't fix). 

Includes significant documentation improvements explaining how ghosts are handled in refinement. This has also highlighted inconsistencies and limitation which should be fixed in the future.

Supersedes #3362


chrisrichardson (approved):
I don't really like `msh` but apart from that, looks fine...

PR #3445, opened 2024-09-30
Bump PETSC release 

jorgensd:
Contains a critical fix for NewtonTR in snes:
https://gitlab.com/petsc/petsc/-/issues/1645


jorgensd:
We could bump to 3.22 for both PETSC and slepc, but I would like to hold it off in case of release related bugs such as the one above.

garth-wells:
We should bump to 3.22. We don't want build failures against the latest PETSc release. Seeing any bugs sooner is better than later.

docker/Dockerfile.test-env
```diff
@@ -18,7 +18,7 @@ ARG GMSH_VERSION=4_13_1
 ARG HDF5_VERSION=1.14.4.3
 ARG KAHIP_VERSION=3.16
 ARG NUMPY_VERSION=2.1.1
-ARG PETSC_VERSION=3.21.5
+ARG PETSC_VERSION=3.21.6
 ARG SLEPC_VERSION=3.21.1
```
jorgensd, line 22:
```suggestion
ARG PETSC_VERSION=3.22.0
ARG SLEPC_VERSION=3.22.0
```

PR #3453, opened 2024-10-03
Fix bug in distribute

chrisrichardson:


jorgensd:
Could you add some context to the PR, what kind of bug and what does it affect?

chrisrichardson:
Error in calculation of `src` in dristribute leads to random SEGV.

garth-wells (approved):
Thanks for fixing!

PR #3468, opened 2024-10-11
Add demo of nonlinear scipy solver

jorgensd:
As we now support DOLFINx without PETSc, it is good to have a show-case of how to interface with scipy to solve problems.

### DIFF
--- a/python/demo/demo_scipy-solvers.py
+++ b/python/demo/demo_scipy-solvers.py
@@ -1,3 +1,18 @@
+# ---
+# jupyter:
+#   jupytext:
+#     text_representation:
+#       extension: .py
+#       format_name: light
+#       format_version: '1.5'
+#       jupytext_version: 1.13.6
+# ---
+# Copyright (C) 2024 Jørgen S. Dokken
+#
+# This file is part of DOLFINx (<https://www.fenicsproject.org>)
+#
+# SPDX-License-Identifier:    LGPL-3.0-or-later
+#
 # # $L^2$ projection through minimization using SciPy solvers
 # In this demo, we will go through how to do a `projection` of a spatially varying function `g`,
 # into a function space `V`.
@@ -18,6 +33,10 @@ import scipy.sparse.linalg
 import dolfinx
 import ufl
 
+if MPI.COMM_WORLD.size > 1:
+    print("This demo works only in serial.")
+    exit(0)
+
 # Next we define the `domain` ($\Omega$) and the function space `V` where we want to
 # project the function `g` into.
 
@@ -31,7 +50,7 @@ g = ufl.sin(ufl.pi * x) * ufl.cos(ufl.pi * y)
 # The functional $G$ can be defined as follows
 
 uh = dolfinx.fem.Function(V)
-G = ufl.inner(uh - g, uh - g) * ufl.dx
+G = ufl.dot(uh - g, uh - g) * ufl.dx
 
 G_compiled = dolfinx.fem.form(G)
 
--- a/python/demo/demo_scipy-solvers.py
+++ b/python/demo/demo_scipy-solvers.py
@@ -1,0 +1,127 @@
+# # $L^2$ projection through minimization using SciPy solvers
+# In this demo, we will go through how to do a `projection` of a spatially varying function `g`,
+# into a function space `V`.
+#
+# An $L^2$ projection can be viewed as the following minimization problem:
+#
+# $$ \min_{u\in V} G(u) = \frac{1}{2}\int_\Omega (u-g)\cdot (u-g)~\mathrm{d}x.$$
+
+# We start by importing the necessary modules
+
+from mpi4py import MPI
+
+import numpy as np
+import numpy.typing as npt
+import scipy.optimize
+import scipy.sparse.linalg
+
+import dolfinx
+import ufl
+
+# Next we define the `domain` ($\Omega$) and the function space `V` where we want to
+# project the function `g` into.
+
+Nx = 15
+Ny = 20
+domain = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, Nx, Ny)
+V = dolfinx.fem.functionspace(domain, ("Lagrange", 1))
+x, y = ufl.SpatialCoordinate(domain)
+g = ufl.sin(ufl.pi * x) * ufl.cos(ufl.pi * y)
+
+# The functional $G$ can be defined as follows
+
+uh = dolfinx.fem.Function(V)
+G = ufl.inner(uh - g, uh - g) * ufl.dx
+
+G_compiled = dolfinx.fem.form(G)
+
+# We can find the minimum of this function by differentiating
+# the function with respect to `u`:
+#
+# $$ F(u, \delta u) = \frac{\mathrm{d}G}{\mathrm{d}u}[\delta u] = 0, \quad \forall \delta u \in V.$$
+
+du = ufl.TestFunction(V)
+residual = ufl.derivative(G, uh, du)
+
+# We generate the integration kernels for the residua, and assemble an initial
+# residual vector `b`.
+
+F = dolfinx.fem.form(residual)
+b = dolfinx.fem.assemble_vector(F)
+
+# A convenience function for computing the value of the functional
+
+
+def compute_functional_value() -> dolfinx.default_scalar_type:
+    local_contribution = dolfinx.fem.assemble_scalar(G_compiled)
+    return domain.comm.allreduce(local_contribution, op=MPI.SUM)
+
+
+# We can use the scipy Newton-Krylov solver to solve this problem.
+# This solver takes in a function that computes the residual of the functional
+# We re-use our structures for `uh` and `b` to avoid unnecessary allocations.
+
+
+def compute_residual(x) -> npt.NDArray[dolfinx.default_scalar_type]:
+    """
+    Evaluate the residual F(x) = 0
+
+    :param x: Input vector with current solution
+    :return: Returns residual array
+    """
+    uh.x.array[:] = x
+    b.array[:] = 0
+    dolfinx.fem.assemble_vector(b.array, F)
+    b.scatter_reverse(dolfinx.la.InsertMode.add)
+    return b.array
+
+
+uh.x.array[:] = 0
+print("Jacobian free Newton Krylov")
+print(f"Initial functional {compute_functional_value():.2e}")
+solution = scipy.optimize.newton_krylov(compute_residual, uh.x.array, method="lgmres", verbose=True)
+uh.x.array[:] = solution
+print(f"Minimal functional {compute_functional_value():.2e}")
+
+# We could also exploit the Jacobian information and use a Newton method with an exact gradient
+#
+#  $$ J(u, \delta u, \delta v) = \frac{\mathrm{d} F}{\mathrm{d}u}[\delta u, \delta v] =
+# \frac{\mathrm{d}^2G}{\mathrm{d}u^2}[\delta u, \delta v].$$
+
+dv = ufl.TrialFunction(V)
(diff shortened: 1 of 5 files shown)


python/demo/demo_scipy-solvers.py
```diff
@@ -0,0 +1,127 @@
...
+
+# If the Jacobian is independent of uh, we can avoid re-assembling the matrix
+
+is_nonlinear = uh._cpp_object in J_compiled._cpp_object.coefficients
+
+# We next create a function for solving the Krylov subspace problem
+# where we use the Jacobian and scipy's sparse LU solver to solve the linear system.
+
+
+def solve_system(_A, x, rtol: float, maxiter: int, M) -> tuple[npt.NDArray[np.float64], int]:
```
jorgensd, line 109:
```suggestion
def solve_system(_A, x, **kwargs) -> tuple[npt.NDArray[np.float64], int]:
```

python/demo/demo_scipy-solvers.py
```diff
@@ -0,0 +1,127 @@
...
+G = ufl.inner(uh - g, uh - g) * ufl.dx
+
+G_compiled = dolfinx.fem.form(G)
+
+# We can find the minimum of this function by differentiating
+# the function with respect to `u`:
+#
+# $$ F(u, \delta u) = \frac{\mathrm{d}G}{\mathrm{d}u}[\delta u] = 0, \quad \forall \delta u \in V.$$
+
+du = ufl.TestFunction(V)
```
jorgensd, line 43:
```suggestion
du = ufl.conj(ufl.TestFunction(V))
```

python/demo/demo_scipy-solvers.py
```diff
@@ -0,0 +1,127 @@
...
+
+# We can find the minimum of this function by differentiating
+# the function with respect to `u`:
+#
+# $$ F(u, \delta u) = \frac{\mathrm{d}G}{\mathrm{d}u}[\delta u] = 0, \quad \forall \delta u \in V.$$
+
+du = ufl.conj(ufl.TestFunction(V))
+residual = ufl.derivative(G, uh, du)
+
+# We generate the integration kernels for the residua, and assemble an initial
```
nate-sime, line 46:
```suggestion
# We generate the integration kernels for the residual, and assemble an initial
```

python/demo/demo_scipy-solvers.py
```diff
@@ -0,0 +1,146 @@
...
+# This solver takes in a function that computes the residual of the functional
+# We re-use our structures for `uh` and `b` to avoid unnecessary allocations.
+
+
+def compute_residual(x) -> npt.NDArray[dolfinx.default_scalar_type]:
+    """
+    Evaluate the residual F(x) = 0
+
+    :param x: Input vector with current solution
+    :return: Returns residual array
```
nate-sime, line 89: Doesn't FEniCS use the Google style for docs? Also does the docstring really matter for a function in a demo?
```suggestion
    Args:
        x: Input vector with current solution
    Returns:
        Residual array
```

jorgensd: It’s good practice to document I guess.

garth-wells:
We do already have a demo that uses a SciPy solver: https://github.com/FEniCS/dolfinx/blob/main/python/demo/demo_types.py

jorgensd:
> We do already have a demo that uses a SciPy solver: https://github.com/FEniCS/dolfinx/blob/main/python/demo/demo_types.py

We only show how to solve linear problems.
I am happy to merge this one into that one.

jorgensd:
I at least think we should make it clearer from the demo title that it shows how to use scipy. 
I guess if we make a separate demo repository we could add more descriptive documentation guiding the user.

python/demo/demo_scipy-solvers.py
```diff
@@ -0,0 +1,148 @@
...
+#       jupytext_version: 1.13.6
+# ---
+# Copyright (C) 2024 Jørgen S. Dokken
+#
+# This file is part of DOLFINx (<https://www.fenicsproject.org>)
+#
+# SPDX-License-Identifier:    LGPL-3.0-or-later
+#
+# # $L^2$ projection through minimization using SciPy solvers
+# In this demo, we will go through how to do a `projection` of a spatially varying function `g`,
```
jhale, line 17: Wrong quotes on `projection`?

python/demo/demo_scipy-solvers.py
```diff
@@ -0,0 +1,148 @@
...
+import numpy as np
+import numpy.typing as npt
+import scipy.optimize
+import scipy.sparse.linalg
+
+import dolfinx
+import ufl
+
+if MPI.COMM_WORLD.size > 1:
+    print("This demo works only in serial.")
```
jhale, line 37: Would be good to say why, at least in a comment, because there is no issue until you use scipy sparse objects.

python/demo/demo_scipy-solvers.py
```diff
@@ -0,0 +1,148 @@
...
+
+    Args:
+        x: Input vector with current solution
+    Returns:
+        Residual array
+    """
+    uh.x.array[:] = x
+    b.array[:] = 0
+    dolfinx.fem.assemble_vector(b.array, F)
+    b.scatter_reverse(dolfinx.la.InsertMode.add)
```
jhale, line 96: There is no lifting applied here - is this still necessary?

python/demo/demo_scipy-solvers.py
```diff
@@ -0,0 +1,148 @@
...
+    return b.array
+
+
+uh.x.array[:] = 0
+print("Jacobian free Newton Krylov")
```
jhale, line 101: I don't consider approximating the Jacobian best practice - best related to the end as an aside, or removed entirely?

python/demo/demo_scipy-solvers.py
```diff
@@ -0,0 +1,148 @@
...
+# We assemble the Jacobian matrix and its inverse
+
+J_compiled = dolfinx.fem.form(jacobian)
+A = dolfinx.fem.assemble_matrix(J_compiled)
+A_scipy = A.to_scipy()
+Ainv = scipy.sparse.linalg.splu(A_scipy)
+
+# If the Jacobian is independent of uh, we can avoid re-assembling the matrix
+
+is_nonlinear = uh._cpp_object in J_compiled._cpp_object.coefficients
```
jhale, line 124: I don't like this - can this be worked out from the UFL form?

Even if it is possible to do it via UFL, would it not be better to solve a non-linear problem in the first place?

python/demo/demo_scipy-solvers.py
```diff
@@ -0,0 +1,148 @@
...
+
+# We next create a function for solving the Krylov subspace problem
```
jhale, line 126: Arguably there is no "Krylov" here if you use LU decomposition on the inner solve.

python/demo/demo_scipy-solvers.py
```diff
@@ -0,0 +1,148 @@
...
+# where we use the Jacobian and scipy's sparse LU solver to solve the linear system.
+
+
+def solve_system(_A, x, **kwargs) -> tuple[npt.NDArray[np.float64], int]:
+    """Compute the Jacobian and solve the linear system Ax"""
```
jhale, line 131: This docstring is wrong. This would be better called something like "inverse_jacobian_action" and then the docstring is "Apply the action of the inverse Jacobian y = A^{-1} x". I would also explicitly define `y` rather than returning the solve directly.

python/demo/demo_scipy-solvers.py
```diff
@@ -0,0 +1,148 @@
...
+#  $$ J(u, \delta u, \delta v) = \frac{\mathrm{d} F}{\mathrm{d}u}[\delta u, \delta v] =
+# \frac{\mathrm{d}^2G}{\mathrm{d}u^2}[\delta u, \delta v].$$
+
+dv = ufl.TrialFunction(V)
+jacobian = ufl.derivative(residual, uh, dv)
+
+# We assemble the Jacobian matrix and its inverse
+
+J_compiled = dolfinx.fem.form(jacobian)
+A = dolfinx.fem.assemble_matrix(J_compiled)
```
jhale, line 118: Link between `A` and `A_scipy` should be made explicit - subtle point.

jhale (changes requested):
Not withstanding issue on where to put this raised by @garth-wells, here are my comments.

python/demo/demo_scipy-solvers.py
jorgensd, line 17:
```suggestion
# In this demo, we will go through how to do a "projection" of a spatially varying function `g`,
```

python/demo/demo_scipy-solvers.py
jorgensd, line 37:
```suggestion
# This demo uses scipy to solve the projection problem, which does not support matrices and vectors
# distributed with MPI.

if MPI.COMM_WORLD.size > 1:
    print("This demo works only in serial.")
```

python/demo/demo_scipy-solvers.py
jorgensd: As the code runs in serial it is not needed at all, but as the residual computation in theory could run in parallel, I think we should always do the correct accumulation of values from each process.

python/demo/demo_scipy-solvers.py
jorgensd: I don't have a strong opinion on keep this or not. Can remove it.

python/demo/demo_scipy-solvers.py
```diff
@@ -0,0 +1,148 @@
...
+
+# We could also exploit the Jacobian information and use a Newton method with an exact gradient
+#
+#  $$ J(u, \delta u, \delta v) = \frac{\mathrm{d} F}{\mathrm{d}u}[\delta u, \delta v] =
+# \frac{\mathrm{d}^2G}{\mathrm{d}u^2}[\delta u, \delta v].$$
+
+dv = ufl.TrialFunction(V)
+jacobian = ufl.derivative(residual, uh, dv)
+
+# We assemble the Jacobian matrix and its inverse
```
jorgensd, line 115:
```suggestion
# We assemble the Jacobian matrix and its inverse.
# Note that we assemble the Jacobian into a :class:`dolfinx.la.MatrixCSR`.
# Then, we get a view of the data in this matrix by using
# :func:`dolfinx.la.MatrixCSR.to_scipy`.
```

python/demo/demo_scipy-solvers.py
jorgensd: Comment added above

python/demo/demo_scipy-solvers.py
jorgensd: Done above

python/demo/demo_scipy-solvers.py
```diff
@@ -0,0 +1,156 @@
...
+# :func:`dolfinx.la.MatrixCSR.to_scipy`.
+
+J_compiled = dolfinx.fem.form(jacobian)
+A = dolfinx.fem.assemble_matrix(J_compiled)
+A_scipy = A.to_scipy()
+Ainv = scipy.sparse.linalg.splu(A_scipy)
+
+# If the Jacobian is independent of uh, we can avoid re-assembling the matrix
+
+is_nonlinear = uh in ufl.algorithms.expand_derivatives(jacobian).coefficients()
```
jorgensd, line 130: @jhale I've here used `ufl to figure it out.
I think it is a quite nice point to make, as some people prefer always using a non-linear solver, but could benefit alot from not having to reassemble J at every step of the Newton solver (even if it is going to be just 1).

jorgensd:
I'm closing this, as I've covered how to properly interface with scipy using a pre-only approach at https://jsdokken.com/FEniCS-workshop/src/form_compilation/alternate_form.html#using-scipy-s-newton-solver

PR #3471, opened 2024-10-14
Allow promotion of geometry dimension for built-in mesh creators.

jhale:
There is currently no easy way to create e.g. a unit interval living in 2d or
3d geometric space.

After discussion of this implementation I will implement the others the same way.


jhale:
I will now do the work on the C++ interface to get this parameter up to `Geometry`.

cpp/dolfinx/refinement/refine.h
```diff
@@ -72,7 +72,7 @@ refine(const mesh::Mesh<T>& mesh,
 
   mesh::Mesh<T> mesh1 = mesh::create_mesh(
       mesh.comm(), mesh.comm(), cell_adj.array(), mesh.geometry().cmap(),
-      mesh.comm(), new_vertex_coords, xshape, partitioner);
+      mesh.comm(), new_vertex_coords, xshape, mesh.geometry().dim(), partitioner);
```
jorgensd, line 75: Shouldn’t xshape[1] and mesh.geometry().dim() be the same?

jhale: Yes - but in terms of the code being correct in meaning, the former is the shape of the geometry data and the latter is the geometric dimension.

jorgensd: Do we check if they match? We should probably add some sanity checks.

jhale:
Will have to look at this post-release - it would still be very useful to have "affine domain" non-affine mesh so that we can push-forward an affine domain with an analytical map with higher-order accuracy.

PR #3477, opened 2024-10-22
Saddle point demo added

niravshah241:


nate-sime:
Are saddle point problems not already covered in the demos? For example:
[demo_mixed-poisson.py](https://github.com/FEniCS/dolfinx/blob/main/python/demo/demo_mixed-poisson.py)
[demo_navier-stokes.py](https://github.com/FEniCS/dolfinx/blob/main/python/demo/demo_navier-stokes.py)
[demo_stokes.py](https://github.com/FEniCS/dolfinx/blob/main/python/demo/demo_stokes.py)
[demo_static-condensation.py](https://github.com/FEniCS/dolfinx/blob/main/python/demo/demo_static-condensation.py)
[demo_hdg.py](https://github.com/FEniCS/dolfinx/blob/main/python/demo/demo_hdg.py)

garth-wells:
I would suggest adding an iterative solver to [demo_mixed-poisson.py](https://github.com/FEniCS/dolfinx/blob/main/python/demo/demo_mixed-poisson.py) rather than a new demo.

Issue #3461, opened 2024-10-09
Replace Boost Timer with `std::chrono`


garth-wells:
We've observed accuracy issue with Boost Timer on some systems, plus removing it in favour of `std::chrono` (https://en.cppreference.com/w/cpp/header/chrono) is one less dependency.

IgorBaratta:
I think there's also the issue that the timer is stored in seconds.
Maybe this conversion should be left to the end when we report timings.


https://github.com/FEniCS/dolfinx/blob/c5644696862635d02193d91b8dd147ea4e385c1d/cpp/dolfinx/common/Timer.cpp#L50

https://github.com/FEniCS/dolfinx/blob/c5644696862635d02193d91b8dd147ea4e385c1d/cpp/dolfinx/common/Timer.cpp#L43

jpdean:
I checked the boost timer directly (`elapsed.wall`), and it doesn't seem to be the conversion that's causing the problem. The accuracy of the timer itself seems much worse than `std::chrono`, but maybe I'm missing something...

IgorBaratta:
> I checked the boost timer directly (`elapsed.wall`), and it doesn't seem to be the conversion that's causing the problem. The accuracy of the timer itself seems much worse than `std::chrono`, but maybe I'm missing something...

I mean it more in general.  Rounding to seconds with floating-point numbers can lead to accumulated rounding errors when many time values are summed over time. Since we more often measure times in the order micro seconds than days.

PR #3478, opened 2024-10-22
Update timer to use `std::chrono`

schnellerhase:
Fixes https://github.com/FEniCS/dolfinx/issues/3461

Timings are now done with `std::chrono`, the `Timer` is now a template over the clock type to use, by default `std::chrono::high_resolution_clock` which is the most precise a system can provide. 

The time may then be extracted in any desired unit with the call to `stop()`, by default microseconds. This allows to use a time scale suitable for the benchmarked code.

Also this removes the differentiation of wall, system and user times and all related functionality is removed.

TODO:
- [x] Reactivate logging
- [x] Remove boost dependency


cpp/dolfinx/common/Timer.h
```diff
@@ -28,36 +28,40 @@ namespace dolfinx::common
...
+  void start() { _start_time = chrono_timer::now(); }
 
   /// Stop timer, return wall time elapsed and store timing data into
   /// logger
-  double stop();
-
-  /// Return wall, user and system time in seconds
-  std::array<double, 3> elapsed() const;
+  template <typename unit = std::chrono::nanoseconds>
+  double stop()
```
garth-wells, line 50: It would be neat to return `std::chrono::duration`. The user can then handle it. `nanobind` can turn it into a Python object.

cpp/dolfinx/common/Timer.h
```diff
@@ -28,36 +28,40 @@ namespace dolfinx::common
...
 
   /// Destructor
-  ~Timer();
+  ~Timer() = default;
 
   /// Zero and start timer
-  void start();
-
-  /// Resume timer. Not well-defined for logging timer
-  void resume();
```
garth-wells, line 47: Any reason for removing this?

garth-wells (commented):
Thanks for picking this up!

Should be have a function to get the current duration from a running timer?

cpp/dolfinx/common/Timer.h
schnellerhase: Yes - the handling of the time points of `std::chrono` are not straight forward to extend to this case. We store the initial `start_time` and then compute the time when `stop` is called by the difference of current time to it. However if we would want to restart we need to store both an elapsed time and the new start point to maintain that property (or do something nasty and move the start time back by the already elapsed time). As it wasn't used anywhere I thought it would be easier to just remove it and not deal with that.

schnellerhase:
The logging functionality currently relies on wall, user and system times. These concepts are not one to one transferable to `std::chrono`. We get a lot more control about how we collect the timings though, so we can also decide on an appropriate output to the log.

How about we remove wall, user, and system time in favor of the single timing of `high_resolution_clock` in seconds then? Any objections?

garth-wells:
> The logging functionality currently relies on wall, user and system times. These concepts are not one to one transferable to `std::chrono`. We get a lot more control about how we collect the timings though, so we can also decide on an appropriate output to the log.
> 
> How about we remove wall, user, and system time in favor of the single timing of `high_resolution_clock` in seconds then? Any objections?

That’s fine.

jhale:
I see the default on elapsed is nanoseconds - it seems a bit excessive as a default, microseconds might be better?

schnellerhase:
Should we remove the `TimeLogManager` altogether? Looks like the only thing it provides is the singleton interface to `TimeLogger`, which we can make part of it as well.

jhale:
You could potentially move to a Monostate design pattern https://wiki.c2.com/?MonostatePattern over the current singleton design.

jhale:
I had another thought: it would be good if we could massively simplify TimeManager so that it just uses e.g. std::map to output data. This would then allow easy serialisation through all of the packages that support dictionaries and maps for saving to disk.

garth-wells:
Let’s just switch to chrono in this PR (to fix an immediate issue) and look at any other changes in new PRs.

schnellerhase:
The removal of the cast to a certain `unit`, i.e. `std::chrono::duration` will make the interface to the Timer more complicated. For example `std::chrono::high_resolution_clock` uses a duration measured in cpu ticks, so calling `timer.elapsed.count()` will yield a tick count as output. Making it again necessary to process this further. I think having a good time unit as default to which we cast (also for consistent output of tests) is a good thing to have in place. This does not limit us from using the standard duration if necessary, calling `stimer.elapsed<std::chrono::high_resolution_clock::duration>()` will result in the current behavior again.

garth-wells:
This PR did the heavy lifting. There were a few issues with logging elapsed times which are resolved in #3484, which built off this PR.

PR #3484, opened 2024-10-27
Revised `Timer`, now using `std::chrono`

### LINKED ISSUE #3461: Replace Boost Timer with `std::chrono`

garth-wells:
Builds on the implementation #3478 by @schnellerhase. Fixes some issues with elapsed time logging and includes some documentation enhancements.

Resolves #3461.


cpp/dolfinx/common/TimeLogger.h
```diff
@@ -35,29 +33,26 @@ class TimeLogger
...
   /// system time) for given task.
-  std::tuple<int, double, double, double> timing(std::string task);
+  std::pair<int, double> timing(std::string task) const;
 
 private:
   // List of timings for tasks, map from string to (num_timings,
-  // total_wall_time, total_user_time, total_system_time)
-  std::map<std::string, std::tuple<int, double, double, double>> _timings;
+  // total_wall_time)
+  std::map<std::string, std::pair<int, double>> _timings;
```
jhale, line 56: Looks good. Can we have an accessor that allows access to this data (perhaps a copy?) directly from C++ and Python? Table always seemed very cumbersome from Python when working with tools like Pandas to do post processing.

cpp/dolfinx/common/TimeLogger.h
```diff
@@ -35,29 +33,26 @@ class TimeLogger
...
-                    Table::Reduction reduction);
+  void list_timings(MPI_Comm comm, Table::Reduction reduction) const;
 
-  /// Return timing
+  /// @brief Return timing.
   /// @param[in] task The task name to retrieve the timing for
   /// @returns Values (count, total wall time, total user time, total
   /// system time) for given task.
-  std::tuple<int, double, double, double> timing(std::string task);
+  std::pair<int, double> timing(std::string task) const;
```
jhale, line 51: See my comment below. A pluralised version that returns the entire std::map would be useful.

PR #3487, opened 2024-10-29
Allow access to logged timer data

garth-wells:


jorgensd:
@garth-wells 
Seems like the `timing` function is broken, as the following doesn't work on nightly
```python
import dolfinx
import time

task = "test"
with dolfinx.common.Timer(task) as timer:
    time.sleep(0.1)

print(dolfinx.common.timing(task))
```
yielding
```bash
  File "/root/shared/debug.py", line 8, in <module>
    print(dolfinx.common.timing(task))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/dolfinx-real/lib/python3.12/dist-packages/dolfinx/common.py", line 48, in timing
    return _cpp.common.timing(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: No timings registered for task "test".
```
Not sure how the tests passed with this.

garth-wells:
> @garth-wells Seems like the `timing` function is broken, as the following doesn't work on nightly
> 
> ```python
> import dolfinx
> import time
> 
> task = "test"
> with dolfinx.common.Timer(task) as timer:
>     time.sleep(0.1)
> 
> print(dolfinx.common.timing(task))
> ```
> 
> yielding
> 
> ```shell
> Traceback (most recent call last):
>   File "/root/shared/debug.py", line 8, in <module>
>     print(dolfinx.common.timing(task))
>           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
>   File "/usr/local/dolfinx-real/lib/python3.12/dist-packages/dolfinx/common.py", line 48, in timing
>     return _cpp.common.timing(task)
>            ^^^^^^^^^^^^^^^^^^^^^^^^
> RuntimeError: No timings registered for task "test".
> ```
> 
> Not sure how the tests passed with this.

See #3488.

cpp/dolfinx/common/timing.h
```diff
@@ -27,10 +27,13 @@ Table timings();
...
-/// @brief Return timing (count, total wall time, total user time, total
-/// system time) for given task.
+/// @brief Return timing (count, total wall time) for given task.
 /// @param[in] task Name of a task
 /// @return The (count, total wall time) for the task.
 std::pair<std::size_t, double> timing(std::string task);
 
+/// @brief Logged elapsed times.
+/// @return Elapsed [task id: (count, total wall time)].
+std::map<std::string, std::pair<int, double>> timings();
```
jhale, line 37: There is a discrepancy in the use of `std::size_t` and `int` between `timing` and `timings`.

jhale (changes requested):

jorgensd:
Could we add docs to the timing routine regarding flush ? Currently the Python routine is undocumented.

jhale (approved):

PR #3488, opened 2024-10-29
Fixes to `Timer` logic

garth-wells:
The logic behind the interaction between `Timer` and `TimeLogger` was messed up. This attempts to fix it without interface changes.

Improves and fixes tests.

### DIFF
--- a/cpp/dolfinx/common/Timer.h
+++ b/cpp/dolfinx/common/Timer.h
@@ -98,6 +98,15 @@ public:
     return _acc;
   }
 
+  /// @brief Resume a stopped timer.
+  ///
+  /// Does nothing if timer has not been stopped.
+  void resume()
+  {
+    if (!_start_time.has_value())
+      _start_time = T::now();
+  }
+
   /// @brief Flush timer duration to the logger.
   ///
   /// Timer can be flushed only once.
--- a/cpp/dolfinx/common/Timer.h
+++ b/cpp/dolfinx/common/Timer.h
@@ -10,6 +10,7 @@
 #include "TimeLogManager.h"
 #include <chrono>
 #include <optional>
+#include <stdexcept>
 #include <string>
 
 namespace dolfinx::common
--- a/cpp/dolfinx/common/Timer.h
+++ b/cpp/dolfinx/common/Timer.h
@@ -50,17 +50,12 @@ public:
   /// registered in the logger.
   ~Timer()
   {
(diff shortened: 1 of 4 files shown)


jorgensd:
Shouldn't "stop" imply a "flush" if we want to maintain the timing API?

There isn't a way to "pause" a timer, as "start" resets the accumulated time. Thus stop should imply flush.

garth-wells:
> Shouldn't "stop" imply a "flush" if we want to maintain the timing API?
> 

The logic has always been that durations are logged with the Timer is destroyed. `flush` forces this to happen earlier. 

> There isn't a way to "pause" a timer, as "start" resets the accumulated time. Thus stop should imply flush.

There was a `resume` - @schnellerhase removed it because it was untested. I could now be added back (I think it should).

garth-wells:
The Python wrappers could call `stop` and `flush`.

schnellerhase:
> There was a `resume` - @schnellerhase removed it because it was untested. I could now be added back (I think it should).

This also allowed for one less state to manage in the Timer and thus simpler logic. I think keeping the states to a minimum is best here.

What do we need the `resume/restart` functionality for? Together with the `TimeLogger` I think we can do without.

jorgensd:
> > There was a `resume` - @schnellerhase removed it because it was untested. I could now be added back (I think it should).
> 
> This also allowed for one less state to manage in the Timer and thus simpler logic. I think keeping the states to a minimum is best here.
> 
> What do we need the `resume/restart` functionality for? Together with the `TimeLogger` I think we can do without.

If we use the `TimeLogger` for resuming timings, then stop should definitely flush, as there is no clear benefit of keeping the object around then?

garth-wells:
> > > There was a `resume` - @schnellerhase removed it because it was untested. I could now be added back (I think it should).
> > 
> > 
> > This also allowed for one less state to manage in the Timer and thus simpler logic. I think keeping the states to a minimum is best here.
> > What do we need the `resume/restart` functionality for? Together with the `TimeLogger` I think we can do without.
> 
> If we use the `TimeLogger` for resuming timings, then stop should definitely flush, as there is no clear benefit of keeping the object around then?

There are users who use stop/resume. It wasn't covered by tests (it is now), but that doesn't mean it's not used.

Timings should only be logged to when finished (by the destructor, or when manually flushed), otherwise the counter get messed up.

Issue #3480, opened 2024-10-24
Timer improvements


schnellerhase:
Collection of possible improvements discussed in https://github.com/FEniCS/dolfinx/pull/3478:

- [ ] `TimeLogManager` solely exists to allow for singleton usage of `TimeLogger` - merge with it? As a Monostate as suggested by @jhale (https://wiki.c2.com/?MonostatePattern)?
- [x] `TimeManager` output as standalone `std::map` - should allow for easy serialization @jhale 
- [x] functionality to retrieve current elapsed time from timer without stopping it @garth-wells

PR #3494, opened 2024-11-02
Remove `TimeLogManager` and make `TimeLogger` singleton

schnellerhase:
Fixes https://github.com/FEniCS/dolfinx/issues/3480


garth-wells:
What's the advantage of a monostate design over a singleton? Looks to me to all intents and purposes to be the same thing.

schnellerhase:
I think in this case it results in the same. We can decide on the interface we prefer, rather calling some kind of an `instance` function in the singleton approach or the constructor as here with the monostate.

schnellerhase:
Shall we stick to the monostate or return to the singleton approach, just within `TimeLogger` now?

jhale:
After reading the comments I would recommend singleton - as @garth-wells says there is little clear advantage, and singleton is at least well-known (and people understand the issues).

PR #3495, opened 2024-11-02
cpp to MPI type mapping improvements

schnellerhase:
`mpi_type` is redesigned in favor of a type trait based system that is easier to extend. Especially this no longer requires a function call to obtain the mpi type, the interface change is
```cpp
dolfinx::MPI::mpi_type<T>() -> dolfinx::MPI::mpi_t<T>
````

Additionally the mapping is extended to previously non matched types. 

Oddly the mapping of the fixed width `std` types to their MPI equivalent, for example `std::int32_t -> MPI_INT32_T` seems to break on some platforms. No clue why, because using the standard `C` type namings works, so are the `MPI` types not pointing to the same ones?

### CODE REVIEW

**Comment by jhale** [comment]
Could you be more specific about which platforms the fixed width conversion fails? And what is the compile error?

**Comment by jhale** [comment]
uint_8 is unsigned char on some platforms see e.g. https://en.cppreference.com/w/cpp/types/integer leading to the error with the redefinition

**Comment by schnellerhase** [comment]
Yes, just missed that one to revert :)

The errors that you see now in the CI were the one I discovered. Locally I couldn't reproduce it, but I also didn't check other versions. 

`long long unsigned int` is somehow not mached here.

**Comment by schnellerhase** [comment]
C standard seems to only guarantee at least 64 bit width of `unsigned long long` not exactly equal, see https://en.cppreference.com/w/cpp/language/types.

However I couldn't find anything hinting a non 64-bit width on these platforms.

File: cpp/dolfinx/common/IndexMap.cpp [hunk 1]
```diff
@@ -116,8 +116,10 @@ communicate_ghosts_to_owners(MPI_Comm comm, std::span<const int> src,
     send_sizes.reserve(1);
     recv_sizes.reserve(1);
     MPI_Request sizes_request;
-    MPI_Ineighbor_alltoall(send_sizes.data(), 1, MPI_INT32_T, recv_sizes.data(),
-                           1, MPI_INT32_T, comm0, &sizes_request);
+    MPI_Ineighbor_alltoall(send_sizes.data(), 1,
+                           dolfinx::MPI::mpi_t<std::int32_t>, recv_sizes.data(),
+                           1, dolfinx::MPI::mpi_t<std::int32_t>, comm0,
```
garth-wells, line 121: I'm not convinced that `dolfinx::MPI::mpi_t<std::int32_t>` adds anything over `MPI_INT32_T`. The latter is simpler.

`dolfinx::MPI::mpi_t` is helpful where we have templated over the type being communicated.

schnellerhase:
> For integers, to avoid any ambiguity I'd suggest providing interfaces only for fixed-width integers with the exception of int. We shouldn't (and don't) use long int, long long int, etc, in DOLFINx.

Agreed, if we can get it to work I'd be happy to see the fixed width version used as well. My understanding was always that the fixed width types of `std` are just replaced with the corresponding C-type on the machine, so if for example `unsigned long long` is 64 bit, `std::uint64_t` would just be a different name for it. 

Its most certainly caused by this type def https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/graph/partitioners.cpp#L622 - but we need it as Kahip expects that type https://github.com/KaHIP/KaHIP/blob/master/parallel/parallel_src/interface/parhip_interface.h.

Anybody got an idea whats causing this?

cpp/dolfinx/common/IndexMap.cpp
schnellerhase: Not the perfect interface for sure, but it allows us to stick to one type system in the code base.

We could also go for something like `dolfinx::MPI::mpi_t<sendsizes::value_type>`, which removes the possibility of a type mismatch, but increases code complexity for the caller.

jhale:
Regarding the kahip interface, does anyone remember why we are we using unsigned long long there?

schnellerhase:
Don't think there is any other interface, at least it is the only occurrence of `ParHIPPartitionKWay` in the repo 😄

schnellerhase:
Couldn't get the `long long unsigned int` to work on all platforms. Switched to the fixed width overloads and made the single occasion of the `long long unsigned int` for Kahip explicit with `MPI_UNSIGNED_LONG_LONG`

jhale:
This is the right approach if KaHiP is really using `long long unsigned` underneath. Still odd they don't provide a native C++ call...

Looks good to me. @garth-wells you still had an open review comment about whether it's best just to use the MPI types directly in some cases. Could you comment further on that now?

We should consider fixed width floats when we move to C++23 https://en.cppreference.com/w/cpp/types/floating-point

cpp/dolfinx/graph/partitioners.cpp
```diff
@@ -631,8 +631,8 @@ graph::partition_fn graph::kahip::partitioner(int mode, int seed,
     common::Timer timer1("KaHIP: build adjacency data");
     std::vector<T> node_disp(dolfinx::MPI::size(comm) + 1, 0);
     const T num_local_nodes = graph.num_nodes();
-    MPI_Allgather(&num_local_nodes, 1, dolfinx::MPI::mpi_type<T>(),
-                  node_disp.data() + 1, 1, dolfinx::MPI::mpi_type<T>(), comm);
+    MPI_Allgather(&num_local_nodes, 1, MPI_UNSIGNED_LONG_LONG,
```
jhale, line 634: Could you stick a comment here about the specific use of unsigned long long in Kahn.

cpp/dolfinx/common/MPI.h
```diff
@@ -269,39 +269,44 @@ struct dependent_false : std::false_type
...
-}
+struct mpi_type_mapping;
+
+/// @brief Retrieves the MPI data type associated to the provided type.
+/// @tparam T cpp type to map
+template <typename T>
+MPI_Datatype mpi_t = mpi_type_mapping<T>::type;
+
+/// @brief Registers for cpp_t the correpsonding mpi_t which can then be
+/// retrieved with mpi_t<cpp_t> form here on.
```
jhale, line 283: form -> from

cpp/dolfinx/common/MPI.h
```diff
@@ -269,39 +269,44 @@ struct dependent_false : std::false_type
...
-    static_assert(!std::is_same_v<T, T>);
-}
+struct mpi_type_mapping;
+
+/// @brief Retrieves the MPI data type associated to the provided type.
+/// @tparam T cpp type to map
+template <typename T>
+MPI_Datatype mpi_t = mpi_type_mapping<T>::type;
+
+/// @brief Registers for cpp_t the correpsonding mpi_t which can then be
```
jhale, line 282: corresponding -> corresponding

cpp/dolfinx/common/MPI.h
```diff
@@ -269,39 +269,44 @@ struct dependent_false : std::false_type
...
+  template <>                                                                  \
+  struct mpi_type_mapping<cpp_t>                                               \
+  {                                                                            \
+    static inline MPI_Datatype type = mpi_t;                                   \
+  };
+
+/// @defgroup MPI type mappings
+/// @{
+/// @cond
+MAP_TO_MPI_TYPE(char, MPI_CHAR)
```
jhale, line 294: I would consider removing this if we don't need it.

cpp/dolfinx/common/MPI.h
```diff
@@ -269,39 +269,44 @@ struct dependent_false : std::false_type
...
+MAP_TO_MPI_TYPE(bool, MPI_C_BOOL)
```
jhale, line 295: Same comment - remove if we don't need it.

IgorBaratta:
> This is the right approach if KaHiP is really using `long long unsigned` underneath. Still odd they don't provide a native C++ call...

https://github.com/KaHIP/KaHIP/blob/master/parallel/parallel_src/interface/parhip_interface.h#L15
KaHiP uses `unsigned long long` for its interface and also internally.

PR #3499, opened 2024-11-06
Demo updates for LU solver bugs

garth-wells:
MUMPS and SuperLU_dist can be shaky for less common types, i.e. other than `float64` and `int32`. This change picks a solver that works for each of the combinations tested on the CI.


python/test/unit/fem/test_petsc_solver_wrappers.py
```diff
@@ -39,10 +39,10 @@ def test_compare_solvers(self, mode):
         sys = PETSc.Sys()
         if MPI.COMM_WORLD.size == 1:
             factor_type = "petsc"
+        if sys.hasExternalPackage("mumps"):
```
francesco-ballarin, line 42: should probably be an `elif`

PR #3503, opened 2024-11-09
Remove loguru and old pugixml references

jorgensd:
Files no longer exist.


jhale:
Looks perfectly reasonable, will merge.

PR #3504, opened 2024-11-09
Expose set original cell index to python

jorgensd:
To be able to make a valid geometry from Python (without using create_topology, as one wants to avoid re-distribution of data), one has to expose the `original_cell_index` attribute to Python.
As we have used this as a public variable in C++, I've gone down the route of creating a setter function in the Python binding, rather than exposing a "list-of-lists" to Python, as we would rather like it to be numpy based.

### DIFF
--- a/python/dolfinx/wrappers/mesh.cpp
+++ b/python/dolfinx/wrappers/mesh.cpp
@@ -612,24 +612,7 @@ void mesh(nb::module_& m)
           nb::rv_policy::reference_internal)
       .def_prop_ro("dim", &dolfinx::mesh::Topology::dim,
                    "Topological dimension")
-      .def(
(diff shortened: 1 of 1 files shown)


garth-wells:
Mixed topology support is still highly experimental and under review, so we shouldn’t expose the mixed topology version of original_cell_index yet.

When it’s ready, no need to add a function with ‘set’ in the name. Could be a nanobind read/write property.

jorgensd:
> Mixed topology support is still highly experimental and under review, so we shouldn’t expose the mixed topology version of original_cell_index yet.
> 
> When it’s ready, no need to add a function with ‘set’ in the name. Could be a nanobind read/write property.

agreed, much cleaner solution 👍🏻

PR #3505, opened 2024-11-10
Unify code path of `DirichletBC::set`

schnellerhase:
Reduces code duplicity in the `set` functionality of the Dirichlet boundary conditions.

This also highlights, that in general the fundamental operation we need to implement to set a Dirichlet boundary condition is the one provided by the `set_fn`. So as long as one can define a mapping from index of Dirichlet boundary condition DOF to a coefficient value the `DirichletBC::set` can be defined. Meaning in the future the `DircihletBC` class or `set` function may be altered to take a general `set_fn` instead of a specific type that contains the values - this would follow the functional style employed by the assembly routines and allow easy extension and remove the type dispatching within the `DirichletBC` class.


cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,71 @@ class DirichletBC
   void set(std::span<T> x, std::optional<std::span<const T>> x0,
            T alpha = 1) const
   {
-    std::int32_t x_size = x.size();
-    if (alpha == T(0)) // Optimisation for when alpha == 0
+    // set_fn is a lambda which gets evaluated for every index in [0,
+    // _dofs0.size()) and its result is assigned to x[_dofs0[i]].
+    auto apply = [&](auto set_fn)
```
jorgensd, line 513: Could we not use auto to define `set_fn`, but explicitly declare the type as
`std::function<T(std::size_t>)` ?

jorgensd: And on that subject should `std::size_t` -> `std::int32_t` as all local dofs are `std::int32_t`?

schnellerhase: If the lambda does not capture a lambda will be inlinable trivially, but a `std::function` not, so I would refrain from using `std::function` in general here.

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,71 @@ class DirichletBC
...
+      apply([](std::size_t) -> T { return 0; });
+      return;
     }
-    else
+
+    auto handle_function = [&](std::shared_ptr<const Function<T, U>> g)
     {
-      if (std::holds_alternative<std::shared_ptr<const Function<T, U>>>(_g))
+      assert(g);
+      auto dofs1_g = _dofs1_g.empty() ? std::span(_dofs0) : std::span(_dofs1_g);
```
jorgensd, line 532: Not necessarily a comment on this pull request. But some documentation here would be nice.

I.e. Why do we need to call the internal variable (to this function) `dofs1_g`, 
when `g` only has 1 set of dofs. I think `dofs_g` is a completely valid name.

I guess the docstring could be something like.

// Extract degrees of freedom associated with `g`. If `g` is in a collapsed sub-space, get the dofs in this space, otherwise the degrees of `g` is the same
as for `_x`

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,71 @@ class DirichletBC
...
-                                [x_size, bs, alpha, &value, &x](auto dof)
-                                {
-                                  if (dof < x_size)
-                                    x[dof] = alpha * value[dof % bs];
-                                });
-        }
+        assert(x.size() <= x0.value().size());
+        apply(
+            [&](std::size_t i) -> T {
+              return alpha * (value[_dofs0[i] % bs] - x0.value()[_dofs0[i]]);
```
jorgensd, line 565: Should a temporary reference be made to `_dofs0[i]` to avoid duplicate access queries?

cpp/dolfinx/fem/DirichletBC.h
schnellerhase: But we could add a `std::invocable` concept to be more specific about what we allow.

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -300,12 +314,11 @@ class DirichletBC
...
   /// mixed spaces.
-  template <typename S, typename X,
-            typename
-            = std::enable_if_t<std::is_convertible_v<S, T>
-                               or std::is_convertible_v<S, std::span<const T>>>>
+  template <typename S, typename X>
     requires std::is_convertible_v<std::remove_cvref_t<X>,
                                    std::vector<std::int32_t>>
+             && (std::is_convertible_v<S, T>
+                 || std::is_convertible_v<S, std::span<const T>>)
```
garth-wells, line 321: It would be easier to read if the order of the `requires` followed the same order as the function arguments.

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,82 @@ class DirichletBC
   void set(std::span<T> x, std::optional<std::span<const T>> x0,
            T alpha = 1) const
   {
-    std::int32_t x_size = x.size();
-    if (alpha == T(0)) // Optimisation for when alpha == 0
+    // set_fn is a lambda which gets evaluated for every index in [0,
+    // _dofs0.size()) and its result is assigned to x[_dofs0[i]].
+    auto apply = [&](std::invocable<std::int32_t> auto set_fn)
```
garth-wells, line 513: Is it possible to avoid capturing everything in the lambda function?

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,82 @@ class DirichletBC
...
     {
-      for (std::int32_t idx : _dofs0)
+      static_assert(
+          std::is_same_v<std::invoke_result_t<decltype(set_fn), std::int32_t>,
+                         T>);
+
+      std::int32_t x_size = x.size();
+      for (std::int32_t i = 0; i < _dofs0.size(); ++i)
```
garth-wells, line 520: Surprised this compiles with our strict flags - `_dofs0.size()` has type `std::size_t`.

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,82 @@ class DirichletBC
...
+    };
+
+    if (alpha == T(0)) // Optimisation for when alpha == 0
+    {
+      apply([](std::int32_t) -> T { return 0; });
+      return;
     }
-    else
+
+    auto handle_function = [&](std::shared_ptr<const Function<T, U>> g)
```
garth-wells, line 533: Lambda function name is cryptic. `function` is a highly overloaded word, not sure what `handle` is.

Function should take a shared pointer - it doesn't hold onto it. Pass a reference.

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,82 @@ class DirichletBC
...
-      if (std::holds_alternative<std::shared_ptr<const Function<T, U>>>(_g))
+      assert(g);
+      std::span<const T> values = g->x()->array();
+
+      // Extract degrees of freedom associated with g. If g is in a collapsed
+      // sub-space, get the dofs in this space, otherwise the degrees of g is
+      // the same as for x
+      auto dofs_g = _dofs1_g.empty() ? std::span(_dofs0) : std::span(_dofs1_g);
+
+      if (x0.has_value())
```
garth-wells, line 543:
```suggestion
      if (x0)
```

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,82 @@ class DirichletBC
...
-        std::span<const T> values = g->x()->array();
-        if (x0.has_value())
-        {
-          std::span<const T> _x0 = x0.value();
-          assert(x.size() <= _x0.size());
-          for (std::size_t i = 0; i < _dofs0.size(); ++i)
-          {
-            if (_dofs0[i] < x_size)
+        std::span<const T> _x0 = x0.value();
+        assert(x.size() <= _x0.size());
```
garth-wells, line 546:
```suggestion
        assert(x.size() <= x0->size());
```

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,82 @@ class DirichletBC
...
-            = _dofs1_g.empty() ? std::span(_dofs0) : std::span(_dofs1_g);
-        std::span<const T> values = g->x()->array();
-        if (x0.has_value())
-        {
-          std::span<const T> _x0 = x0.value();
-          assert(x.size() <= _x0.size());
-          for (std::size_t i = 0; i < _dofs0.size(); ++i)
-          {
-            if (_dofs0[i] < x_size)
+        std::span<const T> _x0 = x0.value();
```
garth-wells, line 545:
```suggestion
```

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,82 @@ class DirichletBC
...
-            }
-          }
-        }
-        else
-        {
-          for (std::size_t i = 0; i < _dofs0.size(); ++i)
-          {
-            if (_dofs0[i] < x_size)
+              assert(dofs_g[i] < static_cast<std::int32_t>(values.size()));
+              return alpha * (values[dofs_g[i]] - _x0[_dofs0[i]]);
```
garth-wells, line 551:
```suggestion
              return alpha * (values[dofs_g[i]] - (*x0)[_dofs0[i]]);
```

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,82 @@ class DirichletBC
...
-          }
-        }
+              assert(dofs_g[i] < static_cast<std::int32_t>(values.size()));
+              return alpha * values[dofs_g[i]];
+            });
       }
-      else if (std::holds_alternative<std::shared_ptr<const Constant<T>>>(_g))
+    };
+
+    auto handle_constant = [&](std::shared_ptr<const Constant<T>> g)
```
garth-wells, line 565: Use more expressive function name, and/or add a descriptive comment.

Pass reference rather than shared pointer.

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,82 @@ class DirichletBC
...
+    {
+      const std::vector<T>& value = g->value;
+      std::int32_t bs = _function_space->dofmap()->bs();
+      if (x0.has_value())
```
garth-wells, line 569:
```suggestion
      if (x0)
```

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,82 @@ class DirichletBC
...
-        else
-        {
-          std::ranges::for_each(_dofs0,
-                                [x_size, bs, alpha, &value, &x](auto dof)
-                                {
-                                  if (dof < x_size)
-                                    x[dof] = alpha * value[dof % bs];
-                                });
-        }
+        assert(x.size() <= x0.value().size());
```
garth-wells, line 571:
```suggestion
        assert(x.size() <= x0->size());
```

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,82 @@ class DirichletBC
...
+        apply(
+            [&](std::int32_t i) -> T
+            {
+              auto dof = _dofs0[i];
+              return alpha * (value[dof % bs] - x0.value()[dof]);
```
garth-wells, line 576:
```suggestion
              return alpha * (value[dof % bs] - *(x0)[dof]);
```

garth-wells (changes requested):
Looks good. Just needs some small clarifications and simplifications.

cpp/dolfinx/fem/DirichletBC.h
schnellerhase: We cannot change the type to just a function here, as we use it for the dispatch of the `std::variant`. Depending on which value it holds the "handle" for the specified type is called by `std::visit`. "Function" is chosen not to show that we are defining a function here, but refers to the type this lambda handles.

cpp/dolfinx/fem/DirichletBC.h
schnellerhase: Same as above regarding reference.

cpp/dolfinx/fem/DirichletBC.h
schnellerhase: It is changed to it, it is not very neat however - we need to be quite verbose to make sure we do not capture a copy or a reference when we don't want it. Revert?

cpp/dolfinx/fem/DirichletBC.h
garth-wells: We tend to use `fn` rather than `function` elsewhere to denote a C++ 'function' rather than a DOLFINx 'Function'.

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -495,76 +508,84 @@ class DirichletBC
...
       }
-    }
+      else
+      {
+        apply([alpha, bs, &value, &dofs0 = _dofs0](std::int32_t i) -> T
+              { return alpha * value[dofs0[i] % bs]; });
+      }
+    };
+
+    std::visit(overloaded{handle_function, handle_constant}, _g);
```
garth-wells, line 588: I think using `std::visit` brings in more obfuscation than is necessary. We only have two cases, so an if-else switch would be clearer, would allow control of the types passed to the lambda functions, and would avoid the 'overloaded' boiler plate code further up in the file.

schnellerhase: Unfortunate, but I get the point and it is really the only occasion where `std::variant` would be beneficial within dolfinx. Will remove.

PR #3507, opened 2024-11-13
Fix SNES NonlinearPDE class when using line search

jhale:
Thanks to @jonas-heinzmann for clearly identifying and explaining this issue and @michalhabera for help creating this fix.

When using a linesearch routine SNES needs to calculate (multiple times) the norm of the residual evaluated at a vector `w` equal to the current solution plus some proposed scaling factor times the Newton increment. To do this SNES calls the user-defined residual evaluation routine passing `w`. Our current implementation of the residual evaluation routine overwrites the residual form state `u` with `w` but does not *replace* `u` with the current solution `x` before returning. This leads to a failure of the linesearch.

This patch fixes this behaviour by preserving the internal state of the residual.

### DIFF
--- a/python/dolfinx/wrappers/mesh.cpp
+++ b/python/dolfinx/wrappers/mesh.cpp
@@ -612,7 +612,7 @@ void mesh(nb::module_& m)
           nb::rv_policy::reference_internal)
       .def_prop_ro("dim", &dolfinx::mesh::Topology::dim,
                    "Topological dimension")
-      .def_prop_ro(
+      .def_prop_rw(
           "original_cell_index",
           [](const dolfinx::mesh::Topology& self)
           {
@@ -623,7 +623,16 @@ void mesh(nb::module_& m)
             return nb::ndarray<const std::int64_t, nb::numpy>(idx[0].data(),
                                                               {idx[0].size()});
           },
-          nb::rv_policy::reference_internal)
+          [](dolfinx::mesh::Topology& self,
+             const nb::ndarray<const std::int64_t, nb::ndim<1>, nb::c_contig>&
+                 original_cell_indices)
+          {
+            self.original_cell_index.resize(1);
+            self.original_cell_index[0].assign(
+                original_cell_indices.data(),
+                original_cell_indices.data() + original_cell_indices.size());
+          },
+          nb::arg("original_cell_indices"))
       .def("connectivity",
            nb::overload_cast<int, int>(&dolfinx::mesh::Topology::connectivity,
                                        nb::const_),
--- a/python/test/unit/nls/test_newton.py
+++ b/python/test/unit/nls/test_newton.py
@@ -11,7 +11,6 @@ import numpy as np
 import pytest
 
 import ufl
-from dolfinx import cpp as _cpp
 from dolfinx import default_real_type
 from dolfinx.fem import Function, dirichletbc, form, functionspace, locate_dofs_geometrical
 from dolfinx.mesh import create_unit_square
@@ -106,6 +105,7 @@ class TestNLS:
     def test_linear_pde(self):
         """Test Newton solver for a linear PDE."""
         from petsc4py import PETSc
+
         from dolfinx.nls.petsc import NewtonSolver
 
         # Create mesh and function space
@@ -154,6 +154,7 @@ class TestNLS:
     def test_nonlinear_pde(self):
         """Test Newton solver for a simple nonlinear PDE"""
         from petsc4py import PETSc
+
         from dolfinx.nls.petsc import NewtonSolver
 
         mesh = create_unit_square(MPI.COMM_WORLD, 12, 5)
--- a/python/test/unit/nls/test_newton.py
+++ b/python/test/unit/nls/test_newton.py
@@ -6,6 +6,7 @@
 """Unit tests for Newton solver assembly"""
 
 from mpi4py import MPI
+from petsc4py import PETSc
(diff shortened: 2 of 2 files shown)


jhale:
Failing line search in parallel run https://petsc.org/release/manualpages/SNES/SNES_DIVERGED_LINE_SEARCH

jhale:
This works but seems unsatisfactory. A point from Jonas:

"...the pointer to the PETSc solution vector is linked directly to the pointer of the solution vector of Dolfinx, while in fact both instead should be independent, i.e. the vector at which Dolfinx assembles the forms and the solution vector PETSc works must not be the same to allow for these temporary solution vectors in PETSc."

michalhabera:
Looking at this again, I think a simpler solution would be to pass a new independent vector to a call to snes.solve().

jorgensd:
Could we move this implementation into the dolfinx namespace?

It is rather hard for users to see how to use PETSC snes as they have to browse the tests.

jonas-heinzmann:
To elaborate a bit more on the PETSc side of things to help the discussion:

In my understanding, the SNES solver object usually only works with a solution vector $X$ at which it needs to evaluate the function $F(X)$ as well as the Jacobian $J(X)$.
To me it seems that with the previously proposed SNESProblem class setup, the PETSc vector $X$ is somehow directly linked to `self.u` in the dolfinx namespace such that we can assemble $a(X)$ and $L(X)$.

This is not an issue _unless_ we want to properly use the SNESLinesearch routines:
They work with another temporary vector $W$ at which the line search routines need to evaluate $F(W)$, while they explicitly rely on $X$ as an unchanged reference point to (iteratively) test different step sizes $\lambda_i$, i.e.

$$W_{i+1} = X + \lambda_i \Delta X$$

See for instance [here](https://gitlab.com/petsc/petsc/-/blob/main/src/snes/linesearch/impls/l2/linesearchl2.c?ref_type=heads#L46).
However, once the line search routines call the `SNESProblem.F` method passing in $W$, since dolfinx wants to assemble $L(X)$, we need to explicitly overwrite $X \leftarrow W$, which then causes $X$ to change in the PETSc namespace, meaning that the line searches can't rely on a fixed reference point to test different step lengths $\lambda$.
This means that the line searches (for iteration $i>1$) get

$$W_{i+1} = W_i + \lambda_i \Delta X$$

instead of the expected

$$W_{i+1} = X + \lambda_i \Delta X$$

which causes them to fail when they try to fit an ansatz e.g. to $||F||_2$ along the step length.
When one looks at the output of the line search routines, this typically means that they shrink the step size $\lambda$ further and further, until the minimum step size is reaches and then exit.
As long as the line search only needs a single evaluation of $F(W)$, everything appears to work since in that case it does not matter if the reference point is not consistent due to $X\leftarrow W$.

So perhaps another idea to fix the issue would be to somehow 'unlink' `u` in the dolfinx namespace from `X` in the PETSc namespace?
Meaning that the solution vector at which dolfinx assembles $a$ and $L$ are independent of the vectors in the PETSc namespace.
Then the `SNESProblem.F` method could assemble the forms at whatever point PETSc passes in, without us interfering with the solution vectors of PETSc.

jhale:
"So perhaps another idea to fix the issue would be to somehow 'unlink' u in the dolfinx namespace from X in the PETSc namespace?"

Agreed, I think that is the best solution and I think it's more consistent with the design of DOLFINx and SNES. We should also consider adjusting our own NewtonSolver if we make the same link there.

michalhabera:
Thanks @jonas-heinzmann for noticing this.

I'd suggest following patch. I've checked and all other uses of SNES (e.g. `test_petsc_nonlinear_assembler.py`) do allocate new solution vector (unrelated to any function in forms),
```
diff --git a/python/test/unit/nls/test_newton.py b/python/test/unit/nls/test_newton.py
index ba56dbcdf1..8ef044f515 100644
--- a/python/test/unit/nls/test_newton.py
+++ b/python/test/unit/nls/test_newton.py
@@ -224,13 +224,16 @@ class TestNLS:
         snes.getKSP().setTolerances(rtol=1.0e-9)
         snes.getKSP().getPC().setType("lu")
 
-        snes.solve(None, u.x.petsc_vec)
+        x = u.x.petsc_vec.copy()
+        x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+
+        snes.solve(None, x)
         assert snes.getConvergedReason() > 0
         assert snes.getIterationNumber() < 6
 
         # Modify boundary condition and solve again
         u_bc.x.array[:] = 0.6
-        snes.solve(None, u.x.petsc_vec)
+        snes.solve(None, x)
         assert snes.getConvergedReason() > 0
         assert snes.getIterationNumber() < 6
         # print(snes.getIterationNumber())
```

jhale:
I think this is fixed - @jonas-heinzmann can you try this out?

In short, users should (almost) never pass the same vector memory to `snes.solve` as used in the specifying the residual or Jacobian form.

@jorgensd We can bring `NonlinearPDE_SNESProblem` out to the user API in another PR.

I also took a look at our `NewtonSolver`. In our current `NonlinearPDEProblem` class we assume that `x` and `u` share the same memory - infact without this the solver will not converge. I would propose for consistency with SNES (that I believe has the right design here) that we remove this implicit coupling in our examples.

jonas-heinzmann:
This seems to do the trick, thanks a lot!

Issue #3483, opened 2024-10-26
Python wrappers for `AdjacencyList` and `FiniteElement` missing


schnellerhase:
Both `AdjacencyList` and `FiniteElement` are exported for multiple data types and thus should be wrapped into a uniform python interface as done for other multi type exports.

chrhansk:
I am experiencing a similar problem: I have so far been using version `0.7.1` of `fenics-dolfinx` using `anaconda`. This version has the class `FiniteElement` in the `ufl` module. 

However, when switching to `0.9.0`, the class is simply gone, breaking my code. Is there some workaround for the current version?

jorgensd:
@chrhansk You are referring to `ufl.FiniteElement`, not `dolfinx.cpp.fem.FiniteElement`.
All finite elements from `ufl` has been removed (`ufl.FiniteElement`, `ufl.VectorElement`, `ufl.TensorElement`, and been replaced by the relevant `basix.ufl.element`. For an explanation about how to write these elements, see for instance: 
https://jsdokken.com/FEniCS-workshop/src/unified_form_language/ufl_elements.html#the-finite-element
or https://jsdokken.com/fenics22-tutorial/comparing_elements.html#piecewise-constant-pressure-spaces
and subsequent sections

chrhansk:
@jorgensd: Thank you for the help and sorry for the noise.

PR #3509, opened 2024-11-13
Add `AdjacencyList` python wrapper

schnellerhase:
Introduces a python wrapper for the multi-type exported `AdjacencyList`.

Addresses (partially) https://github.com/FEniCS/dolfinx/issues/3483


python/dolfinx/geometry.py
```diff
@@ -231,10 +228,14 @@ def compute_colliding_cells(
...
     """
-    return _cpp.geometry.compute_colliding_cells(mesh._cpp_object, candidates, x)
+    return AdjacencyList(
+        _cpp.geometry.compute_colliding_cells(mesh._cpp_object, candidates._cpp_object, x)
+    )
 
 
-def squared_distance(mesh: Mesh, dim: int, entities: list[int], points: npt.NDArray[np.floating]):
+def squared_distance(
+    mesh: Mesh, dim: int, entities: list[int], points: npt.NDArray[np.floating]
```
jorgensd, line 237: Unrelated to this PR, but as we usually use `npt.NDArray[np.int32]` for the entities, do you mind changing it to this here?

python/dolfinx/graph.py
```diff
@@ -28,10 +31,34 @@
     pass
 
 
-__all__ = ["adjacencylist", "partitioner"]
+__all__ = ["AdjacencyList", "adjacencylist", "partitioner"]
+
+
+class AdjacencyList:
+    _cpp_object: Union[_cpp.la.AdjacencyList_int32, _cpp.la.AdjacencyList_int64]
```
jorgensd, line 38: Would be great if we could add some docstrings to this class (as that is the main motivation for having a Python wrapper).

python/dolfinx/graph.py
```diff
@@ -42,15 +69,9 @@ def adjacencylist(data: np.ndarray, offsets=None):
...
-    else:
-        try:
-            return _cpp.graph.AdjacencyList_int32(data, offsets)
-        except TypeError:
-            return _cpp.graph.AdjacencyList_int64(data, offsets)
+    # Switch to np.isdtype(data.dtype, np.int32) once numpy >= 2.0 is enforced
+    is_32bit = data.dtype == np.int32
+    cpp_t = _cpp.graph.AdjacencyList_int32 if is_32bit else _cpp.graph.AdjacencyList_int64
+    cpp_object = cpp_t(data, offsets) if offsets is not None else cpp_t(data)
+    return AdjacencyList(cpp_object)
```
jorgensd, line 77: So we need a guard for invalid input style (ie np.floating), or do we think the general nanobind error is sufficient? Usually the nanobind type arrays aren’t very expressive when it comes to dtype.

schnellerhase: Defending against such calls would require quite a bit of additional logic. It doesn't end there if we did that, then we would need for example also checks of the shapes of the arrays.

Given that it can not cause a false positive, i.e. if we call with a `float` array it will not result in a working code path that produces bad results, and type hints being present, that should flag this in a modern IDE or with `mypy`, I think we should stick to the simplicity of not dealing with it.

python/dolfinx/graph.py
```diff
@@ -38,21 +38,52 @@ class AdjacencyList:
     _cpp_object: Union[_cpp.la.AdjacencyList_int32, _cpp.la.AdjacencyList_int64]
 
     def __init__(self, cpp_object: Union[_cpp.la.AdjacencyList_int32, _cpp.la.AdjacencyList_int64]):
+        """Creates a python wrapper for the exported adjacency list class.
```
jorgensd, line 41:
```suggestion
        """Creates a Python wrapper for the exported adjacency list class.
```

python/dolfinx/graph.py
```diff
@@ -38,21 +38,52 @@ class AdjacencyList:
...
+
+        Note:
+            This constructor does not create a new adjacency list, see `adjacencylist` for that.
```
jorgensd, line 44:
```suggestion
            This constructor does not create a new adjacency list, see  :func:`adjacencylist` for that.
```

python/dolfinx/graph.py
```diff
@@ -38,21 +38,52 @@ class AdjacencyList:
...
+        """Array representation of the adjacency list.
+
+        Returns:
+            Flattened array representation of the adjacency list.
+        """
         return self._cpp_object.array
 
     @property
     def offsets(self) -> npt.NDArray[np.int32]:
+        """Offsets for each node in the `array()`.
```
jorgensd, line 73:
```suggestion
        """Offsets for each node in the :func:`array`.
```

python/test/unit/fem/test_dofmap.py
```diff
@@ -436,7 +436,8 @@ def test_empty_rank_collapse():
     def self_partitioner(comm: MPI.Intracomm, n, m, topo):
         dests = np.full(len(topo[0]) // 2, comm.rank, dtype=np.int32)
         offsets = np.arange(len(topo[0]) // 2 + 1, dtype=np.int32)
-        return dolfinx.graph.adjacencylist(dests, offsets)
+        # TODO: can we improve on this interface? I.e. warp to do cpp type conversion automatically
```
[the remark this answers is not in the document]
schnellerhase: Done, see https://github.com/FEniCS/dolfinx/issues/3531

Issue #3506, opened 2024-11-11
Refactor C++ `dolfinx::mesh::Topology`


jorgensd:
### Describe new/missing feature

Current Topology initializer just takes in an MPI communicator and a cell type, and then all other information is passed through `set_index_map` or `set_connectivity` or `original_cell_index`.
However, a topology isn't really valid (none of the `create_entities` or `create_connectivity` functions can be called without):
- A cell index map
- A vertex index map
- A cell_to_vertex connectivity
- Original cell indices ( Not required for create_*, but for reading external data).

Additionally, users can make invalid connections or index maps through `set_index_map` and `set_connectivity` post initialization.

A benefit of this is that we allow for an "adaptive mesh topology".
However, we do not have the same flexibility for the mesh geometry (cannot replace index map or change the size of `x`, or the indices of the dofmap).

Thus, I propose we refactor topology such that:
- **Initializer**: Takes in communicator, cell_type, cell_index_map, vertex_index_map, c_to_v and original_cell_indices
- **Destruction of connectivity**: Sometimes users might want to delete a connectivity after usage. We could have a `clear_connectivity(i, j)` that removes the connectivity (and guards against removal of c_to_v, c_to_c or v_to_v connectivity).
- **Destruction of entities**: Can currently be done with `set_index_map` (which is unsafe). Should be done with a `clear_entities(i)` with guards against clearing cells or vertices.

PR #3512, opened 2024-11-15
Improve Topology construction and fix MPI/NBX bug

garth-wells:
This change gives a mesh `Topology` a 'fully formed' constructor, i.e. a `Topology` object is complete from construction in that it has vertices and cells. Other entries are derived from these. This follows the DOLFINx design philosophy where we avoid creating half-baked objects with unclear state.

These changes triggered are seemingly unrelated bug in the NBX algorithm related to MPI. Consecutive, fast calls to the NBX could result in corrupted data since the non-blocking MPI calls used the same tag. The tag passed to MPI functions is now an optional argument.

Change also remove duplicated code in `Topology` constructors.

Resolves some of the point in #3506.


jorgensd:
I would suggest refactoring the `set_connectivity` function as well, as well as set_index_map, as explained in https://github.com/FEniCS/dolfinx/issues/3506

garth-wells:
> I would suggest refactoring the `set_connectivity` function as well, as well as set_index_map, as explained in #3506

I think further changes should come later because they are much more intrusive for the interface, and some points need careful thought. This changes doesn't really affect the API (since users do not generally use the `Topology` constructor), and the PR fixes a bug.

jorgensd:
> > I would suggest refactoring the `set_connectivity` function as well, as well as set_index_map, as explained in #3506
> 
> I think further changes should come later because they are much more intrusive for the interface, and some points need careful thought. This changes doesn't really affect the API (since users do not generally use the `Topology` constructor), and the PR fixes a bug.

What I am suggesting isn't very intrusive to the API as most users do not:
1. Attach index maps to the topology themself (as this is done through create_entities)
2. Does not clear the entities post creation.
3. Does set connectivity manually (as it is done through create connectivity).

The `clear options` would be a simple change for anyone using the current interface explicitly.

chrisrichardson (approved):

PR #3520, opened 2024-11-19
Allow mixed element with single element

### LINKED ISSUE #3519: Mixed element with a single element
### Summarize the issue

Since #3500 one can no longer make a mixed function space with a single element.
I see no sensible justification for this, as it massively complicates packages that uses an arbitrary number of mixed elements (which can be >=1).
First reported at https://github.com/festim-dev/FESTIM/pull/922#issuecomment-2479881906

### How to reproduce the bug

Run the below on nightly

### Minimal Example (Python)

```Python
from mpi4py import MPI
import dolfinx

import basix.ufl


mesh = dolfinx.mesh.create_unit_interval(MPI.COMM_WORLD, 10)
el = basix.ufl.element("Lagrange", mesh.basix_cell(), 1)
me = basix.ufl.mixed_element([el])

V = dolfinx.fem.functionspace(mesh, me)
```

### Output (Python)

```bash
  File "/root/shared/mwe_mixed.py", line 11, in <module>
    V = dolfinx.fem.functionspace(mesh, me)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/dolfinx-real/lib/python3.12/dist-packages/dolfinx/fem/function.py", line 630, in functionspace
    cpp_element = _create_dolfinx_element(mesh.topology.cell_type, ufl_e, dtype)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/dolfinx-real/lib/python3.12/dist-packages/dolfinx/fem/function.py", line 578, in _create_dolfinx_element
    return CppElement(elements)
           ^^^^^^^^^^^^^^^^^^^^
RuntimeError: FiniteElement constructor for mixed elements called with a single element.
```

### Version

main branch

### DOLFINx git commit

_No response_

### Installation

_No response_

### Additional information

_No response_

jorgensd:
Resolve #3519

### DIFF
--- a/cpp/dolfinx/fem/FiniteElement.cpp
+++ b/cpp/dolfinx/fem/FiniteElement.cpp
@@ -173,12 +173,6 @@ FiniteElement<T>::FiniteElement(
       _symmetric(false), _needs_dof_permutations(false),
       _needs_dof_transformations(false)
 {
-  if (elements.size() < 2)
-  {
-    throw std::runtime_error("FiniteElement constructor for mixed elements "
-                             "called with a single element.");
-  }
-
   _signature = "Mixed element (";
 
   const std::vector<std::vector<std::vector<int>>>& ed
--- a/cpp/dolfinx/fem/FiniteElement.h
+++ b/cpp/dolfinx/fem/FiniteElement.h
@@ -73,13 +73,32 @@ public:
                 bool symmetric = false);
 
   /// @brief Create a mixed finite element from Basix finite elements.
+  ///
+  /// See FiniteElement(const std::vector<std::shared_ptr<const
+  /// FiniteElement<geometry_type>>>&) for a discussion of mixed
+  /// elements.
+  ///
   /// @param[in] elements List of (Basix finite element, block size,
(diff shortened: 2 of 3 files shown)


garth-wells (changes requested):
This fix is at the wrong level. It needs to be done at a high level when the appropriate constructor is called.

garth-wells:
See https://github.com/FEniCS/basix/pull/882.

jorgensd:
> See [FEniCS/basix#882](https://github.com/FEniCS/basix/pull/882).

This doesn’t resolve the issue, as you then forces the external library to handle two different cases: 
If they only have a single elements, go down one code path, if there are multiple, do something completely different.

this is because one relies on .sub(i) for interpolation, boundary condition etc.

RemDelaporteMathurin (approved):

garth-wells (approved):

Issue #3522, opened 2024-11-20
Unresolved PETSc memory leaks in `dolfinx.fem` when running in parallel with MPI


moonlitfjords:
I have previously come across some of the memory leaks relating to PETSc objects not being garbage collected and was able to get around them using the manual destruction suggestions made on discussions here.

However, I've now come across some akin to those discussed in #2552 and #2559 which seem to break multiprocessing on the current version of `petsc4py` being used. These manifested initially as `yaksa: x leaked handle pool objects` warnings, but while I haven't been able to completely isolate their source, I've managed to track them down to the creation and manipulation of `dolfinx.fem` objects which appear to be creating (and failing to destroy) a number of different MPI objects.

This occurs exclusively when running using `mpirun` across multiple cores, such that if several tasks are running in parallel each using one core there is no problem, but as soon as one or more instances are being distributed across multiple processes, the memory leaks emerge.

Having rebuilt MPI with additional debugging flags, the output from one of the cores upon termination of the script is included below. The objects in question are all arising from MPI-related files, with COMM objects first appearing upon calling `fem.functionspace`; `fem.Function` calls then create ATTR and KEYVAL objects as well, with GROUP and REQUEST objects (I think) being associated with `fem.Form` objects and finally DATATYPE objects being created by vector/matrix manipulations within `fem.petsc`.

It is the DATATYPE objects in particular that get flagged by `yaksa`, and adding in additional `gc.collect()`s does not resolve the problem.

I have not determined precisely which version this becomes broken in, but the behaviour is not present at least within `petsc4py==3.19.6` and earlier, so unless anyone has any ideas on how to resolve this I'd suggest it might be worth regressing the version provided by default in the various premade Docker images?

In the meantime, for anyone else coming across this with similar issues, manually building a clean Docker image on top of Ubuntu using the `apt-get` instructions in the installation guide for  `dolfinx 0.9.0` and then adding on an older version of `petsc4py` has allowed me to bypass the issue. Otherwise a slow buildup of memory will eventually cause a crash when running large numbers of simulations (I was able to manage about 50 on a relatively small mesh using 8 total cores and 2 cores per simulation before approaching my machine's capacity).

Apologies if any of this isn't clear or I am missing something important/obvious, but hopefully it generally makes sense.

_(Output from `mpirun` with debug-build of MPI:)_
```
leaked context IDs detected: mask=0x7fdac4fb9440 mask[0]=0x7fff7e7f
[...]

PR #3524, opened 2024-11-20
Add more PETSc destruction notes to assembly functionality that creates PETSc.Mat/Vec objects

jorgensd:
Continuation of: https://github.com/FEniCS/dolfinx/pull/3329
Uncovered in: https://github.com/FEniCS/dolfinx/issues/3522#issuecomment-2488679142


garth-wells:
I think we should have the 'note' text appear in one place in the docs and refer (link) to it where appropriated. We have duplicated the text in a *lot* of places.

jorgensd:
> I think we should have the 'note' text appear in one place in the docs and refer (link) to it where appropriated. We have duplicated the text in a _lot_ of places.

I guess the most natural place is to have one under create_matrix and another under create_vector, and the have references everywhere else.

PR #3529, opened 2024-11-22
Remove jit options from function-space as we no longer generate code.

jorgensd:
Remove commented out code from create_dolfinx_element.


jorgensd:
Also make sure that newest mixed element test (#3520) works on single precision, as it currently gets into issues with basix due to the float type of the basix element. https://github.com/FEniCS/ffcx/actions/runs/11973090647/job/33381299505?pr=730

jhale (approved):

PR #3530, opened 2024-11-22
Remove pkg-config install from macOS test

jhale:


jhale:
This seems like an issue with the image/homebrew.

schnellerhase:
This looks like the same problem: https://github.com/apache/arrow/pull/44798 - maybe try to switch to `pkgconf` as well?

jhale:
This `brew unlink` can be removed once the following GitHub Action fix is rolled out: https://github.com/Homebrew/homebrew-core/issues/198881

PR #3533, opened 2024-11-25
Make read_meshtags a template function

mleoni-pf:
This PR leverages functionalities introduced some time ago by @jorgensd that makes `xdmf_utils::distribute_entity_data` process data of a generic type, not only `std::int32_t`. In this PR I make `XDMFFile::read_meshtags` a template so that we can read mesh tags of any type [esp. `double`] from file. This will be useful in an upcoming PR to read functions from file.


cpp/dolfinx/io/XDMFFile.h
```diff
@@ -170,10 +170,11 @@ class XDMFFile
   /// GridType="Uniform"
   /// @param[in] attribute_name Name of the attribute to read
   /// @param[in] xpath XPath where MeshTags Grid is stored in file
-  mesh::MeshTags<std::int32_t>
-  read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
-                std::optional<std::string> attribute_name,
-                std::string xpath = "/Xdmf/Domain");
+  template <typename T = std::int32_t>
```
garth-wells, line 173: What types are supported? If the type in the file can't be checked it can lead to all sorts of crashes and mis-behaviours.

What is needed for this to be robust is a function for querying the data type in the file. I started this ages ago for HDF5 files, but it wasn't trivial to make it precise, but getting the type size was easy and may be adequate.

jhale: We hit the same issue with the recent checkpointing project. We should have a chat in #developmemt about the best ways to tackle this.

mleoni-pf: I implemented something like this [untested]:

```C++
// Define a type-trait-kinda thing
template <typename T>
struct xdmf_data
{
};

template <>
struct xdmf_data<std::int32_t>
{
  static const std::string data_type;
  static const std::size_t precision = 4;
};
const std::string xdmf_data<std::int32_t>::data_type = "Int";

template <>
struct xdmf_data<std::int64_t>
{
  static const std::string data_type;
  static const std::size_t precision = 8;
};
const std::string xdmf_data<std::int64_t>::data_type = "Int";

template <>
struct xdmf_data<float>
{
  static const std::string data_type;
  static const std::size_t precision = 4;
};
const std::string xdmf_data<float>::data_type = "Float";

template <>
struct xdmf_data<double>
{
  static const std::string data_type;
  static const std::size_t precision = 8;
};
const std::string xdmf_data<double>::data_type = "Float";

template <typename T>
mesh::MeshTags<T>
XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
                        std::optional<std::string> attribute_name,
                        std::string xpath)
{
  // omitted stuff...
  values_data_node = attribute_node.child("DataItem");

  const std::string data_type = values_data_node.attribute("DataType").value();
  const auto precision
      = std::stol(values_data_node.attribute("Precision").value());

  if (xdmf_data<T>::data_type != data_type
      || xdmf_data<T>::precision != precision)
  {
    throw std::runtime_error(
        "The data in the XDMF file does not match the required data type.");
  }

  // continue...
```

What do you think of this approach? Needs polishing obviously... @garth-wells  @jhale

mleoni-pf: Ok, I just implemented and pushed what I believe is a solution to this problem, I'd be happy to hear your comments!

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -363,10 +382,83 @@ XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
...
+        = values_data_node.attribute("Precision");
+        precision_attribute)
+    {
+      precision = std::stol(values_data_node.attribute("Precision").value());
+    }
+    // otherwhise use the standard's default
+    else
+    {
+      precision = 4;
+    }
```
jorgensd, line 450:
```suggestion
      precision = std::stol(values_data_node.attribute("Precision").value());
    // otherwhise use the standard's default
    else
      precision = 4;
```
I think this would be the "house-style"

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -363,10 +382,83 @@ XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
...
+             = values_data_node.attribute("NumberType");
+             data_type_attribute)
+    {
+      data_type = data_type_attribute.value();
+    }
+    // If unspecified, use the standard's implicit default
+    else
+    {
+      data_type = "Float";
+    }
```
jorgensd, line 435:
```suggestion
    if (const auto& data_type_attribute
        = values_data_node.attribute("DataType");
        data_type_attribute)
      data_type = data_type_attribute.value();
    // ... But "NumberType" is supported too and widely used in its place
    else if (const auto& data_type_attribute
             = values_data_node.attribute("NumberType");
             data_type_attribute)
      data_type = data_type_attribute.value();
    // If unspecified, use the standard's implicit default
    else
      data_type = "Float";
```
I also think this would be the house-style. @garth-wells  can comment further

cpp/dolfinx/io/XDMFFile.cpp
```diff
@@ -397,12 +489,32 @@ XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
...
+template mesh::MeshTags<std::int64_t>
+XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
+                        std::optional<std::string> attribute_name,
+                        std::string xpath);
+template mesh::MeshTags<float>
+XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
+                        std::optional<std::string> attribute_name,
+                        std::string xpath);
+template mesh::MeshTags<double>
+XDMFFile::read_meshtags(const mesh::Mesh<double>& mesh, std::string name,
```
jorgensd, line 513: Maybe we should add complex wrappers as well?

garth-wells:
We're deprecating XDMF in favour of VTKHDF5, see #3674.

Issue #3516, opened 2024-11-18
[BUG]: Interpolation into symmetric rank-2 tensors broken


michalhabera:
### Summarize the issue

Recent changes altered the value shape of symmetric rank-2 tensors, being rank-1 with shape `(dim * (dim + 1)/2, )`. This means that interpolation of objects with shape `(dim, dim)` into symmetric rank-2 tensors does not work anymore.

See https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/FiniteElement.cpp#L118

### How to reproduce the bug

Interpolate into symmetric rank-2 tensors.

### Minimal Example (Python)

```Python
import dolfinx
from mpi4py import MPI

mesh = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, 1, 1)

REG = dolfinx.fem.functionspace(mesh, ("Regge", 1))
Psymm = dolfinx.fem.functionspace(mesh, ("P", 1, (2, 2), True))

print(f"Regge value shape: {REG.element.value_shape}")
print(f"Psymm value shape: {Psymm.element.value_shape}")

f0 = dolfinx.fem.Function(REG)
f1 = dolfinx.fem.Function(Psymm)

f0.interpolate(f1)
```

### Output (Python)

```bash
Regge value shape: [2 2]
Psymm value shape: [3]
Traceback (most recent call last):
  File "/home/UNILU/symm_interp.py", line 15, in <module>
    f0.interpolate(f1)
  File "/home/UNILU/dolfinx/python/dolfinx/fem/function.py", line 459, in interpolate
    _interpolate(u0)
  File "/usr/lib/python3.12/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/UNILU/dolfinx/python/dolfinx/fem/function.py", line 445, in _
    self._cpp_object.interpolate(u0._cpp_object, cells0, cells1)  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Interpolation: elements have different value dimensions
```

### Version

main branch

PR #3535, opened 2024-11-25
Fix symmetric value shape

michalhabera:
Attempt to fix https://github.com/FEniCS/dolfinx/issues/3516.

This PR makes symmetric rank-2 elements have `block size == dim*(dim + 1) / 2` but keeps `value_shape == (dim, dim)`.

### DIFF
--- a/cpp/dolfinx/fem/FiniteElement.cpp
+++ b/cpp/dolfinx/fem/FiniteElement.cpp
@@ -302,30 +302,11 @@ int FiniteElement<T>::space_dimension() const noexcept
 template <std::floating_point T>
 int FiniteElement<T>::value_size() const
 {
-  if (_value_shape)
-  {
-    int vs = std::accumulate(_value_shape->begin(), _value_shape->end(), 1,
-                             std::multiplies{});
-
-    // See comments in constructor on why special handling for the
-    // symmetric case is required.
-    if (_symmetric)
-    {
-      if (vs == 3)
-        return 4;
-      else if (vs == 6)
-        return 9;
-      else
-      {
-        throw std::runtime_error(
-            "Inconsistent size for symmetric rank-2 tensor.");
-      }
-    }
-    else
-      return vs;
-  }
-  else
+  if (!_value_shape)
     throw std::runtime_error("Element does not have a value_shape.");
+
+  return std::accumulate(_value_shape->begin(), _value_shape->end(), 1,
+                         std::multiplies{});
 }
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -115,30 +103,45 @@ FiniteElement<T>::FiniteElement(
...
   if (value_shape and !element.value_shape().empty())
   {
     throw std::runtime_error("Blocked finite elements can be constructed only "
                              "from scalar base elements.");
   }
 
-  if (symmetric) // Consistency check for symmetric elements
-  {
-    if (!value_shape)
+  if (symmetric && value_shape){
```
garth-wells, line 112: Run `clang-format`/

michalhabera: Fixed. I'll still suggest to wait so I test this in more elaborate tests in dolfiny. TBH, I just fixed the apparent places and I do not see completely through if all works downstream (the code is a bit complicated).

cpp/dolfinx/fem/FiniteElement.h
```diff
@@ -64,12 +63,11 @@ class FiniteElement
...
   /// for a vector in 3D or `{2, 2}` for a rank-2 tensor in 2D. Can only
   /// be set for blocked scalar `element`. For other elements and scalar
-  /// elements it should be `std::nullopt`.
+  /// elements it should be empty vector.
   /// @param[in] symmetric Is the element a symmetric tensor? Should ony
   /// set for 2nd-order tensor blocked elements.
   FiniteElement(const basix::FiniteElement<geometry_type>& element,
-                std::optional<std::vector<std::size_t>> value_shape
-                = std::nullopt,
+                std::vector<std::size_t> value_shape = {}, int block_size = 1,
```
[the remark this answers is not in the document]
michalhabera: The whole point of extra changes is to simplify this nontrivial code.

1. My motivation for having block size as an extra argument is that the logic to compute it already lives in basix UFL wrapper, so it is better to avoid duplicating this nontrivial assumption, pass the one from basix instead. It does not have to have a default value.
2. I do not see the point of using `std::optional<std::vector<>>`. Optional is okay for when its value might not be provided. But empty std::vector is semantically correct and nicely maps in Python to an empty list or empty array. For lists/tuples on the Python side one does not use `None` to indicate an empty tuple/list. So here I strongly disagree.

I'll revert the changes to previous commit so the regression is fixed.

garth-wells (approved):

PR #3536, opened 2024-11-25
Add xdmffile read function

mleoni-pf:
NOTE: This PR is meant to be stacked onto #3533 . The diff of this PR shows the edits of both but the only new changes are the ones related to the function `XDMFFile::read_function`. The rest of the diff will disappear once #3533 is merged and this PR is rebased onto `main` but I am uploading it now so it can be reviewed immediately.

This PR introduces much-needed functionalities that let a user read a function from file. This is vital in most scenarios in which a Finite Element element simulation is part of a bigger pipeline and needs to read input data produced by other software.
This PR adds a `read_function` routine to the class `XDMFFile` that can read a P1 function from an XDMF file exploiting the fact that the degrees of freedom of a P1 function are the mesh vertexes and that we can now read a [`double`-valued] meshtag defined on vertexes, so we can define a P1 function.
The implementation closely follows the normal `read_meshtags` function but reads vertex data instead and, as a last step, copies the data in a function's `array` in the correct position.

Co-implemented with @jorgensd .


jorgensd:
Wouldn't it be sufficient to read in the "vertex_data" as a MeshTags object (using `read_meshtags<double/float/complex double/complex float>`, and then cast the data using a map like:
https://fenicsproject.discourse.group/t/how-to-apply-a-boundary-condition-to-a-range-of-dofs/15758/7?u=dokken
as a stand-alone function?
It would separate concerns, and make it more generalizable for a time when XDMFFile is phased out.

mleoni-pf:
@jorgensd This is not possible, as far as I can tell, because `read_meshtags` reads data based on the cell type from the topology node. The reference file I'm reading from is

```
<Xdmf Version="3.0">
<Domain>
    <Grid Name="Grid">
        <Geometry GeometryType="XYZ">
            <DataItem DataType="Float" Dimensions="23103 3" Format="HDF" Precision="4">
            wallWithEdgeRadiationExtruded.h5:/data0</DataItem>
        </Geometry>
        <Topology TopologyType="Hexahedron" NumberOfElements="15000" NodesPerElement="8">
            <DataItem DataType="Int" Dimensions="15000 8" Format="HDF" Precision="8">
            wallWithEdgeRadiationExtruded.h5:/data1</DataItem>
        </Topology>
        <Attribute Name="edge radiation" AttributeType="Scalar" Center="Node">
            <DataItem DataType="Float" Dimensions="23103" Format="HDF" Precision="8">
            wallWithEdgeRadiationExtruded.h5:/data2</DataItem>
        </Attribute>
    </Grid>
</Domain>
</Xdmf>
```

so `read_meshtags` would try to read tags associated with hexahedra here. There is a comment in the code where I clarify this.

garth-wells:
We're deprecating XDMF in favour of VTKHDF5, see #3674.

PR #3537, opened 2024-11-25
Get FEniCSx refs from central workflow

jhale:
Currently during development and on release it is necessary to change hundreds of places to get a consistent set of refs tested. This allows this to be set centrally in a single file and additionally will make sure all checkouts are actually done at refs (many are not currently correct).

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -28,11 +28,18 @@ on:
         type: string
 
 jobs:
+  refs:
+    uses: ./fenicsx-refs.yml
+
   lint:
     runs-on: ubuntu-latest
+    needs: refs
     name: Lint
     steps:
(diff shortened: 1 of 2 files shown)


michalhabera (approved):
Nice!

garth-wells:
This just looks more complicated to me. I didn't see the issue with the previous, simpler setup. We need to resist unnecessary complexity.

jhale:
I think I found a native way to remove the extra workflow and simply read the refs from a json file. Then it will just look like the if: workflow_dispatch that is in most workflows already. It will be next week before I get a chance to do it.

I can assure you this is a huge simplification for me when I make releases where I have to change branches/tags sets iteratively and need those tested across every component. It will also be useful when we move forwards to (attempting!) semantic versioning.

PR #3538, opened 2024-11-25
Fix domain id ordering of manually specified integration entities in Python

jpdean:
The subdomain data passed when a form is created must be ordered by domain id (see [this](https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/utils.h#L337)). However, when passing manually created subdomain data to a UFL measure in Python, the data isn't sorted before creating the form, giving incorrect results.

This PR sorts the data by domain ID before creating the form and modifies a test to check for this. Alternatively, we could not sort the data and document the requirement in Python, but I think it would be easy to miss by users.


jorgensd (approved):

garth-wells (approved):

jhale:
I don't see an issue here with the internal sort - subdomains is unlikely to be very large, so it's better to be safe.

Issue #3528, opened 2024-11-22
Allow `ufl.Measure` without `subdomain_data` to be used in form with another measure with subdomain data


finsberg:
### Summarize the issue

Consider the following variational form for a diffusion problem with a source term `f` defined on some subdomain.

```python
dx_ft = ufl.dx(domain=mesh, subdomain_data=facet_tags)
dx = ufl.dx
u, v = ufl.TrialFunction(V), ufl.TestFunction(V)

du_dt = u - u_n
theta = 0.5
u_mid = theta * u + (1 - theta) * u_n

G = du_dt * v * dx + dt * ufl.dot(ufl.grad(u_mid), ufl.grad(v)) * dx  - dt * f * v * dx_ft(S1_marker)
a, L = ufl.system(G)
```
Here I have intentionally created two measures where one of the measures contain `subdomain_data`.
When solving this problem (see full example below) it yields an `AssertionError` which requires that the `subdomain_data` should be the same for all `dx` (see https://github.com/FEniCS/dolfinx/blob/3e7c708770274ecb27f03c37286f9f1de58e9d32/python/dolfinx/fem/forms.py#L245-L247).

The full backtrace is here
```
Traceback (most recent call last):
  File "/home/shared/mwe.py", line 65, in <module>
    problem = LinearProblem(
              ^^^^^^^^^^^^^^
  File "/usr/local/dolfinx-real/lib/python3.12/dist-packages/dolfinx/fem/petsc.py", line 789, in __init__
    self._L = _create_form(
              ^^^^^^^^^^^^^
  File "/usr/local/dolfinx-real/lib/python3.12/dist-packages/dolfinx/fem/forms.py", line 337, in form
    return _create_form(form)
           ^^^^^^^^^^^^^^^^^^
  File "/usr/local/dolfinx-real/lib/python3.12/dist-packages/dolfinx/fem/forms.py", line 331, in _create_form
    return _form(form)
           ^^^^^^^^^^^
  File "/usr/local/dolfinx-real/lib/python3.12/dist-packages/dolfinx/fem/forms.py", line 249, in _form
    assert all([d is data[0] for d in data])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
Exception ignored in: <function LinearProblem.__del__ at 0xffffa602c7c0>
Traceback (most recent call last):
[...]

PR #3541, opened 2024-11-28
Ignore none subdomain data in form compilation subdomain checker

jorgensd:
Resolve: https://github.com/FEniCS/dolfinx/issues/3528


garth-wells (approved):
Could a test be added?

PR #3542, opened 2024-11-29
Add `FiniteElement` python wrapper

schnellerhase:
Together with https://github.com/FEniCS/dolfinx/pull/3509 fixes https://github.com/FEniCS/dolfinx/issues/3483.


python/dolfinx/fem/element.py
```diff
@@ -160,3 +162,162 @@ def _(e: basix.finite_element.FiniteElement):
...
+        return self._cpp_object == other._cpp_object
+
+    @property
+    def dtype(self):
+        """Geometry type of the Mesh that the FunctionSpace is defined on."""
+        return self._cpp_object.dtype
+
+    @property
+    def basix_element(self):
+        """Return underlying Basix element (if it exists).
```
garth-wells, line 191:
```suggestion
        """Return underlying Basix C++ element (if it exists).
```

python/dolfinx/fem/element.py
garth-wells, line 1: Could you add return type hints?

python/dolfinx/fem/element.py
```diff
@@ -160,3 +162,162 @@ def _(e: basix.finite_element.FiniteElement):
...
+        """Points on the reference cell at which an expression needs to be evaluated in order to
+        interpolate the expression in the finite element space.
+
+        Note:
+            For Lagrange elements the points will just be the nodal positions. For other elements
+            the points will typically be the quadrature points used to evaluate moment degrees of
+            freedom.
+
+        Returns:
+            Interpolation point coordinates on the reference cell, returning the (0) coordinates
```
garth-wells, line 226: The return type is a NumPy ndarray.

python/dolfinx/fem/element.py
```diff
@@ -160,3 +162,162 @@ def _(e: basix.finite_element.FiniteElement):
...
+        Returns:
+            True if DOF transformations are required.
+        """
+        return self._cpp_object.needs_dof_transformations
+
+    @property
+    def signature(self):
+        """String identifying the finite element.
+
+        Returns:
```
garth-wells, line 275: No need for `Returns` for a property.

python/dolfinx/fem/element.py
```diff
@@ -160,3 +162,162 @@ def _(e: basix.finite_element.FiniteElement):
...
+            Element signature
+        """
+        return self._cpp_object.signature
+
+    def T_apply(self, x, cell_permutations, dim):
+        """Transform basis functions from the reference element ordering and orientation to the
```
garth-wells, line 281: Document args. Add comment that these functions are for testing only. Performance will be poor.

python/dolfinx/fem/element.py
```diff
@@ -160,3 +162,162 @@ def _(e: basix.finite_element.FiniteElement):
...
+        """Apply the inverse transpose of the operator applied by T_apply()."""
+        self._cpp_object.Tt_apply(x, cell_permutations, dim)
+
+
+def finite_element(
+    cell_type: _cpp.mesh.CellType,
+    ufl_e: ufl.finiteelement,
+    dtype: np.dtype,
+) -> FiniteElement:
+    """Create a DOLFINx element from a basix.ufl element."""
```
garth-wells, line 300: Document arguments.

python/dolfinx/fem/function.py
```diff
@@ -746,11 +721,9 @@ def ufl_function_space(self) -> ufl.FunctionSpace:
         return self
 
     @property
-    def element(
-        self,
-    ) -> typing.Union[_cpp.fem.FiniteElement_float32, _cpp.fem.FiniteElement_float64]:
+    def element(self) -> FiniteElement:
         """Function space finite element."""
-        return self._cpp_object.element  # type: ignore
+        return FiniteElement(self._cpp_object.element)
```
garth-wells, line 726: This create a new `FiniteElement` wrapper each time. Would it be better to wrap the element in the initialiser?

garth-wells (changes requested):

python/dolfinx/fem/function.py
schnellerhase: That would require an additional state variable of the wrapper class `FunctionSpace` keeping track of the `FiniteElement`, which contradicts it being a wrapper to provide access only to the `cpp_object`, as we would then have two `finite_element`'s around to manage. 

Also the `FiniteElement` object itself is only a wrapper, it performs no computations or produces additional states, so it's fine to be created whenever we expose a `FiniteElement`.

python/dolfinx/fem/element.py
```diff
@@ -160,3 +162,185 @@ def _(e: basix.finite_element.FiniteElement):
...
+    _cpp_object: typing.Union[_cpp.fem.FiniteElement_float32, _cpp.fem.FiniteElement_float64]
+
+    def __init__(
+        self,
+        cpp_object: typing.Union[_cpp.fem.FiniteElement_float32, _cpp.fem.FiniteElement_float64],
+    ):
+        """Creates a Python wrapper for the exported finite element class.
+
+        Note:
+            Do not use this constructor directly. Instead use :func:`finite_element`.
```
jhale, line 177: We're not using _ on e.g. functionspace - probably should be consistent.

python/dolfinx/fem/element.py
```diff
@@ -160,3 +162,185 @@ def _(e: basix.finite_element.FiniteElement):
...
+) -> FiniteElement:
+    """Create a DOLFINx element from a basix.ufl element.
+
+    Args:
+        cell_type: Element cell type, see `mesh.CellType`
+        ufl_e: UFL element, holding quadrature rule and other properties of the selected element.
+        dtype: Geometry type of the element.
+    """
+    if np.issubdtype(dtype, np.float32):
+        CppElement = _cpp.fem.FiniteElement_float32
```
jhale, line 325: Let's call this something consistent with the underlying class e.g. FiniteElement_dtype

On a tangential note it would be nice to automate this type of duck typing across the dtype.

python/dolfinx/fem/element.py
schnellerhase: I am not sure what you are referring to, could you elaborate further?

jhale: The function to create FiniteElement should be called finiteelement to be consistent with e.g. functionspace.

schnellerhase: I see, thanks.

python/dolfinx/fem/element.py
```diff
@@ -160,3 +162,187 @@ def _(e: basix.finite_element.FiniteElement):
...
+        globally consistent physical element ordering and orientation.
+
+        Args:
+            x: Data to transform (in place). The shape is `(m, n)`, where `m` is the number of
+            dgerees-of-freedom and the storage is row-major.
+            cell_permutations: Permutation data for the cell.
+            dim: Number of columns in `data`.
+
+        Note:
+            Exposed for testing only, expect poor performance.
```
jhale, line 277: This isn't very helpful, I'd propose:

Exposed for testing only, expect poor performance. Please see `basix.numba_helpers` for performant versions.

python/dolfinx/fem/function.py
nate-sime: Perhaps this is naive, but could https://docs.python.org/3/library/functools.html#functools.cached_property be used?

schnellerhase: Seems to add exactly what we would need here - however I am not convinced we satisfy the
> Useful for expensive computed properties of instances that are otherwise effectively immutable.

part of it. Calling the constructor of `FiniteElement` is very lightweight and would then compete with the cached lookup.

nate-sime: With non-cached approach calling the constructor each time, won't this break expected outcome of
```python
assert V.element is V.element
```
?

schnellerhase: No, the equal check is performed on the underlying `cpp_object` https://github.com/FEniCS/dolfinx/pull/3542/files#diff-93e5ad3a5a70702c7593dd13933baba17ded94b1c679cdc58129bd3e12610a3fR184

schnellerhase: However I did miss that in the other PR https://github.com/FEniCS/dolfinx/pull/3509.

schnellerhase: Ah right, `is` not the same as `==`, my bad. You are right, then this check will fail.

nate-sime: Another aspect to consider regarding `cached_property`:
> Transform a method of a class into a property whose value is computed once and then cached as a normal attribute for the life of the instance.

In the current implementation, what is the lifespan of all these instanitated `FiniteElement`s?

nate-sime: As I look through the Python layer, this question crops up a lot. Perhaps this can be addressed in a future PR.

schnellerhase: Sounds good, will file an issue when this gets merged.

jorgensd: In `dolfinx.mesh.Mesh`, we store the Python-wrappers: https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/mesh.py#L275-L278 

I guess we need to decide on a unified structure for this.

garth-wells (changes requested):
The wrapping of the C++ FiniteElement needs to be sorted out. See comment on `cached_property` in the discussion.

python/dolfinx/fem/element.py
```diff
@@ -160,3 +162,190 @@ def _(e: basix.finite_element.FiniteElement):
...
+        """
+        return self._cpp_object.value_shape
+
+    @property
+    def interpolation_points(self) -> npt.NDArray[np.floating]:
+        """Points on the reference cell at which an expression needs to be evaluated in order to
+        interpolate the expression in the finite element space.
+
+        Interpolation point coordinates on the reference cell, returning the (0) coordinates data
+        (row-major) storage and (1) the shape `(num_points, tdim)`.
```
garth-wells, line 221: Python docs use reStructured text, so need double ticks for typewriting fonts, e.g. ``(num_points, tdim)``.

Return type descriptions is wrong -  returns an ndarray, not a flat array + shape.

garth-wells (approved):

python/dolfinx/fem/element.py
```diff
@@ -160,3 +162,190 @@ def _(e: basix.finite_element.FiniteElement):
...
+    def Tt_apply(self, x: npt.NDArray[np.floating], cell_permutations: np.int32, dim: int) -> None:
+        """Apply the transpose of the operator applied by T_apply().
+
+        Args:
+            x: Data to transform (in place). The shape is ``(m, n)``, where `m` is the number of
+            dgerees-of-freedom and the storage is row-major.
+            cell_permutations: Permutation data for the cell.
+            dim: Number of columns in `data`.
+
+        Note:
```
jorgensd, line 291: @schnellerhase  This is not correct. These functions have loop access in the python wrapper.

jorgensd: The type hint should be `npt.NDArray[np.int32]`.
https://github.com/FEniCS/dolfinx/blob/fcc7719793b55da5f3736a73e109c76b114bdf8d/python/dolfinx/wrappers/fem.cpp#L187-L196

python/dolfinx/fem/element.py
```diff
@@ -160,3 +162,190 @@ def _(e: basix.finite_element.FiniteElement):
...
+            x: Data to transform (in place). The shape is ``(m, n)``, where ``m`` is the number of
+            dgerees-of-freedom and the storage is row-major.
+            cell_permutations: Permutation data for the cell.
+            dim: Number of columns in `data`.
+
+        Note:
+            Exposed for testing. Function is not vectorised across multiple cells. Please see
+            ``basix.numba_helpers`` for performant versions.
+        """
+        self._cpp_object.Tt_apply(x, cell_permutations, dim)
```
jorgensd, line 312: This is also wrong, as this should apply the inverse transform.

python/dolfinx/fem/element.py
schnellerhase: Thanks for the catches and fixes - my bad!

Issue #3482, opened 2024-10-25
Adding edge_tags and vertex_tags to gmshio


bayswiss:
### Describe new/missing feature

gmshio functions ```model_to_mesh``` and ```read_from_mesh``` return (only) the triplet ```(mesh, cell_tags, facet_tags)```, it would be very useful if it returned also ```edge_tags``` and ```vertex_tags```, like in the following: 

[https://github.com/ComputationalPhysiology/cardiac-geometriesx/blob/f7d304600972992299ef6d3938adeebd4c4e2af2/src/cardiac_geometries/utils.py#L47](url)

jhale:
Seems like a good idea - in terms of implementation we should consider using a namedtuple as the number of return arguments is getting unwieldy.

bayswiss:
The code in the link from Henrik Finsber returns indeed a named tuple: 
```python
class GMshModel(NamedTuple):
    mesh: dolfinx.mesh.Mesh
    cell_tags: dolfinx.mesh.MeshTags
    facet_tags: dolfinx.mesh.MeshTags
    edge_tags: dolfinx.mesh.MeshTags
    vertex_tags: dolfinx.mesh.MeshTags
```
After little modifications, this works flawlessly on my machine. I can try to adapt it for main dolfinx.

PR #3547, opened 2024-12-03
Extract edge tags and vertex tags from gmsh and return a NamedTuple

finsberg:
Fix https://github.com/FEniCS/dolfinx/issues/3482

Extract `edge_tags` and `vertex_tags` from gmsh if exists and return a `NamedTuple` with mesh and meshtags rather than a regular tuple. 

Another thing I also find very useful is to the extract the physical groups from gmsh. This can be done as follows
```python
if comm.rank == rank:
    gmsh.initialize()
    gmsh.model.add("Mesh from file")
    gmsh.merge(str(msh_file))
    mesh, ct, ft, et, vt = model_to_mesh(gmsh.model, comm, 0)
    markers = {
        gmsh.model.getPhysicalName(*v): tuple(reversed(v))
        for v in gmsh.model.getPhysicalGroups()
    }
    gmsh.finalize()
else:
    mesh, ct, ft, et, vt = model_to_mesh(gmsh.model, comm, 0)
    markers = {}
 markers = comm.bcast(markers, root=rank)
```
However, since we here return a `NamedTuple` we could also add these markers as another item to the `NamedTuple`. Let me know if you think it is a good idea, and I can add this as well.

### DIFF
--- a/python/demo/demo_axis.py
+++ b/python/demo/demo_axis.py
@@ -458,9 +458,7 @@ if MPI.COMM_WORLD.rank == 0:
 
 model = MPI.COMM_WORLD.bcast(model, root=0)
 partitioner = dolfinx.cpp.mesh.create_cell_partitioner(dolfinx.mesh.GhostMode.shared_facet)
-msh, cell_tags, facet_tags, _, _ = io.gmshio.model_to_mesh(
-    model, MPI.COMM_WORLD, 0, gdim=2, partitioner=partitioner
-)
+gmsh_model = io.gmshio.model_to_mesh(model, MPI.COMM_WORLD, 0, gdim=2, partitioner=partitioner)
 
 gmsh.finalize()
 MPI.COMM_WORLD.barrier()
@@ -468,13 +466,15 @@ MPI.COMM_WORLD.barrier()
 
 # Visually check of the mesh and of the subdomains using PyVista:
 
-tdim = msh.topology.dim
+tdim = gmsh_model.mesh.topology.dim
 if have_pyvista:
-    topology, cell_types, geometry = plot.vtk_mesh(msh, 2)
+    topology, cell_types, geometry = plot.vtk_mesh(gmsh_model.mesh, 2)
     grid = pyvista.UnstructuredGrid(topology, cell_types, geometry)
     plotter = pyvista.Plotter()
-    num_local_cells = msh.topology.index_map(tdim).size_local
-    grid.cell_data["Marker"] = cell_tags.values[cell_tags.indices < num_local_cells]
+    num_local_cells = gmsh_model.mesh.topology.index_map(tdim).size_local
+    grid.cell_data["Marker"] = gmsh_model.cell_tags.values[
+        gmsh_model.cell_tags.indices < num_local_cells
+    ]
     grid.set_active_scalars("Marker")
     plotter.add_mesh(grid, show_edges=True)
     plotter.view_xy()
@@ -489,15 +489,17 @@ if have_pyvista:
 # will use Lagrange elements:
 
 degree = 3
-curl_el = element("N1curl", msh.basix_cell(), degree, dtype=real_type)
-lagr_el = element("Lagrange", msh.basix_cell(), degree, dtype=real_type)
-V = fem.functionspace(msh, mixed_element([curl_el, lagr_el]))
+curl_el = element("N1curl", gmsh_model.mesh.basix_cell(), degree, dtype=real_type)
+lagr_el = element("Lagrange", gmsh_model.mesh.basix_cell(), degree, dtype=real_type)
+V = fem.functionspace(gmsh_model.mesh, mixed_element([curl_el, lagr_el]))
 
 # The integration domains of our problem are the following:
 
 # +
 # Measures for subdomains
-dx = ufl.Measure("dx", msh, subdomain_data=cell_tags, metadata={"quadrature_degree": 5})
+dx = ufl.Measure(
+    "dx", gmsh_model.mesh, subdomain_data=gmsh_model.cell_tags, metadata={"quadrature_degree": 5}
+)
 dDom = dx((au_tag, bkg_tag))
 dPml = dx(pml_tag)
 # -
@@ -509,10 +511,10 @@ n_bkg = 1  # Background refractive index
 eps_bkg = n_bkg**2  # Background relative permittivity
 eps_au = -1.0782 + 1j * 5.8089
 
-D = fem.functionspace(msh, ("DG", 0))
+D = fem.functionspace(gmsh_model.mesh, ("DG", 0))
 eps = fem.Function(D)
-au_cells = cell_tags.find(au_tag)
-bkg_cells = cell_tags.find(bkg_tag)
+au_cells = gmsh_model.cell_tags.find(au_tag)
+bkg_cells = gmsh_model.cell_tags.find(bkg_tag)
 eps.x.array[au_cells] = np.full_like(au_cells, eps_au, dtype=eps.x.array.dtype)
 eps.x.array[bkg_cells] = np.full_like(bkg_cells, eps_bkg, dtype=eps.x.array.dtype)
 eps.x.scatter_forward()
@@ -556,7 +558,7 @@ I0 = 0.5 * n_bkg / Z0  # Intensity
 # We now now define `eps_pml` and `mu_pml`:
 
 # +
-rho, z = ufl.SpatialCoordinate(msh)
+rho, z = ufl.SpatialCoordinate(gmsh_model.mesh)
 alpha = 5
 r = ufl.sqrt(rho**2 + z**2)
 
@@ -577,7 +579,7 @@ eps_pml, mu_pml = create_eps_mu(pml_coords, rho, eps_bkg, 1)
 Eh_m = fem.Function(V)
 Esh = fem.Function(V)
 
-n = ufl.FacetNormal(msh)
+n = ufl.FacetNormal(gmsh_model.mesh)
 n_3d = ufl.as_vector((n[0], n[1], 0))
 
 # Geometrical cross section of the sphere, for efficiency calculation
@@ -585,10 +587,12 @@ gcs = np.pi * radius_sph**2
 
 # Marker functions for the scattering efficiency integral
 marker = fem.Function(D)
-scatt_facets = facet_tags.find(scatt_tag)
-incident_cells = mesh.compute_incident_entities(msh.topology, scatt_facets, tdim - 1, tdim)
-msh.topology.create_connectivity(tdim, tdim)
-midpoints = mesh.compute_midpoints(msh, tdim, incident_cells)
+scatt_facets = gmsh_model.facet_tags.find(scatt_tag)
+incident_cells = mesh.compute_incident_entities(
+    gmsh_model.mesh.topology, scatt_facets, tdim - 1, tdim
+)
+gmsh_model.mesh.topology.create_connectivity(tdim, tdim)
+midpoints = mesh.compute_midpoints(gmsh_model.mesh, tdim, incident_cells)
 inner_cells = incident_cells[(midpoints[:, 0] ** 2 + midpoints[:, 1] ** 2) < (radius_scatt) ** 2]
 marker.x.array[inner_cells] = 1
 
@@ -596,7 +600,7 @@ marker.x.array[inner_cells] = 1
 dAu = dx(au_tag)
 
 # Define integration facet for the scattering efficiency
-dS = ufl.Measure("dS", msh, subdomain_data=facet_tags)
+dS = ufl.Measure("dS", gmsh_model.mesh, subdomain_data=gmsh_model.facet_tags)
 # -
 
 # We also specify a variable `phi`, corresponding to the $\phi$ angle of
@@ -620,7 +624,7 @@ dS = ufl.Measure("dS", msh, subdomain_data=facet_tags)
 phi = np.pi / 4
 
 # Initialize phase term
-phase = fem.Constant(msh, scalar_type(np.exp(1j * 0 * phi)))
+phase = fem.Constant(gmsh_model.mesh, scalar_type(np.exp(1j * 0 * phi)))
 # -
 
 # We now solve the problem:
@@ -655,7 +659,7 @@ for m in m_list:
     elif sys.hasExternalPackage("superlu_dist"):  # type: ignore
         mat_factor_backend = "superlu_dist"
     else:
-        if msh.comm.size > 1:
+        if gmsh_model.mesh.comm.size > 1:
             raise RuntimeError("This demo requires a parallel linear algebra backend.")
         else:
             mat_factor_backend = "petsc"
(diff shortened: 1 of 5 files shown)


python/dolfinx/io/gmshio.py
```diff
@@ -55,6 +55,14 @@
 }
 
 
+class GMSHModel(typing.NamedTuple):
+    mesh: Mesh
+    cell_tags: _cpp.mesh.MeshTags_int32
```
jhale, line 60: Is MeshTags Python wrapped now?

python/dolfinx/io/gmshio.py
```diff
@@ -330,7 +373,43 @@ def model_to_mesh(
...
+            mesh, tdim - 3, marked_vertices, vertex_values
+        )
+        mesh.topology.create_connectivity(topology.dim - 3, tdim)
+        adj = adjacencylist(local_entities)
+        vt = meshtags_from_entities(mesh, tdim - 3, adj, local_values.astype(np.int32, copy=False))
+        vt.name = "Vertex tags"
+    else:
+        vt = meshtags(mesh, tdim - 3, np.empty(0, dtype=np.int32), np.empty(0, dtype=np.int32))
+
+    return GMSHModel(mesh, ct, ft, et, vt)
```
jhale, line 412: GmshModel is consistent with our variable naming and Gmsh preferred capitalisation.

python/dolfinx/io/gmshio.py
```diff
@@ -330,7 +373,43 @@ def model_to_mesh(
...
+
+        local_entities, local_values = distribute_entity_data(
+            mesh, tdim - 2, marked_edges, edge_values
+        )
+        mesh.topology.create_connectivity(topology.dim - 2, tdim)
+        adj = adjacencylist(local_entities)
+        et = meshtags_from_entities(mesh, tdim - 2, adj, local_values.astype(np.int32, copy=False))
+        et.name = "Edge tags"
+    else:
+        et = meshtags(mesh, tdim - 2, np.empty(0, dtype=np.int32), np.empty(0, dtype=np.int32))
```
jhale, line 392: Is there a good reason not to use None?

python/dolfinx/io/gmshio.py
```diff
@@ -330,7 +373,43 @@ def model_to_mesh(
...
+
+        local_entities, local_values = distribute_entity_data(
+            mesh, tdim - 3, marked_vertices, vertex_values
+        )
+        mesh.topology.create_connectivity(topology.dim - 3, tdim)
+        adj = adjacencylist(local_entities)
+        vt = meshtags_from_entities(mesh, tdim - 3, adj, local_values.astype(np.int32, copy=False))
+        vt.name = "Vertex tags"
+    else:
+        vt = meshtags(mesh, tdim - 3, np.empty(0, dtype=np.int32), np.empty(0, dtype=np.int32))
```
jhale, line 410: Same comment on using None.

python/demo/demo_scattering_boundary_conditions.py
```diff
@@ -442,7 +442,7 @@ def curl_2d(f: fem.Function):
     )
 
 model = MPI.COMM_WORLD.bcast(model, root=0)
-domain, cell_tags, facet_tags = io.gmshio.model_to_mesh(model, MPI.COMM_WORLD, 0, gdim=2)
+domain, cell_tags, facet_tags, _, _ = io.gmshio.model_to_mesh(model, MPI.COMM_WORLD, 0, gdim=2)
```
jhale, line 445: We should encourage users to used the NamedTuple directly in one of these demos.

jhale (changes requested):

python/dolfinx/io/gmshio.py
finsberg: Fixed in https://github.com/FEniCS/dolfinx/pull/3547/commits/bd5d73b916fc73687a0b214d4770dbef317e73b2

python/dolfinx/io/gmshio.py
finsberg: This is a choice to be made. I think using `None` make sense and would signal that there is no physical group for a given dimension. However, it would make the return type `MeshTags | None` which would complicate the type checking further down (i.e the user need to check for `None` before using the attributes). I think both ways are OK, so I leave it to you to decide.

jhale: We're trying to remove sentinel types from the interface, i.e. return values that imply None. I'd prefer to return None. @jorgensd ?

jhale:
> Fix https://github.com/FEniCS/dolfinx/issues/3482
> 
> 
> 
> Extract `edge_tags` and `vertex_tags` from gmsh if exists and return a `NamedTuple` with mesh and meshtags rather than a regular tuple. 
> 
> 
> 
> Another thing I also find very useful is to the extract the physical groups from gmsh. This can be done as follows
> 
> ```python
> 
> if comm.rank == rank:
> 
>     gmsh.initialize()
> 
>     gmsh.model.add("Mesh from file")
> 
>     gmsh.merge(str(msh_file))
> 
>     mesh, ct, ft, et, vt = model_to_mesh(gmsh.model, comm, 0)
> 
>     markers = {
> 
>         gmsh.model.getPhysicalName(*v): tuple(reversed(v))
> 
>         for v in gmsh.model.getPhysicalGroups()
> 
>     }
> 
>     gmsh.finalize()
> 
> else:
> 
>     mesh, ct, ft, et, vt = model_to_mesh(gmsh.model, comm, 0)
> 
>     markers = {}
> 
>  markers = comm.bcast(markers, root=rank)
> 
> ```
> 
> However, since we here return a `NamedTuple` we could also add these markers as another item to the `NamedTuple`. Let me know if you think it is a good idea, and I can add this as well.

This sounds good I think - but is markers the right name? Doesn't seem to map clearly to the gmsh terminology of physical groups.

python/dolfinx/io/gmshio.py
jorgensd: I wonder if there ever is a case where one want to distinguish between having any data on a local process and there not being tags at all in the Gmsh model.

I would prefer a monotype return value, ie meshtags (if the above comment doesn’t raise any other concerns)

finsberg:
> 
> This sounds good I think - but is markers the right name? Doesn't seem to map clearly to the gmsh terminology of physical groups.

I agree that `physical_groups` is a better name

jorgensd:
> We should also allow the user to generate a mesh without any physical groups. Currently the user needs to mark the cells as physical. I added an example to `demo_gmsh.py` [here](https://github.com/FEniCS/dolfinx/pull/3547/commits/5dcb4b551f4153be87a4ff406e14c0a4cd2d801d) which currently doesn't work, but which I think should be possible to do. I can try to submit a fix for this if you agree.

I don’t think it is a good idea, ref https://github.com/FEniCS/dolfinx/issues/3332#issuecomment-2270601302

python/demo/demo_scattering_boundary_conditions.py
finsberg: Fixed in https://github.com/FEniCS/dolfinx/pull/3547/commits/1c78264b4a28a76b442c98632b05c3d7bbb44436

python/dolfinx/io/gmshio.py
```diff
@@ -114,10 +130,13 @@ def extract_topology_and_markers(model, name: typing.Optional[str] = None):
             model will be used.
 
     Returns:
-        A nested dictionary where each key corresponds to a gmsh cell
+        A tuple ``(topologies, groups)``, where ``topologies`` is a
+        nested dictionary where each key corresponds to a gmsh cell
         type. Each cell type found in the mesh has a 2D array containing
         the topology of the marked cell and a list with the
-        corresponding markers.
+        corresponding markers. ``groups`` is a dictionary where the key
```
jhale, line 137: Name `groups` to `physical_groups` here so the naming is consistent throughout.

python/dolfinx/io/gmshio.py
jhale: OK, I guess I'm unsure now. If we're sticking with this, is this behaviour documented?

python/dolfinx/io/gmshio.py
finsberg: Fixed in https://github.com/FEniCS/dolfinx/pull/3547/commits/205662e5ab99923a7727cbf558dee91c99cc204c

python/dolfinx/io/gmshio.py
```diff
@@ -264,10 +290,16 @@ def model_to_mesh(
         num_nodes = cell_information[perm_sort[-1]]["num_nodes"]
         cell_id, num_nodes = comm.bcast([cell_id, num_nodes], root=rank)
 
-        # Check for facet data and broadcast relevant info if True
+        # Check for facet, edge and vertex data and broadcast relevant info if True
         has_facet_data = False
         if tdim - 1 in cell_dimensions:
```
garth-wells, line 295: I know this is existing code that are been replicated below, but it would be better as something like
```python
has_facet_data = (tdim - 1) in cell_dimensions
```

python/dolfinx/io/gmshio.py
garth-wells: A starting point is that in parallel all processes must return the same type. 

If the Gmsh model has no tags, then the return type should be `None` because there are no tags. If the user expects tags and gets `None`, then they know something has gone wrong. Getting empty `MeshTags` because there are no tags but tags were expected is a problem.

In parallel, if I get empty tags on one process I don't know if that means that the mesh has no tags or if the portion of the mesh on the current process has no tags but that other portions might.

garth-wells (changes requested):

python/demo/demo_axis.py
```diff
@@ -458,23 +458,23 @@ def create_eps_mu(pml, rho, eps_bkg, mu_bkg):
 
 model = MPI.COMM_WORLD.bcast(model, root=0)
 partitioner = dolfinx.cpp.mesh.create_cell_partitioner(dolfinx.mesh.GhostMode.shared_facet)
-msh, cell_tags, facet_tags = io.gmshio.model_to_mesh(
-    model, MPI.COMM_WORLD, 0, gdim=2, partitioner=partitioner
-)
+gmsh_model = io.gmshio.model_to_mesh(model, MPI.COMM_WORLD, 0, gdim=2, partitioner=partitioner)
```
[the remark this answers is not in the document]
finsberg: I agree with you that it is unrelated to Gmsh. Do you have a suggestion for a different name? Some other suggestions are `grid`, `data`, `mesh_data`, `grid_data`.

python/dolfinx/io/gmshio.py
```diff
@@ -55,6 +60,15 @@
 }
 
 
+class GmshModel(typing.NamedTuple):
```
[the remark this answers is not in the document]
finsberg: I agree. Do you have a suggestion for a different name? What about,  `Grid`, `GridData` or `MeshData`. Or could use `Container` rather than `Data`.

python/dolfinx/io/gmshio.py
finsberg: I think this is better

python/demo/demo_axis.py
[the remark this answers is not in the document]
garth-wells: Maybe `meshdata` or `mesh_data`?

python/dolfinx/io/gmshio.py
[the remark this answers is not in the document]
garth-wells: `MeshData` sounds goo to me.

python/dolfinx/io/gmshio.py
finsberg: One problem with using `None` is that you would need to add checks everywhere if you want to make the type checker happy and this can make the code rather verbose. For example the following code
```python
mesh_data = gmshio.model_to_mesh(model, comm, rank=0)
mesh_data.mesh.name = name
mesh_data.cell_tags.name = f"{name}_cells"
mesh_data.facet_tags.name = f"{name}_facets"
mesh_data.edge_tags.name = f"{name}_edges"
mesh_data.vertex_tags.name = f"{name}_vertices"
```
will need to be rewritten as
```python
mesh_data = gmshio.model_to_mesh(model, comm, rank=0)
mesh_data.mesh.name = name
assert mesh_data.cell_tags is not None
mesh_data.cell_tags.name = f"{name}_cells"
assert mesh_data.facet_tags is not None
mesh_data.facet_tags.name = f"{name}_facets"
assert mesh_data.edge_tags is not None
mesh_data.edge_tags.name = f"{name}_edges"
assert mesh_data.vertex_tags is not None
mesh_data.vertex_tags.name = f"{name}_vertices"
```
to make `mypy` happy (or some other type of check). Not saying one is better than this other, but just to be aware that users downstream would need add these checks if they type check their code (which also includes the demos). 

Another option could be to add an argument `default_tag` which will be used if no tags are found, e.g `mesh_data = gmshio.model_to_mesh(model, comm, rank=0, default_tag=0)`

python/dolfinx/io/gmshio.py
finsberg: Fixed in https://github.com/FEniCS/dolfinx/pull/3547/commits/0a875880936454e9140c27e051479c30ab335062

python/dolfinx/io/gmshio.py
garth-wells: I think the checker is a secondary consideration. Don't want the tail wagging the dog :).

The asserts are not necessarily bad; if a user is expected mesh tags from the mesh, then the assertions check that they got them.

The most important point is that empty `MeshTags` and `None` are not the same thing, and points like this can be critical in parallel. If a rank has an empty `MeshTags` then other ranks may non-empty have tags. If a rank has `None`, then the mesh has no mesh tags, i.e. no ranks will have tags. Knowing this can be important for handling collective operations.

finsberg: Tags are now set to `None` if they don't exist in https://github.com/FEniCS/dolfinx/pull/3547/commits/58ae9057d4359344b1844c20f3074677ba0f4f09.

python/demo/demo_axis.py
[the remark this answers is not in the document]
finsberg: Fixed in https://github.com/FEniCS/dolfinx/pull/3547/commits/7bb809a9b2cdbbf8604b60534a6d2da8b851aed0

python/dolfinx/io/gmshio.py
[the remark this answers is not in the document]
finsberg: Fixed in https://github.com/FEniCS/dolfinx/pull/3547/commits/7bb809a9b2cdbbf8604b60534a6d2da8b851aed0

jhale:
I've reviewed all of the comments and they've been addressed satisfactorily. Small tweaks can be made in another PR if necessary.

python/dolfinx/io/gmshio.py
```diff
@@ -33,6 +33,11 @@
 ]
 
 
+class TopologyDict(typing.TypedDict):
```
garth-wells, line 36: What is this for? Should it be documented?

finsberg: It seems like `topologies` that that are defined on line 158 (and returned from `extract_topology_and_markers`) should be a dictionary with keys `topology` and `cell_data` with values being numpy arrays. In this case you can use a `TypedDict` to specify this for the type checker (see also https://peps.python.org/pep-0589/). In other words, the main difference from using a regular `dict[str, npt.NDArray[Any]]` is that you will get an error from `mypy` if you try so use a different key that what is specified in the `TypedDict`.

garth-wells: A short docstring would help :)

python/dolfinx/io/gmshio.py
```diff
@@ -55,6 +60,25 @@
 }
 
 
+class MeshData(typing.NamedTuple):
```
garth-wells, line 63: See https://github.com/FEniCS/dolfinx/blob/cee28f63f67e3ec4b4e9f0e260e0dcd8a878d8dd/python/dolfinx/fem/function.py#L548 for an example of documenting a `NamedTuple` with Sphinx.

jhale (approved):

PR #3549, opened 2024-12-03
Introduction local assembling of cell/facet form integral values 

asmeralt:
Hi all! This is my first PR to FEniCS projects, so please consider this more like a call for code review & guidance than an actual patch.

I am working on a bigger task in my research which involves using assembling functions and requires thier modification.
As a first step I decided to do small code refactoring, but end up with few questions. I'll be glad if somebody can answer them. 

1. What code style / formatting rules do you use for C++ part of the project?
2. What C++ standard are you working with? Do you have any limitations on that?
3. In particular I am not satisfied with this piece of code which repeats few times in the code
` using X = scalar_value_type_t<T>;
  std::vector<X> coordinate_dofs_cnt(3 * x_dofmap.extent(1));
  std::span<X> coordinate_dofs(coordinate_dofs_cnt);`
I propose to get rid of using `std::span` for `coordinate_dofs` variable and stick with `std::vector`.
4. I am focused on `assemble_scalar_impl.h` file, but can work on other assemblers too.
5. Building from source was kinda hell on my MacBook Pro with Apple Silicon. I have faced a few platform specific problems which were not raised in Github issues or Community Forum. Are you interested in improving building docs (I can help with build for Mac section)?

### DIFF
--- a/cpp/dolfinx/fem/assemble_scalar_impl.h
+++ b/cpp/dolfinx/fem/assemble_scalar_impl.h
@@ -55,9 +55,7 @@ T assemble_cells(mdspan2_t x_dofmap, std::span<const scalar_value_type_t<T>> x,
     return value;
 
   // Create data structures used in assembly
-  using X = scalar_value_type_t<T>;
-  std::vector<X> coordinate_dofs_cnt(3 * x_dofmap.extent(1));
-  std::span<X> coordinate_dofs(coordinate_dofs_cnt);
+  std::vector<scalar_value_type_t<T>> coordinate_dofs(3 * x_dofmap.extent(1));
 
   // Iterate over all cells
   for (std::size_t index = 0; index < cells.size(); ++index)
@@ -65,7 +63,7 @@ T assemble_cells(mdspan2_t x_dofmap, std::span<const scalar_value_type_t<T>> x,
     std::int32_t cell = cells[index];
 
     // Get cell coordinates/geometry
-    get_cell_geometry(coordinate_dofs, x_dofmap, x, cell);
+    get_cell_geometry(std::span(coordinate_dofs), x_dofmap, x, cell);
 
     const T* coeff_cell = coeffs.data() + index * cstride;
     fn(&value, coeff_cell, constants.data(), coordinate_dofs.data(),
@@ -90,9 +88,7 @@ T assemble_exterior_facets(mdspan2_t x_dofmap,
     return value;
 
   // Create data structures used in assembly
-  using X = scalar_value_type_t<T>;
-  std::vector<X> coordinate_dofs_cnt(3 * x_dofmap.extent(1));
-  std::span<X> coordinate_dofs(coordinate_dofs_cnt);
+  std::vector<scalar_value_type_t<T>> coordinate_dofs(3 * x_dofmap.extent(1));
 
   // Iterate over all facets
   const std::uint8_t step = 4;
@@ -103,7 +99,7 @@ T assemble_exterior_facets(mdspan2_t x_dofmap,
     std::int32_t local_facet = facets[index + 1];
 
     // Get cell coordinates/geometry
-    get_cell_geometry(coordinate_dofs, x_dofmap, x, cell);
+    get_cell_geometry(std::span(coordinate_dofs), x_dofmap, x, cell);
 
     // Permutations
     auto perm = get_cell_permutations(cell, local_facet, num_facets_per_cell, perms);
--- a/cpp/dolfinx/fem/assemble_scalar_impl.h
+++ b/cpp/dolfinx/fem/assemble_scalar_impl.h
@@ -15,6 +15,7 @@
 #include <dolfinx/mesh/Geometry.h>
 #include <dolfinx/mesh/Mesh.h>
 #include <dolfinx/mesh/Topology.h>
+#include <variant>
 #include <memory>
 #include <vector>
 
@@ -35,6 +36,18 @@ inline void get_cell_geometry(std::span<T> coordinate_dofs,
   }
 }
 
+template <dolfinx::scalar T>
+inline T* get_value_ptr(std::variant<std::shared_ptr<T>, std::span<T>> acc, std::size_t i)
+{
+  if (std::holds_alternative<std::shared_ptr<T>>(acc))
+  {
+    return std::get<std::shared_ptr<T>>(acc).get();
+  }
+  else if (std::holds_alternative<std::span<T>>(acc)){
+    return &(std::get<std::span<T>>(acc)[i]);
+  }
+}
+
 inline std::uint8_t get_cell_permutations(std::int32_t cell,
                                           std::int32_t local_facet,
                                           int num_facets_per_cell,
@@ -45,14 +58,17 @@ inline std::uint8_t get_cell_permutations(std::int32_t cell,
 
 /// Assemble functional over cells
 template <dolfinx::scalar T>
-T assemble_cells(mdspan2_t x_dofmap, std::span<const scalar_value_type_t<T>> x,
-                 std::span<const std::int32_t> cells, FEkernel<T> auto fn,
-                 std::span<const T> constants, std::span<const T> coeffs,
-                 int cstride)
+void assemble_cells(std::variant<std::shared_ptr<T>, std::span<T>> acc,
+                    mdspan2_t x_dofmap, std::span<const scalar_value_type_t<T>> x,
+                    std::span<const std::int32_t> cells, FEkernel<T> auto fn,
+                    std::span<const T> constants, std::span<const T> coeffs,
+                    int cstride)
 {
-  T value(0);
   if (cells.empty())
-    return value;
+    return;
+
+  if (!(std::holds_alternative<std::shared_ptr<T>>(acc) || std::holds_alternative<std::span<T>>(acc)))
+    return;
 
   // Create data structures used in assembly
   std::vector<scalar_value_type_t<T>> coordinate_dofs(3 * x_dofmap.extent(1));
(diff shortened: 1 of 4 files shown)


jorgensd:
> What code style / formatting rules do you use for C++ part of the project?

We use clang-format, with the ruleset specifed in: https://github.com/FEniCS/dolfinx/blob/main/.clang-format

> What C++ standard are you working with? Do you have any limitations on that?

As far as I recall, the lower bound we support is C++ 20 (ref: https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/CMakeLists.txt#L8)



> 3\. In particular I am not satisfied with this piece of code which repeats few times in the code
>         ` using X = scalar_value_type_t<T>;   std::vector<X> coordinate_dofs_cnt(3 * x_dofmap.extent(1));   std::span<X> coordinate_dofs(coordinate_dofs_cnt);`
>         I propose to get rid of using `std::span` for `coordinate_dofs` variable and stick with `std::vector`.

Could you elaborate on why you do not want to use `std::span` for the `coordinate_dofs`? 
I don't quite see how the PR improves that (as you now have both a vector and a span).

I see that you are pushing out duplicate code to an implicit function, which I do like though:)

> 5\. Building from source was kinda hell on my MacBook Pro with Apple Silicon. I have faced a few platform specific problems which were not raised in Github issues or Community Forum. Are you interested in improving building docs (I can help with build for Mac section)?

Quite a few developers of DOLFINx use MacBook's (I am not sure if it is the Pro with Apple Silicon though). Any feedback is appreciated, but should probably start as:
- A description of what issues you have faced, and how you have solved then (as a discourse post)

Then developers that use macs can chime in and see if they have alternative solutions to your problems.

jhale:
The reason we use std::span in many function signatures is that it is far more weakly constrained type than std::vector. For example, it does not need to own it's underlying memory. You can think of it as being a safe/modern way of passing a non-owning pointer to the first element of a contiguous array, plus a length.

jhale:
Regarding the issue b. with brew, yes, the entire stack needs to be built on the brew provided OpenMPI. If another MPI implementation or even version gets in the mix then there can of course be runtime issues.

Our GitHub action is quite precise: is there a specific command where you think the issues arise?

jorgensd:
> Regarding the issue b. with brew, yes, the entire stack needs to be built on the brew provided OpenMPI. If another MPI implementation or even version gets in the mix then there can of course be runtime issues.
> 
> Our GitHub action is quite precise: is there a specific command where you think the issues arise?

Maybe we should make a note that tells people that want to install from source to look at specific github actions?

jhale:
Already done: https://docs.fenicsproject.org/dolfinx/main/python/installation

jorgensd:
> Already done: https://docs.fenicsproject.org/dolfinx/main/python/installation

We could add a mac specific link to the CI as well? It wouldn't hurt to make it even more visible?

garth-wells (commented):
It's not clear to me what this PR is setting out to do. It looks more complicated than before, e.g. why are `std::vector`s declared and then wrapped as spans?

On style, we generally avoid `auto` unless (i) it's blindingly obvious what the type is, e.g. the return type from `std::make_shared`, or (ii) it is hard to infer the return type and/or it's very elaborate.

asmeralt:
> It's not clear to me what this PR is setting out to do. It looks more complicated than before, e.g. why are `std::vector`s declared and then wrapped as spans?
To be honest the only goal of this PR for me is to get familiar with dev practices of this project (code style, namings, PR review process, etc.) 
I thought that mooving repeating code chunks to a function would be a good task to do it.
I agree that wrapping `std::vector` in `std::span` looks redundant and more complicated than before, but that was derived by `std::span` template deduction guide discussed in this [topic](https://stackoverflow.com/questions/77021359/why-can-t-not-be-deduced-for-stdspant-when-passing-a-stdvector). 
I had three options:
1. Overload new inline function
2. Stick with `std::vector`and do some pointer arithmetic in lines 152-153. First part is ok for inline functions but pointers are questionable IMHO


3. Do as I did with explicit wrapping to `std::span` (I got rid of unnecessary variables in second commit)  


> On style, we generally avoid `auto` unless (i) it's blindingly obvious what the type is, e.g. the return type from `std::make_shared`, or (ii) it is hard to infer the return type and/or it's very elaborate.
That is exactly what I was hoping to get from this PR.
Here are three places where I introduced `auto`



I guess the third one is obvious, the others should be changed

asmeralt:
Hi all, I did not know if it is ok to start another PR so I decided to edit this one.
In my research I need to get local cell integral values of different forms (energy norms).
In legacy **dolfin** library I found an `assemble_local` method which makes what I need, but I did not found anything similar in **dolfinX.**
So I'd like to introduce my recent piece of work.
I modified 'assemble_scalar_impl.h' functions so that they can either return a single scalar value (as it is right now) or an array of local scalars computed on each cell/facet (actually it is a map of type ```std::map<std::pair<dolfinx::fem::IntegralType, int>, std::span<T>> ```.
And introduce new `integrate_scalar` function on API level in `assembler.h`
I tried to "mostly" avoid code duplication and preserve current algorithmic requirements for scallar assembling 
a. O(n) - time and O(1) - memory  for current `assemble_scalar` function
b. O(n) - time and O(n) - memory for new `integrate_scalar` function

So far I came up with few questions to you (as always more questions then answers from me).

1. What do you think in general about this functionality? Maybe it is already done and I missed it.
2. What do you think about my implementation, namings, docstring, etc.?
3. I'd like to implement some unit testing for this piece of code and will be glad to recieve any ideas, examples or test cases.

jhale:
Did you try writing a UFL form with your scalar expression tested against a cell-wise constant test function?

The majority of this patch, in its current form, is unlikely to be accepted. The overall use case is unclear to me. Variants and shared pointers have large indirection overheads and we avoid having this type of data structure inside hot loops.

jorgensd:
> Hi all, I did not know if it is ok to start another PR so I decided to edit this one. In my research I need to get local cell integral values of different forms (energy norms). In legacy **dolfin** library I found an `assemble_local` method which makes what I need, but I did not found anything similar in **dolfinX.** So I'd like to introduce my recent piece of work. I modified 'assemble_scalar_impl.h' functions so that they can either return a single scalar value (as it is right now) or an array of local scalars computed on each cell/facet (actually it is a map of type `std::map<std::pair<dolfinx::fem::IntegralType, int>, std::span<T>> `. And introduce new `integrate_scalar` function on API level in `assembler.h` I tried to "mostly" avoid code duplication and preserve current algorithmic requirements for scallar assembling a. O(n) - time and O(1) - memory for current `assemble_scalar` function b. O(n) - time and O(n) - memory for new `integrate_scalar` function
> 
> So far I came up with few questions to you (as always more questions then answers from me).
> 
> 1. What do you think in general about this functionality? Maybe it is already done and I missed it.
> 2. What do you think about my implementation, namings, docstring, etc.?
> 3. I'd like to implement some unit testing for this piece of code and will be glad to recieve any ideas, examples or test cases.
> 4. My biggest doubt is that MPI parallelization will reorder cells during assembling and thus the output array would have inconsistent oreder of scalars. Is this something plausible?

Local assembly can be found at:
https://fenicsproject.discourse.group/t/example-of-custom-kernel-for-facet-assembly/16300/2?u=dokken
https://fenicsproject.discourse.group/t/element-wise-post-processing/11837/4?u=dokken

it would help if you could explain your use-case (why do you want the integral for a single cell at the time?)
And as Jack said, integrating against a DG-0 test function, if the original form returns a scalar, is just what you want.

asmeralt:
> The majority of this patch, in its current form, is unlikely to be accepted. The overall use case is unclear to me. Variants and shared pointers have large indirection overheads and we avoid having this type of data structure inside hot loops.

I agree with you and do not to push anything. Moreover I am open for a discussion and further improvements and refactorings if you find value in this PR and ready to discuss.
I used pointer in `for-loop` without any doubts since it was already used oneline above for coeficients. `std:variant` type can be enormously large in the code for dealing with (all this `holds` checks and `get` calls), but all the typechecks are handled at a compile time (unlike virtual functions and polymorfic objects). So I decided that it one of the safest ways of implementation, but I will be glad to hear the oposite opinion (I am not that confident in C++).

> Did you try writing a UFL form with your scalar expression tested against a cell-wise constant test function?

> And as Jack said, integrating against a DG-0 test function, if the original form returns a scalar, is just what you want.

Short anwer: 
No I did not, I find this approach little hacky.
Long answer: 
My thesis is about designing of yet another aposteriore error estimator. I need to evaluate elemetwise four + one, such expressions

- Energy norm in Lagrange space
- Energy norm in Raviart-Thomas space
- Linear functional over Raviart-Thomas based function
- Custom expression of problem coeficients
- (Optional) jumps across interior facets, jumps for exterior facets in case of Neumann-Robing boundary condition



`a` - bilinear form in H^1(\Omega) Lagrange
`ad` - bilinear form in H^1(div\Omega) Raviart-Thomas
`Ld `- linear functional in H^1(div\Omega) Raviart-Thomas
`ufl.elem_pow(f, 2) / sigma) * dx`  - coeficients (can be expressed as an energy notm of `f`

jorgensd:
> No I did not, I find this approach little hacky.

The approach is not hacky, but a very good answer to your question, as long as you have originally `rank`-0 forms that you want to assemble, assembling a `rank`-1 form with DG-0 test functions is the correct and most efficient way to do this (and parallel safe), as a DG-0 basis function is always 1.

If your forms are of higher rank (rank 1 or rank 2), ways to do this without interfering with the source code has been sketched above


> Local assembly can be found at:
> https://fenicsproject.discourse.group/t/example-of-custom-kernel-for-facet-assembly/16300/2?u=dokken
> https://fenicsproject.discourse.group/t/element-wise-post-processing/11837/4?u=dokken

You can also have a look at https://gist.github.com/jorgensd/addf4aab06c6820bb2d8138dbb26c7d1
and LocalSolver by @michalhabera in: https://github.com/search?q=repo%3Amichalhabera%2Fdolfiny%20LocalSolver&type=code

jhale:
You can post code snippets inline using GitHub's three-mark markdown syntax e.g. 

 ```hello```

I honestly don't see the hackyness - it is a correct mathematical formulation that maps directly to the abstractions already present in UFL and DOLFINx.

This PR has gone very far off from the original intent to get used to the development practices of DOLFINx by re-factoring some small pieces. The original version of PR where some of the geometry data gathering was being shifted out to small functions had some merit - having some of these kernels in small reusable components would make development of custom assemblers more straightforward, but we need to be very mindful of performance degradations. At this stage I think it would be best to close this PR and begin a new one.

PR #3550, opened 2024-12-04
ci(oneapi.yml): test build with non-pip (uv) frontend

Saransh-cpp:
This PR introduces `uv` in `oneapi.yml` to ensure that the package's build generalises well and works with non-pip build frontends.

### DIFF
--- a/.github/workflows/oneapi.yml
+++ b/.github/workflows/oneapi.yml
@@ -45,6 +45,7 @@ jobs:
           apt-get -y install binutils libstdc++-14-dev
 
       - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


jhale:
The installation of uv should be done through conda, rather than the astral's own action.

jhale:
https://github.com/FEniCS/dolfinx/blob/main/python/conda-oneapi-test-env.yml

Saransh-cpp:
Thanks, fixed!

jhale:
Looks like `uv pip` doesn't support `--check-build-dependencies`, see https://docs.astral.sh/uv/pip/compatibility/#pep-517-build-isolation Remove?

PR #3551, opened 2024-12-05
Copy x into u inside SNES Solver during Jacobian computation

jhale:
SNES requires J to be evaluated at x - it is only by chance that this was the same x as updated immediately prior in the call to F, but there is no guarantee that this is the case.


michalhabera (approved):
Copy of `x` into `self.u.x.petsc_vec` could be done in the local ghosted form to avoid the extra ghostUpdate call, but it is such in many places so better address this separately later.

jhale:
Indeed and it can also be done most elegantly with our own vector updates. We should consider this when we add our own class for SNES to the public interface.

PR #3552, opened 2024-12-06
Fix regression in FiniteElement Python wrapper

jorgensd:
Introduced in: https://github.com/FEniCS/dolfinx/pull/3542

Exposed by tests in adios4dolfinx:.


nate-sime (approved):
Nice catch!

jorgensd:
> Nice catch! I'm surprised that the `Tt_inv_apply` typo/misplacement made it through the tests. Is there some way to nicely test this?

Test added that fails on main (based of failing test in adios4dolfinx).

PR #3559, opened 2024-12-11
Update demo_tnt-elements.py

jmv2009:
@mscroggs 
According to https://github.com/mscroggs/symfem/blob/main/symfem/elements/tnt.py, we need to multiply by the primary bubble function, and take the Laplacian for the interior dofs. To this end tabulation which also generates the derivatives is used.
This will aid in generating 3D TNT elements in Basix.
[Need to do some extra checks]
The numerical results do not change at this point. This is likely due to mixing in of boundary dofs into the interior dofs, but not really the reverse. Doing more checking.
Please check contribution for form (including python conventions/requirements)


python/demo/demo_tnt-elements.py
```diff
@@ -218,10 +218,12 @@ def create_tnt_quad(degree):
...
-        pts, wts = basix.make_quadrature(basix.CellType.quadrilateral, 2 * degree - 1)
-        poly = basix.tabulate_polynomials(
-            basix.PolynomialType.legendre, basix.CellType.quadrilateral, degree - 2, pts
-        )
+        pts, wts = basix.make_quadrature(basix.CellType.quadrilateral, 2 * degree)
+        u = pts[:, 0]
+        v = pts[:, 1]
+        poly = basix.polynomials.tabulate_polynomial_set(basix.CellType.quadrilateral, basix.PolynomialType.legendre, degree-2, 2, pts)
+        # this assumes the conventional [0 to 1][0 to 1] domain of the reference element, and takes the Laplacian of (1-u)*u*(1-v)*v*poly, cf https://github.com/mscroggs/symfem/blob/main/symfem/elements/tnt.py
+        poly = (poly[5]+poly[3])*(u-1)*u*(v-1)*v+2*(poly[2]*(u-1)*u*(2*v-1)+poly[1]*(v-1)*v*(2*u-1)+poly[0]*((u-1)*u+(v-1)*v))
```
jhale, line 226: Avoid reassigning to poly, use another descriptive variable name.

jhale: This won't pass ruff check ruff format steps in our CI.

jhale:
Could you tighten up the PR title and description and move the extra details to a post further down?

jmv2009:
Details:
The numerical results do not change at this point. This is because the exact definition of internal dofs does not really matter: As long as internal dof functionals are sufficiently independent from the point and edge dof functionals, the full space is spanned. 
Please check contribution for form (including python conventions/requirements). Feel free to make your own version of this.
Note: Also need to check the quadrature degree to get to exactness, as it appears to be higher. [Correct: adjusted by +1]
First documented use of new tabulate_polynomial_set python function?

Reference implementation mathematica shape function tabulation (orders of vertices and edges are different):
k=3
xt={0,1,1,0,0}
yt={0,0,1,1,0}
space=Join[MonomialList[(1+x)^k(1+y)^k],{x,y}^(k+1),{x,y}^k x y]
tst=space.Inverse[Join[Table[space/.{x->xt[[i]],y->yt[[i]]},{i,4}],Flatten[Table[Integrate[space LegendreP[n,2s-1]Sqrt[2 n +1]/.{x->xt[[i]](1-s)+xt[[i+1]]s,y->yt[[i]](1-s)+yt[[i+1]]s},{s,0,1}],{i,4},{n,0,k-1}],1],Flatten[Table[Integrate[space Laplacian[s(s-1)t(t-1)LegendreP[n1,2s-1]Sqrt[2 n1+1]LegendreP[n2,2t-1]Sqrt[2 n2 +1],{s,t}]/.{x->s,y->t},{s,0,1},{t,0,1}],{n1,0,k-2},{n2,0,k-2}],1]]]
Flatten[Table[tst,{y,0,1,1/(k+1)},{x,0,1,1/(k+1)}],0]//N//MatrixForm

The only difference with symfem is that the dof functionals do not correspond to monomials but to normalized Legendre polynomials, as per the dof functional tabulation.

jhale:
I appreciate that this is still a draft but the PR is drifting away from the original purpose of the demo - showing how a relatively simple but non-trivial element can be implemented at runtime with Basix.

jmv2009:
Yeah, this IS getting out of hand which is a good thing. Please be patient for now: I appear on track to a reference implementation for the whole TNT de Rham complex for all elements (including a fully compatible curl - prism). I have already not seen an implementation of the 0-Form TNT prism, which is already included here. Also this variant for the 0-Form for the triangle and tetrahedron may be novel. Pyramids may be more tricky. Let's see how far we get, and then see how about implementing them in Basix and/or Symfem. The advantage of doing it here first is a very simple implementation, which can be easily tested. The TNT family appears to be the most complex, and other families can be done also.

Issue #3557, opened 2024-12-09
Helmholtz demo does not solve for a plane wave as stated


jzitelli:
The python Helmholtz demo is misleading - the solution when run in complex mode is not a plane wave as stated in the introductory comments:
https://github.com/FEniCS/dolfinx/blob/main/python/demo/demo_helmholtz.py#L15-L20

I've created an alternate, complex-mode only script which solves for a plane wave propagating at an angle theta (using mixed dirichet and neumann BCs), which I've pasted below... I'm happy to clean it up and contribute as a PR with any necessary changes if there is interest.

```
# Helmholtz problem in complex mode
# The exact solution is a plane wave propagating at
# an angle theta to the positive x-axis. Chosen for comparison with
# results from Ihlenburg's book "Finite Element Analysis of Acoustic
# Scattering" p138-139.

from mpi4py import MPI

try:
    from petsc4py import PETSc

    import dolfinx

    if not dolfinx.has_petsc:
        print("This demo requires DOLFINx to be compiled with PETSc enabled.")
        exit(0)
except ModuleNotFoundError:
    print("This demo requires petsc4py.")
    exit(0)

import numpy as np

import dolfinx
import ufl
from dolfinx.fem import Function, assemble_scalar, form, functionspace, dirichletbc, locate_dofs_geometrical
from dolfinx.fem.petsc import LinearProblem
from dolfinx.io import XDMFFile
from dolfinx.mesh import create_unit_square
from ufl import dx, grad, inner, ds, SpatialCoordinate, FacetNormal, dot

# Wavenumber
k0 = 4 * np.pi

# Approximation space polynomial degree
deg = 2

# Number of elements in each direction of the mesh
n_elem = 64

msh = create_unit_square(MPI.COMM_WORLD, n_elem, n_elem)

is_complex_mode = np.issubdtype(PETSc.ScalarType, np.complexfloating)
if not is_complex_mode:
    print('you need to be in fenics complex mode!')
    exit(0)

A = 1
theta = 0.25 * np.pi / 4
wave_dir = np.array((np.cos(theta), np.sin(theta)))

# def u_exact(x):
#     return A * np.exp(1j*k0*np.dot(wave_dir, x))

# def grad_u_exact(x):
#     return 1j*k0 * u_exact(x) * wave_dir

# Test and trial function space
V = functionspace(msh, ("Lagrange", deg))

x = SpatialCoordinate(msh)

# Function space for exact solution - need it to be higher than deg
[...]

PR #3563, opened 2024-12-16
Modify demo_helmholtz.py to solve for plane wave solution in complex mode

jzitelli:
This is a minimal change addressing the issue raised in https://github.com/FEniCS/dolfinx/issues/3557

### DIFF
--- a/python/demo/demo_helmholtz.py
+++ b/python/demo/demo_helmholtz.py
@@ -38,11 +38,11 @@ import numpy as np
 
 import dolfinx
 import ufl
-from dolfinx.fem import Function, assemble_scalar, form, functionspace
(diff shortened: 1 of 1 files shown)


garth-wells:
We aim to have demos and tests run (near) identical code in real and complex cases. Otherwise testing is problematic. Can you make the real case run the same code? Setting the NumPy type should discard the complex part when computing with reals.

Also, run `ruff` to format the Python code.

jzitelli:
Thank you for the feedback, I was indeed able to consolidate the real/complex code paths.  In the real mode the script will simply solve for the real part of the complex exponential (I updated the introductory comments accordingly).  I also fixed the lint errors found using ruff.

garth-wells (approved):

garth-wells (approved):

garth-wells:
Resolves #3557

PR #3576, opened 2024-12-24
VTKHDF mesh I/O

chrisrichardson:
Adds a simple reader and writer for `Mesh` using the `VTKHDF` format.
This is currently limited to first order meshes (but can be extended to higher order).
Includes support for mixed-topology meshes.

- [x] Add float32 support in Python layer
- [x] Add Python wrapper
- [x] Add mixed-topology test fixture


jorgensd:
Why have you chosen that restriction (the restriction is not present in my Python implementation)

chrisrichardson:
@jorgensd - partly because higher order pyramids are not working in Paraview, partly just to keep it simpler for now.
I noted above that it can be an extension...

jorgensd:
> @jorgensd - partly because higher order pyramids are not working in Paraview, partly just to keep it simpler for now. I noted above that it can be an extension...

I really think we should include higher order right away.
XDMF only supports second order grids. Thus we have no API for reading in higher order meshes that are not generated through GMSH (max degree 3).

very few (if any uses pyramids atm, especially higher order ones, and if the issue is with Paraview it shouldn’t limit us).

jorgensd:
Additionally, the code complexity for higher order is not really hard (as shown in the mixed topology io branch with the Python implementation).

python/dolfinx/io/vtkhdf.py
```diff
@@ -0,0 +1,40 @@
...
+
+import numpy as np
+
+import basix
+import ufl
+from dolfinx.cpp.io import read_vtkhdf_mesh_float32, read_vtkhdf_mesh_float64, write_vtkhdf_mesh
+from dolfinx.mesh import Mesh
+
+
+def read_mesh(comm, filename, dtype=np.float64):
```
jorgensd, line 15:
```suggestion
import typing
from pathlib import Path

from mpi4py import MPI as _MPI

import numpy as np

import basix
import ufl
from dolfinx.cpp.io import read_vtkhdf_mesh_float32, read_vtkhdf_mesh_float64, write_vtkhdf_mesh
from dolfinx.mesh import Mesh


def read_mesh(comm: _MPI.Comm, filename: typing.Union[str, Path], dtype=np.float64):
    """Read mesh data from file.
    
    Args:
        comm: MPI-communicator to distribute mesh over.
        filename: Path to input file.
        dtype: Data-type of mesh geometry.
    """
```
Can we not determine the dtype when looking for a specific variable with HDF5?

python/dolfinx/io/vtkhdf.py
```diff
@@ -0,0 +1,40 @@
...
+                cell_types[0].name,
+                cell_degree,
+                basix.LagrangeVariant.equispaced,
+                shape=(mesh_cpp.geometry.dim,),
+            )
+        )
+    return Mesh(mesh_cpp, domain)
+
+
+def write_mesh(filename, mesh):
```
jorgensd, line 45:
```suggestion
def write_mesh(filename: typing.Union[str, Path],, mesh: Mesh):
    """Write a ``Mesh`` to file with VTKHDF-format.
    
    Args:
        filename: Output path.
        mesh: Mesh to output.
    """
```

python/dolfinx/io/vtkhdf.py
```diff
@@ -0,0 +1,51 @@
...
+from dolfinx.mesh import Mesh
+
+
+def read_mesh(comm, filename, dtype=np.float64):
+    """Read a mesh from a VTKHDF format file
+    Args:
+           comm: An MPI communicator.
+           filename: File to read from.
+           dtype: Scalar type of mesh geometry (need not match dtype in file)
+    """
```
jorgensd, line 21:
```suggestion
import numpy as np

import basix
import ufl
from dolfinx.cpp.io import read_vtkhdf_mesh_float32, read_vtkhdf_mesh_float64, write_vtkhdf_mesh
from dolfinx.mesh import Mesh


def read_mesh(comm: _MPI.Comm, filename: typing.Union[str, Path], dtype: npt.DTypeLike = np.float64) -> Mesh:
    """Read a mesh from a VTKHDF format file
    Args:
           comm: An MPI communicator to distribute mesh over.
           filename: File to read from.
           dtype: Scalar type of mesh geometry (need not match dtype in file)
    """
```

python/dolfinx/io/vtkhdf.py
```diff
@@ -0,0 +1,51 @@
...
+                cell_types[0].name,
+                cell_degree,
+                basix.LagrangeVariant.equispaced,
+                shape=(mesh_cpp.geometry.dim,),
+            )
+        )
+    return Mesh(mesh_cpp, domain)
+
+
+def write_mesh(filename, mesh):
```
jorgensd, line 45:
```suggestion
def write_mesh(filename: typing.Union[str, Path],, mesh: Mesh):
```

PR #3580, opened 2024-12-28
Use block preconditioner for mixed Poisson problem 

garth-wells:
Switch to a diagonal block preconditioner, with Hypre Auxiliary Space (AMS) for the $H({\rm div})$ block. AMS works in this case because the demo is 2D, in which case RT elements are just a rotation of Nedelec elements. If Hypre is not available (not configured with PETSc or running with complex floats, which are not supported by Hypre) use LU for the $H({\rm div})$ block.

For a 3D mixed Poisson demo, we would need to use the Hypre Auxiliary-space Divergence Solver (ADS) in place of AMS. This requires the 'discrete curl' operator (see  #3587).

This change fixes issues with some LU solvers and saddle point problems.

- [x] Perhaps use `ZeroBaseForm` from UFL for the `empty` block on the form `L` (see https://github.com/FEniCS/ufl/blob/eabdb001d648ae02fb2e04890899501ca471d528/ufl/form.py#L836).

### DIFF
--- a/python/demo/demo_mixed-poisson.py
+++ b/python/demo/demo_mixed-poisson.py
@@ -8,7 +8,7 @@
 #       jupytext_version: 1.14.1
 # ---
 
-# # Mixed formulation for the Poisson equation
+# # Mixed formulation of Poisson equation with a block-preconditioner/solver
 #
 # This demo illustrates how to solve Poisson equation using a mixed
 # (two-field) formulation and block-preconditioned iterative solver. In
@@ -17,7 +17,7 @@
 # * Use mixed and non-continuous finite element spaces.
 # * Set essential boundary conditions for subspaces and $H(\mathrm{div})$ spaces.
 # * Construct a blocked linear system.
-# * Use a block-preconditioned iterative linear solver.
+# * Construct a block-preconditioned iterative linear solver.
 #
 # ```{admonition} Download sources
 # :class: download
@@ -46,15 +46,16 @@
 #   \sigma \cdot n = g \quad {\rm on} \ \Gamma_{N}.
 # $$
 #
-# The same equations arise in connection with flow in porous media, and are
-# also referred to as Darcy flow. Here $n$ denotes the outward pointing normal
-# vector on the boundary. Looking at the variational form, we see that the
-# boundary condition for the flux ($\sigma \cdot n = g$) is now an essential
-# boundary condition (which should be enforced in the function space), while
-# the other boundary condition ($u = u_0$) is a natural boundary condition
-# (which should be applied to the variational form). Inserting the boundary
-# conditions, this variational problem can be phrased in the general form: find
-# $(\sigma, u) \in \Sigma_g \times V$ such that
+# The same equations arise in connection with flow in porous media, and
+# are also referred to as Darcy flow. Here $n$ denotes the outward
+# pointing normal vector on the boundary. Looking at the variational
+# form, we see that the boundary condition for the flux ($\sigma \cdot n
+# = g$) is now an essential boundary condition (which should be enforced
+# in the function space), while the other boundary condition ($u = u_0$)
+# is a natural boundary condition (which should be applied to the
+# variational form). Inserting the boundary conditions, this variational
+# problem can be phrased in the general form: find $(\sigma, u) \in
+# \Sigma_g \times V$ such that
 #
 # $$
 #    a((\sigma, u), (\tau, v)) = L((\tau, v))
@@ -72,14 +73,15 @@
 #   + \int_{\Gamma_D} u_0 \tau \cdot n  \ {\rm d} s,
 # \end{align}
 # $$
-# and $\Sigma_g = \{ \tau \in H({\rm div})$ such that $\tau \cdot n|_{\Gamma_N}
-# = g \}$ and $V = L^2(\Omega)$.
+# and $\Sigma_g = \{ \tau \in H({\rm div})$ such that $\tau \cdot
+# n|_{\Gamma_N} = g \}$ and $V = L^2(\Omega)$.
 #
-# To discretize the above formulation, two discrete function spaces $\Sigma_h
-# \subset \Sigma$ and $V_h \subset V$ are needed to form a mixed function space
-# $\Sigma_h \times V_h$. A stable choice of finite element spaces is to let
-# $\Sigma_h$ be the Brezzi-Douglas-Marini elements of polynomial order
-# $k$ and let $V_h$ be discontinuous elements of polynomial order $k-1$.
+# To discretize the above formulation, two discrete function spaces
+# $\Sigma_h \subset \Sigma$ and $V_h \subset V$ are needed to form a
+# mixed function space $\Sigma_h \times V_h$. A stable choice of finite
+# element spaces is to let $\Sigma_h$ be the Raviart-Thomas elements of
+# polynomial order $k$ and let $V_h$ be discontinuous elements of
+# polynomial order $k-1$.
 #
 # We will use the same definitions of functions and boundaries as in the
 # demo for {doc}`the Poisson equation <demo_poisson>`. These are:
@@ -92,9 +94,10 @@
 # * $f = 10\exp(-((x - 0.5)^2 + (y - 0.5)^2) / 0.02)$  (source term)
 #
 # ## Implementation
+#
+# Import the required modules:
 
 # +
-
 try:
     from petsc4py import PETSc
 
@@ -113,7 +116,9 @@ import numpy as np
 
 import dolfinx.fem.petsc
 from basix.ufl import element
-from dolfinx import default_real_type, default_scalar_type, fem, io, la, mesh
+from dolfinx import default_real_type, default_scalar_type, fem, la, mesh
+from dolfinx.fem.petsc import discrete_gradient, interpolation_matrix
+from dolfinx.mesh import CellType, create_unit_square
 from ufl import (
     Measure,
     SpatialCoordinate,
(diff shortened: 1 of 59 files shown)


python/demo/demo_mixed-poisson.py
```diff
@@ -62,37 +66,28 @@
...
+# and $\Sigma_g := \{ \tau \in H({\rm div})$ such that $\tau \cdot
+# n|_{\Gamma_N} = g \}$ and $V := L^2(\Omega)$.
 #
-# We will use the same definitions of functions and boundaries as in the
-# demo for {doc}`the Poisson equation <demo_poisson>`. These are:
+# To discretize the above formulation, two discrete function spaces
+# $\Sigma_h \subset \Sigma$ and $V_h \subset V$ are needed to form a
+# mixed function space $\Sigma_h \times V_h$. A stable choice of finite
+# element spaces is to let $\Sigma_h$ be the Raviart-Thomas elements of
+# polynomial order $k$ and let $V_h$ be discontinuous elements of
```
jorgensd, line 82:
```suggestion
# polynomial order $k$ and let $V_h$ be discontinuous Lagrange elements of
```

python/demo/demo_mixed-poisson.py
```diff
@@ -109,88 +104,296 @@
...
+)
+
+# Solution scalar (e.g., float32, complex128) and geometry (float32/64)
+# types
+dtype = dolfinx.default_scalar_type
+xdtype = dolfinx.default_real_type
+# -
 
-msh = mesh.create_unit_square(MPI.COMM_WORLD, 32, 32, mesh.CellType.quadrilateral)
+# Create mesh a mesh in two dimensions. The iterative solver constructed
```
jorgensd, line 128:
```suggestion
# Create a two-dimensional mesh. The iterative solver constructed
```

python/demo/demo_mixed-poisson.py
```diff
@@ -109,88 +104,296 @@
...
+# changes to the linear solver.
 
+# +
+msh = create_unit_square(MPI.COMM_WORLD, 48, 48, CellType.quadrilateral, dtype=xdtype)
+# -
+#
+# Here we construct compatible function spaces for the mixed Poisson
+# problem. The `V` Raviart-Thomas ($\mathbb{RT}$) space is a vector-valued $H({\rm
+# div})$ conforming space. The `W` space is a space of discontinuous
+# Lagrange function of degree `k`.
```
jorgensd, line 140: Repetition from above (Line 81-82), maybe we should remove the first mention and introduce the stable pair here?

python/demo/demo_mixed-poisson.py
```diff
@@ -109,88 +104,296 @@
...
 
+# +
 dx = Measure("dx", msh)
-a = inner(sigma, tau) * dx + inner(u, div(tau)) * dx + inner(div(sigma), v) * dx
-L = -inner(f, v) * dx
+a = [[inner(sigma, tau) * dx, inner(u, div(tau)) * dx], [inner(div(sigma), v) * dx, None]]
+L = [
+    inner(fem.Constant(msh, np.zeros(2, dtype=dtype)), tau) * dx,
+    -inner(f, v) * dx,
+]
```
jorgensd, line 186: Is there a reason for not using `MixedFunctionSpace` here, so that the user doesn't have to manually specify the block structure, and can use `ufl.extract_blocks` to get the blocked matrix and vector, as done in https://github.com/FEniCS/dolfinx/pull/3508, ref: https://docs.fenicsproject.org/dolfinx/v0.9.0/python/demos/demo_hdg.html

python/demo/demo_mixed-poisson.py
```diff
@@ -109,88 +104,296 @@
...
-    values = np.zeros((2, x.shape[1]))
-    values[1, :] = np.sin(5 * x[0])
-    return values
+# +
+f_h1 = fem.Function(V, dtype=dtype)
+f_h1.interpolate(lambda x: np.vstack((np.zeros_like(x[0]), np.sin(5 * x[0]))))
+bc_top = fem.dirichletbc(f_h1, dofs_top)
 
+f_h2 = fem.Function(V, dtype=dtype)
+f_h2.interpolate(lambda x: np.vstack((np.zeros_like(x[0]), -np.sin(5 * x[0]))))
```
jorgensd, line 223: Could we instead use a single function `f`, where we highlight the possibility to restrict the interpolation to a sub-set of cells? 
This would highlight:
1. The usage of `top_cells = dolfinx.mesh.compute_incident_entities(msh.topology, top_facets, fdim, fdim+1)`
2. Interpolation into subset of cells:
    ```python
    f.interpolate(lambda x: np.stack((...), cells0=top_cells)
    f.interpolate(lambda x: np.stack((...), cells0=bottom_cells)
    f.x.scatter_forward()
    ```

python/demo/demo_mixed-poisson.py
```diff
@@ -109,88 +104,296 @@
...
+# a_p((\sigma, u), (\tau, v))
+# = \begin{bmatrix} \int_{\Omega} \sigma \cdot \tau + (\nabla \cdot
+# \sigma) (\nabla \cdot \tau) \ {\rm d} x  & 0 \\ 0 &
+# \int_{\Omega} u \cdot v \ {\rm d} x \end{bmatrix}
+# $$
+# and assemble it into the matrix `P`:
 
+# +
+a_p00 = inner(sigma, tau) * dx + inner(div(sigma), div(tau)) * dx
+a_p11 = inner(u, v) * dx
```
jorgensd, line 268: Same comment as above wrt. extract_blocks. 
If we want to highlight both methods in 1 demo, we could keep this one as is as it is the simplest one.

python/demo/demo_mixed-poisson.py
```diff
@@ -109,88 +104,296 @@
...
+ksp = PETSc.KSP().create(msh.comm)
+ksp.setOperators(A, P)
+ksp.setMonitor(lambda ksp, its, rnorm: print(f"Iteration: {its}, residual: {rnorm}"))
+ksp.setType("gmres")
+ksp.setTolerances(rtol=1e-8)
+ksp.setGMRESRestart(100)
+# -
+
+# To apply different solvers/preconditioners to the blocks of `P`, we
+# set the preconditioner to be a `fieldsplit` (block) type and set the
```
jorgensd, line 287: Maybe reference PETSc fieldsplit here: https://petsc.org/release/manual/ksp/#sec-block-matrices

python/demo/demo_mixed-poisson.py
```diff
@@ -109,88 +104,296 @@
...
+pc_sigma = ksp_sigma.getPC()
+if PETSc.Sys().hasExternalPackage("hypre") and not np.issubdtype(dtype, np.complexfloating):
+    pc_sigma.setType("hypre")
+    pc_sigma.setHYPREType("ams")
+
+    opts = PETSc.Options()
+    opts[f"{ksp_sigma.prefix}pc_hypre_ams_cycle_type"] = 7
+    opts[f"{ksp_sigma.prefix}pc_hypre_ams_relax_times"] = 2
+
+    # Construct at set the 'discrete gradient' operator, which maps
```
jorgensd, line 320:
```suggestion
    # Construct and set the 'discrete gradient' operator, which maps
```

python/demo/demo_mixed-poisson.py
```diff
@@ -109,88 +104,296 @@
...
+    # grad H1 -> H(curl), i.e. the gradient of a scalar Lagrange space
+    # to a H(curl) space
+    V_H1 = fem.functionspace(msh, element("Lagrange", msh.basix_cell(), k, dtype=xdtype))
+    V_curl = fem.functionspace(msh, element("N1curl", msh.basix_cell(), k, dtype=xdtype))
+    G = discrete_gradient(V_H1, V_curl)  #
```
jorgensd, line 325:
```suggestion
    G = discrete_gradient(V_H1, V_curl)
```

python/demo/demo_mixed-poisson.py
```diff
@@ -109,88 +104,296 @@
...
+reason = ksp.getConvergedReason()
+assert reason > 0, f"Krylov solver has not converged {reason}."
+ksp.view()
+# -
+
+# We save the solution `u` in VTX format. DOLFINx VTX output support
+# visualisation of discontinuous Lagrange element, hence the finite
+# element field can be visualised exactly. However, VTX does not support
+# cell-wise data, so for the lowest-order `u` case we interpolate to a
+# degree 1 discontinuous space.
```
jorgensd, line 381:
```suggestion
This is not True any longer, ref: https://github.com/FEniCS/dolfinx/blob/fb0b99447b96434a48b44c4a256c29330ffac7c6/cpp/dolfinx/io/ADIOS2Writers.h#L587-L599
```
Paraview currently have some issues with visualizing DG-0 in time-dependent problems, but that is not on our implementation.

jorgensd:
@garth-wells In general a really nice addition to DOLFINx. I added some thoughts regarding the demo above.

python/demo/demo_mixed-poisson.py
garth-wells: I found `MixedFunctionSpace` confusing - sometimes we use it, sometimes we don't. I was going to register a separate issue on it. Why do we have two ways of doing the same thing? Confusion compounded by not always using the `ufl` namespace in demos to clearly distinguish between UFL and DOLFINx functions/classes (see #3532).

python/demo/demo_mixed-poisson.py
garth-wells: Ideally we would use a space that is defined only only BC boundary. How would this looks with the new mixed-domain support?

jpdean: It would look something like [this](https://github.com/jpdean/mixed_domain_demos/blob/main/neumann_bc.py). I can add it if you want?

jpdean: It would look something like [this](https://github.com/jpdean/mixed_domain_demos/blob/main/neumann_bc.py). I can add it if you want?

garth-wells: > It would look something like [this](https://github.com/jpdean/mixed_domain_demos/blob/main/neumann_bc.py). I can add it if you want?

Yes, please do.

python/demo/demo_mixed-poisson.py
jorgensd: It has not been consistently used as it was introduced/fixed right before the release of 0.9.0.
One can do things in two ways because:
1. One want the flexibility to write out the blocks yourself, and the input to assemble_matrix_block/nest is just a list/nested list of variational forms.
2. Using `MixedFunctionSpace` and `ufl.extract_blocks` massively simplifies the readability of the variational form, especially for problems with more than two blocks. (I think @jpdean has some good examples of this in: https://github.com/jpdean/mixed_domain_demos). I also use `extract_blocks` alot in my own projects, as I have more than just two blocks, and filling out all the `None` blocks consistently is utterly painful and error-prone.

jorgensd: This also implicitly links to: https://github.com/FEniCS/ufl/pull/303#issuecomment-2575598818
which would give us yet another option as to how to handle mixed/sub-mesh stuff.

python/demo/demo_mixed-poisson.py
jpdean: Since these boundary conditions are enforced strongly, you have to use the space defined over the whole mesh (since you're setting the degrees of freedom values directly). We could add a non-zero natural boundary condition to demonstrate it though.

jorgensd: I've got a generalized interpolation into normal based dof spaces at: https://fenicsproject.discourse.group/t/setting-scalar-function-for-bdm-boundary-condition-with-controlled-flux-mixed-poisson/13508/9?u=dokken
but it is currently quite verbose (the integration entity packing can now be removed as we have simplified this in DOLFINx).
However, i think we should proceed to add this to the C++ interface or as a Python convenience function.

Issue #3583, opened 2025-01-02
[BUG]: Interval refinement fails (hits assertion) on macos


schnellerhase:
### Summarize the issue

See https://github.com/schnellerhase/dolfinx/pull/39 for minimal (failing) example. 

### How to reproduce the bug

Rerun macos pipeline. 

However can not reproduce on other platform with debug build. 

### Version

main branch

### Additional information

Behavior fist observed in https://github.com/FEniCS/dolfinx/pull/3581

jhale:
Just wanted to note that I can  reproduce using OpenMPI via Homebrew.

PR #3584, opened 2025-01-02
Fix: interval refinement assertion ignores optional type

schnellerhase:
The interval refinement routine falsely ignores the `std::optional` data type of the underlying checked object, causing undefined behavior in the case of it being unassigned. 

Introduces a fix to the assertion and a (previously failing) test case. 

Fixes https://github.com/FEniCS/dolfinx/issues/3583

### DIFF
--- a/cpp/dolfinx/refinement/interval.h
+++ b/cpp/dolfinx/refinement/interval.h
@@ -176,7 +176,7 @@ compute_refinement_data(const mesh::Mesh<T>& mesh,
   }
 
   assert(cell_topology.size() == 2 * refined_cell_count);
-  assert(parent_cell->size() == (compute_parent_cell ? refined_cell_count : 0));
+  assert((!compute_parent_cell) || (parent_cell->size() == refined_cell_count));
 
   std::vector<std::int32_t> offsets(refined_cell_count + 1);
   std::ranges::generate(offsets, [i = 0]() mutable { return 2 * i++; });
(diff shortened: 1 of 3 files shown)


cpp/test/mesh/refinement/interval.cpp
```diff
@@ -239,3 +239,13 @@ TEMPLATE_TEST_CASE("Interval Refinement (parallel)",
           != v_to_e->links((center_index + 2) % 3)[0]);
   }
 }
+
+TEMPLATE_TEST_CASE("Interval uniform refinement", "[refinement][interva]",
+                   double, float)
+{
+  auto interval
+      = dolfinx::mesh::create_interval<TestType>(MPI_COMM_WORLD, 2, {0.0, 1.0});
```
garth-wells, line 247: Will such a small mesh cause problems when testing with more than one or two MPI processes?

schnellerhase: Indeed, fails for $np>3$.

schnellerhase: Seems to be a more general issue, filed https://github.com/FEniCS/dolfinx/issues/3588 to track it.

cpp/dolfinx/refinement/interval.h
```diff
@@ -176,7 +176,7 @@ compute_refinement_data(const mesh::Mesh<T>& mesh,
   }
 
   assert(cell_topology.size() == 2 * refined_cell_count);
-  assert(parent_cell->size() == (compute_parent_cell ? refined_cell_count : 0));
+  assert((!compute_parent_cell) || (parent_cell->size() == refined_cell_count));
```
garth-wells, line 179:
```suggestion
  assert(!compute_parent_cell or parent_cell->size() == refined_cell_count);
```

garth-wells (approved):

PR #3585, opened 2025-01-03
Create binary wheels against ABI-compatible mpi4py

jhale:


garth-wells:
Doesn't seem to be active, so closing.

PR #3587, opened 2025-01-06
Add discrete curl matrix operator

garth-wells:
This is required for the Hypre  Auxiliary-space Divergence Solver (ADS) solver.

- [x] Extend to support non-affine cells

### DIFF
--- a/cpp/dolfinx/fem/discreteoperators.h
+++ b/cpp/dolfinx/fem/discreteoperators.h
@@ -42,7 +42,7 @@ namespace dolfinx::fem
 /// If the finite element degree-of-freedom vectors associated with
 /// \f$u\f$ and \f$v\f$ are \f$a\f$ and \f$b\f$, respectively, then \f$b
 /// = C a\f$, which yields \f$v = \Pi_{V} \nabla \times u\f$. It
-/// essential maps that curl of a function in a degree \f$k > 1\f$
+/// essentially maps that curl of a function in a degree \f$k > 1\f$
 /// Nédélec space into a degree \f$k - 1\f$ Raviart-Thomas space.
 ///
 /// The discerete curl is typically used in constructing algebraic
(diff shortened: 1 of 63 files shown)


cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -22,6 +22,277 @@
 namespace dolfinx::fem
 {
 
+/// @brief Assemble a discrete curl operator.
+///
+/// Curl of Nedelec(k) function -> RT(k-1) function.
+///
+/// @param[in] V0
+/// @param[in] V1
+/// @param[in] mat_set A functor that sets values in a matrix
```
jorgensd, line 31: If the curl-operator is specific to the function space, do we need to send in the spaces?
Would `degree-k` be sufficient, or would the dofmap potentially be different due to the graph re-ordering (GPS)?

If we send in the spaces, what kind of checks do we need?

garth-wells: This PR isn't ready for review (hence it's a draft). PR was created to provide a cross-reference for when the discrete curl is ready.

jorgensd: Ok.

cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -22,6 +23,311 @@
...
+/// \f$V_{1}\f$ is a Raviart-Thomas (\f$H({\rm div})\f$) space of degree of
+/// at least \f$k - 1\f$, then the interpolation is exact.
+///
+/// This function builds a matrix \f$C\f$ (the 'discrete curl'), which
+/// when applied to the degrees-of-freedom of \f$u\f$ gives the
+/// degrees-of-freedom of \f$v\f$ such that \f$v = \nabla \times u\f$.
+/// If the finite element degree-of-freedom vectors associated with
+/// \f$u\f$ and \f$v\f$ are \f$a\f$ and \f$b\f$, respectively, then \f$b
+/// = C a\f$, which yields \f$v = \Pi_{V} \nabla \times u\f$. It
+/// essential maps that curl of a function in a degree \f$k > 1\f$
```
nate-sime, line 45:
```suggestion
/// essentially maps that curl of a function in a degree \f$k > 1\f$
```

PR #3590, opened 2025-01-08
More work towards supporting mixed-topology meshes

jpdean:
This PR builds on @chrisrichardson's work in [this](https://github.com/FEniCS/dolfinx/tree/chris/add-basic-mixed-mesh-demo) branch to support mixed-topology meshes.

The idea is to allow `fem::FunctionSpace` to have multiple elements/dofmaps for each cell type. Similarly, `fem::Form` now can have more than one `integral_data` for a given subdomain ID, one for each kernel type. In the assembler, we simply loop over each kernel type. There are no changes to the lowest level assemblers. Currently, it only works for cell integrals and matrix assembly, but this design should easily generalise to assembling vectors/scalars and exterior facet integrals. The design should also generalise to interior facet integrals, but this will need more work because we need to generate kernels and access the DOF maps for each compatible combination of cell types etc.

It's just a draft for now; any comments on the general design would be much appreciated. If we think this is the right approach, I can modify the other assemblers and add support for exterior facet integrals.

### DIFF
--- a/cpp/dolfinx/fem/Form.h
+++ b/cpp/dolfinx/fem/Form.h
@@ -265,14 +265,14 @@ public:
   /// @return Function to call for `tabulate_tensor`.
   std::function<void(scalar_type*, const scalar_type*, const scalar_type*,
                      const geometry_type*, const int*, const uint8_t*)>
-  kernel(IntegralType type, int i, int cell_type_idx) const
+  kernel(IntegralType type, int i, int kernel_idx) const
   {
     const auto& integrals = _integrals[static_cast<std::size_t>(type)];
     auto it = std::ranges::lower_bound(integrals, i, std::less<>{},
                                        [](const auto& a) { return a.id; });
     // FIXME Do this properly
     if (it != integrals.end() and it->id == i)
-      return (it + cell_type_idx)->kernel;
+      return (it + kernel_idx)->kernel;
     else
       throw std::runtime_error("No kernel for requested domain index.");
   }
@@ -372,16 +372,15 @@ public:
       throw std::runtime_error("No mesh entities for requested domain index.");
   }
 
-  // TODO Add cell_type_idx and return single list
   std::vector<std::int32_t> domain(IntegralType type, int i,
-                                   int cell_type_idx) const
+                                   int kernel_idx) const
   {
     const auto& integrals = _integrals[static_cast<std::size_t>(type)];
     auto it = std::ranges::lower_bound(integrals, i, std::less<>{},
                                        [](const auto& a) { return a.id; });
     if (it != integrals.end() and it->id == i)
       // FIXME Do this properly
-      return (it + cell_type_idx)->entities;
+      return (it + kernel_idx)->entities;
     else
       throw std::runtime_error("No mesh entities for requested domain index.");
   }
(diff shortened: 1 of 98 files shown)


chrisrichardson:
Seems to be crashing in sparsity pattern builder at the moment.

python/demo/demo_mixed-topology.py
```diff
@@ -0,0 +1,187 @@
...
+# FIXME: For the time being, since UFL doesn't understand mixed topology meshes,
+# we have to call mixed_topology_form instead of form.
+a_form = mixed_topology_form(a, dtype=np.float64)
+
+# Assemble the matrix
+A = assemble_matrix(a_form)
+
+# Solve
+A_scipy = A.to_scipy()
+b_scipy = np.ones(A_scipy.shape[1])
```
chrisrichardson, line 134: Can we also assemble RHS? Maybe add an issue and next PR?

chrisrichardson (approved):

python/demo/demo_mixed-topology.py
jpdean: This PR only adds matrix assembly, but adding vector / scalar assembly should be straightforward. It would be a small extra PR.

cpp/dolfinx/fem/FunctionSpace.h
```diff
@@ -44,7 +44,22 @@ class FunctionSpace
...
+      : _mesh(mesh), _dofmaps{dofmap}, _elements{element},
+        _id(boost::uuids::random_generator()()), _root_space_id(_id)
+  {
+    // Do nothing
+  }
+
+  /// @brief Create function space for given mesh, elements and
+  /// degree-of-freedom maps.
+  /// @param[in] mesh Mesh that the space is defined on.
+  /// @param[in] elements Finite element for the space.
```
garth-wells, line 56: How do we know which element belongs to which cell type?

jpdean: At the moment, they're assumed to be in the same order as `mesh.topology.cell_types`, but this relies on the user passing them in the correct order, so I'll add a check

garth-wells: That's fine, but it needs to be documented.

cpp/dolfinx/fem/FunctionSpace.h
```diff
@@ -44,12 +44,43 @@ class FunctionSpace
...
   {
     // Do nothing
   }
 
+  /// @brief Create function space for given mesh, elements and
+  /// degree-of-freedom maps.
+  /// @param[in] mesh Mesh that the space is defined on.
+  /// @param[in] elements Finite elements for the space, one for each cell type.
+  /// The elements must be ordered to be consistent with
+  /// mesh->topology()->cell_types().
```
garth-wells, line 58:
```suggestion
  /// mesh::topology::cell_types.
```

cpp/dolfinx/fem/Form.h
```diff
@@ -262,18 +262,40 @@ class Form
   /// type.
   /// @param[in] type Integral type.
   /// @param[in] i Domain identifier (index).
```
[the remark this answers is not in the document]
jpdean: This hasn't been change, but happy to add more meaningful documentation. Yes, it's just the subdomain ID. What do you think would be the best description?

garth-wells (approved):

python/demo/demo_mixed-topology.py
```diff
@@ -0,0 +1,187 @@
...
+
+xdmf += """
+    </Grid>
+  </Domain>
+</Xdmf>
+"""
+
+fd = open("mixed-mesh.xdmf", "w")
+fd.write(xdmf)
+fd.close()
```
jorgensd, line 187: This can be replaced with the new VTKHDF interface that @chris added in https://github.com/FEniCS/dolfinx/pull/3576

cpp/dolfinx/fem/Form.h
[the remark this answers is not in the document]
jorgensd: I know, it is just that now that we introduce more input indices, each one should be properly documented.
I would call it Subdomain id to link it to `ufl.Measure(...., subdomain_id=i)`.

python/demo/demo_mixed-topology.py
jpdean: Is there a way to write the solution using the stuff from this PR? I can only find `write_mesh`...

PR #3597, opened 2025-01-13
Fix mapping of coefficients and constants for form independent compilation.

jorgensd:
Initially reported at: https://fenicsproject.discourse.group/t/issue-fem-create-form-with-dg0-function-and-ufl-derivative/16718?u=dokken

When constants and/or coefficients are eliminated from a ufl-form due to differentiation, this was not taken into account when creating the maps for coefficients/constants.

This PR corrects this behavior, ensuring that all remaining coefficients in the form are provided, as well as mapping them to the correct index.

### DIFF
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -528,7 +528,7 @@ def create_form(
         original_index = form.ufcx_form.original_coefficient_positions[c]
         original_const = original_constants[original_index]
         try:
-            constants[f"w{c}"] = constant_map[original_const]._cpp_object
+            constants[f"c{c}"] = constant_map[original_const]._cpp_object
         except KeyError:
             raise RuntimeError(f"Missing constant {original_const}")
 
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -508,13 +508,29 @@ def create_form(
     for _, idomain in _subdomain_data.items():
         idomain.sort(key=lambda x: x[0])
 
-    # Extract name of ufl objects and map them to their corresponding C++ object
-    ufl_coefficients = ufl.algorithms.extract_coefficients(form.ufl_form)
-    coefficients = {
-        f"w{ufl_coefficients.index(u)}": uh._cpp_object for (u, uh) in coefficient_map.items()
-    }
-    ufl_constants = ufl.algorithms.analysis.extract_constants(form.ufl_form)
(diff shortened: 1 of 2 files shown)


python/dolfinx/fem/forms.py
```diff
@@ -508,13 +508,29 @@ def create_form(
...
+            coefficients[f"w{c}"] = coefficient_map[original_coeff]._cpp_object
+        except KeyError:
+            raise RuntimeError(f"Missing coefficient {original_coeff}")
+
+    # Extract all constants of the compiled form in correct order
+    constants = {}
+    original_constants = ufl.algorithms.analysis.extract_constants(form.ufl_form)
+    num_constants = form.ufcx_form.num_constants
+    for c in range(num_constants):
+        original_index = form.ufcx_form.original_coefficient_positions[c]
```
ordinary-slim, line 528: I am getting:
```
    original_index = form.ufcx_form.original_coefficient_positions[c]
      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
RuntimeError: cannot dereference null pointer from cdata 'int *'
```
in the following code snippet:
```python3
from dolfinx import fem, mesh
import ufl
from mpi4py import MPI
import numpy as np

def main():
    els_side = 4
    domain  = mesh.create_unit_square(MPI.COMM_WORLD, els_side, els_side)
    V   = fem.functionspace(domain, ("Lagrange", 1),)
    solution = fem.Function(V, name="sol")
    conductivity = fem.Constant(domain, 1.0)
    v = ufl.TestFunction(V)
    dx = ufl.dx
    a_ufl = conductivity * solution * v * dx
    form = ufl.derivative(a_ufl, solution)
    j_compiled = fem.compile_form(domain.comm, form)

    coefficient_map = {}
    for coeff in form.coefficients():
        coefficient_map[coeff] = coeff
    constant_map = {}
    for const in form.constants():
        constant_map[const] = const
    j_instance = fem.create_form(j_compiled,
                                 [V, V],
                                 msh=domain,
                                 subdomains={},
                                 coefficient_map = coefficient_map,
                                 constant_map = constant_map)
    A = fem.assemble_matrix(j_instance)
    print(A.data)

if __name__=="__main__":
    main()
```
It seems that this line is mixing coefficients and constants, although I am unsure.

jorgensd: Should be fixed with latest commit.

garth-wells (approved):

ordinary-slim (approved):
Works in my examples, thank you!

Issue #3599, opened 2025-01-15
Allow user to interact with BoundingBoxTree coordinates


ordinary-slim:
### Describe new/missing feature

I am working with mesh motion and would like to avoid recomputation of BoundingBoxTrees. In my case, the mesh motion is a simple translation so I would like to be able to shift the `_bbox_coordinates` of `BoundingBoxTree` by a 3D vector. The issue is that it is a private member of the class with no Python exposure.
Making the attribute `protected` instead of `private` would allow a user to inherit from the class and interact with it from there. Another option would be to expose it to Python the same way the mesh coordinates are exposed.

jhale:
The latter option (exposing the coordinates) seems like the consistent approach. The docstring needs to reflect that this does not update the underlying tree.

garth-wells:
@ordinary-slim are you working with Python or C++?

ordinary-slim:
> [@ordinary-slim](https://github.com/ordinary-slim) are you working with Python or C++?

I write high level routines in Python and low level ones in C++ as done in your library.

PR #3600, opened 2025-01-16
Expose `_bbox_coordinates` of `BoundingBoxTree` to users

ordinary-slim:
Addressing https://github.com/FEniCS/dolfinx/issues/3599. Expose `_bbox_coordinates` of `BoundingBoxTree` the same way `_x` of `Geometry` is exposed. I added a test with point-tree collisions before and after a translational motion.


ordinary-slim:
I've been profiling `compute_collisions` with `points` and I found some opportunities for speed-ups:
- Early break in `point_in_bbox`. Note that `&=` doesn't short-circuit and always evaluates the RHS so it's better to have `a &= b && c` in the same line rather than have it in two separate lines.
- Avoid calling `BoundingBoxTree.get_bbox` as it spends a good deal of time in `copy_n`. I wrote a small workaround as a lambda function in `compute_collisions` with `points` but I think it would be cleaner if `get_bbox` returned a span of const rather than a new copy.

In this snippet of code
```python3
from dolfinx import mesh, geometry
import numpy as np
from mpi4py import MPI
from line_profiler import LineProfiler

comm = MPI.COMM_WORLD
rank = comm.Get_rank()

def main():
    tdim = 3
    num_elems_side = 20
    domain = mesh.create_unit_cube(MPI.COMM_WORLD,
                                   num_elems_side, num_elems_side, num_elems_side,
                                   mesh.CellType.tetrahedron, dtype=np.float64)
    bbtree = geometry.bb_tree(domain, tdim, padding=0.0)
    rng = np.random.default_rng(0)
    num_points = 200000
    points = rng.random((num_points, 3))
    collisions = geometry.compute_collisions_points(bbtree, points)

if __name__=="__main__":
    lp = LineProfiler()
    lp_wrapper = lp(main)
    lp_wrapper()
    with open(f"profiling_rank{rank}.txt", 'w') as pf:
        lp.print_stats(stream=pf)
```
the line `collisions = geometry.compute_collisions_points(bbtree, points)` takes 2.4 seconds before these changes and 1.95 seconds after these changes.

jhale:
Could you split these changes into two PRs? The algorithmic changes need a different type of review.

cpp/dolfinx/geometry/utils.h
```diff
@@ -310,10 +311,13 @@ void _compute_collisions_point(const geometry::BoundingBoxTree<T>& tree,
 {
   std::deque<std::int32_t> stack;
   std::int32_t next = tree.num_bboxes() - 1;
+  std::span<const T> bbox_coordinates = tree.bbox_coordinates();
+  auto view_bbox = [&bbox_coordinates](std::size_t node)
```
garth-wells, line 315: I don't think this is needed - use [subspan](https://en.cppreference.com/w/cpp/container/span/subspan).

ordinary-slim:
I'll make the changes as soon as I have time, apologies for the delay.
Just leaving a note here that applying a constant shift does work as intented but rotations or reflections do not work since bboxes are defined by their lower left and upper right coordinates.

garth-wells (approved):

garth-wells (approved):
Nice work.

PR #3608, opened 2025-01-29
Add identifiable Python include directory name for Python wrappers

schnellerhase:
Currently the python module installs the header files into the (include) path  `python/dolfinx/wrappers`, which is then accessible through `dolfinx.get_include`.

This however drops the necessity/option of a name giving include path for dolfinx extension modules. For example the `array.h` file of the dolfinx wrappers is then included in a third party lib as `<array.h>` obscuring its origin. This can be for example seen in action in [dolfinx_mpc](https://github.com/jorgensd/dolfinx_mpc/blob/main/python/dolfinx_mpc/mpc.cpp#L8) and [multiphenicsx](https://github.com/multiphenics/multiphenicsx/blob/main/multiphenicsx/cpp/multiphenicsx/wrappers/fem.cpp#L8) .

This PR introduces an additional sub directory `dolfinx_wrappers` for the installed headers, so the new include for the pyhton wrapper headers adds the descriptive additional prefix `<dolfinx_wrappers/...>`.

Also, removes file `python/dolfinx/wrappers/__init__.py`, which is superseded by `python/dolfinx/__init__.py::get_include`, and moves `py_to_cpp_coeffs` into the `dolfinx_wrappers` namespace (previously in global scope).


jhale:
This looks like a sensible change to me - any comments on naming `dolfinx_wrappers/` @chrisrichardson or @garth-wells? A possible alternative could be e.g. `dolfinx_python/`.

PR #3617, opened 2025-02-03
Use strict compiler flags when building nanobind layer in Debug mode.

garth-wells:
This is now possible thanks to a change in the nanobind 2.5.0 CMake configuration.


jhale:
Looks good. We should also investigate `STABLE_ABI`, `FREE_THREADED` and the stub generation.

Issue #3135, opened 2024-04-16
Dockerfiles should use pyproject.toml to install optional dependencies


jhale:
### Describe new/missing feature

We currently attempt to include optional dependencies in our `pyproject.toml` file

https://github.com/FEniCS/dolfinx/blob/main/python/pyproject.toml

But this remains untested as the CI image builds manually type optional dependencies

 https://github.com/FEniCS/dolfinx/blob/main/docker/Dockerfile.test-env#L156

There is some discussion of this issue here with respect to adding e.g. `--only-deps` option to `pip install`: https://github.com/pypa/pip/issues/11440

### Suggested user interface

Not applicable.

jhale:
pip-tools seems to have support for creating requirements.txt files that can then be fed into `pip`.

https://github.com/jazzband/pip-tools

garth-wells:
How about we slim down the images and let the CI runner install Python optional dependencies, documentation dependencies, etc. This makes testing, pinning and un-pinning versions much easier

francesco-ballarin:
> How about we slim down the images and let the CI runner install Python optional dependencies, documentation dependencies, etc. This makes testing, pinning and un-pinning versions much easier

:+1::+1::+1:

The only possible downside I see is that, if pinning was required, one would have to do that several times, for instance:
- in the CI runner of dolfinx,
- (in the CI runner of e.g. basix or ffcx, if the same pin were needed for them too)
- in the end user dockerfile

jhale:
@francesco-ballarin The `pyproject.toml` would be the single source for specifying light Python optional requirements. We're already doing this in e.g. Basix and it works well see e.g. https://github.com/FEniCS/basix/blob/main/.github/workflows/pythonapp.yml#L42

petsc4py, mpi4py and numpy and the other heavy packages will be built inside the Dockerfile as they are now.

365Bit:
Hey, I am still new to open-source contributions and would like to help with this issue.

I have created the pull request #3618 which adds a workflow that installs the Python dependencies with the `pyproject.toml` file as in `basix`.

Regarding the slim down of the Docker images, I wanted to ask which dependencies are all optional and should be moved to `pyproject.toml`.

PR #3618, opened 2025-02-04
Workflow using pyproject.toml for installation of dependencies

365Bit:
As suggested in issue #3135, this adds a prototype workflow where the CI runner uses the `project.toml` file to install optional Python dependencies.

If the proposal is accepted, I would like to adjust the existing workflows accordingly and reduce the size of the Docker images.

### DIFF
--- a/.github/workflows/pythonapp.yml
+++ b/.github/workflows/pythonapp.yml
@@ -51,6 +51,12 @@ jobs:
         with:
           python-version: "3.12"
 
+      - name: Install FEniCS Python components
+        run: |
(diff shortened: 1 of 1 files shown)


365Bit:
I would like to, but when I try to sign up through your [Slack link](https://join.slack.com/t/fenicsproject/shared_invite/zt-1lraknsp1-6_3Js5kueDIyWgF192d3nA), I get the message that the link is no longer active.

jhale:
Thanks for making this PR - there are a number of issues as discussed in Slack so I have decided to close it.

Issue #3553, opened 2024-12-08
Consistent handling of member variables in Python wrappers


schnellerhase:
The  Python wrappers require in some cases an additional wrapping to return the appropriate Python type. This can be achieved in multiple ways, but should be consistent throughout.

For example the `FunctionSpace_float32/float64.element` function returns a `FiniteElement_float32/float64`, in the Python API however these types are not exposed directly but only the corresponding wrappers `FunctionSpace` and `FiniteElement`. This means that at some point of the (wrapped) `FunctionSpace.element` call we need to switch to the wrapped type.

The question is then, how this should be (consistently) facilitated across the Python layer. Options are
1. Python wrapper maintains a (wrapped) member variable that is exposed through the API:

    - Single time construction of wrapper (during construction of super class).
    - Requires additional state that needs to be managed/maintained - updates in cpp layer need to be propagated.
    - Currently used for example in `dolfinx.mesh.Mesh`

2. Construct wrapped object during access

    - Simple syntax in python wrapper. 
    - Produces different objects with different calls, resulting in non consistent behavior for address checking, see https://github.com/FEniCS/dolfinx/pull/3542#discussion_r1868058142

3. Use of `cached_property` to maintain a single instance without explicit management.

    - Resolves the problem of memory addresses of 2. 
    - Property accesses imply lookup - `cached_property` documentation states it should be used for "Useful for expensive computed properties of instances that are otherwise effectively immutable." - Python wrappers are however always lightweight and how correct is the immutable assumption?
    - Currently used in `dolfinx.fem.FunctionSpace`
   
To resolve this one needs to decide on a consistent style, adapt it throughout and maybe document it in the [developer resources](https://docs.fenicsproject.org/dolfinx/main/python/developer.html)

Issue #3238, opened 2024-05-31
Generate Python type stubs using nanobind (v2)


garth-wells:
Nanobind can now auto-generate a type stubs for the wrapper code. 

This has been used with Basix (https://github.com/FEniCS/basix/pull/829).

Jocho-Smith:
This issue can be closed right?

schnellerhase:
Hi @Jocho-Smith, no this issue is still accurate. Changes in https://github.com/FEniCS/dolfinx/pull/4096 have been reverted in https://github.com/FEniCS/dolfinx/pull/4100. In particular the current CI setup did not check the generated stubs adequately. https://github.com/FEniCS/dolfinx/pull/4135 addresses this, together with https://github.com/FEniCS/dolfinx/pull/4131 we can then eventually roll this out (again) in https://github.com/FEniCS/dolfinx/pull/4101.

schnellerhase:
Resolved by https://github.com/FEniCS/dolfinx/pull/4342.

PR #3624, opened 2025-02-10
Add python wrapper for `FiniteElement.basix_element`

schnellerhase:
Switches to use the Basix type for the returned element.

Ping https://github.com/FEniCS/dolfinx/issues/3553

### DIFF
--- a/python/dolfinx/fem/element.py
+++ b/python/dolfinx/fem/element.py
@@ -6,7 +6,7 @@
 """Finite elements."""
 
 import typing
-from functools import singledispatch
+from functools import cached_property, singledispatch
 
 import numpy as np
 import numpy.typing as npt
@@ -189,14 +189,14 @@ class FiniteElement:
         """Geometry type of the Mesh that the FunctionSpace is defined on."""
         return self._cpp_object.dtype
 
-    @property
+    @cached_property
     def basix_element(self) -> basix.finite_element.FiniteElement:
         """Return underlying Basix C++ element (if it exists).
 
         Raises:
             Runtime error if Basix element does not exist.
         """
-        return self._cpp_object.basix_element
+        return basix.finite_element.FiniteElement(self._cpp_object.basix_element)
 
     @property
     def num_sub_elements(self) -> int:
--- a/python/test/unit/fem/test_function_space.py
+++ b/python/test/unit/fem/test_function_space.py
@@ -250,14 +250,11 @@ def test_cell_mismatch(mesh):
 @pytest.mark.skipif(default_real_type != np.float64, reason="float32 not supported yet")
 def test_basix_element(V, W, Q, V2):
     for V_ in (V, W, V2):
-        e = V_.element.basix_element
-        assert isinstance(
-            e, (basix._basixcpp.FiniteElement_float64, basix._basixcpp.FiniteElement_float32)
-        )
+        assert isinstance(V_.element.basix_element, basix.finite_element.FiniteElement)
 
     # Mixed spaces do not yet return a basix element
     with pytest.raises(RuntimeError):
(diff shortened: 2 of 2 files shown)


python/dolfinx/fem/element.py
```diff
@@ -189,14 +189,14 @@ def dtype(self) -> np.dtype:
         """Geometry type of the Mesh that the FunctionSpace is defined on."""
         return self._cpp_object.dtype
 
-    @property
+    @cached_property
```
jhale, line 192: Cached property should be signposted in docstring as it adds hidden state to the method.

python/test/unit/fem/test_function_space.py
```diff
@@ -250,14 +250,11 @@ def test_cell_mismatch(mesh):
...
-        e = V_.element.basix_element
-        assert isinstance(
-            e, (basix._basixcpp.FiniteElement_float64, basix._basixcpp.FiniteElement_float32)
-        )
+        assert isinstance(V_.element.basix_element, basix.finite_element.FiniteElement)
 
     # Mixed spaces do not yet return a basix element
     with pytest.raises(RuntimeError):
-        e = Q.element.basix_element
+        Q.element.basix_element
```
jhale, line 257: Throwaway return argument with `_ = `

jhale (changes requested):

python/dolfinx/fem/element.py
```diff
@@ -193,14 +193,18 @@ def dtype(self) -> np.dtype:
...
 
         Raises:
             Runtime error if Basix element does not exist.
+
+        Note:
+            Cached property: Wrapper constructed on initial call and not updated on subsequent
+            calls.
         """
-        return self._cpp_object.basix_element
+        return basix.finite_element.FiniteElement(self._cpp_object.basix_element)
```
garth-wells, line 207: This doesn't feel quite right because the Basix element returned by this method is not the same Basix element object that the C++ class holds.

Where in the code do we call this method? The test that calls it doesn't really test anything so could be removed.

garth-wells (commented):
A concern/issue in the past has been how tightly we couple Basix Python objects and DOLFINx Python objetcs

python/dolfinx/fem/element.py
schnellerhase: I can't find any specific use case of it in fenics. I faced the problem with the non wrapped type, when I called `V.element.basix_element.tabulate` from the python layer and noted that it does not result in a call of `FiniteElement.tabulate (python)`, which caused some problems with typing.

garth-wells: Maybe best to remove it.

One typing, #3238 would be a good issue to tackle.

jorgensd:
> A concern/issue in the past has been how tightly we couple Basix Python objects and DOLFINx Python objetcs

We need to expose the basix element in some way (being wrapped as a basix python or c++ object), as we do not interface to all the functionality of the C++ element through `dolfinx::fem::FiniteElement`.

As of now, there is not really any other alternative than using a basix element within DOLFINx.

python/dolfinx/fem/element.py
schnellerhase: > This doesn't feel quite right because the Basix element returned by this method is not the same Basix element object that the C++ class holds.

Could you explain why that is the case, I don't fully understand the problem at hand I believe.

garth-wells: > > This doesn't feel quite right because the Basix element returned by this method is not the same Basix element object that the C++ class holds.
> 
> Could you explain why that is the case, I don't fully understand the problem at hand I believe.

Actually, I think it's ok. I thought it was making a copy of the finite element, but it's just holding a reference to it.

PR #3625, opened 2025-02-10
Move `Expression::eval` to a free function with other 'assemblers'

### LINKED ISSUE #3620: Re-factor `fem::Expression` interface
There are number of issues with the `fem::Expression` design which could be fixed by re-factoring.
- The `Expression` class is a mixture data structure and computation, which is something we aim to avoid in the DOLFINx design. This would be largely fixed by changing `fem::Expression::eval` to be a free-function.
- Mixing data structure and computation is causing some circular dependencies w.r.t `fem/utils.h`.
- It should be made possible, and easy, to execute an `Expression` on different 'coefficient' data. Making `fem::Expression::eval` a free-function might achieve this.
- It should be possible to not re-pack coefficient data if the user wants to manage this process and exploit data re-use.

garth-wells:
Resolves #3620.

See #3620 for summary of the issues. Also improves documentation, uses multi-dimensional arrays on the Python side (Python is not C!).

TODO:
- [x] Pack Expression constants and coefficients on the Python side (consistent with assemblers)


cpp/dolfinx/fem/Expression.h
```diff
@@ -8,30 +8,34 @@
...
-/// @brief Represents a mathematical expression evaluated at a
-/// pre-defined set of points on the reference cell.
+template <dolfinx::scalar T, std::floating_point U>
+class Function;
+
+/// @brief An Expression represents a mathematical expression evaluated
+/// at a pre-defined set of points on a reference cell.
 ///
-//// This class closely follows the concept of a UFC Expression.
+/// An Expression can be evaluated using ::assemble_expression.
```
jorgensd, line 30: Wouldn't `tabulate_expression` be more appropriate? For me assembly indicates some summation into a tensor.

cpp/dolfinx/fem/assemble_expression_impl.h
```diff
@@ -0,0 +1,126 @@
+// Copyright (C) 2019-2020 Garth N. Wells
```
jorgensd, line 1: Wrong year in copyright header?

cpp/dolfinx/fem/assemble_expression_impl.h
```diff
@@ -0,0 +1,126 @@
...
+{
+  namespace md = MDSPAN_IMPL_STANDARD_NAMESPACE;
+
+  std::size_t estride;
+  if (mesh.topology()->dim() == Xshape[1])
+    estride = 1;
+  else if (mesh.topology()->dim() == Xshape[1] + 1)
+    estride = 2;
+  else
+    throw std::runtime_error("Invalid dimension of evaluation points.");
```
jorgensd, line 54: Since we are doing this rewrite. Should we think about how one could tabulate an expression on an interior facet?
The abstraction with checking the shape of `X` does not work, as an interior facet would have the same dimension for the reference points. 
Do we need a similar enum to `IntegralType`, or just use it?

garth-wells: Let's leave this to another PR.

cpp/dolfinx/fem/assemble_expression_impl.h
```diff
@@ -0,0 +1,126 @@
...
+  std::span<const std::uint32_t> cell_info;
+  std::function<void(std::span<T>, std::span<const std::uint32_t>, std::int32_t,
+                     int)>
+      post_dof_transform
+      = [](std::span<T>, std::span<const std::uint32_t>, std::int32_t, int)
+  {
+    // Do nothing
+  };
+
+  if (V)
```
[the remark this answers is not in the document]
garth-wells: Addressed in the new docstrings.

cpp/dolfinx/fem/assemble_expression_impl.h
```diff
@@ -0,0 +1,203 @@
...
+#include <dolfinx/common/IndexMap.h>
+#include <dolfinx/mesh/Geometry.h>
+#include <dolfinx/mesh/Mesh.h>
+#include <dolfinx/mesh/Topology.h>
+#include <memory>
+#include <vector>
+
+namespace dolfinx::fem::impl
+{
+/// @brief Tabulate an Expression at points.
```
jorgensd, line 23: Remove double brief?

cpp/dolfinx/fem/assemble_expression_impl.h
```diff
@@ -0,0 +1,201 @@
...
+///
+/// Executes an Expression kernel over a list of mesh entities.
+///
+/// @tparam estride 1 for cell expressions, 2 for facets.
+/// @tparam T Scalar type of expression.
+/// @tparam U Geometry type.
+/// @param[in,out] values Array in which the tabulated expressions are
+/// set. If should have shape (`entities.size(), Xshape[0], value_size,
```
jorgensd, line 31: I guess this should be `entities.size() / estride` ?

cpp/dolfinx/fem/assemble_expression_impl.h
```diff
@@ -0,0 +1,201 @@
...
+/// a second-order tensor in 2D.
+/// @param[in] num_argument_dofs Dimension of an argument function.
+/// Greater than 1 when computing a 1-form expression, e.g. can be used
+/// to create a matrix that when applied to a degree-of-freedom vector
+/// gives the expression values at the evaluation points.
+/// @param[in] x_dofmap Geometry degree-of-freedom map.
+/// @param[in] x Geometry coordinate of the mesh.
+/// @param[in] coeffs Coefficient data that appears in the expression.
+/// Usually packed using fem::pack_coefficients.
+/// @param[in] cstride Coefficient stride. Coefficient data for teh ith
```
jorgensd, line 48:
```suggestion
/// @param[in] cstride Coefficient stride. Coefficient data for the ith
```

cpp/dolfinx/fem/assemble_expression_impl.h
```diff
@@ -0,0 +1,201 @@
...
+/// entity starts at `coeffs.data() + i*cstride`.
+/// @param[in] constants Constant (coefficient) data that appears in
+/// expression. Usually packed using em::pack_constants.
```
jorgensd, line 51:
```suggestion
/// expression. Usually packed using fem::pack_constants.
```

cpp/dolfinx/fem/assemble_expression_impl.h
```diff
@@ -0,0 +1,201 @@
...
+}
+
+/// @brief Tabulate an Expression at points.
+///
+/// Executes an Expression kernel over a list of mesh entities.
+///
+/// @tparam T Scalar type
+/// @tparam U Geometry type
+/// @param[in,out] values Array in which the tabulated expressions are
+/// set. If `V` is available, it should have shape (`entities.size(),
```
jorgensd, line 120: Same comment as above.

cpp/dolfinx/fem/assemble_expression_impl.h
```diff
@@ -0,0 +1,201 @@
...
+/// @param[in] Xshape Shape `(num_points, geometric dimensions)` of
+/// points array at which the expression is evaluated by the kernel.
+/// @param[in] value_size Value size of the evaluated expression at a
+/// point, e.g. 1 for a scalar field and 3 for a vector field in 3D.
+/// @param[in] coeffs Coefficient data that appears in the expression.
+/// Usually packed using fem::pack_coefficients.
+/// @param[in] cstride Coefficient stride. Coefficient data for teh ith
+/// entity starts at `coeffs.data() + i*cstride`.
+/// @param[in] constant_data Constant (coefficient) data that appears in
+/// expression. Usually packed using em::pack_constants.
```
jorgensd, line 134:
```suggestion
/// expression. Usually packed using fem::pack_constants.
```

cpp/dolfinx/fem/assembler.h
```diff
@@ -20,15 +23,115 @@
...
+{
+  auto [X, Xshape] = e.X();
+  impl::tabulate_expression(values, e.kernel(), Xshape, e.value_size(), coeffs,
+                            cstride, constants, mesh, entities, element);
+}
+
+/// @brief Evaluate an Expression on cells or facets.
+/// @tparam T Scalar type.
+/// @tparam U Geometry type
+/// @param[in,out] values Array to fil with computed values. Row major
```
jorgensd, line 83:
```suggestion
/// @param[in,out] values Array to fill with computed values. Row major
```

jorgensd (commented):
Only some typos. Otherwise this looks good to me.

jorgensd:
@garth-wells the latest change that you introduced to entities will make it way harder to extend this to interior facets, and differs from how we treat entities in other assemblies. Please consider reverting it.

garth-wells:
> @garth-wells the latest change that you introduced to entities will make it way harder to extend this to interior facets, and differs from how we treat entities in other assemblies. Please consider reverting it.

What in the change will make it harder?

jorgensd:
> > @garth-wells the latest change that you introduced to entities will make it way harder to extend this to interior facets, and differs from how we treat entities in other assemblies. Please consider reverting it.
> 
> What in the change will make it harder?

*Summary of current inconsistency*:
- We use `FetchCells` in normal packing of coefficients as well as in expression
- We currently use a flat structure for entitites of all types:
  - cell integrals: (num_entities, )
  - exterior facets: (2* num_entities, )
  - interior facets: (4 * num_entities, )
- The original argument for this pre-dates the introduction of `mdspan`, when we wanted all flat data-structures to ensure data locality and native structures.

*Solution*:
We rework cell integrals to be rank one mdspans, exterior facets rank 2 and interior facets rank 3 (num_entities, 2, 2) across all functions, rather than having fetch_cell.

jorgensd:
> > @garth-wells the latest change that you introduced to entities will make it way harder to extend this to interior facets, and differs from how we treat entities in other assemblies. Please consider reverting it.
> 
> What in the change will make it harder?

To be more clear, I am happy with the proposed change, if we also use the same change within: https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/pack.h#L99-L168

garth-wells:
> To be more clear, I am happy with the proposed change, if we also use the same change within: https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/pack.h#L99-L168

That change is already in the PR, see https://github.com/FEniCS/dolfinx/blob/187c58b79017ed2e21e26af10e4ed8418fbec569/cpp/dolfinx/fem/pack.h#L93.

PR #3632, opened 2025-02-14
Mixed topology: find dS cells

chrisrichardson:
Adds a utility to compute the lists of pairs of cells for `dS` integrals, in the mixed-topology case.

`mesh::compute_mixed_cell_pairs(const Topology& topology, mesh::CellType facet_type)`

returns a list for each possible combination of cells, containing the cell indices and local facet indices for each pair.

### DIFF
--- a/cpp/dolfinx/mesh/Topology.cpp
+++ b/cpp/dolfinx/mesh/Topology.cpp
@@ -1411,7 +1411,7 @@ mesh::entities_to_index(const Topology& topology, int dim,
   return indices;
 }
 //-----------------------------------------------------------------------------
-std::vector<std::vector<std::pair<std::int32_t, std::int32_t>>>
+std::vector<std::vector<std::int32_t>>
 mesh::compute_mixed_cell_pairs(const Topology& topology,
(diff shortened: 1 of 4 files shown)


cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -1411,3 +1411,85 @@ mesh::entities_to_index(const Topology& topology, int dim,
   return indices;
 }
 //-----------------------------------------------------------------------------
+std::vector<std::vector<std::int32_t>>
+mesh::compute_mixed_cell_pairs(const Topology& topology,
+                               mesh::CellType facet_type)
+{
+  int tdim = topology.dim();
+  std::vector<mesh::CellType> cell_types = topology.entity_types(tdim);
+  std::vector<mesh::CellType> facet_types = topology.entity_types(tdim - 1);
```
mscroggs, line 1420: Do we want to generalise this for other sub-entities? I think all we'd need to change is replacing `tdim - 1` with `dim(facet_type)` everywhere and update the docstring

mscroggs (approved):

PR #3635, opened 2025-02-16
Unpin cffi version (fixes tests for Python 3.13)

### LINKED ISSUE #3340: Unpin cffi from version < 1.17

garth-wells:
The CFFI version was pinned to <1.17 due to the bug https://github.com/python-cffi/cffi/issues/110, which hasn't made its way into a CFFI release. However, older CFFI releases do not support Python 3.13.

This changes removes the CFFI version pin, and for CFFI version 1.17.0/1 adds skips to tests that fail due to the CFFI bug. 

Resolves #3340.

### DIFF
--- a/python/demo/demo_static-condensation.py
+++ b/python/demo/demo_static-condensation.py
@@ -133,7 +133,9 @@ kernel10 = getattr(ufcx10.form_integrals[0], f"tabulate_tensor_{np.dtype(PETSc.S
 ffi = cffi.FFI()
 if np.issubdtype(PETSc.ScalarType, np.complexfloating):
     if cffi.__version_info__ > (1, 16, 99) and cffi.__version_info__ <= (1, 17, 1):
-        print("CFFI 1.17.0 and 1.17.1 has a bug for complex type. Exiting.")
+        print(
+            "CFFI 1.17.0 and 1.17.1 has a bug for complex type. See https://github.com/FEniCS/dolfinx/pull/3635. Exiting."
+        )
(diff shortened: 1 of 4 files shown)


python/demo/demo_static-condensation.py
```diff
@@ -130,8 +129,13 @@ def sigma_u(u):
 ufcx10, _, _ = ffcx_jit(msh.comm, a10, form_compiler_options={"scalar_type": PETSc.ScalarType})  # type: ignore
 kernel10 = getattr(ufcx10.form_integrals[0], f"tabulate_tensor_{np.dtype(PETSc.ScalarType).name}")  # type: ignore
 
+
 ffi = cffi.FFI()
-cffi_support.register_type(ffi.typeof("double _Complex"), numba.types.complex128)
+if np.issubdtype(PETSc.ScalarType, np.complexfloating):
+    if cffi.__version_info__ > (1, 16, 99) and cffi.__version_info__ <= (1, 17, 1):
```
michalhabera, line 135: Can you add comment with the link to the cffi bug report?

michalhabera (approved):

PR #3641, opened 2025-02-20
SNES solver interface for standard, block and nest systems

jorgensd:
It is non-trivial to use snes (#3507, #3551), especially with NEST and Block structures.

As we already have these in tests, I see no reason for us not having an API that is accessible and easily extendible.

This adds a solver interface to DOLFINx for snes.


python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,339 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        """The MPI communicator associated with the problem"""
+        return self._comm
+
+    @property
+    def L(self) -> Form:
+        """Compiled linear form (the residual form)"""
+        return self._L
+
+    @property
+    def a(self) -> Form:
```
jhale, line 1189: Similar method for the compiled preconditioner?

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,339 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        """
+        self.u.x.petsc_vec.copy(x)
+        x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+
+    def F(self, snes: PETSc.SNES, x: PETSc.Vec, F: PETSc.Vec):
+        """Assemble the residual F into the vector b.
+
+        Args:
+            snes: The solver instance
+            x: The vector containing the latest solution
```
jhale, line 1222: This wording isn't correct - this is the point at which SNES is asking you to evaluate the residual.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,339 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        dolfinx.fem.petsc.set_bc(F, self.bcs, x, -1.0)
+
+    def J(self, snes, x, J, P):
+        """Assemble the Jacobian matrix.
+
+        Args:
+            snes: The solver instance
+            x: The vector containing the latest solution
+            J: Matrix to assemble the Jacobian into
+            P: Matrix to assemble the preconditioner into
```
jhale, line 1242: What happens if the preconditioner depends on `x`? This is infact almost always the case.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,339 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        self.u.x.petsc_vec.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+        with F.localForm() as f_local:
+            f_local.set(0.0)
+        dolfinx.fem.petsc.assemble_vector(F, self.L)
+        dolfinx.fem.petsc.apply_lifting(F, [self.a], bcs=[self.bcs], x0=[x], alpha=-1.0)
+        F.ghostUpdate(addv=PETSc.InsertMode.ADD, mode=PETSc.ScatterMode.REVERSE)
+        dolfinx.fem.petsc.set_bc(F, self.bcs, x, -1.0)
+
+    def J(self, snes, x, J, P):
+        """Assemble the Jacobian matrix.
```
jhale, line 1236: and preconditioner?

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +79,102 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
 
         """
         super().setP(P, Pmat)
+
+
+class SNESSolver:
```
jhale, line 84: Why wrap SNESSolver? We don't do this for e.g. KSP.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,344 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        self,
+        F: typing.Union[dolfinx.fem.Form, ufl.form.Form],
+        u: dolfinx.fem.Function,
+        J: typing.Optional[typing.Union[dolfinx.fem.Form, ufl.form.Form]] = None,
+        P: typing.Optional[typing.Union[dolfinx.fem.Form, ufl.form.Form]] = None,
+        bcs: typing.Optional[list[dolfinx.fem.DirichletBC]] = None,
+        form_compiler_options: typing.Optional[dict] = None,
+        jit_options: typing.Optional[dict] = None,
+    ):
+        """Class for constructing the residual and Jacobian constructors for a SNES problem.
```
garth-wells, line 1137: A second paragraph summarising what the class solver, etc, would be helpful.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,344 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+
+        Args:
+            F: The PDE residual F(u, v)
```
garth-wells, line 1140: `F(u, v)` doesn't mean much in isolation. A description of what the class does, and in the argument docs refer to what the argument is the math formulation.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,344 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+            u: The unknown
+            bcs: List of Dirichlet boundary conditions
+            J: UFL representation of the Jacobian (Optional)
```
garth-wells, line 1143: No need to mentioned 'optional' (that's in the type hint), but do need to explain what happens if it is s`None`.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,344 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+            P: UFL representation of a preconditioner (Optional)
```
garth-wells, line 1144: See above comment.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,344 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        )
+        self._a = _create_form(
+            J, form_compiler_options=form_compiler_options, jit_options=jit_options
+        )
+
+        self.bcs = bcs
+        self.u = u
+
+    @property
+    def comm(self):
```
garth-wells, line 1179: Is this property needed? If yes, should it come from the SNES object?

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,344 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        """
+        self.u.x.petsc_vec.copy(x)
+        x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+
+    def F(self, snes: PETSc.SNES, x: PETSc.Vec, F: PETSc.Vec):
+        """Assemble the residual F into the vector b.
+
+        Args:
+            snes: The solver instance
+            x: The vector containing the latest solution
```
jorgensd, line 1222:
```suggestion
            x: The vector containing the point to evaluate the residual at.
```

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,344 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        self.u.x.petsc_vec.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+        with F.localForm() as f_local:
+            f_local.set(0.0)
+        dolfinx.fem.petsc.assemble_vector(F, self.L)
+        dolfinx.fem.petsc.apply_lifting(F, [self.a], bcs=[self.bcs], x0=[x], alpha=-1.0)
+        F.ghostUpdate(addv=PETSc.InsertMode.ADD, mode=PETSc.ScatterMode.REVERSE)
+        dolfinx.fem.petsc.set_bc(F, self.bcs, x, -1.0)
+
+    def J(self, snes, x, J, P):
+        """Assemble the Jacobian matrix.
```
jorgensd, line 1236:
```suggestion
        """Assemble the Jacobian matrix and preconditioner.
```

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,344 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        self.bcs = bcs
+        self.u = u
+
+    @property
+    def comm(self):
+        """The MPI communicator associated with the problem"""
+        return self._comm
+
+    @property
+    def L(self) -> Form:
```
garth-wells, line 1184: Docs need to make clear how `a` and `L` relate to `F` and `J`.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,344 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        """Compiled bilinear form (the Jacobian form)"""
+        return self._a
+
+    @property
+    def P(self) -> Form:
+        """Compiled bilinear form (the preconditioner form)"""
+        return self._a_prec
+
+    def update_solution(self, x: PETSc.Vec):
+        """Update the solution vector in the function used in the residual and Jacobian.
```
garth-wells, line 1199: Add math description of what the method does.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,344 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+
+        Args:
+            x: The vector containing the latest solution
+        """
+        x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+        x.copy(self.u.x.petsc_vec)
```
garth-wells, line 1205: Why is there a copy made? Should the Function be called `replace`? The two ghost updates look weird. Are they necessary?

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,344 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        """Copy the solution vector from the function used in the residual and Jacobian.
+
+        Args:
+            x: The vector containing the latest solution
+        """
+        self.u.x.petsc_vec.copy(x)
+        x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+
+    def F(self, snes: PETSc.SNES, x: PETSc.Vec, F: PETSc.Vec):
+        """Assemble the residual F into the vector b.
```
garth-wells, line 1218: Not sure about relevance of `b` because `b` doesn't appear anywhere in the args.

python/dolfinx/nls/petsc.py
jorgensd: We wrap `KSP` in `LinearProblem`.

garth-wells (changes requested):

python/dolfinx/fem/petsc.py
jorgensd: Should be `F`.

jorgensd, line 1218:
```suggestion
        """Assemble the residual evaluated at ``x`` into the vector ``F``.
```

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,344 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        dolfinx.fem.petsc.assemble_matrix(J, self.a, self.bcs)
+        J.assemble()
+
+        if self._a_prec is not None:
+            P.zeroEntries()
+            dolfinx.fem.petsc.assemble_matrix(P, self._a_prec, self.bcs, diagonal=1.0)
+            P.assemble()
+
+
+class BlockedSNESProblem(SNESProblem):
```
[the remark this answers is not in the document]
jorgensd: The constructor becomes incredibly complicated if we have to handle all cases (standard, blocked and nest) at the same time, as they create different data structures, requires different ways of transferring solutions from the `dolfinx.fem.Function` and the `PETSc.Vec`, as seen below.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,344 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        ]
+        _cpp.la.petsc.scatter_local_vectors(
+            x,
+            u_petsc,
+            index_maps,
+        )
+        x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+
+
+class NestSNESProblem(BlockedSNESProblem):
```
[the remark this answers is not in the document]
jorgensd: I don't see how it can be one class, as one can see by the functions below, they vastly differ for the different structures. If we only want one class, one would have to do something similar with `create_matrix`, `create_matrix_nest`, `create_matrix_block`, `assemble_vector`, `assemble_vector_block` etc.

python/dolfinx/fem/petsc.py
jorgensd: It is needed to construct the SNES object.
We don't easily store the communicator within compiled forms, especially when it can be a form, a list of forms or a nested list of forms.

python/dolfinx/nls/petsc.py
jorgensd: It also matches how we have split the `NewtonSolver` into two instances. You have a pure solver class, doing the algorithmic thing, getting information from  `NonLinearProblem`

python/dolfinx/fem/petsc.py
jorgensd: I've changed `update_solution` to replace solution. Yes, both updates are needed (The solver fails to converge otherwise in the tests).

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,345 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+
+        Args:
+            x: Data that is inserted into the solution vector
+        """
+        x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+        x.copy(self.u.x.petsc_vec)
+        self.u.x.petsc_vec.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+
+    def copy_solution(self, x: PETSc.Vec):
+        """Copy the solution for the unknown `u` into the vector `x`.
```
jorgensd, line 1213:
```suggestion
        """Copy the data in `u` into the vector `x`.
```

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,344 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        """Update the solution vector in the function used in the residual and Jacobian.
+
+        Args:
+            x: The vector containing the latest solution
+        """
+        x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+        x.copy(self.u.x.petsc_vec)
+        self.u.x.petsc_vec.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+
+    def copy_solution(self, x: PETSc.Vec):
```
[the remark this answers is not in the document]
jorgensd: I've updated the docstring to reflect that it updates an internal variable.

jorgensd:
I've tried to address all comments. 
I do not see how we can easily make a maintainable class for the various snes solver/PETSc matrix/vector instances.
We have a clear split between `assemble_vector`, `assemble_vector_block` and `assemble_vector_nest` which means that this is mirrored in the solver instance. 

This PR add no new functionality, it just tries to unify what we already have inside our test-suite, where we use SNES with standard, blocked and nest matrices. Thus, even if the diff makes it look like there is more code in this PR, it is simply documenting and generalizing the various steps required to use snes.

python/dolfinx/fem/petsc.py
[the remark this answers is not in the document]
jorgensd: How would you differentiate between a nest and block instance of this class, as the input to them would be the same structure? I.e. a list of residuals and list of unknowns.

we also do not really use these factories for pure python classes, we only use them to attach the correct c++ object. 

If you sketch out how you would do that, on a branch, I am happy to review and consider it.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,342 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        ]
+        _cpp.la.petsc.scatter_local_vectors(
+            x,
+            u_petsc,
+            index_maps,
+        )
+        x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+
+
+class NestSNESProblem(BlockedSNESProblem):
```
michalhabera, line 1392: Why does it inherit from `BlockedSNESProblem`?

python/dolfinx/nls/petsc.py
michalhabera: I don't like this split ala legacy code into a FooSolver and FooProblem classes. I never understood why this is needed and always confused me. I think of one FooProblem class that has the algorithmic active `.solve()` etc methods.

jorgensd: I'm fine with merging them (also in `NonLinearProblem`/`NewtonSolver` if everyone else is fine with it.
Seems like a history reason (@garth-wells can probably give some context: https://github.com/FEniCS/dolfinx/blob/089c89b1e3746e6b875d98a2a31ebc2775e191a4/src/kernel/nls/NonlinearProblem.cpp) 
which in turn was mirrored in: https://github.com/FEniCS/dolfinx/pull/1478

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,342 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        ]
+        _cpp.la.petsc.scatter_local_vectors(
+            x,
+            u_petsc,
+            index_maps,
+        )
+        x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+
+
+class NestSNESProblem(BlockSNESProblem):
```
michalhabera, line 1392: I see your point with repeated constructor, but in that case I think a cleaner solution is to refactor the common logic into a free function that does the Jacobian AD and call it in both problem types. Otherwise `class NestSNESProblem(BlockSNESProblem)` is not the right relation,  nest problem is not a special type of block problem.

michalhabera (commented):
I like that the existing SNES wrappers that are in tests anyways will get exposed to users. I see also how tempting it is to ask for a single SNESProblem, but I agree with Jorgen that there the first step would probably be to unify assembly functions.

python/dolfinx/nls/petsc.py
garth-wells: The C++ Newton solver was never properly reviewed with dolfin -> dolfinx. It shouldn't necessarily be taken as a good design model to follow.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,342 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        """Class for constructing the residual and Jacobian constructors for a SNES problem.
+
+        Solves problems of the form :math:`F_i(u, v) = 0, i=0,...N\\ \\forall v \\in V`
+        where :math:`u=(u_0,...,u_N), v=(v_0,...,v_n)` using PETSc SNES as the non-linear solver.
+
+        Args:
+            F: List of PDE residuals :math:`F_i`.
+            u: List of unknowns of the block system.
+            bcs: List of Dirichlet boundary conditions.
+            J: Rectangular array of bi-linear forms representing the
```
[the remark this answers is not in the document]
jorgensd: We do support this (as PETSc supports it).
I tested it locally by modifying the demo to do:
```python
   a_block = form(
            [
                [derivative(F0, u, du), derivative(F0, p, dp)],
                [derivative(F1, u, du), derivative(F1, p, dp)],
                [derivative(F1, u, du), derivative(F1, p, dp)],
            ]
        )
A = assemble_matrix_block(a_block)
print(A.size)
```
returning `(351, 198)`

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,342 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+            x: The vector containing the point to evaluate at
+            J: Matrix to assemble the Jacobian into
+            P: Matrix to assemble the preconditioner into
+        """
+        # Copy existing soultion into the function used in the residual and Jacobian
+        self.replace_solution(x)
+
+        # Assemble Jacobian
+        J.zeroEntries()
+        dolfinx.fem.petsc.assemble_matrix(J, self.a, self.bcs)
```
francesco-ballarin, line 1244: For consistency you may want to add a `diagonal=1.0` even here, since you have it the call below.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1121,342 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        """Assemble the Jacobian matrix.
+
+        Args:
+            snes: The solver instance
+            x: The vector containing the latest solution
+            J: Matrix to assemble the Jacobian into
+            P: Matrix to assemble the preconditioner into
+        """
+        self.replace_solution(x)
+        assert x.getType() != "nest" and J.getType() != "nest" and P.getType() != "nest"
```
francesco-ballarin, line 1347: Split into 3 asserts, as in `def F`?

python/dolfinx/fem/petsc.py
jorgensd: Inheritance removed

python/dolfinx/fem/petsc.py
jorgensd: @michalhabera I've now removed the inheritance, and instead introduced a `typing.Protocol` that illustrates what a class that is passed into `SNESSolver` expects.

python/dolfinx/fem/petsc.py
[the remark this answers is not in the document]
garth-wells: It's not necessary to distinguish between nest, block, etc, versions of a class because the interface is the same. When creating a class, an option can be passed to indicate what type of linear algebra structures should be used. The methods can then dispatch appropriately. Take for example PETSc nested matrices. PETSc has a function to created a nested matrix, and the new nested matrix is still a `Mat`. It's not a "`MatNest`".

The design pattern with factories has nothing to do with the C++ backend. Python doesn't support overloading, so a common pattern is factory functions that do the prep work for a common backend, e.g. how many NumPy function work.

python/dolfinx/fem/petsc.py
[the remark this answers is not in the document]
garth-wells: I don't follow `FooSNESProblem` is a class, `create_matrix_nest`, etc are functions.

jhale:
I would like to forget all this 'NonlinearProblem' and `NewtonSolver` baggage for the moment - it's not clear that it is the best thing to emulate, we wrote it so long ago and a lot of the thinking has changed.

In terms of design goals, what I would like to see is a more functional approach. Why does everything need to be in classes? SNES already has a perfectly fine class design. As a user I have UFL F and J (and maybe P) what I want back from a factory function are new assembly functions matching the interface that SNES expects. Then with those functions, I want to pass them up to another function which constructs a native SNES solver setup in the correct way, so that I can continue to modify and solve using the native SNES API.

jorgensd:
> I would like to forget all this 'NonlinearProblem' and `NewtonSolver` baggage for the moment - it's not clear that it is the best thing to emulate, we wrote it so long ago and a lot of the thinking has changed.
> 
> In terms of design goals, what I would like to see is a more functional approach. Why does everything need to be in classes? SNES already has a perfectly fine class design. As a user I have UFL F and J (and maybe P) what I want back from a factory function are new assembly functions matching the interface that SNES expects. Then with those functions, I want to pass them up to another function which constructs a native SNES solver setup in the correct way, so that I can continue to modify and solve using the native SNES API.

I feel like I have now approach it in that way, by stating that there is only **one** SNESSolver, which does not take in a specific class, but a `typing.Protocol` that the user has to follow.

python/dolfinx/nls/petsc.py
jorgensd: Factory function has been added here. There is now only one solver class, that accepts any kind of input that follows the typing protocol above. This makes it as general as I can imagine.

python/dolfinx/fem/petsc.py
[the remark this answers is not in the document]
jorgensd: I have now done this within `SNESSolver`, which only has a single point of entry, with a SNESType as input.

python/dolfinx/fem/petsc.py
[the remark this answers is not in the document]
jorgensd: There is now no inheritance and the `SNESSolver` only expects that a class provide a certain set of functions.

jorgensd:
Suggested "unified approach" using a factory can be found at: https://github.com/FEniCS/dolfinx/pull/3648/

PR #3642, opened 2025-02-20
Changes to type system

schnellerhase:
Contains multiple changes to the type system.

1. Generalizes `dolfinx::scalar` concept to allow for general `std::complex<T>` types (before only `float` and `double` supported).
2. Renames trait `dolfinx::scalar_value_type` to `dolfinx::scalar_value` and its short notation `scalar_value_type_t` to `scalar_value_t` (removes duplicate 'type'). Also renames internal trait variable from `value_type` to `type` to highlight different implications. Also, switches to usage of short notation when feasible.
3. ~Introduces `DofMapSpan` type in `types.h` in favor of multiple definitions of `mdspan2_t` types. (This caused undefined definition errors when including assembler implementation files)~
4. Introduce namespace alias `md` for `MDSPAN_IMPL_STANDARD_NAMESPACE` - following style chosen in https://github.com/FEniCS/basix/pull/902  (should one globally rename `MDSPAN_IMPL_STANDARD_NAMESPACE` to `md` with this?)

### DIFF
--- a/cpp/dolfinx/common/types.h
+++ b/cpp/dolfinx/common/types.h
@@ -16,7 +16,7 @@ namespace dolfinx
 /// point real or complex types. Note that this concept is different to
 /// std::floating_point which does not include std::complex.
 template <class T>
-concept scalar = std::is_floating_point_v<T>
+concept scalar = std::floating_point<T>
                  || std::is_same_v<T, std::complex<typename T::value_type>>;
 
 /// @private These structs are used to get the float/value type from a
--- a/cpp/dolfinx/common/types.h
+++ b/cpp/dolfinx/common/types.h
@@ -10,6 +10,8 @@
 #include <concepts>
 #include <type_traits>
 
+#include <basix/mdspan.hpp>
+
 namespace dolfinx
 {
 /// @private This concept is used to constrain the a template type to floating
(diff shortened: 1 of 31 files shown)


cpp/dolfinx/common/types.h
```diff
@@ -10,30 +10,36 @@
 #include <concepts>
 #include <type_traits>
```
garth-wells, line 12:
```suggestion
```

cpp/dolfinx/common/types.h
```diff
@@ -10,30 +10,36 @@
...
-  typedef typename T::value_type value_type;
+  typedef typename T::value_type type;
 };
 /// @private Convenience typedef
 template <scalar T>
-using scalar_value_type_t = typename scalar_value_type<T>::value_type;
+using scalar_value_t = typename scalar_value<T>::type;
+
+/// @private mdspan/mdarray namespace
+namespace md = MDSPAN_IMPL_STANDARD_NAMESPACE;
```
garth-wells, line 43: Any issues with exposing this globally through inclusion in a header file?

garth-wells (commented):
Nice. Any issues/down sides to exposing the `md` namespace  globally?

cpp/dolfinx/common/types.h
schnellerhase: Maybe that we introduce a transitive dependency on the `mdspan.hpp` header by including `types.h` now. Besides that it is still contained in `dolfinx::` so no other global side effects should be exposed.

garth-wells (approved):

Issue #3371, opened 2024-08-28
Rewrite entity maps to be `MeshTags`


jorgensd:
### Describe new/missing feature

Instead of having `entity_maps` as a dictionary, it should be a `dolfinx::mesh::MeshTags` object, 
where the topology of the `MeshTag` is the mesh we are mapping to, the indices the entities in the input mesh topology, while the values are the entity number in the mesh we are mapping from.

This means that inverting the map is as easy as switching order of `indices` and `values` (and attaching the correct mesh topology), and we avoid storing `-1` indices (that currently are resulting in segfaults whenever there are issues with user input, making it very hard to debug). 

We would also no longer need a dictionary for the maps, as the topology is stored as part of the MeshTag, and we can create a topology -> list_index_in_input_entity_tags.

Internally in DOLFINx, we could unravel these to avoid calling `find` many times (similar to what we do with Dirichlet BC dofs), but then with explicit handling of the negative/non-existent entities.

Also note that creating the "parent_to_sub_mesh" map is a simpler operation than its inverse (as we do not have to sort the input entities to `entity_map`.

### Suggested user interface

```python3
mesh = .....
sub_indices = ....
submesh, sub_to_parent, _, _ = dolfinx.mesh.create_submesh(mesh, dim, entities)
entity_map = dolfinx.mesh.meshtags(submesh, dim, np.arange(len(sub_to_parent), dtype=np.int32), sub_to_parent)

# Given a parent cell `c` find cell in submesh
sub_cell = entity_map.find(c)

# Invert map
parent_sort = np.argsort(sub_to_parent)
sub_cells = np.arange(len(sub_to_parent), dtype=np.int32)[parent_sort]
inverted_map = dolfinx.mesh.meshtags(mesh, dim, sub_to_parent[parent_sort], sub_cells)

# Given a child cell `d` find parent_cell
p_cell = inverted_map.find(d)
```

jorgensd:
Alternative comment from @jpdean:
we could try doing, 
`entity_maps: list[meshtags]`
```python
coeff_arg_mesh = ...
coeff_arg_topology = coeff_arg_mesh.topology
coeff_arg_tdim = coeff_arg_topology.dim
entity_tag = dolfinx.mesh.meshtags(coeff_arg_mesh, coeff_arg_tdim, coeff_arg_entities, integration_entities)
```
Example:
```python
#J1
sub_cells = np.arange(len(sub_to_parent), dtype=np.int32)
[...]

Issue #3598, opened 2025-01-13
Make `Form::domain` functions consistent


jpdean:
As pointed out by @jorgensd, currently there are some inconsistencies in the `Form::domain` functions. [One](https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/Form.h#L350) simply returns a `span`, whilst [the other ](https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/Form.h#L369) does some computation (to map entities from one mesh to another) and returns a `vector`. Since all the data computed on is immutable, we should probably compute it once in `Form`'s constructor and store it. Then, both `domain` functions can return a `span` and we avoid unnecessary computation.

PR #3646, opened 2025-02-23
Performance fixes and simplifications for `fem::Form`

### LINKED ISSUE #3644: Reduce and fix `fem::Form::domain` methods
`fem::Form` has four `domain` methods, which is confusing, and they handle memory differently. Some return a std:span and some return a copy of a `std::vector`. This easily leads to memory errors, and in low-level assembly code we have large `std::vector` copies being made.

The number of methods should be reduced, methods with very different memory handling should not have the same name, and copies inside the assembly code should be avoided.

garth-wells:
`fem::Form` had gotten unmanageably  complex, which contributed to some performance bugs (e.g., #3644) and #3645. This change substantially simplifies `fem::Form` without removing functionality, gets it in decent shape for handling 'mixed topology' meshes, and fixes some performance bugs where data was copy returned in performance critical code sections and data was repeatedly re-computed.

The `fem::Form` now always requires a `Mesh` argument. The clears up various ambiguities and possible sources of error.  There are more improvements that can follow this PR. The bugs in `fem::Form` were exposed by improvements that were being made elsewhere.

Resolves #3644 and #3598. See https://github.com/FEniCS/dolfinx/pull/3645#issuecomment-2676765039 for one example of performance improvement.

### DIFF
--- a/cpp/dolfinx/fem/Form.h
+++ b/cpp/dolfinx/fem/Form.h
@@ -42,9 +42,13 @@ enum class IntegralType : std::int8_t
 
 namespace impl
 {
-/// @brief Compute the list of entity indices in `mesh` for the ith
-/// integral (kernel) of a given type (i.e. cell, exterior facet, or
-/// interior facet).
+/// @brief Compute "`entity_map[entities[i]]`", with suitable handling
+/// for facets.
+/// @param entities
+/// @param entity_map
+/// @param codim
+/// @param c_to_f
+/// @return entity_map[entities[i]]
 std::vector<std::int32_t> compute_domain(
     auto entities, std::span<const std::int32_t> entity_map,
     std::optional<int> codim = std::nullopt,
@@ -188,13 +192,13 @@ public:
   /// @param[in] needs_facet_permutations Set to `true` is any of the
   /// integration kernels require cell permutation data.
   /// @param[in] entity_maps If any trial functions, test functions, or
-  /// coefficients in the form are not defined over the same mesh as the
-  /// integration domain, `entity_maps` must be supplied. For each key
-  /// (a mesh, different to the integration domain mesh) a map should be
-  /// provided relating the entities in the integration domain mesh to
-  /// the entities in the key mesh e.g. for a pair (msh, emap) in
-  /// `entity_maps`, `emap[i]` is the entity in `msh` corresponding to
-  /// entity `i` in the integration domain mesh.
+  /// coefficients in the form are not defined on `mesh` (the
+  /// 'integration domain'),`entity_maps` must be supplied. For each key
+  /// (a mesh, which different to `mesh`) an array map must be provided
+  /// which relates the entities in `mesh` to the entities in the key
+  /// mesh e.g. for a key/value pair `(mesh0, emap)` in `entity_maps`,
+  /// `emap[i]` is the entity in `mesh0` corresponding to entity `i` in
+  /// `mesh`.
   /// @param[in] mesh Mesh of the domain to integrate over (the
   /// 'integration domain').
   ///
@@ -361,11 +365,11 @@ public:
   ///
   /// bilinear form = 2, linear form = 1, functional = 0, etc.
   ///
-  /// @return The rank of the form
+  /// @return The rank of the form.
   int rank() const { return _function_spaces.size(); }
 
-  /// @brief Extract common mesh for the form.
-  /// @return The mesh.
+  /// @brief Common mesh for the form (the 'integration domain').
+  /// @return The integration domain mesh.
   std::shared_ptr<const mesh::Mesh<geometry_type>> mesh() const
   {
     return _mesh;
@@ -379,10 +383,9 @@ public:
     return _function_spaces;
   }
 
-  /// @brief Get the kernel function for integral `i` on given domain
-  /// type.
+  /// @brief Get the kernel function for an integral.
   /// @param[in] type Integral type.
-  /// @param[in] id The subdomain ID.
+  /// @param[in] id Integral subdomain ID.
   /// @param[in] kernel_idx Index of the kernel (we may have multiple
   /// kernels for a given ID in mixed-topology meshes).
   /// @return Function to call for `tabulate_tensor`.
@@ -472,9 +475,9 @@ public:
   ///
   /// @param[in] type Integral type.
   /// @param[in] id Integral ID, i.e. (sub)domain index.
-  /// @param[in] kernel_idx Index of the kernel (we may have multiple kernels
-  /// for a given ID in mixed-topology meshes).
-  /// @return List of active entities for the given integral (kernel).
+  /// @param[in] kernel_idx Index of the kernel (we may have multiple
+  /// kernels for a given ID in mixed-topology meshes).
+  /// @return Entity indices in the 'integration domain' mesh.
   std::span<const std::int32_t> domain(IntegralType type, int id,
                                        int kernel_idx) const
   {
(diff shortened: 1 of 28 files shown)


jorgensd:
This is a good step towards resolving: https://github.com/FEniCS/dolfinx/issues/3598

cpp/dolfinx/fem/Form.h
```diff
@@ -39,17 +41,109 @@ enum class IntegralType : std::int8_t
...
+  const int tdim = topology.dim();
+
+  std::vector<std::int32_t> mapped_entities;
+  mapped_entities.reserve(entities.size());
+  switch (type)
+  {
+  case IntegralType::cell:
+  {
+    std::ranges::transform(entities, std::back_inserter(mapped_entities),
+                           [&entity_map](auto e) { return entity_map[e]; });
```
jorgensd, line 64: Until we have resolved https://github.com/FEniCS/dolfinx/issues/3371 we need to have a check that `entity_map` does not map to a negative number, as this would cause a segfault later in the code.

cpp/dolfinx/fem/Form.h
```diff
@@ -39,17 +41,109 @@ enum class IntegralType : std::int8_t
...
+  }
+  case IntegralType::exterior_facet:
+  {
+    if (codim == 0)
+    {
+      for (std::size_t i = 0; i < entities.size(); i += 2)
+      {
+        // Add cell and the local facet index
+        mapped_entities.insert(mapped_entities.end(),
+                               {entity_map[entities[i]], entities[i + 1]});
```
jorgensd, line 75: Same comment as above.

cpp/dolfinx/fem/Form.h
```diff
@@ -39,17 +41,109 @@ enum class IntegralType : std::int8_t
...
+      // in `mesh`, so we need to get the facet number from the (cell,
+      // local_facet pair) first.
+      auto c_to_f = topology.connectivity(tdim, tdim - 1);
+      assert(c_to_f);
+      for (std::size_t i = 0; i < entities.size(); i += 2)
+      {
+        // Get the facet index, and add cell and the local facet index
+        std::int32_t facet = c_to_f->links(entities[i])[entities[i + 1]];
+        mapped_entities.insert(mapped_entities.end(),
+                               {entity_map[facet], entities[i + 1]});
```
jorgensd, line 90: Is this correct?
Say we have an exterior facet integral `(cell_0, facet_index_0)` which should be mapped onto a submesh of codim-1. Then the integration entity on the "submesh" is a cell integral for that mesh, and thus the "local facet index" (entities[i+1])" doesn;t really make sense to store. @jpdean am I missing some logic here?

garth-wells: Code is unchanged from `main`, https://github.com/FEniCS/dolfinx/blob/ead49d0fdd57d2273162ede3484505fa77981001/cpp/dolfinx/fem/Form.h#L511.

cpp/dolfinx/fem/Form.h
garth-wells: Why would it be negative?

This code is unchanged from `main`,  it's just just been moved into a different function: https://github.com/FEniCS/dolfinx/blob/ead49d0fdd57d2273162ede3484505fa77981001/cpp/dolfinx/fem/Form.h#L428

cpp/dolfinx/fem/Form.h
jorgensd: I am not stating that you have introduced a bug. What I am trying to say is that I don't think we need to add entities[i+1]  for codim 1, and that we could send in 0. I'll let @jpdean comment if he agrees.

cpp/dolfinx/fem/Form.h
```diff
@@ -188,44 +219,137 @@ class Form
...
       {
-        throw std::runtime_error("Integral IDs not sorted");
+        for (auto& [key, integral] : _integrals)
+          vdata.insert({key, std::span(integral.entities)});
       }
+      else
+      {
+        auto it = entity_maps.find(mesh0);
+        if (it == entity_maps.end())
+          throw std::runtime_error("No entity map for the mesh.");
```
jpdean, line 271: This was already checked above, so maybe we could remove above check or just `assert` here?

cpp/dolfinx/fem/Form.h
```diff
@@ -188,44 +219,137 @@ class Form
...
+        if (auto mesh0 = coefficients.at(c)->function_space()->mesh();
+            mesh0 == _mesh)
+        {
+          _cdata.insert({{type, id, c}, std::span(integral.entities)});
+        }
+        else
+        {
+          auto it = entity_maps.find(mesh0);
+          if (it == entity_maps.end())
+            throw std::runtime_error("No entity map for the mesh.");
```
jpdean, line 322: Also checked above

cpp/dolfinx/fem/Form.h
```diff
@@ -378,173 +474,77 @@ class Form
...
+  /// @param id Integral identifier index.
+  /// @param kernel_idx Kernel index (cell type).
+  /// @return An array `domain`, such that:
+  /// - For cell integrals, `domain[i]` is the index of the ith cell `i`
+  /// in the integration domain mesh.
+  /// - For exterior/interior facet integrals, `domain[2*i]` is the
+  /// index of a cell in the integration domain mesh attached to the
+  /// facets `i`, and `domain[2*i + 1]` is the local index of the facet
+  /// relative to the cell.
+  std::span<const std::int32_t> domain_arg(IntegralType type, int rank, int id,
```
jpdean, line 507: I don't think the docs for this function are correct. The returned list contains indices in the meshes that the trial or test functions are defined over, not the integration mesh.

cpp/dolfinx/fem/Form.h
```diff
@@ -378,173 +474,77 @@ class Form
...
   /// @param type Integral type.
-  /// @param i Integral ID, i.e. the (sub)domain index.
-  /// @param mesh The mesh the entities are numbered with respect to.
-  /// @return List of active entities in `mesh` for the given integral.
-  std::vector<std::int32_t> domain(IntegralType type, int i,
-                                   const mesh::Mesh<geometry_type>& mesh) const
+  /// @param id Integral identifier index.
+  /// @param c Coefficient index.
+  /// @return Domain indices for the coefficient function.
+  std::span<const std::int32_t> domain_coeff(IntegralType type, int id,
```
jpdean, line 534: Same issue here

cpp/dolfinx/fem/Form.h
```diff
@@ -606,9 +606,18 @@ class Form
   // True if permutation data needs to be passed into these integrals
   bool _needs_facet_permutations;
 
-  // Entity maps (see Form documentation)
-  std::map<std::shared_ptr<const mesh::Mesh<geometry_type>>,
-           std::vector<std::int32_t>>
-      _entity_maps;
+  // Domain data for argument functions
+  //  [rank_i][IntegralType, integral(id), cell_type] -> entity map
+  std::vector<std::map<
```
jpdean, line 611: I'm think it's fine with storing these here, rather than just the entitiy maps. But, this approach might use more memory than before for some integrals

cpp/dolfinx/fem/Form.h
```diff
@@ -606,9 +606,18 @@ class Form
   // True if permutation data needs to be passed into these integrals
   bool _needs_facet_permutations;
 
-  // Entity maps (see Form documentation)
-  std::map<std::shared_ptr<const mesh::Mesh<geometry_type>>,
-           std::vector<std::int32_t>>
-      _entity_maps;
+  // Domain data for argument functions
+  //  [rank_i][IntegralType, integral(id), cell_type] -> entity map
```
jpdean, line 610: I think "-> entity map" might be a bit misleading here. This gives the mapped entities for the requested [rank_i][IntegralType, integral(id), cell_type]

cpp/dolfinx/fem/Form.h
garth-wells: Any memory change would be zero to minimal because the data can be shared across integrals. The re-computation cost is avoided. Some more work would allow more data sharing for more complex cases.

Major change is avoiding need to store shared pointer to meshes, which was ugly.

cpp/dolfinx/fem/Form.h
jpdean: Yes, I agree. Though it's probably not a major issue for now

jpdean (approved):

cpp/dolfinx/fem/Form.h
jorgensd: Entity maps with it current design, stores -1 for entities of the parent mesh that does not exist in the submesh. That is why I have suggested we use a `MeshTags` instance instead, as it allows for a sparse 1-1 map.

PR #3648, opened 2025-02-24
Snes solver functional

jorgensd:


jhale:
Could you do the pull request against main? Thanks :)

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +91,453 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
...
+        self.copy_solution(self._u, self._x)
+
+        # Solve problem
+        self._snes.solve(None, self._x)
+
+        converged_reason = self._snes.getConvergedReason()
+        self.replace_solution(self._u, self._x)
+        return self._x, converged_reason, self._snes.getIterationNumber()
+
+    def __del__(self):
```
jhale, line 210: This class design is now very nice, but the issue to decide on is whether we want to smoothen out the solve and memory management in the first place. Another option would be to provide a high-level NonLinearProblem class, in the same style as LinearProblem.

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +91,453 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
...
+    J.zeroEntries()
+    assemble_matrix_block(J, jacobian, bcs=bcs, diagonal=1.0)  # type: ignore
+    J.assemble()
+    if preconditioner is not None:
+        P.zeroEntries()
+        assemble_matrix_block(P, preconditioner, bcs=bcs, diagonal=1.0)  # type: ignore
+        P.assemble()
+
+
+def replace_solution_block(u: list[dolfinx.fem.Function], x: PETSc.Vec):  # type: ignore
```
jhale, line 378: Personally I quite like having these in DOLFINx as I can never remember how to write them myself. Should they be somewhere else though? They aren't isolated to SNES or NLS.

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +91,453 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
...
+        Vi = ui.function_space
+        num_sub_dofs = Vi.dofmap.index_map.size_local * Vi.dofmap.index_map_bs
+        ui.x.petsc_vec.array_w[:num_sub_dofs] = x.array_r[
+            offset_start : offset_start + num_sub_dofs
+        ]
+        ui.x.petsc_vec.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)  # type: ignore
+        offset_start += num_sub_dofs
+
+
+def copy_solution_block(u: list[dolfinx.fem.Function], x: PETSc.Vec):  # type: ignore
```
jhale, line 397: We should review the naming in the more general context - u doesn't need to be a solution, necessarily.

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +91,453 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
 
         """
         super().setP(P, Pmat)
+
+
+def create_data_structures(
```
jhale, line 97: The naming is a bit generic - they are all either Mats or Vecs?

python/test/unit/fem/test_petsc_nonlinear_assembler.py
```diff
@@ -391,76 +271,48 @@ def initial_guess_p(x):
...
 
-            x = create_vector_block(F)
-            scatter_local_vectors(
-                x,
-                [u.x.petsc_vec.array_r, p.x.petsc_vec.array_r],
-                [
-                    (u.function_space.dofmap.index_map, u.function_space.dofmap.index_map_bs),
-                    (p.function_space.dofmap.index_map, p.function_space.dofmap.index_map_bs),
-                ],
+            snes_options = {"snes_rtol": 1.0e-15, "snes_max_it": 10, "snes_monitor": None}
```
jhale, line 277: Thanks, I think this is very useful.

python/test/unit/fem/test_petsc_nonlinear_assembler.py
```diff
@@ -391,76 +271,48 @@ def initial_guess_p(x):
...
+            snes, x = dolfinx.nls.petsc.create_snes_solver(
```
jhale, line 278: In addition, could we see an example of use/test where the very lowest level components are used outside of `create_snes_solver`? I.e. the user sets up SNES manually. I'd like to see how much boiler plate that involves.

python/test/unit/fem/test_petsc_nonlinear_assembler.py
```diff
@@ -492,30 +344,18 @@ def monolithic_solve():
...
-            snes.solve(None, x)
-            assert snes.getKSP().getConvergedReason() > 0
-            assert snes.getConvergedReason() > 0
+            solver = dolfinx.nls.petsc.SNESSolver(
+                F, U, J=J, bcs=bcs, assembly_type=dolfinx.fem.AssemblyType.default
+            )
+            solver.set_options(snes_options)
+            x, converged_reason, _ = solver.solve()
+            assert converged_reason > 0
+            solver.replace_solution(U, x)
```
jhale, line 357: We should discuss where this should go, inside or outside solve, now that there is a lower level path as well.

python/test/unit/fem/test_petsc_nonlinear_assembler.py
```diff
@@ -602,86 +433,53 @@ def initial_guess_p(x):
...
-            x.set(0.0)
-            snes.solve(None, x)
-            assert snes.getConvergedReason() > 0
-            snes.destroy()
-            Jnorm = nest_matrix_norm(Jmat)
-            Fnorm = Fvec.norm()
+            solver = dolfinx.nls.petsc.SNESSolver(
+                F, [u, p], J=J, bcs=bcs, assembly_type=dolfinx.fem.AssemblyType.nest, P=P
+            )
+            nested_IS = solver.snes.getJacobian()[0].getNestISs()
```
jhale, line 468: This is an advanced example, so the user would probably go with the direct SNES interface.

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +91,453 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
...
+        elif assembly_type == fem.AssemblyType.block:
+            self._copy_solution = copy_solution_block
+            self._replace_solution = replace_solution_block
+        elif assembly_type == fem.AssemblyType.nest:
+            self._copy_solution = copy_solution_nest
+            self._replace_solution = replace_solution_nest
+        else:
+            raise ValueError("Unsupported Assembly type")
+
+    def set_options(self, options: dict):
```
jhale, line 185: Remove this and pass up the options at construction time? This is consistent with e.g. LinearProblem.

python/test/unit/fem/test_petsc_nonlinear_assembler.py
jhale: I can do this.

python/dolfinx/nls/petsc.py
jhale: I agree, so in effect I'm saying there is an argument this should just be called `NonlinearSolver` which wraps low-level SNES routines, in the same way `LinearSolver` wraps low-level KSP routines.

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +92,500 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
...
+    elif assembly_type == fem.AssemblyType.nest:
+        matrix_creator = create_matrix_nest
+        vector_creator = create_vector_nest
+    else:
+        raise ValueError("Unsupported SNES type")
+    A = matrix_creator(a)  # type: ignore
+    b = vector_creator(L)  # type: ignore
+    x = vector_creator(L)  # type: ignore
+    P = None if P is None else matrix_creator(P)  # type: ignore
+    return A, x, b, P
```
jhale, line 131: `A, b, P, x` matches the input argument ordering.

jorgensd:
> The only downside I see with the free function approach is that now `dolfinx.nls.petsc` is polluted with functions like `F_default`, `F_block`, `J_default`, `J_block` which I guess we don't want the user to see (and think that they must be using those functions somehow). Maybe prefix them with an underscore?

They should be public as a "low-level" user (like Jack) would like to access these and not wrap the snes solver.

python/dolfinx/nls/petsc.py
jorgensd: I think we could remove the interface to `NewtonSolver` from Python, as I don't really see the benefit of it any longer. It is just a very naive implementation of `Newton` that somewhat interfaces with PETSc.

jorgensd:
> The only downside I see with the free function approach is that now `dolfinx.nls.petsc` is polluted with functions like `F_default`, `F_block`, `J_default`, `J_block` which I guess we don't want the user to see (and think that they must be using those functions somehow). Maybe prefix them with an underscore?

I have now added an example in the tests that uses these very low level blocks to set up a snes solver.

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +98,418 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
...
+    Args:
+        F: List of PDE residuals :math:`F_i`.
+        u: List of unknowns of the block system.
+        bcs: List of Dirichlet boundary conditions.
+        J: Rectangular array of bi-linear forms representing the
+            Jacobian :math:`J_ij = dF_i/du_j`.
+        P: Rectangular array of bi-linear forms representing the preconditioner.
+        assembly_type: Determines what kind of matrix the problem is assembled into.
+            If `N=0`, i.e one is solving for a single form, the assembly type is `default`.
+            For blocked, the assembly type is `block` or `nest`.
```
jhale, line 473: I wonder if we could just make default mean non-blocked or blocked, depending on if F is a Form or a list of Forms.

python/dolfinx/fem/forms.py
```diff
@@ -21,13 +21,20 @@
...
 
 if typing.TYPE_CHECKING:
     from mpi4py import MPI
 
     from dolfinx.fem import function
     from dolfinx.mesh import Mesh, MeshTags
+from enum import Enum
+
+
+class AssemblyType(Enum):
```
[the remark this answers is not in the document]
garth-wells: The first point still stands - PETSc concepts leaking outside of our PETSc containment.

python/dolfinx/nls/petsc.py
jorgensd: You would still need an argument to determine nest or not, so Im not sure if it gives much.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1118,91 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+    ]
+    _cpp.la.petsc.scatter_local_vectors(
+        x,
+        u_petsc,
+        index_maps,
+    )
+    x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)  # type: ignore
+
+
+def copy_nest_vec_to_functions(u: list[dolfinx.fem.Function], x: PETSc.Vec):  # type: ignore
```
jhale, line 1183: At least in C, it's common to do in then out ordering of arguments.

python/dolfinx/fem/forms.py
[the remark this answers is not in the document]
jorgensd: That has been resolved by moving it into `dolfinx.fem.petsc`.

python/dolfinx/nls/petsc.py
jorgensd: It is a side-track from the PR at hand. Whatever we want to do with NewtonSolver (make it purely native for instance) would require access to mumps or superlu dist. Which we currently do not have.

The current NewtonSolver is tied to PETSc, and is used in C++ by people not wanting to set up their own PETSc SNES solver there, so lets keep it for a rainy day.

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +98,424 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
...
+    x = vector_creator(L)  # type: ignore
+    P = None if P is None else matrix_creator(P)  # type: ignore
+    return A, b, P, x
+
+
+class SNESSolver:
+    def __init__(
+        self,
+        F: typing.Union[dolfinx.fem.Form, ufl.form.Form],
+        u: dolfinx.fem.Function,
```
jorgensd, line 150:
```suggestion
        u: typing.Union[dolfinx.fem.Function, list[dolfinx.fem.Function]],
```

michalhabera (commented):
I like the layered approach and I like the high-level SNESSolver class. But I am missing the point in this specific design of the lower- or mid- level API. What is the typical use case there? There should be an example. In the current code you help users with creating matrices `create_snes_matrices_and_vectors` and you provide some `dolfinx.nls.petsc.F_nest` that user controls what data is passed into it. But when would this very API be needed?

I think I struggle with `create_snes_matrices_and_vectors` in particular. Why not just a dictionary that maps `AssemblyType` to a vector and matrix factories and let user call these? Maybe as a user I want to only create the matrix, because I already have a vector in the correct layout. And maybe, why the need for a helper at all? The more layers in-between you expose to public API, the more ways to do things appear, and "There should be one-- and preferably only one --obvious way to do it."

jorgensd:
> But I am missing the point in this specific design of the lower- or mid- level API. What is the typical use case there? There should be an example

Please see the current tests. They show all different usages of the API.

jorgensd:
> Why not just a dictionary that maps `AssemblyType` to a vector and matrix factories and let user call these? Maybe as a user I want to only create the matrix, because I already have a vector in the correct layout. And maybe, why the need for a helper at all

If you want to use the lowest level API the users can call the appropriate `create_vector_*`, `create_matrix_*` and shouldn’t really need the create_snes_data_structures.

jorgensd:
> There should be one-- and preferably only one --obvious way to do it."

The others wanted a low level API. We can of course remove the mid-layer, ie not expose `create_snes_solver`, and only expose the wrappers for F,J and moving data, and then let anyone that wants to do something else than `SNESSolver` do everything manually.

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +98,432 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
...
+    @property
+    def copy_function_to_vec(self):
+        return self._copy_function_to_vec
+
+    @property
+    def copy_vec_to_function(self):
+        return self._copy_vec_to_function
+
+
+def F_default(
```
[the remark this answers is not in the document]
jorgensd: Will change to "standard" to reflect the enum.

jorgensd:
> like the layered approach and I like the high-level SNESSolver class. But I am missing the point in this specific design of the lower- or mid- level API. What is the typical use case there? There should be an example. In the current code you help users with creating matrices `create_snes_matrices_and_vectors` and you provide some `dolfinx.nls.petsc.F_nest` that user controls what data is passed into it. But when would this very API be needed?

The low level API can for instance be used to set your own `F_nest`,`J_nest` that takes into account the external operator. However, you might not want to manually control all the other things by hand, such as matrices/vec creation.

jhale:
> I think I struggle with create_snes_matrices_and_vectors in particular. Why not just a dictionary that maps AssemblyType to a vector and matrix factories and let user call these? Maybe as a user I want to only create the matrix, because I already have a vector in the correct layout. And maybe, why the need for a helper at all? The more layers in-between you expose to public API, the more ways to do things appear, and "There should be one-- and preferably only one --obvious way to do it."

Seems like a good suggestion to break this out further.

> I like the layered approach and I like the high-level SNESSolver class. But I am missing the point in this specific design of the lower- or mid- level API. What is the typical use case there? There should be an example. In the current code you help users with creating matrices create_snes_matrices_and_vectors and you provide some dolfinx.nls.petsc.F_nest that user controls what data is passed into it. But when would this very API be needed?

My view is that SNESSolver will end up being re-born as NonlinearProblem (in the same way LinearProblem wraps KSP), but that's another discussion to be had.

So then in terms of using SNES, there will be the `create_snes_solver` which will be the high-level API which returns a fully setup petsc4py SNES, and the low-level components that can be put together as the user wishes by interacting with petsc4py SNES directly. The idea here is that we reduce boilerplate - for example, I have applications where we simply need to redefine the function `J`.

python/test/unit/fem/test_petsc_nonlinear_assembler.py
```diff
@@ -718,33 +537,30 @@ def monolithic():
...
+                U,
+                J=J,
+                bcs=bcs,
+                P=P,
+                assembly_type=dolfinx.fem.petsc.AssemblyType.standard,
+                snes_options=snes_options,
+            )
+            x, converged_reason, _ = solver.solve()
+            assert converged_reason > 0
+            solver.copy_vec_to_function(x, U)
```
jhale, line 560: If we can agree that `SNESSolver` should be `NonlinearProblem` then this call should go inside solve and also solve should do any necessary ghost updates, in the same way that `LinearProblem` works.

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +98,432 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
 
         """
         super().setP(P, Pmat)
+
+
+def create_snes_matrices_and_vectors(
```
jhale, line 103: @michalhabera Questions the need for this in the public API - should users just use `create_*` directly?

python/test/unit/fem/test_petsc_nonlinear_assembler.py
jhale: Let's do that in this PR so that we can review this important set of changes as a whole.

python/dolfinx/nls/petsc.py
jhale: This is now resolved in favour of offering a NonlinearProblem class aimed at beginner users, in the same style as LinearProblem.

michalhabera:
> > I think I struggle with create_snes_matrices_and_vectors in particular. Why not just a dictionary that maps AssemblyType to a vector and matrix factories and let user call these? Maybe as a user I want to only create the matrix, because I already have a vector in the correct layout. And maybe, why the need for a helper at all? The more layers in-between you expose to public API, the more ways to do things appear, and "There should be one-- and preferably only one --obvious way to do it."
> 
> Seems like a good suggestion to break this out further.
> 
> > I like the layered approach and I like the high-level SNESSolver class. But I am missing the point in this specific design of the lower- or mid- level API. What is the typical use case there? There should be an example. In the current code you help users with creating matrices create_snes_matrices_and_vectors and you provide some dolfinx.nls.petsc.F_nest that user controls what data is passed into it. But when would this very API be needed?
> 
> My view is that SNESSolver will end up being re-born as NonlinearProblem (in the same way LinearProblem wraps KSP), but that's another discussion to be had.
> 
> So then in terms of using petsc4py SNES, there will be the `create_snes_solver` which will be the high-level API which returns a fully setup petsc4py SNES, and the low-level components that can be put together as the user wishes by interacting with petsc4py SNES directly. The idea here is that we reduce boilerplate - for example, I have applications where we simply need to redefine the function `J`.
> 
> Edit: @michalhabera There is an "example" of using SNES directly at https://github.com/FEniCS/dolfinx/pull/3648/files#diff-f9b6597717f57bdb47379a64d265490d51aa09a0320d4d47511b09fba2371b1bR304



> > But I am missing the point in this specific design of the lower- or mid- level API. What is the typical use case there? There should be an example
> 
> Please see the current tests. They show all different usages of the API.

I've seen the tests and "how" to use the API. But my question was more about "why" to use it, in the tests all could be neatly done with the high level SNESSolver.

jorgensd:
> > > I think I struggle with create_snes_matrices_and_vectors in particular. Why not just a dictionary that maps AssemblyType to a vector and matrix factories and let user call these? Maybe as a user I want to only create the matrix, because I already have a vector in the correct layout. And maybe, why the need for a helper at all? The more layers in-between you expose to public API, the more ways to do things appear, and "There should be one-- and preferably only one --obvious way to do it."
> > 
> > 
> > Seems like a good suggestion to break this out further.
> > > I like the layered approach and I like the high-level SNESSolver class. But I am missing the point in this specific design of the lower- or mid- level API. What is the typical use case there? There should be an example. In the current code you help users with creating matrices create_snes_matrices_and_vectors and you provide some dolfinx.nls.petsc.F_nest that user controls what data is passed into it. But when would this very API be needed?
> > 
> > 
> > My view is that SNESSolver will end up being re-born as NonlinearProblem (in the same way LinearProblem wraps KSP), but that's another discussion to be had.
> > So then in terms of using petsc4py SNES, there will be the `create_snes_solver` which will be the high-level API which returns a fully setup petsc4py SNES, and the low-level components that can be put together as the user wishes by interacting with petsc4py SNES directly. The idea here is that we reduce boilerplate - for example, I have applications where we simply need to redefine the function `J`.
> > Edit: @michalhabera There is an "example" of using SNES directly at https://github.com/FEniCS/dolfinx/pull/3648/files#diff-f9b6597717f57bdb47379a64d265490d51aa09a0320d4d47511b09fba2371b1bR304
> 
> > > But I am missing the point in this specific design of the lower- or mid- level API. What is the typical use case there? There should be an example
> > 
> > 
> > Please see the current tests. They show all different usages of the API.
> 
> I've seen the tests and "how" to use the API. But my question was more about "why" to use it, in the tests all could be neatly done with the high level SNESSolver.

This was a request of Jack. As he stated, he wants to work directly with the `SNES`-object, and handle memory desctruction, but maybe only replace the `F` or `J`.

jhale:
> I've seen the tests and "how" to use the API. But my question was more about "why" to use it, in the tests all could be neatly done with the high level SNESSolver.

Two points here:

1. The basic principle in DOLFINx is that we return first-class petsc4py objects, except at a couple of high-level entry points (LinearProblem and the new proposed NonlinearProblem). I think this has been very successful. Personally, as I've discussed with Jorgen, I'd be happy to not use LinearProblem and NonlinearProblem, but I appreciate that this is not very welcoming.
2. Your point is correct, these tests could be implemented using SNESSolver/NonlinearProblem. They are not really meant to be examples of use, but experiments + tests of how the lower-level API looks + works. I'm inclined to agree with you that `create_snes_matrices_and_vectors` could be removed from the public API.

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +98,432 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
...
+        snes.setJacobian(partial(J_standard, u, jacobian, preconditioner, bcs), A, P)  # type: ignore
+    elif assembly_type == fem.petsc.AssemblyType.block:
+        snes.setFunction(partial(F_block, u, residual, jacobian, bcs), b)
+        snes.setJacobian(partial(J_block, u, jacobian, preconditioner, bcs), A, P)
+    elif assembly_type == fem.petsc.AssemblyType.nest:
+        snes.setFunction(partial(F_nest, u, residual, jacobian, bcs), b)
+        snes.setJacobian(partial(J_nest, u, jacobian, preconditioner, bcs), A, P)
+    else:
+        raise ValueError(f"Unsupported SNES type {assembly_type}")
+    return snes, x
```
jhale, line 529: The thinking behind returning `x` is that users must not under any circumstances use the data behind `u` to call SNES `solve`. So in a sense, this is signposting that may not be necessary for users of the lower-level interface.

python/dolfinx/fem/forms.py
```diff
@@ -631,3 +631,27 @@ def create_form(
...
     )
     return Form(f, form.ufcx_form, form.code)
+
+
+def compute_jacobian(
+    F: typing.Union[ufl.Form | list[ufl.Form]],
+    u: typing.Union[Function | list[Function]],
+    du: typing.Optional[typing.Union[ufl.Argument, list[ufl.Argument]]] = None,
+) -> typing.Union[ufl.Form, list[list[ufl.Form]]]:
+    """Compute the Jacobian :math:`J = \\frac{\\partial F}{\\partial u}[\\delta u]`.
```
garth-wells, line 641: It's not clear from the docs what is 'computed'. It should say something like it computes the derivative of a UFL form to generate a UFL form for the Jacobian.

Need to document the arguments.

python/dolfinx/fem/petsc.py
```diff
@@ -939,12 +947,12 @@ def solve(self) -> _Function:
 
     @property
     def L(self) -> Form:
-        """The compiled linear form"""
+        """The compiled linear form `F`."""
```
garth-wells, line 950: It's a bit odd for a property named `L` to return `F`.

python/dolfinx/fem/petsc.py
```diff
@@ -939,12 +947,12 @@ def solve(self) -> _Function:
...
         return self._L
 
     @property
     def a(self) -> Form:
-        """The compiled bilinear form"""
+        """The compiled bilinear form of `a`."""
```
garth-wells, line 955: Odd docstring - more-or-less says "property `a` returns `a`". The docstring needs to make clear what the compiled bilinear form of `a` is.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1126,94 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
         Interpolation matrix.
     """
     return _interpolation_matrix(space0._cpp_object, space1._cpp_object)
+
+
+def copy_vec_to_function(
+    x: PETSc.Vec,
+    u: dolfinx.fem.Function,
```
garth-wells, line 1133: No need to pass a `dolfinx.fem.Function` here, only  PETSc vec from it is needed.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1126,94 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+):  # type: ignore
+    """Update the solution for the unknown `u` with the values in `x`.
+
+    Args:
+        x: Data that is inserted into the solution vector.
```
garth-wells, line 1138:
```suggestion
        x: Data that is copied to the vector ``u``.
```

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1126,94 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+        u: Function data should be inserted into.
```
garth-wells, line 1139:
```suggestion
        u: Function data should be coped to.
```
`u` should be a vector (see comment above)

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1126,94 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+    """
+    x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)  # type: ignore
+    x.copy(u.x.petsc_vec)
+    u.x.petsc_vec.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)  # type: ignore
```
garth-wells, line 1143: This ghost update can be avoided - get the local form and then copy.

python/dolfinx/nls/petsc.py
```diff
@@ -23,9 +29,29 @@
...
 from dolfinx import fem
-from dolfinx.fem.petsc import create_matrix, create_vector
+from dolfinx.fem.forms import form as _create_form
+from dolfinx.fem.petsc import (
+    apply_lifting,
+    assemble_matrix_block,
+    assemble_matrix_nest,
+    assemble_vector,
+    assemble_vector_block,
+    copy_block_vec_to_functions,
```
garth-wells, line 39: These functions names don't read well and there are a lot. In most cases (maybe all?) a single function would suffice for one direction of copying - the necessary operation can be inferred from the `Vec` type and whether a `Function` or a list of `Function`s  is passed. 


A simple way to reduce the number would be to pass an argument.

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +98,434 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
 
         """
         super().setP(P, Pmat)
+
+
+def _create_snes_matrices_and_vectors(
+    a: typing.Union[list[list[fem.Form]], fem.Form],
+    L: typing.Union[list[fem.Form], fem.Form],
+    P: typing.Union[list[list[fem.Form]], list[fem.Form], fem.Form, None],
+    assembly_type: fem.petsc.AssemblyType,
```
garth-wells, line 107: I think `assembly_type` should be a PETSc Mat/Vec type. This would allow the caller to not only specify 'nest`, but also other PETSc matrix types.

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +98,434 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
...
+    @property
+    def copy_function_to_vec(self):
+        return self._copy_function_to_vec
+
+    @property
+    def copy_vec_to_function(self):
+        return self._copy_vec_to_function
+
+
+def F_standard(
```
garth-wells, line 249: The `F_foo` function names looks weird.

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +98,434 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
...
+    u: dolfinx.fem.Function,
+    residual: dolfinx.fem.Form,
+    jacobian: dolfinx.fem.Form,
+    bcs: list[dolfinx.fem.DirichletBC],
+    snes: PETSc.SNES,  # type: ignore
```
garth-wells, line 254: Why is `snes` passed? It's not used.

garth-wells (changes requested):
A first pass of comments.

I would prefer to see this built up in stages, perhaps by starting with bottom-up functionality that simplifies the test examples. An example to start could be a single function copies data from DOLFINx `Function`s in/out of blocked PETSc data structures, and get that as simple as possible. This could be merged aheads of a SNES interface.

python/dolfinx/nls/petsc.py
jhale: Already discussed this, it was also my preference, but @jorgensd told me it's not possible to discern between nest and block types. I do see `MATNEST` and `MATBLOCK` though: https://petsc.org/release/manualpages/Mat/MatType/ - would these work @jorgensd ?

python/dolfinx/fem/forms.py
jhale: I'm going to review and edit all of the documentation in due course. Many issues, also with consistency with respect to `LinearProblem`.

python/dolfinx/nls/petsc.py
jorgensd: We don't use `createBlockMat` (https://petsc.org/release/manualpages/Mat/MatCreateBlockMat/).
We use `createMat` and attach an index set to it based on the concatenated index maps.

python/dolfinx/fem/petsc.py
jorgensd, line 955:
```suggestion
        """The compiled bilinear form."""
```

python/dolfinx/fem/petsc.py
```diff
@@ -1015,12 +1023,12 @@ def __init__(
 
     @property
     def L(self) -> Form:
-        """Compiled linear form (the residual form)"""
+        """The compiled linear form `F`."""
```
jorgensd, line 1026:
```suggestion
        """The compiled linear form (the residual form)."""
```

python/dolfinx/nls/petsc.py
jorgensd: Again, how do we differentiate between block and nest copy? There is no clear way of distinguishing them on the PETSc level. If you have one way of doing it, please show it. 

This also then goes back to you wanting to remove the "Function input in `copy_function_to_vec`, which if you make it a PETSc vector, there is no way of distinguishing the copy direction.

python/dolfinx/nls/petsc.py
jorgensd: What should we call them then?
Just calling them weird isn't very helpful.

python/dolfinx/nls/petsc.py
jorgensd: Updated to use `_snes` as the solver has to be passed in to adhere with the SNES Residual and Jacobian function.

python/dolfinx/nls/petsc.py
garth-wells: The PETSc Mat/Vec has a `getType()` method, which can be used to check if it is blocked. 

The direction can be inferred by the argument order.

jorgensd: Nope, see Jack's PR: https://github.com/FEniCS/dolfinx/pull/3652
And you couldn't infer the direction if one would go for your suggestion above:
https://github.com/FEniCS/dolfinx/pull/3648#discussion_r1973473558
> No need to pass a dolfinx.fem.Function here, only PETSc vec from it is needed.
as the operation would be PETSc.Vec <-> PETSc.Vec. 
However, lets just review Jack's PR instead of discussing here.

python/dolfinx/nls/petsc.py
garth-wells: We don't need to discern between block and nest types. If the forms are a list, then it's block or nest. Nest is used if the user asks for nest, otherwise `MATAIJ` or whatever else they want.

The interface should probably have options for the vector and the matrix (matrices for nest), since they can differ. Even for a single matrix or a matrix in a nest there are options, e.g. block CSR. See https://github.com/FEniCS/dolfinx/blob/192222e278fb2809abe97c3d46ecda534fb6bcdf/python/dolfinx/fem/petsc.py#L174 where the users passes the `MatType`. The other "create_matrix functions should have that option too.

garth-wells: Focusing on the matrix, the user should do one of the following:
1. Send in `None` for the matrix type. In this case the default PETSc matrix type will be used. If the forms are blocked, the assembly will be blocked; or
2. Send in `MATNEST` as the matrix type. In this case each block will have the default matrix type; or
3. Send in a list-of-lists with the matrix type for each block in a `MATNEST` matrix. 

The C++ interfaces for creating the matrices support these options. The code should handle a PETSc prefix to support setting of matrix types from the command line.

python/dolfinx/nls/petsc.py
```diff
@@ -72,3 +90,405 @@ def setP(self, P: types.FunctionType, Pmat: PETSc.Mat):  # type: ignore
...
+
+    @property
+    def snes(self) -> PETSc.SNES:  # type: ignore
+        return self._snes
+
+    @property
+    def copy_function_to_vec(self):
+        return self._copy_function_to_vec
+
+    @property
```
jorgensd, line 206: @jorgensd  please remove these

jhale:
I propose I code some further adjustments:

- API change: `SNESSolver` immediately becomes `NonlinearProblem`. Current `NonlinearProblem` becomes `NewtonSolverNonlinearProblem` and is marked as deprecated.
- API change: Python `NewtonSolver` is marked as deprecated.
- `compute_jacobian` has name changed and documentation string added.
- Consider further factoring into private functions in `assemble_residual`.

python/dolfinx/fem/petsc.py
```diff
@@ -922,6 +1036,14 @@ def __init__(
 
             problem = NonlinearProblem(F, u, [bc0, bc1])
         """
+        warnings.warn(
+            (
+                "dolfinx.nls.petsc.NewtonSolver is deprecated. "
+                + "Use dolfinx.fem.NonlinearProblem, a high level interface to PETSc SNES, instead."
```
```suggestion
                + "Use dolfinx.fem.petsc.NonlinearProblem, a high level interface to PETSc SNES, instead."
```

python/dolfinx/nls/petsc.py
```diff
@@ -43,14 +44,22 @@
...
 
 class NewtonSolver(_cpp.nls.petsc.NewtonSolver):
     def __init__(self, comm: MPI.Intracomm, problem: NonlinearProblem):
         """A Newton solver for non-linear problems."""
         super().__init__(comm)
 
+        warnings.warn(
+            (
+                "dolfinx.nls.petsc.NewtonSolver is deprecated. "
+                + "Use dolfinx.fem.NonlinearProblem, a high level interface to PETSc SNES, instead."
```
jorgensd:
```suggestion
                + "Use dolfinx.fem.petsc.NonlinearProblem, a high level interface to PETSc SNES, instead."
```

jhale:
Various modifications to present a consistent interface as `LinearProblem`. Quite happy with how the tests look - I will merge this, and then modify the Cahn Hilliard demo next week.

PR #3649, opened 2025-02-26
Simplify assembler implementation interfaces

garth-wells:
Use `mdspan` for multi-dimensional arrays in the (private) assembler implementations. `mdspan` has crept into the interface in other places, so we might as well use it where it makes things simpler. This change:
- Reduces the number of arguments to assembly interfaces, which improves readability. 
- Exposed an error in the Python interface for wrapping 2D arrays
- Removes code lint where data was passed that was not used.
- Has documentation improvements (still more to do).
- In simplifying the code exposed sub-performant code, which has been fixed (#3645).

### DIFF
--- a/cpp/dolfinx/fem/assemble_matrix_impl.h
+++ b/cpp/dolfinx/fem/assemble_matrix_impl.h
@@ -27,9 +27,7 @@ namespace dolfinx::fem::impl
 namespace md = MDSPAN_IMPL_STANDARD_NAMESPACE;
 
 /// @brief Typedef
-using mdspan2_t = MDSPAN_IMPL_STANDARD_NAMESPACE::mdspan<
-    const std::int32_t,
-    MDSPAN_IMPL_STANDARD_NAMESPACE::dextents<std::size_t, 2>>;
+using mdspan2_t = md::mdspan<const std::int32_t, md::dextents<std::size_t, 2>>;
 
 /// @brief Execute kernel over cells and accumulate result in matrix.
 /// @tparam T Matrix/form scalar type.
@@ -51,9 +49,7 @@ using mdspan2_t = MDSPAN_IMPL_STANDARD_NAMESPACE::mdspan<
 /// @param bc0 Marker for rows with Dirichlet boundary conditions applied
 /// @param bc1 Marker for columns with Dirichlet boundary conditions applied
 /// @param kernel Kernel function to execute over each cell.
-/// @param coeffs The coefficient data array of shape (cells.size(), cstride),
-/// flattened into row-major format.
-/// @param cstride The coefficient stride
+/// @param coeffs The coefficient data array of shape (cells.size(), cstride).
 /// @param constants The constant data
 /// @param cell_info0 The cell permutation information for the test function
 /// mesh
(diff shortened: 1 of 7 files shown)


jorgensd:
Summarizing:
All 2D structures input to assembly is now an mdspan. This includes:
- integration entities (for geometry, test and trial functions for interior and exterior face integrals)
- Cell_permutations for interior and exterior facets
- Coefficients 
- cell permutations for exterior and interior facets

In turn, this means that we do not need to send in the cstride or num facets per cell to the assembler.

Could you be more specific about what bugs you stumbled over, and what the redundant data was?

Finally, should we profile this before merging? This is one of the performance critical places in dolfinx, and is a real test of the performance of md span.

As a side note, should we enforce that the md spans are contiguous in memory?

garth-wells:
> Summarizing: All 2D structures input to assembly is now an mdspan. This includes:
> 
> * integration entities (for geometry, test and trial functions for interior and exterior face integrals)
> * Cell_permutations for interior and exterior facets
> * Coefficients
> * cell permutations for exterior and interior facets
> 
> In turn, this means that we do not need to send in the cstride or num facets per cell to the assembler.
> 
> Could you be more specific about what bugs you stumbled over, and what the redundant data was?
> 

1. The line https://github.com/FEniCS/dolfinx/blob/192222e278fb2809abe97c3d46ecda534fb6bcdf/python/dolfinx/wrappers/dolfinx_wrappers/pycoeff.h#L30 should use `e.second.size()` in place of `e.second.shape(0)`.
2. The redundant data was just code lint in private implementations left over from previous refactorings.  

> Finally, should we profile this before merging? This is one of the performance critical places in dolfinx, and is a real test of the performance of md span.
> 

I have tested and there is no measurable performance difference. 

> As a side note, should we enforce that the md spans are contiguous in memory?

It is enforced (and is row-major) by the extents that have been used.

garth-wells:
> Looks good. I would also be interested to know about performance.
> 

No measurable difference.

> On a tangential note, what's the general status of C++23 and vendored std::mdspan?

It's available in all major implementations except GCC/libstdc++, see https://en.cppreference.com/w/cpp/23.

jorgensd:
> 1. The line https://github.com/FEniCS/dolfinx/blob/192222e278fb2809abe97c3d46ecda534fb6bcdf/python/dolfinx/wrappers/dolfinx_wrappers/pycoeff.h#L30
>     should use `e.second.size()` in place of `e.second.shape(0)`

Should this be picked up in a test?

garth-wells:
> > 1. The line https://github.com/FEniCS/dolfinx/blob/192222e278fb2809abe97c3d46ecda534fb6bcdf/python/dolfinx/wrappers/dolfinx_wrappers/pycoeff.h#L30
> >    
> >    should use `e.second.size()` in place of `e.second.shape(0)`
> 
> Should this be picked up in a test or is it naturally picked up now that it is an md span and not flattened?

It is now picked up by an assertion.

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -24,39 +24,44 @@
...
+/// `(cells.size(), num_cell_coeffs)`. `coeffs(i, j)` is the `j`th
+/// coefficient for cell `i`.
+/// @param constants Constant data.
+/// @param cell_info0 Cell permutation information for the test
+/// function mesh.
+/// @param cell_info1 Cell permutation information for the trial
+/// function mesh.
 template <dolfinx::scalar T>
 void assemble_cells(
     la::MatSet<T> auto mat_set, mdspan2_t x_dofmap,
```
jorgensd, line 67: Why is `x` still flattened? It is always width 3.

PR #3652, opened 2025-02-27
Add functions to copy data between Function(s) and PETSc Vecs

jhale:
This PR adds functions that can copy data between function(s) and standard/blocked/nested PETSc Vecs. Dispatch to correct function is performed through weak typing, always duck-typing through to standard routine.

This work was based on https://github.com/FEniCS/dolfinx/pull/3648 by @jorgensd


jhale:
Some outstanding points:

- Use localForm in `_copy_vec_to_function`
- Should we assume `x` and underlying `u` enters in a consistent state?
- Are all of the ghost updates necessary?

jorgensd:
> Some outstanding points:
> 
>     * Use localForm in `_copy_vec_to_function`
> 
>     * Should we assume `x` and vector underlying `u` enters in a consistent state?
> 
>     * Are all of the ghost updates necessary?

They do not do not always come in as a consistent state (observed in the SNESSolver).

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1118,106 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+    else:
+        # Fall through to standard copy
+        _copy_vec_to_function(x, u)
+
+
+def _copy_vec_to_function(x: PETSc.Vec, u: dolfinx.fem.Function):  # type: ignore
+    """Copy values of x into vectors backing function(s) u. Normal version."""
+    x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)  # type: ignore
+    x.copy(u.x.petsc_vec)
+    u.x.petsc_vec.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)  # type: ignore
```
jorgensd, line 1149:
```suggestion
    x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)  # type: ignore
    with x.localForm() as _x, u.x.petsc_vec.localForm() as _u:
        _u.x.array[:] = _x.array_r
```

garth-wells:
I've reduced the number of functions and made them much simpler, and demonstrated usage in the SNES tests.

The functions should not have ghost updates. Copying the ghost values eliminates any need for a 'post' update. Any 'pre' ghost update should be performed by the caller. Throwing in mis-guided ghost updates is something we have eliminated in DOLFINx.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1122,141 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
         Interpolation matrix.
     """
     return _interpolation_matrix(space0._cpp_object, space1._cpp_object)
+
+
+@functools.singledispatch
+def assign(
+    x0: typing.Union[npt.NDArray[np.floating], list[npt.NDArray[np.floating]]], x1: PETSc.Vec
```
jorgensd, line 1129: Shouldn't the typing also include `np.complexfloating` for complex values?

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1122,141 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+            assign(u.x.array, x)
+
+
+@assign_function.register(PETSc.Vec)
+def _(x: PETSc.Vec, u: typing.Union[_Function, list[_Function]]):
+    """Assign vector entries to :class:`.Function` degrees-of-freedom.
+
+    Assigns values in ``x`` to the degrees-of-freedom of ``u``, which is
+    possibly a collection of ``Functions``s. When ``u`` is a collecion
+    of ``Functions``s, values in ``x`` are assigned block-wise ro the
```
jorgensd, line 1241:
```suggestion
    of ``Functions``s, values in ``x`` are assigned block-wise to the
```

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1122,141 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+                data0.append(v.x.array[: bs * n])
+                data1.append(v.x.array[bs * n :])
+            assign(data0 + data1, x)
+        else:
+            assign(u.x.array, x)
+
+
+@assign_function.register(PETSc.Vec)
+def _(x: PETSc.Vec, u: typing.Union[_Function, list[_Function]]):
+    """Assign vector entries to :class:`.Function` degrees-of-freedom.
```
jorgensd, line 1237:
```suggestion
    """Assign vector entries to :class:`Function` degrees-of-freedom.
```
Should there really be a `.` here?

jorgensd (commented):
Only minor things to consider. 

Should we move it into `dolfinx.la.petsc` right away?

garth-wells:
> Should we move it into `dolfinx.la.petsc` right away?

We don't have a `dolfinx.ls.petsc.py` file at the moment, so rather than creating one it might be better to make the linear algebra functions private.

jorgensd:
> > Should we move it into `dolfinx.la.petsc` right away?
> 
> We don't have a `dolfinx.ls.petsc.py` file at the moment, so rather than creating one it might be better to make the linear algebra functions private.

The `assign` vec<->function are very useful, even for experienced programmers. This was pointed out by @jhale in the snes-problem PR. I don’t understand the reason for making them if you want them to be private.

python/dolfinx/fem/petsc.py
jhale: From my reading of the numpy documentation this concept is expressed as `np.inexact`. https://numpy.org/doc/stable/reference/arrays.scalars.html#inexact-types

I suspect we're making a number of issues overtyping/undertyping/incorrectly typing throughout the codebase. This could be looked at in another PR.

jhale: Clearly we should get this one agreed on so we can look elsewhere!

jhale:
> > > Should we move it into `dolfinx.la.petsc` right away?
> > 
> > 
> > We don't have a `dolfinx.ls.petsc.py` file at the moment, so rather than creating one it might be better to make the linear algebra functions private.
> 
> The `assign` vec<->function are very useful, even for experienced programmers. This was pointed out by @jhale in the snes-problem PR. I don’t understand the reason for making them if you want them to be private.

My interpretation of @garth-wells comment was that the lower-level routine should be made private, rather than putting it in la.petsc (which would be the correct placement). The routines that take `Function`s belong in `fem.petsc`.

garth-wells:
> My interpretation of @garth-wells comment was that the lower-level routine should be made private, rather than putting it in a new location `la.petsc` (which would be the strictly correct placement). The routines that take `Function`s belong in `fem.petsc`.

Correct.

jhale:
I've done the legwork to split la into a native and PETSc part in a manner consistent with the rest of our interface. This creates an API change `dolfinx.la.create_petsc_vector` -> `dolfinx.la.petsc.create_petsc_vector`. The second `petsc` now seems redundant. Should it be renamed e.g. `dolfinx.la.petsc.create_vector` and `dolfinx.la.petsc.create_vector_wrap`?

https://github.com/FEniCS/dolfinx/tree/jhale/refactor-la-petsc

jhale:
Closed in favour of #3659

PR #3654, opened 2025-03-01
CI use `.env` configuration in favor of run dependency

schnellerhase:


jhale:
Looks much better. Few comments:

.env isn't very discoverable. How about just fenicsx-refs.env?

Does it work with PowerShell on Windows?

schnellerhase:
Sounds good, renamed. And tested windows, see https://github.com/FEniCS/dolfinx/actions/runs/13606788609/job/38039199644?pr=3654.

jhale (approved):

.github/workflows/oneapi.yml
```diff
@@ -41,11 +36,14 @@ jobs:
     steps:
       - name: Install compiler dependencies
         run: |
-          apt-get -y update
-          apt-get -y install binutils libstdc++-14-dev
+          sudo apt-get -y update
```
garth-wells, line 39: Has something changed that now makes `sudo` required?

schnellerhase: Yes, that became necessary due to the removal of the container image above (which was just the same as the runner image itself)

jhale: I missed this on my review, thanks @garth-wells - the Ubuntu Docker images are much more minimal than the GitHub ones, hence we prefer the Docker images. Can you change back?

garth-wells: The Ubuntu images are 'more' minimal than the Github VM images which is good, and also reproducible locally which is very good.

schnellerhase: I see will change back. One other problem I faced was that in the container the `.github` directory does not seem to be present. Not sure why.

schnellerhase: Figured it out, if no `git` install is present the checkout action defaults to the `Github REST API`, this seems to miss those files for some reason. Installing `git` fixes it.

PR #3659, opened 2025-03-05
Refactor LA into native and PETSc part and add assigners

jhale:
Supersedes #3652 which made it apparent that `dolfinx.la` should have a dedicated `dolfinx.la.petsc` module.

This PR introduces API changes:

`dolfinx.la.create_petsc_vector` -> `dolfinx.la.petsc.create_vector`
`dolfinx.la.create_petsc_vector_wrap` -> `dolfinx.la.petsc.create_vector_wrap`

Note that `dolfinx.fem.create_vector` is now a very light wrapper over `dolfinx.la.create_vector`, perhaps bringing into question the need for the former.


jorgensd:
Couldn’t we place all PETSC related structures in `dolfinx.petsc` instead of having spread namespaces as `nls`,`fem`,`la` that has all to be explicitly imported in DOLFINx?

jhale:
> Couldn’t we place all PETSC related structures in `dolfinx.petsc` instead of having spread namespaces as `nls`,`fem`,`la` that has all to be explicitly imported in DOLFINx?

Well, yes, possibly, but not in this PR. But it's a huge API change. Let's discuss it on Slack.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1119,61 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
         Interpolation matrix.
     """
     return _interpolation_matrix(space0._cpp_object, space1._cpp_object)
+
+
+@functools.singledispatch
+def assign(u: typing.Union[_Function, Sequence[_Function]], x: PETSc.Vec):
```
jorgensd, line 1125: Is the reasoning for keeping this assign in `fem.petsc` that it works on a `Function` rather than a set of arrays?

garth-wells: Yes. 

The linear algebra module should not see/include any finite element concepts.

jorgensd (approved):

python/test/unit/fem/test_petsc_nonlinear_assembler.py
```diff
@@ -403,15 +385,10 @@ def blocked_solve():
...
-                x,
-                [u.x.petsc_vec.array_r, p.x.petsc_vec.array_r],
-                [
-                    (u.function_space.dofmap.index_map, u.function_space.dofmap.index_map_bs),
-                    (p.function_space.dofmap.index_map, p.function_space.dofmap.index_map_bs),
-                ],
-            )
-            x.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+
+            # Assign u, p values to x
```
michalhabera, line 389: I think this comment can go, the code below is self-explanatory.

python/test/unit/fem/test_petsc_nonlinear_assembler.py
```diff
@@ -445,13 +422,9 @@ def nested_solve():
...
             assert x.getType() == "nest"
-            for x_soln_pair in zip(x.getNestSubVecs(), (u, p)):
-                x_sub, soln_sub = x_soln_pair
-                soln_sub.x.petsc_vec.ghostUpdate(
-                    addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD
-                )
-                soln_sub.x.petsc_vec.copy(result=x_sub)
-                x_sub.ghostUpdate(addv=PETSc.InsertMode.INSERT, mode=PETSc.ScatterMode.FORWARD)
+
+            # Assign u, p values to x
```
michalhabera, line 426: Same for this comment.

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1119,61 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
...
+    'stacked' and assigned to ``x``. See :func:`assign` for documentation on
+    how stacked assignment is handled.
+
+    Args:
+        u: ``Function`` (s) to assign degree-of-freedom value from.
+        x: Vector to assign degree-of-freedom values in ``u`` to.
+    """
+    if x.getType() == PETSc.Vec.Type().NEST:
+        dolfinx.la.petsc.assign([v.x.array for v in u], x)
+    else:
```
michalhabera, line 1140: I think a single branch of
```python
if x.getType() == PETSc.Vec.Type().NEST:
  ...
elif isinstance(u, Sequence):
  ...
else:
  ...

```
 would be more readable here. Similar for the assign below.

python/dolfinx/la/__init__.py
```diff
@@ -19,9 +20,9 @@
     "MatrixCSR",
     "Norm",
     "Vector",
-    "create_petsc_vector",
     "is_orthonormal",
     "matrix_csr",
+    "norm",
```
michalhabera, line 25: Norm used anywhere?

python/dolfinx/fem/petsc.py
```diff
@@ -1118,3 +1119,61 @@ def interpolation_matrix(space0: _FunctionSpace, space1: _FunctionSpace) -> PETS
         Interpolation matrix.
     """
     return _interpolation_matrix(space0._cpp_object, space1._cpp_object)
+
+
+@functools.singledispatch
+def assign(u: typing.Union[_Function, Sequence[_Function]], x: PETSc.Vec):
+    """Assign :class:`Function` degrees-of-freedom to to a vector.
```
michalhabera, line 1126:
```suggestion
    """Assign :class:`Function` degrees-of-freedom to a vector.
```

python/dolfinx/la/petsc.py
```diff
@@ -0,0 +1,140 @@
...
+              [x0_owned[n-1]]
+              [x0_ghost[0]]
+              [.....]
+              [x0_ghost[n-1]]
+
+    Args:
+        x0: An array or list of arrays that will be assigned to ``x1``.
+        x1: Vector to assign values to.
+    """
+    try:
```
michalhabera, line 97: Is this try block meant to check if the vector is of type NEST? If yes, why not check that directly?

michalhabera (commented):
The separation into fem.petsc and la.petsc reads very nice! And also follows the C++ structure.

Why are all these helpers (la.petsc.create_vector_wrap, la.petsc.create_vector, la.petsc.assign, fem.petsc.assign, ...) written in Python only? Is the argument that C++ users would not need/want such helpers?

jhale:
> Why are all these helpers (la.petsc.create_vector_wrap, la.petsc.create_vector, la.petsc.assign, fem.petsc.assign, ...) written in Python only? Is the argument that C++ users would not need/want such helpers?

I discussed this with @garth-wells - providing similar functionality for C++ users would be beneficial as interfacing SNES in particular is apparently "very hard". However, it's not clear we would want to wrap PETSc in C++ and then expose to Python, or just provide two good designs that suits both languages for PETSc C and petsc4py.

garth-wells:
> > Why are all these helpers (la.petsc.create_vector_wrap, la.petsc.create_vector, la.petsc.assign, fem.petsc.assign, ...) written in Python only? Is the argument that C++ users would not need/want such helpers?
> 
> I discussed this with @garth-wells - providing similar functionality for C++ users would be beneficial as interfacing SNES in particular is apparently "very hard". However, it's not clear we would want to write helpers in C++ using PETSc and then expose to Python, or just provide two good designs that suits C++ and Python via PETSc C and petsc4py, respectively.

There is a helper at https://github.com/FEniCS/dolfinx/blob/43dafba21c4333adb3cf345a13a33f7ab315b52b/cpp/dolfinx/la/petsc.h#L107.

More C++ 'helpers' could be helpful, but we should do it with a purpose, e.g. when adding SNES helpers, to get it right rather than adding it for hypothetical reasons.

python/test/unit/fem/test_petsc_nonlinear_assembler.py
garth-wells: Or is it "assign x values to u, p" :)

python/dolfinx/la/__init__.py
jhale: Yes, extensively in the unit tests for example.

PR #3667, opened 2025-03-20
Unify PETSc matrix and vector factory functions

jorgensd:
Stemming from discussions in #3648, here is an attempt at unifying the PETSc matrix and vector constructors.

It avoids docs duplication, and very small wrapper function mapping Python <-> C++ one to one.

### DIFF
--- a/python/dolfinx/fem/petsc.py
+++ b/python/dolfinx/fem/petsc.py
@@ -138,7 +138,7 @@ def create_vector(
         ]
         if kind == PETSc.Vec.Type.NEST:
             return _cpp.fem.petsc.create_vector_nest(maps)
-        elif kind in (None, PETSc.Vec.Type.NEST, PETSc.Vec.Type.MPI):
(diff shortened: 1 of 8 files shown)


python/dolfinx/fem/petsc.py
```diff
@@ -136,62 +136,46 @@ def create_vector(
...
         ]
-
-        if vec_type is None:
-            # Create block vector
-            return _cpp.fem.petsc.create_vector_block(maps)
-        elif vec_type == PETSc.Vec.Type.NEST or vec_type == "nest":
-            # Create nest vector
+        if kind == PETSc.Vec.Type.NEST:
             return _cpp.fem.petsc.create_vector_nest(maps)
+        elif kind in (None, PETSc.Vec.Type.NEST, PETSc.Vec.Type.MPI):
```
jorgensd, line 141: Nest should be removed here?

PR #3668, opened 2025-03-23
Unify assembly functions for PETSc data structures (Python)

garth-wells:
This change unifies (i) 'standard' assembly, (ii) blocked assembled and (iii) PETSc 'nest' assembly. Previously, each case had its own 'assembly' functions.  Common `assemble_matrix` and `assemble_vector` functions now support all three cases.  `assemble_matrix(vector)_block` and `assemble_matrix(vector)_nest` have been removed.

This change allow higher level classes and functions, e.g. a PETSc SNES support class, to be simpler since common functions can be called that transparently support different cases. The unification also eliminates some inconsistencies in behaviour across the three cases.

PR also include substantial improvements to the documentation of the affected functions, and improvements to type hints.


cpp/dolfinx/fem/assembler.h
```diff
@@ -371,7 +446,7 @@ void assemble_matrix(
 /// @param[in] constants Constants that appear in `a`.
 /// @param[in] coefficients Coefficients that appear in `a`.
 /// @param[in] bcs Boundary conditions to apply. For boundary condition.
```
jorgensd, line 448:
```suggestion
/// @param[in] bcs Boundary conditions to apply. For boundary condition
```

python/dolfinx/fem/assemble.py
```diff
@@ -11,36 +11,39 @@
...
 from dolfinx.cpp.fem import pack_coefficients as _pack_coefficients
 from dolfinx.cpp.fem import pack_constants as _pack_constants
+from dolfinx.fem import IntegralType
 from dolfinx.fem.bcs import DirichletBC
 from dolfinx.fem.forms import Form
 
 
 def pack_constants(
     form: typing.Union[Form, typing.Sequence[Form]],
 ) -> typing.Union[np.ndarray, typing.Sequence[np.ndarray]]:
```
jorgensd, line 31: Should we make the type-hint more specific, and use `typing.Union[npt.NDArray[np.inexact], typing.Sequence[npt.NDArray[np.inexact]]` to highlight that the constants shouldn't be integers?

python/dolfinx/fem/assemble.py
```diff
@@ -117,14 +131,14 @@ def create_matrix(a: Form, block_mode: typing.Optional[la.BlockMode] = None) ->
 # -- Scalar assembly ---------------------------------------------------------
 
 
-def assemble_scalar(M: Form, constants=None, coeffs=None):
+def assemble_scalar(M: Form, constants: typing.Optional[np.ndarray] = None, coeffs=None):
```
jorgensd, line 134: Same typehint-comment about the np-array.
Should we also type-hint the coefficients, as done in pack_coefficients?

python/dolfinx/fem/petsc.py
```diff
@@ -634,86 +559,147 @@ def _assemble_matrix_block_mat(
...
+    Args:
+        b: Vector to modify in-place.
+        a: List of bilinear forms. If ``b`` is not blocked or a nest,
+            then ``a`` is a 1D sequence. If ``b`` is blocked or a nest,
+            then ``a`` is  a 2D array of forms, with the ``a[i]`` forms
+            used to modify the block/nest vector ``b[i]``.
+        bcs: Boundary conditions to use to modify ``b`` (see
+            :func:`dolfinx.fem.apply_lifting`). The boundary conditions
+            in ``bcs[j]`` are associated with the forms in the ``j``th
+            colulmn of ``a``.
```
jorgensd, line 586:
```suggestion
            column of ``a``.
```

python/dolfinx/fem/petsc.py
```diff
@@ -634,86 +559,147 @@ def _assemble_matrix_block_mat(
...
+        x0 = [] if x0 is None else x0.getNestSubVecs()
+        constants = [pack_constants(forms) for forms in a] if constants is None else constants
+        coeffs = [pack_coefficients(forms) for forms in a] if coeffs is None else coeffs
+        for b_sub, a_sub, const, coeff in zip(b.getNestSubVecs(), a, constants, coeffs):
+            const_ = list(
+                map(lambda x: np.array([], dtype=PETSc.ScalarType) if x is None else x, const)
+            )
+            apply_lifting(b_sub, a_sub, bcs, x0, alpha, const_, coeff)
+    else:
+        with contextlib.ExitStack() as stack:
```
jorgensd, line 617: This is where I think it would be natural to extract he appropriate bcs from the list of BCs.

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -581,22 +583,13 @@ def test_mixed_measures():
     dx_smsh = ufl.Measure("dx", smsh)
 
     # Trial and test functions
-    u, v = ufl.TrialFunction(V), ufl.TestFunction(V)
-    p, q = ufl.TrialFunction(Q), ufl.TestFunction(Q)
+    _, v = ufl.TrialFunction(V), ufl.TestFunction(V)
+    _, q = ufl.TrialFunction(Q), ufl.TestFunction(Q)
```
jorgensd, line 587:
```suggestion
    v = ufl.TestFunction(V)
    q = ufl.TestFunction(Q)
```

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -581,22 +583,13 @@ def test_mixed_measures():
...
-            fem.form(ufl.inner(p, v) * dx_smsh, entity_maps={msh: smsh_to_msh}),
-        ],
-        [
-            fem.form(ufl.inner(u, q) * dx_smsh, entity_maps={msh: smsh_to_msh}),
-            fem.form(ufl.inner(p, q) * dx_smsh),
-        ],
-    ]
     L = [fem.form(ufl.inner(2.3, v) * dx_msh), fem.form(ufl.inner(1.3, q) * dx_smsh)]
-    b0 = assemble_vector_block(L, a)
+    b0 = assemble_vector(L, kind=PETSc.Vec.Type.MPI)
```
jorgensd, line 591: This simplification of the test is very nice!

python/test/unit/fem/test_assembler.py
```diff
@@ -357,34 +356,41 @@ def test_matrix_assembly_block(self, mode):
 
         def blocked():
             """Monolithic blocked"""
-            A = petsc_assemble_matrix_block(a_block, bcs=[bc])
+            A = petsc_assemble_matrix(a_block, bcs=[bc])
```
jorgensd, line 359: Couldn't this now be simplified to `from dolfinx.fem.petsc import assemble_matrix? Is there a reason for the renaming?

jorgensd:
Really good in general. Only one remark regarding the extraction of boundary condition on blocked forms for lifting.

python/demo/demo_mixed-poisson.py
```diff
@@ -234,13 +234,14 @@
 # conditions. Then set Dirichlet boundary values in the RHS vector `b`:
 
 # +
-b = fem.petsc.assemble_vector_nest(L)
-fem.petsc.apply_lifting_nest(b, a, bcs=bcs)
+b = fem.petsc.assemble_vector(L, kind="nest")
+bcs1 = fem.bcs_by_block(fem.extract_function_spaces(a, 1), bcs)
+fem.petsc.apply_lifting(b, a, bcs=bcs1)
```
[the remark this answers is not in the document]
garth-wells: It could, but some things to consider:
1. For repeated assembly, the `DirichletBC`s are repeatedly 'partitioned'.
2. More substantially, at the moment we cannot do:
```python
V = FunctionSpace(....)
u0, v0 = TrialFunction(V), TestFunction(V)
u1, v1 = TrialFunction(V), TestFunction(V)
a = inner(u0, v0) * dx + inner(u1, v1) * dx
...
bcs = [....]
A = assemble_matrix(a, bcs=bcs)
``` 
because we can't distinguish between the `A00` and `A11` blocks in `A` because they are both on $V \times V$, or between the `A01` and `A01` blocks in `A` because they are also on.  $V \times V$. This would work if we could do:
```python
V = FunctionSpace(....)
u0, v0 = TrialFunction(V), TestFunction(V)
a = inner(u0, v0) * dx + inner(u1, v1) * dx
...
bcs0 = [[...], [...]]   # Constrained 'rows', where bcs0[i] are the bcs on row block i
bcs1 = [[...], [...]]   # Constrained 'columns', , where bcs1[i] are the bcs on column block i
A = assemble_matrix(a, bcs0=bcs0, bcs1=bcs1)
``` 
At present we get around this by doing (note the use of `FunctionSpace.clone` below))
```python
V0 = FunctionSpace(....)
V1 = V0.clone()
u0, v0 = TrialFunction(V0), TestFunction(V0)
u1, v1 = TrialFunction(V1), TestFunction(V1)
a = inner(u0, v0) * dx + inner(u1, v1) * dx
...
bcs = [....]
A = assemble_matrix(a, bcs=bcs)
``` 
The above applies to `apply_lifting` too. With the change, we could do
```python
V = FunctionSpace(....)
u0, v0 = TrialFunction(V), TestFunction(V)
u1, v1 = TrialFunction(V), TestFunction(V)
a = inner(u0, v0) * dx + inner(u1, v1) * dx
...
bcs1 = [[...], [...]]
apply_lifting(b0, [a00, a01],  bcs1=bcs1)
``` 
(note the absence of `FunctionSpace.clone`).

python/dolfinx/fem/assemble.py
garth-wells: I really don't know how NumPy arrays should be properly typed. I would prefer to not change this in this PR, and in a separate PR fix the typing across the boards.

python/demo/demo_mixed-poisson.py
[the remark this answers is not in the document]
jorgensd: > ```python
> V = FunctionSpace(....)
> u0, v0 = TrialFunction(V), TestFunction(V)
> u1, v1 = TrialFunction(V), TestFunction(V)
> a = inner(u0, v0) * dx + inner(u1, v1) * dx
> ...
> bcs = [....]
> A = assemble_matrix(a, bcs=bcs)
> ```
I would consider this a way of trying to create a "blocked"-problem that shouldn't really be valid. 
I don't think this code should distinguish between `u0` and `u1`, and similarly `v0` and `v1` as they are from the same function space.

I think I understand why you want to support this, as it would reduce the memory footprint by avoiding recreation of another equivalent dofmap. 

Therefore, I am up for discussing this further. My main motivation behind the suggestion is to have a single way of defining the SNES operators. However, I guess that will be challenging anyhow, due to the fact that the DirichletBC extraction for `set_bc` is not possible to unify to a single function.

python/demo/demo_navier-stokes.py
```diff
@@ -325,13 +325,20 @@ def jump(phi, n):
...
 
+
 # Assemble Stokes problem
-A = assemble_matrix_block(a_blocked, bcs=bcs)
+A = assemble_matrix(a_blocked, bcs=bcs)
 A.assemble()
-b = assemble_vector_block(L_blocked, a_blocked, bcs=bcs)
+
+b = assemble_vector(L_blocked, kind=PETSc.Vec.Type.MPI)
+bcs1 = fem.bcs_by_block(fem.extract_function_spaces(a_blocked, 1), bcs)
```
jorgensd, line 334: Do you want to keep it in this way for coverage of the API?

jorgensd (approved):
Overall a very good unification of the API.
Future work is to use proper type hints everywhere and reconsider how we pass bcs to lifiting (what options to support).

python/dolfinx/fem/petsc.py
```diff
@@ -112,18 +110,48 @@ def create_vector(
...
+           (0, 12, 28)
+           >>> offsets1
+           (28, 32, 35)
+           >>> b0_owned = b.array[offsets0[0]:offsets0[1]]
+           >>> b0_ghost = b.array[offsets1[0]:offsets1[1]]
+           >>> b1_owned = b.array[offsets0[1]:offsets0[2]]
+           >>> b1_ghost = b.array[offsets1[1]:offsets1[2]]
+
+
+    3. If ``kind`` is ``PETSc.Vec.Type.NEST``, a PETSc nested vector (a
```
jhale, line 144: Should this also say whether L is a sequence of linear forms, a single linear form, or both?

python/dolfinx/fem/petsc.py
```diff
@@ -182,30 +241,96 @@ def create_matrix(
...
+       (ii) ghost entries for each ``L[i]``. See :func:`create_vector`
+       for a description of the offset blocks.
+
+    3. If ``L`` is a sequence of linear forms and ``kind`` is
+       ``PETSc.Vec.Type.NEST``, the forms are assembled into a PETSc
+       nested vector ``b`` (a nest of ghosted PETSc vectors) such that
+       ``L[i]`` is assembled into into the ith nested matrix in ``b``.
+
+    Constants and coefficients that appear in the forms(s) can be passed
+    to avoid re-computation of constants and coefficients. The functions
```
jhale, line 280: to avoid their re-computation.

jhale, line 280: re-computation -> re-packing (consistent with terminology of function names, makes link with functions mentioned below clearer)

python/dolfinx/fem/petsc.py
```diff
@@ -182,30 +241,96 @@ def create_matrix(
...
+        typing.Union[
+            dict[tuple[IntegralType, int], npt.NDArray],
+            Iterable[dict[tuple[IntegralType, int], npt.NDArray]],
+        ]
+    ] = None,
+) -> PETSc.Vec:
+    """Assemble linear form(s) into a PETSc vector.
+
+    The vector ``b`` must have been initialized with a size/layout that
+    is consistent with the linear form. The PETSc vector ``b`` is
```
jhale, line 327: Remove PETSc, already explicit by typing.

python/dolfinx/fem/petsc.py
```diff
@@ -214,201 +339,97 @@ def _assemble_vector_vec(b: PETSc.Vec, L: Form, constants=None, coeffs=None) ->
...
+    #. If ``a`` is a rectangular array of forms the forms in ``a`` are
+       assembled into a matrix such that::
 
-    return b
+            A = [a_00 ... a_0n]
+                [a_10 ... a_1n]
+                [     ...     ]
+                [a_m0 ..  a_mn]
 
+       a. If ``kind`` is a ``PETSc.Mat.Type`` (other than
```
jhale, line 410: Couldn't easily follow the logic in the brackets at the back end of this statement.

python/dolfinx/fem/petsc.py
```diff
@@ -214,201 +339,97 @@ def _assemble_vector_vec(b: PETSc.Vec, L: Form, constants=None, coeffs=None) ->
...
 
-    The returned matrix is not finalised, i.e. ghost values are not
-    accumulated.
+    Constants and coefficients that appear in the forms(s) can be passed
+    to avoid re-computation of constants and coefficients. The functions
+    :func:`dolfinx.fem.pack_constants` and
+    :func:`dolfinx.fem.pack_coefficients` can be called.
 
     Note:
         The returned matrix is not 'assembled', i.e. ghost contributions
```
jhale, line 426: The returned matrix is not finalised, i.e. ghost values are not accumulated.

python/demo/demo_mixed-poisson.py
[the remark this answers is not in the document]
garth-wells: Code updated to handle both cases. Not ideal for testing, and allows us to postpone a decision.

python/demo/demo_navier-stokes.py
garth-wells: Yes

PR #3679, opened 2025-04-01
Create `EntityMap` class for handling relations between mesh topologies

jorgensd:
An entity map takes in:
Two mesh topologies, the topological dimension one wants to have a relation between, and two equally long lists
where `entities0[i]` is in `topology0`, while `entities1[i]` is the corresponding entity in `topology1`.

This makes the `entitymaps` constructor alot easier in many cases.
The "map" is two-sided, meaning that it doesn't need to change if we switch the integration domain.

Internally (in the form constructor), we use the same logic, mapping entities, with some extra sanity checks.


garth-wells:
Close in favour of #3780.

PR #3680, opened 2025-04-01
Correct usage of input coefficients for `apply_lifting` of blocked operators.

jorgensd:
Patch for #3668.
Also generalizes the assembly checks to different kinds of matrices.


garth-wells (approved):
Could a test be added?

PR #3684, opened 2025-04-03
Proposal for supporting blocked problems (not nest) in `LinearProblem`

jorgensd:
With the unification of `create_*`, `assemble_*` and `assign` for various PETSc operators, I suggest we let `LinearProblem` support non-blocked and blocked problems, i.e.

```python
problem = LinearProblem(a, L, [bc0, bc1], petsc_options={"ksp_type": "preonly",
                                                         "pc_type": "lu",
                                                         "pc_factor_mat_solver_type":
                                                           "mumps"})
problem = LinearProblem([[a00,a01],[None, a11]], [L0, L1], bcs=[bc0, bc1],
                        u=[uh0, uh1])
```
For now, I've chosen to force the user to specify `u` for blocked problems, as it is hard extract the correct function spaces (`dolfinx.fem.FunctionSpace`, python class) from the nested list of forms with the current functions (`extract_function_spaces` work on the C++ function-spaces).

This also introduces an immediate solution vector `_x`, in the same fashion as within the proposed SNES solver.


francesco-ballarin:
Looks like a reasonable interface to me. The only issue I see: what if the user wants a MatNest assembly?

jorgensd:
> Looks like a reasonable interface to me. The only issue I see: what if the user wants a MatNest assembly?

MatNest can't be used for direct solves (as far as I am aware).

francesco-ballarin:
> MatNest can't be used for direct solves (as far as I am aware).

I do understand that, but in principle the user could choose something different from

```
petsc_options={"ksp_type": "preonly",
                                                         "pc_type": "lu",
                                                         "pc_factor_mat_solver_type":
                                                           "mumps"}
```

that is compatible with MatNest. I do understand that it will be a kind of advanced user that probably will have no trouble to set up the KSP object on their own, but still it would feel odd to have Nest-compatible SNES and Nest-incompatible linear solver.

garth-wells:
The logic around `kind` looks complicated. I'd suggest passing `kind` to initialiser. If the user asks for a matrix type and a solver that are not compatible, let PETSc raise the exception.

python/dolfinx/fem/petsc.py
```diff
@@ -771,41 +772,58 @@ def __init__(
...
         if u is None:
-            # Extract function space from TrialFunction (which is at the
-            # end of the argument list as it is numbered as 1, while the
-            # Test function is numbered as 0)
-            self.u = _Function(a.arguments()[-1].ufl_function_space())
+            try:
+                # Extract function space from TrialFunction (which is at the
+                # end of the argument list as it is numbered as 1, while the
+                # Test function is numbered as 0)
+                self._u = _Function(a.arguments()[-1].ufl_function_space())
```
jorgensd, line 813:
```suggestion
                self._u = _Function(L.arguments()[0].ufl_function_space())
```

python/dolfinx/fem/petsc.py
```diff
@@ -771,41 +772,58 @@ def __init__(
...
 
         if u is None:
-            # Extract function space from TrialFunction (which is at the
-            # end of the argument list as it is numbered as 1, while the
-            # Test function is numbered as 0)
-            self.u = _Function(a.arguments()[-1].ufl_function_space())
+            try:
+                # Extract function space from TrialFunction (which is at the
+                # end of the argument list as it is numbered as 1, while the
+                # Test function is numbered as 0)
```
jorgensd, line 812:
```suggestion
                # Extract function space for unknown from the right hand side of the equation.
```

jorgensd:
I've made "kind" input to `LinearProblem`.
It now accepts "nest" as well, and it creates the functions that stores the solution, regardless of the type of system (if not supplied by the user).

I've added a test that check that all modes work in serial/parallel (here we can use mumps+nest as it can invert each of the systems individually).

python/dolfinx/la/petsc.py
```diff
@@ -29,6 +29,15 @@
 __all__ = ["assign", "create_vector", "create_vector_wrap"]
 
 
+def ghostUpdate(x: PETSc.Vec, insert_mode: PETSc.InsertMode, scatter_mode: PETSc.ScatterMode):  # type: ignore
```
francesco-ballarin, line 32: I do understand where the name is coming from, but for consistency with the rest of dolfinx codebase maybe we want to call it `ghost_update`?

garth-wells: I think this function should be 'private'. We aim to minimise the wrapping of PETSc functionality and prefer users to use the petsc4py interfaces.

garth-wells:
Looking at his more closely, it's not clear to me what need this change addresses. Block/nest systems are usually constructed to support block-type preconditioners. But, it's not obvious that the changes support this. E.g., how should the standard block-preconditioned solvers implemented in the Stokes demo be implemented via the `LinearProblem` interface?

jorgensd:
> Looking at his more closely, it's not clear to me what need this change addresses.

For mixed dimensional/submesh problems, one has to use blocked problems (either manually or through `ufl.MixedFunctionSpace`), where one doesn't necessarily use blocked-preconditioning.

> s. But, it's not obvious that the changes support this. E.g., how should the standard block-preconditioned solvers implemented in the Stokes demo be implemented via the `LinearProblem` interface

We can of course add a `set_preconditioner` function to the `LinearProblem`, which creates the `P` matrix, and sets it to the krylov solver. I guess it could even be an input argument to the constructor.

python/dolfinx/fem/petsc.py
```diff
@@ -771,41 +772,54 @@ def __init__(
...
+            kind: The PETSc matrix and vector type. See :func:`create_matrix` for options.
 
         Example::
 
             problem = LinearProblem(a, L, [bc0, bc1], petsc_options={"ksp_type": "preonly",
                                                                      "pc_type": "lu",
                                                                      "pc_factor_mat_solver_type":
                                                                        "mumps"})
+
+            problem = LinearProblem([[a00,a01],[None, a11]], [L0, L1], bcs=[bc0, bc1],
```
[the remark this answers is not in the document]
jorgensd: Since it is i a comment?

jorgensd:
> Looking at his more closely, it's not clear to me what need this change addresses. Block/nest systems are usually constructed to support block-type preconditioners. But, it's not obvious that the changes support this. E.g., how should the standard block-preconditioned solvers implemented in the Stokes demo be implemented via the `LinearProblem` interface?

I've now added the option to pass in a preconditioner to the LinearProblem at construction.

francesco-ballarin (approved):

Issue #3582, opened 2024-12-30
Improvements to VTKHDF format I/O


chrisrichardson:
### Describe new/missing feature

Various features can be added to the VTKHDF I/O format:

- [ ] Handling of input markers for cell and boundary data
- [ ] Support of higher-order geometry
- [ ] Paraview compatibility
- [ ] Function output

* Paraview does not currently support the vtkLagrangePyramid type, which is needed for higher-order visualisation of meshes containing pyramids. See https://gitlab.kitware.com/paraview/paraview/-/issues/22840
* Paraview fails to recognise 2D data for geometry for 2D meshes (probably needs padding to 3D).

jorgensd:
> * Paraview fails to recognise 2D data for geometry for 2D meshes (probably needs padding to 3D).

Ref the Python implementation we had; I stored a variable gdim which is then used to prune the input coordinates.

PR #3686, opened 2025-04-04
Extend VTKHDF file format to read/write higher order grids

jorgensd:
- VTKHDF now supports reading and writing higher order grids (arbitrary order).
    - CoordinateElement for degree > 2 will be equispaced.
    - The support includes mixed topology.
         - Test using a mesh with second order quadrilateral and second order triangular cells have been added.
- Added special handling of 1D and 2D geometries (`gdim` argument in read constructor).
    - This is because VTKHDF doesn't have a notion of 2D geometries, in the same way XDMF supported `XY` points and `XYZ` points.

Partially addresses #3582


cpp/dolfinx/io/VTKHDF.h
```diff
@@ -199,7 +200,40 @@ mesh::Mesh<U> io::VTKHDF::read_mesh(MPI_Comm comm, std::string filename)
   std::vector<std::uint8_t> types
       = io::hdf5::read_dataset<std::uint8_t>(dset_id, local_cell_range, true);
   H5Dclose(dset_id);
-  std::vector<std::uint8_t> types_unique(types.begin(), types.end());
+  std::vector<std::array<std::uint8_t, 2>> types_unique;
+  types_unique.reserve(types.size());
+
+  // Create reverse map from VTK to dolfinx cell types
+  std::map<std::uint8_t, mesh::CellType> vtk_to_dolfinx;
+  const std::vector<mesh::CellType> dolfinx_cells
```
garth-wells, line 208: Use a `constexpr std::array` for `dolfinx_cells`

cpp/dolfinx/io/VTKHDF.h
```diff
@@ -215,36 +249,31 @@ mesh::Mesh<U> io::VTKHDF::read_mesh(MPI_Comm comm, std::string filename)
...
                    recv_offsets.begin() + 1);
-  std::vector<std::uint8_t> recv_types(recv_offsets.back());
-  MPI_Allgatherv(types_unique.data(), count, MPI_UINT8_T, recv_types.data(),
-                 recv_count.data(), recv_offsets.data(), MPI_UINT8_T, comm);
+
+  MPI_Datatype compound_type;
+  MPI_Type_contiguous(2, MPI_UINT8_T, &compound_type);
+  MPI_Type_commit(&compound_type);
+
+  std::vector<std::array<std::uint8_t, 2>> recv_types(recv_offsets.back());
```
garth-wells, line 257: This look very wrong. You need to use a flat `std::vector` with length `2 * recv_offsets.back()`.

Passing `recv_types.data()` is a pointer to the first `std::array`, rather than a pointer to the first `std::uint8_t`

cpp/dolfinx/io/VTKHDF.h
```diff
@@ -215,36 +249,31 @@ mesh::Mesh<U> io::VTKHDF::read_mesh(MPI_Comm comm, std::string filename)
...
-  }
-
-  // Map from VTKCellType to index in list of cell types
-  std::map<std::uint8_t, std::int32_t> type_to_index;
+  // Map from VTKCellType to index in list of (cell types, degree)
+  std::map<std::array<std::uint8_t, 2>, std::int32_t> type_to_index;
   std::vector<mesh::CellType> dolfinx_cell_type;
+  std::vector<std::uint8_t> dolfinx_cell_degree;
+  std::vector<std::uint32_t> num_nodes_per_cell;
   for (std::size_t i = 0; i < recv_types.size(); ++i)
```
garth-wells, line 271: Looks like a range-based loop could be used.

cpp/dolfinx/io/VTKHDF.h
```diff
@@ -277,18 +303,20 @@ mesh::Mesh<U> io::VTKHDF::read_mesh(MPI_Comm comm, std::string filename)
...
     for (std::size_t k = 0; k < offsets[j + 1] - offsets[j]; ++k)
       cells_local[type_index].push_back(topology[perm[k] + offsets[j]]);
   }
 
-  // Make first order coordinate elements
+  // Make coordinate elements
   std::vector<fem::CoordinateElement<U>> coordinate_elements;
-  for (auto& cell : dolfinx_cell_type)
-    coordinate_elements.push_back(fem::CoordinateElement<U>(cell, 1));
+  for (std::size_t i = 0; i < dolfinx_cell_type.size(); ++i)
```
garth-wells, line 315: Could use `std::transform` here.

chrisrichardson (approved):

garth-wells (dismissed):
Needs some tests.

Would be good to make the title clearer - does it support write, read, or both?

jorgensd:
> Needs some tests.
> 
> Would be good to make the title clearer - does it support write, read, or both?

Do you have any good examples at hand for mixed meshes?
I can hand-code a 2D higher order mixed mesh (1 second order triangle + 1 second order quadrilateral).

The code supports read and write.

chrisrichardson:
There is a mixed mesh fixture (see last test in `io/test_vtkdhf.py`) although this is only first-order.

garth-wells:
Single cell type tests would be enough for now.

cpp/dolfinx/io/VTKHDF.h
```diff
@@ -29,9 +29,10 @@ void write_mesh(std::string filename, const mesh::Mesh<U>& mesh);
 /// @tparam U Scalar type of mesh
 /// @param comm MPI Communicator for reading mesh
 /// @param filename Filename
+/// @param gdim Geometric dimension
 /// @return a mesh
 template <typename U>
-mesh::Mesh<U> read_mesh(MPI_Comm comm, std::string filename);
+mesh::Mesh<U> read_mesh(MPI_Comm comm, std::string filename.std::size_t gdim);
```
jorgensd, line 35:
```suggestion
mesh::Mesh<U> read_mesh(MPI_Comm comm, std::string filename, std::size_t gdim);
```

cpp/dolfinx/io/VTKHDF.h
```diff
@@ -29,9 +29,10 @@ void write_mesh(std::string filename, const mesh::Mesh<U>& mesh);
 /// @tparam U Scalar type of mesh
 /// @param comm MPI Communicator for reading mesh
 /// @param filename Filename
+/// @param gdim Geometric dimension
 /// @return a mesh
 template <typename U>
-mesh::Mesh<U> read_mesh(MPI_Comm comm, std::string filename);
+mesh::Mesh<U> read_mesh(MPI_Comm comm, std::string filename, std::size_t gdim);
```
jorgensd, line 35:
```suggestion
mesh::Mesh<U> read_mesh(MPI_Comm comm, std::string filename, std::size_t gdim=3);
```

PR #3694, opened 2025-04-08
Change interface of  `la.petsc.create_vector` to allow for general list of function spaces

schnellerhase:
The creation of PETSc vectors relies internally on a list of function spaces. Given a list of function spaces and a vector kind we can create the corresponding coefficient vector(s). Previously this functionality was not exposed to the user directly and only supported indirectly by passing a `Form` from which the spaces would then be extracted. 

This introduces/changes `la.petsc.create_vector()` which now takes a list of function spaces and returns  a `PETSc.Vec` object. The interface of `fem.petsc.create_vector()`, operating on forms, remains unchanged and now acts as a convenience wrapper to this implementation. 

Additionally, the dispatching internally has been reworked, especially it no longer relies on `try-except` blocks.


python/dolfinx/la/petsc.py
```diff
@@ -22,29 +23,14 @@
 import numpy.typing as npt
 
 import dolfinx
-from dolfinx.la import IndexMap, Vector
+from dolfinx.fem.function import FunctionSpace
```
garth-wells, line 26: This is a design we avoid. Anything in an `la` module must be oblivious to `fem`.

schnellerhase: I see. How about the interface in `la` operates on list of `DofMaps` (or tuples `(IndexMap, bs)`) and the `FunctionSpace` callback is also handled in `fem`?

garth-wells: `DofMap` is a no-no because it's a `fem` concept, but `(IndexMap, bs)` would be ok.

python/dolfinx/la/petsc.py
```diff
@@ -63,6 +49,79 @@ def create_vector_wrap(x: Vector) -> PETSc.Vec:  # type: ignore[name-defined]
...
+        A PETSc vector with the prescribed layout. The vector is not initialised to zero.
+    """
+    if len(maps) == 1:
+        # Single space case
+        index_map, bs = maps[0]
+        ghosts = index_map.ghosts.astype(PETSc.IntType)  # type: ignore[attr-defined]
+        size = (index_map.size_local * bs, index_map.size_global * bs)
+        return PETSc.Vec().createGhost(ghosts, size=size, bsize=bs, comm=index_map.comm)  # type: ignore
+
+    if kind is None or kind == PETSc.Vec.Type.MPI:
```
francesco-ballarin: Can this part be factored out in a standalone function? It's basically repeated verbatim in `dolfinx.fem.petsc._assign_block_data`, and it would be good to have a function from `dolfinx.la` to call from there

python/dolfinx/fem/petsc.py
```diff
@@ -112,27 +112,29 @@ def fn(form):
 
 
 def create_vector(
-    L: typing.Union[Form, Iterable[Form]], kind: typing.Optional[str] = None
+    container: typing.Union[Form, _FunctionSpace, Iterable[Form], Iterable[_FunctionSpace]],
```
francesco-ballarin: If possible I would change the signature to `create_vector(container, /, kind)`, where "/" is to disallow people from using a kwarg for the first input. The rationale is that the name of the first input changed, and arguably container is uglier (less expressive) than L

python/test/unit/fem/test_assembler.py
```diff
@@ -1217,6 +1217,7 @@ def test_lifting_coefficients(self, kind):
         # Apply lifting with input coefficient
         coeffs = pack_coefficients(J)
         b = petsc_create_vector(L, kind=kind)
+        assert b.equal(create_vector([V, Q]))
```
francesco-ballarin:
```suggestion
        assert b.equal(petsc_create_vector([V, Q]))
```

python/dolfinx/la/petsc.py
```diff
@@ -87,6 +72,75 @@ def create_vector_wrap(x: Vector) -> PETSc.Vec:  # type: ignore[name-defined]
...
+        index_map, bs = maps[0]
+        ghosts = index_map.ghosts.astype(PETSc.IntType)  # type: ignore[attr-defined]
+        size = (index_map.size_local * bs, index_map.size_global * bs)
+        return PETSc.Vec().createGhost(ghosts, size=size, bsize=bs, comm=index_map.comm)  # type: ignore
+
+    if kind is None or kind == PETSc.Vec.Type.MPI:  # type: ignore[attr-defined]
+        b = dolfinx.cpp.fem.petsc.create_vector_block(maps)
+        _assign_block_data(maps, b)
+        return b
+
```
garth-wells: Close gap

python/dolfinx/la/petsc.py
```diff
@@ -87,6 +72,75 @@ def create_vector_wrap(x: Vector) -> PETSc.Vec:  # type: ignore[name-defined]
...
+    elif kind == PETSc.Vec.Type.NEST:  # type: ignore[attr-defined]
+        return dolfinx.cpp.fem.petsc.create_vector_nest(maps)
+
```
garth-wells: Close gap

python/dolfinx/fem/petsc.py
```diff
@@ -83,27 +82,30 @@
 
 
 def create_vector(
-    L: typing.Union[Form, Sequence[Form]], kind: typing.Optional[str] = None
+    container: typing.Union[Form, _FunctionSpace, Sequence[Form], Sequence[_FunctionSpace]],
```
garth-wells: This function is taking quite a few different types now. Since a `Form` always has a function space, I would suggest
```python
def create_vector(V : typing.Union[_FunctionSpace, Sequence[_FunctionSpace]], 
                  kind: typing.Optional[str] = None) -> PETSc.Vec
````

Comment (schnellerhase) [reply python/dolfinx/fem/petsc.py]: Dropping support for creation of vectors based on forms completely?

Comment (garth-wells) [reply python/dolfinx/fem/petsc.py]: > Dropping support for creation of vectors based on forms completely?

Yes. Keep the interface tight.

**Review by garth-wells — APPROVED** [review]

**Comment by schnellerhase** [comment]
> Shouldn't then this also be done for matrices to stay consistent, i.e. have an extract_function_spaces(bilinear_form), and create_matrix(nested_list_of_spaces)?

Sounds good - lets do that in another PR.

File: python/test/unit/fem/test_assembler.py [hunk 8]
```diff
@@ -1206,7 +1206,7 @@ def test_lifting_coefficients(self, kind):
         p = ufl.TrialFunction(Q)
         q = ufl.TestFunction(Q)
 
-        L = form([ufl.ZeroBaseForm((v,)), ufl.ZeroBaseForm((q,))])
+        # L = form([ufl.ZeroBaseForm((v,)), ufl.ZeroBaseForm((q,))])
```
garth-wells: Remove line?

schnellerhase:
92d1a1b32113bfe22995f47ccc345e5654d9953b  adapt to vert fresh ruff release https://pypi.org/project/ruff/0.13.0/

PR #3695, opened 2025-04-09
Simplify SCOTCH configuration

garth-wells:
Find SCOTCH in CMake 'CONFIG' mode.

Resolves a CI build configuration failure with PETSc 3.23.


schnellerhase:
Maybe, try with `git clone -b v23.0.0 https://gitlab.com/petsc/petsc.git petsc`? - could be that `release` behaves different at the moment.

PR #3696, opened 2025-04-09
Remove overflow issue in pybinding for `entities_to_geometry`

jorgensd:
The current logic in `entities_to_geometry` (nanobind layer), causes a division by 0/overflow error when one supplies no entities on a process.

This PR rewrites `entities_to_geometry` so that it also returns the shape of the 2D array.

Exposed with using a non-scotch partitioner in ADIOS4DOLFINx: https://github.com/jorgensd/adios4dolfinx/actions/runs/14352168908/job/40234091707

### DIFF
--- a/cpp/dolfinx/mesh/utils.h
+++ b/cpp/dolfinx/mesh/utils.h
@@ -245,10 +245,9 @@ std::vector<T> h(const Mesh<T>& mesh, std::span<const std::int32_t> entities,
     return std::vector<T>(entities.size(), 0);
 
   // Get the geometry dofs for the vertices of each entity
-  const std::vector<std::int32_t> vertex_xdofs
+  const auto [vertex_xdofs, xdof_shape]
       = entities_to_geometry(mesh, dim, entities, false);
-  assert(!entities.empty());
-  const std::size_t num_vertices = vertex_xdofs.size() / entities.size();
+  const std::size_t num_vertices = xdof_shape[1];
 
   // Get the  geometry coordinate
   std::span<const T> x = mesh.geometry().x();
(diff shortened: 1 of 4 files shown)


python/dolfinx/wrappers/mesh.cpp
```diff
@@ -482,10 +482,13 @@ void declare_mesh(nb::module_& m, std::string type)
          nb::ndarray<const std::int32_t, nb::ndim<1>, nb::c_contig> entities,
          bool permute)
       {
-        std::vector<std::int32_t> idx = dolfinx::mesh::entities_to_geometry(
-            mesh, dim, std::span(entities.data(), entities.size()), permute);
-        return as_nbarray(std::move(idx),
-                          {entities.size(), idx.size() / entities.size()});
+        std::pair<std::vector<std::int32_t>, std::array<std::size_t, 2>>
```
garth-wells, line 485: Use
```cpp
auto [foo, idx_shape] = dolfinx::mesh::entities_to_geometry(mesh, dim, std::span(entities.data(), entities.size()), permute);
```

cpp/dolfinx/mesh/utils.h
```diff
@@ -309,10 +308,10 @@ std::vector<T> cell_normals(const Mesh<T>& mesh, int dim,
 
   // Find geometry nodes for topology entities
   std::span<const T> x = mesh.geometry().x();
-  std::vector<std::int32_t> geometry_entities
+  const auto [geometry_entities, eshape]
       = entities_to_geometry(mesh, dim, entities, false);
 
-  const std::size_t shape1 = geometry_entities.size() / entities.size();
+  const std::size_t shape1 = eshape[1];
```
garth-wells, line 314: No need for a temporary.

garth-wells (changes requested):

PR #3697, opened 2025-04-09
Add dolfinx dependencies in user docker image as requirement and run end-user Docker image build in CI

schnellerhase:
Bump of PETSc version last night https://github.com/FEniCS/dolfinx/pull/3689 caused SCOTCH to not be picked up anymore in the end user image build, the pipeline passed however. 

This changes the used dolfinx dependencies in the end user image to required - thus resulting in a failed pipeline if they are only partially found.  

Additionally activates the end-user docker builds as part of the standard CI runs (dev-env not feasible due to runtime) to continuously test for it working (or at least building).


jhale:
Looks good but I think there is a typo on the new cmake flags, or, we use the other form -DREQUIRE_*

schnellerhase:
I think with leading `_` is correct, see https://github.com/FEniCS/dolfinx/blob/main/cpp/CMakeLists.txt#L115C7-L115C22

jhale:
Ah right. They are private implementation variables - use the ones at the outer if statement to make a dependency required.

schnellerhase:
Now I understand, `DOLFINX_ENABLE_*` sets the corresponding  `_REQUIRE_*` and thus enforces it, if not provided just uses everything that it finds. Did not know `DOLFINX_ENABLE_*` implied non optional already - will change.

PR #3699, opened 2025-04-12
Use `DOLFINX_ENABLE_*` for end-user Docker

schnellerhase:
Follow up to discussion in https://github.com/FEniCS/dolfinx/pull/3697


jhale:
This may break partial reconfigure with cmakr behaviour - I didn't do it the original way for no reason!

PR #3706, opened 2025-04-13
Set max doc length to 75 for `python/dolfinx/` and `python/demo/`

schnellerhase:
Sphinx requires max line length of 79 see https://documentation-style-guide-sphinx.readthedocs.io/en/latest/style-guide.html#line-length.  Dolfinx seems to be already mostly formatted with `max-doc-length = 75`  - this is however not checked with ruff and causes reformatting issues when the doc strings are changed over time.

Proposes to enable a ruff check for the specified doc string line length for the documentation relevant Python directories.


python/dolfinx/pyproject.toml
```diff
@@ -0,0 +1,5 @@
+[tool.ruff]
+extend = "../pyproject.toml"
```
garth-wells, line 2: What does this line do?

schnellerhase: Tells ruff to pick up the config file given by the relative path and extend it with the definitions in this file. Thefore the doc line formatting does not get applied to for example `python/test`

garth-wells: Should we apply the line length check for tests too?

schnellerhase: As tests are not passed to sphinx it is not strictly necessary. Consistency wise, it would not require these specialised ruff configs, which sounds preferable.

schnellerhase: After having a look into what changes would be necessary, it breaks the syntax for commented out test cases. They would need to be reformatted to account for the different max line length.

garth-wells (approved):

PR #3712, opened 2025-04-16
Fix MacOS actions

chrisrichardson:
* Adds /usr/local/lib to library path
* Fixes some compiler warnings


cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -368,7 +368,7 @@ class MatrixCSR
     std::adjacent_difference(std::next(_val_recv_disp.begin()),
                              _val_recv_disp.end(), val_recv_count.begin());
 
-    int status = MPI_Ineighbor_alltoallv(
+    [[maybe_unused]] int status = MPI_Ineighbor_alltoallv(
```
garth-wells, line 371: Below shouldn't be an assertion on `status`. It should call `dolfinx::MPI::check_error`.

PR #3716, opened 2025-04-27
Introduce `BlockSize`

schnellerhase:
In performance critical parts some block sizes are optimized for by compiling explicit versions with the block size being provided as a compile time constant. At the same time general runtime block sizes are supported through an argument to these functions.

This causes
1. Code duplication: one path for the runtime and one for the compile time definitions of the block sizes, and
2. duplicate input of the block sizes: once as template argument once as argument (matching of both is only asserted does not raise in `Release`)

Introduces a `BlockSize` concept that either holds a runtime `int` or a compile time `std::integral_constant<int, bs>` which allows to generate code paths explicitly for certain sizes, while maintaining a shared code path in both cases.  

This is based on a more general concept of an optionally compile time valued `ConstexprType<T, V>`. It stores a value of type `T` in the container type `V`. If runtime valued, then `T = V`. For compile time, usually `T = std::integral_constant<T, ...>`.

Future applications:
- [ ] form packing optimizes for block sizes 1,2,3 - vector assembly for 1,3: is this miss match intentional?
- [ ] matrix operation routines (in particular, custom CSR assembler routines for arbitrary block sizes, ref https://github.com/FEniCS/dolfinx/blob/f1daede196ad068fdf8a508ea7795c21c2150add/python/dolfinx/wrappers/assemble.cpp#L348-L441. )


jhale:
Looks very nice. Could we review the basic approach before you spend lots more time on it?

schnellerhase:
Sure thing. Should be good to go as is and can be extended further when approved. One neat byproduct, that these changes would allow for, are non compile time sized operations on the `MatrixCSR` which we are currently missing.

chrisrichardson (approved):
Looking good.

garth-wells:
Looks really neat.
- Should the name be more generic, it's basically a runtime or templated integer. I can think of applications outside of block size, e.g. geometric dimension, where it could be useful.
- Should it support different integer types?
- Could tests be added to check that when it's a compile time integer that it really is a compiler time integer?

schnellerhase:
For points 1 and 2 that should be no problem - how about: `ConstexprType` as name for the general concept?

Regarding 3: the interface to retrieve the value (here `block_size`) needs to be able to produce both a runtime value and a compile time value. Therefore it can not be marked `constexpr`. Testing for in lining of the compile time variant is also not straight forward as this remains in all cases a compiler decision. Best way to check for its effect, I assume, would be with a benchmark of those cases.

garth-wells:
> Regarding 3: the interface to retrieve the value (here `block_size`) needs to be able to produce both a runtime value and a compile time value. Therefore it can not be marked `constexpr`. Testing for in lining of the compile time variant is also not straight forward as this remains in all cases a compiler decision. Best way to check for its effect, I assume, would be with a benchmark of those cases.

I don't like relying on the compiler to inline things that we know are known at compile time. We have avoided this in the past and preferred being explicit over relying on the compiler and then not knowing what the compiler does.

schnellerhase:
It would be best if the `block_size/value` function would be `constexpr` for the compile time case. I will try if I can recover that behaviour.

schnellerhase:
It think I have a fix: `value(ConxtexprType<T, V>)` is now `constexpr` for `is_compile_v<T, V> == True` and otherwise not. The test case showcases that we can assert during compile time now.  (Block size is not yet adapted).

jhale:
I have reviewed the PR and the above discussion and all comments have been addressed.

garth-wells:
I’d like to discuss this one more. Not quite convinced that it’s a simplification.

cpp/dolfinx/fem/pack.h
```diff
@@ -51,29 +52,19 @@ get_cell_orientation_info(const Function<T, U>& coefficient)
...
-    }
-    else
-    {
-      assert(_bs == bs);
-      const int pos_c = _bs * i;
-      const int pos_v = _bs * dofs[i];
-      for (int k = 0; k < _bs; ++k)
-        coeffs[pos_c + k] = v[pos_v + k];
-    }
+    int bs = block_size(_bs);
```
garth-wells: What's the point of passing `BlockSize auto _bs` if the implementation using `int bs`?

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -95,10 +97,11 @@ void _lift_bc_cells(
...
 
-  const auto [dmap0, bs0, cells0] = dofmap0;
-  const auto [dmap1, bs1, cells1] = dofmap1;
-  assert(_bs0 < 0 or _bs0 == bs0);
-  assert(_bs1 < 0 or _bs1 == bs1);
+  const auto [dmap0, _bs0, cells0] = dofmap0;
+  const auto [dmap1, _bs1, cells1] = dofmap1;
+
+  int bs0 = block_size(_bs0);
+  int bs1 = block_size(_bs1);
```
garth-wells: This isn't `constexpr`, so we lose the compile-time 'const-ness'.

cpp/dolfinx/fem/pack.h
schnellerhase: It allows for inovocations with both a  compile time block size, of  `std::integral_constant` type, or a runtime one, of  `int` type.  In both cases however we then extract an integral block size `bs` which gets used for the computations.

cpp/dolfinx/fem/assemble_vector_impl.h
schnellerhase: We loose the `constexpr int = ...` version. However in the case of a compile time value the `block_size` callback has signature `constexpr int block_size()`. Meaning the compiler will see `int bs = constexpr int`.

cpp/dolfinx/fem/pack.h
garth-wells: My point is that `int bs;` isn't `constexpr`

cpp/dolfinx/fem/assemble_vector_impl.h
garth-wells: The block size get assigned to the runtime `int` - it's not clear or guaranteed that the compiler can exploit that the block size is a compile time known.

schnellerhase: Type of the unpacked value is now changed to `auto`. Therefore the type is inferred from  the return type of the `block_size` function, which is `int` for the runtime case and `consteval int` for a compile time value.

cpp/dolfinx/common/constexpr_type.h
```diff
@@ -0,0 +1,53 @@
...
+/// @private Concept defining a variadic compile time or runtime variable. T
+/// indicates the type that is stored and V the value. Either V equals T, i.e.
+/// it is a runtime variable or V defines a compile time value V::value of type
+/// T.
+/// @tparam T type of the value to be stored.
+/// @tparam V container type. Usually T for a runtime variable or a
+/// std::integral_constant<T, ...> for a compile time constant.
+template <typename T, typename V>
+concept ConstexprType = std::is_same_v<T, V> || (requires {
+                          std::is_fundamental_v<T>;
```
jhale: I think there should be a `requires` here - without requires,  isn't it just an expression?

schnellerhase: That's a tricky pitfall.

cpp/dolfinx/common/constexpr_type.h
```diff
@@ -0,0 +1,54 @@
...
+/// T.
+/// @tparam T type of the value to be stored.
+/// @tparam V container type. Usually T for a runtime variable or a
+/// std::integral_constant<T, ...> for a compile time constant.
+template <typename T, typename V>
+concept ConstexprType = std::is_same_v<T, V> || (requires {
+                          requires std::is_fundamental_v<T>;
+                          typename V::value_type;
+                          requires std::same_as<typename V::value_type, T>;
+                          requires std::same_as<decltype(V::value), const T>;
```
schnellerhase:
```suggestion
                          requires std::same_as<std::remove_cv_t<decltype(V::value)>, T>;
```

schnellerhase: That would allow for non const returns, we don't care if `V::value` is a `const T` or `T`, they are fundamental types anyways.

garth-wells:
One concern with the `BlockSize`/`ConstexprType` mechanism as currently used: `block_size()` collapses the tag to a plain `int` *before* it reaches the hot loops, e.g. in `assemble_cells`:

```cpp
const auto [dmap, _bs, cells0] = dofmap;
auto bs = block_size(_bs);          // BS<1>, BS<3>, or int -> always `int` here
...
for (int k = 0; k < bs; ++k)
  b[bs * dofs[i] + k] += be[bs * i + k];
```

`block_size()` being `consteval` only constrains *when the call itself is evaluated* — it says nothing about the plain `auto bs` variable it initializes. So the `BS<1>`/`BS<3>` instantiations of `assemble_cells`/`assemble_entities`/`assemble_interior_facets`/`spmv` now generate the *same loop* as the generic runtime-`bs` path. The old `if constexpr (_bs > 0)` gave a type-system guarantee that the specialized branch has a literal trip count and is eligible for unrolling/vectorization unconditionally; here that guarantee is gone and replaced by "hopefully the inliner + constant propagation recovers it." That won't happen at all in a Developer/Debug build (no optimization), and isn't guaranteed in Release either once there's enough call-chain depth (tuple destructure + function call boundary). So this risks quietly losing the optimization the compile-time paths exist for, while keeping the appearance of specialized code paths via the `BS<1>()`/`BS<3>()` call sites.

Suggestion: keep the `if constexpr` dispatch at the loop itself instead of erasing the compile-time-ness beforehand, but factor only the loop body into a shared generic lambda so the surrounding function isn't duplicated:

```cpp
auto scatter = [&](auto bs)   // bs: BS<N> (has ::value) or plain int
{
  if constexpr (common::is_compile_time_v<std::int32_t, decltype(bs)>)
  {
    constexpr int n = decltype(bs)::value;   // genuine literal, unroll guaranteed
    for (int k = 0; k < n; ++k)
      b[n * dofs[i] + k] += be[n * i + k];
  }
  else
  {
    for (int k = 0; k < bs; ++k)
      b[bs * dofs[i] + k] += be[bs * i + k];
  }
};
scatter(_bs);
```

This still removes the original problem (duplicating the *entire* function per block size), but keeps the compile-time literal alive all the way to the loop bound rather than relying on the optimizer to reconstruct it from a runtime `int`.

garth-wells:
Closing as what happens it too implicit and compiler dependent in what is a performance critical part of code. Other ways to reduced code duplicate whilst preserving transparency, e.g. with lambda functions.

PR #3722, opened 2025-05-04
Clang tidy support for cpp

schnellerhase:
Introduces `clang-tidy` support to the cmake config (so far for the `cpp` build, enables all `performance` checks and introduces a CI workflow for it.

Closely follows setup of previous `basix` PR https://github.com/FEniCS/basix/pull/917


cpp/dolfinx/common/MPI.cpp
```diff
@@ -85,7 +85,7 @@ void dolfinx::MPI::check_error(MPI_Comm comm, int code)
     std::string error_string(MPI_MAX_ERROR_STRING, ' ');
     MPI_Error_string(code, error_string.data(), &len);
     error_string.resize(len);
-    std::cerr << error_string << std::endl;
+    std::cerr << error_string << '\n';
```
jhale, line 88: Is this recommended? Surprised.

cpp/dolfinx/graph/partitioners.h
```diff
@@ -17,7 +17,7 @@ namespace scotch
 /// @brief PT-SCOTCH partitioning strategies.
 ///
 /// See PT-SCOTCH documentation for details.
-enum class strategy
+enum class strategy : std::uint8_t
```
jhale, line 20: What's the advantage of specifying the underlying type? Why not leave it to the compiler?

jhale (commented):
Would be good to get this in as its difficult to merge. Quite a lot of changes from pass by value to pass by const reference. Not sure about the removal of std::endl.

cpp/dolfinx/graph/partitioners.h
schnellerhase: https://clang.llvm.org/extra/clang-tidy/checks/performance/enum-size.html - default type (`int`, I believe) is larger than necessary for the number of specified options.

cpp/dolfinx/common/MPI.cpp
schnellerhase: yes, https://clang.llvm.org/extra/clang-tidy/checks/performance/avoid-endl.html - buffer flushing that is implied with `std::endl` is discouraged.

schnellerhase:
Both of these seem not super important to be honest, if we do not want the changes that those imply, we can always deactivate them one by one to our liking.

jhale:
Looks good, let's look at other clang-tidy features after 0.10 is released. I feel like SonarCloud is good but not getting a lot of active use, something useable from the terminal like clang-tidy would be great.

garth-wells:
Many of the `clang-tidy` are not desired. We deliberately pass by value whenever we can. For example, there is no need to pass `nb::ndarray` by reference. They are basically spans. `std:shared_ptr` should be passed by value, not reference.

garth-wells:
Our style approach has been to use the simplest syntax possible without measurable performance hits. `ndarray` is analogous to `std::span` which we pass by value.

We decided quite a while back to pass `std::shared_ptr` by value because it is essentially just a pointer with a reference counter. It's not passed inside hot loops so performance isn't an issue. The idea of a reference to a `std::shared_ptr` is a bit weird because it breaks the concept of reference counting.

PR #3726, opened 2025-05-10
Vertex integrals

schnellerhase:


.github/workflows/fenicsx-refs.env
```diff
@@ -1,3 +1,6 @@
+basix_repository=FEniCS/basix
 basix_ref=main
+ufl_repository=FEniCS/ufl
 ufl_ref=main
-ffcx_ref=main
+ffcx_repository=schnellerhase/fenics-ffcx
+ffcx_ref=point_measure
```
schnellerhase: Revert before merge

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -907,6 +907,94 @@ void assemble_interior_facets(
   }
 }
 
+/// @brief Execute kernel over cells and accumulate result in vector.
```
jorgensd, line 910:
```suggestion
/// @brief Execute kernel over a set of vertices and accumulate result in vector.
```

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -907,6 +907,94 @@ void assemble_interior_facets(
...
+///
+/// @tparam T  Scalar type
+/// @tparam _bs Block size of the form test function dof map. If less
+/// than zero the block size is determined at runtime. If `_bs` is
+/// positive the block size is used as a compile-time constant, which
+/// has performance benefits.
+/// @param[in] P0 Function that applies transformation `P0.b` in-place
+/// to `b` to transform test degrees-of-freedom.
+/// @param[in,out] b Aray to accumulate into.
```
jorgensd, line 919:
```suggestion
/// @param[in,out] b Array to accumulate into.
```

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -907,6 +907,94 @@ void assemble_interior_facets(
...
+/// @param[in] x Mesh geometry (coordinates).
+/// @param[in] vertices Vertex indices `(vertices.size(), 2)` - first entry
+/// holds the cell index over which the vertex is to evaluated, and the second
+/// stores the local index of the vertex within the cell.
+/// @param[in] dofmap Test function (row) degree-of-freedom data holding
+/// the (0) dofmap, (1) dofmap block size and (2) dofmap cell indices.
+/// @param[in] kernel Kernel function to execute over each cell.
+/// @param[in] constants Constant coefficient data in the kernel.
+/// @param[in] coeffs Coefficient data in the kernel. It has shape
+/// `(cells.size(), num_cell_coeffs)`. `coeffs(i, j)` is the `j`th
```
jorgensd, line 930:
```suggestion
/// `(vertices.size(), num_cell_coeffs)`. `coeffs(i, j)` is the `j`th
```

python/test/unit/fem/test_point_source.py
```diff
@@ -0,0 +1,150 @@
...
+        mesh.CellType.triangle,
+        mesh.CellType.quadrilateral,
+        mesh.CellType.tetrahedron,
+        # mesh.CellType.pyramid,
+        mesh.CellType.prism,
+        mesh.CellType.hexahedron,
+    ],
+)
+@pytest.mark.parametrize("ghost_mode", [mesh.GhostMode.none, mesh.GhostMode.shared_facet])
+def test_point_source_rank_0(cell_type, ghost_mode):
```
jorgensd, line 29: Maybe parametrize over dtype

.github/workflows/ci-spack.yml
```diff
@@ -86,9 +86,9 @@ jobs:
         run: |
           . ./spack-src/share/spack/setup-env.sh
           spack env activate .
-          pip install git+https://github.com/fenics/ufl.git@${{ env.ufl_ref }}
-          pip install git+https://github.com/fenics/basix.git@${{ env.basix_ref }}
-          pip install git+https://github.com/fenics/ffcx.git@${{ env.ffcx_ref }}
+          pip install git+https://github.com/fenics/${{ env.ufl_repository }}.git@${{ env.ufl_ref }}
+          pip install git+https://github.com/fenics/${{ env.basix_repository }}.git@${{ env.basix_ref }}
+          pip install git+https://github.com/fenics/${{ env.ffcx_repository }}.git@${{ env.ffcx_ref }}
```
schnellerhase:
```suggestion
          pip install git+https://github.com/${{ env.ufl_repository }}.git@${{ env.ufl_ref }}
          pip install git+https://github.com/${{ env.basix_repository }}.git@${{ env.basix_ref }}
          pip install git+https://github.com/${{ env.ffcx_repository }}.git@${{ env.ffcx_ref }}
```

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -942,6 +942,95 @@ void assemble_interior_facets(
...
+  if (vertices.empty())
+    return;
+
+  const auto [dmap, bs, vertices0] = dofmap;
+  assert(_bs < 0 or _bs == bs);
+
+  // Create data structures used in assembly
+  std::vector<scalar_value_t<T>> cdofs(3 * x_dofmap.extent(1));
+  std::vector<T> be(bs * dmap.extent(1));
+  std::span<T> _be(be);
```
garth-wells: Why is `_be` needed?

cpp/dolfinx/fem/pack.h
```diff
@@ -322,6 +322,39 @@ void pack_coefficients(const Form<T, U>& form,
...
+          std::span<const std::int32_t> vertices_b
+              = form.domain_coeff(IntegralType::vertex, id, coeff);
+          md::mdspan<const std::int32_t,
+                     md::extents<std::size_t, md::dynamic_extent, 2>>
+              vertices(vertices_b.data(), vertices_b.size() / 2, 2);
+          std::span<const std::uint32_t> cell_info
+              = impl::get_cell_orientation_info(*coefficients[coeff]);
+          impl::pack_coefficient_entity(
+              std::span(c), cstride, *coefficients[coeff], cell_info,
+              std::submdspan(vertices, md::full_extent, 0), offsets[coeff]);
```
garth-wells: `std::submdspan` is C++26. We're not that far a long yet!

cpp/dolfinx/fem/utils.cpp
```diff
@@ -147,15 +165,9 @@ fem::compute_integration_domains(fem::IntegralType integral_type,
...
-  switch (integral_type)
-  {
-  case IntegralType::cell:
-  {
-    entity_data.insert(entity_data.begin(), entities.begin(), entities.end());
-    break;
-  }
-  default:
+  auto cell_connectivity
+      = [&]() -> std::pair<std::shared_ptr<const graph::AdjacencyList<int>>,
```
garth-wells: I really like to avoid that capture everything `[&]`. Capture what's needed only.

cpp/dolfinx/fem/utils.cpp
```diff
@@ -170,61 +182,70 @@ fem::compute_integration_domains(fem::IntegralType integral_type,
...
+            "Use \"shared facet\"  ghost mode when creating the mesh.");
       }
     }
     break;
-    default:
-      throw std::runtime_error(
-          "Cannot compute integration domains. Integral type not supported.");
-    }
   }
+    // C++ 23
```
garth-wells: Remove comments.

cpp/dolfinx/fem/utils.h
```diff
@@ -86,6 +90,48 @@ get_cell_facet_pairs(std::int32_t f, std::span<const std::int32_t> cells,
 
   return cell_local_facet_pairs;
 }
+
+/// @brief Kernel C-pointer type.
+/// @tparam T scalar type.
+/// @tparam U geometry type.
+template <dolfinx::scalar T, std::floating_point U = scalar_value_t<T>>
```
garth-wells: Could we put the typedefs in a separate header file?

More generally, our 'utils' headers are getting very long in some cases, and they drag a lot of declarations into parts of the code where most declarations are not needed.

python/test/unit/fem/test_point_source.py
```diff
@@ -0,0 +1,160 @@
...
+    elif cell_dim == 2:
+        msh = mesh.create_unit_square(
+            comm, 4, 4, cell_type=cell_type, dtype=rdtype, ghost_mode=ghost_mode
+        )
+    elif cell_dim == 3:
+        msh = mesh.create_unit_cube(
+            comm, 4, 4, 4, cell_type=cell_type, dtype=rdtype, ghost_mode=ghost_mode
+        )
+
+    def check(form, coordinate_range, weighted=False):
```
garth-wells: Use more descriptive name?

python/test/unit/fem/test_point_source.py
```diff
@@ -0,0 +1,160 @@
...
+            comm, 4, 4, cell_type=cell_type, ghost_mode=ghost_mode, dtype=rdtype
+        )
+    elif cell_dim == 3:
+        msh = mesh.create_unit_cube(
+            comm, 4, 4, 4, cell_type=cell_type, ghost_mode=ghost_mode, dtype=rdtype
+        )
+
+    num_vertices = msh.topology.index_map(0).size_local
+
+    def check(form, coordinate_range, weighted=False):
```
garth-wells: Use more descriptive name?

garth-wells (changes requested):

cpp/dolfinx/fem/pack.h
schnellerhase: Indeed, surprised to see this compile with the set cxx standard.

cpp/dolfinx/fem/assemble_vector_impl.h
schnellerhase: The DOF transform `P0`, which gets applied to `be`, takes the `be`-argument by value. To avoid a copy `_be` (sharing the data with `be`) is created to pass. Usage is consistent with `assemble_cells` implementation https://github.com/FEniCS/dolfinx/blob/42644d68aee423790fa23f98ac3b7b7729e50b8f/cpp/dolfinx/fem/assemble_vector_impl.h#L684-L702.

cpp/dolfinx/fem/pack.h
schnellerhase: Works because the mdspan implementation just overwrites the `std` namespace. `md = std` here.

cpp/dolfinx/fem/assemble_vector_impl.h
garth-wells: I removed `_be` in a branch elsewhere.

I don't think we need it. The copy 'span -> span' and the copy 'vector -> span' are the same. A copy of the pointer and a copy of the size.

schnellerhase: Right, missed `DofTransformKernel` already accepts `span`'s only.

schnellerhase: Changed also the other occurrences.

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -228,6 +265,25 @@ T assemble_scalar(
...
+    assert(fn);
+
+    auto& [coeffs, cstride] = coefficients.at({IntegralType::vertex, i});
+
+    std::span<const std::int32_t> data = M.domain(IntegralType::vertex, i, 0);
+    assert(data.size() * cstride == coeffs.size());
+
+    md::mdspan<const std::int32_t,
+               md::extents<std::size_t, md::dynamic_extent, 2>>
+        cell_and_vertex(data.data(), data.size() / 2, 2);
```
jorgensd: If we want to be consistent with the other integral types, we should remove the temporary variable declaration cell_and_vertex, and just do the casting inside `assemble_vertices` below.

cpp/dolfinx/fem/kernel.h
```diff
@@ -0,0 +1,57 @@
...
+
+/// @brief Kernel callback type.
+/// @tparam T scalar type.
+/// @tparam U geometry type.
+template <dolfinx::scalar T, std::floating_point U = scalar_value_t<T>>
+using kern_t = std::function<void(T*, const T*, const T*, const U*, const int*,
+                                  const std::uint8_t*, void*)>;
+
+/// @brief Extract correct kernel by type from UFCx integral.
+/// @tparam T scalar type of kernel to extract.s
```
jorgensd:
```suggestion
/// @tparam T scalar type of kernel to extract.
```

cpp/dolfinx/fem/pack.h
```diff
@@ -322,6 +322,39 @@ void pack_coefficients(const Form<T, U>& form,
...
+      {
+        // Iterate over coefficients that are active in cell integrals
+        for (int coeff : form.active_coeffs(IntegralType::vertex, id))
+        {
+          // Get coefficient mesh
+          auto mesh = coefficients[coeff]->function_space()->mesh();
+          assert(mesh);
+
+          // Other integrals in the form might have coefficients defined
+          // over entities of codim > 0, which don't make sense for vertex
```
jorgensd: Is this really true? Couldn't one have a coefficient from a submesh that shared the vertex within a `dP` integral?

cpp/dolfinx/fem/utils.cpp
```diff
@@ -147,15 +165,9 @@ fem::compute_integration_domains(fem::IntegralType integral_type,
...
-  std::vector<std::int32_t> entity_data;
-  switch (integral_type)
-  {
-  case IntegralType::cell:
-  {
-    entity_data.insert(entity_data.begin(), entities.begin(), entities.end());
-    break;
-  }
-  default:
+  auto cell_connectivity = [tdim, &topology]()
```
jorgensd: Could have a more descriptive name, as it is only used for `facet` integrals, i.e get_cell_facet_connectivity?

cpp/dolfinx/fem/utils.cpp
```diff
@@ -170,60 +182,66 @@ fem::compute_integration_domains(fem::IntegralType integral_type,
...
-    switch (integral_type)
-    {
-    case IntegralType::exterior_facet:
+  std::vector<std::int32_t> entity_data;
+  switch (integral_type)
+  {
+  case IntegralType::cell:
+  case IntegralType::vertex:
+  {
+    entity_data.insert(entity_data.begin(), entities.begin(), entities.end());
```
jorgensd, line 199: Should the entities (vertices) be mapped to (cell, local_entity_index) pairs as for an exterior facet integral?

cpp/dolfinx/fem/utils.h
```diff
@@ -450,10 +456,17 @@ Form<T, U> create_form_factory(
...
+  std::array<int, 4> num_integrals_type;
+  for (std::size_t i = 0; i < num_integrals_type.size(); ++i)
     num_integrals_type[i] = integral_offsets[i + 1] - integral_offsets[i];
 
+  // Create vertices, if required
+  if (num_integrals_type[vertex] > 0)
+  {
+    mesh->topology_mutable()->create_connectivity(0, tdim);
+    mesh->topology_mutable()->create_connectivity(tdim, 0);
+  }
```
jorgensd, line 502: Usually we do not create connectivity implicitly. Maybe @garth-wells can comment on the suitability here.
At least we know that `(tdim, 0)` should exist, the inverse however, does not exist by default.

cpp/dolfinx/fem/utils.h
```diff
@@ -704,29 +669,7 @@ Form<T, U> create_form_factory(
...
-#ifndef DOLFINX_NO_STDC_COMPLEX_KERNELS
-        else if constexpr (std::is_same_v<T, std::complex<double>>)
-        {
-          k = reinterpret_cast<void (*)(T*, const T*, const T*,
-                                        const scalar_value_t<T>*, const int*,
-                                        const unsigned char*, void*)>(
-              integral->tabulate_tensor_complex128);
-        }
-#endif // DOLFINX_NO_STDC_COMPLEX_KERNELS
+        impl::kern_t<T, U> k = impl::extract_kernel<T>(integral);
```
jorgensd: This is nice!

cpp/dolfinx/fem/utils.h
```diff
@@ -780,6 +723,96 @@ Form<T, U> create_form_factory(
...
+            // Find local index of vertex within cell
+            auto cell_vertices = c_to_v->links(cell);
+            auto it = std::ranges::find(cell_vertices, vertex);
+            assert(it != cell_vertices.end());
+            std::int32_t local_index = std::distance(cell_vertices.begin(), it);
+            cell_and_vertex.push_back(local_index);
+          }
+          assert(cell_and_vertex.size() == 2 * vertices_range.size());
+          return cell_and_vertex;
+        };
```
jorgensd: This packing is usually done inside `compute_integration_domains`, and would allow for custom/controllable interfaces, as for instance shown in: https://github.com/FEniCS/dolfinx/blob/396700836046033ebaa4e71add97db3c94997168/python/test/unit/fem/test_assemble_mesh_independent_form.py#L124-L131

jorgensd: Ah, sorry, this is fine where it is, but my comment is still relevant in the above bit.

cpp/dolfinx/fem/utils.h
schnellerhase: The same happens right below for exterior or interior facets, not saying this is what we should go for, but it's consistent with other integral types. See https://github.com/FEniCS/dolfinx/pull/3726/files#diff-dd0900d50cefcee4dc6c16f86bdb12cf6fea0921ad30075c3edcd217ae504864R474

cpp/dolfinx/fem/pack.h
schnellerhase: Agree, copy paste error from the cell assembler and did not re-consider if it applied.

cpp/dolfinx/fem/utils.cpp
schnellerhase: Indeed, should behave the same way as its also an assembly of a  lower dimensional entity associated with a cell. 

This also addresses https://github.com/FEniCS/dolfinx/pull/3726/files#diff-dd0900d50cefcee4dc6c16f86bdb12cf6fea0921ad30075c3edcd217ae504864R773 - allowing the choice of cell to assemble the vertex over to be injected by the caller by different packing.

cpp/dolfinx/fem/utils.cpp
```diff
@@ -137,7 +138,24 @@ fem::compute_integration_domains(fem::IntegralType integral_type,
...
   const int tdim = topology.dim();
-  const int dim = integral_type == IntegralType::cell ? tdim : tdim - 1;
+
+  int dim = -1;
+  switch (integral_type)
+  {
+  case IntegralType::cell:
+    dim = tdim;
+    break;
+  case IntegralType::exterior_facet:
```
garth-wells, line 148: I think adding `dim = tdim - 1;` here would help readability.

cpp/dolfinx/fem/utils.h
schnellerhase: Addressed as is with the changes to `compute_integration_domain`?

cpp/dolfinx/fem/utils.cpp
schnellerhase: ~As a default in line 142?~ Got it now.

cpp/dolfinx/fem/pack.h
```diff
@@ -322,6 +322,28 @@ void pack_coefficients(const Form<T, U>& form,
         }
         break;
       }
+      case IntegralType::vertex:
+      {
+        // Iterate over coefficients that are active in cell integrals
```
michalhabera:
```suggestion
        // Iterate over coefficients that are active in vertex integrals
```

michalhabera:
> We've discussed with @schnellerhase that this needs some sanity checks.
> 
> Vertex integral, just like interior facet integral, is a bit tricky when discontinuous functions are involved. Contributions from all cells connected to the vertex are required, which is possibly not bounded (arbitrary number of cells, say N , can be connected to single vertex). This would require prepacking coefficients from all N cells, and for discontinuous Arguments assembling variable sized local tensor, etc.
> 
> Regardless, vertex integrals are very useful for continuous spaces so possibly raising an error if discontinuous space is involved might be be good enough.

@jorgensd @garth-wells Would be good to have you opinion on the above issue.

jorgensd:
> > We've discussed with @schnellerhase that this needs some sanity checks.
> > Vertex integral, just like interior facet integral, is a bit tricky when discontinuous functions are involved. Contributions from all cells connected to the vertex are required, which is possibly not bounded (arbitrary number of cells, say N , can be connected to single vertex). This would require prepacking coefficients from all N cells, and for discontinuous Arguments assembling variable sized local tensor, etc.
> > Regardless, vertex integrals are very useful for continuous spaces so possibly raising an error if discontinuous space is involved might be be good enough.
> 
> @jorgensd @garth-wells Would be good to have you opinion on the above issue.

As mentioned before (similar to the ridge integral in the previously mentioned PR) this PR assumes that we view stuff from 1 cell (kind of exterior facet/ds)-view.  I am fine with that, as a proper handling of "interior" vertex integrals would requires us to change how restrictions work in `FFCx`.

This issue also arise on the branching manifolds, where a ("+"), ("-") restriction is not sufficient.

I think what we should do is:
1. Merge this as supporting "exterior" vertex integrals. One can still use it on an interior vertex, but only with the view from a single cell, by manually packing the integration domains.
2. Start thinking about how arbitrary restrictions should be done in both UFL and FFCx, as we would need to take this into account in the generated code. An option here would be similar to the one proposed by @schnellerhase in the dual graph PR, where one specify the max number of connections as a form compiler parameter, and we add an additional index as input to FFCx kernels that would instruct on how many cells is packed for the given kernel.

schnellerhase:
Regarding point 2: Our feeling is, the underlying problem is to find a notion of defining and executing a kernel over patches - which is the collection of connected entities (possibly also for bigger distances?). A first impression we had was that maybe the kernel should be responsible for providing the necessary patch for its assembly as well (or rather a routine to compute it during packing). So given the mesh connectivity one could run such a 'patch computation' and pack this data. The kernel needs to be able to deal with variable patch sizes then. Such a setup would probably remove the necessity to differentiate between `ds` and `dS` altogether, the difference being solely in the altered patch computation if the assembled forms requires the neighbour information.

.github/workflows/fenicsx-refs.env
```diff
@@ -1,4 +1,6 @@
+basix_repository=FEniCS/basix
 basix_ref=main
+ufl_repository=FEniCS/ufl
 ufl_ref=main
-ffcx_ref=main
-
+ffcx_repository=schnellerhase/fenics-ffcx
+ffcx_ref=point_measure
```
schnellerhase:
```suggestion
ffcx_repository=FEniCS/ffcx
ffcx_ref=main
```

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -228,6 +265,24 @@ T assemble_scalar(
...
   }
 
+  for (int i : M.integral_ids(IntegralType::vertex))
+  {
+    auto fn = M.kernel(IntegralType::vertex, i, 0);
+    assert(fn);
+
+    auto& [coeffs, cstride] = coefficients.at({IntegralType::vertex, i});
+
+    std::span<const std::int32_t> data = M.domain(IntegralType::vertex, i, 0);
```
michalhabera: Would be more readable as `vertices`.

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -228,6 +265,24 @@ T assemble_scalar(
...
+    assert(data.size() * cstride == coeffs.size());
+
+    value += impl::assemble_vertices(
+        x_dofmap, x,
+        md::mdspan<const std::int32_t,
+                   md::extents<std::size_t, md::dynamic_extent, 2>>(
+            data.data(), data.size() / 2, 2),
```
michalhabera: These assumption on 2, or 4 above are really hard to find when later needed. I'd personally name these constants.

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -942,6 +941,94 @@ void assemble_interior_facets(
...
+/// than zero the block size is determined at runtime. If `_bs` is
+/// positive the block size is used as a compile-time constant, which
+/// has performance benefits.
+/// @param[in] P0 Function that applies transformation `P0.b` in-place
+/// to `b` to transform test degrees-of-freedom.
+/// @param[in,out] b Array to accumulate into.
+/// @param[in] x_dofmap Dofmap for the mesh geometry.
+/// @param[in] x Mesh geometry (coordinates).
+/// @param[in] vertices Vertex indices `(vertices.size(), 2)` - first entry
+/// holds the cell index over which the vertex is to evaluated, and the second
```
michalhabera:
```suggestion
/// holds the index of the cell adjacent to the vertex, and the second
```

cpp/dolfinx/fem/kernel.h
```diff
@@ -0,0 +1,57 @@
...
+#include <dolfinx/common/types.h>
+#include <ufcx.h>
+
+namespace dolfinx::fem::impl
+{
+/// @brief Kernel C-pointer type.
+/// @tparam T scalar type.
+/// @tparam U geometry type.
+template <dolfinx::scalar T, std::floating_point U = scalar_value_t<T>>
+using kern_c_t = void (*)(T*, const T*, const T*, const U*, const int*,
```
michalhabera: My preference would be `kernelptr_t` and `kernel_t`

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -215,19 +252,42 @@ T assemble_scalar(
     auto& [coeffs, cstride]
         = coefficients.at({IntegralType::interior_facet, i});
     std::span facets = M.domain(IntegralType::interior_facet, i, 0);
-    assert((facets.size() / 4) * 2 * cstride == coeffs.size());
+
+    constexpr std::size_t shape1 = 4;
```
michalhabera: Not what I meant here, `shape1` is not helping the reader. Something like
```cpp
constexpr int num_adjacent_cells = 2;
// Two values per each adj. cell (cell index and local facet index).
constexpr int shape1 = 2 * num_adjacent_cells;
```
in this case of interior facet and for vertices and exterior facets
```cpp
constexpr int num_adjacent_cells = 1;
// Two values per adj. cell (cell index and local vertex index).
constexpr int shape1 = 2 * num_adjacent_cells;
```
I find it very hard to read the code here and see what assumptions were made.

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -942,6 +941,94 @@ void assemble_interior_facets(
...
+    return;
+
+  const auto [dmap, bs, vertices0] = dofmap;
+  assert(_bs < 0 or _bs == bs);
+
+  // Create data structures used in assembly
+  std::vector<scalar_value_t<T>> cdofs(3 * x_dofmap.extent(1));
+  std::vector<T> be(bs * dmap.extent(1));
+
+  // Iterate over active cells
```
michalhabera:
```suggestion
  // Iterate over active vertices
```

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -942,6 +941,94 @@ void assemble_interior_facets(
...
+    std::int32_t cell = vertices(index, 0);
+    std::int32_t local_index = vertices(index, 1);
+    std::int32_t c0 = vertices0(index, 0);
+
+    // Get cell coordinates/geometry
+    auto x_dofs = md::submdspan(x_dofmap, cell, md::full_extent);
+    for (std::size_t i = 0; i < x_dofs.size(); ++i)
+      std::copy_n(&x(x_dofs[i], 0), 3, std::next(cdofs.begin(), 3 * i));
+
+    // Tabulate vector for cell
```
michalhabera:
```suggestion
    // Tabulate vector for vertex
```

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -942,6 +941,94 @@ void assemble_interior_facets(
...
+    std::ranges::fill(be, 0);
+    kernel(be.data(), &coeffs(index, 0), constants.data(), cdofs.data(),
+           &local_index, nullptr, nullptr);
+    P0(be, cell_info0, c0, 1);
+
+    // Scatter cell vector to 'global' vector array
```
michalhabera:
```suggestion
    // Scatter vertex vector to 'global' vector array
```

cpp/dolfinx/fem/kernel.h
```diff
@@ -0,0 +1,57 @@
...
+/// @tparam U geometry type.
+template <dolfinx::scalar T, std::floating_point U = scalar_value_t<T>>
+using kernelptr_t = void (*)(T*, const T*, const T*, const U*, const int*,
+                             const std::uint8_t*, void*);
+
+/// @brief Kernel callback type.
+/// @tparam T scalar type.
+/// @tparam U geometry type.
+template <dolfinx::scalar T, std::floating_point U = scalar_value_t<T>>
+using kern_t = std::function<void(T*, const T*, const T*, const U*, const int*,
```
michalhabera: You don't like `kernel_t`? Also the function is called `extract_kernel`.

python/test/unit/fem/test_vertex_integral.py
```diff
@@ -0,0 +1,184 @@
+# Copyright (C) 2025 Paul T. Kühner
```
michalhabera: These tests are a bit misplaced, there are interior/exterior facet integral tests in `test_assembler.py`.

python/test/unit/fem/test_vertex_integral.py
```diff
@@ -0,0 +1,184 @@
...
+    elif cell_dim == 2:
+        msh = mesh.create_unit_square(
+            comm, 4, 4, cell_type=cell_type, dtype=rdtype, ghost_mode=ghost_mode
+        )
+    elif cell_dim == 3:
+        msh = mesh.create_unit_cube(
+            comm, 4, 4, 4, cell_type=cell_type, dtype=rdtype, ghost_mode=ghost_mode
+        )
+
+    def check_vertex_integral_against_sum(form, coordinate_range, weighted=False):
```
michalhabera: This looks like your custom parametrization system? Why not precompute the triple of form, some marker and expected results and parametrize over that? It is not good to hide executions from pytest.

schnellerhase: I am not sure this is this simple. The marking is dependant on the mesh parallelisation which we do not control. The forms change for every invocation, either in the integrand or the measure, no two forms are the same. We can construct them all from one form, but I don't think this would increases readability.

cpp/dolfinx/fem/utils.cpp
```diff
@@ -147,15 +167,9 @@ fem::compute_integration_domains(fem::IntegralType integral_type,
...
-  std::vector<std::int32_t> entity_data;
-  switch (integral_type)
-  {
-  case IntegralType::cell:
-  {
-    entity_data.insert(entity_data.begin(), entities.begin(), entities.end());
-    break;
-  }
-  default:
+  auto get_cell_facet_connectivity = [tdim, &topology]()
```
michalhabera: Not sure these helpers are improvement. Could all be done in single switch clause, the one you have above, incl. connectivity computation and the for loop?

python/test/unit/fem/test_assembler.py
```diff
@@ -1507,3 +1507,174 @@ def test_vector_types():
...
+    elif cell_dim == 2:
+        msh = mesh.create_unit_square(
+            comm, 4, 4, cell_type=cell_type, dtype=rdtype, ghost_mode=ghost_mode
+        )
+    elif cell_dim == 3:
+        msh = mesh.create_unit_cube(
+            comm, 4, 4, 4, cell_type=cell_type, dtype=rdtype, ghost_mode=ghost_mode
+        )
+
+    def check_vertex_integral_against_sum(form, coordinate_range, weighted=False):
```
michalhabera: The algebraic marking is a bit weird, very artificial. Why not use `locate_entities`?

michalhabera (approved):
Nice work, I think this is a very useful feature.

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -772,6 +772,7 @@ void assemble_exterior_facets(
   const int num_dofs = dmap.extent(1);
   std::vector<scalar_value_t<T>> cdofs(3 * x_dofmap.extent(1));
   std::vector<T> be(bs * num_dofs);
+
```
jorgensd:
```suggestion
```

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -793,6 +794,7 @@ void assemble_exterior_facets(
     std::ranges::fill(be, 0);
     kernel(be.data(), &coeffs(f, 0), constants.data(), cdofs.data(),
            &local_facet, &perm, nullptr);
+
```
jorgensd:
```suggestion
```

cpp/dolfinx/fem/utils.cpp
```diff
@@ -170,59 +184,104 @@ fem::compute_integration_domains(fem::IntegralType integral_type,
...
-      std::ranges::set_intersection(entities, bfacets,
-                                    std::back_inserter(facets));
-      for (auto f : facets)
-      {
-        // Get the facet as a pair of (cell, local facet)
-        auto facet
-            = impl::get_cell_facet_pairs<1>(f, f_to_c->links(f), *c_to_f);
-        entity_data.insert(entity_data.end(), facet.begin(), facet.end());
-      }
+      throw std::runtime_error(
```
jorgensd: I do not think this should ever happen, as it is required to initialize the Topology (post the last refactoring):
https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/mesh/Topology.h#L56-L57
I don't think we at the moment are able to delete connectivity (even if it was made so by design).
So I guess this doesn't hurt.

cpp/dolfinx/fem/utils.cpp
```diff
@@ -170,59 +184,104 @@ fem::compute_integration_domains(fem::IntegralType integral_type,
...
+      // Use first cell for assembly over by default
+      std::int32_t cell = cells[0];
+      entity_data.push_back(cell);
+
+      // Find local index of vertex within cell
+      auto cell_vertices = c_to_v->links(cell);
+      auto it = std::ranges::find(cell_vertices, vertex);
+      assert(it != cell_vertices.end());
+      std::int32_t local_index = std::distance(cell_vertices.begin(), it);
+      entity_data.push_back(local_index);
```
jorgensd: Should this be a convenience function `get_cell_vertex_pairs`, similar to how we compute exterior and interior facets? Then we could parameterize it and prepare for "multi-vertex" views?

cpp/dolfinx/fem/utils.h
```diff
@@ -779,6 +729,95 @@ Form<T, U> create_form_factory(
...
+            // Use first cell for assembly over by default
+            std::int32_t cell = cells[0];
+            cell_and_vertex.push_back(cell);
+
+            // Find local index of vertex within cell
+            auto cell_vertices = c_to_v->links(cell);
+            auto it = std::ranges::find(cell_vertices, vertex);
+            assert(it != cell_vertices.end());
+            std::int32_t local_index = std::distance(cell_vertices.begin(), it);
+            cell_and_vertex.push_back(local_index);
```
jorgensd: Same comment as above, if we follow the structure from interior and exterior facets, having a templated function to get these entries, we remove code duplicates and future-proof ourselves for more advanced vertex integrals (especially as `v_to_c` is ordered in a non-deterministic fashion when change num procs)

python/test/unit/fem/test_assembler.py
```diff
@@ -1560,3 +1560,159 @@ def test_mixed_quadrature(dtype, method):
...
+        msh = mesh.create_unit_interval(comm, 4, dtype=rdtype, ghost_mode=ghost_mode)
+    elif cell_dim == 2:
+        msh = mesh.create_unit_square(
+            comm, 4, 4, cell_type=cell_type, dtype=rdtype, ghost_mode=ghost_mode
+        )
+    elif cell_dim == 3:
+        msh = mesh.create_unit_cube(
+            comm, 4, 4, 4, cell_type=cell_type, dtype=rdtype, ghost_mode=ghost_mode
+        )
+
```
jorgensd, line 1622: Should typing complain here as there is no catch for cell_dim!=(1,2,3) and then msh would be none?
Maybe add an explicit error message here to just avoid this.

python/test/unit/fem/test_assembler.py
```diff
@@ -1560,3 +1560,159 @@ def test_mixed_quadrature(dtype, method):
...
+    vertex_map = msh.topology.index_map(0)
+
+    def check_vertex_integral_against_sum(form, vertices, weighted=False):
```
jorgensd, line 1625: Maybe add a comment here, for future us, explaining that weighting is a concept introduced in the last part of the test, where a function `u` is introduced (which is a `P`-1 function whose value at a vertex corresponds to the global vertex index).

python/test/unit/fem/test_assembler.py
```diff
@@ -1560,3 +1560,159 @@ def test_mixed_quadrature(dtype, method):
...
+    dP = ufl.Measure("dP", domain=msh, subdomain_data=meshtags)
+
+    # Combinations of sub domains
+    check_vertex_integral_against_sum(x[0] * dP(1), vertices_1)
+    check_vertex_integral_against_sum(x[0] * dP(2), vertices_2)
+    check_vertex_integral_against_sum(x[0] * (dP(1) + dP(2)), np.arange(num_vertices))
+
+    V = fem.functionspace(msh, ("P", 1))
+    u = fem.Function(V, dtype=dtype)
+    u.x.array[:] = vertex_map.local_to_global(np.arange(num_vertices + vertex_map.num_ghosts))
```
jorgensd: Are you sure that this will always be true. Sometimes, I've seen the ghosts of the vertex maps in topology be in a different order than those of the dofs. This is why I have a `dof-to-vertex-map` in scifem, to ensure that this always holds: https://github.com/scientificcomputing/scifem/blob/a3400de35f0c4868df4840a6670bc1944a4d228b/src/scifem.cpp#L113-L151

python/test/unit/fem/test_assembler.py
```diff
@@ -1560,3 +1560,159 @@ def test_mixed_quadrature(dtype, method):
...
+    check_vertex_integral_against_sum(x[0] * v * (dP(1) + dP(2)), np.arange(num_vertices))
+
+    V = fem.functionspace(msh, ("P", 1))
+    u = fem.Function(V, dtype=dtype)
+    u.x.array[:] = vertex_map.local_to_global(np.arange(num_vertices + vertex_map.num_ghosts))
+
+    check_vertex_integral_against_sum(u * x[0] * v * ufl.dP, np.arange(num_vertices), True)
+    check_vertex_integral_against_sum(u * x[0] * v * dP(1), vertices_1, True)
+    check_vertex_integral_against_sum(u * x[0] * v * dP(2), vertices_2, True)
+    check_vertex_integral_against_sum(u * x[0] * v * (dP(1) + dP(2)), np.arange(num_vertices), True)
```
jorgensd, line 1791: Could we add an additional check where the user manually packs the `(cell, local_vertex_index)` pair in the integration domain (could just be an extra line on one of the existing tests)`, to check that manual specification of integration entities also works.

python/test/unit/fem/test_assembler.py
schnellerhase: mypy is not fully run on demos and tests yet, so still spotty.

python/test/unit/fem/test_assembler.py
schnellerhase: I think we are fine here, as we are working on the global indices. If I understand correctly the problem arises when we have the local dof packing not aligning with the order of ghost vertices.

cpp/dolfinx/fem/utils.cpp
```diff
@@ -170,59 +184,93 @@ fem::compute_integration_domains(fem::IntegralType integral_type,
...
+      throw std::runtime_error(
+          "Topology vertex-to-cell connectivity has not been computed.");
+    }
+
+    auto c_to_v = topology.connectivity(tdim, 0);
+    // Any mesh already has c->v connectivity.
+    assert(c_to_v);
+
+    return {v_to_c, c_to_v};
+  };
```
jorgensd: Remove code duplication, ref chat on slack:)

cpp/dolfinx/fem/utils.h
```diff
@@ -86,6 +91,34 @@ get_cell_facet_pairs(std::int32_t f, std::span<const std::int32_t> cells,
 
   return cell_local_facet_pairs;
 }
+
+/// Helper function to get an array of of (cell, local_vertex) pairs
+/// corresponding to a given vertex index.
```
jorgensd, line 96: Add a note in the documentation that if a vertex is connected to multiple cells (which is almost always the case),
this function picks the first one in `cells`.

Just good to have in main docs for later extensions:)

cpp/dolfinx/fem/utils.h
```diff
@@ -779,6 +756,87 @@ Form<T, U> create_form_factory(
...
+        // pack for a range of vertices a flattened list of cell index c_i and
+        // local vertex index l_i:
+        //  [c_0, l_0, ..., c_n, l_n]
+        auto get_cells_and_vertices = [v_to_c, c_to_v](auto vertices_range)
+        {
+          std::vector<std::int32_t> cell_and_vertex;
+          cell_and_vertex.reserve(2 * vertices_range.size());
+          for (std::int32_t vertex : vertices_range)
+          {
+            std::array<std::int32_t, 2> pair = impl::get_cell_vertex_pairs<1>(
```
jorgensd, line 803: Is there a reason for `get_cell_vertex_pairs` not having the same kind of call signature as `get_cell_facet_pairs`, i.e. that `v_to_c` is passed into the function, and there `v_to_c->links(vertex)` is extracted?

.github/workflows/ci.yml
```diff
@@ -65,7 +65,7 @@ jobs:
           . $GITHUB_WORKSPACE/spack-src/share/spack/setup-env.sh
           spack info fenics-dolfinx
           spack env create cxx dolfinx-src/.github/workflows/spack-config/gh-actions-env-test.yml
-          spack -e cxx develop --path $GITHUB_WORKSPACE/dolfinx-src fenics-dolfinx@main
+          spack -e cxx develop --path $GITHUB_WORKSPACE/dolfinx-src fenics-dolfinx@git.${{ env.CURRENT_BRANCH }}=main
```
garth-wells: This shouldn't be needed. The brach has already been cloned above. `fenics-dolfinx@main` just says 'treat this like `main`'.

schnellerhase: It looks like the ran version in the merge queue does not align with the state of the branch. Guessing it's the state of `main`. Do you have another idea what's causing this?

garth-wells: Try
```
spack -e cxx develop --no-clone --path $GITHUB_WORKSPACE/dolfinx-src fenics-dolfinx@main
```

Spack-based CI is experimental, so while it would be nice to get it working here don't let it block merging.

schnellerhase: Unfortunately, did not resolve it. Could we then change the status check to not-required to get this merged?

garth-wells: It hasn't been made a requirement, but it is a good chance to debug the CI. Maybe you need to handle all FEniCS dependencies in the same as DOLFINx is handled, i.e. use `actions/checkout` to get code from a fork.

garth-wells:
@schnellerhase see https://github.com/FEniCS/dolfinx/tree/garth/point_source_2.

.github/workflows/ci.yml
```diff
@@ -39,6 +39,28 @@ jobs:
...
         with:
           path: dolfinx-src
+
+      - name: Load dev branch environment variables
+        run: cat dolfinx-src/.github/workflows/fenicsx-refs.env >> $GITHUB_ENV
+
+      - name: Get UFL code
+        uses: actions/checkout@v5
+        with:
+          repository: fenics/ufl
```
garth-wells: Repository value is available in `fenicsx-refs.env`.

.github/workflows/ci.yml
```diff
@@ -65,12 +84,12 @@ jobs:
...
           spack info fenics-dolfinx
           spack env create cxx dolfinx-src/.github/workflows/spack-config/gh-actions-env-test.yml
-          spack -e cxx develop --path $GITHUB_WORKSPACE/dolfinx-src fenics-dolfinx@main
-          spack -e cxx develop fenics-basix@git.${{ env.basix_ref }}=main
-          spack -e cxx develop fenics-ufcx@git.${{ env.ffcx_ref }}=main
-          spack -e cxx develop py-fenics-basix@git.${{ env.basix_ref }}=main
-          spack -e cxx develop py-fenics-ffcx@git.${{ env.ffcx_ref }}=main
-          spack -e cxx develop py-fenics-ufl@git.${{ env.ufl_ref }}=main
+          spack -e cxx develop --no-clone --path $GITHUB_WORKSPACE/dolfinx-src fenics-dolfinx@main
+          spack -e cxx develop --no-clone --path $GITHUB_WORKSPACE/basix-src fenics-basix@main
```
garth-wells: py-basix is needed here too because `py-ffcx` depends on it. See my branch.

garth-wells (dismissed):

.github/workflows/ci.yml
schnellerhase: Installs are aligned now with branch.

Issue #3729, opened 2025-05-13
Support for arbitrary boundary vertices for topology computation


schnellerhase:
### Describe new/missing feature

The problem described in https://github.com/FEniCS/dolfinx/issues/2374 may be tackled by allowing the `create_mesh` function to accept a custom `boundary_v_fn` callback.  For arbitrary manifold meshes this should allow partitioning by just setting all vertices as boundary vertices. 

### Suggested user interface

```python3

```

jorgensd:
I like this idea!

Issue #2374, opened 2022-09-24
Generalised topology construction (#2306) yields invalid partitioning for simple mesh


anzil:
For a simple mesh (4 nodes, 3 cells, gdim=3, tdim=1) the partitioning using #2306 results in

- an invalid global size of the index map for selected topological dimensions,
- incorrect global indices in the index map for selected topological dimensions

as demonstrated in the minimal example below.

```python
import dolfinx
import ufl
import numpy as np
from mpi4py import MPI

comm = MPI.COMM_WORLD

#         (3)
#          |
#          |[2]
#          |
# (1)-----(0)-----(2)
#     [0]     [1]

if comm.rank == 0:
	coord = np.array([[0.0, 0.0, 0.0], [-1.0, 0.0, 0.0], [+1.0, 0.0, 0.0], [0.0, +1.0, 0.0]])
	cells = np.array([[0, 1], [0, 2], [0, 3]])
else:
	coord = np.ndarray((0, 3))
	cells = np.ndarray((0, 2))

ufl_cell = ufl.Cell("interval", geometric_dimension=3)
ufl_elem = ufl.FiniteElement("Lagrange", ufl_cell, 1)
ufl_mesh = ufl.Mesh(ufl.VectorElement(ufl_elem))

mesh = dolfinx.mesh.create_mesh(comm, cells, coord, ufl_mesh)

for tdim in range(mesh.topology.dim + 1):
	imap = mesh.topology.index_map(tdim)
	lidx = np.arange(imap.size_local, dtype=np.int32)
	size = imap.size_global
	gidx = imap.local_to_global(lidx)
	print(f"(tdim={tdim}, rank={comm.rank}): size_global = {size}, global_indices = {gidx}")
```

Before the changes in #2306:
```shell
dolfinx@136aa0acdeeb7e29b336949ad9ed8537237f5f88# mpirun -n 2 python3 mfe.py 
(tdim=0, rank=0): size_global = 4, global_indices = [0]
(tdim=1, rank=0): size_global = 3, global_indices = [0]
(tdim=0, rank=1): size_global = 4, global_indices = [1 2 3]
(tdim=1, rank=1): size_global = 3, global_indices = [1 2]
```
After the changes in #2306:
```
dolfinx@dfbcb44ad7e109a3367f8e7c0f973487f58eafbe# mpirun -n 2 python3 mfe.py 
(tdim=0, rank=0): size_global = 5, global_indices = [0 1]
[...]

PR #3730, opened 2025-05-13
Local dual graph for branching meshes

schnellerhase:
The dual graph computation previously assumed that at most two cells could be connected through a facet. This is not true for branching/non-orientable meshes. This extends the local dual graph computation to allow for such meshes. It implies no overhead for meshes without branches/joints. 

Therefore we are now able to deal with non-orientable meshes correctly where the branching does not happen over a process boundary.

This fixes https://github.com/FEniCS/dolfinx/issues/2374. Since the input is provided on one process. 

The problem still persists if the joint lies on a process boundary facet (Global dual graph computation fails to detect joint edge). This behavior is now documented and its fixing is tracked in https://github.com/FEniCS/dolfinx/issues/3733.

Fixes https://github.com/FEniCS/dolfinx/issues/3729 (Solution strategy in there turned out to be not the right approach - needs to be fixed on dual graph level, boundary vertices need to be computed based on a correct dual graph, this can not be corrected for on vertex level).

Additionally
- adds documentation to `build_local_dual_graph`.
- moves `boundary_vertices_fn` to own function which could be moved to `cpp` at some point. 

The important changes are in https://github.com/FEniCS/dolfinx/pull/3730/files#diff-c24d3c0de0058214385657cf17a71f410f74a4e709bde7d858de41b5b94e52d7R510-R521. It changes from introducing edges `cell_0 -> cell_i` to adding all combinations `cell_a -> cell_b` for `a<b`


schnellerhase:
Output for mfe from https://github.com/FEniCS/dolfinx/issues/2374

```bash
(tdim=0, rank=0): size_global = 4, global_indices = []
(tdim=1, rank=0): size_global = 3, global_indices = []
(tdim=0, rank=1): size_global = 4, global_indices = [0 1 2 3]
(tdim=1, rank=1): size_global = 3, global_indices = [0 1 2]
```

jorgensd (approved):

michalhabera:
Any more comments @chrisrichardson ? This is a blocker for us, will merge tomorrow if no more comments.

PR #3731, opened 2025-05-14
XDMF IO for meshes using serendipity degree 2

mscroggs:
- Added permutation to vtk mesh ordering
- Made coordinate element use serendipity variant with point evaluations on edges (defined as a custom element)


jorgensd:
This was initially added in: https://github.com/FEniCS/dolfinx/pull/3662 but then later removed in https://github.com/FEniCS/dolfinx/pull/3665

mscroggs:
> This was initially added in: #3662 but then later removed in #3665

This PR doesn't include S2 on quads, so we should add that back in

chrisrichardson:
Is this OK to merge?

PR #3737, opened 2025-05-20
Update scientificcomputing arm64 vtk wheel + bump pyvista and PETSc/SLEPc

schnellerhase:
https://github.com/finsberg/vtk-aarch64 has moved to https://github.com/scientificcomputing/vtk-aarch64 plus version bumps


jorgensd:
One should use: https://github.com/scientificcomputing/vtk-aarch64/releases/tag/v9.3.0-py3.12
which is compatible with Pyvista.

schnellerhase:
Thanks, indeed pyvista is not compatible with vtk 9.4.0 and 9.4.1, see https://github.com/pyvista/pyvista/blob/v0.45.2/pyproject.toml#L27-L29. However, 9.4.2 is supported - should we instead bump to this?

jorgensd (approved):

Issue #3735, opened 2025-05-19
[BUG]: Problem when assembling different integrals for different tags


pantolin:
### Summarize the issue

I encountered this problem that I suspect it is a bug when calling the different kernels for different cell tags.

### How to reproduce the bug

Just run the MWE below with dolfinx v0.9.0

### Minimal Example (Python)

```Python
from mpi4py import MPI

import numpy as np
import ufl
from dolfinx.cpp.mesh import CellType
from dolfinx.fem import assemble_scalar, form
from dolfinx.mesh import create_unit_square, meshtags

msh = create_unit_square(MPI.COMM_WORLD, 1, 1, CellType.triangle)


# Setting different quadrature degrees and values for a and b, makes it fail
cells = np.array([0, 1])
tags = np.array([0, 1])
cell_tags = meshtags(msh, msh.topology.dim, cells, tags)
dx = ufl.dx(domain=msh, subdomain_data=cell_tags)
dx0 = dx(subdomain_id=(0, 1), degree=2)
dx1 = dx(subdomain_id=(0), degree=3)

a, b = 1.0, 2.0
ufl_form_0 = a * dx0
ufl_form_1 = b * dx1
ufl_form = ufl_form_0 + ufl_form_1

expected = 1.0 * a + 0.5 * b

computed_0 = assemble_scalar(form(ufl_form_0)) + assemble_scalar(form(ufl_form_1))
print(f"Two assemblies -- Expected: {expected}. Computed: {computed_0}")

computed_1 = assemble_scalar(form(ufl_form))
print(f"One assembly   -- Expected: {expected}. Computed: {computed_1}")
```

### Output (Python)

```bash
Two assemblies -- Expected: 2.0. Computed: 2.0
One assembly   -- Expected: 2.0. Computed: 2.5
```

### Version

0.9.0

### Installation

conda on osx arm64

jorgensd:
I think the issue is this around these lines ( https://github.com/FEniCS/dolfinx/blame/f1c5f307c70f414872a090e349f1d88211f54b58/python/dolfinx/fem/forms.py#L369), as it seems like the order of the integrals here are different that the one in the generated code.

FFCx orders the `b` integral prior to the two `a` integrals, while the referenced code goes through it as
`a` (id 0), `b` (id 0), `a` (id 1)

jorgensd:
So I think we can simplify the Python code to:
```python
        # Extract subdomain ids from ufcx_form
[...]

PR #3740, opened 2025-05-23
Refactor `dolfinx::fem::Form` to use local indexing of forms, rather than integral ids

jorgensd:
As shown in: #3735, there are many UFL forms (especially in the case of mixed quadrature rules), where the `subdomain_id` is not mapped 1-1 with an integration kernel.

This PR refactors the Form class, to rather use the local indexing from the generated code (where integrals have been grouped) as the lookup key for the integral.

Only a minor change occurs in the user-interface, as `Form::subdomain_ids` no longer maps to the UFL subdomain_ids.
However, this is currently only used for internal looping, as the integral_idx -> subdomain_id mapping happens within the `create_form_factory` function.

This allows for different quadrature degrees in different parts of a form.

Example:
```python
import basix.ufl
import ufl

cell = basix.CellType.triangle
coord_element = basix.ufl.element("Lagrange", cell, 1, shape=(2,))
domain = ufl.Mesh(coord_element)

degree = 2
el = basix.ufl.element("Lagrange", cell, degree)
V = ufl.FunctionSpace(domain, el)
u = ufl.TrialFunction(V)
v = ufl.TestFunction(V)

dx = ufl.Measure("dx", domain=domain)

a_mass = ufl.inner(u, v)*dx((1, 2), degree=2*degree)
a_stiffness = ufl.inner(ufl.grad(u), ufl.grad(v))*dx((1,), degree=2*(degree-1))

a = a_mass + a_stiffness
```
This generates two different integration kernels in FFCx, with the  integral information as:
```C
uint64_t finite_element_hashes_form_e26395feca75c800a7470ee9a844c8cdc67aff23[2] = {UINT64_C(16933917890882727401), UINT64_C(16933917890882727401)};
int form_integral_offsets_form_e26395feca75c800a7470ee9a844c8cdc67aff23[4] = {0, 3, 3, 3};
static ufcx_integral* form_integrals_form_e26395feca75c800a7470ee9a844c8cdc67aff23[3] = {&integral_0dcf3cd8d98806a1645751b277f55c83190bebb9_triangle, &integral_be233757c48f146683317dc29c6fe8fa75e720a8_triangle, &integral_be233757c48f146683317dc29c6fe8fa75e720a8_triangle};
int form_integral_ids_form_e26395feca75c800a7470ee9a844c8cdc67aff23[3] = {1, 1, 2};
```
As seen in the latter line here, `form_integral_ids_form_*` is not a list of unique integers.
In the main branch of DOLFINx, we assume that this id is unique, which is the root cause of #3735.


garth-wells:
Can you add some gentle background to the PR description?

jorgensd:
> Can you add some gentle background to the PR description?

I've added an example to the description of the PR.

garth-wells:
Yes, will add comments asap.

chrisrichardson (approved):

cpp/dolfinx/fem/Form.h
```diff
@@ -188,9 +188,8 @@ class Form
   /// @param[in] V Function spaces for the form arguments, e.g. test and
   /// trial function spaces.
   /// @param[in] integrals Integrals in the form, where
-  /// `integrals[IntegralType, domain ID, kernel index]` returns the
-  /// integral (`integral_data`) of type `IntegralType` over domain `ID`
-  /// with kernel index `kernel index`.
+  /// `integrals[IntegralType, i, kernel index]` returns the `i`th integral
```
garth-wells, line 132: It's not clear what the "`i`th integral" means.

cpp/dolfinx/fem/Form.h
```diff
@@ -432,31 +431,22 @@ class Form
...
-  /// The IDs correspond to the domain IDs which the integrals are
-  /// defined for in the form. `ID=-1` is the default integral over the
-  /// whole domain.
+  /// @brief Get number of integrals (kernels) for a given integral type and
+  /// kernel index.
   ///
   /// @param[in] type Integral type.
-  /// @return List of IDs for given integral type.
-  std::vector<int> integral_ids(IntegralType type) const
+  /// @param[in] kernel_idx Index of the kernel (we may have multiple
```
garth-wells, line 437: Would be helpful to add a more in-depth discussion/explanation to to the body of the docs.

cpp/dolfinx/fem/Form.h
```diff
@@ -432,31 +431,22 @@ class Form
...
+  /// kernels for a integral type in mixed-topology meshes).
+  int num_integrals(IntegralType type, int kernel_idx) const
   {
-    std::vector<int> ids;
+    int count = 0;
```
garth-wells: Could this be replaced by `std::count_if`?

cpp/dolfinx/fem/Form.h
```diff
@@ -512,21 +502,21 @@ class Form
   /// `entities0[i]` point to the same mesh entity, but with respect to
   /// different mesh views.
   ///
-  /// @param type Integral type.
-  /// @param rank Argument index, e.g. `0` for the test function space, `1`
+  /// @param[in] type Integral type.
```
garth-wells, line 514: Doc is just a repeat of the type. Does it need more explanation?

garth-wells (changes requested):

cpp/dolfinx/fem/Form.h
jorgensd: A form is split into `N` integrals by UFL. Each of these integrals can have a sequence of `subdomain_ids` (non-unique).

For instance, a form could be `integral_a((3, 4)) + integral_b((4, 1))`.
which would be ordered as
```
integrals = (integral_b, integral_a, integral_b, integral_a)
ids = (1, 3, 4, 4)
```
internally in the generated code of FFCx. It does not have any other meaning than a list entry of the generated kernels.

jorgensd:
@garth-wells I've addressed your comments. Good to merge?

michalhabera:
This is more of a critical bug-fix, so should go in asap. Btw. is the local index now needed? Looks like it will always be 0, 1, 2, ..., num_integrals

PR #3734, opened 2025-05-16
Specify Python binding language in CMakeLists.txt

garth-wells:

PR #3742, opened 2025-06-03
Fix python compilation against adios2 2.10.2

jmv2009:
Refer to:

https://github.com/TESSEorg/ttg/issues/86
https://github.com/FEniCS/dolfinx/pull/3734

to build on current debian testing (trixie), which is on adios2 2.10.2


francesco-ballarin:
See also https://github.com/FEniCS/dolfinx/pull/3734#issuecomment-2906449973

PR #3744, opened 2025-06-10
Restrict numpy version for oneAPI

jpdean:


garth-wells:
This is a conda bug - the conda should pin the NumPy version. Seems there have been some package updates in the past day - might be working again without any changes on our side.

PR #3746, opened 2025-06-11
Remove dynamic memory allocations inside lowest level assembly functions

garth-wells:
This will allow the lowest lever assembly functions to be executed efficiently for small batches of cells.

Also makes a number of simplifications following careful review of code.


cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -29,6 +29,9 @@ using mdspan2_t = md::mdspan<const std::int32_t, md::dextents<std::size_t, 2>>;
 
 /// @brief Execute kernel over cells and accumulate result in a matrix.
 ///
+/// @note This function does no dynamic memory allocation. Developers
+/// must make any changes that introduce a dynamic memory allocation.
```
jorgensd:
```suggestion
/// must not make any changes that introduce a dynamic memory allocation.
```

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -160,6 +168,9 @@ void assemble_cells(
 /// @brief Execute kernel over exterior facets and accumulate result in
 /// a matrix.
 ///
+/// @note This function does no dynamic memory allocation. Developers
+/// must make any changes that introduce a dynamic memory allocation.
```
jorgensd:
```suggestion
/// must not make any changes that introduce a dynamic memory allocation.
```

michalhabera:
It would be also neat to check no heap allocation in certain scopes. Clang (>=19) has `[[clang::nonallocating]]` attribute, https://clang.llvm.org/docs/FunctionEffectAnalysis.html#the-nonblocking-and-nonallocating-attributes
Requires extra `-Wfunction-effects` pass.

schnellerhase:
That flag could be passed to the `clang-tidy` linting run in https://github.com/FEniCS/dolfinx/pull/3722. Clang tidy hooks into a  `LLVM` based compile anyways.

PR #3751, opened 2025-06-16
Simplify input to non-linear problem

### LINKED ISSUE #3749: Port Cahn-Hilliard demo to NonlinearProblem class

Cahn-Hilliard is one of our main non-linear demos and should be ported to the new NonlinearProblem class.

### Suggested user interface

```python3

```

jorgensd:
The matrix kind should determine what the vector kind should be.
The naming of inputs are now consistent with `dolfinx.fem.petsc.LinearProblem` (https://github.com/FEniCS/dolfinx/pull/3684)

Use interface to solve Cahn-Hilliard problem (resolves #3749)


python/demo/demo_cahn-hilliard.py
```diff
@@ -266,40 +265,31 @@
...
-ksp = solver.krylov_solver
-opts = PETSc.Options()  # type: ignore
-option_prefix = ksp.getOptionsPrefix()
-opts[f"{option_prefix}ksp_type"] = "preonly"
-opts[f"{option_prefix}pc_type"] = "lu"
-sys = PETSc.Sys()  # type: ignore
-# For factorisation prefer MUMPS, then superlu_dist, then default
+# For the factorisation of the underlying linearized problems, prefer MUMPS,
+# then superlu_dist, then default
+linear_solver = "petsc"
```
jhale: Clearer to fall back to "petsc" in else below?

python/demo/demo_cahn-hilliard.py
```diff
@@ -266,40 +265,31 @@
...
+    "snes_linesearch_type": "none",
+    "snes_rtol": np.sqrt(np.finfo(default_real_type).eps) * 1e-2,
+    "ksp_type": "preonly",
+    "pc_type": "lu",
+    "pc_factor_mat_solver_type": linear_solver,
+}
+problem = NonlinearProblem(F, u, petsc_options=petsc_options)
 # -
 
-# The setting of `convergence_criterion` to `"incremental"` specifies
```
jhale, line 297: How does this convergence criterion transfer to the SNES?

jhale:
Looks good - two small comments then it can be merged.

jhale (approved):

python/demo/demo_cahn-hilliard.py
jorgensd: https://web.cels.anl.gov/projects/petsc/vault/petsc-3.21/docs/manual/snes.html#convergence-tests
- `stol`: is the convergence criteria for the increment (`successive tol`)
- `atol`, `rtol` is the absolute and relative tolerance of the norm of the `setFunction` output.

jhale: Isn't the old version better mapped to `stol` then?

jorgensd: I've added in options and docs to illustrate this. 
If we get converged reason == 4, then it is due to the increments being too small.

PR #3752, opened 2025-06-16
Illustrate how `dolfinx.fem.petsc.LinearProblem` can be used with fieldsplit

jorgensd:


python/demo/demo_mixed-poisson.py
```diff
@@ -165,33 +164,23 @@
...
-# and `L` correspond to the $\tau$ and $v$ test functions, respectively.
-# The columns of `a` correspond to the $\sigma$ and $u$ trial functions,
-# respectively. Note that `a[1][1]` is empty, which is denoted by
-# `None`. This zero block is typical of a saddle-point problem. In the
-# `L[0]` block, the test function $\tau$ is multiplied by a zero
-# `Constant`, which is evaluated at runtime. We do this to preserve
-# knowledge of the test space in the block. *Note that the defined `L`
+# We now declare the blocked bilinear and linear forms. We use `ufl.extract_blocks`
+# to extract the block structure of the bi-linear and linear form.
+# For the first block of the right-hand-side, we provide a form that efficiently is 0.
```
jhale: Remove second hyphen right-hand-side

python/demo/demo_mixed-poisson.py
```diff
@@ -165,33 +164,23 @@
...
-# We now declare the blocked bilinear and linear forms. The rows of `a`
-# and `L` correspond to the $\tau$ and $v$ test functions, respectively.
-# The columns of `a` correspond to the $\sigma$ and $u$ trial functions,
-# respectively. Note that `a[1][1]` is empty, which is denoted by
-# `None`. This zero block is typical of a saddle-point problem. In the
-# `L[0]` block, the test function $\tau$ is multiplied by a zero
-# `Constant`, which is evaluated at runtime. We do this to preserve
-# knowledge of the test space in the block. *Note that the defined `L`
+# We now declare the blocked bilinear and linear forms. We use `ufl.extract_blocks`
+# to extract the block structure of the bi-linear and linear form.
```
jhale: Remove hyphen bi-linear

python/demo/demo_mixed-poisson.py
```diff
@@ -258,38 +222,50 @@
...
+# +
+
+problem = fem.petsc.LinearProblem(
+    a,
+    L,
+    u=[sigma, u],
+    P=a_p,
+    kind="nest",
+    bcs=bcs,
+    petsc_options={
```
jhale, line 255: Is it potentially confusing to use both programmatic and dictionary based options in the same solver?

jorgensd: I don’t really think so. I think it shows the flexibility. These options are more verbose to add directly to PETSC ksp than they are through the dictionary. 

Additionally, some things are simply not options, as they pass along data and can’t be passed through an option interface.

python/demo/demo_mixed-poisson.py
jorgensd:
```suggestion
# to extract the block structure of the bilinear and linear form.
```

python/demo/demo_mixed-poisson.py
jorgensd:
```suggestion
# For the first block of the right-hand side, we provide a form that efficiently is 0.
```

python/demo/demo_mixed-poisson.py
```diff
@@ -258,38 +222,50 @@
...
+
+
 # +
-ksp.getPC().setType("fieldsplit")
-ksp.getPC().setFieldSplitType(PETSc.PC.CompositeType.ADDITIVE)
-nested_IS = P.getNestISs()
+# We get the nest index sets for the two fields
+nested_IS = problem.A.getNestISs()
+ksp = problem.solver
+ksp.setMonitor(lambda ksp, its, rnorm: print(f"Iteration: {its}, residual: {rnorm}"))
```
nate-sime: Consider formatting, flushing and outputting only on the root MPI rank.
```suggestion
ksp.setMonitor(lambda ksp, its, rnorm: PETSc.Sys.Print(f"Iteration: {its:>4d}, residual: {rnorm:.3e}"))
```

If you don't want to use PETSc then something like
```suggestion
def monitor(ksp: PETSc.KSP, its: int, rnorm: float):
    if msh.comm.rank == 0:
        print(f"Iteration: {its:>4d}, residual: {rnorm:.3e}", flush=True)
ksp.setMonitor(monitor)
```

jorgensd: I don’t mind using PETSc for this. Will adapt it

python/demo/demo_mixed-poisson.py
```diff
@@ -258,40 +223,57 @@
...
+
+
 # +
-ksp.getPC().setType("fieldsplit")
-ksp.getPC().setFieldSplitType(PETSc.PC.CompositeType.ADDITIVE)
-nested_IS = P.getNestISs()
+# We set the nest index sets for the preconditioner.
+nested_IS = problem.A.getNestISs()
+ksp = problem.solver
 ksp.getPC().setFieldSplitIS(("sigma", nested_IS[0][0]), ("u", nested_IS[0][1]))
```
jhale: Should this information be set on the PC automatically? Are there any circumstances when you wouldn't want the PC to share the fieldsplit IS with the assembled preconditioning operator `a_p`?

jorgensd: I don't think so. Feel free to modify it.
I guess within the solver, the name should be taken from `u[i].name`?
I guess the only issue is that we default the name of all functions to `f`?
Maybe we should set a default name within the solver, i.e. `uh0`, `uh1` etc?

jhale:
The current implementation gives the user a fieldsplit name of e.g. `sigma_0` if the DOLFINx function is named sigma and is in position 0, otherwise just `0`.

PR #3756, opened 2025-06-20
Uniform refinement of mixed topology meshes

chrisrichardson:
Allows mixed topology meshes, e.g. triangle+quad, or hex+pyramid+prism+tet to be refined globally.

This also works for pure or quad hex meshes, for example.


cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,266 @@
...
+      e_index.push_back(std::distance(entity_types[3].begin(), it));
+  }
+
+  // Add up all local vertices, edges, quad facets and hex cells.
+  std::int64_t nlocal = 0;
+  for (std::size_t dim = 0; dim < e_index.size(); ++dim)
+  {
+    if (topology->index_maps(dim).empty())
+      throw std::runtime_error(
+          "Missing topology for dimension " + std::to_string(dim)
```
jorgensd:
```suggestion
          "Missing entities of dimension " + std::to_string(dim)
```

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+  spdlog::info("Topology dim = {}", tdim);
+
+  // Collect up entity types on each dimension
+  std::vector<std::vector<mesh::CellType>> entity_types;
+  for (int i = 0; i < tdim + 1; ++i)
+    entity_types.push_back(topology->entity_types(i));
+
+  // Get index maps for vertices and edges and create maps for new vertices
+  std::vector<std::vector<std::int64_t>> new_v;
```
jorgensd, line 35: Maybe add a note here that the vertices in the new grid will be:
1. the original vertices
2. All edges get a new vertex at its midpoint.
3. Quadrilaterals (cells or facets) gets a new vertex at its midpoint
4. Hexahedra adds a node at the cell midpoint

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,266 @@
...
+      throw std::runtime_error(
+          "Missing topology for dimension " + std::to_string(dim)
+          + ", need to call create_entities(" + std::to_string(dim) + ")");
+    index_maps.push_back(topology->index_maps(dim)[e_index[dim]]);
+    new_v.push_back(std::vector<std::int64_t>(
+        index_maps.back()->size_local() + index_maps.back()->num_ghosts()));
+    nlocal += index_maps.back()->size_local();
+  }
+
+  // std::int64_t nv_global = index_maps[0]->size_global();
```
jorgensd: Remove line?

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+  // Get current geometry and put into new array for vertices
+  std::vector<std::int32_t> vertex_to_x(index_maps[0]->size_local()
+                                        + index_maps[0]->num_ghosts());
+
+  // Iterate over cells
```
jorgensd:
```suggestion
  // Iterate over cells per topology type
```

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,266 @@
...
+  for (int j = 0; j < static_cast<int>(entity_types.back().size()); ++j)
+  {
+    // Get geometry for each cell type
+    auto x_dofmap = mesh.geometry().dofmap(j);
+    auto c_to_v = topology->connectivity({tdim, j}, {0, 0});
+    auto dof_layout = mesh.geometry().cmaps().at(j).create_dof_layout();
+    std::vector<int> entity_dofs(dof_layout.num_dofs());
+    for (int k = 0; k < dof_layout.num_dofs(); ++k)
+      entity_dofs[k] = dof_layout.entity_dofs(0, k)[0];
```
jorgensd:
```suggestion
      entity_dofs[k] = dof_layout.entity_dofs(0, k).front();
```

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,266 @@
...
+
+    // Iterate over cells of this type
+    for (std::int32_t c = 0;
+         c < topology->index_maps(tdim)[j]->size_local()
```
jorgensd: Maybe extract index map above to avoid double access?

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,266 @@
...
+  std::array<std::int64_t, 2> local_range = {nscan - nlocal, nscan};
+
+  std::int32_t w_off = 0;
+  for (int j = 0; j < static_cast<int>(index_maps.size()); ++j)
+  {
+    std::int32_t num_entities = index_maps[j]->size_local();
+    std::iota(new_v[j].begin(), std::next(new_v[j].begin(), num_entities),
+              local_range[0] + w_off);
+
+    if (j > 0)
```
jorgensd: Maybe add a note here that with the current supported cell types, there is only one type of entity type per topological dimension that gets a new vertex.

It makes it easier to parse the connectivity call below.

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,266 @@
...
+    {
+      auto e_to_v = topology->connectivity({j, e_index[j]}, {0, 0});
+      for (std::int32_t w = 0; w < num_entities; ++w)
+      {
+        auto vt = e_to_v->links(w);
+        std::size_t nv_ent = vt.size();
```
jorgensd: Should we assert this against `cell_num_entities` https://github.com/FEniCS/dolfinx/blob/13c9ce4a8d700b5be57cd4c8727dd0d0b0948c4f/cpp/dolfinx/mesh/cell_types.cpp#L139 ?

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,266 @@
...
+        std::array<T, 3> v = {0, 0, 0};
+        for (std::size_t i = 0; i < nv_ent; ++i)
+        {
+          for (int k = 0; k < 3; ++k)
+            v[k] += x_g[3 * vertex_to_x[vt[i]] + k];
+        }
+        for (int k = 0; k < 3; ++k)
```
jorgensd: This only works for linear geometries, right? Should we add the non-linear extension here right away?

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,264 @@
...
+          new_x[(w + w_off) * 3 + k] = v[k] / static_cast<T>(nv_ent);
+      }
+    }
+    common::Scatterer sc(*index_maps[j], 1);
+    sc.scatter_fwd(std::span<const std::int64_t>(new_v[j]),
+                   std::span(std::next(new_v[j].begin(), num_entities),
+                             index_maps[j]->num_ghosts()));
```
jorgensd, line 148: Why is the scatterer after the if-test? The if-test does not alter the `new_v` array.

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,264 @@
...
+  std::vector<int> refined_cell_list;
+  for (int k = 0; k < static_cast<int>(cell_entity_types.size()); ++k)
+  {
+    // Reserve an estimate of space for the topology of each type
+    mixed_topology[k].reserve(mesh.topology()->index_maps(tdim)[k]->size_local()
+                              * 8 * 6);
+
+    // Select correct subdivision for celltype
+    // Hex -> 8 hex, Prism -> 8 prism, Tet -> 8 tet, Pyr -> 5 pyr + 4 tet
+    if (cell_entity_types[k] == mesh::CellType::hexahedron)
```
jorgensd: Shouldn't this use  a switch?

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,264 @@
...
+    {
+      spdlog::info("Hex subdivision [{}]", k);
```
jorgensd: I guess this is rather debug info than general info?

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,264 @@
...
+      spdlog::info("Prism subdivision [{}]", k);
+      refined_cell_list
+          = {0,  6,  7,  8,  15, 16, 6,  1,  9,  15, 10, 17, 7,  9,  2,  16,
+             17, 11, 6,  9,  7,  15, 17, 16, 15, 17, 16, 12, 14, 13, 8,  15,
+             16, 3,  12, 13, 11, 17, 16, 5,  14, 13, 10, 15, 17, 4,  12, 14};
+    }
+    else if (cell_entity_types[k] == mesh::CellType::pyramid)
+    {
+      spdlog::info("Pyramid subdivision [{}]", k);
+      refined_cell_list = {0,  5,  6, 13, 7,  1,  8,  5, 13, 9,  3,  10, 8,
```
jorgensd, line 206: Could you explain what is in the list? I though it would be the local connectivity for the entires 
Above it i stated that a pyramid is subdivided in 5 pyramids and 4 tetrahedra. I would then expect the array to be (5*5 + 4*4) entries, while this list has only 25 entries. 
One can later observe that you have append the array: `pyr_to_tet_list` to mixed topology, which accounts for the missing tetrahedra.

This should be better documented.

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,264 @@
...
+
+    spdlog::debug("On {}, over {} cells", k,
+                  topology->index_maps(tdim)[k]->size_local());
+    for (int c = 0; c < topology->index_maps(tdim)[k]->size_local(); ++c)
+    {
+      std::vector<std::int64_t> entities;
+      for (std::int32_t i : c_to_v->links(c))
+        entities.push_back(new_v[0][i]);
+      for (std::int32_t i : c_to_e->links(c))
+        entities.push_back(new_v[1][i]);
```
jorgensd:
```suggestion
      # Cell topology defined through its globally numbered vertices
      std::vector<std::int64_t> entities;
      # Extract new global vertex number for existing vertices
      for (std::int32_t i : c_to_v->links(c))
        entities.push_back(new_v[0][i]);
      # Extract the index of each vertex inserted on an edge
      for (std::int32_t i : c_to_e->links(c))
        entities.push_back(new_v[1][i]);
```

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,264 @@
...
+      if (e_index.size() > 2)
```
jorgensd, line 240:
```suggestion
      # If a quadrilateral or hexahedra has been refined, extra vertices are inserted 
      # in the corresponding topological dimension.
      if (e_index.size() > 2)
```

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,264 @@
...
+      {
+        if (tdim == 3)
+        {
+          auto conn = topology->connectivity({3, k}, {2, e_index[2]});
+          if (conn)
```
jorgensd, line 246: Shouldn't this rather throw an error if `conn` does not exist, or am I missing something?

python/test/unit/refinement/test_uniform.py
```diff
@@ -0,0 +1,58 @@
...
+    mesh = create_unit_cube(MPI.COMM_WORLD, 3, 2, 1, cell_type=ctype)
+
+    ncells0 = mesh.topology.index_map(3).size_local
+    mesh.topology.create_entities(1)
+    mesh.topology.create_entities(2)
+
+    m2 = Mesh(dolfinx.cpp.refinement.uniform_refine(mesh._cpp_object), None)
+    ncells1 = m2.topology.index_map(3).size_local
+
+    assert mesh.comm.allreduce(ncells0) * 8 == mesh.comm.allreduce(ncells1)
```
jorgensd, line 20: Maybe add an assembly check to ensure that the vertices of each sub-cell is correctly ordered?
Should test also be parametrized for shared_facet/none ghost mode?

cpp/dolfinx/refinement/uniform.cpp
chrisrichardson: Let's leave for a future PR. Higher-order geometry will be more difficult because we need to create new edge nodes too.

cpp/dolfinx/refinement/uniform.cpp
chrisrichardson: I have reorganised

cpp/dolfinx/refinement/uniform.cpp
chrisrichardson: Remember C++ uses `//` for comments

cpp/dolfinx/refinement/uniform.cpp
chrisrichardson: There is no connectivity from tet cells to quadrilaterals.

python/dolfinx/mesh.py
```diff
@@ -189,6 +189,24 @@ def index_map(self, dim: int) -> _cpp.common.IndexMap:
...
+        Returns:
+            List of IndexMaps for the entities of dimension ``dim``.
+        """
+        if (imaps := self._cpp_object.index_maps(dim)) is not None:
+            return imaps
+        else:
+            raise RuntimeError(
+                f"Entities of dimension {dim} have not been computed."
+                f"Call `dolfinx.mesh.Topology.create_entities({dim}) first."
+            )
```
schnellerhase, line 208: Should this not be raised on the cpp side?

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,285 @@
...
+      if (cell_entity_types[k] == mesh::CellType::pyramid)
+      {
+        for (int i : pyr_to_tet_list)
+          mixed_topology[ktet].push_back(entities[i]);
+      }
+    }
+  }
+
+  spdlog::debug("Create partitioner");
+  auto partitioner = mesh::create_cell_partitioner(mesh::GhostMode::none);
```
schnellerhase: Partitioner as argument? (Consistent with `refine`)

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,285 @@
...
+  // Requires edges (and facets for some 3D meshes) to be built already
+  auto topology = mesh.topology();
+  int tdim = topology->dim();
+  if (tdim < 2)
+    throw std::runtime_error("Uniform refinement only for 2D and 3D meshes");
+
+  spdlog::info("Topology dim = {}", tdim);
+
+  // Collect up entity types on each dimension
+  std::vector<std::vector<mesh::CellType>> entity_types;
```
schnellerhase, line 27:
```suggestion
  std::vector<std::vector<mesh::CellType>> entity_types;
  entity_types.reserve(tdim + 1);
```

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,285 @@
...
+  }
+
+  // Add up all local vertices, edges, quad facets and hex cells.
+  std::int64_t nlocal = 0;
+  for (std::size_t dim = 0; dim < e_index.size(); ++dim)
+  {
+    if (topology->index_maps(dim).empty())
+      throw std::runtime_error(
+          "Missing entities of dimension " + std::to_string(dim)
+          + ", need to call create_entities(" + std::to_string(dim) + ")");
```
schnellerhase, line 63: I think this should be raised in `topology->index_maps()`, which is consistent with `topology->connectivity`. See also other comment on python layer equivalent.

cpp/dolfinx/refinement/uniform.cpp
```diff
@@ -0,0 +1,285 @@
...
+  std::vector<std::vector<mesh::CellType>> entity_types;
+  for (int i = 0; i < tdim + 1; ++i)
+    entity_types.push_back(topology->entity_types(i));
+  const std::vector<mesh::CellType>& cell_entity_types = entity_types.back();
+
+  // Get index maps for vertices and edges and create maps for new vertices
+  // New vertices are created for each old vertex, and also at the centre of
+  // each edge, each quadrilateral facet, and each hexahedral cell.
+  std::vector<std::vector<std::int64_t>> new_v;
+  std::vector<std::shared_ptr<const common::IndexMap>> index_maps;
```
schnellerhase, line 36:
```suggestion
  std::vector<std::vector<std::int64_t>> new_v;
  new_v.reserve(e_index.size());
  std::vector<std::shared_ptr<const common::IndexMap>> index_maps;
  index_maps.reserve(e_index.size());
```

schnellerhase (commented):
Should maybe the entry point for `uniform_refine` and the already present `refine` share a common API? `refine` already supports uniform refinements by marking of all edges, and in the future the refinement routines implemented here could be extended to non uniform cases.

Also the partitioner creation, especially for refinement aware `identity_partitioner` to come https://github.com/FEniCS/dolfinx/pull/3661, should be available for the uniform refinement. This would be easiest to accomplish if the entry point is shared one.

cpp/dolfinx/refinement/uniform.cpp
chrisrichardson: Sure, I had this in mind - will do.

cpp/dolfinx/refinement/uniform.cpp
chrisrichardson: I am mostly uninclined to `reserve` vectors which are known to be very small (say 8 or less) since it just adds code clutter, and does nothing really for performance.

cpp/dolfinx/refinement/uniform.cpp
chrisrichardson: Could do.

cpp/dolfinx/refinement/uniform.cpp
chrisrichardson: I'm not sure about this. We want to have the possibility of a mesh which has no edges or facets initialised, so how then should we check for this case?

schnellerhase: For the connectivity we have the same problem/interface: If a connectivity is not computed before accessing it through `connectivity(...)` will raise and it's the callers responsibility to guarantee it has been pre-computed, if requested. Here this would translate to `index_map` raising if none are found, which would be the case for the valid state of not initialised entities, and otherwise returning a (guaranteed) non-empty vector of maps. Or, am I missing something else here?

python/dolfinx/mesh.py
chrisrichardson: I really just copyed this from the code for `index_map` - not sure about where "errors" versus "checks" should take place for missing topology. It is legal to have a mesh with no edges or facets computed...

PR #3764, opened 2025-06-24
Use bitset in VTKHDF read mesh, removing MPI_Allgatherv call.

chrisrichardson:
Rather than communicating all mesh cell types globally using `MPI_Allgatherv` use a bitset to do a set union on the cell type and degree. Using 8 cell types, and degrees 1-8, this fits into a 64-bit bitset.


garth-wells:
I think a medium term fix would be that the mesh construction/factory function takes only the cell types on the calling process. Then this issue (largely) disappears.

Once the mesh topology is built, information on the cell polynomial degree can be communicated (at the same time that the geometry data is communicated.

jorgensd:
@chrisrichardson do we still want this?

garth-wells:
@chrisrichardson can this be brought up-to-date?

PR #3765, opened 2025-06-24
NonlinearProblem docstring improvements

jhale:
- Improves API consistency across LinearSolver and NonlinearSolver.
- Documentation improvements, particularly various non-obvious points.
- Applies unique options prefixes to all held PETSc objects.
- Explicitly points out that `x` (Vector solution used by KSP/SNES) and `u` (Function solution returned by solve).
- Added some new property accessors.
- `petsc_options` are now only applied to KSP or the SNES, not the linear algebra objects - this is made explicit in the docstring. The previous LinearSolver also applied the options to `A` and `b`, but it wasn't clear why these were not applied to `x` and the new `P` object. It's also not clear whether one would want the same options applied to multiple different linear algebra objects.

### DIFF
--- a/python/dolfinx/fem/petsc.py
+++ b/python/dolfinx/fem/petsc.py
@@ -1381,7 +1381,7 @@ class NonlinearProblem:
                 opts[k] = v
 
             self.solver.setFromOptions()
-            
+
             # Tidy up global options
             for k in petsc_options.keys():
                 del opts[k]


python/dolfinx/fem/petsc.py
```diff
@@ -1315,6 +1324,10 @@ def solve(self) -> tuple[PETSc.Vec, int, int]:  # type: ignore
         """Solve the problem and update the solution in the problem
         instance.
 
+        Note:
+            The user is responsible for asserting convergence e.g. `assert
+            converged_reason > 0`.
+
```
jorgensd: The user can also set `"snes_error_if_not_converged": True` in the same way one can do it for KSP as `"ksp_error_if_not_converged":True`.
Thus if the provide both these options to non-linear problem, it will terminate if the underlying ksp fails to converge, or if snes fails to converge.

python/dolfinx/fem/petsc.py
```diff
@@ -1208,6 +1207,16 @@ def __init__(
...
         non-linear solver.
 
+        By default, the underlying SNES solver uses PETSc's default
+        options. To use the robust combination of LU via MUMPS with
+        a basic linesearch, pass:
+
+            petsc_options = {"ksp_type": "preonly",
+                             "pc_type": "lu",
+                             "pc_factor_mat_solver_type": "mumps",
+                             "snes_linesearch_type": "basic",
```
michalhabera: "basic" aka. "none", is the full Newton step, so is not the most robust choice here. PETSc's default "bt" is surely more robust.

Issue #3738, opened 2025-05-20
LBYL vs. EAFP (isinstance vs. try/catch)


michalhabera:
Easier to Ask Forgiveness than Permission (EAFP) style recently added to some of the python interfaces, see e.g. https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/la/petsc.py#L130 leads to

1. large performance overhead, see example below and,
2. opaque error handling. If any of the deep nested try/except blocks fails then user sees the outer exception caught which looks like an error message, but is not an error. The logic in the code also depends on the type of exception caught thus relies on the code being called raising consistently.

The following benchmark
```python
import time
import platform
import datetime
import textwrap
import numpy as np

N = 1_000_000


def eafp():
    total = 0
    for k in range(N):
        try:
            raise RuntimeError
        except RuntimeError:
            try:
                raise RuntimeError
            except RuntimeError:
                total += k
    return total


def lbyl():
    total = 0
    for k in range(N):
        if isinstance(k, ()):
            pass
        else:
            if isinstance(k, ()):
                pass
            else:
                total += k
    return total


def run_benchmark(fn, repeats=3):
    times = []
    for _ in range(repeats):
        start = time.perf_counter()
        fn()
        times.append(time.perf_counter() - start)

    times = np.array(times)
    mean = times.mean()
    std = times.std()
    return mean, std  # use best-of-repeats to reduce noise


lbyl_time_mean, lbyl_time_std = run_benchmark(lbyl)
eafp_time_mean, eafp_time_std = run_benchmark(eafp)


benchmark_info = textwrap.dedent(f"""
Python {platform.python_version()} on {platform.python_implementation()}
Date: {datetime.datetime.now().isoformat(sep=" ", timespec="seconds")}

Iterations            : {N:,}

RESULTS
------------------------
EAFP (try/except)      : mean {eafp_time_mean:.4f} s, std {eafp_time_std:.4f} s
LBYL (isinstance)      : mean {lbyl_time_mean:.4f} s, std {lbyl_time_std:.4f} s
Speed-up (LBYL/EAFP)   : mean {eafp_time_mean / lbyl_time_mean:.1f} × faster
""")

print(benchmark_info)

```

executes for me in dolfin/dolfinx:nightly on M4 as
```
Python 3.12.3 on CPython
Date: 2025-05-20 13:51:58

Iterations            : 1,000,000

RESULTS
------------------------
[...]

PR #3768, opened 2025-06-25
Fix mypy errors for newly typed ufl

schnellerhase:
Depends on https://github.com/FEniCS/ufl/pull/383 and https://github.com/FEniCS/basix/pull/944

Basix mypy check in the dolfinx CI appeared to be more exhaustive then the one in the linting step here. Therefore, this adds a mypy check in the non matrix 'Build and test' run.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -144,6 +144,12 @@ jobs:
           pip install --check-build-dependencies --no-build-isolation --config-settings=cmake.build-type="Debug" 'python/[test]'
           python -c "from mpi4py import MPI; import dolfinx; assert not dolfinx.has_petsc4py"
 
+      - name: Run mypy
+        working-directory: python
+        run: |
+          pip install mypy
+          mypy dolfinx
+
       - name: Run demos (Python, serial)
         run: |
           pip install pytest-xdist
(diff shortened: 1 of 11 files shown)


schnellerhase:
Highlights another problem with the EAFP style https://github.com/FEniCS/dolfinx/issues/3738 - typing info can not be correctly deduced.

.github/workflows/fenicsx-refs.env
```diff
@@ -1,4 +1,4 @@
-basix_ref=main
-ufl_ref=main
+basix_ref=schnellerhase/mypy
+ufl_ref=schnellerhase/mypy
```
schnellerhase:
```suggestion
basix_ref=main
ufl_ref=main
```

.github/workflows/ccpp.yml
```diff
@@ -144,6 +144,12 @@ jobs:
           pip install --check-build-dependencies --no-build-isolation --config-settings=cmake.build-type="Debug" 'python/[test]'
           python -c "from mpi4py import MPI; import dolfinx; assert not dolfinx.has_petsc4py"
 
+      - name: Run mypy
+        working-directory: python
+        run: |
+          pip install mypy
+          mypy dolfinx
```
michalhabera, line 151: Why not `mypy -p dolfinx`?

schnellerhase, line 151:
```suggestion
      - name: Run mypy
        run: |
          pip install mypy
          mypy -p dolfinx
```

schnellerhase: Good catch, trying it resulted in quite a few additional mypy errors. Will file an issue to address at a later stage, so we can merge these fixes for now.

Issue #3653, opened 2025-02-27
[BUG]: GJK fails on simple convex hull


jorgensd:
### Summarize the issue

Running the following minimal example makes GJK not converge.
First reported at:
https://fenicsproject.discourse.group/t/gjk-error-in-interpolation-between-non-matching-second-ordered-3d-meshes/16086/8?u=dokken

### How to reproduce the bug

See below

### Minimal Example (Python)

```Python
import numpy as np


p0 = np.array([[44.2919, 39.7508, 18.8208]])
p1 = np.array([[46.0659, 38.6251, 16.3637], 
                 [44.2764, 42.3867, 18.239],
                 [  50.0397, 38.4649, 19.9843], 
                  [ 47.1963, 40.4563, 19.1057] ,
                  [ 48.0266, 38.5593, 18.2475] ,
                   [45.1735, 40.495, 17.3336] ,
                   [48.0415, 40.4456, 18.1963 ]])
print(p1.shape)
import dolfinx
from dolfinx.geometry import compute_distance_gjk

compute_distance_gjk(p0, p1)
```

### Output (Python)

```bash
Traceback (most recent call last):
  File "/home/dokken/Documents/src/debug/mwe.py", line 16, in <module>
    compute_distance_gjk(p0, p1)
    ~~~~~~~~~~~~~~~~~~~~^^^^^^^^
  File "/home/dokken/src/mambaforge/envs/test_mpl/lib/python3.13/site-packages/dolfinx/geometry.py", line 272, in compute_distance_gjk
    return _cpp.geometry.compute_distance_gjk(p, q)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
RuntimeError: GJK error - max iteration limit reached
```

### Version

main branch

### Installation

Tested both docker and conda, stable and nightly

jorgensd:
Further reduced to:
```python
import numpy as np


# p0 = np.array([[44.2919, 39.7508, 18.8208]])
p0 = np.array([[44.2919, 39.7508, 18.8208]])
p1 = np.array(
    [
        [46.0659, 38.6251, 16.3637],
        [44.2764, 42.3867, 18.239],
        [50.0397, 38.4649, 19.9843],
        [45.1735, 40.495, 17.3336],
        [48.0266, 38.5593, 18.2475],
    ]
)
import matplotlib.pyplot as plt
import matplotlib

fig = plt.figure(figsize=(8, 8))
[...]

PR #3770, opened 2025-06-27
Use boost multiprecision for GJK

chrisrichardson:


jorgensd:
Have you tested this on: https://github.com/FEniCS/dolfinx/issues/3653

In general, it seems like you have rewritten the algorithm quite a bit. Could you comment on why and how it improves the situation (of course higher precision helps).

chrisrichardson:
Yes it works, the answer is:
`[-1.03648825 -1.01316152  1.08986529]`

I tried to make the algorithm clearer - there are some comments throughout. It is possible there is more simplification that can be done.

jhale (commented):
Looks nice. Multiprecision is header only and we already require boost so that's good.

Do we want the caller to be able to control the internal precision?

chrisrichardson:
It would be easy to add a template parameter for the internal precision, good for c++, but probably not so much for Python. We could use presets for double and float, I guess.

The functions are already in `impl_gjk` namespace. Some are duplicated from dolfinx/math.h (e.g. det3) because fma is used there which is not compatible with multiprecision libraries.

jhale (approved):

chrisrichardson:
On oneapi seems to be dispatching to float32 incorrectly...

cpp/dolfinx/geometry/gjk.h
```diff
@@ -22,208 +23,249 @@ namespace dolfinx::geometry
...
 {
 
-/// @brief Find the resulting sub-simplex of the input simplex which is
-/// nearest to the origin. Also, return the shortest vector from the
-/// origin to the resulting simplex.
-template <std::floating_point T>
-std::pair<std::vector<T>, std::array<T, 3>>
-nearest_simplex(std::span<const T> s)
+template <typename T>
+T det3(std::span<const T, 9> A)
```
garth-wells: Should this be in `math.h`?

cpp/dolfinx/geometry/gjk.h
```diff
@@ -22,208 +23,249 @@ namespace dolfinx::geometry
...
+{
+  T w0 = A[3 + 1] * A[2 * 3 + 2] - A[3 + 2] * A[3 * 2 + 1];
+  T w1 = A[3] * A[3 * 2 + 2] - A[3 + 2] * A[3 * 2];
+  T w2 = A[3] * A[3 * 2 + 1] - A[3 + 1] * A[3 * 2];
+  return A[0] * w0 - A[1] * w1 + A[2] * w2;
+}
+
+template <typename T>
```
garth-wells: Document.

cpp/dolfinx/geometry/gjk.h
```diff
@@ -22,208 +23,249 @@ namespace dolfinx::geometry
...
+T dot3(std::span<const T, 3> a, std::span<const T, 3> b)
+{
+  return a[0] * b[0] + a[1] * b[1] + a[2] * b[2];
+}
+
+/// @brief Find the barycentric coordinates in the simplex s, of the point in s
+/// which is closest to the origin.
+/// @param s Simplex described by a set of points in 3D, flattened.
```
garth-wells: 'flattened' isn't enough. Describe layout fully.

cpp/dolfinx/geometry/gjk.h
```diff
@@ -22,208 +23,249 @@ namespace dolfinx::geometry
...
+/// @return Barycentric coordinates of the point in s closest to the origin.
```
garth-wells, line 50: Layout?

cpp/dolfinx/geometry/gjk.h
```diff
@@ -22,208 +23,249 @@ namespace dolfinx::geometry
...
+/// @note s may be an interval (2 points), a triangle (3) or a tetrahedron (4).
```
garth-wells: (3)?, (4)? Make docs complete.

Format `s` properly.

cpp/dolfinx/geometry/gjk.h
```diff
@@ -22,208 +23,249 @@ namespace dolfinx::geometry
...
+template <typename T>
+std::vector<T> nearest_simplex(std::span<const T> s)
 {
   assert(s.size() % 3 == 0);
   const std::size_t s_rows = s.size() / 3;
+
+  spdlog::debug("GJK: nearest_simplex({})", s_rows);
```
garth-wells: Is this a very hot function? If yes, it shouldn't include logging.

cpp/dolfinx/geometry/gjk.h
```diff
@@ -22,208 +23,249 @@ namespace dolfinx::geometry
...
+
+    T lm = dot3(s0, s0) - dot3(s0, s1);
+    if (lm < 0.0)
     {
-      // The origin is between A and B
-      // v = s0 + lm * (s1 - s0);
-      std::array v
-          = {s0[0] + lm * ds[0], s0[1] + lm * ds[1], s0[2] + lm * ds[2]};
-      return {std::vector<T>(s.begin(), s.end()), v};
+      spdlog::debug("GJK: line point A");
```
garth-wells: Is function too hot for logging without preprocessor guard?

cpp/dolfinx/geometry/gjk.h
```diff
@@ -22,208 +23,249 @@ namespace dolfinx::geometry
...
+      return {1.0, 0.0};
+    }
+    T mu = dot3(s1, s1) - dot3(s1, s0);
+    if (mu < 0.0)
+    {
+      spdlog::debug("GJK: line point B");
```
garth-wells: See debug comment.

cpp/dolfinx/geometry/gjk.h
```diff
@@ -22,208 +23,249 @@ namespace dolfinx::geometry
...
-    std::size_t pos = 0;
+    spdlog::debug("GJK: triangle ABC");
+    T f1 = 1.0 / (va + vb + vc);
+    return {va * f1, vb * f1, vc * f1};
+  }
+  case 4:
+  {
+    // Most complex case, where simplex is a tetrahedron, with 15 possible
+    // outcomes (4 vertices, 6 edges, 4 facets and the interior).
+    std::vector<T> rv = {0.0, 0.0, 0.0, 0.0};
```
garth-wells, line 160:
```suggestion
    std::vector<T> rv = {0, 0, 0, 0};
```

garth-wells, line 160: Should this a `std::array`?

cpp/dolfinx/geometry/gjk.h
```diff
@@ -242,40 +284,50 @@ std::array<T, 3> support(std::span<const T> bd, std::array<T, 3> v)
 }
 } // namespace impl_gjk
 
-/// @brief Compute the distance between two convex bodies p and q, each
+/// @brief Compute the distance between two convex bodies p0 and q0, each
```
garth-wells:
```suggestion
/// @brief Compute the distance between two convex bodies `p0` and `q0`, each
```

cpp/dolfinx/geometry/gjk.h
```diff
@@ -242,40 +284,50 @@ std::array<T, 3> support(std::span<const T> bd, std::array<T, 3> v)
...
 /// defined by a set of points.
 ///
 /// Uses the Gilbert–Johnson–Keerthi (GJK) distance algorithm.
 ///
-/// @param[in] p Body 1 list of points, shape (num_points, 3). Row-major
+/// @param[in] p0 Body 1 list of points, shape (num_points, 3). Row-major
```
garth-wells:
```suggestion
/// @param[in] p0 Body 1 list of points, `shape=(num_points, 3)`. Row-major
```

cpp/dolfinx/geometry/gjk.h
```diff
@@ -291,28 +343,50 @@ std::array<T, 3> compute_distance_gjk(std::span<const T> p,
...
+    spdlog::debug("GJK: vw={}/{}", static_cast<double>(vw),
+                  static_cast<double>(eps));
+
     // Add new vertex to simplex
     s.insert(s.end(), w.begin(), w.end());
 
     // Find nearest subset of simplex
-    auto [snew, vnew] = impl_gjk::nearest_simplex<T>(s);
+    std::vector<U> lmn = impl_gjk::nearest_simplex<U>(s);
+    v = {0.0, 0.0, 0.0};
```
garth-wells, line 368: This would be better as `std::fill`.

cpp/dolfinx/geometry/gjk.h
chrisrichardson: There is a version in math.h, but that uses fma.

cpp/dolfinx/geometry/gjk.h
chrisrichardson: No, it has variable size.

cpp/dolfinx/geometry/gjk.h
garth-wells: That can be easily handled in `math.h` with type checks.

cpp/dolfinx/geometry/gjk.h
chrisrichardson: I have changed to use `SPDLOG_DEBUG` instead of `spdlog::debug`. This way, the debug statement is not compiled when in Release mode.

jhale:
Probably the latter because I'd consider non-convergence 'normal' rather than exceptional behaviour. Unfortunately we don't have std::expected in C++20 (C++23).

PR #3775, opened 2025-07-02
Generalise interior facet assembler for interface coupling

jpdean:
Some problems require coupling between two domains that do not overlap but have a common interface. The interior facet assembler is *almost* suitable for this coupling, but currently assumes all arguments in the form are defined on  both sides of the interface. This PR generalises the interior facet assembler to handle this case.

Whilst it does add some complexity to the assembler, it should allow the user interface for mixed-domain problems to be significantly simplified in a later PR. The extra complexity may potentially impact the performance of interior facet integrals, but it should be possible to mitigate this with some templating etc.


jpdean:
One thing we could do is to let `create_submesh` return an `EntityMap`, rather than just the list representing the map. That would keep track of which topologies related by which maps and mean users can just put them in a list to pass to `form`.

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -700,3 +700,118 @@ def f(x):
...
+
+
+def test_interior_interface():
+    """
+    This is a test for assembling a form over an interface between two domains
+    that don't overlap. The test function is defined on one domain, and the trial
+    function is defined on the other.
+    """
+
+    def interface_int_entities(msh, interface_facets, marker):
```
jorgensd: This should probably use my more efficient implementation: 
```python
def compute_interface_data(
    cell_tags: dolfinx.mesh.MeshTags, facet_indices: npt.NDArray[np.int32]
) -> npt.NDArray[np.int32]:
    """
    Compute interior facet integrals that are consistently ordered according to the `cell_tags`,
    such that the data `(cell0, facet_idx0, cell1, facet_idx1)` is ordered such that
    `cell_tags[cell0]`<`cell_tags[cell1]`, i.e the cell with the lowest cell marker is considered the
    "+" restriction".

    Args:
        cell_tags: MeshTags that must contain an integer marker for all cells adjacent to the `facet_indices`
        facet_indices: List of facets (local index) that are on the interface.
    Returns:
        The integration data.
    """
    # Future compatibilty check
    integration_args: tuple[int] | tuple
    if Version("0.10.0") <= Version(dolfinx.__version__):
        integration_args = ()
    else:
        fdim = cell_tags.dim - 1
        integration_args = (fdim,)
    idata = dolfinx.cpp.fem.compute_integration_domains(
        dolfinx.fem.IntegralType.interior_facet,
        cell_tags.topology,
        facet_indices,
        *integration_args,
    )
    ordered_idata = idata.reshape(-1, 4).copy()
    switch = cell_tags.values[ordered_idata[:, 0]] > cell_tags.values[ordered_idata[:, 2]]
    if True in switch:
        ordered_idata[switch, :] = ordered_idata[switch][:, [2, 3, 0, 1]]
    return ordered_idata
```

garth-wells:
Close in favour of #3780.

Issue #3733, opened 2025-05-15
Global dual graph construction for non orientable meshes


schnellerhase:
Consider the following example of a mesh (black) and its dual graph (blue) - that would be produced by `build_dual_graph` in a sequential call.

![Image](https://github.com/user-attachments/assets/ee29afa8-e01f-49e5-8ae4-1b8e67c7c611)

If we now consider the mesh to split across two processes and consider the computation of `build_dual_graph` we hope to get, where the dashed blue lines indicate ghost edges that should exist on both processes:

![Image](https://github.com/user-attachments/assets/3960b90a-6ed8-4821-bd76-0be9c0e835a1)

Currently the `build_dual_graph` will not produce these ghost edges. `build_local_dual_graph` will not mark the necessary vertices as they are connected to more than one cell and thus considered 'matched'. 

To resolve this is general the `compute_nonlocal_dual_graph` would need to consider all vertices for possible ghost edges (we can not deduce from local data if there exists a joint across the process boundary).

PR #3782, opened 2025-07-05
Local dual graph computation support for custom matching criteria 

schnellerhase:
Computation of dual graphs for branching meshes requires to consider of facets which are connected to more than two facets for global dual graph computations - compare image in https://github.com/FEniCS/dolfinx/issues/3733.

This introduces an optional argument to `build_dual_graph` (and `build_local_dual_graph`) - `max_facet_to_cell_links` - which allows to control the matching condition of a facet.

A facet is considered *matched*, i.e. not considered for a possible edge in the dual graph across process boundary, if it connects at least `max_facet_to_cell_links ` cells. Otherwise *unmatched*. The default value is `max_facet_to_cell_links`$=2$, corresponding to the non-branching setting where a facet is matched - not on the boundary - if it is shared by two cells. For `std::nullopt` there is no bound, all facets are considered unmatched.

Since the identification to be unmatched implies heavy communication in the (to follow) global construction this fine control is desirable. Consider for example the (skeleton) mesh formed by the edges of a regular hexahedral mesh. We know a vertex will be connected to at most 6 edges. If these 6 connections are found locally we can rule out this vertex (facet) to have another ghost edge in the dual graph which we have not computed locally and it should therefore not be considered in the parallel communication step. This corresponds to passing `max_facet_to_cell_links=6`.

First step to complete https://github.com/FEniCS/dolfinx/issues/3733.


chrisrichardson:
I think this is OK, though it may incur huge costs for certain meshes, e.g. suppose a mesh has cells mostly connected by two facets, but a few connected by three facets. All the two-facet connections will be considered "unmatched" and will need to go to the non-local dual graph part (which I think is your next PR).

cpp/dolfinx/mesh/graphbuild.h
```diff
@@ -46,10 +52,15 @@ enum class CellType;
...
+///
+/// @note Facet (2) and cell (4) data will contain multiple entries for the same
+/// facet for branching meshes with `max_facet_to_cell_links>2` to account for
+/// all facet cell connectivies.
 std::tuple<graph::AdjacencyList<std::int32_t>, std::vector<std::int64_t>,
            std::size_t, std::vector<std::int32_t>>
 build_local_dual_graph(std::span<const CellType> celltypes,
-                       const std::vector<std::span<const std::int64_t>>& cells);
+                       const std::vector<std::span<const std::int64_t>>& cells,
+                       std::optional<std::int32_t> max_facet_to_cell_links = 2);
```
garth-wells, line 63: It's a bit weird having `std::optional<std::int32_t>` for something that has a default value of type `std::int32_t`. What happens if `std::nullopt` is passed?

schnellerhase: Passing `std::nullopt` corresponds to an unbounded matching condition (`max_facet_to_cell_links`=∞). Every facet is considered unmatched, independent of connectivity, in this case.

garth-wells: Could you update the docstrings to explain this?

PR #3785, opened 2025-07-06
Add mandatory options prefix to high-level solver classes and general tidy

francesco-ballarin:
Follow up to #3781 . Further motivations in the comments below.


python/dolfinx/fem/petsc.py
```diff
@@ -1050,26 +1038,6 @@ def u(self) -> typing.Union[_Function, Iterable[_Function]]:
 # -- High-level interface for SNES ---------------------------------------
 
 
-def _ghostUpdate(x: PETSc.Vec, insert_mode: PETSc.InsertMode, scatter_mode: PETSc.ScatterMode):  # type: ignore
```
francesco-ballarin, line 1053: Removing in favor of the one in `la.petsc`. This would have been the third duplicate!

python/dolfinx/la/petsc.py
```diff
@@ -34,10 +34,23 @@ def _ghost_update(x: PETSc.Vec, insert_mode: PETSc.InsertMode, scatter_mode: PET
     if x.getType() == PETSc.Vec.Type.NEST:  # type: ignore[attr-defined]
         for x_sub in x.getNestSubVecs():
             x_sub.ghostUpdate(addv=insert_mode, mode=scatter_mode)
+            x_sub.destroy()
```
francesco-ballarin, line 37: I think these subvectors need to be explicitly destroyed in PETSc if we do not want them to stay in memory until python garbage collection happens.

python/dolfinx/la/petsc.py
```diff
@@ -34,10 +34,23 @@ def _ghost_update(x: PETSc.Vec, insert_mode: PETSc.InsertMode, scatter_mode: PET
...
     else:
         x.ghostUpdate(addv=insert_mode, mode=scatter_mode)
 
 
+def _zero_vector(x: PETSc.Vec):  # type: ignore
+    """Helper function for zeroing out PETSc vectors"""
+    if x.getType() == PETSc.Vec.Type.NEST:
```
francesco-ballarin, line 44: When copying from the `_zero_vector` I deleted in `fem.petsc`, I switched from duck typing to `if`. I don't think it is safe to do duck typing here with a `except PETSc.Error:`, because the failure could come for reasons other than `x` not being a nest vector.

schnellerhase: Related to https://github.com/FEniCS/dolfinx/issues/3738.

francesco-ballarin:
Ready for review now, the former CI error was a silly mistake on my end.

python/dolfinx/fem/petsc.py
```diff
@@ -1217,9 +1185,9 @@ def __init__(
         J: typing.Optional[typing.Union[ufl.form.Form, Sequence[Sequence[ufl.form.Form]]]] = None,
         P: typing.Optional[typing.Union[ufl.form.Form, Sequence[Sequence[ufl.form.Form]]]] = None,
         kind: typing.Optional[typing.Union[str, typing.Iterable[typing.Iterable[str]]]] = None,
+        petsc_options: typing.Optional[dict] = None,
```
francesco-ballarin, line 1206: Move up for consistency with `LinearProblem`. Not that it will make much of a difference, since almost everybody will be using `petsc_options` as a kwarg.

python/dolfinx/fem/petsc.py
```diff
@@ -1372,7 +1340,7 @@ def __init__(
             )
             self.solver.getKSP().getPC().setFieldSplitIS(*fieldsplit_IS)
 
-    def solve(self) -> tuple[PETSc.Vec, int, int]:  # type: ignore
+    def solve(self) -> tuple[typing.Union[_Function, Iterable[_Function]], int, int]:  # type: ignore
```
francesco-ballarin: Changed the first return argument to be the `dolfinx.fem.Function`, rather than the solution `PETSc.Vec`. Again, I did this for consistency with `LinearProblem`.

python/dolfinx/fem/petsc.py
```diff
@@ -1388,16 +1356,17 @@ def solve(self) -> tuple[PETSc.Vec, int, int]:  # type: ignore
         """
 
         # Move current iterate into the work array.
-        assign(self._u, self.x)
+        assign(self.u, self.x)
 
         # Solve problem
         self.solver.solve(None, self.x)
+        dolfinx.la.petsc._ghost_update(self.x, PETSc.InsertMode.INSERT, PETSc.ScatterMode.FORWARD)
```
francesco-ballarin, line 1395: This ghost update was missing, but is present in `LinearProblem.solve`.

python/dolfinx/fem/petsc.py
jhale: It's not necessary except when using ksp_only - the update is done during the computation of the residual norm before exit. Nevertheless, we can leave it in, or, check if the SNES is set as ksp_only and do it?

https://petsc.org/release/manualpages/SNES/SNESKSPONLY/

jhale: Sorry, this was aimed at the comment on ghost update after solve.

python/dolfinx/la/petsc.py
jhale: Let's remove all of the petsc duck typing in one big PR.

python/test/unit/fem/test_petsc_nonlinear_assembler.py
```diff
@@ -500,9 +500,9 @@ def nested():
                 ["u", nested_IS[0][0]], ["p", nested_IS[1][1]]
             )
 
-            x, converged_reason, _ = problem.solve()
```
jhale, line 503: Prefer the return method for getting x.

python/dolfinx/fem/petsc.py
francesco-ballarin: @jhale 

> It's not necessary [...] - the update is done during the computation of the residual norm before exit

I thought so too, but then got test failures downstream due to the missing ghost update when computing the L^2 norm of the error on the function. I'll try to see if one of the existing tests can be adapted to show the same behavior: the few tests I've seen are all computing the R^n norm of the output vector, and hence won't catch the (possible) mistake.

python/test/unit/fem/test_petsc_solver_wrappers.py
```diff
@@ -18,9 +18,11 @@
 @pytest.mark.petsc4py
 class TestPETScSolverWrappers:
     @pytest.mark.parametrize(
-        "mode", [dolfinx.mesh.GhostMode.none, dolfinx.mesh.GhostMode.shared_facet]
+        "mode",
+        [dolfinx.mesh.GhostMode.none, dolfinx.mesh.GhostMode.shared_facet],
     )
-    def test_compare_solvers(self, mode):
+    @pytest.mark.parametrize("nonlinear_solver", ["snes high level", "legacy Newton"])
+    def test_compare_solvers(self, mode, nonlinear_solver):
```
francesco-ballarin: @jhale This is a test that confirms that the ghost update after SNES solve is really needed. If I run this on `main` I get a failure for the case `nonlinear_solver=="snes high level"`

python/dolfinx/fem/petsc.py
```diff
@@ -1388,16 +1356,17 @@ def solve(self) -> tuple[PETSc.Vec, int, int]:  # type: ignore
         """
 
         # Move current iterate into the work array.
```
garth-wells: This isn't a move. It's a copy.

python/dolfinx/fem/petsc.py
garth-wells: A ghost update (forward) is required at some point for a nonlinear solver, but it should come before the vector is used to assemble the residual rather than after the solve.

python/dolfinx/fem/petsc.py
```diff
@@ -1431,6 +1400,15 @@ def A(self) -> PETSc.Mat:
         """
         return self._A
 
+    @property
+    def P_mat(self) -> typing.Optional[PETSc.Mat]:
+        """Preconditioner matrix.
+
+        Note:
+            The matrix has an options prefix set.
```
garth-wells: And the options prefix is?

python/test/unit/fem/test_petsc_solver_wrappers.py
garth-wells: Test is comparing solutions, not solvers.

python/test/unit/fem/test_petsc_solver_wrappers.py
```diff
@@ -18,9 +18,11 @@
 @pytest.mark.petsc4py
 class TestPETScSolverWrappers:
     @pytest.mark.parametrize(
-        "mode", [dolfinx.mesh.GhostMode.none, dolfinx.mesh.GhostMode.shared_facet]
+        "mode",
+        [dolfinx.mesh.GhostMode.none, dolfinx.mesh.GhostMode.shared_facet],
     )
-    def test_compare_solvers(self, mode):
+    @pytest.mark.parametrize("nonlinear_solver", ["snes high level", "legacy Newton"])
```
garth-wells: I think this parameterisation degrades readability. It triggers and an if/else in the test, so might as well test both cases at the same time.

garth-wells (changes requested):

python/dolfinx/fem/petsc.py
francesco-ballarin: Indeed there is a ghost update in the `assemble_residual` function. That one has always been there.

python/dolfinx/fem/petsc.py
francesco-ballarin: Will fix this, although this line is in `main` already (i.e., I didn't change it).

python/dolfinx/fem/petsc.py
francesco-ballarin: Can fix this, although this line is in `main` already (i.e., I didn't change it, I just moved it up in the file).

python/test/unit/fem/test_petsc_solver_wrappers.py
francesco-ballarin: True, will give it a try later.

python/test/unit/fem/test_petsc_solver_wrappers.py
francesco-ballarin: I guess the point of the name is that the test compare solutions for linear vs nonlinear solvers. Still, as above I didn't choose this name so I can't comment on the rationale they had in mind.

python/dolfinx/fem/petsc.py
francesco-ballarin: The user can call the `getOptionsPrefix()` method of `PETSc.Mat`/`PETSc.Vec`. Do we want to write this explicitly (every time there is this note?)

python/test/unit/fem/test_petsc_solver_wrappers.py
francesco-ballarin: I gave it another try.

python/test/unit/fem/test_petsc_solver_wrappers.py
francesco-ballarin: I gave it another try (name change)

python/dolfinx/fem/petsc.py
francesco-ballarin: Fixed here and a few lines after.

python/dolfinx/fem/petsc.py
```diff
@@ -868,7 +862,7 @@ def __init__(
         self.solver.setOperators(self.A, self.P_mat)
 
         # Give PETSc objects a unique prefix
-        problem_prefix = f"dolfinx_linearproblem_{id(self)}_"
+        problem_prefix = f"dolfinx_linearproblem_{comm.bcast(id(self), root=0)}_"
```
garth-wells: This is an example of trying to be too clever. A user would never expect that setting a prefix would involve a collective MPI operation. We got in a pickle in the past trying to force PETSc to conform to our desired behaviour rather than using PETSc as it was designed to be used.

If the prefix is important, we should require the user to pass it.

francesco-ballarin: I changed this to a class-level counter, without any MPI operation. I think that the rationale for generating the `problem_prefix` internally rather than asking to the user is that if the user provides the same prefix twice then, the second time, the options from the two calls will be joined together, which will come as a unpleasant surprise for the average user.

python/dolfinx/fem/petsc.py
```diff
@@ -753,6 +747,8 @@ class LinearProblem:
     using PETSc as a linear algebra backend.
     """
 
+    _prefix_counter = 0
```
garth-wells: What happens if `LinearProblem`s are created on some ranks and not others? Looks like the prefix will get messed up.

python/dolfinx/fem/petsc.py
jhale: Sorry for the bug on `id(self)`, I blindly copied that without thinking about the MPI context.

On balance we should probably just require an explicit prefix - good to get users into the habit of using prefixes early.

python/dolfinx/fem/petsc.py
jhale: My understanding was as @garth-wells so this new ghost update is unexpected to me, so a bit more investigation/thought should be put into why it's necessary before merging.

python/dolfinx/fem/petsc.py
jhale: Of the three options proposed (sharing state on comm, explicit passing prefix by user, a counter) the prefix_counter and different handling whether we are on COMM_WORLD etc. is my least favourite.

python/test/unit/fem/test_petsc_nonlinear_assembler.py
jhale: Why return the Function `u`? The user invariably has that already because they must pass it themselves to the constructor. This is quite different to the `LinearProblem` where the user doesn't usually have `u` and it is often created within the class. NonlinearProblem's `x` is the same in that sense.

python/dolfinx/fem/petsc.py
francesco-ballarin: I now tried with a new approach that mixes input argument and default internal construction

python/test/unit/fem/test_petsc_nonlinear_assembler.py
francesco-ballarin: Does the average user ever really need `x`? I'd say that they typically need the `fem.Function`, not the internal vector. Your point that the `fem.Function` is already passed in is valid, but my overall rationale was to make the return type consistent between linear and nonlinear.

python/dolfinx/fem/petsc.py
francesco-ballarin: Let's see if this minimal example helps (which basically mimicks the test in `test_compare_solution_linear_vs_nonlinear_problem`)

```
import sys

import mpi4py.MPI
import numpy as np
import petsc4py.PETSc

import dolfinx.mesh
import dolfinx.fem
import dolfinx.fem.petsc
import dolfinx.la.petsc
import ufl

assert len(sys.argv) > 1

if sys.argv[1] == "main":
    do_ghost_update = False
elif sys.argv[1] == "PR":
    do_ghost_update = True
else:
    raise RuntimeError("Invalid argument")

msh = dolfinx.mesh.create_unit_square(mpi4py.MPI.COMM_WORLD, 3, 3)
V = dolfinx.fem.functionspace(msh, ("Lagrange", 2))
uh = dolfinx.fem.Function(V)
v = ufl.TestFunction(V)
x = ufl.SpatialCoordinate(msh)
f = x[0] * x[1]
F = ufl.inner(uh, v) * ufl.dx - ufl.inner(f, v) * ufl.dx
u = ufl.TrialFunction(V)

eps = 100 * np.finfo(dolfinx.default_scalar_type).eps
petsc_options = {
    "ksp_type": "preonly",
    "pc_type": "lu",
    "pc_factor_mat_solver_type": "mumps",
    "snes_atol": eps,
    "snes_rtol": eps,
}

nonlinear_problem = dolfinx.fem.petsc.NonlinearProblem(
    F,
    uh,
    petsc_options=petsc_options
)
nonlinear_problem.solve()

if do_ghost_update:
    dolfinx.la.petsc._ghost_update(
        nonlinear_problem.x, petsc4py.PETSc.InsertMode.INSERT, petsc4py.PETSc.ScatterMode.FORWARD)
    dolfinx.fem.petsc.assign(nonlinear_problem.x, uh)

error_ufl = dolfinx.fem.form(ufl.inner(uh - f, uh - f) * ufl.dx)
error = np.sqrt(msh.comm.allreduce(dolfinx.fem.assemble_scalar(error_ufl), op=mpi4py.MPI.SUM))
print(error)
```

These are the calls I made and the correspding results:
```
$ python3 standalone.py main
8.902307474921183e-17
$ python3 standalone.py PR
8.902307474921183e-17
$ mpirun -n 2 python3 standalone.py main
0.08304706201897356
0.08304706201897356
$ mpirun -n 2 python3 standalone.py PR
9.234168709261581e-17
9.234168709261581e-17
```

python/dolfinx/fem/petsc.py
francesco-ballarin: > Of the three options proposed (sharing state on comm, explicit passing prefix by user, a counter) the prefix_counter and different handling whether we are on COMM_WORLD etc. is my least favourite.

I don't have a strong opinion, just let me know @garth-wells @jhale what is a reasonable tradeoff. The only strong opinion I have is that the user must not be forced to provide the prefix themselves, because it gets trick to remember to change the string value every time, and most users will surely do a copy-paste if they are solving two problems in the same python file.

jhale: Our users are smart people and what smart people don't like are unexpected behaviours, or things that change based on edge cases (COMM_WORLD vs others).

1. We're agreed that unique prefixes are good, and that not setting a prefix by default (None) is probably a bad thing to do.
2. There is no good way to set a unique prefix across ranks, especially when dealing with complex communicators.
3. So let the user pass in an options prefix as a mandatory argument. Explain clearly in the docstring that it must be unique.

python/test/unit/fem/test_petsc_solver_wrappers.py
jhale: Thanks, I'm convinced - we should go back through our tests and check carefully whether we are missing anything.

python/dolfinx/fem/petsc.py
francesco-ballarin: I can do that, but I'd like to hear the opinion from a few others first.

IMHO "let the user pass in an options prefix as a mandatory argument" is a tough ask for the majority of the users, considering that the case where the default construction is not working is a less frequent case. Still, I understand your point about the fact that it's better to be explicit, especially in an instance like this one that educates the user.

For reference, this afternoon I tried making `_prefix_counter` a `defauldict(lambda: 0)`, hoping to be able to use the communicator as key. Unfortunately, when testing this attempt I realized that communicators are not hashable and their `.Get_name()` does not get preserved by `MPI_Comm_dup`, so there is no good way to prepare a key for this dict.

wence-: If you store your own (library) communicator on the user's communicator that they passed in (effectively `MPI_Comm_dup` the user communicator and then do `MPI_Comm_setattr(user_comm, library_comm)`, and then query the user comm for the library comm when you're given it again), then you can use a communicator attribute to store the counter.

jhale: Thanks @wence- that's potentially a nice technique, but I'd think we'd have to look at how we're dealing with MPI comms throughout in a separate PR before trying that.

francesco-ballarin: Marking this as resolved in order to clean up the page a bit, you'll receive another message shortly after with the latest attempt ("forcing" users to provide a prefix)

python/test/unit/fem/test_petsc_nonlinear_assembler.py
francesco-ballarin: Marking this as resolved in order to clean up the page a bit, you'll receive another message shortly after with the latest attempt (a compromise that would make both of us happy)

python/test/unit/fem/test_petsc_solver_wrappers.py
francesco-ballarin: I think what we missed was that we were only comparing the output `x`, rather than anything that had to do with the solution `u`.

francesco-ballarin:
@garth-wells @jhale I tried to address your concerns:
1. the user must now provided a `petsc_options_prefix` argument. Both `petsc_options_prefix` and `petsc_options` are still optional arguments, but the documentation clearly states that the only case in which the default arguments are accepted is the case in which the user does not want to set any option at all. An error is raised if `petsc_options` is provided but  `petsc_options_prefix` is missing to ease the upgrade of user codes from 0.9.0.
2. `def solve` now returns the solution function, the solution vector, convergence reason and number of iterations. Hopefully this will make both myself and Jack happy.

jhale:
I've rewritten the docstring and even with my best efforts, the proposed design just creates a lot of difficult conjunctions.

I don't follow your second argument.

Regarding the signature, it is Pythonic that if something is required, it's at the beginning of the signature. We agree that this is required, and that the no option prefix path is an edge case with potential pitfalls.

Regarding prettiness of the actual call, in all of our examples we can pass `petsc_options_prefix` as a named mandatory argument and so we can place it right up where `petsc_options` is - order in the call is irrelevant for named arguments.

```
>>> def f(a, b, petsc_options_prefix, bcs=None, petsc_options=None):
...     pass
...
>>> a = 0
>>> b = 1
>>> f(a, b, bcs=[bc], petsc_options_prefix="prefix", petsc_options={})
```

Runs without error.

jhale:
`LinearProblem(my_a, my_L, my_bcs)` - if users were relying on a kwarg being always in a position, then they will need to fix their code.

jhale:
There is another Pythonic option but it hasn't yet been used in DOLFINx's Python interface:

`def f(a, b, *, petsc_options_prefix, bcs=None):`

This makes `petsc_options_prefix` a mandatory keyword-only argument.

jhale:
I have implemented `def f(a, b, *, petsc_options_prefix, bcs=None):` which makes `petsc_options_prefix` a mandatory named argument. If the user does not pass `petsc_options_prefix` they receive a clear error message created without any internal logic in the constructor. This is a standard Python feature. Users relying on `bcs` being the third positional argument will need to adjust their code.

jhale:
I have reviewed all of the comments regarding `petsc_options_prefix` and the solution proposed now appears to satisfy all of the requirements.

1. Only one behaviour.
2. Explicit.
3. No MPI tricks.
4. `petsc_options_prefix` can be mandatory and sit up near `petsc_options` (or anywhere else the user desires).

I would point developers towards considering `\` and `*` in function signatures - these are features we are possibly underusing at the moment.

francesco-ballarin:
Thanks @jhale , the * in the function signature did the trick!

python/dolfinx/fem/petsc.py
```diff
@@ -970,7 +985,7 @@ def solve(self) -> tuple[typing.Union[_Function, Iterable[_Function]], int, int]
         self.solver.solve(self.b, self.x)
         dolfinx.la.petsc._ghost_update(self.x, PETSc.InsertMode.INSERT, PETSc.ScatterMode.FORWARD)
         dolfinx.fem.petsc.assign(self.x, self.u)
-        return self.u, self.solver.getConvergedReason(), self.solver.getIterationNumber()
+        return self.u, self.x, self.solver.getConvergedReason(), self.solver.getIterationNumber()
```
nate-sime, line 988: Why do we have to make an interface change here? Further to @jorgensd's comments, `x` and `u` should be accessible members.

Furthermore I feel like the pendulum is swinging here back to the legacy DOLFIN style. The DOLFINx PETSc module should facilitate PETSc and not wrap it. The user/developer should know what PETSc is when they invoke `import dolfinx.fem.petsc`. Therefore they can be trusted to understand `linear_solver.solver.getConvergedReason()` for their own purpose.

PR #3787, opened 2025-07-06
Make `bcs` handling in `fem.petsc.set_bc` and `fem.petsc.apply_lifting` uniform

### LINKED ISSUE #3786: `bcs` handling is asymmetric in `fem.petsc.set_bc` and `fem.petsc.apply_lifting`
In block/nest cases in `fem.petsc`, the `bcs` argument is treated differently when applying lifting and when setting the bc value in a vector.

`fem.petsc.set_bc` forces the user to call `fem.bcs_by_block` beforehand, see https://github.com/FEniCS/dolfinx/blob/a6dc610a5cdd10e31c12cf8d0d765235f528ff1a/python/dolfinx/fem/petsc.py#L715-L720

Instead, `fem.petsc.apply_lifting` accepts as input either the output of the appropriate call to `fem.bcs_by_block`, or calls it internally, see
https://github.com/FEniCS/dolfinx/blob/a6dc610a5cdd10e31c12cf8d0d765235f528ff1a/python/dolfinx/fem/petsc.py#L597-L614
https://github.com/FEniCS/dolfinx/blob/a6dc610a5cdd10e31c12cf8d0d765235f528ff1a/python/dolfinx/fem/petsc.py#L637-L640
https://github.com/FEniCS/dolfinx/blob/a6dc610a5cdd10e31c12cf8d0d765235f528ff1a/python/dolfinx/fem/petsc.py#L664-L667
https://github.com/FEniCS/dolfinx/blob/a6dc610a5cdd10e31c12cf8d0d765235f528ff1a/python/dolfinx/fem/petsc.py#L685-L688

Do we really want this asymmetry? While I understand the convenience of not having to manually call `fem.bcs_by_block`, the asymmetry is reflected in our demos, which may complicate the end-user understanding of them. For instance `demo_hdg.py` manually extracts `bcs1` 
https://github.com/FEniCS/dolfinx/blob/a6dc610a5cdd10e31c12cf8d0d765235f528ff1a/python/demo/demo_hdg.py#L184-L189
and passes it to `apply_lifting`, while `demo_stokes.py` let `apply_lifting` do the extraction internally https://github.com/FEniCS/dolfinx/blob/a6dc610a5cdd10e31c12cf8d0d765235f528ff1a/python/demo/demo_stokes.py#L407-L411

francesco-ballarin:
Fixes #3786


python/dolfinx/fem/petsc.py
```diff
@@ -594,25 +594,12 @@ def apply_lifting(
...
-                   bcs1 = fem.bcs_by_block(
-                    fem.extract_function_spaces(a, 1), bcs
-                   )
-
-            2. ``bcs`` is a sequence of :class:`dolfinx.fem.DirichletBC`
-               objects. The function deduces which `DiricletBC` objects
-               apply to each column of ``a`` by matching the
-               :class:`dolfinx.fem.FunctionSpace`.
-
+        bcs: Boundary conditions to apply. If ``b`` is nested or
```
francesco-ballarin: Well, I copied the documentation from `set_bc` as a reminder of what I am trying to achieve, but I am not there yet.

Comparison between the two functions:
- `set_bc` requires 1D `bcs` input for non-blocked problems, and 2D `bcs` for blocked problem.
- ideally, `apply_lifting` should do the same, but:
   - on `main`, it accepts 1D `bcs` input for either blocked or non-blocked problems. For non-blocked problems, it internally converts into a 2D array.
   - the goal would be to requires 1D `bcs` input for non-blocked problems, and 2D `bcs` for blocked problem. However:
   - this branch, as it is now, forces 2D `bcs` input, which is still not what I would have wanted in #3786. I'll fix this, but I'll probably wait for a discussion if #3786 is worth fixing or not.

francesco-ballarin: No, scratch this. `dolfinx.fem.apply_lifting` requires the input to be 2D, so that same requirement will go into `dolfinx.fem.petsc.apply_lifting`.

PR #3789, opened 2025-07-09
Go back to LBYL programming style (instead of EAFP) in `fem.petsc` and `la.petsc`

### LINKED ISSUE #3738: LBYL vs. EAFP (isinstance vs. try/catch)

francesco-ballarin:
Fixes #3738


francesco-ballarin:
Sorry, this was built upon #3785, which I didn't mean to do. Will cherry pick on a new branch and open a new PR.

PR #3790, opened 2025-07-09
Go back to LBYL programming style (instead of EAFP) and several `mypy` fixes in `fem.petsc` and `la.petsc`

### LINKED ISSUE #3738: LBYL vs. EAFP (isinstance vs. try/catch)

francesco-ballarin:
- Works towards #3738 for `fem.petsc` and `la.petsc`
- Replace `Iterable` with `Sequence` in all public function signatures in `fem.petsc` and `la.petsc`. See https://github.com/FEniCS/dolfinx/pull/3790#discussion_r2199648315 .
- Remove module level `# mypy: ignore-errors` in `fem.petsc`: this has uncovered several typing errors that are fixed here.

(replaces #3789 which was not branched off of `main`)


python/dolfinx/la/petsc.py
```diff
@@ -99,20 +99,20 @@ def assign(x0: typing.Union[npt.NDArray[np.inexact], list[npt.NDArray[np.inexact
...
+    if x1.getType() == PETSc.Vec.Type().NEST:
         x1_nest = x1.getNestSubVecs()
         for _x0, _x1 in zip(x0, x1_nest):
             with _x1.localForm() as x:
                 x.array_w[:] = _x0
-    except PETSc.Error:  # type: ignore[attr-defined]
+    else:
         with x1.localForm() as _x:
-            try:
+            if isinstance(x0, list):
```
schnellerhase:
```suggestion
            if isinstance(x0, Iterable):
```

python/dolfinx/la/petsc.py
```diff
@@ -129,18 +129,18 @@ def _(x0: PETSc.Vec, x1: typing.Union[npt.NDArray[np.inexact], list[npt.NDArray[
...
+    if x0.getType() == PETSc.Vec.Type().NEST:
         x0_nest = x0.getNestSubVecs()
         for _x0, _x1 in zip(x0_nest, x1):
             with _x0.localForm() as x:
                 _x1[:] = x.array_r[:]  # type: ignore
-    except PETSc.Error:  # type: ignore[attr-defined]
+    else:
         with x0.localForm() as _x0:
-            try:
+            if isinstance(x1, list):
```
schnellerhase:
```suggestion
            if isinstance(x1, Iterable):
```

python/dolfinx/fem/petsc.py
```diff
@@ -1132,26 +1135,24 @@ def assemble_residual(
     # Assign the input vector to the unknowns
     assign(x, u)
 
+    # Assign block data if block assembly is requested
+    if isinstance(residual, Iterable) and b.getType() != PETSc.Vec.Type.NEST:
+        _assign_block_data(residual, b)  # type: ignore
```
```suggestion
        _assign_block_data(residual, b)
```

python/dolfinx/fem/petsc.py
```diff
@@ -1132,26 +1135,24 @@ def assemble_residual(
...
+        _assign_block_data(residual, x)  # type: ignore
```
```suggestion
        _assign_block_data(residual, x)
```

python/dolfinx/fem/petsc.py
```diff
@@ -1132,26 +1135,24 @@ def assemble_residual(
...
     # Assemble the residual
     _zero_vector(b)
-    try:
-        # Single form and nest assembly
-        assemble_vector(b, residual)
-    except TypeError:
-        # Block assembly
-        _assign_block_data(residual, b)  # type: ignore
-        assemble_vector(b, residual)  # type: ignore
+    assemble_vector(b, residual)  # type: ignore
```
```suggestion
    assemble_vector(b, residual)
```

python/dolfinx/la/petsc.py
francesco-ballarin: I agree that this would work, but the type hints in the function signature have `list[npt.NDArray[np.inexact]]` so I used the same type. I am happy with using `Iterable`, but we should do so in both places.

python/dolfinx/la/petsc.py
francesco-ballarin: As above.

python/dolfinx/la/petsc.py
jhale: Let's change everything to Iterable. It would also be nice if functions returned the same iterable as they were passed, but for another time... :)

francesco-ballarin: There currently is a mixture of Iterable and Sequence in many typing annotations, especially in `fem/petsc.py.` Will we ever use iterables, or will they actually always be sequences? I can't understand from the code why some are iterables and others are just sequences.

jhale: Bottom line, although there is some more complexity than this, Iterable supports `for x in a:` while Sequence supports `for x in a:` and `a[0]`. So use Iterable if sufficient, otherwise Sequence.

jhale: By the way, type checking is disabled in `fem/petsc.py` so I have no idea if what's there is actually correct at all.

francesco-ballarin: > By the way, type checking is disabled in fem/petsc.py so I have no idea if what's there is actually correct at all.

Good catch. I think that such line was added in the past because `petsc4py` typing support is very lacking, and to avoid needing to add an ignore at every single line that uses `petsc4py`. I'll try to have a look at how tedious it would be to enable mypy globally in the file and disable line by line `petsc4py` related errors.

> Bottom line, although there is some more complexity than this, Iterable supports for x in a: while Sequence supports for x in a: and a[0]. So use Iterable if sufficient, otherwise Sequence.

That's the textbook definition, but our case is much more complicated than that, for at least the following reasons:
1. a true iterable (for instance a generator, rather than a list which is an iterable but also a sequence) consumes its content when iterating. If I look at the third line of
https://github.com/FEniCS/dolfinx/blob/4ca0f4534c1fb06f657f7db3b9001c074ee5c8fa/python/dolfinx/fem/petsc.py#L452-L454
we claim that having `a: Iterable[...]` is ok. Yet,
https://github.com/FEniCS/dolfinx/blob/4ca0f4534c1fb06f657f7db3b9001c074ee5c8fa/python/dolfinx/fem/petsc.py#L461-L462
iterate over the input `a` twice. If `a` was a true iterable, the loop in the second line would result in `coeffs=[]`.
Hence, at least in this case, the claim that `a` can be an iterable is false even if we never use `a[0]`. If the user were to pass `a` as a true iterable `mypy` would not complain, but the end user would surely either get an error as soon as `coeffs` is used, or worst case scenario simply not get the right result
2. the `bcs` argument in https://github.com/FEniCS/dolfinx/blob/4ca0f4534c1fb06f657f7db3b9001c074ee5c8fa/python/dolfinx/fem/petsc.py#L689-L691
is currently marked as `bcs: Iterable[...]`, yet 
https://github.com/FEniCS/dolfinx/blob/4ca0f4534c1fb06f657f7db3b9001c074ee5c8fa/python/dolfinx/fem/petsc.py#L720
uses `bcs[0]`, and thus this will definitely be needed to changed to `Sequence[...]`. My point now is that as soon as we are forced to make a change from `Iterable[...]` to `Sequence[...]` in one of the functions in this file, then we are better off in making the same change in the argument of every other function in this file that has the same name. The rationale for my statement is that the user will typically define a variable (say `my_bcs`), and then pass that same variable to `assemble_matrix`, `set_bc`, `apply_lifting`, so even if they tried to stick with an iterable `my_bcs` and `mypy` would be ok in they passing it to the function that is more permessive (say, `apply_lifting`), as soon as they get to the call that is more restrictive (in this case `set_bc`) `mypy` would complain, and the end result is that they would be forced to change their definition of `my_bcs` to a sequence anyways.  [See also footnote A]
3. we clearly don't have any test that covers using generators (or a similar true iterable) for those arguments, otherwise it would have caught item 1. Apart from the technical issue, I don't really see any valid reason for users to ever need to pass in a generator.
4. less convincing, but I'll write this anyway [see footnote B]. For this iterable arguments, the C++ backend typically has `vector<T>`. I haven't checked the details of `nanobind`, but if I had a `def function_python(arg_python: Iterable[T])` that calls `void function_cpp(vector<T> arg_cpp)`, I would imagine that internally `nanobind` constructs `arg_cpp` as someting like `arg_cpp = [x for x in arg_python]`, as there would be no way to go back and forth to the python generator and generate on the fly the next element on the C++ side by calling the python side. Hence, in practice, even if `arg_python` can be an iterable, it is basically converted to a sequence before being used.

If @jhale is convinced by at least a few of these arguments, then I'll go through the `Iterable[...]` annotations in these two files and change them to `Sequence[...]` as soon as the fail at least one of the items 1-4 above. I'll do the check one by one for each `Iterable[...]` annotation, but I am pretty much convinced that it will end up with every `Iterable[...]` annotation become a `Sequence[...]`.

Your honour, I rest my case now :joy:


Footnote A: if you wanted to play devil's advocate you could argue that this specific counterexample is my fault, as it is a side effect of having moved from LBYL to EAFP. I don't think that would really disprove my point, I'd probably go and find another more involved counterexample.

Footnote B: I am saying that this is less convincing because I've checked every use of `_cpp.fem.petsc` in `fem.petsc` and we always pass in an auxiliary list that is created in the function body, rather than the original input argument, for instance
https://github.com/FEniCS/dolfinx/blob/4ca0f4534c1fb06f657f7db3b9001c074ee5c8fa/python/dolfinx/fem/petsc.py#L234

python/dolfinx/fem/assemble.py
```diff
@@ -50,7 +50,7 @@ def pack_constants(
     def _pack(form):
         if form is None:
             return None
-        elif isinstance(form, collections.abc.Iterable):
+        elif isinstance(form, collections.abc.Sequence):
```
schnellerhase:
```suggestion
        elif isinstance(form, Sequence):
```
Maybe, also remove the `import collections`?

python/dolfinx/fem/assemble.py
```diff
@@ -84,7 +84,7 @@ def pack_coefficients(
     def _pack(form):
         if form is None:
             return {}
-        elif isinstance(form, collections.abc.Iterable):
+        elif isinstance(form, collections.abc.Sequence):
```
schnellerhase:
```suggestion
        elif isinstance(form, Sequence):
```

python/dolfinx/fem/forms.py
```diff
@@ -140,24 +141,24 @@ def get_integration_domains(
         return []
     else:
         domains = []
-        try:
+        if not isinstance(subdomain, list):
```
schnellerhase, line 144:
```suggestion
        if not isinstance(subdomain, Sequence):
```

python/dolfinx/la/petsc.py
```diff
@@ -112,25 +116,28 @@ def assign(x0: typing.Union[npt.NDArray[np.inexact], list[npt.NDArray[np.inexact
...
+    if x1.getType() == PETSc.Vec.Type().NEST:  # type: ignore[attr-defined]
         x1_nest = x1.getNestSubVecs()
         for _x0, _x1 in zip(x0, x1_nest):
             with _x1.localForm() as x:
                 x.array_w[:] = _x0
-    except PETSc.Error:  # type: ignore[attr-defined]
+    else:
         with x1.localForm() as _x:
-            try:
+            if isinstance(x0, collections.abc.Sequence):
```
schnellerhase:
```suggestion
            if isinstance(x0, Sequence):
```

python/dolfinx/la/petsc.py
```diff
@@ -86,7 +87,10 @@ def create_vector_wrap(x: Vector) -> PETSc.Vec:  # type: ignore[name-defined]
 
 
 @functools.singledispatch
-def assign(x0: typing.Union[npt.NDArray[np.inexact], list[npt.NDArray[np.inexact]]], x1: PETSc.Vec):  # type: ignore
+def assign(
+    x0: typing.Union[npt.NDArray[np.inexact], typing.Sequence[npt.NDArray[np.inexact]]],
```
schnellerhase:
```suggestion
    x0: typing.Union[npt.NDArray[np.inexact], Sequence[npt.NDArray[np.inexact]]],
```

python/dolfinx/la/petsc.py
```diff
@@ -142,18 +149,18 @@ def _(x0: PETSc.Vec, x1: typing.Union[npt.NDArray[np.inexact], list[npt.NDArray[
...
         x0_nest = x0.getNestSubVecs()
         for _x0, _x1 in zip(x0_nest, x1):
             with _x0.localForm() as x:
-                _x1[:] = x.array_r[:]  # type: ignore
-    except PETSc.Error:  # type: ignore[attr-defined]
+                _x1[:] = x.array_r[:]  # type: ignore[index]
+    else:
         with x0.localForm() as _x0:
-            try:
+            if isinstance(x1, collections.abc.Sequence):
```
schnellerhase:
```suggestion
            if isinstance(x1, Sequence):
```

schnellerhase (commented):
We should maybe aim to activate `warn_unused_ignores` in the mypy settings to get notified when all these PETSc ignores are no longer necessary.

francesco-ballarin:
> We should maybe aim to activate `warn_unused_ignores` in the mypy settings to get notified when all these PETSc ignores are no longer necessary.

Agreed, but this will not happen in this PR.

python/dolfinx/fem/forms.py
francesco-ballarin: As before, my rationale is that the typing annotation says "list", and I'll stick with what the annotation says.

python/dolfinx/la/petsc.py
```diff
@@ -112,25 +116,28 @@ def assign(x0: typing.Union[npt.NDArray[np.inexact], list[npt.NDArray[np.inexact
...
+            else:
                 _x.array_w[:] = x0
 
 
-@assign.register(PETSc.Vec)  # type: ignore[attr-defined]
-def _(x0: PETSc.Vec, x1: typing.Union[npt.NDArray[np.inexact], list[npt.NDArray[np.inexact]]]):  # type: ignore
+@assign.register
+def _(
+    x0: PETSc.Vec,  # type: ignore[name-defined]
+    x1: typing.Union[npt.NDArray[np.inexact], typing.Sequence[npt.NDArray[np.inexact]]],
```
schnellerhase:
```suggestion
    x1: typing.Union[npt.NDArray[np.inexact], Sequence[npt.NDArray[np.inexact]]],
```

francesco-ballarin:
Principle agreed upon in previous discussions. Queuing this to merge to unblock other PRs (e.g. #3794).

PR #3794, opened 2025-07-10
Switch from  EAFP to LBYL

schnellerhase:


francesco-ballarin:
Sorry @schnellerhase , I am moving it back to draft to avoid that anyone merges this, as it must be merged after https://github.com/FEniCS/dolfinx/pull/3785 and https://github.com/FEniCS/dolfinx/pull/3790 otherwise issue https://github.com/FEniCS/dolfinx/issues/3738 would be closed prematurely.

garth-wells:
We need to find balance - best is the simplest solution that meets a need. It's not the case that everything should be EAFP or that everything should be LBYL.

python/dolfinx/fem/bcs.py
```diff
@@ -219,17 +219,16 @@ def dirichletbc(
     # Unwrap value object, if required
     if isinstance(value, np.ndarray):
         _value = value
+    elif isinstance(value, (Function, Constant)):
+        _value = value._cpp_object  # type: ignore
     else:
-        try:
```
michalhabera: I agree with @garth-wells 's latest comment that we need to find a balance. Examples as here, getting `_cpp_object` reads better with EAFP I think. The `isinstance` is unnecessary explicit and narrowing in these cases. However, a simple `_value = getattr(value, "_cpp_object", value)` might be even better.

python/dolfinx/fem/bcs.py
```diff
@@ -219,17 +219,16 @@ def dirichletbc(
...
-            _value = value._cpp_object
-        except AttributeError:
-            _value = value  # type: ignore[assignment]
+        raise ValueError(f"Unsupported value tpye {type(value)}.")
```
schnellerhase:
```suggestion
        raise ValueError(f"Unsupported value type {type(value)}.")
```

python/dolfinx/fem/function.py
```diff
@@ -142,10 +142,7 @@ def __init__(
 
         # Attempt to deduce dtype
         if dtype is None:
-            try:
-                dtype = e.dtype  # type: ignore
-            except AttributeError:
-                dtype = default_scalar_type
+            dtype = e.dtype if hasattr(e, "dtype") else default_scalar_type
```
schnellerhase:
```suggestion
            dtype = getattr(e, "dtype",  default_scalar_type)
```

PR #3796, opened 2025-07-11
Check PETSc options prefix is not empty

schnellerhase:
Follow up to https://github.com/FEniCS/dolfinx/pull/3785.

Passing an empty `petsc_options_prefix` string breaks. This is composed of PETSc's optional type handling  of `getOptionsPrefix` and input of `setOptionsPrefix` not aligning, and the internal empty c-string being encoded  as "\0" instead of "". 

As an empty prefix is not desirable, we just raise to avoid this.

Debugged with @jhale.

### DIFF
--- a/python/dolfinx/fem/petsc.py
+++ b/python/dolfinx/fem/petsc.py
@@ -888,6 +888,9 @@ class LinearProblem:
         self._solver = PETSc.KSP().create(self.A.comm)
         self.solver.setOperators(self.A, self.P_mat)
 
+        if petsc_options_prefix == "":
+            raise ValueError("PETSc options prefix can not be empty.")
+
         self._petsc_options_prefix = petsc_options_prefix
         self.solver.setOptionsPrefix(petsc_options_prefix)
         self.A.setOptionsPrefix(f"{petsc_options_prefix}A_")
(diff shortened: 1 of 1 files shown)


python/dolfinx/fem/petsc.py
```diff
@@ -888,6 +888,9 @@ def __init__(
         self._solver = PETSc.KSP().create(self.A.comm)
         self.solver.setOperators(self.A, self.P_mat)
 
+        if petsc_options_prefix == "":
+            raise ValueError("PETSc options prefix can not be empty.")
```
chrisrichardson:
```suggestion
            raise ValueError("PETSc options prefix cannot be empty.")
```

python/dolfinx/fem/petsc.py
```diff
@@ -1333,6 +1336,9 @@ def __init__(
         )
         self.solver.setFunction(partial(assemble_residual, u, self.F, self.J, bcs), self.b)
 
+        if petsc_options_prefix == "":
+            raise ValueError("PETSc options prefix can not be empty.")
```
chrisrichardson:
```suggestion
            raise ValueError("PETSc options prefix cannot be empty.")
```

Issue #2106, opened 2022-03-30
Replace edge_maps in refinement with AdjacencyList


jorgensd:
When computing refinement data, we currently create a `std::map` from the local edge numbers to the processes sharing that edge 
(https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/refinement/utils.h#L38-L39) 
In turn we use this map multiple times, with `map.find()`, which has logarithmic complexity;
https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/refinement/plaza.cpp#L645
https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/refinement/plaza.cpp#L87-L88
https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/refinement/utils.cpp#L256-L257

I wonder if it would be faster to use an `AdjacencyList<std::int32_t>` which has `num_local_edges` as number of nodes, and each link containing the neighborhood ranks.
Then every find would be replaced with:
```c++
auto map_it = shared_edges.links(local_edge)
if (!map_it.empty())
{
   // Do something...
}
```
Any thoughts @garth-wells ?

garth-wells:
In general, anything that replaces a `std::map` with a O(1) lookup and contiguous memory is good. The main question is probably how to compute the data efficiently in the first place.

chrisrichardson:
This is a good idea, let's do it once the current work on MeshTag refinement is in main.

jorgensd:
I wonder if we should rather go for something along the lines of what is done in: 
https://github.com/FEniCS/dolfinx/pull/3396
where we use a sorted `std::vector<std::int32_t, std::int64_t>>` and a `std::lower_bound` for searches.

Any thoughts on this?

PR #3797, opened 2025-07-11
Replace `std::map` in refinement

chrisrichardson:
Addresses #2106

### DIFF
--- a/cpp/dolfinx/refinement/utils.h
+++ b/cpp/dolfinx/refinement/utils.h
@@ -10,10 +10,10 @@
 #include <concepts>
 #include <cstdint>
 #include <dolfinx/common/MPI.h>
+#include <dolfinx/common/log.h>
 #include <dolfinx/graph/AdjacencyList.h>
 #include <dolfinx/mesh/Mesh.h>
 #include <dolfinx/mesh/utils.h>
-#include <map>
 #include <memory>
 #include <set>
 #include <span>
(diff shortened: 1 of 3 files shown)


cpp/dolfinx/refinement/utils.h
```diff
@@ -42,18 +42,17 @@ std::int64_t local_to_global(std::int32_t local_index,
...
 
 /// @brief Create geometric points of new Mesh, from current Mesh and a
-/// edge_to_vertex map listing the new local points (midpoints of those
-/// edges).
+/// list of edges (with new points created at midpoints of those edges).
 ///
 /// @param mesh Current mesh.
-/// @param local_edge_to_new_vertex A map from a local edge to the new
-/// global vertex index that will be inserted at its midpoint.
+/// @param local_edge_new_vertices A list of local edges requiring new vertices
```
schnellerhase:
```suggestion
/// @param marked_edge_list A list of local edges requiring new vertices
```

(and other occurrences)

cpp/dolfinx/refinement/utils.h
```diff
@@ -244,22 +239,42 @@ create_new_vertices(MPI_Comm comm,
...
-  std::vector<std::int32_t> recv_local_edge(recv_global_edge.size());
-  mesh.topology()->index_map(1)->global_to_local(recv_global_edge,
-                                                 recv_local_edge);
-  for (std::size_t i = 0; i < received_values.size() / 2; ++i)
   {
-    assert(recv_local_edge[i] != -1);
-    [[maybe_unused]] auto it = local_edge_to_new_vertex.insert(
-        {recv_local_edge[i], received_values[i * 2 + 1]});
-    assert(it.second);
+    std::vector<std::int64_t> recv_global_edge;
```
schnellerhase:
```suggestion
    std::vector<std::int64_t> recv_global_edge;
    recv_global_edge.reserve(received_values.size() / 2);
```

cpp/dolfinx/refinement/utils.h
```diff
@@ -244,22 +239,42 @@ create_new_vertices(MPI_Comm comm,
...
   }
 
-  return {std::move(local_edge_to_new_vertex), std::move(new_vertex_coords),
+  // Compile marked_edge_list and local_edge_to_new_vertex into an
+  // AdjacencyList
+  // Permute data into ascending order of marked_edge_list
+  std::vector<std::int32_t> perm(marked_edge_list.size());
+  std::iota(perm.begin(), perm.end(), 0);
+  std::sort(perm.begin(), perm.end(), [&marked_edge_list](int a, int b)
+            { return marked_edge_list[a] < marked_edge_list[b]; });
```
schnellerhase:
```suggestion
  std::ranges::sort(perm, std::less{}, [&marked_edge_list](int i) {return marked_edge_list[i]; });
```

Should we use radix sort here, we are sorting indices? The same question came up when looking into the dual graph computation. When should we opt for radix sort and when not?

chrisrichardson: TBD. But this is fine for now.

Issue #3783, opened 2025-07-06
Add high-level interface to `SLEPc EPS`


francesco-ballarin:
### Describe new/missing feature

We now have a `dolfinx.fem.petsc.LinearProblem` and `dolfinx.fem.petsc.NonlinearProblem`.
It would be nice to also have a `dolfinx.fem.slepc.LinearEigenvalueProblem` that interfaces `EPS`.

I will work on this in the next few days.

### Suggested user interface

```python3

```

jhale:
How far down the rabbit hole do we want to go with adding interfaces? LinearProblem and NonlinearProblem were added because  the first is common place for beginners and the second non-trivial. EPS seems, on the face of it, to be neither.

jhale:
TAO is more niche. I'd imagine that writing one's own assemblers is non-trivial, so we should see how the recently added assemble_jacobian and assemble_residual work there and if anything can be modified. @schnellerhase can probably advise.

I've never used TS so I couldn't comment.

jorgensd:
> TAO is more niche. I'd imagine that writing one's own assemblers is non-trivial, so we should see how the recently added assemble_jacobian and assemble_residual work there and if anything can be modified. [@schnellerhase](https://github.com/schnellerhase) can probably advise.
> 
> I've never used TS so I couldn't comment.

I tried to implement som TS stuff prior to the conference, and it was trickier than I hoped for.

I think having an EPS interface makes sense, as eigenvalue problems are commonplace.

jorgensd:
To throw away dirichlet BC dofs, you can use PETSc quite efficiently, as shown in:
https://fenicsproject.discourse.group/t/structural-modal-analysis/19479/2?u=dokken
i.e.
```python
def extract_sub_matrix(A:PETSc.Mat, V: dolfinx.fem.FunctionSpace, bcs:list[dolfinx.fem.DirichletBC])->PETSc.Mat:
    num_dofs_local = V.dofmap.index_map.size_local*V.dofmap.index_map_bs
    num_ghosts = V.dofmap.index_map.num_ghosts*V.dofmap.index_map_bs

    not_dirichlet = np.full(num_dofs_local+num_ghosts, 1, dtype=np.int32)

    for bc in bcs:
        not_dirichlet[bc.dof_indices()[0]] = 0

    idx_set = np.flatnonzero(not_dirichlet).astype(np.int32)
    idx_set_row = idx_set[idx_set<num_dofs_local]
[...]

PR #3799, opened 2025-07-12
Extend `fem.extract_function_space` to also work with UFL forms

francesco-ballarin:
Current implementation of `fem.extract_function_space` requires as inputs C++ wrapped forms, and returns as output C++ wrapped function spaces.

This PR allows as input UFL forms, and returns the corresponding python function spaces.

Came up while working on #3783, because I was trying to extract the function spaces from a bilinear form (as there wouldn't be a linear form in linear eigenvalue problems) and use them to construction a `dolfinx.fem.Function`, which can be created from the python function spaces but not from the C++ wrapped function spaces. Still, please review this regardless on what is your opinion about #3783, because this feature may be used independently on that issue being accepted or not.


garth-wells:
This doesn't look right to me. The return type doesn't look correct - with the change the return type could be a DOLFINx space or a UFL space. Right approach is probably that UFL operations are in UFL.

PR #3800, opened 2025-07-12
Extend test coverage for `fem.Function(..., dtype=dtype)` for `dtype` different from `PETSc.ScalarType`

francesco-ballarin:
Context: while working on #3783 I wanted to create a complex-valued `fem.Function` even if my `PETSc.ScalarType` is real. The rationale there is that the eigenfunctions may be complex-valued even if the PETSc matrix is real.
I can't remember this specific discussion in the past, and I can't seem to find any *basic* test coverage for this case.

Please review this regardless of your opinion about #3783. I think that having this case tested is a nice way to further show that we can untie ourselves from PETSc, and it is a low hanging fruit with all the `dtype` infrastracture already in place. After all, I can't think of a reason why one should be forbidden from creating a complex-valued function even if, for whatever reason (say, PETSc installation), they computed its real and its imaginary parts separately.

Feel free to suggest further tests to be added. For instance, in the case of #3783 I would expect that I would need to compute the L^2 norm of the eigenfunction. If this is not as straightforward as passing the custom `fem.Function.dtype` to `fem.form` then we may want to check if that works too. Seems to me however that the parametrization `dtype_parametrize` in `test/unit/fem/test_assembler.py`  already covers that.

### DIFF
--- a/python/dolfinx/__init__.py
+++ b/python/dolfinx/__init__.py
@@ -21,11 +21,33 @@ try:
 
     default_scalar_type = _PETSc.ScalarType  # type: ignore
     default_real_type = _PETSc.RealType  # type: ignore
+
+    import numpy as _np
+
+    if _np.issubdtype(default_scalar_type, _np.complexfloating):
+        default_complex_type = default_scalar_type
+    elif _np.issubdtype(default_scalar_type, _np.float32):
+        default_complex_type = _np.complex64
+    elif _np.issubdtype(default_scalar_type, _np.float64):
+        default_complex_type = _np.complex128
+    else:
+        import warnings
+
+        warnings.warn(f"Unable to determine default complex type from {default_scalar_type}")
+
+        default_complex_type = None
+
+        del warnings
+
+    del _np
 except ImportError:
     import numpy as _np
 
     default_scalar_type = _np.float64
     default_real_type = _np.float64
+    default_complex_type = _np.complex128
+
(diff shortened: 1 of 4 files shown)


garth-wells:
I don't quite follow: we've been able to create Functions with different scalar types, and independently of the PETSc build, for some time and quite a few tests are parameterised over the scalar type.

Is the main point of this PR the changes in `def petsc_vec(self):`?

I'd rather not make the changes in `python/dolfinx/__init__.py`. What we have already is a step towards being less explicit than we might like, so I'd rather not take it further.

francesco-ballarin:
> I don't quite follow: we've been able to create Functions with different scalar types, and independently of the PETSc build, for some time and quite a few tests are parameterised over the scalar type.

Yes, but I couldn't find a super simple test like the one I added at the end of `test_function.py`. They were all more involved, like using the Function in a form.

> Is the main point of this PR the changes in `def petsc_vec(self):`?

Arguably, yes. It's just a safeguard, as `numpy` would give an error anyways.

> I'd rather not make the changes in `python/dolfinx/__init__.py`. What we have already is a step towards being less explicit than we might like, so I'd rather not take it further.

OK, I will undo them.

francesco-ballarin:
Closing this, as it doesn't add much. If there will be anything from here I truly need when working on #3783 I will simply add it to that PR.

PR #3820, opened 2025-07-22
Fine tune `clang-tidy`

schnellerhase:
Addresses undesired changes introduced in https://github.com/FEniCS/dolfinx/pull/3722. And reactivates the clang tidy check temporary deactivated in https://github.com/FEniCS/dolfinx/pull/3814.

- All occurrences of `const std::shared_ptr<>&` are now removed. `performance-unnecessary-value-param`-check either expects pass by const ref or by value together with move semantics. The later is used throughout, consistent with previous use.
- Introduces custom special case handling for `nanobind::ndarray` to allow for it to be passed (as non-trivially copyable type) by value and not by `const &`.
- The same special case treatment could be applied to the `std::string` and `std::filesystem::path` arguments. However this would also disallow the checks in other critical parts of the code and it is preferable to stay consistent. Therefore this returns to the initial changes for those in the python exports. If not desired, we could also use another clang-tidy config file for the python exports to only disregard in this section.
- Deactivates by hand clang-tidy check for small array which should be passed by value. This could be used for any such cases that might arise in the future.


python/dolfinx/wrappers/refinement.cpp
```diff
@@ -165,7 +166,7 @@ void refinement(nb::module_& m)
             parent_meshtag, *topology1,
             std::span(parent_cell.data(), parent_cell.size()));
         return dolfinx::mesh::MeshTags<std::int32_t>(
-            topology1, tdim, std::move(entities), std::move(values));
+            std::move(topology1), tdim, std::move(entities), std::move(values));
```
garth-wells: `std::move` on `topology1`  is unnecessary and will have no effect since `MeshTags` takes the shared pointer by value.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -757,11 +758,12 @@ Topology::Topology(
 #endif
 
   // Set data
-  _index_maps.insert({{0, 0}, vertex_map});
   _connectivity.insert(
       {{{0, 0}, {0, 0}},
        std::make_shared<graph::AdjacencyList<std::int32_t>>(
            vertex_map->size_local() + vertex_map->num_ghosts())});
+  _index_maps.insert(
```
garth-wells: This change adds complexity for no meaningful performance benefit.

python/dolfinx/wrappers/fem.cpp
```diff
@@ -87,7 +87,7 @@ auto ptr_to_ref_wrapper_vec(auto& x)
 };
 
 template <typename T>
-void declare_function_space(nb::module_& m, std::string type)
+void declare_function_space(nb::module_& m, const std::string& type)
```
garth-wells: Passing strings by reference isn't really needed in cases like this because (i) the code here is isn't performance critical and (ii) the strings are usually short and compilers have 'short string optimisation'.

I prefer to let the developer make the decision because they know the context.

python/dolfinx/wrappers/fem.cpp
```diff
@@ -356,7 +357,9 @@ void declare_objects(nb::module_& m, std::string type)
              std::shared_ptr<const dolfinx::fem::FunctionSpace<U>> V)
           {
             new (bc) dolfinx::fem::DirichletBC<T, U>(
-                g, std::vector(dofs.data(), dofs.data() + dofs.size()), V);
+                std::move(g),
+                std::vector(dofs.data(), dofs.data() + dofs.size()),
+                std::move(V));
```
garth-wells: `std::move` on `V` not required and will have no effect.

python/dolfinx/wrappers/fem.cpp
```diff
@@ -383,7 +387,8 @@ void declare_objects(nb::module_& m, std::string type)
                                V_g_dofs[0].data() + V_g_dofs[0].size()),
                    std::vector(V_g_dofs[1].data(),
                                V_g_dofs[1].data() + V_g_dofs[1].size())};
-            new (bc) dolfinx::fem::DirichletBC(g, std::move(dofs), V);
+            new (bc) dolfinx::fem::DirichletBC(std::move(g), std::move(dofs),
```
garth-wells: No need to move `g`.

python/dolfinx/wrappers/fem.cpp
```diff
@@ -690,8 +695,9 @@ void declare_form(nb::module_& m, std::string type)
             }
 
             new (fp) dolfinx::fem::Form<T, U>(
-                spaces, std::move(_integrals), mesh, coefficients, constants,
-                needs_permutation_data, ptr_to_ref_wrapper_vec(entity_maps));
+                spaces, std::move(_integrals), std::move(mesh), coefficients,
```
garth-wells: `std::move` on `mesh` not needed.

python/dolfinx/wrappers/refinement.cpp
schnellerhase: The `Meshtags` argument (taken by value) is then move constructed, not copy constructed. https://godbolt.org/z/zKb3h46qj

cpp/dolfinx/mesh/Topology.cpp
schnellerhase: Initialiser lists are not movable. This removes a somewhat hidden copy of `vertex_map`.

python/dolfinx/wrappers/fem.cpp
schnellerhase: Moving `std::shared_ptr` removes the implied overhead by the atomic reference count increase and decrease associated with copying (and deleting) a `std::shared_ptr` - moving requires no reference count update.

garth-wells: `std::move` makes `V` an rvalue, but the `DirichletBC` constructor doesn't accept rvalues for the `FunctionSpace` argument (it takes it by value), which is why it may have no effect unless the compiler can do some copy elision.

Even if it did move the shared pointer, the cost of copying a shared pointer outside a hot loop is trivial. We shouldn't be making code syntactically more complex if there is no measurable performance benefit. Happy to reconsider if a realistic test can demonstrate a meaningful performance difference.

PR #3821, opened 2025-07-22
Use cache for Spack CI

garth-wells:
Build cache for dependencies means tests run faster. 

- Also splits C++ and Python builds, which exposed some issues.
- Split into multiple jobs.


jhale:
Now spack tooling has hit v1 we should probably use stable rather than develop. The package index is now a separate repo so they can be moved independently.

garth-wells:
> Now spack tooling has hit v1 we should probably use stable rather than develop. The package index is now a separate repo so they can be moved independently.

Will merge now because I have another CI branching depending on this. Can then test against release.

PR #3823, opened 2025-07-23
Rationalise and simplify some CI by using Spack

garth-wells:
This PR combines Redhat/Rocky and openAPI CI in a single action (separate jobs).

- Dependencies are installed using Spack. This is nice because the testing is self-contained in one file (no custom Docker image) we have full control over dependency versions.
- Spack caching to Github Packages makes installing the dependencies fast.
- Removes our somewhat odd testing of oneAPI via conda.
- Removes the Rocky Dockerfile.
- Can be trivially extended to test with more compilers.
- New versions of dependencies can be easily tested in branches. No messing with Docker image builds.


jhale:
This makes sense but the Rocky Dockerfile and build also provides very clean set of working instructions for a source install on Unix systems. It would be nice to keep something like that.

jhale:
I was never a fan of building against Anaconda OneAPI, glad to see that go.

Issue #3828, opened 2025-07-25
[BUG]: XDMFFile time series unusable


michalhabera:
### Summarize the issue

Some recent changes caused function name from XDMFFile in ParaView always always contain "/"+time at the end, which makes it look like different function. This makes time series playback in PV not possible.

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/9f1a49fd-23cc-4704-979d-174596e64edf" />

### Minimal Example (Python)

```Python
import dolfinx
from mpi4py import MPI

mesh = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, 10, 10)
V = dolfinx.fem.functionspace(mesh, ("CG", 1))
u = dolfinx.fem.Function(V, name="u")

with dolfinx.io.XDMFFile(MPI.COMM_WORLD, "u.xdmf", "w") as ofile:
    ofile.write_mesh(mesh)
    for t in range(10):
        ofile.write_function(u, float(t))
```

### Output (Python)

```bash

```

### Version

main branch

schnellerhase:
Changing https://github.com/FEniCS/dolfinx/blob/e180dbdaedcdfe180dc452ba141e337c47d58068/cpp/dolfinx/io/xdmf_function.cpp#L175 to `std::string dataset_name = "/Function/" + attr_name;` fixes `ASCII`, not `HDF5`.

PR #3829, opened 2025-07-25
Fix XDMF time series

schnellerhase:
Time info in function name should not be part of the attributes, only the data item's name needs to be differentiated between the different time steps by name.

This caused time step data for a function `u` to not be named `u/0`, `u/1`, ... - which broke paraviews time step visualisation. 

Fixes https://github.com/FEniCS/dolfinx/issues/3828

### DIFF
--- a/cpp/dolfinx/io/xdmf_function.cpp
+++ b/cpp/dolfinx/io/xdmf_function.cpp
@@ -7,6 +7,7 @@
 #include "xdmf_function.h"
 #include "xdmf_mesh.h"
 #include "xdmf_utils.h"
+#include <algorithm>
 #include <basix/mdspan.hpp>
 #include <boost/lexical_cast.hpp>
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/io/xdmf_function.cpp
```diff
@@ -7,6 +7,7 @@
 #include "xdmf_function.h"
 #include "xdmf_mesh.h"
 #include "xdmf_utils.h"
+#include <algorithm>
```
michalhabera, line 10: Used?

schnellerhase: Yes, `std::replace` is part of it https://en.cppreference.com/w/cpp/algorithm/replace.html.

michalhabera (approved):

PR #3830, opened 2025-07-26
Global dual graph computation for branching meshes

schnellerhase:
Fixes https://github.com/FEniCS/dolfinx/issues/3733.

### DIFF
--- a/cpp/dolfinx/mesh/graphbuild.cpp
+++ b/cpp/dolfinx/mesh/graphbuild.cpp
@@ -404,7 +404,7 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
 
   std::vector<int> send_mf_sendcounts(num_items_recv.size(), 0);
   int index = 0;
-  for (int i = 0; i < num_items_recv.size(); i++)
+  for (std::size_t i = 0; i < num_items_recv.size(); i++)
   {
     for (int j = 0; j < num_items_recv[i]; j++)
     {
@@ -440,7 +440,7 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
 
   std::vector<int> recv_mf_sendcounts(num_items_per_dest.size(), 0);
   index = 0;
-  for (int i = 0; i < num_items_per_dest.size(); i++)
+  for (std::size_t i = 0; i < num_items_per_dest.size(); i++)
   {
     for (int j = 0; j < num_items_per_dest[i]; j++)
     {
@@ -547,7 +547,7 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
     {
       std::size_t pos = send_indx_to_pos[i];
       std::size_t cell = cells[pos];
-      for (std::size_t j = 0; j < recv_matched_facet_counts[i]; j++)
+      for (int j = 0; j < recv_matched_facet_counts[i]; j++)
       {
         data[disp[cell]++] = recv_matched_facets[offset + j];
       }
--- a/cpp/dolfinx/mesh/graphbuild.cpp
+++ b/cpp/dolfinx/mesh/graphbuild.cpp
@@ -459,11 +459,11 @@ mesh::build_local_dual_graph(
     graph::AdjacencyList<std::int32_t> cell_facets
         = mesh::get_entity_vertices(cell_type, tdim - 1);
 
-    for (std::int32_t i = 0; i < cell_facets.num_nodes(); ++i)
-    {
-      max_vertices_per_facet
-          = std::max(max_vertices_per_facet, cell_facets.num_links(i));
-    }
+    // Determine/update maximum number of vertices for facet
+    std::ranges::for_each(
+        std::views::iota(0, cell_facets.num_nodes()),
+        [&max = max_vertices_per_facet, &cell_facets](auto node)
+        { max = std::max(max, cell_facets.num_links(node)); });
   }
 
   // 2) Build a list of (all) facets, defined by sorted vertices, with
--- a/cpp/dolfinx/mesh/graphbuild.cpp
+++ b/cpp/dolfinx/mesh/graphbuild.cpp
@@ -12,7 +12,10 @@
 #include <dolfinx/common/log.h>
 #include <dolfinx/common/sort.h>
 #include <dolfinx/graph/AdjacencyList.h>
+#include <mpi.h>
+#include <numeric>
 #include <optional>
+#include <ranges>
 #include <span>
 #include <utility>
 #include <vector>
@@ -89,7 +92,7 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
   //      to dual graph
   //    - prepare info of matched facet for recipients
   //  d) return info of remotely matched/unmatched facets from PO
-  //    - contruct locally the paralel aware dual graph (with ghost edges).
+  //    - construct locally the parallel aware dual graph (with ghost edges).
 
   assert(local_max_vertices_per_facet == 0
          or facets.size() % local_max_vertices_per_facet == 0);
@@ -194,8 +197,8 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
 
       // Set entry in map from local facet row index (position) to local
       // destination rank
-      for (auto e = it; e != it1; ++e)
-        pos_to_neigh_rank[(*e)[1]] = neigh_rank;
+      for (auto& e : std::ranges::subrange(it, it1))
+        pos_to_neigh_rank[e[1]] = neigh_rank;
 
       // Advance iterator
       it = it1;
@@ -283,16 +286,14 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
     // Compute sort permutation for received data
     std::vector<int> sort_order(recv_buffer.size() / buffer_shape1);
     std::iota(sort_order.begin(), sort_order.end(), 0);
-    std::ranges::sort(
-        sort_order,
-        [&recv_buffer, buffer_shape1, max_vertices_per_facet](auto f0, auto f1)
-        {
-          auto it0 = std::next(recv_buffer.begin(), f0 * buffer_shape1);
-          auto it1 = std::next(recv_buffer.begin(), f1 * buffer_shape1);
-          return std::lexicographical_compare(
-              it0, std::next(it0, max_vertices_per_facet), it1,
-              std::next(it1, max_vertices_per_facet));
-        });
+    std::ranges::sort(sort_order, std::ranges::lexicographical_compare,
+                      [&](auto f)
+                      {
+                        auto begin
+                            = std::next(recv_buffer.begin(), f * buffer_shape1);
+                        return std::ranges::subrange(
+                            begin, std::next(begin, max_vertices_per_facet));
+                      });
 
     for (auto it = sort_order.begin(); it != sort_order.end();)
     {
@@ -316,10 +317,12 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
       if (std::size_t num_matches = std::distance(it, it1); num_matches == 2)
       {
         // Store the global cell index from the other rank
-        send_buffer1[*it]
-            = recv_buffer[*(it + 1) * buffer_shape1 + max_vertices_per_facet];
-        send_buffer1[*(it + 1)]
-            = recv_buffer[*it * buffer_shape1 + max_vertices_per_facet];
+        int facet = *it;
+        int next_facet = *(it + 1);
+        send_buffer1[facet]
+            = recv_buffer[next_facet * buffer_shape1 + max_vertices_per_facet];
+        send_buffer1[next_facet]
+            = recv_buffer[facet * buffer_shape1 + max_vertices_per_facet];
       }
       else if (num_matches > 2)
       {
@@ -456,7 +459,6 @@ mesh::build_local_dual_graph(
     graph::AdjacencyList<std::int32_t> cell_facets
         = mesh::get_entity_vertices(cell_type, tdim - 1);
 
-    // Determine maximum number of vertices for facet
     for (std::int32_t i = 0; i < cell_facets.num_nodes(); ++i)
     {
       max_vertices_per_facet
@@ -510,17 +512,16 @@ mesh::build_local_dual_graph(
   // 3) Sort facets by vertex key
   std::vector<std::size_t> perm(facets.size() / shape1, 0);
   std::iota(perm.begin(), perm.end(), 0);
-  // TODO: radix_sort?
-  std::sort(perm.begin(), perm.end(),
-            [&facets, shape1](auto f0, auto f1)
-            {
-              auto it0 = std::next(facets.begin(), f0 * shape1);
-              auto it1 = std::next(facets.begin(), f1 * shape1);
-              return std::lexicographical_compare(it0, std::next(it0, shape1),
-                                                  it1, std::next(it1, shape1));
-            });
+  // TODO: radix_sort? This is a heavy sort call.
+  std::ranges::sort(perm, std::ranges::lexicographical_compare,
+                    [&](auto f)
+                    {
+                      auto begin = std::next(facets.begin(), f * shape1);
+                      return std::ranges::subrange(begin,
+                                                   std::next(begin, shape1));
+                    });
 
-  // 4) Iterate over sorted list of facets. Facets shared by more than
+  // // 4) Iterate over sorted list of facets. Facets shared by more than
   //    one cell lead to a graph edge to be added. Facets that are not
   //    shared are stored as these might be shared by a cell on another
   //    process.
@@ -528,23 +529,26 @@ mesh::build_local_dual_graph(
   std::vector<std::int32_t> local_cells;
   std::vector<std::array<std::int32_t, 2>> edges;
   {
-    auto it = perm.begin();
-    while (it != perm.end())
+    for (auto it = perm.begin(); it != perm.end();)
     {
       std::size_t facet_index = *it;
       std::span facet(facets.data() + facet_index * shape1, shape1);
 
       // Find iterator to next facet different from f0 -> all facets in
       // [it, it_next_facet) describe the same facet
-      auto it_next_facet = std::find_if_not(
-          it, perm.end(),
-          [facet, &facets, shape1](auto idx) -> bool
-          {
-            auto f1_it = std::next(facets.begin(), idx * shape1);
-            return std::equal(facet.begin(), std::prev(facet.end()), f1_it);
-          });
+      // TODO: use a subrange here
+      auto matching_facets = std::ranges::subrange(
+          it, std::find_if_not(it, perm.end(),
+                               [facet, &facets, shape1](auto idx) -> bool
+                               {
+                                 auto f1_it
+                                     = std::next(facets.begin(), idx * shape1);
+                                 return std::equal(facet.begin(),
+                                                   std::prev(facet.end()),
+                                                   f1_it);
+                               }));
 
-      std::int32_t cell_count = std::distance(it, it_next_facet);
+      std::int32_t cell_count = matching_facets.size();
       assert(cell_count >= 1);
       if (!max_facet_to_cell_links.has_value()
           or (cell_count < *max_facet_to_cell_links))
@@ -562,12 +566,13 @@ mesh::build_local_dual_graph(
       // Add dual graph edges (one direction only, other direction is
       // added later). In the range [it, it_next_facet), all
       // combinations are added.
-      for (auto facet_a_it = it; facet_a_it != it_next_facet; facet_a_it++)
+      for (auto facet_a_it = it; facet_a_it != matching_facets.end();
+           facet_a_it++)
       {
         std::span facet_a(facets.data() + *facet_a_it * shape1, shape1);
         std::int32_t cell_a = facet_a.back();
         for (auto facet_b_it = std::next(facet_a_it);
-             facet_b_it != it_next_facet; facet_b_it++)
+             facet_b_it != matching_facets.end(); facet_b_it++)
         {
           std::span facet_b(facets.data() + *facet_b_it * shape1, shape1);
           std::int32_t cell_b = facet_b.back();
@@ -576,7 +581,7 @@ mesh::build_local_dual_graph(
       }
 
       // Update iterator
-      it = it_next_facet;
+      it = matching_facets.end();
     }
   }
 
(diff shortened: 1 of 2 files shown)


cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -12,7 +12,11 @@
 #include <dolfinx/common/log.h>
 #include <dolfinx/common/sort.h>
 #include <dolfinx/graph/AdjacencyList.h>
+#include <iostream>
```
jorgensd: Can probably remove iostream (used for debugging? :) )

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -276,98 +285,158 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
   MPI_Type_free(&compound_type);
   MPI_Comm_free(&comm_po_post);
 
+  // TODO: update
```
jorgensd: Is this a future todo or a  current todo? As the code ha sbeen altered below, it's hard to tell if this should still be there.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -79,17 +83,22 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
...
   //    - dolfinx::MPI::index_owner deduces ownership
-  //  b) every facet is send to owning PO
-  //    - data for facet i: list of vertices + associated global cell idx
-  //  c) check (on PO) if multiple same facets have been received
-  //    - if so, found matched facet across process boundary -> introduce edge
-  //      to dual graph
-  //    - prepare info of matched facet for recipients
-  //  d) return info of remotely matched/unmatched facets from PO
-  //    - contruct locally the paralel aware dual graph (with ghost edges).
+  //  b) every unmatched facet is send to owning PO
```
garth-wells, line 88: Need to define or spell out 'PO'.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -276,98 +285,158 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
   MPI_Type_free(&compound_type);
   MPI_Comm_free(&comm_po_post);
 
+  // TODO: update
   // Search for consecutive facets (-> dual graph edge between cells)
   // and pack into send buffer
-  std::vector<std::int64_t> send_buffer1(recv_disp.back(), -1);
+  std::vector<std::vector<std::int64_t>> matched_facets(recv_disp.back());
```
garth-wells: How large can `recv_disp.back()` be?

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -276,98 +285,158 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
...
-        [&recv_buffer, buffer_shape1, max_vertices_per_facet](auto f0, auto f1)
-        {
-          auto it0 = std::next(recv_buffer.begin(), f0 * buffer_shape1);
-          auto it1 = std::next(recv_buffer.begin(), f1 * buffer_shape1);
-          return std::lexicographical_compare(
-              it0, std::next(it0, max_vertices_per_facet), it1,
-              std::next(it1, max_vertices_per_facet));
-        });
+    std::ranges::sort(sort_order, std::ranges::lexicographical_compare,
+                      [&](auto f)
```
garth-wells: Avoid `[&]` - it breaks encapsulation more than is needed and it is no longer obvious what the function depends on.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -276,98 +285,158 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
...
+      // backedge).
+      for (auto facet_a_it = matching_facets.begin();
+           facet_a_it != matching_facets.end(); facet_a_it++)
       {
-        throw std::runtime_error(
-            "A facet is connected to more than two cells.");
+        for (auto facet_b_it = std::next(facet_a_it);
+             facet_b_it != matching_facets.end(); facet_b_it++)
+        {
+          int facet_a = *facet_a_it;
```
jorgensd: Add a comment here reminding the reader that the cell was packed as an additional entry at the end of each facet sequence.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -276,98 +285,158 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
...
+    std::partial_sum(send_mf_sendcounts.begin(), send_mf_sendcounts.end(),
+                     std::next(send_mf_displs.begin()));
+  }
 
-  // Count number of adjacency list edges
-  std::vector<std::int32_t> num_edges(local_dual_graph.num_nodes(), 0);
-  std::adjacent_difference(std::next(local_dual_graph.offsets().begin()),
-                           local_dual_graph.offsets().end(), num_edges.begin());
-  for (std::size_t i = 0; i < recv_buffer1.size(); ++i)
+  // Compute matched facet receive counts and displacements.
```
jorgensd, line 391: Should the comment be moved to after the MPI_Wait three lines down, as it relates to the for-loop, rather than the assignment of sendcounts?

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -383,16 +452,31 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
                              [cell_offset](auto x) { return x + cell_offset; });
     }
 
-    // Add non-local data
-    for (std::size_t i = 0; i < recv_buffer1.size(); ++i)
+    int offset = 0;
```
jorgensd, line 478: Add back comment of adding non-local data?

python/test/unit/mesh/test_branching_manifold.py
```diff
@@ -0,0 +1,144 @@
...
+    skeleton_top = skeleton_mesh.topology
+
+    if comm.size > 1:
+        pytest.skip("Branching mesh c->e, e->v connectivity not yet implemented.")
+
+    skeleton_top.create_connectivity(1, 2)
+    skeleton_f_to_c = skeleton_top.connectivity(1, 2)
+
+    # import febug
+    # febug.plot_entity_indices(mesh, 2).show() # .save_graphic(f"test_{comm.rank}.svg")
```
jorgensd: Remove once ready for merge

python/test/unit/mesh/test_branching_manifold.py
```diff
@@ -0,0 +1,144 @@
...
+        create_cell_partitioner(GhostMode.shared_facet, max_facet_to_cell_links),
+    )
+
+    skeleton_top = skeleton_mesh.topology
+    skeleton_top.create_connectivity(0, 1)
+    skeleton_f_to_c = skeleton_top.connectivity(0, 1)
+
+    # debug ouput
+    # import febug
+    # febug.plot_entity_indices(skeleton_mesh, 1).show() # .save_graphic(f"test_{comm.rank}.svg")
```
jorgensd: Remove when ready for merge

cpp/dolfinx/mesh/graphbuild.cpp
schnellerhase: Depends on matching condition. If `max_facet_to_cell_links==2` this corresponds to something of order number of local boundary facets, for larger values, up to order of local facet count.

python/test/unit/geometry/test_bounding_box_tree.py
```diff
@@ -421,9 +421,13 @@ def test_sub_bbtree_codim1(dtype):
...
     cells = compute_collisions_points(bbtree, point)
-    if MPI.COMM_WORLD.rank in ranks.array:
-        assert len(cells.links(0)) > 0
-    else:
+
+    # Relationship to test: if cells has link => ranks must also hold a value.
+    if len(cells.links(0)) > 0:
+        assert len(ranks.links(0)) > 0
+
+    # Negated: ranks holds no value => cells has no link
```
schnellerhase, line 430: Not entirely clear why this test case breaks in its previous form. Checked the dual graphs, attached, which up to reorder are equivalent (compared to `main`).
[dual_graph_bb_test_main.txt](https://github.com/user-attachments/files/21666361/dual_graph_bb_test_main.txt)
[dual_graph_bb_test_global_dual_graph.txt](https://github.com/user-attachments/files/21666362/dual_graph_bb_test_global_dual_graph.txt)

cpp/dolfinx/mesh/graphbuild.cpp
garth-wells: That's a performance problem. `std::vector<std::vector<...>>` can be convenient, but the performance hit can be big when the size of the vector is large. Each entry is a dynamic memory allocation.

Can you return a flat data array +offset array instead?

schnellerhase: Should be possible, by first computing the offsets and then assembling data. The later on send data structure is pretty much the flattened layout anyways. So we would also save the additional packing before sending.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -278,96 +286,155 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
 
   // Search for consecutive facets (-> dual graph edge between cells)
   // and pack into send buffer
-  std::vector<std::int64_t> send_buffer1(recv_disp.back(), -1);
+  std::vector<std::vector<std::int64_t>> matched_facets(recv_disp.back());
```
schnellerhase: Change variable naming to setting of dual graph from here onwards: no more facets and matching but edges in dual graph - PO does not send back facets, but edges.

cpp/dolfinx/mesh/utils.h
```diff
@@ -947,10 +948,10 @@ entities_to_geometry(const Mesh<T>& mesh, int dim,
 /// cells on this rank by applying the default graph partitioner to the
 /// dual graph of the mesh.
 /// @return Function that computes the destination ranks for each cell.
-CellPartitionFunction create_cell_partitioner(mesh::GhostMode ghost_mode
-                                              = mesh::GhostMode::none,
-                                              const graph::partition_fn& partfn
-                                              = &graph::partition_graph);
+CellPartitionFunction create_cell_partitioner(
```
garth-wells, line 957: Document the function args, especially the new one.

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -779,12 +779,16 @@ void mesh(nb::module_& m)
...
+         std::optional<std::int32_t> max_facet_to_cell_links)
           -> part::impl::PythonCellPartitionFunction
       {
         return part::impl::create_cell_partitioner_py(
-            dolfinx::mesh::create_cell_partitioner(mode));
+            dolfinx::mesh::create_cell_partitioner(
+                mode, &dolfinx::graph::partition_graph,
+                max_facet_to_cell_links));
       },
+      nb::arg("mode"), nb::arg("max_facet_to_cell_links") = 2,
```
garth-wells, line 791: Is this function wrapped in Python? We tend to avoid default args in the nanobind layer.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -277,59 +285,92 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
...
 
-      // Advance iterator and increment entity
-      it = it1;
-    }
+    // Iterate matching facets to compute count/offset information of dual edges
+    for_each_matched_pair(
+        [&dedge_send_count](int facet_a, std::int64_t /* cell_a */, int facet_b,
+                            std::int64_t /* cell_b */)
+        {
+          dedge_send_count[facet_a]++;
```
garth-wells: `++dedge_send_count[facet_a]` is clearer - the operation appears first.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -277,59 +285,92 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
...
+          dedge_send_count[facet_b]++;
+        });
+
+    std::partial_sum(dedge_send_count.begin(), dedge_send_count.end(),
+                     std::next(dedge_send_displs.begin()));
+
+    auto send_dual_edges_size
```
garth-wells: Don't use `auto`, be explicit with the type.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -277,59 +285,92 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
...
+        = std::accumulate(dedge_send_count.begin(), dedge_send_count.end(), 0);
+    dedge_send_data.resize(send_dual_edges_size);
+
+    // Iterate matching facets to store dual edge connectivity
+    auto offset = dedge_send_displs;
```
garth-wells: Don't use auto, be explicit with the type.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -384,15 +476,31 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
...
-      {
-        std::size_t pos = send_indx_to_pos[i];
-        std::size_t cell = cells[pos];
-        data[disp[cell]++] = recv_buffer1[i];
-      }
+      std::int32_t cell_idx = send_indx_to_pos[i];
+      std::int32_t cell = cells[cell_idx];
+
+      for (int j = 0; j < dedge_recv_count[i]; j++)
+        data[disp[cell]++] = recv_dual_edges[offset + j];
```
garth-wells: This isn't super easy to read. Maybe break into more the one line?

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -384,15 +476,31 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
...
+      auto links
+          = std::ranges::subrange(std::next(data.begin(), offsets[node]),
+                                  std::next(data.begin(), offsets[node + 1]));
+      std::ranges::sort(links);
+      auto duplicate_links = std::ranges::unique(links);
+      data.erase(duplicate_links.begin(), duplicate_links.end());
+      for (std::size_t i = node + 1; i < offsets.size(); i++)
+        offsets[i] -= std::ranges::size(duplicate_links);
+    }
+    // TOOD: shrink_to_fit here?
```
garth-wells: Resolve this.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -510,41 +617,42 @@ mesh::build_local_dual_graph(
...
-              auto it0 = std::next(facets.begin(), f0 * shape1);
-              auto it1 = std::next(facets.begin(), f1 * shape1);
-              return std::lexicographical_compare(it0, std::next(it0, shape1),
-                                                  it1, std::next(it1, shape1));
-            });
-
-  // 4) Iterate over sorted list of facets. Facets shared by more than
+  // TODO: radix_sort? This is a heavy sort call.
+  std::ranges::sort(perm, std::ranges::lexicographical_compare,
+                    [&](auto f)
```
garth-wells: Avoid capturing everything with `[&]` - limited scope is best.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -585,24 +694,24 @@ mesh::build_local_dual_graph(
...
   std::vector<std::int32_t> data(offsets.back());
-  {
-    std::vector<std::int32_t> pos = offsets;
-    for (auto e : edges)
-    {
-      data[pos[e[0]]++] = e[1];
-      data[pos[e[1]]++] = e[0];
-    }
-  }
+  std::ranges::for_each(edges,
```
garth-wells, line 717: I think this is less readable than the code is replaces.

garth-wells (changes requested):
Any performance implications for 'standard' meshes?

python/dolfinx/wrappers/mesh.cpp
schnellerhase: Not at the moment.

schnellerhase:
> Any performance implications for 'standard' meshes?

I see two changes with possible performance impact:
1. Change from one to two neighbourhood communications. 
a) Size communication with fixed size neighbourhood all-to-all (non-blocking)
b) Data communication (this improved in terms of data amount as we no longer send unmatched facet information)
2. Removal of possible duplicate edges during assembly of global dual graph (post-communications) - requires sort + remove on very small vectors. (If this becomes a measurable performance hit, one can deactivate this for the non-branching case).

schnellerhase:
https://github.com/schnellerhase/fenics-dolfinx/pull/24 with `np=64` on `main`.

```
Benchmark                               Time             CPU   Iterations
-------------------------------------------------------------------------
BM_create_box/10/iterations:5        43.5 ms         43.5 ms            5
BM_create_box/50/iterations:5        62.7 ms         62.7 ms            5
BM_create_box/100/iterations:5        579 ms          579 ms            5
BM_create_box/200/iterations:5       5895 ms         5880 ms            5
BM_create_box/300/iterations:5      22676 ms        22618 ms            5
```

schnellerhase:
https://github.com/schnellerhase/fenics-dolfinx/pull/24 with `np=64` on `schnellerhase/global-dual-graph`.

```
-------------------------------------------------------------------------
Benchmark                               Time             CPU   Iterations
-------------------------------------------------------------------------
BM_create_box/10/iterations:5        46.1 ms         45.9 ms            5
BM_create_box/50/iterations:5        74.2 ms         74.1 ms            5
BM_create_box/100/iterations:5       1172 ms         1171 ms            5
BM_create_box/200/iterations:5      47608 ms        47577 ms            5

```
negligible performance hit 😅

garth-wells (approved):

schnellerhase:
That was a bad joke, there is something really wrong here - need to investigate. If we would see a hit I was expecting it to be most prominent for small meshes, but it is the opposite.

garth-wells (changes requested):
Undoing approval until performance issues resolved.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -384,14 +476,33 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
...
+        std::int32_t _cell_offset = disp[cell]++;
+        std::int64_t node = recv_dual_edges[offset + j];
+        data[_cell_offset] = node;
       }
+
+      offset += dedge_recv_count[i];
+    }
+    // local connections are possibly introduced again by remote -> remove
+    // duplicates
+    for (std::size_t node = 0; node < offsets.size() - 1; node++)
```
schnellerhase, line 496: Stalls in this duplicate removal loop - not in communications.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -500,6 +502,9 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
                                   std::next(data.begin(), offsets[node + 1]));
       std::ranges::sort(links);
       auto duplicate_links = std::ranges::unique(links);
+      if (duplicate_links.size() == 0)
```
jorgensd: Could this be changed to `duplicate_links.empty()`? Or does it not have that option?

chrisrichardson:
How does it reduce the amount of communication in the setup phase?

jorgensd:
> One such example would be the mesh in the PR description - where `max_facet_to_cell_links=4` will allow to rule out locally interior facets prior to communicating them. They are never transferred to the post office for matching.
> 
> I agree that this is is not as general as one would like to have it, as it requires some very detailed a-priori knowledge about the mesh. However, we can extend this freely in the future to a facet wise input from the user.

For 1D grids/graphs one can easily get this information through software such as networkx, and then the input will be very useful indeed

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -277,59 +285,92 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
   MPI_Comm_free(&comm_po_post);
 
   // Search for consecutive facets (-> dual graph edge between cells)
-  // and pack into send buffer
-  std::vector<std::int64_t> send_buffer1(recv_disp.back(), -1);
+  // and pack into send buffer. We store for every cell the number of matches,
+  // the offsets of each cell and the continuous data.
+  // Note: deges is short dual dual edges.
```
schnellerhase:
```suggestion
  // Note: deges is short for dual edges.
```

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -277,59 +285,92 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
...
+        });
+
+    std::partial_sum(dedge_send_count.begin(), dedge_send_count.end(),
+                     std::next(dedge_send_displs.begin()));
+
+    std::int32_t send_dual_edges_size
+        = std::accumulate(dedge_send_count.begin(), dedge_send_count.end(), 0);
+    dedge_send_data.resize(send_dual_edges_size);
+
+    // Iterate matching facets to store dual edge connectivity
```
schnellerhase:
```suggestion
    // Iterate matching facets to store dual edges
```

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -339,35 +380,86 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
...
 
-  // Send back data
-  std::vector<std::int64_t> recv_buffer1(send_disp.back());
-  MPI_Neighbor_alltoallv(send_buffer1.data(), num_items_recv.data(),
-                         recv_disp.data(), dolfinx::MPI::mpi_t<std::int64_t>,
-                         recv_buffer1.data(), num_items_per_dest.data(),
-                         send_disp.data(), dolfinx::MPI::mpi_t<std::int64_t>,
-                         comm_po_receive);
-  MPI_Comm_free(&comm_po_receive);
+  // Send po->recipient matched cell counts (non-blocking)
```
schnellerhase:
```suggestion
  // Send PO->recipient: matched cell counts (non-blocking)
```

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -384,14 +476,36 @@ graph::AdjacencyList<std::int64_t> compute_nonlocal_dual_graph(
...
+      auto links
+          = std::ranges::subrange(std::next(data.begin(), offsets[node]),
+                                  std::next(data.begin(), offsets[node + 1]));
+      std::ranges::sort(links);
+      auto duplicate_links = std::ranges::unique(links);
+      if (duplicate_links.empty())
+        continue;
+
+      data.erase(duplicate_links.begin(), duplicate_links.end());
+      for (std::size_t i = node + 1; i < offsets.size(); i++)
```
schnellerhase: Remove quadratic complexity - replace with accumulated duplicate count and update of current `node`.

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -510,41 +623,42 @@ mesh::build_local_dual_graph(
...
-            [&facets, shape1](auto f0, auto f1)
-            {
-              auto it0 = std::next(facets.begin(), f0 * shape1);
-              auto it1 = std::next(facets.begin(), f1 * shape1);
-              return std::lexicographical_compare(it0, std::next(it0, shape1),
-                                                  it1, std::next(it1, shape1));
-            });
-
-  // 4) Iterate over sorted list of facets. Facets shared by more than
+  // TODO: radix_sort? This is a heavy sort call.
```
schnellerhase:
```suggestion
```
Not possible (compared elements are spans, not ints).

chrisrichardson:
This line looks wrong...

https://github.com/FEniCS/dolfinx/blob/1a13864509499e48a4dafd8cb1d27fa16ccfda8f/cpp/dolfinx/mesh/graphbuild.cpp#L138

garth-wells (approved):

PR #3833, opened 2025-07-29
Add default scalar type cast

schnellerhase:
Fixes https://github.com/FEniCS/ufl/issues/399.

Caused by default scalar type being complex in the ufl dolfinx integration test. In this case form, function and coefficients are constructed from default type (complex), but the boundary condition from a float value. This caused a mismatch.

Tested at https://github.com/FEniCS/ufl/actions/runs/16607427384/job/46982707793

### DIFF
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -98,9 +98,9 @@ class Form:
         return self._cpp_object.function_spaces
 
     @property
-    def dtype(self) -> np.floating:
+    def dtype(self) -> np.dtype:
         """Scalar type of this form."""
-        return np.dtype(self._cpp_object.dtype).type
+        return np.dtype(self._cpp_object.dtype)
 
     @property
     def mesh(self) -> typing.Union[_cpp.mesh.Mesh_float32, _cpp.mesh.Mesh_float64]:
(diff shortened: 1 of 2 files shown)


garth-wells:
Maybe `fem.dirichletbc` should be handling the cast. It's pretty natural to cast a float to complex.

jorgensd:
> So, changing from extracting the `dtype` from the value passed to `dirichletbc`, to instead extracting it from the function space (on which we want to apply it) and casting the value (array) to its `dtype`?

The function space doesn’t know the `dtype`?  It is only the function we are applying it to that knows the dtype.

jorgensd:
I guess this goes back to `Constant` not having a default dtype. I raised this a long time ago, and here is the discussion: https://github.com/FEniCS/dolfinx/pull/1997

schnellerhase:
`ufl` CI rerun https://github.com/FEniCS/ufl/actions/runs/16607427384/job/47052249727.

jorgensd:
> > Ah right, function space only knows the geometry `dtype`. Then I don't know what to change it to.
> 
> I'd suggest (i) moving `bc` closer to where it's used, and (ii) getting the `dtype` from the `Form` it acts on. `default_scalar_type` is something a bit ugly we need to live with but should use as little as possible.

That has to be slightly adapted to work with `bc.set`:
https://github.com/FEniCS/dolfinx/blob/d1caaf2d6bf0f0d858b1392aea2d81cc7cea0000/python/dolfinx/fem/bcs.py#L134

PR #3834, opened 2025-07-30
Correct several typos in Stokes example and improve documentation formatting

jorgensd:


python/demo/demo_stokes.py
```diff
@@ -74,18 +74,20 @@
 # ## Implementation
 #
 # The Stokes problem using Taylor-Hood elements is solved using:
-# 1. [Block preconditioner using PETSc MatNest and VecNest data
-#    structures. Each 'block' is a standalone object.]
-#    (#nested-matrix-solver)
+# 1. [Block preconditioner using PETSc Nest data structures using
```
garth-wells, line 77: "using PETSc Nest data structures using the high-level interface" reads strangely.

Also not clear what the "high-level interface" is.

python/demo/demo_stokes.py
```diff
@@ -636,23 +641,28 @@ def mixed_direct():
 
 
 # Solve using LinearProblem class
+
```
garth-wells, line 645: I don't think the extra line breaks help. Association with code is clearer without.

garth-wells (changes requested):

python/demo/demo_stokes.py
jorgensd: They do when you look at the web-page, as it currently looks as:
image
while with the linebreaks it looks like
image

python/demo/demo_stokes.py
jorgensd: Do you have a better suggestion? Currently the link is completely broken
image

Issue #3593, opened 2025-01-12
[BUG]: (Fixed in this post) The "eps" function fails when the "Omega" functions are tweaked in the tutorial "Electromagnetic modal analysis for a waveguide"


FyDininno:
### Summarize the issue

Instead of using "Omega" functions to assign the values to eps, what you have to do is write an "eps_expr(x)" function which returns the scalar value of eps over the domain at each x. Next, you have to run "eps.interpolate(eps_expr)" to correctly assign the dielectric values to the eps function. This way, it does not break. 

### How to reproduce the bug

To reproduce the bug, use these omega functions instead of the ones given in the tutorial, and set all of the dielectric permittivities to 1 just to illustrate that the bug arises from the assignment of eps, and not the values of the permittivities. As you can see, simply by making the dielectric domain two dimensional instead of one dimensional completely breaks the simulation, giving nonsensical results.

### Minimal Example (Python)

```Python
eps_v = 1
eps_d = 1

def Omega_d(x): # Returns True within the correct domain, false outside the domain
    return (x[0] <= w/2) & (x[1] <= h/2)

def Omega_v(x):
    return (x[0] >= w/2) | (x[1] >= h/2)

D = fem.functionspace(msh, ("DQ", 0)) # This defines the FEM function space over the mesh with linear (degree 0) discontinuous functions defined on quadilateral elements
eps = fem.Function(D) # A template function created by elements within D, as of now, all the basis functions have a coefficient of zero

cells_v = locate_entities(msh, msh.topology.dim, Omega_v)
cells_d = locate_entities(msh, msh.topology.dim, Omega_d) # locates the cells which have the correct truth value

eps.x.array[cells_v] = np.full_like(cells_v, eps_v, dtype=scalar_type) # .x gives acces to the DOF vector for eps. The argument specifies the cell values that are being modified. The numpy array formats an array to the likes of the argument with scalar values of the correct dielectric permittivities in their place.
eps.x.array[cells_d] = np.full_like(cells_d, eps_d, dtype=scalar_type)
```

### Version

main branch

### Installation

I used conda and set the vscode environment to use the fenicsx conda environment.

### Additional information

If you simply add
```
def eps_expr(x):
  cond_d = Omega_d(x)
return np.where(cond_d, eps_d, eps_v)

D = fem.functionspace(msh, ("CG", 1))
eps = fem.Function(D)
eps.interpolate(eps_expr)
```
It should work

PR #3843, opened 2025-08-01
Use interpolate in half loaded waveguide demo

schnellerhase:
The cell based assignment breaks if the conditional domains do not align with the underlying mesh. Caused by `locate_entites` only returning cells for which all vertices fulfil condition. This breaks for cells crossing the subdomain boundary. Using interpolate fixes this.

Fixes https://github.com/FEniCS/dolfinx/issues/3593.

### DIFF
--- a/python/demo/demo_half_loaded_waveguide.py
+++ b/python/demo/demo_half_loaded_waveguide.py
@@ -187,7 +187,7 @@ eps_d = 2.45
 
 D = fem.functionspace(msh, ("DQ", 0))
 eps = fem.Function(D)
-eps.interpolate(lambda x: np.less_equal(x[1], d) * eps_d + np.greater_equal(x[1], d) * eps_v)
+eps.interpolate(lambda x: np.less(x[1], d) * eps_d + np.greater_equal(x[1], d) * eps_v)
 # -
 
 # In order to find the weak form of our problem, the starting point are
--- a/python/demo/demo_half_loaded_waveguide.py
+++ b/python/demo/demo_half_loaded_waveguide.py
@@ -49,7 +49,7 @@ import ufl
 from basix.ufl import element, mixed_element
 from dolfinx import fem, plot
 from dolfinx.fem.petsc import assemble_matrix
-from dolfinx.mesh import CellType, create_rectangle, exterior_facet_indices, locate_entities
+from dolfinx.mesh import CellType, create_rectangle, exterior_facet_indices
 
 try:
     import pyvista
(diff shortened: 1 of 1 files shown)


python/demo/demo_half_loaded_waveguide.py
```diff
@@ -185,23 +185,9 @@ def verify_mode(
...
-
 D = fem.functionspace(msh, ("DQ", 0))
 eps = fem.Function(D)
-
-cells_v = locate_entities(msh, msh.topology.dim, Omega_v)
-cells_d = locate_entities(msh, msh.topology.dim, Omega_d)
-
-eps.x.array[cells_d] = np.full_like(cells_d, eps_d, dtype=PETSc.ScalarType)
-eps.x.array[cells_v] = np.full_like(cells_v, eps_v, dtype=PETSc.ScalarType)
+eps.interpolate(lambda x: np.less(x[1], d) * eps_d + np.greater_equal(x[1], d) * eps_v)
```
nate-sime: Shouldn't this interpolation be enacted on each subdomain? Then you don't have to worry about precision issues, nonconforming bases, etc.

```suggestion
eps.interpolate(lambda x: np.full_like(x[0], eps_d), cells_d)
eps.interpolate(lambda x: np.full_like(x[0], eps_v), cells_v)
```

jorgensd: > Shouldn't this interpolation be enacted on each subdomain? Then you don't have to worry about precision issues, nonconforming bases, etc.

I kind of agree. I think a cleaner solution would be to add a tolerance to the "cells_d" and "cells_v" locater.

schnellerhase: `cells_d` and `cells_v` will only be an exact split of the domain's cells if the internal subdomain boundary aligns with edges of the mesh. If this is not given we will have cells which cross this boundary. Such cells are neither part of `cells_d` nor `cells_v` (since there is always one vertex not fulfilling the conditional). So if we have any construction relying on the cell information, we will always miss these boundary cells and not assign any value there at all.

schnellerhase: > I kind of agree. I think a cleaner solution would be to add a tolerance to the "cells_d" and "cells_v" locater.

A matches any vertex option for `locate_entites` would be great here.

jorgensd: Since this is "just" a demo, where we control the subdivision:
```python
msh = create_rectangle(MPI.COMM_WORLD, np.array([[0, 0], [w, h]]), (nx, ny), CellType.quadrilateral)
```
we could just ensure that `d` aligns with grid (maybe have an assert checking this).

An alternative would be to just make an array of all cells, and just mark `cells_v`, i.e.
```python
cell_map = msh.topology.index_map(msh.topology.dim)
num_cells_on_process = cell_map.size_local + cell_map.num_ghosts
eps.x.array[:] = eps_d # Initially assign eps_d to all cells
eps.x.array[cells_v] = eps_v # Assign eps_v to cells that satisfy criterion
```

```

schnellerhase: In the issue it came up when moving the subdomains. Since this works, I think we should support it in the demo. I am under the impression that both the current state and the resolution strategy are a custom reimplementation of `interpolate`. Why is `interpolate` not the correct approach here?

jorgensd: Interpolate would allow for any other finite element (say a continuous space), that doesn’t conform to the original function. Additionally, for DG-0 data, the above approach is way faster, as it does not involve:
1. computing interpolation coordinates
2. Evaluating a function at said coordinates.

PR #3846, opened 2025-08-11
Get nestis from A instead of P

jorgensd:
There is no reason the nestIS should be gotten from P, and not A, as there are situations where you do not supply a preconditioner form, where you would need the index sets.

The index sets for A and P should be identical.

See for instance: https://github.com/jorgensd/dolfinx-tutorial/pull/264

### DIFF
--- a/python/dolfinx/fem/petsc.py
+++ b/python/dolfinx/fem/petsc.py
@@ -896,10 +896,10 @@ class LinearProblem:
 
             opts.prefixPop()
 
-        if self.P_mat is not None and kind == "nest":
-            # Transfer nest IS on self.P_mat to PC of main KSP. This allows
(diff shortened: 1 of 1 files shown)


python/dolfinx/fem/petsc.py
```diff
@@ -896,10 +896,10 @@ def __init__(
 
             opts.prefixPop()
 
-        if self.P_mat is not None and kind == "nest":
-            # Transfer nest IS on self.P_mat to PC of main KSP. This allows
+        if kind == "nest":
+            # Transfer nest IS on self.A_mat to PC of main KSP. This allows
```
```suggestion
            # Transfer nest IS on self.A to PC of main KSP. This allows
```

jhale (approved):

PR #3847, opened 2025-08-12
Allow `dolfinx.fem.Constant` to take native float as value type

finsberg:
Currently mypy complains on the following code
```python
from mpi4py import MPI
import dolfinx
import numpy as np


x = dolfinx.fem.Constant(mesh, 1.0)
```
with the error
```
error: Argument 2 to "Constant" has incompatible type "float"; expected "ndarray[Any, Any] | Sequence[Any] | floating[Any] | complexfloating[Any, Any]"  [arg-type]
```
while the following works
```python
from mpi4py import MPI
import dolfinx
import numpy as np


x = dolfinx.fem.Constant(mesh, np.float64(1.0))
```
In my opinion it is quite annoying to have to wrap all floats in a `numpy` float, so here I am suggesting to add (native) `float` as a valid type to `dolfinx.fem.Constant`

As a side note, the main issue here is really that 
```python
import numpy as np

x: np.floating = 1.0
```
fails in mypy.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -50,7 +50,7 @@ jobs:
         run: |
           export PATH="$(brew --prefix gfortran)/bin:$(brew --prefix bison)/bin:$PATH"
           export PATH="$(brew --prefix make)/libexec/gnubin:$PATH"
-          git clone -b release https://gitlab.com/petsc/petsc.git petsc
+          git clone --depth 1 --branch release https://gitlab.com/petsc/petsc.git petsc
           cd petsc
           python ./configure \
             --with-64-bit-indices=no \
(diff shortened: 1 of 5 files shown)


python/dolfinx/fem/function.py
```diff
@@ -38,7 +38,7 @@ class Constant(ufl.Constant):
     ]
 
     def __init__(
-        self, domain, c: typing.Union[np.ndarray, Sequence, np.floating, np.complexfloating]
+        self, domain, c: typing.Union[float, np.ndarray, Sequence, np.floating, np.complexfloating]
```
schnellerhase:
```suggestion
        self, domain, c: typing.Union[float, np.floating, complex, np.complexfloating, Sequence, np.ndarray]
```

garth-wells (approved):

Issue #3848, opened 2025-08-13
Blocked linear solver default arguments are not accepted


SteinStoter:
When blocked ufl forms are sent to the (linear) solver, the default arguments for `petsc_options_prefix`, `kind` and `bcs` are not accepted:

```
from dolfinx import mesh, fem
import dolfinx.fem.petsc
from mpi4py import MPI
import ufl

domain = mesh.create_unit_square(MPI.COMM_WORLD, 5, 5, mesh.CellType.triangle)
RT = fem.functionspace(domain, ("RT", 1))
DG0 = fem.functionspace(domain, ("DG", 0))
W = ufl.MixedFunctionSpace(RT, DG0)
sigma, p = ufl.TrialFunctions(W)
tau, q = ufl.TestFunctions(W)
a = ufl.extract_blocks( ufl.inner(sigma, tau) * ufl.dx + ufl.inner(p, ufl.div(tau)) * ufl.dx + ufl.inner(ufl.div(sigma), q) * ufl.dx )
L = [ufl.ZeroBaseForm((sigma,)), ufl.ZeroBaseForm((q,))]

# Expected runnable version 1
problem = fem.petsc.LinearProblem(a,L) # This does not run due to missing `petsc_options_prefix`

# Expected runnable version 2
problem = fem.petsc.LinearProblem(a,L, petsc_options_prefix="mysolver") # This does not run due to missing `kind`

# # Expected runnable version 3
problem = fem.petsc.LinearProblem(a,L, kind="nest", petsc_options_prefix="mysolver") 
sigma,u = problem.solve() # This does not run due to missing `bcs`

# Running version
problem = fem.petsc.LinearProblem(a,L, kind="nest", petsc_options_prefix="mysolver", bcs=[]) 
sigma,u = problem.solve()
```

SteinStoter:
Just my 2cts on these considerations:

For novice users, I think it'd be valuable for `fem.petsc.LinearProblem(a,L)` to simply work.

# Expected runnable version 1
So the reason why a valid default argument is not specified in the solvers is that if a user creates _multiple_ nonlinear solvers, PETSc SNES has issues? Isn't that like 1% of use cases? It seems to me this is creating a roadblock (and errors raised) for 99% of use cases just to avoid an error raised in the 1% of use cases.

# Expected runnable version 2:
[...]

PR #3849, opened 2025-08-13
Consistency of input arguments to `LinearProblem`

jorgensd:
As pointed out in https://github.com/FEniCS/dolfinx/issues/3848
there are some inconsistencies in `LinearProblem` and `create_vector`.

According to the documentation, `create_vector` should create a `dolfinx`-blocked vector **if** kind is None. This was currently not the case. 

Similarly, this means that the default behavior of `LinearProblem` should reflect this, and `kind=None` gives you the DOLFINx blocked system.

Finally, default bcs=None was not correctly handled.

### DIFF
--- a/python/dolfinx/fem/petsc.py
+++ b/python/dolfinx/fem/petsc.py
@@ -864,7 +864,7 @@ class LinearProblem:
         else:
             self._u = u
 
-        self.bcs = bcs
+        self.bcs = bcs if bcs is not None else []
 
         self._solver = PETSc.KSP().create(self.A.comm)  # type: ignore[attr-defined]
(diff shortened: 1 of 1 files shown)


python/dolfinx/fem/petsc.py
```diff
@@ -864,7 +864,7 @@ def __init__(
         else:
             self._u = u
 
-        self.bcs = bcs
+        self.bcs = bcs if bcs is not None else []
```
francesco-ballarin: For my OCD maybe make it `bcs = [] if bcs is None else bcs`, which is what we used in `NonlinearProblem` :)

```suggestion
        self.bcs = [] if bcs is None else bcs
```

PR #3851, opened 2025-08-13
Wrap FunctionSpace.collapse

### LINKED ISSUE #3850: `FunctionSpace.collapse` returns list, not ndarray
```Python
from mpi4py import MPI

import basix
import dolfinx

import numpy as np

msh = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, 10, 10)
P1 = basix.ufl.element("Lagrange", msh.basix_cell(), 1, dtype=np.float64)
V = dolfinx.fem.functionspace(msh, basix.ufl.mixed_element([P1, P1]))

ME_sub, dofmap = V.sub(0).collapse()
print(type(dofmap))
# > <class 'list'>

# requires conversion before usage
dofmap = np.asarray(dofmap)

my_dofs = np.array([0, 1, 2], dtype=np.int32)
dofs = dofmap[my_dofs]
```

Should return a numpy array.

michalhabera:
Fixes #3850

### DIFF
--- a/python/dolfinx/wrappers/fem.cpp
+++ b/python/dolfinx/wrappers/fem.cpp
@@ -103,7 +103,16 @@ void declare_function_space(nb::module_& m, std::string type)
                      std::shared_ptr<const dolfinx::fem::FiniteElement<T>>>,
                  std::vector<std::shared_ptr<const dolfinx::fem::DofMap>>>(),
              nb::arg("mesh"), nb::arg("elements"), nb::arg("dofmaps"))
-        .def("collapse", &dolfinx::fem::FunctionSpace<T>::collapse)
+        .def("collapse",
+             [](const dolfinx::fem::FunctionSpace<T>& self)
+                 -> std::pair<dolfinx::fem::FunctionSpace<T>,
+                              nanobind::ndarray<std::int32_t, nanobind::numpy>>
+             {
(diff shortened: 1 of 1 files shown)


schnellerhase:
Thanks, should we add a test case for this? There seems to be no test case checking the returned DOF-map.

python/dolfinx/wrappers/fem.cpp
```diff
@@ -103,7 +103,16 @@ void declare_function_space(nb::module_& m, std::string type)
                      std::shared_ptr<const dolfinx::fem::FiniteElement<T>>>,
                  std::vector<std::shared_ptr<const dolfinx::fem::DofMap>>>(),
              nb::arg("mesh"), nb::arg("elements"), nb::arg("dofmaps"))
-        .def("collapse", &dolfinx::fem::FunctionSpace<T>::collapse)
+        .def("collapse",
+             [](const dolfinx::fem::FunctionSpace<T>& self)
+                 -> std::pair<dolfinx::fem::FunctionSpace<T>,
+                              nanobind::ndarray<std::int32_t, nanobind::numpy>>
+             {
+               auto [collapsed_fs, dofs] = self.collapse();
```
schnellerhase:
```suggestion
               auto&& [collapsed_fs, dofs] = self.collapse();
```

schnellerhase (approved):

PR #3852, opened 2025-08-14
Force non random order of execution for C++ tests

schnellerhase:
Catch2 [`>=3.9.0`](https://github.com/catchorg/Catch2/releases/tag/v3.9.0) changes execution order to random by default. This needs to be deactivated for MPI parallel testing.

### DIFF
--- a/cpp/test/main.cpp
+++ b/cpp/test/main.cpp
@@ -10,7 +10,13 @@ int main(int argc, char* argv[])
   // Parallel tests require MPI initialization before any tests run and
   // termination only after all tests complete.
   MPI_Init(&argc, &argv);
-  int result = Catch::Session().run(argc, argv);
+
+  // Configure test session
+  auto session = Catch::Session();
+
+  // Default order is for Catch version >= 3.9.0 changed to rnadom
+  session.configData().runOrder = Catch::TestRunOrder::Declared;
+  int result = session.run(argc, argv);
   MPI_Finalize();
   return result;
 }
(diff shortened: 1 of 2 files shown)


schnellerhase:
`MPI_Dist_graph_create_adjacent` causes an MPI error - without any further output.

schnellerhase:
But the problem appears to be open-mpi 5.0.7 -> 5.0.8. After a local update I can reproduce it.

garth-wells:
> `MPI_Dist_graph_create_adjacent` causes an MPI error - without any further output.

Any particular line in DOLFINx, or is it any call to `MPI_Dist_graph_create_adjacent`? OpenMPI and MPICH have had bugs in the past with `MPI_Dist_graph_create_adjacent` - we have some odd looking `std::vector::reserve` calls in places to work around these.

schnellerhase:
Any call.

garth-wells:
I can't reproduce failure locally on macOS and with OpenMPI 5.0.8 from Homebrew.

schnellerhase:
Maybe race condition. Cycling `mpiexec -np 3 ./unittests [la_vector]` crashes most times, but passes occasionally.

cpp/test/main.cpp
```diff
@@ -10,7 +10,13 @@ int main(int argc, char* argv[])
   // Parallel tests require MPI initialization before any tests run and
   // termination only after all tests complete.
   MPI_Init(&argc, &argv);
-  int result = Catch::Session().run(argc, argv);
+
+  // Configure test session
+  auto session = Catch::Session();
+
+  // Default order is for Catch version >= 3.9.0 changed to rnadom
```
jorgensd:
```suggestion
  // Default order is for Catch version >= 3.9.0 changed to random
```

PR #3853, opened 2025-08-15
Warn about missing IndexMap in Topology

### LINKED ISSUE #3844: Improve `refine` output when edge entities are not computed
```Python
import dolfinx
from mpi4py import MPI

# mesh.topology.create_entities(1) # This is missing
dolfinx.mesh.refine(mesh)

# Output:
# > IndexError: vector
```

This should result in an output of dim 1 entities not being created.

chrisrichardson:
Throw a more appropriate error when there is a missing IndexMap in Topology.

Resolves #3844


schnellerhase:
Should we throw in `index_maps` right above instead, if `maps` is empty? Then we get the same interface behaviour for mixed topologies.

chrisrichardson:
I don't think that would be right. It might be that you want to check if there are any IndexMaps of dim(1) or not, without throwing an error.

schnellerhase (approved):

PR #3854, opened 2025-08-16
Expose original cell indices

jorgensd:
Required to be able to read higher order meshes from external software, such as NetGen.
With this binding exposed, I am able to make a code for reading netgen mixed meshes, such as
image
into DOLFINx.

### DIFF
--- a/python/dolfinx/wrappers/mesh.cpp
+++ b/python/dolfinx/wrappers/mesh.cpp
@@ -703,14 +703,11 @@ void mesh(nb::module_& m)
             const std::vector<std::vector<std::int64_t>>& indices
                 = self.original_cell_index;
             std::vector<nb::ndarray<const std::int64_t, nb::numpy>> idx_nb;
-            idx_nb.reserve(indices.size());
-            std::ranges::for_each(
-                indices,
-                [&idx_nb](const std::vector<std::int64_t>& idx)
-                {
-                  idx_nb.push_back(nb::ndarray<const std::int64_t, nb::numpy>(
-                      idx.data(), {idx.size()}));
-                });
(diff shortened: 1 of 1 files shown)


python/dolfinx/wrappers/mesh.cpp
```diff
@@ -696,6 +696,23 @@ void mesh(nb::module_& m)
...
           nb::arg("original_cell_indices"))
+      .def_prop_ro(
+          "original_cell_indices",
+          [](const dolfinx::mesh::Topology& self)
+          {
+            const std::vector<std::vector<std::int64_t>>& indices
+                = self.original_cell_index;
+            std::vector<nb::ndarray<const std::int64_t, nb::numpy>> idx_nb;
+            idx_nb.reserve(indices.size());
+            std::ranges::for_each(
```
garth-wells: This is hard to read. Would be a lot easier to read as a for loop.

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -696,6 +696,23 @@ void mesh(nb::module_& m)
...
           },
           nb::arg("original_cell_indices"))
+      .def_prop_ro(
+          "original_cell_indices",
+          [](const dolfinx::mesh::Topology& self)
+          {
+            const std::vector<std::vector<std::int64_t>>& indices
+                = self.original_cell_index;
+            std::vector<nb::ndarray<const std::int64_t, nb::numpy>> idx_nb;
+            idx_nb.reserve(indices.size());
```
garth-wells: No need for this. Won't make any measurable performance difference.

garth-wells (changes requested):

PR #3855, opened 2025-08-17
Generalise `la::Vector` to support GPUs

garth-wells:
Allows the user to specify the container types for `la::Vector` and `common::Scatterer`. The containers can, for example, be `thrust::device_vector` for on-device storage of data for a multi-GPU `Vector`.

- Changes the data return type for `la::Vector::array` from `std::span` to `container_type&`.
- Changes some  `common::Scatterer` args to plains pointers, which can be device pointers. Using spans was misleading because device entries can't be accessed by `operator[]`.
- Communication of  on-device data requires a GPU-ware MPI build.
- Improves `common::Scatterer` documentation.
- Improve type deduction in assembler functions, reducing need to use to supply template arguments.

### DIFF
--- a/cpp/dolfinx/common/Scatterer.h
+++ b/cpp/dolfinx/common/Scatterer.h
@@ -10,6 +10,7 @@
 #include "MPI.h"
 #include "sort.h"
 #include <algorithm>
+#include <concepts>
 #include <functional>
 #include <memory>
 #include <mpi.h>
@@ -20,11 +21,12 @@
 
 namespace dolfinx::common
 {
-/// Types of MPI communication pattern used by the Scatterer.
+/// @brief Types of MPI communication pattern that can be used by a
+/// Scatterer.
 enum class ScattererType : std::uint8_t
 {
-  neighbor, // use MPI neighborhood collectives
-  p2p       // use MPI Isend/Irecv for communication
+  neighbor, // MPI neighborhood collectives
+  p2p       // MPI Isend/Irecv for communication
 };
 
 /// @brief A Scatterer supports the MPI scattering and gathering of data
@@ -33,11 +35,23 @@ enum class ScattererType : std::uint8_t
 /// Scatter and gather operations use MPI neighbourhood collectives. The
 /// implementation is designed for sparse communication patterns, as it
 /// typical of patterns based on an IndexMap.
+
+/// @brief A Scatterer supports the MPI scattering and gathering of data
+/// that is associated with a common::IndexMap.
+///
+/// Scatter and gather operations use MPI neighbourhood collectives. The
+/// implementation is designed for sparse communication patterns, as it
+/// typical of patterns based on an IndexMap.
+///
+/// @tparam Container Container type for storing local and remote
+/// indices. On CPUs this is normally `std::vector<std::int32_t>`. For
+/// GPUs the container should store the indices on the device.
 template <class Container = std::vector<std::int32_t>>
+  requires std::is_integral_v<typename Container::value_type>
 class Scatterer
 {
 public:
-  /// Container type
+  /// Container type used to store local and remote indices.
   using container_type = Container;
 
   /// @brief Create a scatterer.
@@ -48,7 +62,10 @@ public:
   /// `map` that will be scattered/gathered.
   Scatterer(const IndexMap& map, int bs)
       : _bs(bs), _src(map.src().begin(), map.src().end()),
-        _dest(map.dest().begin(), map.dest().end())
+        _dest(map.dest().begin(), map.dest().end()),
+        _sizes_remote(_src.size(), 0), _displs_remote(_src.size() + 1),
+        _sizes_local(_dest.size()), _displs_local(_dest.size() + 1)
+
   {
     if (dolfinx::MPI::size(map.comm()) == 1)
       return;
@@ -94,13 +111,13 @@ public:
     // disp[i] is the first entry in the buffer sent to neighbourhood
     // rank i, and disp[i + 1] - disp[i] is the number of values sent to
     // rank i.
-    _sizes_remote.resize(_src.size(), 0);
-    _displs_remote.resize(_src.size() + 1, 0);
+    assert(_sizes_remote.size() == _src.size());
+    assert(_displs_remote.size() == _src.size() + 1);
     std::vector<std::int32_t>::iterator begin = owners_sorted.begin();
     for (std::size_t i = 0; i < _src.size(); i++)
     {
       auto upper = std::upper_bound(begin, owners_sorted.end(), _src[i]);
-      int num_ind = std::distance(begin, upper);
+      std::size_t num_ind = std::distance(begin, upper);
       _displs_remote[i + 1] = _displs_remote[i] + num_ind;
       _sizes_remote[i] = num_ind;
       begin = upper;
@@ -113,8 +130,8 @@ public:
 
     // Compute sizes and displacements of local data (how many local
     // elements to be sent/received grouped by neighbors)
-    _sizes_local.resize(_dest.size());
-    _displs_local.resize(_sizes_local.size() + 1);
+    assert(_sizes_local.size() == _dest.size());
+    assert(_displs_local.size() == _dest.size() + 1);
     _sizes_remote.reserve(1);
     _sizes_local.reserve(1);
     MPI_Neighbor_alltoall(_sizes_remote.data(), 1, MPI_INT32_T,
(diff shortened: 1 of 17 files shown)


cpp/demo/poisson_matrix_free/main.cpp
```diff
@@ -95,7 +95,7 @@ int cg(auto& x, auto& b, auto action, int kmax = 50, double rtol = 1e-8)
   // Create p work vector
   la::Vector p(r);
 
-  // Iterations of CG
+  // // Iterations of CG
```
jorgensd: Is this on purpose?

cpp/dolfinx/common/Table.cpp
```diff
@@ -105,7 +105,7 @@ Table Table::reduce(MPI_Comm comm, Table::Reduction reduction) const
   // Get keys, values into containers for int doubles
   std::string keys;
   std::vector<double> values;
-  for (const auto& it : _values)
+  for (auto& it : _values)
```
jorgensd, line 108: what is the reason for this change and the one below?

cpp/dolfinx/la/Vector.h
```diff
@@ -22,161 +22,328 @@
...
-        _x(bs * (map->size_local() + map->num_ghosts()))
+      : _map(map), _bs(bs), _x(bs * (map->size_local() + map->num_ghosts())),
+        _scatterer(
+            std::make_shared<common::Scatterer<ScatterContainer>>(*_map, bs)),
+        _buffer_local(_scatterer->local_indices().size()),
+        _buffer_remote(_scatterer->remote_indices().size())
   {
   }
 
+  // TODO: update
```
jorgensd: Remove?

cpp/dolfinx/la/Vector.h
```diff
@@ -22,161 +22,328 @@
...
-      : _map(x._map), _scatterer(x._scatterer), _bs(x._bs),
-        _request(1, MPI_REQUEST_NULL), _buffer_local(x._buffer_local),
-        _buffer_remote(x._buffer_remote), _x(x._x)
+      : _map(x._map), _bs(x._bs), _x(x._x), _scatterer(x._scatterer),
+        _request(MPI_REQUEST_NULL), _buffer_local(x._buffer_local),
+        _buffer_remote(x._buffer_remote)
   {
   }
 
+  // TODO: update
```
jorgensd: Remove?

cpp/dolfinx/la/Vector.h
```diff
@@ -22,161 +22,328 @@
...
-        _buffer_remote(std::move(x._buffer_remote)), _x(std::move(x._x))
+        _buffer_remote(std::move(x._buffer_remote))
   {
   }
 
   // Assignment operator (disabled)
   Vector& operator=(const Vector& x) = delete;
 
-  /// Move Assignment operator
+  // TODO: update
```
jorgensd: Remove?

cpp/dolfinx/common/Table.cpp
garth-wells: It’s not needed. The function is const so all member data is const. Unnecessary const is clutter.

garth-wells:
> In general looks good to me. Good documentation of deprecated functions helps a lot! Looking forward to having GPU CI at some point so one could add some tests.

I got GPU CI on Azure or AWS working in https://github.com/ukri-bench/benchmark-dolfinx. Will soon add GPU examples with CI to https://github.com/FEniCS/dolfinx-gpu-solvers.

cpp/dolfinx/la/Vector.h
```diff
@@ -22,161 +22,313 @@
...
 class Vector
 {
   static_assert(std::is_same_v<typename Container::value_type, T>);
 
-public:
-  /// Scalar type
-  using value_type = T;
+private:
+  /// @brief Return a 'pack' function for packing a send buffer.
+  ///
```
chrisrichardson, line 55: Is this CPU only? Add note... same for other unpack functions

cpp/dolfinx/la/Vector.h
```diff
@@ -22,161 +22,313 @@
...
-
-    auto unpack = [](auto&& in, auto&& idx, auto&& out, auto op)
-    {
-      for (std::size_t i = 0; i < idx.size(); ++i)
-        out[idx[i]] = op(out[idx[i]], in[i]);
-    };
-
-    unpack(_buffer_remote, _scatterer->remote_indices(), x_remote,
-           [](auto /*a*/, auto b) { return b; });
+    this->scatter_fwd_end(get_unpack());
```
chrisrichardson, line 234: Remove `this->` (not needed, and inconsistent usage) ?

cpp/dolfinx/la/Vector.h
```diff
@@ -249,8 +412,9 @@ auto inner_product(const V& a, const V& b)
   return result;
 }
 
-/// Compute the squared L2 norm of vector
-/// @note Collective MPI operation
+/// @brief Compute the squared L2 norm of vector.
+///
```
chrisrichardson, line 417: CPU only?

cpp/dolfinx/la/Vector.h
```diff
@@ -185,53 +337,64 @@ class Vector
...
+}; // namespace dolfinx::la
 
-  // Vector data
-  container_type _x;
-};
-
-/// Compute the inner product of two vectors. The two vectors must have
-/// the same parallel layout
+/// @brief Compute the inner product of two vectors.
+///
```
chrisrichardson, line 379: CPU only?

cpp/dolfinx/la/Vector.h
garth-wells: I added this for readability to be really clear then the function comes from `Vector` vs when it comes from `Scatterer`.

cpp/dolfinx/la/Vector.h
garth-wells: Probably, but I don't actually know. Would need testing of how GPU-aware backends behave.

cpp/dolfinx/la/Vector.h
garth-wells: Comment added.

cpp/dolfinx/common/Scatterer.h
```diff
@@ -20,40 +21,49 @@
...
-/// Scatter and gather operations use MPI neighbourhood collectives. The
-/// implementation is designed for sparse communication patterns, as it
-/// typical of patterns based on an IndexMap.
-template <class Allocator = std::allocator<std::int32_t>>
+/// Scatter and gather operations can use:
+/// 1. MPI neighbourhood collectives (recommended), or
+/// 2. Non-blocking point-to-point communication modes.
+///
+/// The implementation is designed for sparse communication
+/// patterns, as it typical of patterns based on an IndexMap.
```
schnellerhase:
```suggestion
/// patterns, as is typical of patterns based on an IndexMap.
```

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -76,9 +76,11 @@ using mdspan2_t = md::mdspan<const std::int32_t, md::dextents<std::size_t, 2>>;
 /// conditions applied.
 /// @param[in] x0 Vector used in the lifting.
 /// @param[in] alpha Scaling to apply.
-template <dolfinx::scalar T, int _bs0 = -1, int _bs1 = -1>
+template <int _bs0 = -1, int _bs1 = -1, typename V,
+          dolfinx::scalar T = typename std::remove_cvref_t<V>::value_type>
+  requires std::is_same_v<typename std::remove_cvref_t<V>::value_type, T>
```
schnellerhase, line 81: Can the template type `T` here and below be removed? Will always be `V::value_type`.

garth-wells: Yes, it can be removed but it's there for convenience. If it's removed, the template code gets hard to read because everywhere you see `T` (which appears quite a bit) in the concepts constraint it becomes `typename std::remove_cvref_t<V>::value_type>`. It's then long and very cluttered.

schnellerhase: I see. Could use a declaration `using scalar_t = typename std::remove_cvref_t<V>::value_type>` inside the functions - but that does not resolve the verbose signature for `x`.

garth-wells: I originally didn't have `T` in the template code and declared it inside the function, but readability of the template code was poor.

Issue #3856, opened 2025-08-17
Create entity permutations for mixed topology meshes


jorgensd:
### Describe new/missing feature

Currently, a lot of the internal logic in entity permutations assumes a single topology, meaning that we cannot use elements requiring this info.

This includes higher order geometries (order>2), as this info is used to permute the geometry dofmap.

### Suggested user interface

```python3

```

PR #3857, opened 2025-08-17
Start handling mixed topology more properly

jorgensd:
Start removing usage of connectivity(a, b) in the core of the code. Progress towards #3856

### DIFF
--- a/cpp/dolfinx/mesh/Geometry.h
+++ b/cpp/dolfinx/mesh/Geometry.h
@@ -271,8 +271,9 @@ create_geometry(const Topology& topology,
   // If the mesh has higher order geometry, permute the dofmap
   if (elements.front().needs_dof_permutations())
   {
-    const std::int32_t num_cells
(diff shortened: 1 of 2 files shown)


cpp/dolfinx/mesh/Geometry.h
```diff
@@ -271,8 +271,9 @@ create_geometry(const Topology& topology,
   // If the mesh has higher order geometry, permute the dofmap
   if (elements.front().needs_dof_permutations())
   {
-    const std::int32_t num_cells
-        = topology.connectivity(topology.dim(), 0)->num_nodes();
+    std::int32_t num_cells = 0;
+    for (const auto& imap : topology.index_maps(topology.dim()))
+      num_cells += imap->size_local() + imap->num_ghosts();
```
chrisrichardson, line 276: This is not right. Can't just add up all cells. Each type must be treated differently.

PR #3858, opened 2025-08-17
Integrate C++ tests into project cmake

schnellerhase:
Includes the c++ tests into the root cmake project with the standard `CMake` testing functionality.

- Standard flag [`BUILD_TESTING`](https://cmake.org/cmake/help/latest/variable/BUILD_TESTING.html) enables/disables build of tests
- All compile time flags, settings etc are inherited now from the central project setup and do not require to be reconfigured.
- Test build/execution do not require a prior installation.
- Parallel test cases are run through `ctetst` automatically (Removing the need to rerun with different parallelisation settings
- New option `DOLFINX_TEST_MPI_PROCESSES` encodes the process counts to run during the test phase

This is motivated by
1. Development experience on the C++ end, the install + change directory + compile/run tests setup simplifies to
```bash
ninja && ctest
```
which automatically tests the setup as performed in the CIs.
2. Combines compile of dolfinx (library) and tests into a combined step.
3.  compliance with the cmake tooling.

This could/should be extended to include the demos in another PR.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -109,6 +109,7 @@ jobs:
           --warn-uninitialized
           -G Ninja
           -DCMAKE_BUILD_TYPE=Developer
+          -DBUILD_TESTING=true
           -DDOLFINX_ENABLE_PETSC=false
 
       - name: Build and install (C++)
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -108,8 +108,7 @@ jobs:
           -Werror=dev
           --warn-uninitialized
           -G Ninja
-          -DCMAKE_BUILD_TYPE=Developer 
-          -DBUILD_TESTING=true
+          -DCMAKE_BUILD_TYPE=Developer
           -DDOLFINX_ENABLE_PETSC=false
 
       - name: Build and install (C++)
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -101,23 +101,26 @@ jobs:
           pip install --no-build-isolation git+https://github.com/fenics/basix.git@${{ env.basix_ref }}
           pip install --no-build-isolation git+https://github.com/fenics/ffcx.git@${{ env.ffcx_ref }}
 
-      - name: Configure and install C++
+      - name: Configure (C++)
+        working-directory: cpp
+        run: >
+          cmake -B build -S .
+          -Werror=dev
+          --warn-uninitialized
+          -G Ninja
+          -DCMAKE_BUILD_TYPE=Developer 
+          -DBUILD_TESTING=true
+          -DDOLFINX_ENABLE_PETSC=false
+
+      - name: Build and install (C++)
+        working-directory: cpp/build
         run: |
-          cmake -Werror=dev --warn-uninitialized -G Ninja -DDOLFINX_ENABLE_PETSC=false -DCMAKE_BUILD_TYPE=Developer -B build -S cpp/
-          cmake --build build
-          sudo cmake --install build
-      - name: Build C++ unit tests
-        run: |
-          cmake -Werror=dev --warn-uninitialized -G Ninja -DCMAKE_BUILD_TYPE=Developer -B build/test/ -S cpp/test/
-          cmake --build build/test
-      - name: Run C++ unit tests (serial)
-        run: |
-          cd build/test
-          ctest -V --output-on-failure -R unittests
-      - name: Run C++ unit tests (MPI)
-        run: |
-          cd build/test
-          mpirun -n 3 ctest -V --output-on-failure -R unittests
+          cmake --build .
+          sudo cmake --install .
+
+      - name: Run tests (C++)
+        working-directory: cpp/build
+        run: ctest -V --output-on-failure -R unittests
 
       - name: Build Python interface
         run: |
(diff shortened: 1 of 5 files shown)


cpp/CMakeLists.txt
```diff
@@ -479,6 +479,15 @@ endif()
 # Installation of DOLFINx library
 add_subdirectory(dolfinx)
 
+# ------------------------------------------------------------------------------
+# Unit testing
+option(BUILD_TESTING "Build DOLFINx unit tests." OFF)
```
jhale, line 484: `BUILD_TESTS` would be better.

.github/workflows/ccpp.yml
```diff
@@ -186,24 +189,30 @@ jobs:
...
+          -DDOLFINX_ENABLE_PETSC=true
+          -DDOLFINX_ENABLE_SCOTCH=true
+          -DDOLFINX_ENABLE_SLEPC=true 
+          -DDOLFINX_ENABLE_PARMETIS=false
+
+      - name: Build and install (C++)
+        working-directory: cpp/build
+        run: cmake --build . --target install
+
+      - name: Run tests (C++)
```
jhale, line 212: How is MPI with 3 ranks handled now?

.github/workflows/ccpp.yml
```diff
@@ -186,24 +189,30 @@ jobs:
...
-      - name: Configure, build and install C++ library
-        run: |
-          cmake -Werror=dev --warn-uninitialized -G Ninja -DCMAKE_BUILD_TYPE=Developer -DDOLFINX_ENABLE_ADIOS2=true -DDOLFINX_ENABLE_KAHIP=true -DDOLFINX_ENABLE_PARMETIS=false -DDOLFINX_ENABLE_PETSC=true -DDOLFINX_ENABLE_SCOTCH=true -DDOLFINX_ENABLE_SLEPC=true -B build -S cpp/
-          cmake --build build
-          cmake --install build
+      - name: Configure (C++)
+        working-directory: cpp
+        run: >
+          cmake -B build -S .
+          -Werror=dev
```
jhale, line 196: Would be more readable if the lines following the first were indented two spaces.

.github/workflows/ccpp.yml
```diff
@@ -101,23 +101,26 @@ jobs:
...
-          ctest -V --output-on-failure -R unittests
-      - name: Run C++ unit tests (MPI)
-        run: |
-          cd build/test
-          mpirun -n 3 ctest -V --output-on-failure -R unittests
+      - name: Configure (C++)
+        working-directory: cpp
+        run: >
+          cmake -B build -S .
+          -Werror=dev
```
jhale, line 108: Indent two spaces from here.

cpp/test/CMakeLists.txt
```diff
@@ -78,29 +62,42 @@ if(WIN32)
...
+    unittests
+    PROPERTIES
+      CXX_CLANG_TIDY
+      "${CLANG_TIDY};--config-file=${CMAKE_CURRENT_SOURCE_DIR}/../../.clang-tidy"
+  )
 endif()
 
-# Enable testing
-enable_testing()
+# Nice for debugging tests one at a time - but extremely slow due to process
```
jhale, line 98: Don't understand the comment. And if it's slow should it be enabled?

schnellerhase: `catch_discover_tests` will register for every test case an own ctest test. Every test case corresponds to a process being started and the (same) test executable being invoked. This overhead is a >x100. Benefit is however that one can then filter directly with ctest for exact test runs, which is nice for debugging.

cpp/CMakeLists.txt
schnellerhase: `BUILD_TESTING` is a cmake default https://cmake.org/cmake/help/latest/variable/BUILD_TESTING.html.

.github/workflows/ccpp.yml
schnellerhase: Automatically run through the ctest invocation. The processes tested can be configured with the (newly introduced) cmake flag `TEST_PROCESSES`, which defaults to "1;3" - maintaining previous parallel testing behaviour.

.github/workflows/ccpp.yml
schnellerhase: Unfortunately that's not possible - https://github.com/orgs/community/discussions/26105

schnellerhase:
> Main global comment: Is it still possible to build and configure the tests separately, and if so, can one of the CI suites be kept on the separate path?

With the current setup this would not be possible anymore. Mainly because the test configuration does not create a fully featured cmake project anymore (with the upside of not having to duplicate all the necessary flags). Do we want/need the support of building tests without the dolfinx cmake context?

jhale:
The main thing we'd be looking to include in the future is automated testing via Spack where it seems that `make check` and `make test` targets are expected.

https://spack.readthedocs.io/en/latest/packaging_guide_testing.html

On that note it would be nice if targets were created that further abstracted the process of running `ctest`.

schnellerhase:
`make test` we get 'for free' with cmake if we are using make instead of ninja - which just invokes ctest. So the changes here move us towards that, no longer need to invoke ctest multiple times.

Not sure about check, we aren't using that one so far.

jhale:
Thanks for the clarifications. Does something like `cmake --build . --target test` give you a build tool independent run of the tests?

cpp/CMakeLists.txt
```diff
@@ -479,6 +479,15 @@ endif()
...
+set(TEST_PROCESSES "1;3" CACHE STRING "List of process counts to run tests with.")
```
jhale: Why isn't this an `option`? Should it be called e.g. `DOLFINX_TEST_MPI_PROCESSES`? or similar?

schnellerhase: ~No particular reason, will change.~ option only stores booleans https://cmake.org/cmake/help/latest/command/option.html

jhale:
Works fine with:

```
cmake --build . --target test
```

No need to change all of them, good to have coverage of both methods. This should allow us to add automated testing into our Spack package.

schnellerhase:
In a fresh build directory this does not compile the dolfinx library before running the tests. If before the build was already run it passes fine. Maybe we are missing a dependency.

jhale:
Let's not override the default internally, tends to lead to confusion. For me the default behaviour makes sense - for Spack set it as an explicit argument.

jhale:
On reflection, let's revert the last commit. The default two step makes sense, and it's going to be very hard to debug build build and test in a single job step.

schnellerhase:
> I find the change inconvenient for developing and debugging tests.

The main advantage we should have gotten with this change is managing compile of tests from within the same cmake setup. It should have gotten easier to debug tests as this is now possible without the previous necessary two step of installing the library and building/running tests from another build directory.

> It forces a build framework to know about `ctest`.

I think we can move the `Ctest` include into the test cmake file. That should make `ctest` a test only dependency.

garth-wells:
> * Could you be more precise on the workflow that no longer works?

If I already have a build of DOLFINx, how can I build and run the tests?

Our Spack config doesn't support tests yet (my experience has been that Spack test support was shaky, to may have improved). I have a Spack build of DOLFINx and I would like to then build and run the tests.

schnellerhase:
Then you are running a local spack [`developer mode`](https://spack-tutorial.readthedocs.io/en/latest/tutorial_developer_workflows.html#developer-workflows-tutorial) for `dolfinx`? I haven't used this myself, but one should be able to access the build directory. From there it should behave like a source build: reconfigure to enable `BUILD_TESTING` and build/run the test targets.

PR #3863, opened 2025-08-19
Fix indentation of create_mesh

jorgensd:
Wrong indentation in documentation

### DIFF
--- a/python/dolfinx/mesh.py
+++ b/python/dolfinx/mesh.py
@@ -685,8 +685,7 @@ def create_mesh(
 
     Args:
         comm: MPI communicator to define the mesh on.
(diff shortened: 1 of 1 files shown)


python/dolfinx/mesh.py
```diff
@@ -685,8 +685,7 @@ def create_mesh(
 
     Args:
         comm: MPI communicator to define the mesh on.
-            cells: Cells of the mesh. ``cells[i]`` are the 'nodes' of cell
-            ``i``.
+        cells: Cells of the mesh. ``cells[i]`` are the 'nodes' of cell ``i``.
```
jorgensd:
```suggestion
        cells: Cells of the mesh. ``cells[i]`` are the 'nodes' of
            cell ``i``.
```

PR #3870, opened 2025-08-23
Remove `mpiexec` in cmake test target

jhale:
Exporting a test target is nice but the fixed `mpiexec` hits walls quickly - we have no idea what downstream consumers need to use.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -118,9 +118,12 @@ jobs:
           cmake --build .
           sudo cmake --install .
 
-      - name: Run tests (C++)
+      - name: Run tests (C++, serial)
         working-directory: cpp/build
-        run: cmake --build . --target test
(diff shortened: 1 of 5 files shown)


jhale:
By colleagues here at uni.lu I've been persuaded that we should use the output of FindMPI to discover the correct way to invoke `mpiexec`:

https://cmake.org/cmake/help/latest/module/FindMPI.html

```
find_package(MPI)

if(MPI_FOUND)
  execute_process(
    COMMAND
      ${MPIEXEC_EXECUTABLE}
      ${MPIEXEC_NUMPROC_FLAG}
      ${MPIEXEC_MAX_NUMPROCS}
      ${MPIEXEC_PREFLAGS}
      <executable>
      ${MPIEXEC_POSTFLAGS}
      <args>
  )
endif()
```

jhale:
Partial reversion to old behaviour completed in #3984.

PR #3872, opened 2025-08-24
Make `ctest` test only dependency

schnellerhase:
Only include  `ctest` if testing is enabled, removes the dependency on `ctest`, for builds without `BUILD_TESTING`.

Follow up to https://github.com/FEniCS/dolfinx/pull/3858


jhale:
I've done a bit of reading here - even if `ctest` is not built/provided as part of a cmake package (a very rare occurrence), the `CTest` module is always included with the `cmake`. I don't think that changes needing this PR though.

PR #3875, opened 2025-08-25
allow stable abi builds

minrk:
when stable abi is requested in scikit-build-core's [standard way](https://scikit-build-core.readthedocs.io/en/latest/guide/cmakelists.html#limited-api-stable-abi) (e.g. `pip install . -C wheel.py-api=cp312`), pass STABLE_API to nanobind.

This doesn't change defaults, it just accepts the flag from scikit-build-core and passes it through to the right places. It doesn't mean stable abi builds work, it just means they can be attempted.

I believe this will require mpi4y and possibly petsc4py to also be built with the stable api, which is possible in mpi4py 4.1 and will be possible in petsc4py 3.24.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -31,6 +31,11 @@ execute_process(
 list(APPEND CMAKE_PREFIX_PATH "${NB_DIR}")
 find_package(nanobind CONFIG REQUIRED)
 
+# Enabling C here is a workaround for an ADIOS2 bug, see
+# https://github.com/ornladios/ADIOS2/pull/4255. Remove when the ADIOS2
+# fix is released.
+enable_language(C)
+
 find_package(DOLFINX REQUIRED CONFIG)
 
 if(DOLFINX_FOUND)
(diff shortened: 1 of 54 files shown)


jhale:
This looks quite exciting in the context of delivering conda packages and binary wheels of FEniCS that are compatible with many versions of Python and any MPI compatible with the new ABI initiative.

It seems like petsc4py is the last piece of the puzzle (I checked mpi4py recently and everything looks ready there already). Do you know if they plan to ship binary wheels of petsc4py on pypi?

minrk:
> Do you know if they plan to ship binary wheels of petsc4py on pypi?

I don't believe there are any plans to do that. mpi4py wheels [depend on findable libmpi](https://github.com/mpi4py/mpi4py/blob/master/src/mpi4py/_mpiabi.py) on the system with a compatible abi. They _do not_ include libmpi itself. That's not _so_ hard for mpich and openmpi, which tend to have very stable ABIs, which are _not_ very sensitive to the build configuration.

A similar approach could be taken for petsc4py, but petsc's ABI is much more dependent on build configuration and has many more axes of ABI breakage (even than across mpi providers). So if libpetsc itself is not bundled, I'm not sure what to choose.

There is a `petsc` on PyPI that I assume is part of making `pip install petsc4py` work, but I don't know how they would choose which options of:

- dependencies
- scalar type
- precision
- cuda
- mpi
- etc.

all of which contribute to the ABI of petsc itself
for a binary release, as opposed to the current source-only user-compiled approach.

python/CMakeLists.txt
```diff
@@ -56,6 +62,10 @@ nanobind_add_module(
   dolfinx/wrappers/petsc.cpp
   dolfinx/wrappers/refinement.cpp
 )
+# mpi4py and petsc4py use generated Cython headers
+# make sure to tell them that the limited API is in use
+target_compile_definitions(cpp PRIVATE CYTHON_COMPILING_IN_LIMITED_API=1)
```
jhale: Should this only be enabled if `NANOBIND_SABI` is `STABLE_ABI`?

python/CMakeLists.txt
```diff
@@ -56,6 +62,10 @@ nanobind_add_module(
...
+target_compile_definitions(cpp PRIVATE MPI4PY_LIMITED_API=1)
```
jhale: Same comment.

python/CMakeLists.txt
minrk: Yes, I'll add that condition.

minrk: I added the condition and a comment that it's actually a Cython bug that they are needed. If you want, we could leave that out and leave it to users' to add CYTHON_COMPILING_IN_LIMITED_API via CFLAGS.

PR #3879, opened 2025-08-26
Make tests work as standalone project

schnellerhase:
Allows to also use the c++ tests as a standalone project. Building on top of an already installed dolfinx version.

Also fixes a bad linkage against `Catch2::Catch2WithMain` - since we provide our own `main` - we need to link against `Catch2::Catch2`.

### DIFF
--- a/cpp/test/CMakeLists.txt
+++ b/cpp/test/CMakeLists.txt
@@ -11,8 +11,6 @@ if (PROJECT_IS_TOP_LEVEL)
   set(CMAKE_CXX_STANDARD_REQUIRED ON)
   set(CMAKE_CXX_EXTENSIONS OFF)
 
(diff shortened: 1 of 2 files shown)


cpp/test/CMakeLists.txt
```diff
@@ -1,3 +1,17 @@
+cmake_minimum_required(VERSION 3.21)
+
+project(dolfinx-tests LANGUAGES C CXX)
+
+if (PROJECT_IS_TOP_LEVEL)
+  include(CTest) # enables testing
+  find_package(DOLFINX REQUIRED)
+
+  set(CMAKE_C_STANDARD 17) # For FFCx generated .c files.
```
jhale: This is old fashioned and probably should be applied to the target.

PR #3881, opened 2025-08-27
Small simplifications of gmsh IO interface

schnellerhase:
- Rename `dolfinx.io.gmsio` to `dolfinx.io.gmsh` - remove duplicate `io`. If one needs to differentiate between `gmsh` and the `io` module, this is possible with `from dolfin.io import gmsh as gmshio `
- Physical group tuple of `dim` and `tag` now is a named tuple: unpacking those before as `[0]` and `[1]` was hard to follow in user code.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -1,11 +1,11 @@
-name: DOLFINx CI
+name: CI
 
 on:
-  push:
-    branches:
-      - "**"
-    tags:
-      - "v*"
+  # push:
+  #   branches:
+  #     - "**"
+  #   tags:
+  #     - "v*"
   pull_request:
     branches:
       - main
@@ -97,9 +97,9 @@ jobs:
 
       - name: Install FEniCS Python components
         run: |
-          pip install --no-build-isolation git+https://github.com/fenics/ufl.git@${{ env.ufl_ref }}
-          pip install --no-build-isolation git+https://github.com/fenics/basix.git@${{ env.basix_ref }}
-          pip install --no-build-isolation git+https://github.com/fenics/ffcx.git@${{ env.ffcx_ref }}
+          pip install --no-build-isolation git+https://github.com/${{ env.ufl_repository }}.git@${{ env.ufl_ref }}
+          pip install --no-build-isolation git+https://github.com/${{ env.basix_repository }}.git@${{ env.basix_ref }}
+          pip install --no-build-isolation git+https://github.com/${{ env.ffcx_repository }}.git@${{ env.ffcx_ref }}
 
       - name: Configure (C++)
         working-directory: cpp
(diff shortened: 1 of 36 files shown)


michalhabera (approved):

schnellerhase:
Including this way this is possible. However, if one does not import dolfinx modules, so usage through `import dolfinx; dolfinx.io.gmshio` - this becomes harder to read.

If one has both `import gmsh` and `from dolfinx.io import gmsh` - ruff will scream at you anyways. So that should protect against that going unnoticed.

jorgensd:
> If one does not import dolfinx modules, so usage through `import dolfinx; dolfinx.io.gmshio` - this becomes harder to read and can not be simplified.
> 
> If one has both `import gmsh` and `from dolfinx.io import gmsh` - ruff will scream at you anyways. So that should protect against that going unnoticed.

I don't really see the benefit of changing the API, as you have changed the demos to `import dolfinx.io.gmsh as gmshio`, which is equivalent to `from dolfinx.io. import gmshio`.

michalhabera:
> > If one does not import dolfinx modules, so usage through `import dolfinx; dolfinx.io.gmshio` - this becomes harder to read and can not be simplified.
> > If one has both `import gmsh` and `from dolfinx.io import gmsh` - ruff will scream at you anyways. So that should protect against that going unnoticed.
> 
> I don't really see the benefit of changing the API, as you have changed the demos to `import dolfinx.io.gmsh as gmshio`, which is equivalent to `from dolfinx.io. import gmshio`.

How about importing alias `gmshio` in `dolfinx/io/__init__.py` for legacy compatibility and not breaking existing API, but embracing the use of `import dolfinx.io.gmsh as gmshio`. I don't think it is a good habit to mix the module structre (io) into module name (gmsh).

francesco-ballarin:
I don't have an opinion on `dolfinx.io.gmshio` vs `dolfinx.io.gmsh`. Good to keep backward compatibility if we decide to go for the latter, but the legacy version should be explicitly marked as deprecated, otherwise people will keep using it without ever noticing we've changed it.

python/dolfinx/io/gmshio.py
```diff
@@ -1,463 +1,5 @@
...
-    if comm.rank == rank:
-        gmsh.initialize()
-        gmsh.model.add("Mesh from file")
-        gmsh.merge(str(filename))
-        msh = model_to_mesh(gmsh.model, comm, rank, gdim=gdim, partitioner=partitioner)
-        gmsh.finalize()
-        return msh
-    else:
-        return model_to_mesh(gmsh.model, comm, rank, gdim=gdim, partitioner=partitioner)
+warnings.deprecated("dolfinx.io.gmshio is deprected, use dolfinx.io.gmsh.")
```
francesco-ballarin:
```suggestion
warnings.deprecated("dolfinx.io.gmshio is deprecated, use dolfinx.io.gmsh.")
```

jhale:
Marginal but correct - best to do this type of change before we release `1.0`.

PR #3884, opened 2025-08-28
CI testing via Spack (experimental)

garth-wells:
The PR adds experimental 'single source' CI testing via Spack. All other tests are unchanged. The proposal is to merge this PR and monitor how it goes. While experimental, it will not be a merge requirement.

## Background
Issues this PR addresses are:
- Our test framework is not 'single source', it's a mish mash of Docker and Actions. 
- Our Docker files are complicated and bloated. 
- Control over versioning of Python packages in the test system is weak; some things are (built and) installed in the Dockerfile, others are installed in the CI framework. This has caused some recent testing bugs, especially on maintaining an older version of NumPy that is compatible with Numba. 
- We can't practicably test against new/different versions of dependencies and compilers. This requires changes to already complicated Docker files, docker builds and the simplest approach clobbers the previous Docker image tag, which can cause CI failures.

## Spack-based CI
This PR address the above issues.
- The test framework is single-source, the Actions yaml file controls the build and execution.
- Versions of all dependencies can be controlled in a single yaml file.
- Different dependencies can be tested in a branch, with no side-effects.
- Spack packages are cached using the Github container repository (private at the moment, rather than being a public cache).
- The test suite is a bit slower than what we presently have since the dependencies are pulled in from the cache one-by-one by Spack. There is scope to reduce the runtime.
- The core Spack test system for running CI tests through Spack has several issues (some are on the Spack issue tracker). It is not suitable for running our tests at present.

This PR exposed a number of shortcomings/bugs in FEniCS Spack specs, and bugs in dependency Spack spec. This PR uses the specs in a branch spack-packages. The changes will eventually be upstreamed.

## Issues and questions
- This PR uses minimal dependency pinning. Should we pin significant dependencies, e.g. PETSc, and let others float?
- If Spack cannot find a suitable cached dependency, it will re-build and cache. The re-build of some heavy dependencies, e.g. LLVM (for Numba) and VTK (for pyvista) can be very heavy. 'Controlled' rebuilds can be managed and just require patience or running locally on a powerful system and pushing to the cache. What will be problematic is unexpected, heavy re-builds since the CI will suddenly become very slow. Version pinning of heavy dependencies may avoid this.

  If we see too frequent re-builds, we can figure out what to do.

- In this PR,  all Python test dependencies are installed using Spack. 
  
   Advantages:
   - Tight control over compatible dependency versions. 
   - Self-contained, i.e. not dependency the host system libraries.
   
   Disadvantages:
   - Some Python packages have very deep and heavy dependency graphs, e.g. numba (LLVM dep),  (py-)gmsh (lots of graphics library deps) and pyvista (VTK dep).
   -  Quite a few Python packages in Spack are not actively/well maintained. This PR exposed bugs in some Spack Python specs.

   A question for later is: should Python test dependencies be installed via Spack or `pip`? `pip` would be faster and less package maintenance, but would require greater care with Python dependency versions.
   
   The Spack version is presently the development version. We may want to pin this. There are a few details still to add on specifying FEniCS dependency branches.


jhale:
On your questions, I wouldn't mix pip and Spack - let's see how the package caching goes in practice.

garth-wells:
> I think this is a great idea.
> 
> The Docker images have become a mess trying to cater for users (who would like something minimal that works, not bloat) and our testing (which due to the increasingly varied requirements needs a concretiser, which Spack really does well).
> 
> A few small comments:
> 
> * The biggest risk I see with Spack is this issue of packages specs not being updated or being poor quality. Particularly for Python, we can say that most of pypi is curated by package developers, whereas Spack package specs are often community maintained. Compiled system dependencies are often of the highest quality, although of course older.

We can always take on some responsibility for some specs. Updating a few specs is usually pretty straightforward. The difficulty is the rabbit hole of dependencies - the 'blocker' can be an esoteric Python library at the bottom of a long stack.

Or if necessary we use pip for some test dependencies, which comes with pros and cons.

> * Spack `spack/spack` has weekly tags, perhaps it is possible to pin on one of those before moving to a release version later?

Yes, there are plenty of tags we can pin to later.

> * My view is we should still keep one very boring build, e.g. PETSc int 32-bit double precision, on e.g. Ubuntu LTS or RedHat-like image using system-provided dependencies as much as possible.

Agree. This PR is for testing of the code. A simple bare-bones install and test (1) demonstrates how to build DOLFINx from scratch, and (2) provides some build system testing.

jorgensd:
> On your questions, I wouldn't mix pip and Spack - let's see how the package caching goes in practice.

I agree that we shouldn't mix spack and pip, as it easily becomes tricky to ensure compatibility.

With respect to our docker hiearchy, I think we should keep a `Dockerfile.test-env` and `Dockerfile.end-user`, 
with a simplified setup. For instance only support
- PETSc 32 bit - real and complex mode
   - Compatible SLEPc's
- Pyvista (which i think can be installed via pip now that VTK have arm images)
- GMSH (some apt packages required to work)
- ADIOS2 (I think we still need to do this from source to get a working python interface).
and use the `apt` mpi compiler etc.

Im in favor of this as the current dockerfile's make it super easy for end-user/third party vendors to set up CI with minor additional dependencies, which would be incredibly slow to install with spack and a reconcretization.

jorgensd:
> > With respect to our docker hiearchy, I think we should keep a `Dockerfile.test-env` and `Dockerfile.end-user`, with a simplified setup. For instance only support
> > 
> > * PETSc 32 bit - real and complex mode
> >   
> >   * Compatible SLEPc's
> > * Pyvista (which i think can be installed via pip now that VTK have arm images)
> > * GMSH (some apt packages required to work)
> > * ADIOS2 (I think we still need to do this from source to get a working python interface).
> >   and use the `apt` mpi compiler etc.
> > 
> > Im in favor of this as the current dockerfile's make it super easy for end-user/third party vendors to set up CI with minor additional dependencies, which would be incredibly slow to install with spack and a reconcretization.
> 
> What I didn't mention because it's adjacent rather than directly related is that Spack can produce Docker images (very easy, see https://spack.readthedocs.io/en/latest/containers.html#from-existing-installations), or Dockerfiles (https://spack.readthedocs.io/en/latest/containers.html#generating-recipes-for-docker-and-singularity), or build Docker images from an environment spec. This would be a lot cleaner than what we do now.
> 
> Spack CI builds are not slow when used with a build cache. Something to consider for the future is if we want to provide a public Spack mirror/cache (which are the same thing in Spack) for FEniCS.

I am aware of this feature, and it looks very neat. The problem is if a user has a dependency that isn't on spack (or is installed through pip, or you need to compile C++ code against dolfinx, as in the case of DOLFINx_MPC), which we then need to add to the spack spec (i.e. we need to install `py-pip` in the runtime images).
If we do not add `py-pip, cmake, nanobind` and all of our build-deps to those images, a user has to add then, reconcretize and possible go through the whole installation-phase of DOLFINx, as there might be incompatibilities with our build-cache.

jhale:
Let's not get sidetracked on Docker image design for now - none of us have even tried the Spack Docker generation yet.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,156 @@
...
+
+      - name: Build library (Python)
+        run: |
+          . $GITHUB_WORKSPACE/spack-src/share/spack/setup-env.sh
+          spack -e py add py-fenics-dolfinx@main +petsc4py +slepc4py \
+            %fenics-dolfinx@main build_type=Developer +petsc +slepc partitioners=parmetis \
+            ^openmpi ^adios2~fortran~libcatalyst \
+            ^petsc +mpi +hypre +mumps +superlu-dist
+          spack -e py add py-gmsh ^gmsh~fltk+external py-matplotlib py-networkx \
+            py-numba py-pyamg@5.3.0 py-pytest py-pytest-xdist py-scipy py-pyvista
```
jorgensd: Should the pyamg pin be moved to a config?

jorgensd: I see that this is due to it missing from spack atm and is added in your branch. Ignore comment

.github/workflows/ci.yml
```diff
@@ -0,0 +1,156 @@
...
+          spack -e py concretize -j 4
+          spack -e py install -j 4 -p 2 --use-buildcache=auto
+
+      - name: Run Python demos
+        run: |
+          . $GITHUB_WORKSPACE/spack-src/share/spack/setup-env.sh
+          spack env activate py
+          spack load gcc
```
jorgensd, line 147: Is this to ensure jit compilation works?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,156 @@
...
+          ctest -V -R demo -R serial
+          ctest --output-on-failure -V -R demo -R mpi_2
+
+      - name: Prepare Python environment
+        run: |
+          . $GITHUB_WORKSPACE/spack-src/share/spack/setup-env.sh
+          spack info py-fenics-dolfinx
+          spack env create py dolfinx-src/.github/workflows/spack-config/gh-actions-env-test.yml
+          spack -e py develop --path $GITHUB_WORKSPACE/dolfinx-src fenics-dolfinx@main
+          spack -e py develop fenics-basix@git.main=main
```
jhale: These should be checked out using the GitHub action at the git ref specified in the yml file.

cpp/test/CMakeLists.txt
```diff
@@ -4,23 +4,18 @@ project(dolfinx-tests LANGUAGES C CXX)
 
 if (PROJECT_IS_TOP_LEVEL)
   include(CTest) # enables testing
-  find_package(DOLFINX REQUIRED)
-endif()
 
-find_package(Catch2 3)
+  set(CMAKE_C_STANDARD 17) # For FFCx generated .c files.
```
jhale: Is this necessary? Modern cmake specifies these on the target.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,156 @@
...
+          . $GITHUB_WORKSPACE/spack-src/share/spack/setup-env.sh
+          spack repo set --destination $GITHUB_WORKSPACE/spack-packages builtin
+
+      - name: Prepare C++ environment
+        run: |
+          . $GITHUB_WORKSPACE/spack-src/share/spack/setup-env.sh
+          spack info fenics-dolfinx
+          spack env create cxx dolfinx-src/.github/workflows/spack-config/gh-actions-env-test.yml
+          spack -e cxx develop --path $GITHUB_WORKSPACE/dolfinx-src fenics-dolfinx@main
+          spack -e cxx develop fenics-basix@git.main=main
```
jhale: Same comment as below on checkout action and yml git ref.

cpp/test/CMakeLists.txt
```diff
@@ -4,23 +4,18 @@ project(dolfinx-tests LANGUAGES C CXX)
...
+  set(CMAKE_CXX_STANDARD 20)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
+  set(CMAKE_CXX_EXTENSIONS OFF)
 
-if(NOT Catch2_FOUND)
```
jhale, line 12: It's also my preference to remove this, hardly difficult to install Catch2.

python/CMakeLists.txt
```diff
@@ -31,6 +31,11 @@ execute_process(
 list(APPEND CMAKE_PREFIX_PATH "${NB_DIR}")
 find_package(nanobind CONFIG REQUIRED)
 
+# Enabling C here is a workaround for an ADIOS2 bug, see
+# https://github.com/ornladios/ADIOS2/pull/4255. Remove when the ADIOS2
+# fix is released.
+enable_language(C)
```
jhale, line 37: Fine, but would be good to only enable this is adios2 was found.

.github/workflows/ci.yml
garth-wells: > Is this to ensure jit compilation works?

Yes.

.github/workflows/ci.yml
garth-wells: That would be more code for no advantage.

What remains is to get the branch from the refs env file, i.e.
```
- name: Load environment variables
  run: cat .github/workflows/fenicsx-refs.env >> $GITHUB_ENV

- name: Set up env
   run: |
     ....
     spack -e py develop fenics-basix@git.${{ env.basix_ref }}=main
```

The only one we should checkout using actions is DOLFINx - the commit hash is different for 'pull request runs' and 'push runs' (Github creates internal commits for PR triggers). Getting the correct hash ref isn't trivial.

cpp/test/CMakeLists.txt
garth-wells: Maybe not - it was there before.

python/CMakeLists.txt
garth-wells: It's chicken and egg - ADIOS2 can't be found if C is not enabled.

.github/workflows/ci.yml
jhale: There is an advantage: the GitHub action does very shallow clones and handles authentication even into private forks.

PR #3891, opened 2025-08-29
Cache PETSc on macos CI

garth-wells:
Speeds up tests considerably.


jhale:
Thanks for actually doing this I had the same idea a while back. I think this strategy would also work for our proposed "boring build" as long as the cache action works inside a running container.

PR #3897, opened 2025-09-02
Fix kernel extraction on windows

schnellerhase:


jorgensd (approved):

schnellerhase:
Looks like msvc can not deal with friend + requires: https://godbolt.org/z/zTvrPMMn1

jhale:
One to revert when MSVC catches up.

PR #3900, opened 2025-09-03
Unification of one-sided sub-entity integrals + ridge integrals

jorgensd:
## New features
- Introduces ridge integrals (`ufl.dr`) (tdim - 2) (for rank 0, 1 and  2-tensors)
- Unifies assembly routines in `assemble_vector` for all exterior entity integrals.
- Adds `assemble_matrix` and `apply_lifting` for vertex integrals

## Context
Any one-sided sub entity integral boils down to getting the correct (cell, local_entity_index), and assemble over these.
The lower level assembly routines (assemble_exterior_facet, assemble_vertex) does not really differ in this respect. It is only due to the fact that facets might require permutations that codes were not identical.
This became crystal clear when introducing ridge  integrals.

## Consequences
- The `assemble_vertex` routines recently introduced are removed in favor of an `assemble_exterior_facet` function, now renamed as `assemble_exterior_entities`. This is also used for ridge integrals.

## Bugs encountered
- There is a subtle bug in the coefficient allocation of vertex integrals, where the double amount of memory is allocated. Adjusted packing to divide by two for all integral types but cells.

## Future work
Mixed dimensional support for codim-2.
For this to work, a few things have to be introduced:
- require_ridge_permutation has to be introduced in FFCx
- Computing ridge-permutation has to be a member of DOLFINx (similar to compute facet permutations)

### DIFF
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -364,6 +364,9 @@ def form(
 
         # For each argument in form extract its function space
         V = [arg.ufl_function_space()._cpp_object for arg in form.arguments()]
+        part = form_compiler_options.get("part", "full")
+        if part == "diagonal":
+            V = [V[0]]
 
         # Prepare coefficients data. For every coefficient in form take
         # its C++ object.
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -23,7 +23,7 @@ jobs:
     env:
       PETSC_ARCH: arch-darwin-c-opt
       PETSC_DIR: ${{ github.workspace }}/petsc
-      PETSC_VERSION: 3.23.6
+      PETSC_VERSION: 3.24.0
       PYTEST_ADDOPTS: "-W error"
 
     steps:
--- a/.github/workflows/sonarcloud.yml
+++ b/.github/workflows/sonarcloud.yml
@@ -17,7 +17,7 @@ jobs:
     container: fenicsproject/test-env:current-mpich
     env:
       SONAR_SCANNER_VERSION:
-        6.1.0.4477 # Find the latest version at:
+        7.2.0.5079 # Find the latest version at:
         # https://github.com/SonarSource/sonar-scanner-cli/tags
       SONAR_SERVER_URL: "https://sonarcloud.io"
       BUILD_WRAPPER_OUT_DIR: build_wrapper_output_directory # Directory where build-wrapper output will be placed
--- a/cpp/dolfinx/fem/DirichletBC.h
+++ b/cpp/dolfinx/fem/DirichletBC.h
@@ -255,7 +255,7 @@ class DirichletBC
 {
 private:
   /// Compute number of owned dofs indices. Will contain 'gaps' for
-  /// sub-spaces.
+  /// sub-spaces. The dofs must be unrolled.
   std::size_t num_owned(const DofMap& dofmap,
                         std::span<const std::int32_t> dofs)
   {
@@ -325,8 +325,7 @@ public:
                                    std::vector<std::int32_t>>
   DirichletBC(std::shared_ptr<const Constant<T>> g, X&& dofs,
               std::shared_ptr<const FunctionSpace<U>> V)
-      : _function_space(V), _g(g), _dofs0(std::forward<X>(dofs)),
-        _owned_indices0(num_owned(*V->dofmap(), _dofs0))
+      : _function_space(V), _g(g), _dofs0(std::forward<X>(dofs))
   {
     assert(g);
     assert(V);
@@ -352,10 +351,9 @@ public:
 
     // Unroll _dofs0 if dofmap block size > 1
     if (const int bs = V->dofmap()->bs(); bs > 1)
-    {
-      _owned_indices0 *= bs;
       _dofs0 = unroll_dofs(_dofs0, bs);
-    }
+
+    _owned_indices0 = num_owned(*_function_space->dofmap(), _dofs0);
   }
 
   /// @brief Create a representation of a Dirichlet boundary condition
@@ -375,17 +373,15 @@ public:
                                    std::vector<std::int32_t>>
   DirichletBC(std::shared_ptr<const Function<T, U>> g, X&& dofs)
       : _function_space(g->function_space()), _g(g),
-        _dofs0(std::forward<X>(dofs)),
-        _owned_indices0(num_owned(*_function_space->dofmap(), _dofs0))
+        _dofs0(std::forward<X>(dofs))
   {
     assert(_function_space);
 
     // Unroll _dofs0 if dofmap block size > 1
     if (const int bs = _function_space->dofmap()->bs(); bs > 1)
-    {
-      _owned_indices0 *= bs;
       _dofs0 = unroll_dofs(_dofs0, bs);
-    }
+
+    _owned_indices0 = num_owned(*_function_space->dofmap(), _dofs0);
   }
 
   /// @brief Create a representation of a Dirichlet boundary condition
--- a/docker/Dockerfile.test-env
+++ b/docker/Dockerfile.test-env
@@ -22,8 +22,8 @@ ARG KAHIP_VERSION=3.18
 # the most recent Numba release, see
 # https://numba.readthedocs.io/en/stable/user/installing.html#version-support-information
 ARG NUMPY_VERSION=2.2.6
-ARG PETSC_VERSION=3.23.5
-ARG SLEPC_VERSION=3.23.2
+ARG PETSC_VERSION=3.24.0
+ARG SLEPC_VERSION=3.24.0
 ARG SPDLOG_VERSION=1.15.3
 
 ARG MPICH_VERSION=4.3.1
--- a/python/demo/demo_axis.py
+++ b/python/demo/demo_axis.py
@@ -2,10 +2,13 @@
 #
 # Copyright (C) 2022 Michele Castriotta, Igor Baratta, Jørgen S. Dokken
 #
-# This demo is implemented in three files: one for the mesh generation
-# with gmsh, one for the calculation of analytical efficiencies, and one
-# for the variational forms and the solver. It illustrates how to:
+# ```{admonition} Download sources
+# :class: download
+# * {download}`Python script <./demo_axis.py>`
+# * {download}`Jupyter notebook <./demo_axis.ipynb>`
+# ```
 #
+# This demo illustrates how to:
 # - Setup and solve Maxwell's equations for axisymmetric geometries
 # - Implement (axisymmetric) perfectly matched layers
 #
@@ -388,7 +391,7 @@ def create_eps_mu(pml, rho, eps_bkg, mu_bkg):
 
 # We can now define some constants and geometrical parameters, and then
 # we can generate the mesh with Gmsh, by using the function
-# `generate_mesh_sphere_axis` in `mesh_sphere_axis.py`:
+# `generate_mesh_sphere_axis`:
 
 
 # +
--- a/python/demo/demo_biharmonic.py
+++ b/python/demo/demo_biharmonic.py
@@ -14,9 +14,12 @@
 
 # # Biharmonic equation
 #
-# This demo is implemented in a single Python file,
-# {download}`demo_biharmonic.py`, which contains both the variational forms
-# and the solver. It illustrates how to:
+# ```{admonition} Download sources
+# :class: download
+# * {download}`Python script <./demo_biharmonic.py>`
+# * {download}`Jupyter notebook <./demo_biharmonic.ipynb>`
+# ```
+# This demo illustrates how to:
 #
 # - Solve a linear partial differential equation
 # - Use a discontinuous Galerkin method
@@ -143,6 +146,9 @@ V = fem.functionspace(msh, ("Lagrange", 2))
 # degree)`, where `family` is the finite element family, and `degree`
 # specifies the polynomial degree. in this case `V` consists of
 # second-order, continuous Lagrange finite element functions.
+# For further details of how one can specify
+# finite elements as tuples, see {py:class}`ElementMetaData
+# <dolfinx.fem.ElementMetaData>`.
 #
 # Next, we locate the mesh facets that lie on the boundary
 # $\Gamma_D = \partial\Omega$.
--- a/python/demo/demo_cahn-hilliard.py
+++ b/python/demo/demo_cahn-hilliard.py
@@ -10,6 +10,11 @@
 
 # # Cahn-Hilliard equation
 #
+# ```{admonition} Download sources
+# :class: download
+# * {download}`Python script <./demo_cahn-hilliard.py>`
+# * {download}`Jupyter notebook <./demo_cahn-hilliard.ipynb>`
+# ```
 # This example demonstrates the solution of the Cahn-Hilliard equation,
 # a nonlinear, time-dependent fourth-order PDE.
 #
@@ -21,8 +26,6 @@
 # - Interpolation of functions
 # - Visualisation of a running simulation with
 #   [PyVista](https://pyvista.org/)
-#
-# This demo is implemented in {download}`demo_cahn-hilliard.py`.
 #
 # ## Equation and problem definition
 #
@@ -113,9 +116,6 @@
 # - $\theta = 0.5$
 #
 # ## Implementation
-#
-# This demo is implemented in the {download}`demo_cahn-hilliard.py`
-# file.
 
 # +
 import os
@@ -154,6 +154,7 @@ log.set_output_file("log.txt")
 lmbda = 1.0e-02  # surface parameter
 dt = 5.0e-06  # time step
 theta = 0.5  # time stepping family, e.g. theta=1 -> backward Euler, theta=0.5 -> Crank-Nicholson
+t = 0.0  # Current time
 
 # A unit square mesh with 96 cells edges in each direction is created,
 # and on this mesh a
@@ -197,8 +198,6 @@ c0, mu0 = ufl.split(u0)
 # The initial conditions are interpolated into a finite element space:
 
 # +
-# Zero u
-u.x.array[:] = 0.0
 
 # Interpolate initial condition
 rng = np.random.default_rng(42)
@@ -230,8 +229,7 @@ dfdc = ufl.diff(f, c)
 #
 # It is convenient to introduce an expression for $\mu_{n+\theta}$:
 
-# mu_(n+theta)
-mu_mid = (1.0 - theta) * mu0 + theta * mu
+mu_mid = (1.0 - theta) * mu0 + theta * mu  # mu_(n+theta)
 
 # which is then used in the definition of the variational forms:
 
@@ -256,9 +254,7 @@ F = F0 + F1
 #
 # To solve the nonlinear system of equations,
 # {py:class}`NonlinearProblem<dolfinx.fem.petsc.NonlinearProblem>` object
-# to solve a system of nonlinear equations
-
-# +
+# to solve a system of nonlinear equations.
 # For the factorisation of the underlying linearized problems, prefer
 # MUMPS, then superlu_dist, then default.
 # We measure convergence by looking at the norm of the increment of the
@@ -266,6 +262,7 @@ F = F0 + F1
 # [`SNES convegence tests`](https://petsc.org/release/manual/snes/#convergence-tests)
 # for further details.
 
+# +
 use_superlu = PETSc.IntType == np.int64  # or PETSc.ScalarType == np.complex64
 sys = PETSc.Sys()  # type: ignore
 if sys.hasExternalPackage("mumps") and not use_superlu:
@@ -290,29 +287,19 @@ problem = NonlinearProblem(
 )
 # -
 
-# To run the solver and save the output to a VTK file for later
-# visualization, the solver is advanced in time from $t_{n}$ to
-# $t_{n+1}$ until a terminal time $T$ is reached:
+# We prepare output files and pyvista for time-dependent visualization:
 
-# +
-# Output file
-file = XDMFFile(MPI.COMM_WORLD, "demo_ch/output.xdmf", "w")
+file = XDMFFile(MPI.COMM_WORLD, "demo_ch/output.xdmf", "w")  # Output file
 file.write_mesh(msh)
 
-# Step in time
-t = 0.0
-
-#  Reduce run time if on test (CI) server
-if "CI" in os.environ.keys() or "GITHUB_ACTIONS" in os.environ.keys():
-    T = 3 * dt
-else:
-    T = 50 * dt
-
 # Get the sub-space for c and the corresponding dofs in the mixed space
-# vector
+# vector. This is used for visualization on the collapsed subspace with
+# pyvista.
+
 V0, dofs = ME.sub(0).collapse()
 
 # Prepare viewer for plotting the solution during the computation
+
 if have_pyvista:
     # Create a VTK 'mesh' with 'nodes' at the function dofs
     topology, cell_types, x = plot.vtk_mesh(V0)
@@ -327,6 +314,17 @@ if have_pyvista:
     p.view_xy(True)
     p.add_text(f"time: {t}", font_size=12, name="timelabel")
 
+
+# Reduce run time if on test (CI) server
+if "CI" in os.environ.keys() or "GITHUB_ACTIONS" in os.environ.keys():
+    T = 3 * dt
+else:
+    T = 50 * dt
+
+# The solver is advanced in time from $t_{n}$ to
+# $t_{n+1}$ until a terminal time $T$ is reached
+
+# +
 c = u.sub(0)
 u0.x.array[:] = u.x.array
 step = 0
@@ -348,10 +346,11 @@ while t < T:
         p.app.processEvents()
 
 file.close()
+# -
 
-# Update ghost entries and plot
+# Update plot
+
 if have_pyvista:
-    u.x.scatter_forward()
     grid.point_data["c"] = u.x.array[dofs].real
     screenshot = None
     if pv.OFF_SCREEN:
--- a/python/demo/demo_comm-pattern.py
+++ b/python/demo/demo_comm-pattern.py
@@ -10,10 +10,15 @@
 
 # # Parallel communication pattern analysis
 #
-# This demo is implemented in {download}`demo_comm-pattern.py`. It
-# illustrates how build a graph that represents a parallel communication
-# pattern and how to analyse the parallel communication pattern using
-# [NetworkX](https://networkx.org/).
+# ```{admonition} Download sources
+# :class: download
+# * {download}`Python script <./demo_comm-pattern.py>`
+# * {download}`Jupyter notebook <./demo_comm-pattern.ipynb>`
+# ```
+# This demo illustrates how to:
+# - Build a graph that represents a parallel communication pattern
+# - Analyse the parallel communication pattern using
+#   [NetworkX](https://networkx.org/).
 #
 # The layout of a distributed array across processes (MPI ranks) is
 # described in DOLFINx by an {py:class}`IndexMap
--- a/python/demo/demo_elasticity.py
+++ b/python/demo/demo_elasticity.py
@@ -12,10 +12,14 @@
 #
 # Copyright © 2020-2022 Garth N. Wells and Michal Habera
 #
-# This demo ({download}`demo_elasticity.py`) solves the equations of
-# static linear elasticity using a smoothed aggregation algebraic
-# multigrid solver. It illustrates how to:
-#
+# ```{admonition} Download sources
+# :class: download
+# * {download}`Python script <./demo_elasticity.py>`
+# * {download}`Jupyter notebook <./demo_elasticity.ipynb>`
+# ```
+# This demo solves the equations of static linear elasticity using
+# a smoothed aggregation algebraic multigrid solver.
+# It illustrates how to:
 # - Use a smoothed aggregation algebraic multigrid solver
 # - Use {py:class}`Expression <dolfinx.fem.Expression>` to compute
 #   derived quantities of a solution
@@ -94,7 +98,7 @@ def build_nullspace(V: FunctionSpace):
 
 # ## Problem definition
 
-# Create a box Mesh:
+# Create a {py:func}`box mesh<dolfinx.mesh.create_box>`:
 
 
 msh = create_box(
@@ -258,8 +262,9 @@ with XDMFFile(msh.comm, "out_elasticity/von_mises_stress.xdmf", "w") as file:
 # -
 
 # Finally, we compute the $L^2$ norm of the displacement solution
-# vector. This is a collective operation (i.e., the method `norm` must
-# be called from all MPI ranks), but we print the norm only on rank 0.
+# vector. This is a collective operation (i.e., the method
+# {py:func}`norm<dolfinx.la.norm>` must be called from all MPI ranks),
+# but we print the norm only on rank 0.
 
 # +
 unorm = la.norm(uh.x)
--- a/python/demo/demo_gmsh.py
+++ b/python/demo/demo_gmsh.py
@@ -12,8 +12,13 @@
 #
 # Copyright (C) 2020-2023 Garth N. Wells and Jørgen S. Dokken
 #
+# ```{admonition} Download sources
+# :class: download
+# * {download}`Python script <./demo_gmsh.py>`
+# * {download}`Jupyter notebook <./demo_gmsh.ipynb>`
+# ```
+
 # This demo shows how to create meshes using the Gmsh Python interface.
-# It is implemented in {download}`demo_gmsh.py`.
 #
 # The Gmsh module is required for this demo.
 
--- a/python/demo/demo_half_loaded_waveguide.py
+++ b/python/demo/demo_half_loaded_waveguide.py
@@ -17,18 +17,19 @@
 
 # Copyright (C) 2022 Michele Castriotta, Igor Baratta, Jørgen S. Dokken
 #
-# This demo is implemented in two files, one for defining and solving
-# the eigenvalue problem for a half-loaded electromagnetic waveguide
-# with perfect electric conducting walls, and one for verifying if the
-# numerical eigenvalues are consistent with the analytical modes of the
-# problem.
+# ```{admonition} Download sources
+# :class: download
+# * {download}`Python script <./demo_half_loaded_waveguide.py>`
+# * {download}`Jupyter notebook <./demo_half_loaded_waveguide.ipynb>`
+# ```
 #
 # The demo shows how to:
-#
 # - Setup an eigenvalue problem for Maxwell's equations
+# - Setup a half-loaded electromagnetic waveguide with perfect electric
+#   conducting walls
 # - Use SLEPc for solving eigenvalue problems
+# - Verification of numerical eigenvalues with analytical modes
 #
-
 # ## Equations and problem definition
 #
 # In this demo, we are going to show how to solve the eigenvalue problem
@@ -407,6 +408,7 @@ eps.solve()
 eps.view()
 eps.errorView()
 
+# ## Verification of computed eigenvalues
 # Now we can get the eigenvalues and eigenvectors calculated by SLEPc
 # with the following code. We also verify if the numerical $k_z$ are
 # consistent with the analytical equations of the half-loaded waveguide
--- a/python/demo/demo_hdg.py
+++ b/python/demo/demo_hdg.py
@@ -10,14 +10,20 @@
 
 # # HDG scheme for the Poisson equation
 #
-# This demo is implemented in {download}`demo_hdg.py`. It
-# illustrates how to:
+# ```{admonition} Download sources
+# :class: download
+# * {download}`Python script <./demo_hdg.py>`
+# * {download}`Jupyter notebook <./demo_hdg.ipynb>`
+# ```
+# This demo illustrates how to:
 #
 # - Solve Poisson's equation using an HDG scheme.
+# - Defining custom integration domains
+# - Create a submesh over all facets of the mesh
+# - Use `ufl.MixedFunctionSpace` to defined blocked problems.
+# - Assemble mixed systems with multiple, related meshes
 
 # +
-import sys
-
 from mpi4py import MPI
 from petsc4py import PETSc
 
@@ -28,22 +34,40 @@ import ufl
 from dolfinx import fem, mesh
 from dolfinx.cpp.mesh import cell_num_entities
 from dolfinx.fem import extract_function_spaces
-from dolfinx.fem.petsc import apply_lifting, assemble_matrix, assemble_vector, set_bc
+from dolfinx.fem.petsc import (
+    apply_lifting,
+    assemble_matrix,
+    assemble_vector,
+    assign,
+    create_vector,
+    set_bc,
+)
+
+# -
 
 
-def par_print(comm, string):
-    if comm.rank == 0:
-        print(string)
-        sys.stdout.flush()
+# We start by creating two convenience functions: One to compute
+# the L2 norm of a UFL expression, and one to define the integration
+# domains for the facets of all cells in a mesh.
 
 
-def norm_L2(comm, v, measure=ufl.dx):
-    return np.sqrt(
-        comm.allreduce(fem.assemble_scalar(fem.form(ufl.inner(v, v) * measure)), op=MPI.SUM)
-    )
+def norm_L2(v: ufl.core.expr.Expr, measure: ufl.Measure = ufl.dx) -> np.inexact:
+    """Convenience function to compute the L2 norm of a UFL expression."""
+    compiled_form = fem.form(ufl.inner(v, v) * measure)
+    comm = compiled_form.mesh.comm
+    return np.sqrt(comm.allreduce(fem.assemble_scalar(compiled_form), op=MPI.SUM))
 
 
-def compute_cell_boundary_facets(msh):
+# In DOLFINx, we represent integration domains over entities of
+# codimension > 0 as a tuple `(cell_idx, local_entity_idx)`,
+# where `cell_idx` is the index of the cell in the mesh
+# (local to process), and `local_entity_idx` is the local index
+# of the sub entity (local to cell). For the HDG scheme, we will
+# integrate over the facets of each cell (for internal facets,
+# there will be repeat entries, from the viewpoint of the connected cells).
+
+
+def compute_cell_boundary_facets(msh: dolfinx.mesh.Mesh) -> np.ndarray:
     """Compute the integration entities for integrals around the
     boundaries of all cells in msh.
 
@@ -73,15 +97,15 @@ comm = MPI.COMM_WORLD
 rank = comm.rank
 dtype = PETSc.ScalarType
 
-# Number of elements in each direction
-n = 8
-
 # Create the mesh
+
+n = 8  # Number of elements in each direction
 msh = mesh.create_unit_cube(comm, n, n, n, ghost_mode=mesh.GhostMode.none)
 
 # We need to create a broken Lagrange space defined over the facets of
 # the mesh. To do so, we require a sub-mesh of the all facets. We begin
 # by creating a list of all of the facets in the mesh
+
 tdim = msh.topology.dim
 fdim = tdim - 1
 msh.topology.create_entities(fdim)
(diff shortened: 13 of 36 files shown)


cpp/dolfinx/fem/Form.h
```diff
@@ -40,7 +40,8 @@ enum class IntegralType : std::int8_t
   cell = 0,           ///< Cell
   exterior_facet = 1, ///< Exterior facet
   interior_facet = 2, ///< Interior facet
-  vertex = 3          ///< Vertex
+  vertex = 3,         ///< Vertex
+  ridge = 4           ///< Ridge
```
schnellerhase, line 44: Should we maybe switch this order? Feels more natural to have
1. cell - `tdim`
2. facet - `tdim -1`
3. ridge - `tdim-2`
4. vertex - 0
 Would require a change in FFCx as well.

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -290,13 +255,38 @@ T assemble_scalar(
...
                    md::extents<std::size_t, md::dynamic_extent, 2>>(
             vertices.data(), vertices.size() / shape1, shape1),
         fn, constants,
-        md::mdspan(coeffs.data(), vertices.size() / shape1, cstride));
+        md::mdspan(coeffs.data(), vertices.size() / shape1, cstride),
+        vertex_perms);
+  }
+
+  md::mdspan<const std::uint8_t, md::dextents<std::size_t, 2>> ridge_perms;
+  for (int i = 0; i < M.num_integrals(IntegralType::ridge, 0); ++i)
```
schnellerhase: Can we combine these loops here as well now? Something like `for (auto exterior_cell_type : {vertices, ridge})` (and even `exterior_facet`)?

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -1466,34 +1379,85 @@ void assemble_vector(
...
+            P0, b, x_dofmap, x, vertices, {dofs, bs, vertices1}, fn, constants,
+            md::mdspan(coeffs.data(), vertices.size() / 2, cstride), cell_info0,
+            vertex_perms);
+      }
+    }
+
+    // NOTE: Empty for now, but I guess we require this for mixed-dimensional
+    // assembly
+    md::mdspan<const std::uint8_t, md::dextents<std::size_t, 2>> ridge_perms;
+    for (int i = 0; i < L.num_integrals(IntegralType::ridge, 0); ++i)
```
schnellerhase: Same comment.

cpp/dolfinx/fem/utils.h
```diff
@@ -838,6 +846,86 @@ Form<T, U> create_form_factory(
     }
   }
 
+  // Attach ridge kernels
```
schnellerhase: Combinable? As above.

schnellerhase (commented):
Nice code duplication removal!

jorgensd:
> The unification reads really well.
> 
> One point to consider, for future design, given the increase in type of integral we are seeing: Vertex integrals are effectively defined by dimension (vertex: dim=0), while all other come with a codimension (cell: codim=0, facets: codim=-1, ridge: codim=-2). The root of this is also in UFL, and surely we want to keep it for user convenience, but we should try to separate the logic early enough, strip all data to a common pattern for both dim/codim and follow unified codepath onwards. I think your PR is the step in right direction as it moves all integral assemble routines to the codimension format. However, we should also avoid `IntegralType` with duplicate values as much as possible: in 2D ridge integral in == vertex integral.

I think if we ever introduce "peak" integrals, they would effectively replace the vertex integrals, as we should always go down the "co-dimension" route (for the sake of mixed assembly support).

cpp/dolfinx/fem/assemble_scalar_impl.h
jorgensd: I'll see what i can do:)

cpp/dolfinx/fem/Form.h
jorgensd: I think we do this in the future.

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -57,41 +57,41 @@ T assemble_cells(mdspan2_t x_dofmap,
...
 
     // Get cell coordinates/geometry
     auto x_dofs = md::submdspan(x_dofmap, cell, md::full_extent);
     for (std::size_t i = 0; i < x_dofs.size(); ++i)
       std::copy_n(&x(x_dofs[i], 0), 3, std::next(cdofs.begin(), 3 * i));
 
     // Permutations
-    std::uint8_t perm = perms.empty() ? 0 : perms(cell, local_facet);
-    fn(&value, &coeffs(f, 0), constants.data(), cdofs.data(), &local_facet,
+    std::uint8_t perm = perms.empty() ? 0 : perms(cell, local_entity);
```
schnellerhase, line 100: This is (sometimes) compile time information, but now it is deduced at runtime. Change to `std::optional` permutation argument?  `has_value` is a `constexpr` check.

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -272,31 +212,40 @@ T assemble_scalar(
...
-    assert(vertices.size() * cstride == coeffs.size());
-
-    constexpr std::size_t num_adjacent_cells = 1;
-    // Two values per adj. cell (cell index and local vertex index).
-    constexpr std::size_t shape1 = 2 * num_adjacent_cells;
+    md::mdspan<const std::uint8_t, md::dextents<std::size_t, 2>> perms;
+    if (itg_type == fem::IntegralType::exterior_facet && !facet_perms.empty())
+      perms = facet_perms;
+    else
+      perms = md::mdspan<const std::uint8_t, md::dextents<std::size_t, 2>>();
```
schnellerhase: Switch over `itg_type`.

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -1466,34 +1345,53 @@ void assemble_vector(
...
-                     md::extents<std::size_t, md::dynamic_extent, 2>>(
-              vertices.data(), vertices.size() / 2, 2),
-          {dofs, bs,
-           md::mdspan<const std::int32_t,
-                      md::extents<std::size_t, md::dynamic_extent, 2>>(
-               vertices0.data(), vertices0.size() / 2, 2)},
-          fn, constants,
-          md::mdspan(coeffs.data(), vertices.size() / 2, cstride), cell_info0);
+      md::mdspan<const std::uint8_t, md::dextents<std::size_t, 2>> perms;
+      if (itg_type == fem::IntegralType::exterior_facet && !facet_perms.empty())
```
schnellerhase: Switch over `itg_type`.

cpp/dolfinx/fem/utils.h
```diff
@@ -757,82 +696,106 @@ Form<T, U> create_form_factory(
...
-  // Attach vertex kernels
+  // Attach exterior entity integrals
   {
-    for (std::size_t form_idx = 0; form_idx < ufcx_forms.size(); ++form_idx)
+    for (IntegralType itg_type : {IntegralType::exterior_facet,
+                                  IntegralType::vertex, IntegralType::ridge})
     {
-      const ufcx_form& form = ufcx_forms[form_idx];
+      std::size_t dim;
+      if (itg_type == IntegralType::exterior_facet)
```
schnellerhase: Switch over `itg_type`.

cpp/dolfinx/fem/assemble_scalar_impl.h
jorgensd: This is only know at compile time for the vertex integrals, so i don’t see the point of doing it in general.
Once we get mixed dimensional ridges it is no longer compile time.

cpp/dolfinx/fem/assemble_scalar_impl.h
jorgensd: Wouldn't that just add a lot of extra lines here?
If there had been many special cases (which will happen with ridge mixed integrals and potentially peak integrals) it is natural to go over to a switch.

schnellerhase: The problem I see is `else` catching everything and not being verbose about which integral types are handled in this manner and whether this is intentional.

cpp/dolfinx/fem/utils.h
jorgensd: Done

cpp/dolfinx/fem/assemble_scalar_impl.h
jorgensd: Since it is within the for-loop where the entries are hard-coded, it is very unlikely that one should ever get there.

cpp/dolfinx/fem/assemble_scalar_impl.h
schnellerhase: `std::mdpsan::empty` is also constexpr - so with C++23 we get this for free.

cpp/dolfinx/fem/assemble_scalar_impl.h
schnellerhase: Looking at it again, I think the `else` branch is not even needed here, initialised already to default constructed .

jorgensd: > Looking at it again, I think the `else` branch is not even needed here, initialised already to default constructed .

The problem is that if one parses exterior facets prior to ridges or vertices it is set to non-zero. That is why the else branch is needed. 
It will become even more clear with mixed-dimensional ridge integrals.

jorgensd: Never mind, I forgot my own coding structure

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -191,14 +191,14 @@ void assemble_cells_matrix(
 /// @param[in] perms Facet permutation integer. Empty if facet
 /// permutations are not required.
 template <dolfinx::scalar T>
-void assemble_exterior_facets(
+void assemble_exterior_entities(
```
garth-wells: What is an 'exterior entity'?

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -685,7 +653,54 @@ void assemble_matrix(
...
-          cell_info0, cell_info1, perms);
+          cell_info0, cell_info1, facet_perms);
+    }
+
+    for (fem::IntegralType itg_type :
+         {fem::IntegralType::exterior_facet, fem::IntegralType::vertex,
+          fem::IntegralType::ridge})
+    {
+      md::mdspan<const std::uint8_t, md::dextents<std::size_t, 2>> perms;
+      switch (itg_type)
```
garth-wells: This switch looks odd. Can't it be a ternary conditional operator?

garth-wells:
I don't like `dr`. We should fix it properly with something like `dx(codim=2)`. 

`dx`, `ds` and `dS` is manageable if imperfect, but doesn't scale for more types.

jorgensd:
> I don't like `dr`. We should fix it properly with something like `dx(codim=2)`.
> 
> `dx`, `ds` and `dS` is manageable if imperfect, but doesn't scale for more types.

`dr` is a ufl concept that I merged over half a year ago. It perfectly fits the co-dimensionality of entities.
Adding special handling for co-dimensions other than 1 seems wrong.

As already pointed out in this PR, the vertex integrals should be replaced by `dr` and `dp` (peak integrals), such that the integral over a vertex in 2D becomes `dr`, while in 3D a vertex integral is `dp`.

cpp/dolfinx/fem/assemble_matrix_impl.h
jorgensd: An exterior entity is any sub-entity of a cell; peak, ridge or facet, viewed from a single cell.
In DOLFINx, we view every integration entity as a tuple (cell, local_entity_index).
This assembler now works for any co-dimension as illustrated by the simplifications within the code.

cpp/dolfinx/fem/assemble_matrix_impl.h
jorgensd: This is made like this to support the future 3D-1D coupling, i.e. ridge integrals,
which I've started on in: https://github.com/FEniCS/dolfinx/pull/3904

cpp/dolfinx/fem/assemble_matrix_impl.h
garth-wells: In that case the terminology is confusing and should be improved.

jorgensd: We can call it `assemble_one_sided_entity`? As it indicates that it is from the view-point of a single cell.

I simply used similar terminology as already existed, with `exterior_facets`.

schnellerhase: How about `assemble_entities_over_cells`? With the extension being something like `assemble_entities_over_patches` for patches.

jorgensd:
> I don't like `dr`. We should fix it properly with something like `dx(codim=2)`.
> 
> `dx`, `ds` and `dS` is manageable if imperfect, but doesn't scale for more types.

Your point of not liking `dr` is a tangential issue, that should be handled by a rewrite of integration measures in UFL.
Then all integration measures should be handled as `dx` with a special codim argument, i.e.
- `dx(codim=0)` integral over cells
- `dx(codim=1) = ds` integral over facets from the viewpoint of a single cell, default is exterior facets.
- `dx(codim=2) = dr` integral over ridges from the viewpoint of a single cell.
- `dx(codmin=3) = dPeak` integral over peaks from the viewpoint of a single cell.

The current integration measure `dP` (vertex integrals) does not fit in this abstraction, as it will be `dx(codim=1) `, `dx(codim=2)` or `dx(codim=3)` for meshes with topological dimension 1, 2 and 3 respectively, exactly like what I propose in this PR.

This PR only unifies the already supported integration measures in FFCx to work in DOLFINx, without having code duplication everywhere.

cpp/dolfinx/fem/assemble_matrix_impl.h
garth-wells: A design approach in DOLFINx has been to keep things simple and avoid designing for hypotheticals. Changes can be made as needed to extend functionality.

mscroggs:
Long term, I agree that introducing `dr` is probably not the best thing: UFL has too many integral measures that have patchy support and are confusing. But given that we'll need a fair bit of work in UFL to tidy up measure names before we can improve that, I think it may be worth merging this for now then revisiting later once someone's had time to do the necessary work in UFL

cpp/dolfinx/fem/Form.h
```diff
@@ -40,7 +40,8 @@ enum class IntegralType : std::int8_t
   cell = 0,           ///< Cell
   exterior_facet = 1, ///< Exterior facet
```
garth-wells: Having `exterior_facet` is inconsistent with `vertex` and `ridge`. Better name would be `facet`, with `interior_facet` being a special case.

Would it be better to remove the enum and use `(n, codim)` pairs? At present there are ambiguities between facet, ridge and vertex in dimensions less than 3.

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -156,16 +156,16 @@ void assemble_cells_matrix(
   }
 }
 
-/// @brief Execute kernel over exterior facets and accumulate result in
+/// @brief Execute kernel over a set of entities and accumulate result in
 /// a matrix.
 ///
```
garth-wells, line 169:
```
/// Each entity is represented by (i) a cell that the entity is attached to 
/// and (ii) the local index of the entity  with respect to the cell. The 
/// kernel is executed for each entity. The kernel can access data 
/// (e.g., coefficients, basis functions) associated with the attached cell. However, 
/// entities may be attached to more than one cell. This function therefore 
/// computes 'one-sided' integrals, i.e. evaluates integrals as seen from cell used to
/// define the entity.  
```

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -156,16 +156,16 @@ void assemble_cells_matrix(
   }
 }
 
-/// @brief Execute kernel over exterior facets and accumulate result in
+/// @brief Execute kernel over a set of entities and accumulate result in
```
garth-wells:
```suggestion
/// @brief Execute kernel over entities of codimension > 1 and accumulate result in
```

jpdean:
I think this is a natural way to unify the assembly of integrals over cell sub-entities that only require data from one cell at a time. It’s what we’ve had in mind for a while to deal with more general mixed-dimensional problems.

I think we should avoid names like “exterior_entities”, which doesn’t really make sense. Since this assembler can only be used when we only need data from a single cell, in contrast to a patch of cells, a name that reflects this would be better. 

There are quite a lot of mysterious measures in UFL. If we don’t want to go down the path of adding `dr` and `dp` (which is the natural extension of what we have already with `dx` and `ds` etc.), then I agree that we should change `dx`, `ds`, and `dS` so that everything is consistent with the new approach. This is (probably) a lot of work in UFL, though.

Maybe as a first step, this PR could be split into two: a part that generalises the assembler and a part that adds ridge integrals and `dr` etc? The generalisation of the assembler is quite natural and shouldn't depend too heavily on how we define the integrals in UFL. This way, we can get the nice generalisation merged first and have more time to think about what the right way to handle measures is in UFL.

I haven’t been through the PR in detail yet, but I’ll have a look tomorrow!

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -685,7 +653,54 @@ void assemble_matrix(
            mdspanx22_t(facets1.data(), facets1.size() / 4, 2, 2)},
           P1T, bc0, bc1, fn,
           mdspanx2x_t(coeffs.data(), facets.size() / 4, 2, cstride), constants,
-          cell_info0, cell_info1, perms);
+          cell_info0, cell_info1, facet_perms);
+    }
+
+    for (fem::IntegralType itg_type :
```
garth-wells: Could use `auto` here - the type is obvious.

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -55,41 +55,41 @@ T assemble_cells(mdspan2_t x_dofmap,
   return value;
 }
 
-/// Execute kernel over exterior facets and accumulate result
+/// Execute kernel over a set of entities and accumulate result
```
garth-wells: See suggested doc string for matrixes.

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -55,41 +55,41 @@ T assemble_cells(mdspan2_t x_dofmap,
...
 template <dolfinx::scalar T>
-T assemble_exterior_facets(
+T assemble_entities_over_cells(
```
garth-wells:
```suggestion
T assemble_entities(
```

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -273,31 +211,47 @@ T assemble_scalar(
         md::mdspan<const T, md::extents<std::size_t, md::dynamic_extent, 2,
                                         md::dynamic_extent>>(
             coeffs.data(), facets.size() / shape1, 2, cstride),
-        perms, cdofs_b);
+        facet_perms, cdofs_b);
   }
 
-  for (int i = 0; i < M.num_integrals(IntegralType::vertex, 0); ++i)
+  for (fem::IntegralType itg_type :
```
garth-wells: See comment for matrices.

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -273,31 +211,47 @@ T assemble_scalar(
...
-    constexpr std::size_t shape1 = 2 * num_adjacent_cells;
-
-    value += impl::assemble_vertices(
-        x_dofmap, x,
-        md::mdspan<const std::int32_t,
-                   md::extents<std::size_t, md::dynamic_extent, 2>>(
-            vertices.data(), vertices.size() / shape1, shape1),
-        fn, constants,
-        md::mdspan(coeffs.data(), vertices.size() / shape1, cstride), cdofs_b);
+    md::mdspan<const std::uint8_t, md::dextents<std::size_t, 2>> perms;
```
garth-wells: Make this simple with a ternary.

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -273,31 +211,47 @@ T assemble_scalar(
...
+    {
+      auto fn = M.kernel(itg_type, i, 0);
+      assert(fn);
+      auto& [coeffs, cstride] = coefficients.at({itg_type, i});
+
+      std::span entities = M.domain(itg_type, i, 0);
+
+      constexpr std::size_t num_adjacent_cells = 1;
+      // Two values per each adj. cell (cell index and local entity index).
+      constexpr std::size_t shape1 = 2 * num_adjacent_cells;
```
garth-wells: This is trivial - why is it needed?

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -263,15 +263,15 @@ void _lift_bc_cells(
 template <typename V,
           dolfinx::scalar T = typename std::remove_cvref_t<V>::value_type>
   requires std::is_same_v<typename std::remove_cvref_t<V>::value_type, T>
-void _lift_bc_exterior_facets(
+void _lift_bc_exterior_entities(
```
garth-wells:
```suggestion
void _lift_bc_entities(
```

And update docstring.

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -718,7 +718,8 @@ void assemble_cells(
   }
 }
 
-/// @brief Execute kernel over cells and accumulate result in vector.
+/// @brief Execute kernel over a set of entities and accumulate result in
```
garth-wells: See docstring suggestions for matrix assembly.

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -740,19 +741,19 @@ void assemble_cells(
...
 /// @param[in] cell_info0 The cell permutation information for the test
 /// function mesh.
-/// @param[in] perms Facet permutation integer. Empty if facet
+/// @param[in] perms Entity permutation integer. Empty if entity
 /// permutations are not required.
 template <int _bs = -1, typename V,
           dolfinx::scalar T = typename std::remove_cvref_t<V>::value_type>
   requires std::is_same_v<typename std::remove_cvref_t<V>::value_type, T>
-void assemble_exterior_facets(
+void assemble_entities_over_cells(
```
garth-wells:
```suggestion
void assemble_entities(
```

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -1192,7 +1081,47 @@ void lift_bc(V&& b, const Form<T, U>& a, mdspan2_t x_dofmap,
         b, x_dofmap, x, kernel, facets, {dofmap0, bs0, facets0}, P0,
         {dofmap1, bs1, facets1}, P1T, constants,
         mdspanx2x_t(coeffs.data(), facets.extent(0), 2, cstride), cell_info0,
-        cell_info1, bc_values1, bc_markers1, x0, alpha, perms);
+        cell_info1, bc_values1, bc_markers1, x0, alpha, facet_perms);
+  }
+
+  for (fem::IntegralType itg_type :
```
garth-wells: See comment for matrices

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -1192,7 +1081,47 @@ void lift_bc(V&& b, const Form<T, U>& a, mdspan2_t x_dofmap,
...
+       {fem::IntegralType::exterior_facet, fem::IntegralType::vertex,
+        fem::IntegralType::ridge})
+  {
+    md::mdspan<const std::uint8_t, md::dextents<std::size_t, 2>> perms;
+    switch (itg_type)
```
garth-wells: Use ternary to simplify.

jorgensd:
> I think this is a natural way to unify the assembly of integrals over cell sub-entities that only require data from one cell at a time. It’s what we’ve had in mind for a while to deal with more general mixed-dimensional problems.
> 
> I think we should avoid names like “exterior_entities”, which doesn’t really make sense. Since this assembler can only be used when we only need data from a single cell, in contrast to a patch of cells, a name that reflects this would be better.
> 
> There are quite a lot of mysterious measures in UFL. If we don’t want to go down the path of adding `dr` and `dp` (which is the natural extension of what we have already with `dx` and `ds` etc.), then I agree that we should change `dx`, `ds`, and `dS` so that everything is consistent with the new approach. This is (probably) a lot of work in UFL, though.
> 
> Maybe as a first step, this PR could be split into two: a part that generalises the assembler and a part that adds ridge integrals and `dr` etc? The generalisation of the assembler is quite natural and shouldn't depend too heavily on how we define the integrals in UFL. This way, we can get the nice generalisation merged first and have more time to think about what the right way to handle measures is in UFL.
> 
> I haven’t been through the PR in detail yet, but I’ll have a look tomorrow!

Splitting it doesn’t make sense. The only thing that would be left out is the addition of IntegralType::ridge and the corresponding tests. 

If we want to refactor the whole ufl.Measure class, that is a separate issue that should be dealt with elsewhere. It would probably mean that we replace "integral_type» in ufl by codim=0,1…,tdim.
But it also means that we need a new way of specifying `dS` integrals, as both dS and ds is codim=1. I think the natural thing would be to add another integer that would be number of cells (1 for dx, ds, dr, do), 2 for ds, N for any other interior codim entity.

Note that adding a Measure to ufl doesn’t really mean much. the only place where it matters is in FFCx, where we convert the integral_type of ufl to our implementation.

as another sidenote. This pr would also allow for peak integrals, as that is efficiently what a vertex integral currently is in 3D.

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -1466,34 +1361,59 @@ void assemble_vector(
              mdspanx22_t(facets1.data(), facets1.size() / 4, 2, 2)},
             fn, constants,
             mdspanx2x_t(coeffs.data(), facets.size() / 4, 2, cstride),
-            cell_info0, perms);
+            cell_info0, facet_perms);
       }
     }
 
-    for (int i = 0; i < L.num_integrals(IntegralType::vertex, 0); ++i)
+    for (fem::IntegralType itg_type :
```
garth-wells: See earlier comments.

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -1466,34 +1361,59 @@ void assemble_vector(
...
-                     md::extents<std::size_t, md::dynamic_extent, 2>>(
-              vertices.data(), vertices.size() / 2, 2),
-          {dofs, bs,
-           md::mdspan<const std::int32_t,
-                      md::extents<std::size_t, md::dynamic_extent, 2>>(
-               vertices0.data(), vertices0.size() / 2, 2)},
-          fn, constants,
-          md::mdspan(coeffs.data(), vertices.size() / 2, cstride), cell_info0);
+      md::mdspan<const std::uint8_t, md::dextents<std::size_t, 2>> perms;
+      switch (itg_type)
```
garth-wells: Simplify.

cpp/dolfinx/fem/utils.h
```diff
@@ -757,82 +698,118 @@ Form<T, U> create_form_factory(
...
+        dim = 0;
+        break;
+      }
+      default:
+        throw std::runtime_error("Unsupported integral type");
+      }
 
-        std::vector<int> active_coeffs;
-        for (int j = 0; j < form.num_coefficients; ++j)
+      auto get_default_integration_entities
```
garth-wells: Don't use `auto` here - be explicit with the type.

cpp/dolfinx/fem/utils.h
```diff
@@ -757,82 +698,118 @@ Form<T, U> create_form_factory(
...
-      for (int i = 0; i < num_integrals_type[vertex]; ++i)
+      std::size_t dim;
+      switch (itg_type)
       {
-        const int id = ids[i];
-        ufcx_integral* integral
-            = form.form_integrals[integral_offsets[vertex] + i];
-        assert(integral);
-        check_geometry_hash(*integral, form_idx);
+      case IntegralType::exterior_facet:
```
garth-wells, line 710: This would be neater with the suggestion to remove enum and use integer representation with dimensions.

cpp/dolfinx/fem/utils.h
```diff
@@ -757,82 +698,118 @@ Form<T, U> create_form_factory(
...
         {
-          if (integral->enabled_coefficients[j])
-            active_coeffs.push_back(j);
+          // Integrate over all owned exterior facets
+          return mesh::exterior_facet_indices(topology);
+        }
+        else
+        {
+          // Integrate over all owned entities
+          // Default vertex kernel operates on all (owned) vertices
```
garth-wells: What does this mean here? Can ridges be captured in this block?

python/test/unit/fem/test_assembler.py
```diff
@@ -224,7 +224,32 @@ def test_assembly_bcs(self, mode):
...
+        vertex_tags = dolfinx.mesh.meshtags(
+            mesh, 0, exterior_vertices, np.ones_like(exterior_vertices, dtype=np.int32)
+        )
+        dP_exterior = dP(subdomain_data=vertex_tags, subdomain_id=1)
+        x = ufl.SpatialCoordinate(mesh)
+        f = ufl.exp(x[0] + x[1])
+        a = form(
+            inner(u, v) * dx
+            + inner(u, v) * ds
+            - inner(u, v) * dr_exterior
```
garth-wells, line 261: What does 'exterior' mean in this context on `dr` and `dP`?

cpp/dolfinx/fem/Form.h
jorgensd: We can change exterior_facet to facet.

jorgensd: > Having `exterior_facet` is inconsistent with `vertex` and `ridge`. Better name would be `facet`, with `interior_facet` being a special case.
> 
> Would it be better to remove the enum and use `(n, codim)` pairs? At present there are ambiguities between facet, ridge and vertex in dimensions less than 3.

Is `n` here num_neighbours?

cpp/dolfinx/fem/utils.h
jorgensd: This I think should be done in a separate PR. I have made an issue that relates to this at: https://github.com/FEniCS/dolfinx/issues/3917

cpp/dolfinx/fem/assemble_scalar_impl.h
jorgensd: I didn't add this. It exists in the previous code. Can remove it.

cpp/dolfinx/fem/utils.h
jorgensd: Ah, this is a comment that remains from the vertex integration kernel. This should just be the first comment.
It captures cells, facets, ridges and vertices.

cpp/dolfinx/fem/assemble_scalar_impl.h
michalhabera: If the whole function works under the assumption of assembling only over single adjacent cell, and this assumption is well documented then it could be removed. Added it in the vertex PR to improve readability and make easier to find these assumptions buried in literal integers all over the code.

python/test/unit/fem/test_assembler.py
jorgensd: It means that I have picked specific ridges (line 228) and vertices  (line 238) that are on the actual exterior boundary of the mesh (they are connected to a facet that is classified as an exterior facet).

cpp/dolfinx/fem/Form.h
garth-wells: > Is `n` here num_neighbours?

Topological dimension of the cell.

Special cases, i.e. `interior_facet`, could possibly be handled for the time being with a negative index.

jorgensd:
@garth-wells Renaming `exterior_facet` to `facet` cascades deep into FFCx, and should be handled in a separate PR, as there are tons of bad use-cases of `ufl` strings in there. Especially if we now want to remove the enum altogether, I think we leave it in this PR and change in soon.

cpp/dolfinx/fem/Form.h
michalhabera: As the code stands the `interior_facet` is not a special case of `exterior_facet (renamed as) facet`. These two are disjoint, having exactly 2 or exactly 1 adjacent cells respectively.

jorgensd: I think what we should do (in a separate PR) is to replace IntegralType alltogether with a tuple (codim, num_neighbours). It would capture all our existing cases, plus future cases were sub-entities are connected to more than two cells (as for instance is the case in the T-joint meshes.

python/dolfinx/fem/forms.py
```diff
@@ -217,9 +222,10 @@ def form_cpp_class(
 
 _ufl_to_dolfinx_domain = {
     "cell": IntegralType.cell,
-    "exterior_facet": IntegralType.exterior_facet,
+    "facet": IntegralType.exterior_facet,
```
```suggestion
    "exterior_facet": IntegralType.exterior_facet,
```

cpp/dolfinx/fem/utils.h
```diff
@@ -146,13 +146,13 @@ get_cell_vertex_pairs(std::int32_t v, std::span<const std::int32_t> cells,
 /// @param[in] entities List of mesh entities. Depending on the `IntegralType`
 /// these are associated with different entities:
 ///     `IntegralType::cell`:             cells
-///     `IntegralType::exterior_facet`:   facets
+///     `IntegralType::exterior_facet`:            facets
```
jorgensd:
```suggestion
///     `IntegralType::exterior_facet`: facets
```

cpp/dolfinx/fem/utils.h
```diff
@@ -146,13 +146,13 @@ get_cell_vertex_pairs(std::int32_t v, std::span<const std::int32_t> cells,
...
 ///     `IntegralType::interior_facet`:   facets
 ///     `IntegralType::vertex`:           vertices
 /// @return List of integration entity data, depending on the `IntegralType` the
 /// data per entity has different layouts
 ///     `IntegralType::cell`:             cell
-///     `IntegralType::exterior_facet`:   (cell, local_facet)
+///     `IntegralType::exterior_facet`:            (cell, local_facet)
```
jorgensd:
```suggestion
///     `IntegralType::exterior_facet`: (cell, local_facet)
```

cpp/dolfinx/fem/utils.h
```diff
@@ -146,13 +146,13 @@ get_cell_vertex_pairs(std::int32_t v, std::span<const std::int32_t> cells,
...
 ///     `IntegralType::cell`:             cells
-///     `IntegralType::exterior_facet`:   facets
+///     `IntegralType::exterior_facet`: facets
 ///     `IntegralType::interior_facet`:   facets
 ///     `IntegralType::vertex`:           vertices
 /// @return List of integration entity data, depending on the `IntegralType` the
 /// data per entity has different layouts
 ///     `IntegralType::cell`:             cell
-///     `IntegralType::exterior_facet`:   (cell, local_facet)
+///     `IntegralType::exterior_facet`: (cell, local_facet)
```
jorgensd:
```suggestion
///     `IntegralType::exterior_facet`:   (cell, local_facet)
```

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -158,7 +158,7 @@ def interior_marker(x):
 
 def a_ufl(u, v, f, g, measure):
     "Helper function to create a UFL bilinear form. The form depends on the integral type"
-    if measure.integral_type() == "cell" or measure.integral_type() == "exterior_facet":
+    if measure.integral_type() == "cell" or measure.integral_type() == "facet":
```
jorgensd:
```suggestion
    if measure.integral_type() == "cell" or measure.integral_type() == "exterior_facet":
```

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -167,15 +167,15 @@ def a_ufl(u, v, f, g, measure):
 
 def L_ufl(v, f, g, measure):
     "Helper function to create a UFL linear form. The form depends on the integral type"
-    if measure.integral_type() == "cell" or measure.integral_type() == "exterior_facet":
+    if measure.integral_type() == "cell" or measure.integral_type() == "facet":
```
jorgensd:
```suggestion
    if measure.integral_type() == "cell" or measure.integral_type() == "exterior_facet":
```

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -175,7 +175,7 @@ def L_ufl(v, f, g, measure):
 
 
 def M_ufl(f, g, measure):
-    if measure.integral_type() == "cell" or measure.integral_type() == "exterior_facet":
+    if measure.integral_type() == "cell" or measure.integral_type() == "facet":
```
jorgensd:
```suggestion
    if measure.integral_type() == "cell" or measure.integral_type() == "exterior_facet":
```

garth-wells:
> > > @garth-wells Renaming `exterior_facet` to `facet` cascades deep into FFCx, and should be handled in a separate PR, as there are tons of bad use-cases of `ufl` strings in there. Especially if we now want to remove the enum altogether, I think we leave it in this PR and change in soon.
> > 
> > 
> > Something has gone very badly wrong in the DOLFINX design if this is the case. It should be relatively easy to rename an enum, or change the enum to a pair of integers, in DOLFINx without changing UFCx or FFCx. What specifically causes a 'cascade' into FFCx?
> 
> This is being handled separately in: [FEniCS/ffcx#787](https://github.com/FEniCS/ffcx/pull/787)

But what's the issue? I can't see it. If we have such tight coupling between DOLFINx and FFCx/UFCx there is a design mistake that needs to be fixed.

jorgensd:
> > > > @garth-wells Renaming `exterior_facet` to `facet` cascades deep into FFCx, and should be handled in a separate PR, as there are tons of bad use-cases of `ufl` strings in there. Especially if we now want to remove the enum altogether, I think we leave it in this PR and change in soon.
> > > 
> > > 
> > > Something has gone very badly wrong in the DOLFINX design if this is the case. It should be relatively easy to rename an enum, or change the enum to a pair of integers, in DOLFINx without changing UFCx or FFCx. What specifically causes a 'cascade' into FFCx?
> > 
> > 
> > This is being handled separately in: [FEniCS/ffcx#787](https://github.com/FEniCS/ffcx/pull/787)
> 
> But what's the issue? I can't see it. If we have such tight coupling between DOLFINx and FFCx/UFCx there is a design mistake that needs to be fixed.

Currently in DOLFINx we look things up by strings, and assume that they match with things in FFCx.
This is bad design and something I have taken care of in:
- https://github.com/FEniCS/dolfinx/pull/3919
- https://github.com/FEniCS/ffcx/pull/787

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -228,30 +236,30 @@ void assemble_exterior_facets(
...
   const int ndim1 = bs1 * num_dofs1;
   std::vector<T> Ae(ndim0 * ndim1);
-  assert(facets0.size() == facets.size());
-  assert(facets1.size() == facets.size());
-  for (std::size_t f = 0; f < facets.extent(0); ++f)
+  assert(entities0.size() == entities.size());
+  assert(entities1.size() == entities.size());
+  for (std::size_t f = 0; f < entities.extent(0); ++f)
   {
     // Cell in the integration domain, local facet index relative to the
```
michalhabera:
```suggestion
    // Cell in the integration domain, local entity index relative to the
```

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -273,31 +220,36 @@ T assemble_scalar(
...
-
-    value += impl::assemble_vertices(
-        x_dofmap, x,
-        md::mdspan<const std::int32_t,
-                   md::extents<std::size_t, md::dynamic_extent, 2>>(
-            vertices.data(), vertices.size() / shape1, shape1),
-        fn, constants,
-        md::mdspan(coeffs.data(), vertices.size() / shape1, cstride), cdofs_b);
+    md::mdspan<const std::uint8_t, md::dextents<std::size_t, 2>> perms
+        = (itg_type == fem::IntegralType::exterior_facet)
```
michalhabera, line 230: What are you checking here? If the IntegralType happens to be codim=1, i.e. if the entities we integrate over happen to be facets?

cpp/dolfinx/fem/pack.h
```diff
@@ -179,11 +179,8 @@ allocate_coefficient_storage(const Form<T, U>& form, IntegralType integral_type,
     const std::vector<int> offsets = form.coefficient_offsets();
     cstride = offsets.back();
     num_entities = form.domain(integral_type, id, 0).size();
-    if (integral_type == IntegralType::exterior_facet
-        or integral_type == IntegralType::interior_facet)
-    {
+    if (integral_type != IntegralType::cell)
```
michalhabera, line 182: Can you explain this a bit?

jorgensd: Any celltype that is not a cell integral has its entities represented as a (cell, local_entity_index) tuple. Thus we have to divide by 2 to get the correct size of those.

cpp/dolfinx/fem/assemble_scalar_impl.h
jorgensd: This was initially an if statement, then a switch then moved back after Garth’s comment.

in general any integral of codim>0 needs permutations (except codim=tdim). At the time of writing, we only have facet permutations implemented.
It matters for mesh-submesh coupling, ie this PR does not introduce 3D-1D mixed problems, as I need ridge permutations, which is in another PR (https://github.com/FEniCS/dolfinx/pull/3904)

michalhabera: Should the condition be what you said then, that codim > 0 (or better codim=1, and warning for codim > 1)? I don't mind either of the if, switch or ternary, I am asking about the logic.

jorgensd: With what we currently have implemented, it should be equality.

if I get ridge permutations into DOLFINx, it will be an if codim>0. Garth asked to not implement for features that are not there yet, which is why I removed the switch and went for a tenary logic for what we support.

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -156,16 +156,24 @@ void assemble_cells_matrix(
   }
 }
 
-/// @brief Execute kernel over exterior facets and accumulate result in
-/// a matrix.
+/// @brief Execute kernel over entities of codimension > 1 and accumulate result
```
schnellerhase:
```suggestion
/// @brief Execute kernel over entities of codimension ≥ 1 and accumulate result
```

cpp/dolfinx/fem/assemble_scalar_impl.h
```diff
@@ -55,41 +55,50 @@ T assemble_cells(mdspan2_t x_dofmap,
   return value;
 }
 
-/// Execute kernel over exterior facets and accumulate result
+/// @brief Execute kernel over entities of codimension > 1 and accumulate result
```
schnellerhase:
```suggestion
/// @brief Execute kernel over entities of codimension ≥ 1 and accumulate result
```

cpp/dolfinx/fem/assemble_vector_impl.h
```diff
@@ -718,7 +718,16 @@ void assemble_cells(
   }
 }
 
-/// @brief Execute kernel over cells and accumulate result in vector.
+/// @brief Execute kernel over entities of codimension > 1 and accumulate result
```
schnellerhase:
```suggestion
/// @brief Execute kernel over entities of codimension ≥ 1 and accumulate result
```

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -191,14 +199,14 @@ void assemble_cells_matrix(
 /// @param[in] perms Facet permutation integer. Empty if facet
```
schnellerhase:
```suggestion
/// @param[in] perms Entity permutation integer. Empty if entity
```

cdaversin:
Glad to see the ridge integrals are coming their way! The unification makes real sense to me. In addition to the significant duplicated code removal, the "re-naming" of the assembly functions with `entities` instead of `facets` makes the code more generic and less confusing when working with various dimension / co-dimensions.

I agree with @michalhabera 's point on moving assembly routines to "the codimension format". I think @jorgensd 's [PR#3919](https://github.com/FEniCS/dolfinx/pull/3919) with a dedicated class `IntegralType` is a smart way to define the integral types from their codimension and number of adjacent cells. This class is simple/light but flexible to include future type of integrals (some which can be difficult to name / define as a string)

This is a big step towards codimension > 1 :)

chrisrichardson (approved):

PR #3904, opened 2025-09-03
Compute entity permutations only where they are needed

jorgensd:
`Topology::create_entity_permutations()` computed every permutation a mesh
could have, whether or not the caller used any of them: face orientations,
edge orientations, the packed per-cell info, and a
`num_cells × facets_per_cell` facet array. There was no way to ask for less,
so a form either computed all of it or none of it.

This splits the computation so one only get one is required for any given computation.
The permutations are now requested per entity dimension, matching
`create_entities(dim)`:

```cpp
void create_entity_permutations(int dim, int num_threads = 1);
void create_cell_permutations(int num_threads = 1);
const std::vector<std::uint8_t>&  get_entity_permutations(int dim) const;
const std::vector<std::uint32_t>& get_cell_permutation_info() const;
```

`mesh::compute_entity_permutations` takes an entity dimension and returns only
that dimension; the packed cell info moves to `mesh::compute_cell_permutations`.
`Topology::get_facet_permutations` is removed — it is
`get_entity_permutations(tdim - 1)`.

## What was computed before
Assembling on a unit cube (tetrahedra). The first two columns count
O(num_cells) orientation computations; the last two say which arrays end up
retained on the topology. On a tetrahedron the sub-entity permutation array is
always the one for dimension 2, since facets are faces.

| form | face pass | edge pass | packed cell info | entity perm array |
|---|---|---|---|---|
| `inner(u, v) * dx`, P1 | 0 | 0 | – | – |
| `inner(u, v) * dx`, N1curl | 1 | 1 | yes | **dim 2, never read** |
| `inner(u, v) * ds`, P1 | 0 | 0 | – | – |
| `inner(avg(u), avg(v)) * dS`, P1 | 1 | **1, never read** | **yes, never read** | dim 2 |
| `inner(avg(u), avg(v)) * dS`, N1curl | 1 | 1 | yes | dim 2 |

## What is computed now

Same columns, same forms.

| form | face pass | edge pass | packed cell info | entity perm array |
|---|---|---|---|---|
| `inner(u, v) * dx`, P1 | 0 | 0 | – | – |
| `inner(u, v) * dx`, N1curl | 1 | 1 | yes | **–** |
| `inner(u, v) * ds`, P1 | 0 | 0 | – | – |
| `inner(avg(u), avg(v)) * dS`, P1 | 1 | **0** | **–** | dim 2 |
| `inner(avg(u), avg(v)) * dS`, N1curl | 1 | 1 | yes | dim 2 |

Pass counts measured from the `"* Compute triangle/quad face permutations"` and
`"* Compute edge reflections"` timers.

Reading the two tables together: nothing gained a pass or an array, three cells
lost one, and the last row is unchanged -- see
Unpacking rather than recomputing for why
that row does not become two face passes.

Concretely:

- A **P1 interior-facet** form no longer runs the edge-orientation pass at all,
  and no longer builds the packed cell info that nothing reads.
- An **N1curl cell** form no longer builds and retains the facet permutation
  array.
- The `element->needs_dof_transformations() or needs_facet_permutations()`
  guard in the vector and matrix assemblers is gone. `cell_info` is consumed
  only by the `P0`/`P1T` dof-transform calls, which are no-ops for an element
  needing no transformations, so a P1 interior-facet form was computing a full
  cell-permutation pass and never reading it.

Storage is `std::array<std::optional<std::vector<std::uint8_t>>, 3>` indexed by
dimension — no map, and no allocation for dimensions never asked for.
`optional` distinguishes "not computed" from "computed and legitimately empty"
(vertices, or a rank with no cells).

Assemblers call `impl::entity_permutations(topology, integral_type, cell_type)`,
which maps the integral type to an entity dimension and computes on demand, so
the decision lives in one place rather than being duplicated per assembler.

## Unpacking rather than recomputing

Splitting the computation creates a hazard worth calling out. The packed cell
info *is* the per-dimension permutations concatenated — 3 bits per face, then
1 bit per edge — so a form that needs both (N1curl on `dS`) would, naively,
compute the face orientations twice: once to pack, once to answer
`get_entity_permutations(2)`. That would make this PR a regression for exactly
the forms that are most expensive to assemble.

`Topology::create_entity_permutations(dim)` therefore checks whether the packed
info already exists and, if so, unpacks the requested dimension from it instead
of recomputing. The inversion is exact, and the assemblers compute the packed
info first, so this is the order that actually occurs. The last row of the
table above confirms it: one face pass, not two.

## Side effect: ridge integrals

Because the computation is now indexed by entity dimension rather than
hardcoded to facets, edge permutations fall out of the same code path. Ridge
integrals consequently receive real permutations instead of the
unconditionally-zero span they were given before. This is a consequence of the
rewrite, not its purpose, and nothing here is specific to ridges.

There is no end-to-end test of that path: `needs_facet_permutations` is false
for a single-domain ridge integral, and making it true needs a codimension-2
submesh coefficient, which `Form.h` currently rejects with
`"Integral type not supported."`. The path is covered at the `Topology` level
only.

## Tests

`test_cell_permutation_info_matches_entity_permutations` decodes the packed
`uint32` and checks it against the per-dimension arrays across
triangle/quadrilateral/tetrahedron/hexahedron — the two are now separate code
paths and must agree. It computes the per-dimension permutations *first*, so it
compares two independent code paths rather than comparing the unpacking against
its own source, and asserts the data is non-trivial so it cannot pass
vacuously.

`test_entity_permutations_unpacked_from_cell_info` checks the unpacked values
equal the independently computed ones.

Three tests pin the efficiency claims directly: cell assembly with N1curl
computes the packed info and no sub-entities; interior-facet assembly computes
dimension 2 and neither dimension 1 nor the packed info; computing one
dimension leaves the others uncomputed.

## Notes for review

- `Form::needs_facet_permutations()` is a form-wide OR over all integrals, so
  gating on it can over-compute but never under-compute. FFCx sets its
  per-integral flag from `("+" in restrictions and "-" in restrictions) or
  is_mixed_dim`, which means "this kernel reads `quadrature_permutation`" —
  nothing facet-specific. Per-integral granularity plus a rename to
  `needs_entity_permutations` on both sides would tighten this further; left
  for a separate change.
- Vertices have no orientation, so `dim == 0` returns an empty vector rather
  than raising. This matters: a 1D mesh's facets *are* vertices, so
  interior-facet assembly on an interval legitimately asks for dimension 0.

## AI disclosure
This PR was initially made by me without the help of AI.
The latter revision was done by me with the help of CLAUDE (OPUS 5). I've reviewed, edited and simplified all the additions made by CLAUDE.


HoseongJeong:
Thanks for your support Jorgensd.  Has it become possible to insert 1D rebar elements into a 3D concrete mesh?

cpp/dolfinx/mesh/permutationcomputation.cpp
```diff
@@ -284,17 +284,22 @@ compute_face_permutations(const mesh::Topology& topology)
 } // namespace
 
 //-----------------------------------------------------------------------------
-std::pair<std::vector<std::uint8_t>, std::vector<std::uint32_t>>
+std::tuple<std::vector<std::uint8_t>, std::vector<std::uint8_t>,
+           std::vector<std::uint32_t>>
 mesh::compute_entity_permutations(const mesh::Topology& topology)
```
mscroggs, line 289: We should make it so that the user can decide which permutations to create rather than it being all or none. But I think that may be for a future PR.

cpp/dolfinx/mesh/permutationcomputation.cpp
```diff
@@ -284,17 +284,22 @@ compute_face_permutations(const mesh::Topology& topology)
...
 {
+
   common::Timer t_perm("Compute entity permutations");
   const int tdim = topology.dim();
   CellType cell_type = topology.cell_type();
   const std::int32_t num_cells = topology.connectivity(tdim, 0)->num_nodes();
   const int facets_per_cell = cell_num_entities(cell_type, tdim - 1);
+  const int ridges_per_cell = cell_num_entities(cell_type, tdim - 2);
```
mscroggs, line 297: If `tdim` is 1, What will this do?

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -886,6 +886,20 @@ const std::vector<std::uint8_t>& Topology::get_facet_permutations() const
   return _facet_permutations;
 }
 //-----------------------------------------------------------------------------
+const std::vector<std::uint8_t>& Topology::get_ridge_permutations() const
```
mscroggs, line 889: `get_entity_permutations(int codim)` might be a more general interface, but as above this might be for a future PR

jorgensd:
> Thanks for your support Jorgensd.  Has it become possible to insert 1D rebar elements into a 3D concrete mesh?

There is the possibility to do ridge integrals (from the viewpoint of a single cell with dR) in dolfinx. 

Furthermore, for non-conforming coupling with rebar you can consider https://github.com/scientificcomputing/fenicsx_ii

cpp/dolfinx/mesh/permutationcomputation.cpp
```diff
@@ -332,69 +333,106 @@ compute_face_permutations(const mesh::Topology& topology, int num_threads)
...
+  case 2:
+  {
+    spdlog::info("Compute face permutations");
+    const std::vector<std::bitset<_BITSETSIZE>> face_perm
+        = compute_face_permutations<_BITSETSIZE>(topology, num_threads);
+    // Three bits encode each face: one reflection bit and two rotation
+    // bits.
+    for (std::int32_t c = 0; c < num_cells; ++c)
+      for (int i = 0; i < entities_per_cell; ++i)
+        perms[c * entities_per_cell + i]
```
garth-wells, line 386: Use curly braces (statement is multi-line).

cpp/dolfinx/mesh/permutationcomputation.h
```diff
@@ -7,67 +8,81 @@
...
+/// @param[in] dim Topological dimension of the entities to permute.
+/// Must satisfy `0 <= dim < topology.dim()`. Vertices (`dim == 0`) have
+/// no orientation, so an empty vector is returned for them.
+/// @param[in] num_threads Number of threads to use.
+/// @return Permutation of each cell-local entity of dimension `dim`,
+/// flattened row-wise.
+std::vector<std::uint8_t> compute_entity_permutations(const Topology& topology,
+                                                      int dim, int num_threads);
+
+/// @brief Compute the cell permutation data used by non-Lagrange
```
garth-wells: What about higher-order Lagrange elements?

cpp/dolfinx/mesh/Topology.h
```diff
@@ -210,9 +214,25 @@ class Topology
...
+  /// Does nothing if the permutations for `dim` have already been
+  /// computed.
+  ///
+  /// @param[in] dim Topological dimension of the entities, e.g.
+  /// `dim() - 1` for facets. Must satisfy `0 <= dim < dim()`. Vertices
+  /// have no orientation, so their permutations are empty.
+  /// @param[in] num_threads Number of threads to use. Must be >= 1.
+  void create_entity_permutations(int dim, int num_threads = 1);
+
+  /// @brief Compute the cell permutation info used by non-Lagrange
```
garth-wells: What about higher-order Lagrange?

cpp/dolfinx/mesh/Topology.h
```diff
@@ -210,9 +214,25 @@ class Topology
   /// @param[in] d1 Topological dimension.
   void create_connectivity(int d0, int d1);
 
-  /// @brief Compute entity permutations and reflections.
+  /// @brief Compute the permutations of the cell-local entities of a
```
garth-wells, line 217: The docs (or PR description) don't make clear the difference between `create_entity_permutations` and `create_cell_permutations`. Spell it out simply in the docs.

python/dolfinx/mesh.py
```diff
@@ -291,13 +291,23 @@ def create_entities(self, dim: int, num_threads: int = 1) -> bool:
...
         Args:
+            dim: Topological dimension of the entities, e.g. ``tdim - 1``
+                for facets. Must satisfy ``0 <= dim < tdim``.
             num_threads: Number of CPU threads to use. Must be >= 1.
         """
-        self._cpp_object.create_entity_permutations(num_threads)
+        self._cpp_object.create_entity_permutations(dim, num_threads)
+
+    def create_cell_permutations(self, num_threads: int = 1) -> None:
+        """Compute the cell permutation info used by non-Lagrange elements.
```
garth-wells: Higher-order Lagrange?

python/dolfinx/mesh.py
```diff
@@ -291,13 +291,23 @@ def create_entities(self, dim: int, num_threads: int = 1) -> bool:
...
+
+        Args:
+            num_threads: Number of CPU threads to use. Must be >= 1.
```
garth-wells:
```suggestion
            num_threads: Number of threads to use. Must be >= 1.
```

cpp/dolfinx/mesh/utils.h
```diff
@@ -717,7 +717,7 @@ std::vector<std::int32_t> locate_entities_boundary(const Mesh<T>& mesh, int dim,
 /// @param[in] entities Entity indices (local to process).
 /// @param[in] permute If `true`, permute the DOFs such that they are
 /// consistent with the orientation of `dim`-dimensional mesh entities.
-/// This requires `create_entity_permutations` to be called first.
+/// This requires `create_cell_permutations` to be called first.
```
garth-wells: The is a precondition. What happens if it's not called first?

python/doc/source/release_notes.md
```diff
@@ -2,6 +2,43 @@
...
+The single `create_entity_permutations()` call has been split in two:
+
+- {py:meth}`Topology.create_entity_permutations
+  <dolfinx.mesh.Topology.create_entity_permutations>` now takes a `dim`
+  argument, matching {py:meth}`Topology.create_entities
+  <dolfinx.mesh.Topology.create_entities>`, and computes only that dimension.
+  Vertices have no orientation, so `dim` 0 gives an empty array.
+- {py:meth}`Topology.create_cell_permutations
+  <dolfinx.mesh.Topology.create_cell_permutations>` computes the packed
+  per-cell permutation info used by non-Lagrange elements, previously a side
```
garth-wells: Higher-order Lagrange?

python/test/unit/mesh/test_mesh.py
```diff
@@ -915,3 +934,144 @@ def test_point_mesh(gdim, dtype):
...
+    # A Lagrange element needs no dof transformations, so the packed
+    # cell info is not needed either
+    with pytest.raises(RuntimeError):
+        msh.topology.get_cell_permutation_info()
+
+
+def test_cell_assembly_computes_only_cell_permutations():
+    """Dof transformations need the packed cell info, but no sub-entities.
+
+    The cell permutation info used by non-Lagrange elements is no longer
```
garth-wells: Higher-order Lagrange?

garth-wells (dismissed):
`create_cell_permutations` is a new function, but the PR description does say what it does.

jorgensd:
@garth-wells with respect to higher order Lagrange elements;

- Higher order Lagrange elements requires permutations at **dofmap creation** where they are permuted to a consistent ordering.
- They do not need it at cell permutations at **assembly**

| family | needs_dof_permutations | needs_dof_transformations |
|---|---|---|
| Lagrange P>=3 | ✓ — consumed once, at dofmap build | — |
| N1curl / RT / BDM | — | ✓ — consumed per cell, at assembly |


With respect to the cell vs entity permutations.
- `cell_permutations` are required for permuting":
    -  the higher order Lagrange (P>=3) dofmap at construction.
    - Applying transformations to non-permutable dofmaps in assembly.
- `entity_permutations` is used to align dS and mixed dimensional integrals as the two cells (cell-cell or cell-subentity) has different local ordering. It can be derived from `cell_permutations`.

python/test/unit/mesh/test_mesh.py
```diff
@@ -26,6 +26,7 @@
     DiagonalType,
     GhostMode,
     cell_num_vertices,
+    cell_num_entities,
```
jorgensd:
```suggestion
    cell_num_entities,
    cell_num_vertices,
```

jorgensd:
@garth-wells I've tried to address all your comments.

jhale (approved):

mscroggs (approved):

PR #3905, opened 2025-09-04
Ensure `dolfinx` does not expose warning to user

schnellerhase:
Ensures and fixes warning propagation to caller. 

Triggered by: `import dolfinx` triggered now a deprecation warning after deprecation of `gmshio` in https://github.com/FEniCS/dolfinx/pull/3881. This is caused by including this module by default. Removes this include and passes `-W error` to the pytest calls in the CI which should catch such problems in the future.

Removes `gmshio`.

### DIFF
--- a/python/dolfinx/graph.py
+++ b/python/dolfinx/graph.py
@@ -53,7 +53,7 @@ class AdjacencyList:
         cpp_object: Union[
             _cpp.graph.AdjacencyList_int32,
             _cpp.graph.AdjacencyList_int64,
-            _cpp.graph.AdjacencyList_AdjacencyList_int_sizet_int8__int32_int32,
+            _cpp.graph.AdjacencyList_int_sizet_int8__int32_int32,
         ],
     ):
         """Creates a Python wrapper for the exported adjacency list class.
--- a/python/dolfinx/graph.py
+++ b/python/dolfinx/graph.py
@@ -53,7 +53,7 @@ class AdjacencyList:
         cpp_object: Union[
             _cpp.graph.AdjacencyList_int32,
             _cpp.graph.AdjacencyList_int64,
-            _cpp.graph.AdjacencyList_int_sizet_int8__int32_int32,
+            _cpp.graph.AdjacencyList_AdjacencyList_int_sizet_int8__int32_int32,
         ],
     ):
         """Creates a Python wrapper for the exported adjacency list class.
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -72,6 +72,7 @@ jobs:
       OMPI_ALLOW_RUN_AS_ROOT_CONFIRM: 1
       PRTE_MCA_rmaps_default_mapping_policy: :oversubscribe   # Newer OpenMPI
       OMPI_MCA_rmaps_base_oversubscribe: true                 # Older OpenMPI
+      PYTEST_ADDOPTS: "-W error"
 
     name: Build and test
     steps:
@@ -176,6 +177,7 @@ jobs:
       OMPI_ALLOW_RUN_AS_ROOT: 1
       OMPI_ALLOW_RUN_AS_ROOT_CONFIRM: 1
       PRTE_MCA_rmaps_default_mapping_policy: :oversubscribe
+      PYTEST_ADDOPTS: "-W error"
 
     name: Build and test (${{ matrix.petsc_arch }}, ${{ matrix.docker_image }})
     steps:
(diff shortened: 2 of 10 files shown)


python/dolfinx/graph.py
```diff
@@ -53,7 +53,7 @@ def __init__(
         cpp_object: Union[
             _cpp.graph.AdjacencyList_int32,
             _cpp.graph.AdjacencyList_int64,
-            _cpp.graph.AdjacencyList_int_sizet_int8__int32_int32,
+            _cpp.graph.AdjacencyList_AdjacencyList_int_sizet_int8__int32_int32,
```
jorgensd:
```suggestion
            _cpp.graph.AdjacencyList_int_sizet_int8__int32_int32,
```

jorgensd: Revert my accidental push

jorgensd:
> @jorgensd Any opinions? I think you were the one in a need of backward compatible API. Given how ugly this is with deprecation warning we can:
> 
>     1. drop `gmshio` completely, not pretending we care about backward compatible API now,
> 
>     2. remove deprecation warning on import but keep `gmshio`,
> 
>     3. do the magic as in the MR to treat the deprecation warning in custom way.
> 
> 
> I am suggesting to go with 1.

What I have raised as points before is:
- The initial suggestion when I implemented this functionality was to call it `dolfinx.io.gmsh`. @garth-wells pointed out that this can lead to confusion, as if one imports it as `from dolfinx.io import gmsh`, it would seem like the full `gmsh` Python API, which it is not.
- Therefore I am not really convinced that we should change it to `dolfinx.io.gmsh`, and rather keep what we are having as `dolfinx.io.gmshio`, as there is no real benefit of renaming it to `dolfinx.io.gmsh`, as you would either import it as `import dolfinx.io.gmsh` or `from dolfinx.io import gmsh as gmshio`, which is equally verbose (you only save 2 characters for the first option, and the second option is just longer).

michalhabera:
> > @jorgensd Any opinions? I think you were the one in a need of backward compatible API. Given how ugly this is with deprecation warning we can:
> > ```
> > 1. drop `gmshio` completely, not pretending we care about backward compatible API now,
> > 
> > 2. remove deprecation warning on import but keep `gmshio`,
> > 
> > 3. do the magic as in the MR to treat the deprecation warning in custom way.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I am suggesting to go with 1.
> 
> What I have raised as points before is:
> 
> * The initial suggestion when I implemented this functionality was to call it `dolfinx.io.gmsh`. @garth-wells pointed out that this can lead to confusion, as if one imports it as `from dolfinx.io import gmsh`, it would seem like the full `gmsh` Python API, which it is not.
> * Therefore I am not really convinced that we should change it to `dolfinx.io.gmsh`, and rather keep what we are having as `dolfinx.io.gmshio`, as there is no real benefit of renaming it to `dolfinx.io.gmsh`, as you would either import it as `import dolfinx.io.gmsh` or `from dolfinx.io import gmsh as gmshio`, which is equally verbose (you only save 2 characters for the first option, and the second option is just longer).

I don't see a problem with `import dolfinx.io.gmsh as gmshio`. It is nice to have a clear pythonic naming separation, just as we do with `dolfinx.fem.petsc`. Similarly, we do not have `dolfinx.fem.petscfem` and `dolfinx.la.petscla` to prevent users from colliding these two.

michalhabera:
> > I don't see a problem with `import dolfinx.io.gmsh as gmshio`. It is nice to have a clear pythonic naming separation, just as we do with `dolfinx.fem.petsc`. Similarly, we do not have `dolfinx.fem.petscfem` and `dolfinx.la.petscla` to prevent users from colliding these two.
> 
> If you read Garth's original comment, he points out that the python module for `PETSc` is called `petsc4py` and not `petsc`, thus there isn't a nameclash.

I did not mean a clash with petsc4py but between dolfinx.fem.petsc and dolfinx.la.petsc.

jorgensd:
> > > I don't see a problem with `import dolfinx.io.gmsh as gmshio`. It is nice to have a clear pythonic naming separation, just as we do with `dolfinx.fem.petsc`. Similarly, we do not have `dolfinx.fem.petscfem` and `dolfinx.la.petscla` to prevent users from colliding these two.
> > 
> > 
> > If you read Garth's original comment, he points out that the python module for `PETSc` is called `petsc4py` and not `petsc`, thus there isn't a nameclash.
> 
> I did not mean a clash with petsc4py but between dolfinx.fem.petsc and dolfinx.la.petsc.

Ok, so one can have a name-clash if one insists of calling `from dolfinx.fem import petsc` and `from dolfinx.la import petsc` in the same script. But since they are `sub-packages` of DOLFINx, and there is no python package called `petsc` it is a different issue than the one raised by Garth.

 If one always has to call `from dolfinx import gmsh as gmshio` then I really don't see a point for changing the API.

PR #3906, opened 2025-09-04
Pep-0604 - Removing `typing.Union`

jorgensd:
With 3.10 being minimal version, we can use | in favor of typing.Union, ref: https://peps.python.org/pep-0604/


schnellerhase:
Nice, missed that one - can you remove the corresponding ignore of `UP007` in `pyproject.toml`?https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/#non-pep604-annotation-union-up007

jhale (approved):

PR #3908, opened 2025-09-06
Handle diagonal assembly

jorgensd:
Requires: https://github.com/FEniCS/ffcx/pull/783


jhale:
It would be odd to suggest to users Jacobi for preconditioning Poisson without going into why you shouldn't actually use it, which is confusing in a demo. The canonical demo would be the L^2 projection using the mass matrix, where Jacobi works very well.

PR #3915, opened 2025-09-09
Remove dynamic memory allocations in low-level matrix assembly functions

garth-wells:
This makes assembly functions 'static' in that there are no dynamic memory allocations. This make functions suitable for being called fora very small number of cells.

Used Claude Sonnet-5


jorgensd:
Prior to merging this, can we consider #3900 as it simplifies the assembly interface.

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -562,11 +589,20 @@ void assemble_matrix(
...
+    md::mdspan<const std::int32_t, md::dextents<std::size_t, 2>> dofs0
+        = dofmap0->map();
     const int bs0 = dofmap0->bs();
-    auto dofs1 = dofmap1->map();
+    md::mdspan<const std::int32_t, md::dextents<std::size_t, 2>> dofs1
+        = dofmap1->map();
     const int bs1 = dofmap1->bs();
 
+    std::vector<T> Ab((2 * bs0 * dofs0.extent(1))
+                      * (2 * bs1 * dofs1.extent(1)));
```
schnellerhase, line 700: Should this be moved outside the `cell_type` loop and allocate the max. required storage for all cell types?

garth-wells: It will create more noise for what should be no performance benefit. Can be changed later if profiling indicates that it would be beneficial.

Issue #1465, opened 2021-03-25
Are non-manifold meshes (e.g. a "T-junction" in a shell geometry) unsupported in dolfinx?


austinherrema:
I am hoping to use `dolfinx` to perform structural analysis of a relatively complex shell geometry. At the moment I am hoping to use `dolfinx` as opposed to legacy `dolfin` due to the better support of quad elements as discussed [here](https://bitbucket.org/fenics-project/dolfin/issues/1089/quadrilateral-mesh-reordering-error).

I made a simple T-junction geometry (.3dmf attached with .txt extension)

![image](https://user-images.githubusercontent.com/29387274/112530782-78f15b80-8d6c-11eb-8661-23e1bac83592.png)

and am trying to load it into `dolfinx` using:

``` python
import dolfinx
from dolfinx import io
from dolfinx.io import XDMFFile
from mpi4py import MPI

infile = XDMFFile(MPI.COMM_WORLD, "tTopology.xdmf", "r", encoding=dolfinx.cpp.io.XDMFFile.Encoding.ASCII)
mesh = infile.read_mesh(name="Grid")
infile.close()
```

This gives the following error:
```
2021-03-23 15:08:14.799 ( 0.291s) [main ] graphbuild.cpp:93 ERR| Found three identical facets in mesh (local)
Traceback (most recent call last):
File “xdmfToPvd.py”, line 11, in
mesh = infile.read_mesh(name=“Grid”)
File “/usr/local/dolfinx-real/lib/python3.8/dist-packages/dolfinx/io.py”, line 59, in read_mesh
mesh = cpp.mesh.create_mesh(self.comm(), cpp.graph.AdjacencyList_int64(cells),
RuntimeError: Inconsistent mesh data in GraphBuilder: three identical facets
```

Error is clearly stemming from [`graphbuild.cpp`](https://github.com/FEniCS/dolfinx/blob/4cdc251c325adbf884150aa00914da5eec1c2f40/cpp/dolfinx/mesh/graphbuild.cpp). Is this completely expected in this case and, if so, are there any known workarounds? I can imagine, for example, dividing the mesh into multiple manifold meshes and then handling the geometry coupling manually, but this would obviously be less straightforward,

[tTopology.txt](https://github.com/FEniCS/dolfinx/files/6207313/tTopology.txt)

chrisrichardson:
It's a check, which can be disabled... I'll take a look.

garth-wells:
[...]

PR #3916, opened 2025-09-09
Fix "jointed" manifold mesh creation

### LINKED ISSUE #3912: Connectivity computation for branching meshes
With the merge of https://github.com/FEniCS/dolfinx/pull/3830 we have support for parallel branching meshes. The non trivial entity connectivity computations need to be updated to account for the new cases. These are the connectivities $d_a \to d_b$ and $d_b \to d_a$ for  $d_a \neq d_b$ and $`d_a \notin \{ \text{tdim}, 0 \}`$.

jorgensd:
This PR finalizes the long-standing issue in: https://github.com/FEniCS/dolfinx/issues/1465 which is reading in meshes which has joints (up to an arbitrary number).

This PR builds on top of https://github.com/FEniCS/dolfinx/pull/3830, which improved the dual graph creator to accept
that a facet can be connected to a user-specified number of cells.

This PR propagates this change into `create_mesh`, where it is required to identify the correct "boundary_vertices".
A "boundary_vertex" is a vertex that has been assigned to this process after partitioning, but one doesn't know if it is on the process boundary at the present time. These are later used to identify the correct neighbouring cells and set up the correct ghosting and vertex ownership.

With this PR, the user can specify the number of cells a facet can be connected to. This value should by default be 2, and it is only for branching meshes that this will be something different.

To verify that this works, and resolve #3912 I have extended the tests to check that the topology is sane, as #3830 did not quite resolve it, resulting in vertices being duplicated on the process boundaries.
The test is also extended to tetrahedral cells.

### DIFF
--- a/python/demo/demo_comm-pattern.py
+++ b/python/demo/demo_comm-pattern.py
@@ -95,7 +95,7 @@ def plot_bar(G: nx.MultiGraph):
     num_edges = [len(nbrs) for _, nbrs in G.adj.items()]
     weights = [sum(data["weight"] for nbr, data in nbrs.items()) for _, nbrs in G.adj.items()]
 
-    fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(12, 5))
+    _fig, (ax1, ax2) = plt.subplots(1, 2, figsize=(12, 5))
 
     ax1.bar(ranks, num_edges)
     ax1.set_xlabel("rank")
--- a/python/dolfinx/pkgconfig.py
+++ b/python/dolfinx/pkgconfig.py
@@ -18,7 +18,7 @@ def _pkgconfig_query(s):
     pkg_config_exe = os.environ.get("PKG_CONFIG", None) or "pkg-config"
     cmd = [pkg_config_exe, *s.split()]
     proc = subprocess.Popen(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-    out, err = proc.communicate()
+    out, _err = proc.communicate()
     rc = proc.returncode
     return (rc, out.rstrip().decode("utf-8"))
 
--- a/python/test/unit/fem/test_custom_assembler.py
+++ b/python/test/unit/fem/test_custom_assembler.py
@@ -327,7 +327,7 @@ def test_custom_mesh_loop_rank1(dtype):
     # Assemble using generated tabulate_tensor kernel and Numba
     # assembler
     b3 = Function(V, dtype=dtype)
-    ufcx_form, module, code = dolfinx.jit.ffcx_jit(
+    ufcx_form, _module, _code = dolfinx.jit.ffcx_jit(
         mesh.comm, L, form_compiler_options={"scalar_type": dtype}
     )
 
--- a/python/test/unit/io/test_xdmf_mesh.py
+++ b/python/test/unit/io/test_xdmf_mesh.py
@@ -131,14 +131,14 @@ def test_read_write_p2_mesh(tempdir, encoding):
         assert np.all(idx[srt] == np.arange(len(idx)))
         x = points[srt]
 
-        element_types, element_tags, node_tags = model.mesh.getElements(dim=3)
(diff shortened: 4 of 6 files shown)


python/test/unit/mesh/test_branching_manifold.py
```diff
@@ -88,50 +89,70 @@ def on_boundary(x):
...
+    assert num_vertices_global == skeleton_mesh.topology.index_map(0).size_global
 
     skeleton_top.create_connectivity(1, 2)
     skeleton_f_to_c = skeleton_top.connectivity(1, 2)
 
     skeleton_im_f = skeleton_mesh.topology.index_map(1)
 
-    def on_boundary(x):
-        return np.any(np.isclose(x, 0)) or np.any(np.isclose(x, 1))
+    if cell_type == CellType.hexahedron:
```
schnellerhase, line 149: Why is this necesasry?

jorgensd: Because there isn't a fixed number of connecting cells for an interior facet grid of tetrahedrons.
I think it is either 4 or 6.

schnellerhase (approved):
Thanks for the quick catch and even quicker fix.

python/dolfinx/mesh.py
```diff
@@ -671,6 +671,7 @@ def create_mesh(
     e: ufl.Mesh | basix.finite_element.FiniteElement | basix.ufl._BasixElement | _CoordinateElement,
     x: npt.NDArray[np.floating],
     partitioner: Callable | None = None,
+    max_facet_links_per_cell: int = 2,
```
schnellerhase:
```suggestion
    max_facet_to_cell_links: int = 2,
```

python/dolfinx/mesh.py
```diff
@@ -684,6 +685,8 @@ def create_mesh(
             type of ``e``.
         partitioner: Function that determines the parallel distribution of
             cells across MPI ranks.
+        max_facet_links_per_cell: Maximum number of cells a facet can
```
schnellerhase:
```suggestion
        max_facet_to_cell_links: Maximum number of cells a facet can
```

python/dolfinx/mesh.py
```diff
@@ -736,7 +739,7 @@ def create_mesh(
     x = np.asarray(x, dtype=dtype, order="C")
     cells = np.asarray(cells, dtype=np.int64, order="C")
     msh: _cpp.mesh.Mesh_float32 | _cpp.mesh.Mesh_float64 = _cpp.mesh.create_mesh(
-        comm, cells, cmap._cpp_object, x, partitioner
+        comm, cells, cmap._cpp_object, x, partitioner, max_facet_links_per_cell
```
schnellerhase:
```suggestion
        comm, cells, cmap._cpp_object, x, partitioner, max_facet_to_cell_links
```

python/test/unit/mesh/test_branching_manifold.py
```diff
@@ -88,50 +89,70 @@ def on_boundary(x):
...
+        max_facet_to_cell_links = 16
+    else:
+        raise ValueError("Unknown cell type")
     skeleton_mesh = create_mesh(
         comm,
         cells,
         element,
         new_x,
         create_cell_partitioner(GhostMode.shared_facet, max_facet_to_cell_links),
+        max_facet_links_per_cell=max_facet_to_cell_links,
```
schnellerhase:
```suggestion
        max_facet_to_cell_links =max_facet_to_cell_links,
```

cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -660,7 +660,7 @@ mesh::build_local_dual_graph(
                                                    f1_it);
                                }));
 
-      std::int32_t cell_count = matching_facets.size();
+      std::uint16_t cell_count = matching_facets.size();
```
jorgensd:
```suggestion
      std::int32_t cell_count = matching_facets.size();
```

PR #3918, opened 2025-09-09
Remove experimental from pull request, as it then becomes  a required check

jorgensd:
This stops the merge queue from starting.
See: https://github.com/FEniCS/dolfinx/pull/3916


jorgensd:
Not really sure this is the right fix, as Mac doesn’t show as required
IMG_0474

But something is wrong. Maybe it is because it is called ci.yml?

PR #3919, opened 2025-09-10
Change facet type in FFCx

jorgensd:
depends on https://github.com/FEniCS/ffcx/pull/787

### DIFF
--- a/cpp/dolfinx/fem/utils.h
+++ b/cpp/dolfinx/fem/utils.h
@@ -502,7 +502,7 @@ Form<T, U> create_form_factory(
   }
 
   // Create facets, if required
-  if (num_integrals_type[exterior_facet] > 0
+  if (num_integrals_type[facet] > 0
       or num_integrals_type[interior_facet] > 0)
   {
     mesh->topology_mutable()->create_entities(tdim - 1);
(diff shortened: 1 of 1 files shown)


jorgensd:
@garth-wells I've now replaced the `dolfinx::fem::IntegralType` with a `struct` that currently holds `codim` and `num_cells`, which expresses all integral types we currently have (dx, ds, dS, dP), future ones as `dr` (ridge) and even integral types we haven't looked at yet, such as interior peak or ridge integrals, as well as `interior_facet` on branched manifolds.

jorgensd:
Thinking even more about this, I think the integral type should also contain a `default_domain` enum, which should be `all` or `exterior` to distinguish between dP and ds when tdim is 1.

currently these two measure would give different behavior for the two measures, which is unintuitive, at least to me. This would affect `compute_integration_domains` only. @schnellerhase ; a penny for your thoughts:)

garth-wells:
> I don't think `IntegralType` carrying the `num_cells` information is the right abstraction. This is connectivity information from the mesh, and changes across the domain. As we have not figured out how to do patch assemblies for now, best to not add interface support for it.

Agree - just need the dimension and special case handling for interior facets for now.

PR #3922, opened 2025-09-11
Fix Windows CI for `.env`-file config variables.

jorgensd:
Correct way of adding values in Windows Powershell is described in: https://stackoverflow.com/questions/66733076/github-actions-set-environment-variable-for-windows-build-with-powershell

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -50,7 +50,6 @@ jobs:
       - name: Load environment variables
         run: |
           cat dolfinx-src/.github/workflows/fenicsx-refs.env >> $env:GITHUB_ENV
-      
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -49,8 +49,8 @@ jobs:
 
       - name: Load environment variables
         run: |
-          cat dolfinx-src/.github/workflows/fenicsx-refs.env >> $GITHUB_ENV
-
+          cat dolfinx-src/.github/workflows/fenicsx-refs.env >> $env:GITHUB_ENV
+
```
jorgensd:
```suggestion

```

PR #3924, opened 2025-09-11
Fix vector block data for single function space case

schnellerhase:
With the new interface, the `create_vector([V], kind="mpi")` case no longer holds block data. This adds the block data, allowing to treat a single function space vector as a block vector when passing `type="mpi"`. Caught downstream.

### DIFF
--- a/python/test/unit/fem/test_assembler.py
+++ b/python/test/unit/fem/test_assembler.py
@@ -178,6 +178,17 @@ def nest_matrix_norm(A):
 
 
 @pytest.mark.petsc4py
+def test_vector_single_space_as_block():
+    from dolfinx.fem.petsc import create_vector as petsc_create_vector
+
+    mesh = create_unit_square(MPI.COMM_WORLD, 3, 3)
+    gdim = mesh.geometry.dim
+    V = functionspace(mesh, ("Lagrange", 1, (gdim,)))
+    assert petsc_create_vector(V).getAttr("_blocks") is None
+    assert petsc_create_vector(V, kind="mpi").getAttr("_blocks") is not None
+
+
+@pytest.mark.petsc4py
 class TestPETScAssemblers:
     @pytest.mark.parametrize("mode", [GhostMode.none, GhostMode.shared_facet])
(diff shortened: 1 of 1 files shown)


python/dolfinx/la/petsc.py
```diff
@@ -123,7 +123,10 @@ def create_vector(
         index_map, bs = maps[0]
         ghosts = index_map.ghosts.astype(PETSc.IntType)  # type: ignore[attr-defined]
         size = (index_map.size_local * bs, index_map.size_global * bs)
-        return PETSc.Vec().createGhost(ghosts, size=size, bsize=bs, comm=index_map.comm)  # type: ignore
+        b = PETSc.Vec().createGhost(ghosts, size=size, bsize=bs, comm=index_map.comm)  # type: ignore
+        if kind == PETSc.Vec.Type.MPI:
```
michalhabera, line 127: What if `kind == None`?

schnellerhase: For the previous behaviour that would have been equivalent to the call `create_vector(L)` which did not attach block data, therefore we don't want to attach data.

michalhabera: If creating vector for list of a single space, i.e. `create_vector([V])` is almost the same as creating a vector on a single space `create_vector(V)`, except that the former attaches block data, then why don't we remove the whole concept of single space `create_vector(V)` and always require a list of spaces, be it of length one. That would simplify meaning of `kind`, in my opinion, so either mpi or nest. @garth-wells ?

garth-wells: I know I added the 'block data' concept but I can't recall the details ;). Does it introduce any memory overhead for the single space case?

In general, I'm in favour of as few special cases as possible, and has few interface types and possible.

schnellerhase: How about merging this, which recovers behaviour as before (and fixes down stream usage) and having a look into removing the difference in another PR?

michalhabera: Agree.

PR #3929, opened 2025-09-19
Remove NewtonSolver from hyperelasticity demo

chrisrichardson:


cpp/demo/hyperelasticity/main.cpp
```diff
@@ -59,82 +59,163 @@ class HyperElasticProblem
...
+    int iteration = 0;
+    PetscReal residual0 = 0;
+
+    auto converged
+        = [&iteration, &residual0, this](const Vec r) -> std::pair<double, bool>
     {
-      VecGhostUpdateBegin(x, INSERT_VALUES, SCATTER_FORWARD);
-      VecGhostUpdateEnd(x, INSERT_VALUES, SCATTER_FORWARD);
+      PetscReal residual = 0;
+      VecNorm(r, NORM_2, &residual);
```
michalhabera, line 90: Should we use NORM_INFINITY for residual norms instead (as a naive fix)? I always struggle that NORM_2 grows with numbers of elements so for a finer mesh you'll violate `atol` very soon.

Btw. are C++ demos run in CI? Could not find the job.

jhale:
Yes, the plan is to remove NewtonSolver and provide some tooling/examples to help people interface SNES from C++. However, I've never done it myself, so I have no idea where to even begin.

chrisrichardson:
I think that the example here is sufficiently simple that reproducing the logic of NewtonSolver (and simplifying) is OK (only adds 70 lines of code).
We could add a deprecation message to `NewtonSolver` at the same time. I am not sure of the best sequencing for this. I don't think I want to add a SNES interface in this PR.

cpp/demo/hyperelasticity/main.cpp
schnellerhase: > Btw. are C++ demos run in CI? Could not find the job.

https://github.com/FEniCS/dolfinx/blob/main/.github/workflows/ccpp.yml#L230 - CI step has a confusing name.

cpp/demo/hyperelasticity/main.cpp
```diff
@@ -59,82 +59,163 @@ class HyperElasticProblem
...
+      // Output iteration number and residual
+      spdlog::info("Newton iteration {}"
+                   ": r (abs) = {} (tol = {}), r (rel) = {} (tol = {})",
+                   iteration, residual, atol, relative_residual, rtol);
+
+      // Return true if convergence criterion is met
+      if (relative_residual < rtol or residual < atol)
+        return {residual, true};
+      else
+        return {residual, false};
```
schnellerhase:
```suggestion
      bool converged = relative_residual < rtol or residual < atol;
      return {residual, converged};
```

cpp/demo/hyperelasticity/main.cpp
```diff
@@ -59,82 +59,163 @@ class HyperElasticProblem
...
+    }
+
+    if (newton_converged)
+    {
+      spdlog::info("Newton solver finished in {} iterations and {} linear "
+                   "solver iterations.",
+                   iteration, krylov_iterations);
+    }
+    else
+      throw std::runtime_error("Newton solver did not converge.");
```
schnellerhase:
```suggestion
    if (not newton_converged)
      throw std::runtime_error("Newton solver did not converge.");
    
    spdlog::info("Newton solver finished in {} iterations and {} linear "
                   "solver iterations.",
                   iteration, krylov_iterations);  
```

chrisrichardson:
@jhale @michalhabera @schnellerhase - any further comments, or shall I merge?

cpp/dolfinx/nls/NewtonSolver.cpp
```diff
@@ -72,6 +73,8 @@ nls::petsc::NewtonSolver::NewtonSolver(MPI_Comm comm)
       _krylov_iterations(0), _iteration(0), _residual(0), _residual0(0),
       _solver(comm), _dx(nullptr), _comm(comm)
 {
+  std::cout << "Deprecation warning: NewtonSolver is deprecated and will be "
```
schnellerhase:
```suggestion
  std::clog << "Deprecation warning: NewtonSolver is deprecated and will be "
```

schnellerhase: Or maybe just `spdlog::warn`? don't think `std::cout` is the right stream.

python/dolfinx/wrappers/petsc.cpp
```diff
@@ -279,7 +280,11 @@ void petsc_nls_module(nb::module_& m)
           "__init__",
           [](dolfinx::nls::petsc::NewtonSolver* ns,
              const dolfinx_wrappers::MPICommWrapper comm)
-          { new (ns) dolfinx::nls::petsc::NewtonSolver(comm.get()); },
+          {
+            new (ns) dolfinx::nls::petsc::NewtonSolver(comm.get());
+            std::cout << "NewtonSolver is deprecated, and will be removed in a "
```
schnellerhase: Same as above.

cpp/dolfinx/nls/NewtonSolver.cpp
chrisrichardson: I thought about using spdlog, but it is likely to be turned off by most users. What is the advantage of `std::clog` ?

cpp/demo/hyperelasticity/main.cpp
```diff
@@ -247,15 +324,11 @@ int main(int argc, char* argv[])
...
-    newton_solver.set_form(problem.form());
-    newton_solver.rtol = 10 * std::numeric_limits<T>::epsilon();
-    newton_solver.atol = 10 * std::numeric_limits<T>::epsilon();
+    problem.rtol = 10 * std::numeric_limits<T>::epsilon();
+    problem.atol = 10 * std::numeric_limits<T>::epsilon();
 
     la::petsc::Vector _u(la::petsc::create_vector_wrap(*u->x()), false);
-    auto [niter, success] = newton_solver.solve(_u.vec());
+    auto [niter, success] = problem.solve(_u.vec());
     std::cout << "Number of Newton iterations: " << niter << std::endl;
```
schnellerhase, line 332:
```suggestion
    spdlog::info("Number of Newton iterations: {}", niter);
```

cpp/dolfinx/nls/NewtonSolver.cpp
schnellerhase: It is the buffered version of `cerr`, so it will be flagged as error output to the caller.

chrisrichardson: I don't see the advantage of buffering for a simple message like this.

schnellerhase: I was trying to say we should mark the error as non-standard output, so either `clog` or `cerr` - picking `clog` over `cerr` due to no need for immediate write of a warning. (`cerr` implies immediate flush)

PR #3931, opened 2025-09-23
Remove PETSc C-binding utils from default namespace and add all spd LogLevels to Python-API

jorgensd:
All other PETSc dependencies requires explicit import.
Having this in the default import of `dolfinx` and with the changes introduced in https://github.com/FEniCS/dolfinx/pull/3905
the user gets the following logging message at import if `numba` is not installed:
```bash
 python3 -c "import dolfinx"
[2025-09-23 09:15:48.457] [info] Could not import numba, so cffi/numba complex types were not registered.
```
which also makes it impossible to get the include path with the following cmake without numba
```cmake
execute_process(
  COMMAND ${Python3_EXECUTABLE} -c "import dolfinx, sys; sys.stdout.write(str(dolfinx.get_include()))"
  OUTPUT_VARIABLE DOLFINX_PY_DIR
  RESULT_VARIABLE DOLFINX_PY_COMMAND_RESULT OUTPUT_STRIP_TRAILING_WHITESPACE)
```
as one would get the following include path
```bash
"/__w/dolfinx_mpc/dolfinx_mpc/python/[2025-09-23 08:09:25.491] [info] Could not import numba, so cffi/numba complex types were not registered./dolfinx-env/lib/python3.12/site-packages/dolfinx/wrappers"
```

This PR also adds debug as a log-level in Python.

### DIFF
--- a/python/dolfinx/wrappers/log.cpp
+++ b/python/dolfinx/wrappers/log.cpp
@@ -21,11 +21,13 @@ void log(nb::module_& m)
 {
   // log level enums
   nb::enum_<spdlog::level::level_enum>(m, "LogLevel", nb::is_arithmetic())
-      .value("OFF", spdlog::level::level_enum::off)
+      .value("TRACE", spdlog::level::level_enum::trace)
       .value("DEBUG", spdlog::level::level_enum::debug)
       .value("INFO", spdlog::level::level_enum::info)
       .value("WARNING", spdlog::level::level_enum::warn)
-      .value("ERROR", spdlog::level::level_enum::err);
+      .value("ERROR", spdlog::level::level_enum::err)
+      .value("CRITICAL", spdlog::level::level_enum::critical)
+      .value("OFF", spdlog::level::level_enum::off);
 
   m.def(
       "set_output_file",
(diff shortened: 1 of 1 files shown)


python/dolfinx/wrappers/log.cpp
```diff
@@ -72,6 +72,9 @@ void log(nb::module_& m)
         case (spdlog::level::level_enum::err):
           spdlog::error(s.c_str());
           break;
+        case (spdlog::level::level_enum::debug):
+          spdlog::debug(s.c_str());
+          break;
```
schnellerhase: Can we reorder by log level in spdlog https://github.com/gabime/spdlog/blob/f1d748e5e3edfa4b1778edea003bac94781bc7b7/include/spdlog/common.h#L231? and should we support the other cases?

jorgensd:
@chrisrichardson @michalhabera 
I split the custom assembler tests into those PETSc dependent and those independent of PETSc.
Additionally, I removed: `get_matsetvalues_cffi_api` as it is never called in our code.

python/dolfinx/wrappers/log.cpp
jorgensd: I don't really see the point of re-ordering them.
If we want to support other log levels, that should be a separate PR.

jorgensd:
The AlmaLinux issues doesn't seem to related to DOLFINx issues, but a gmsh issue, that I hoped was resolved with #3932, but clearly there is something really funky going on there.

PR #3936, opened 2025-09-25
Bugfix: mesh construction from mesh data on single process

schnellerhase:
Minimal reproducible example (fails on `main`):

```python
from mpi4py import MPI

import numpy as np

import basix
import ufl
from dolfinx.mesh import create_mesh

comm = MPI.COMM_WORLD


if comm.rank == 0:
    cells = np.zeros((0, 2), dtype=np.int32)
    x = np.zeros((0, 3), dtype=np.float64)
else:
    #    2
    #  / |
    # 0--1
    cells = np.array([[0, 1], [1, 2], [2, 0]], dtype=np.int32)
    x = np.array([[0, 0, 0], [1, 0, 0], [0, 1, 0]], dtype=np.float64)

element = ufl.Mesh(basix.ufl.element("Lagrange", "interval", 1, shape=(3,)))
mesh = create_mesh(MPI.COMM_WORLD, cells, element, x)
```

Problem arises in the cases where the dual graph is already the global one, thus no facets are available on one process and the default assigned value breaks the parallel reduction of the size/range.

### DIFF
--- a/cpp/dolfinx/mesh/graphbuild.cpp
+++ b/cpp/dolfinx/mesh/graphbuild.cpp
@@ -12,6 +12,7 @@
 #include <dolfinx/common/log.h>
 #include <dolfinx/common/sort.h>
 #include <dolfinx/graph/AdjacencyList.h>
+#include <iostream>
 #include <mpi.h>
 #include <numeric>
 #include <optional>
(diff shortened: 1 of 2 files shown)


cpp/dolfinx/mesh/graphbuild.cpp
```diff
@@ -12,6 +12,7 @@
 #include <dolfinx/common/log.h>
 #include <dolfinx/common/sort.h>
 #include <dolfinx/graph/AdjacencyList.h>
+#include <iostream>
```
jorgensd:
```suggestion
```
REmove debug print include

jorgensd (approved):

PR #3937, opened 2025-09-26
`DirichletBC.dof_indices()` are wrong for blocked spaces

jorgensd:
For blocked function spaces, `locate_dofs_topological` and `DirichletBC` assumes blocked indices.
However, the private functions `num_owned` assumes that the dofs are unrolled, which makes the number of owned indices returned by `DirichletBC::dof_indices` wrong.

### DIFF
--- a/cpp/dolfinx/fem/DirichletBC.h
+++ b/cpp/dolfinx/fem/DirichletBC.h
@@ -259,9 +259,8 @@ private:
   std::size_t num_owned(const DofMap& dofmap,
                         std::span<const std::int32_t> dofs)
   {
-    int bs = dofmap.index_map_bs();
     std::int32_t map_size = dofmap.index_map->size_local();
-    std::int32_t owned_size = bs * map_size;
+    std::int32_t owned_size = map_size;
     auto it = std::ranges::lower_bound(dofs, owned_size);
     return std::distance(dofs.begin(), it);
   }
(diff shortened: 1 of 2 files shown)


cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -383,8 +382,8 @@ class DirichletBC
     // Unroll _dofs0 if dofmap block size > 1
     if (const int bs = _function_space->dofmap()->bs(); bs > 1)
     {
-      _owned_indices0 *= bs;
       _dofs0 = unroll_dofs(_dofs0, bs);
+      _owned_indices0 *= bs;
```
chrisrichardson, line 385: Is there any need to switch these lines?

jorgensd: Not really, it ended up like this due to me first fixing the problem in another, way, relying on `_dofs0`. Will revert.

chrisrichardson:
Looks like `bs` is being applied twice at the moment (is that right?), once in `num_owned` and then again in

chrisrichardson (approved):

michalhabera:
> > Good catch. But what is `_owned_indices0` used for apart from returning in `dof_indices()`? Seems to me a bit weird to return it. With your fix it will return unrolled dofs and number of not unrolled owned dofs, so a bit inconsistent.
> 
> It still returns the number of unrolled dofs, as it is multiplied by `bs`. I caught this in: #3938 where I require it to set boundary conditions on the Python side of things (to the local PETSc vector)

Ah right, missed the multiplication later somewhere in constructor... This is good to merge as a bugfix, but unrelated to this PR, returning an array and number of owned elements in the array is not a common dolfinx pattern, isn't it? We've always relied on user figuring this out.

PR #3938, opened 2025-09-26
Add matrix free demo using PETSc

### LINKED ISSUE #3779: Add example/demo using PETSc Python context PC

The new high-level interfaces should be checked for sanity when using a PC Python context

### Suggested user interface

```python3

```

jorgensd:
Requires: https://github.com/FEniCS/dolfinx/pull/3937

A demo showcasing how to use PETSc's Python type matrix to create a matrix-free operator.
Would partially resolve: https://github.com/FEniCS/dolfinx/issues/3779

A separate question is where one should put the `MatrixFreeOperator`, as it doesn't change with the bi-linear form.


python/demo/demo_matrix_free_petsc.py
```diff
@@ -0,0 +1,495 @@
...
+
+# We are ready to solve our problem.
+# In this demo we will consider two approaches, using
+# `basix.ufl.mixed_element`and using a `ufl.MixedFunctionSpace`.
+
+# We start by definng our mesh and finite element spaces.
+
+N = 25
+mesh = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, N, N)
+el_0 = basix.ufl.element("Lagrange", mesh.basix_cell(), 2, shape=(2,))
```
jorgensd:
```suggestion
el_0 = basix.ufl.element("Lagrange", mesh.basix_cell(), 2, shape=(2,), dtype=dolfinx.default_real_type)
```

python/demo/demo_matrix_free_petsc.py
```diff
@@ -0,0 +1,495 @@
...
+el_1 = basix.ufl.element("Lagrange", mesh.basix_cell(), 1)
```
jorgensd:
```suggestion
el_1 = basix.ufl.element("Lagrange", mesh.basix_cell(), 1,  dtype=dolfinx.default_real_type)
```

python/demo/demo_matrix_free_petsc.py
```diff
@@ -0,0 +1,497 @@
...
+#       jupytext_version: 1.13.6
+# ---
+
+# # Matrix-free solvers in DOLFINx using PETSc
+#
+# Author: Jørgen S. Dokken
+#
+# This demo can be downloaded as a single Python file
+# {download}`demo_matrix_free_petsc.py`.
+# In this demo, we will demonstrate how to set up a fully matrix-free
```
michalhabera:
```suggestion
# In this demo, we will demonstrate how to set up a matrix-free
```

python/demo/demo_matrix_free_petsc.py
```diff
@@ -0,0 +1,497 @@
...
+# solver using PETSc.
+# We will start by defining our variational problem, and then in turn
+# define a custom PETSc-matrix that will handle assembly without ever
```
michalhabera:
```suggestion
# define a custom PETSc SHELL matrix that will handle assembly without ever
```

python/demo/demo_matrix_free_petsc.py
```diff
@@ -0,0 +1,497 @@
...
+# forming the full system matrix.
```
michalhabera:
```suggestion
# forming the PETSc MATAIJ system matrix.
```

michalhabera (commented):
I don't like that you go for a mixed problem and in addition try to solve it with two approaches. It obscures the core of matrix-free demo.
Also, why bother with BCs for projection problem?

jorgensd:
> I don't like that you go for a mixed problem and in addition try to solve it with two approaches. It obscures the core of matrix-free demo. Also, why bother with BCs for projection problem?

Because it is non-trivial to set up, as seen in the demo. It already exposed a bug in DOLFINx (the bc one). It also serves as a general matrixfree solver, rather than one that is specialised for projections.

I can move this demo to my own web page if that is more suitable.

michalhabera:
> > I don't like that you go for a mixed problem and in addition try to solve it with two approaches. It obscures the core of matrix-free demo. Also, why bother with BCs for projection problem?
> 
> Because it is non-trivial to set up, as seen in the demo. It already exposed a bug in DOLFINx (the bc one). It also serves as a general matrixfree solver, rather than one that is specialised for projections.
> 
> I can move this demo to my own web page if that is more suitable.

I think that PETSc's

> > I don't like that you go for a mixed problem and in addition try to solve it with two approaches. It obscures the core of matrix-free demo. Also, why bother with BCs for projection problem?
> 
> Because it is non-trivial to set up, as seen in the demo. It already exposed a bug in DOLFINx (the bc one). It also serves as a general matrixfree solver, rather than one that is specialised for projections.
> 
> I can move this demo to my own web page if that is more suitable.

I think that PETSc' matrix free solver and diagonal assembly should be demoed. But why not keep it as concise, readable and simple as possible now, and then maybe add some form of `MatrixFreeOperator` into our petsc namespace later?

IgorBaratta:
I like the idea,  much better than https://github.com/FEniCS/dolfinx/tree/main/cpp/demo/poisson_matrix_free.
But also think the demo is a bit complex, and would suggest to make the demo less general and more concise.
And perhaps add `MatrixFreeOperator` to the core as @michalhabera suggested.

jorgensd:
> I think that PETSc' matrix free solver and diagonal assembly should be demoed. But why not keep it as concise, readable and simple as possible now, and then maybe add some form of `MatrixFreeOperator` into our petsc namespace later?

Adding it later means that I need to keep on maintaining it somewhere (as later might be a year from now).

I additionally thought that we wanted more advanced demos in DOLFINx, not just simple setups.

schnellerhase:
> Adding it later means that I need to keep on maintaining it somewhere (as later might be a year from now).

How about moving the general interface of the `MatrixFreeOperator` into the tests, as done for `NonlinearPDEProblem` and `NonlinearPDE_SNESProblem`?

jorgensd:
> The right approach for the demo depends on what you want to achieve:
> 
>     1. An example showing how to interface DOLFINx with PETSc at a lower level - in this case, the example should probably be simplified.
> 
>     2. An example showing how to use a fully featured and library based `NonlinearOperator` - in this case, the code should probably go in the library.
> 
> 
> On the second point, some lower-level functions like https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/fem/petsc.py#L1032 might be a better way to expose some of the routines for assembling actions/diagonals - these could then be placed into a context within the user's scope.

The main complication is that setting bcs in the diagonalized mat-vec (mult) case for PETSc does not fit our current application of BCs in `assemble_residual`: https://github.com/FEniCS/dolfinx/pull/3938/files#diff-0b48cc54bef51011d79955f12bee4a1caae04c7ddf91564444e148a592049c5fR194-R216

I would argue that a 150 Line class (`MatrixFreeOperator`, https://github.com/FEniCS/dolfinx/pull/3938/files#diff-0b48cc54bef51011d79955f12bee4a1caae04c7ddf91564444e148a592049c5fR194-R216) is not very complex. It is an advanced demo, for advanced users that know what they are doing. 

I struggle to understand where we want to draw the line on "more complicated"/"too complicated" for our demos.

jhale:
Well, I would leave the complication of handling mixed or block spaces to the library version.

jhale (approved):
Looks good. Early days on this functionality so let's leave the class in the demo for now.

python/demo/demo_matrix_free_petsc.py
```diff
@@ -0,0 +1,497 @@
...
+import numpy as np
+
+import basix.ufl
+import dolfinx.fem.petsc
+import ufl
+
+# -
+
+# ## Matrix-free operator
+# Many interative methods, such as the conjugate gradient method,
```
jhale: Line lengths look like they weren't done by text editor.

python/demo/demo_matrix_free_petsc.py
```diff
@@ -0,0 +1,497 @@
...
+            for bcs, off0 in zip(bcs0, offset0[:-1], strict=True):  # type: ignore[assignment]
+                for bc in bcs:
+                    di = bc.dof_indices()
+                    odi = di[0][: di[1]]
+                    self._diagonal.array_w[off0 + odi] = 1
+        self._diagonal.ghostUpdate(PETSc.InsertMode.INSERT, PETSc.ScatterMode.FORWARD)
+        vec.setArray(self._diagonal)
+
+
+# ## Setting up a krylov subspace solver with the matrix-free operator
```
jhale: Krylov

python/demo/demo_matrix_free_petsc.py
```diff
@@ -0,0 +1,497 @@
...
+                    di = bc.dof_indices()
+                    odi = di[0][: di[1]]
+                    v_array[odi] = x_array[odi]
+        dolfinx.la.petsc._ghost_update(
+            self._vector, PETSc.InsertMode.INSERT, PETSc.ScatterMode.FORWARD
+        )
+        Y.setArray(self._vector)
+        Y.ghostUpdate(PETSc.InsertMode.INSERT, PETSc.ScatterMode.FORWARD)
+
+    def getDiagonal(self, mat, vec):
```
jhale, line 225: Explain why this is necessary (Jacobi preconditioning)

python/demo/demo_matrix_free_petsc.py
```diff
@@ -0,0 +1,497 @@
...
+    v, q = ufl.TestFunctions(W)
+    residual = ufl.inner(u_h - f, v) * ufl.dx + ufl.inner(p_h - g, q) * ufl.dx
+    return ufl.system(residual)
+
+
+# We also define a convenience function for creating the Krylov subspace
+# solver and attaching the matrix free operator.
+
+
+def define_matrix_free_ksp(
```
jhale: Define -> create

jhale:
I think we should get this in and decide on moving `MatrixFreeOperator` to a library class, or perhaps simplifying it as an inline class, later.

PR #3941, opened 2025-09-27
More fixes to DirichletBC.dof_indices

michalhabera:
PR https://github.com/FEniCS/dolfinx/pull/3937 missed the case of subspaces -- there is a third DirichletBC constructor that does not unroll and multiply with blocksize, as the incoming dofs must be already unrolled.

I've reverted changes in `num_owned`, but compute `_owned_indices0` from unrolled dofs always, later in the constructor.


jorgensd (changes requested):
Looks good to me, but let's add  test to check the second constructor that I missed in the first round?

jorgensd (approved):

PR #3945, opened 2025-10-03
Enable XDMF file closing and reopening.

adrianjhpc:
I've a program that runs for a long time. It writes a mesh and meshtags to the XDMF file (hdf5 format) at the beginning, and then periodically writes function data. This is all fine, and works. 

However, if I try to use the file to get intermediate results at any point it fails to load in something like Paraview. I'm assuming this is because the file has not been closed, this only happens at the end of the program.

Therefore, I've proposed functionality to allow the program to close and reopen the file to allow it to leave the file in a usable state unless it's actively being updated.

It also strikes me this is useful for long running simulations that may hit wallclock limits on schedulers (although there is no guarantee that wouldn't happen during I/O updates).

There are reasons you might not want to do this as well, including the fact the way it is implemented assumes the file isn't modified externally between the close() and reopen() operations (a reasonably safe assumption, but users being users....).

### DIFF
--- a/cpp/dolfinx/io/XDMFFile.cpp
+++ b/cpp/dolfinx/io/XDMFFile.cpp
@@ -30,6 +30,13 @@ XDMFFile::XDMFFile(MPI_Comm comm, const std::filesystem::path& filename,
     : _comm(comm), _filename(filename), _file_mode(file_mode),
       _xml_doc(new pugi::xml_document), _encoding(encoding)
 {
+  open();
+}
+//-----------------------------------------------------------------------------
+XDMFFile::~XDMFFile() { close(); }
+//-----------------------------------------------------------------------------
+void XDMFFile::open(bool reopen)
+{
   // Handle HDF5 and XDMF files with the file mode. At the end of this
   // we will have _hdf5_file and _xml_doc both pointing to a valid and
   // opened file handles.
@@ -44,7 +51,7 @@ XDMFFile::XDMFFile(MPI_Comm comm, const std::filesystem::path& filename,
         = xdmf_utils::get_hdf5_filename(_filename);
     const bool mpi_io = dolfinx::MPI::size(_comm.comm()) > 1 ? true : false;
     _h5_id
-        = io::hdf5::open_file(_comm.comm(), hdf5_filename, file_mode, mpi_io);
+        = io::hdf5::open_file(_comm.comm(), hdf5_filename, _file_mode, mpi_io);
     assert(_h5_id > 0);
     spdlog::info("Opened HDF5 file with id \"{}\"", _h5_id);
   }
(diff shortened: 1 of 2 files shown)


schnellerhase:
Testable with 
```
import dolfinx

from mpi4py import MPI
import time

V = dolfinx.fem.functionspace(mesh, ("CG", 1))

with dolfinx.io.XDMFFile(MPI.COMM_WORLD, "file.xdmf", "w") as f:
    f.write_mesh(mesh)
    f.write_function(dolfinx.fem.Function(V))
    time.sleep(1000) # move outside of managed context for working example.
```

schnellerhase:
```diff
diff --git a/cpp/dolfinx/io/XDMFFile.cpp b/cpp/dolfinx/io/XDMFFile.cpp
index c1c99caeb8..61bb0f25e7 100644
--- a/cpp/dolfinx/io/XDMFFile.cpp
+++ b/cpp/dolfinx/io/XDMFFile.cpp
@@ -9,6 +9,7 @@
 #include "xdmf_function.h"
 #include "xdmf_mesh.h"
 #include "xdmf_utils.h"
+#include <H5Fpublic.h>
 #include <boost/lexical_cast.hpp>
 #include <dolfinx/common/log.h>
 #include <dolfinx/fem/Function.h>
@@ -148,6 +149,7 @@ void XDMFFile::write_mesh(const mesh::Mesh<U>& mesh, const std::string& xpath)
 
   // Add the mesh Grid to the domain
   xdmf_mesh::add_mesh(_comm.comm(), node, _h5_id, mesh, mesh.name);
+  H5Fflush(_h5_id, H5F_SCOPE_LOCAL);
 
   // Save XML file (on process 0 only)
   if (MPI::rank(_comm.comm()) == 0)
@@ -283,7 +285,7 @@ void XDMFFile::write_function(const fem::Function<T, U>& u, double t,
 
   // Add the mesh Grid to the domain
   xdmf_function::add_function(_comm.comm(), u, t, grid_node, _h5_id);
-
+  H5Fflush(_h5_id, H5F_SCOPE_LOCAL);
   // Save XML file (on process 0 only)
   if (dolfinx::MPI::rank(_comm.comm()) == 0)
     _xml_doc->save_file(_filename.c_str(), "  ");
```

and setting `HDF5_USE_FILE_LOCKING=0` resolves the described problem. Should we maybe add an optional flag for flushing in the write calls instead?

adrianjhpc:
Possibly another way to do it. The challenge with putting flushing within the XDMF file is you're forcing a flush per "write", whereas giving users the close() and reopen() option lets users control when the "flush" happens.

However, it's maybe a safer or more saner way of doing it, so I wouldn't particularly object.

adrianjhpc:
Mainly because I'm using the C++ dolfinx api, not the Python one.

adrianjhpc:
If you're going to deprecate XDMF what's is the preferred replacement?

jorgensd:
> Mainly because I'm using the C++ dolfinx api, not the Python one.

But the code above isn't specific to the Python API (Except the context manager bit).
XDMFFile has an append mode, one can then close the file after it has been written to, and re-open and add more data later.



> If you're going to deprecate XDMF what's is the preferred replacement?

We hope that VTKHDF will become fully-featured (from the Paraview-viewpoint).
Currently it is missing some features that would be very nice to have (storing mesh, cell tags and facet tags in the same file). Currently VTKHDF can write a mesh, and point data (similar to what XDMFFile does).

adrianjhpc:
I'm not sure how you'd re-create the python "with" approach for C++ without creating new objects within the same scope.

adrianjhpc:
The close and reopen approach does not require rebuilding the whole I/O context.

adrianjhpc:
Thanks, I'll check

jorgensd:
Before me merge anything, I'd like a profiling of the speedup compared to open/close/append.
For instance, running the following on 28 processes yields:
```python
from mpi4py import MPI
import dolfinx
import time
import numpy as np

M = 500
mesh = dolfinx.mesh.create_unit_cube(MPI.COMM_WORLD, M, M, M)


u = dolfinx.fem.Function(V)
u.interpolate(lambda x: np.sin(2 * np.pi * x[0]))
if mesh.comm.rank == 0:
    print(f"Number of dofs: {V.dofmap.index_map.size_global * V.dofmap.index_map_bs}", flush=True)
with dolfinx.io.XDMFFile(MPI.COMM_WORLD, "file.xdmf", "w") as f:
    f.write_mesh(mesh)


mesh.comm.barrier()
start = time.perf_counter()
with dolfinx.io.XDMFFile(MPI.COMM_WORLD, "file.xdmf", "a") as f:
    end = time.perf_counter()
    total_time = end - start
    total_opening_time = mesh.comm.allreduce(total_time, op=MPI.MAX)

    mesh.comm.barrier()
    start = time.perf_counter()
    f.write_function(u, 0.0)
    end = time.perf_counter()
    total_time = end - start
    total_writing_time = mesh.comm.allreduce(total_time, op=MPI.MAX)
    if mesh.comm.rank == 0:
        print(f"Time to open XDMF file: {total_opening_time:.4e} s", flush=True)
        print(f"Time to write function: {total_writing_time:.4e} s", flush=True)
```

```bash
Number of dofs: 125751501

Time to open XDMF file: 4.4185e-03 s
Time to write function: 1.6293e+00 s
```
so the overhead of opening/close is quite minimal.


> Implemented a `flush` function at https://github.com/FEniCS/dolfinx/tree/schnellerhase/xdmf-flush. Flushing after write calls plus setting `HDF5_USE_FILE_LOCKING=0` should resolve it and is a minimal change. Could you verify this indeed works in your use case?

That being said, the change suggested by @schnellerhase is so minimal, that I wouldn't be against merging it, as it just makes it possible to ensure flushing the data.

adrianjhpc:
The flush() works for my case, so should be sufficient.

PR #3947, opened 2025-10-06
Add XDMF-file flush functionality

schnellerhase:
Ref https://github.com/FEniCS/dolfinx/pull/3945


jorgensd:
As the flush functionality is quite neat, could you add it to the demos that use xdmffile, and add a comment about why it is smart to use it?

PR #3948, opened 2025-10-14
Fix `interpolation_matrix` for continuous spaces

jorgensd:
Currently, the interpolation matrix only works properly for "broken", i.e. entries are accumulated for dofs shared across multiple cells.

This changes the default behavior of `interpolation_matrix` to add_values to the existing matrix, and then accumulate across processes.
Each process only adds entries for row DOFs that:
1. Is owned by the processes
2. Has not been added previously on the process.

### DIFF
--- a/cpp/dolfinx/fem/discreteoperators.h
+++ b/cpp/dolfinx/fem/discreteoperators.h
@@ -536,12 +536,13 @@ void interpolation_matrix(const FunctionSpace<U>& V0,
   auto cell_map = mesh->topology()->index_map(tdim);
   assert(cell_map);
   std::int32_t num_cells = cell_map->size_local();
-  std::int32_t num_owned_rows
-      = dofmap1->index_map->size_local() * dofmap1->index_map_bs();
-  std::int32_t num_ghosted_rows
-      = dofmap1->index_map->num_ghosts() * dofmap1->index_map_bs();
   int row_bs = dofmap1->bs();
+  std::int32_t num_owned_rows
+      = dofmap1->index_map->size_local() * dofmap1->index_map_bs() / row_bs;
+  std::int32_t num_ghosted_rows
+      = dofmap1->index_map->num_ghosts() * dofmap1->index_map_bs() / row_bs;
   std::vector<std::int8_t> row_added(num_owned_rows + num_ghosted_rows, 0);
+
   for (std::int32_t c = 0; c < num_cells; ++c)
   {
     // Get cell geometry (coordinate dofs)
@@ -641,14 +642,16 @@ void interpolation_matrix(const FunctionSpace<U>& V0,
                                                     space_dim0);
       auto row_dofs = dofmap1->cell_dofs(c);
       for (std::size_t i = 0; i < row_dofs.size(); ++i)
-        for (int k = 0; k < row_bs; ++k)
+      {
+        std::int32_t r = row_dofs[i];
+        if (r >= num_owned_rows || row_added[r])
         {
-          std::int32_t r = row_dofs[i] * row_bs + k;
-          T scale = ((r >= num_owned_rows) | row_added[r]) ? 0.0 : 1.0;
           for (std::size_t j = 0; j < space_dim0; ++j)
-            A(i * row_bs + k, j) *= scale;
-          row_added[r] = 1;
+            for (int k = 0; k < row_bs; ++k)
+              A(i * row_bs + k, j) = 0.0;
         }
+        row_added[r] = 1;
+      }
     }
     mat_add(dofmap1->cell_dofs(c), dofmap0->cell_dofs(c), Ab);
   }


cpp/dolfinx/fem/discreteoperators.h
```diff
@@ -627,7 +633,24 @@ void interpolation_matrix(const FunctionSpace<U>& V0,
...
+    // and only add owned rows
+    {
+      md::mdspan<T, md::dextents<std::size_t, 2>> A(Ab.data(), space_dim1,
+                                                    space_dim0);
+      auto row_dofs = dofmap1->cell_dofs(c);
+      for (std::size_t i = 0; i < row_dofs.size(); ++i)
+        for (int k = 0; k < row_bs; ++k)
+        {
+          std::int32_t r = row_dofs[i] * row_bs + k;
+          T scale = ((r >= num_owned_rows) | row_added[r]) ? 0.0 : 1.0;
```
chrisrichardson: Could make this cleaner. I would suggest blanking the non-owned rows in a separate loop, and then you can just have something like:
```
if (row_added[r])
  for (std::size_t j = 0; j < space_dim0; ++j)
    A(i * row_bs + k, j) = 0.0;
else
  row_added[r] = 1;
```

Also, it seems to me, you don't need to keep separate account for each entry in a block. `k` could go into an inner loop, and the `row_added` array could be just `index_map->size_local` without bs or ghosts.

jorgensd: @chrisrichardson  It seems like you are making the assumption that `V.dofmap.bs` is the same as `V.dofmap.index_map_bs`,  which we cannot assume (or if we can, why do we have two variables for it?).

chrisrichardson: Yes, I guess I am. But it looks like that assumption is already in the code? Line 646 uses `row_bs` but the vector is sized in line 544 using `index_map_bs`. Is it ever actually different?

jorgensd: Well, once you multiply out the block size for each of them, i.e.

unroll_dofmap = [row_dofs[i]*row_bs + k for k in range(row_bs)]

they should have the same span as 

(V.dofmap.index_map.size_local + V.dofmap.index_map.num_ghosts)*V.dofmap.index_map_bs

This means that at least in theory, the dofmap could be unrolled, with `dofmap.bs=1`, while the index map is blocked (with bs>1). That is why I have to unroll the marker with respect to index map block size, which is in turn comparable with the unrolled dofmap (`r` in line 464).

I know some year ago this was the case for `V.sub(i)`. This seems to no longer be the case, which raises a general issue of should the dofmap be allowed to have a different block size that its index map?

chrisrichardson: OK, well, forget that for now then. But I still think it can be a bit nicer like my code snippet above. And no need to count the ghost rows?

jorgensd: Non-collapsed spaces of blocked elements yields different dofmap.bs and index_map_bs:
```python
from mpi4py import MPI
import dolfinx
import numpy as np
import basix.ufl

mesh = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, 4, 4)
el = basix.ufl.element("Lagrange", mesh.basix_cell(), 1, shape=(3,))

Vi = V.sub(0)
print(Vi.dofmap.bs, Vi.dofmap.index_map_bs)

print(len(np.unique(Vi.dofmap.list.flatten())), (Vi.dofmap.index_map.size_local +
      Vi.dofmap.index_map.num_ghosts)*Vi.dofmap.index_map_bs)
```
which makes sense, as the sub space refers to the original dof indices.

jorgensd: > OK, well, forget that for now then. But I still think it can be a bit nicer like my code snippet above. And no need to count the ghost rows?

It wouldn't really clean up the logic if you blank out the ghosted rows separately, and remove the count of ghost rows, as you wouldn't be able to access `row_added[r]`.

Thus it would look like:
```c++
    {
      md::mdspan<T, md::dextents<std::size_t, 2>> A(Ab.data(), space_dim1,
                                                    space_dim0);
      auto row_dofs = dofmap1->cell_dofs(c);
      for (std::size_t i = 0; i < row_dofs.size(); ++i)
        for (int k = 0; k < row_bs; ++k)
        {
          std::int32_t r = row_dofs[i] * row_bs + k;
          if (r >= num_owned_rows)
            for (std::size_t j = 0; j < space_dim0; ++j)
              A(i * row_bs + k, j) = 0.0;
          else
            if (row_added[r])
              for (std::size_t j = 0; j < space_dim0; ++j)
                A(i * row_bs + k, j) = 0.0;
            else
              row_added[r] = 1;
        }
```
which i don't really think is any more readable.

chrisrichardson: Not quite what I had in mind. Can't you just clear all non-owned rows of `A` first?

jorgensd: My point is that yes, you could clear them first, but you would still have to keep the (size_local+num_ghosts)*bs in `row_added`, as the rows could be ghost rows.
The above suggestion assumes `rows_added.size()=num_dofs_local*bs`.

If we keep `rows_added.size()= (num_dofs_local+num_ghosted)*bs`, you can split it, but it doesn't really make the code much more readable in my opinion, as you still have to have the blocking in rows_added and you do the same looping twice.

chrisrichardson (approved):

PR #3951, opened 2025-10-17
Safer handling of PETSc destruction in high-level classes

jorgensd:
If any of our high-level wrappers of PETSc (`LinearProblem`, `NewtonSolver` or `NonlinearProblem` fails at initialization,
before creating PETSc objects, users get an exeception due to how we destroy the objects (as they haven't been intialized yet. (See for instance: https://fenicsproject.discourse.group/t/error-running-linearproblem-on-poisson-demo-on-macos-conda-env/18272/2)

This PR resolves this by checking if we have defined the attribute.

### DIFF
--- a/python/dolfinx/fem/petsc.py
+++ b/python/dolfinx/fem/petsc.py
@@ -879,10 +879,10 @@ class LinearProblem:
             self.solver.getPC().setFieldSplitIS(*fieldsplit_IS)
 
     def __del__(self):
-        for o in filter(
-            lambda o: o is not None, (self._solver, self._A, self._b, self._x, self._P_mat)
+        for obj in filter(
+            lambda obj: obj is not None, (self._solver, self._A, self._b, self._x, self._P_mat)
         ):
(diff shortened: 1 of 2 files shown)


python/dolfinx/fem/petsc.py
```diff
@@ -879,12 +879,9 @@ def __init__(
...
     def __del__(self):
-        self._solver.destroy()
-        self._A.destroy()
-        self._b.destroy()
-        self._x.destroy()
-        if self._P_mat is not None:
-            self._P_mat.destroy()
+        for attr in ("_solver", "_A", "_b", "_x", "_P_mat"):
+            if _attr := getattr(self, attr, None) is not None:
+                _attr.destroy()
```
```suggestion
        for o in filter(lambda o: o is not None, (self._solver, self._A, self._b, self._x, self._P_mat)):

            o.destroy()
```
`getattr` breaks all kinds of linting.

chrisrichardson (approved):

PR #3952, opened 2025-10-17
Fix shape for vtkOriginalPointIds and vtkGhostType

chrisrichardson:
This fixes a problem with ADIOS2 VTX BP files, which become corrupted if one process has no cells.
It seems that ADIOS2 cannot accept local arrays with size `(0)` but needs to be `(0, 1)`.


jorgensd:
I guess this is fine to me if it fixes @afasil9's issue.

PR #3957, opened 2025-10-19
Introduce `common::memory`

schnellerhase:
Introduces free function `common::memory`. Which provides a unit based access (in `B`, `KB`, `MB`, ...) to the implementation detail `common::impl::memory`, which computes for given types the used memory in bytes. This needs to be implemented for every type of which one wants to have access to a memory consumption (and its member types!).

This provides a user controllable memory profiling capability, up to a desired/required precision.

First step towards https://github.com/FEniCS/dolfinx/issues/1741.

TODO: 
- [x] consistency with [`__sizeof__`/`sys.getsizeof`](https://docs.python.org/3/library/sys.html#sys.getsizeof)


garth-wells:
Looks like more-or-less the right direction to me.

We should consider how pedantic we want to be, e.g. for a large mesh the memory used by the Basix element will be trivial and do we really care about this. Do we want the truest possible measure of the memory used by an object, or just the storage memory used for parts that depend on problem size?

jhale:
My understanding is that C++26 reflections could remove a lot of tedious boiler plate required to implement something like this.

schnellerhase:
> We should consider how pedantic we want to be, e.g. for a large mesh the memory used by the Basix element will be trivial and do we really care about this. Do we want the truest possible measure of the memory used by an object, or just the storage memory used for parts that depend on problem size?

Problem size dependant memory will give the most relevant insights, the other memory consumptions might help track down bugs, but probably do not add any relevant information for non-dev usage.

jhale:
It would still be good to get this done in a pragmatic way to cover all of our main classes and their heavy members with big heap allocations.

PR #3960, opened 2025-10-21
Add version depndent intersphinx mapping

jorgensd:


python/doc/source/conf.py
```diff
@@ -152,10 +152,23 @@
...
+    basix_version = basix.__version__
+    ffcx_version = ffcx.__version__
 intersphinx_mapping = {
-    "basix": ("https://docs.fenicsproject.org/basix/main/python/", None),
-    "ffcx": ("https://docs.fenicsproject.org/ffcx/main/", None),
-    "ufl": ("https://docs.fenicsproject.org/ufl/main/", None),
+    "basix": (f"https://docs.fenicsproject.org/basix/v{basix_version}/python/", None),
+    "ffcx": (f"https://docs.fenicsproject.org/ffcx/v{ffcx_version}/python/", None),
+    "ufl": (f"https://docs.fenicsproject.org/ufl/{ufl_version}/python/", None),
     "mpi4py": ("https://mpi4py.readthedocs.io/en/stable", None),
```
jhale: It's going to be best to do the same with these dependencies as well.

python/doc/source/conf.py
```diff
@@ -152,10 +152,23 @@
 
 codeautolink_concat_default = True
 
+# The version info for ufl, basix and ffcx used for intersphinx
+if "dev0" in dolfinx.__version__:
```
jhale: Would be clearer just to import each one and then do a one line if contains dev0 then main else version?

jhale: I would rather not bake in any assumptions on everything being a dev0 version.

python/doc/source/conf.py
jorgensd: We can do it for `mpi4py`, but not for `petsc4py`

python/doc/source/conf.py
jorgensd: I've made an attempt to improve this. let me know what you think

PR #3961, opened 2025-10-21
Fix typing error

mscroggs:
Fix mypy error that the Basix CI (eg https://github.com/FEniCS/basix/actions/runs/18676031000/job/53246108920?pr=965) is picking up.

### DIFF
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -133,7 +133,9 @@ class Expression:
         # Get MPI communicator
         if comm is None:
             try:
-                mesh = ufl.domain.extract_unique_domain(e).ufl_cargo()
+                domain = ufl.domain.extract_unique_domain(e)
+                assert isinstance(domain, ufl.Mesh)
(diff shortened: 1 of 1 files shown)


python/dolfinx/fem/function.py
```diff
@@ -133,7 +133,9 @@ def __init__(
         # Get MPI communicator
         if comm is None:
             try:
-                mesh = ufl.domain.extract_unique_domain(e).ufl_cargo()
+                domain = ufl.domain.extract_unique_domain(e)
+                assert isinstance(domain, ufl.Mesh)
+                mesh = domain.ufl_cargo()
```
mscroggs, line 138: Not sure if we'd rather add an assertion of a type or add `type: ignore` to this line?
```suggestion
                mesh = ufl.domain.extract_unique_domain(e).ufl_cargo()  # type: ignore
```

jorgensd: I think it is fine to have the assert.

Issue #1671, opened 2021-08-17
`create_entity_permutations` should take a dimension as an input


mscroggs:
For facet integrals, only the permutations of facets are needed, but all the permutations are created. `create_entity_permutations should take a dimension, then only create the permutations of that dimension so that we can avoid creating those we don't use.

See also FEniCS/ffcx#383

Issue #1250, opened 2020-12-12
Can mesh entity creation be simplified in Topology::create_entity_permutations()?


garth-wells:
This step is rather costly. Perhaps it could be made simpler?

mscroggs:
- [ ] Can we compute per dimension? (ie add a dimension input to the function)
- [x] Can we do it threaded?

jorgensd:
Relates to: https://github.com/FEniCS/dolfinx/pull/3904

chrisrichardson:
Can we close this, following #4270 ?

PR #3974, opened 2025-11-03
Undefined call to cell_num_entities for point meshes.

jorgensd:
For point/vertex meshes, creating facet permutations does not make sense.
Currently cell_num_entities(vertex,-1) results in different behavior in Release and Debug mode.

This is fix until we do a fundamental redesign of `compute_entity_permutations` (ref https://github.com/FEniCS/dolfinx/issues/1671 https://github.com/FEniCS/dolfinx/issues/3856 https://github.com/FEniCS/dolfinx/issues/1250)

I would really like this to be backported to 0.10, as it is breaking some of my packages that do 1D-0D coupling.

### DIFF
--- a/cpp/dolfinx/mesh/permutationcomputation.cpp
+++ b/cpp/dolfinx/mesh/permutationcomputation.cpp
@@ -293,9 +293,7 @@ mesh::compute_entity_permutations(const mesh::Topology& topology)
   const std::int32_t num_cells = topology.connectivity(tdim, 0)->num_nodes();
   // Point meshes have no facets per cell and cell_num_entities(vertex, -1) is
   // undefined
-  int facets_per_cell = 0;
-  if (tdim > 0)
-    facets_per_cell = cell_num_entities(cell_type, tdim - 1);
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/mesh/permutationcomputation.cpp
```diff
@@ -291,7 +291,11 @@ mesh::compute_entity_permutations(const mesh::Topology& topology)
   const int tdim = topology.dim();
   CellType cell_type = topology.cell_type();
   const std::int32_t num_cells = topology.connectivity(tdim, 0)->num_nodes();
-  const int facets_per_cell = cell_num_entities(cell_type, tdim - 1);
+  // Point meshes have no facets per cell and cell_num_entities(vertex, -1) is
+  // undefined
+  int facets_per_cell = 0;
+  if (tdim > 0)
+    facets_per_cell = cell_num_entities(cell_type, tdim - 1);
```
chrisrichardson:
```suggestion
  int facets_per_cell = (tdim > 0) ? cell_num_entities(cell_type, tdim - 1) : 0;
```

chrisrichardson (approved):

PR #3977, opened 2025-11-06
Remove `num_entity_dofs` and `num_entity_closure_dofs` as they are inconsistent for pyramid/prism cells

chrisrichardson:
I came across this problem when trying to create a `Mesh` with quadratic prism cells. The topology creation only considered `num_entity_dofs` on the first (triangular) facet (=0), which then caused a crash for the quad facets.
The number of entity dofs can easily be computed from `entity_dofs_all()` anyway.

### DIFF
--- a/cpp/dolfinx/fem/utils.cpp
+++ b/cpp/dolfinx/fem/utils.cpp
@@ -39,7 +39,7 @@ fem::DofMap fem::create_dofmap(
   // Create required mesh entities
   const int D = topology.dim();
 
-  auto entity_dofs = layout.entity_dofs_all();
+  const auto& entity_dofs = layout.entity_dofs_all();
   for (int dim = 1; dim < topology.dim(); ++dim)
   {
     // Accumulate count of all dofs on this dimension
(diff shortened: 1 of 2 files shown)


cpp/dolfinx/io/cells.cpp
```diff
@@ -347,6 +347,8 @@ std::vector<std::uint16_t> vtk_wedge(int num_nodes)
     return {0, 1, 2, 3, 4, 5};
   case 15:
     return {0, 1, 2, 3, 4, 5, 6, 9, 7, 12, 14, 13, 8, 10, 11};
+  case 18:
+    return {0, 1, 2, 3, 4, 5, 6, 9, 7, 12, 14, 13, 8, 10, 11, 15, 17, 16};
```
jorgensd, line 351: Does this add any features?

chrisrichardson: 18 point prism

chrisrichardson:
ElementDofLayout already holds the structure, so there is no allocation needed when calling `entity_dofs_all`. It is a const reference.

PR #3978, opened 2025-11-10
Fix: copy drops function name

schnellerhase:
Also adapts to `pytest` changes enforcing either `ini_options` or TOML based config https://github.com/pytest-dev/pytest/releases/tag/9.0.0 not both.


schnellerhase:
Failure in spack ci should be resolved as soon as https://github.com/spack/spack-packages/pull/2347 goes in.

PR #3984, opened 2025-11-12
More robust ctest setup for MPI C++ tests.

jhale:
Moves MPI launching back inside ctest by using the output of CMake's FindMPI.

### DIFF
--- a/cpp/test/CMakeLists.txt
+++ b/cpp/test/CMakeLists.txt
@@ -102,7 +102,7 @@ endif()
 # Uncommenting line allows for debugging individual tests, but slow.
 #catch_discover_tests(unittests)
 
-# To subset of tests use:
+# To select one np use, e.g.:
 # ctest -R unittests_np_3
 foreach(N 1 2 3)
   add_test(NAME unittests_np_${N}


.github/workflows/ccpp.yml
```diff
@@ -123,10 +123,7 @@ jobs:
 
       - name: Run tests via target (C++, serial)
```
```suggestion
      - name: Run tests via target (C++, serial and parallel)
```

cpp/test/CMakeLists.txt
```diff
@@ -100,4 +102,18 @@ endif()
...
+           COMMAND
+           ${MPIEXEC_EXECUTABLE}
+           ${MPIEXEC_NUMPROC_FLAG} ${N}
+           ${MPIEXEC_PREFLAGS}
+           $<TARGET_FILE:unittests>
+           ${MPIEXEC_POSTFLAGS}
+  )
+  # This tells ctest how many physical cores (i.e., ${N}) are needed for each
+  # test, preventing oversubscription when using e.g. `ctest -j2`.
+  set_tests_properties(unittests_np_${N} PROPERTIES PROCESSORS ${N})
```
schnellerhase, line 118: Nice! Should we add a global `CTEST_PARALLEL_LEVEL: 4` to the CI runs then?

jhale: I don't see a practical need at the minute, our C++ tests run very fast anyway. But at least it is setup properly for local users and the future. More generally, not a big fan of environment variables, if we do it in the future we should use `-j4` or similar.

PR #3992, opened 2025-11-17
Add Conda 3.13 to weekly conda install check

jhale:
- **Remove deprecated macos-13 runner, add Python 3.13.**
- **Exclude Windows 3.13.**


jhale:
Also removes macos-13 due to upcoming deprecation, adds macos-15. Note that Windows only has old 0.9 version - @minrk any updates on Conda's Intel MPI stack? 0.10 does build on Windows. although, I also cannot get it to run with Intel MPI with np=2.

PR #3996, opened 2025-11-20
Fix unsigned int bug in sort test and C++ standard library assertions

jhale:
This PR:

- fixes the unsigned int bug discovered by @drew-parsons related to generating a uniform distribution using negative lower bound, despite the integers type being unsigned.
- Ensures that our Developer mode triggers exceptions in the C++ standard library which allowed Drew to discover the bug. I suspect this was accidentally turned off by using `-O2` but not completely certain.

As an aside, there are quite a number of useful compile warnings not included `-Wall` which we should consider turning on.

These extra checks produce no discernable increase in test time.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -237,13 +237,13 @@ endif()
 # Debug flags
 check_cxx_compiler_flag(-g HAVE_DEBUG)
 if(HAVE_DEBUG)
-  list(APPEND DOLFINX_CXX_DEVELOPER_FLAGS "-g")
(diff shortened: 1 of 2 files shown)


jhale:
My understanding is that Debian and Ubuntu are starting to use GCC options like -fhardening which enables hardened C++ standard libraries with more asserts enabled, amongst other things. This picked up this bug, but we didn't see it in our CI.

To avoid this, we will be enabling more hardening on our CI systems, which use our custom CMake Developer build type, over the coming weeks.

We do not plan to adjust the standard CMake built types, e.g. Release - we leave this to downstream packagers discretion.

PR #3998, opened 2025-11-20
Experiments with more compiler flags

jhale:


cpp/test/CMakeLists.txt
```diff
@@ -19,6 +23,19 @@ if (PROJECT_IS_TOP_LEVEL)
...
+    list(APPEND DOLFINX_CXX_DEVELOPER_FLAGS -fno-omit-frame-pointer)
+  endif()
+
+  # Add fast memory error detector
+  check_cxx_compiler_flag("-fsanitize=address" HAVE_ADDRESS_SANITIZER)
+  if(HAVE_ADDRESS_SANITIZER)
+    list(APPEND DOLFINX_CXX_DEVELOPER_FLAGS -fsanitize=address)
+    list(APPEND DOLFINX_LINK_DEVELOPER_FLAGS -fsanitize=address)
+  endif()
+
```
schnellerhase, line 38: We should think about a nicer way to maintain these centrally and not twice in `cpp` and `cpp/test`.

schnellerhase (commented):
Add the flags to `generate-cmakefiles.py` and regenerate, for the same behaviour on the `C++` demos?

cpp/test/CMakeLists.txt
jhale: We should probably write one or more CMakePreset.json file, e.g.: https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html and remove this.

jhale: Or/plus a toolchain file:

https://cmake.org/cmake/help/latest/manual/cmake-toolchains.7.html

jhale: Not sure actually, an include looks simplest.

schnellerhase: I'd go with include as well. Maybe we can also export it through the install of dolfinx? For both the standalone tests and the demos we require anyways the presence of a dolfinx install, so we could inherit the flags from that one, needs to be exported in that case.

PR #4001, opened 2025-11-24
Adds `xfail` test for scotch partitioner

schnellerhase:
Discovered by @drew-parsons 

Scotch partitioner fails on `test_mesh_single_process_distribution` in parallel, while KaHIP and Parmetis produce the expected result. This needs further debugging. This PR ensures a passing CI.

### DIFF
--- a/python/test/unit/mesh/test_mesh.py
+++ b/python/test/unit/mesh/test_mesh.py
@@ -782,6 +782,5 @@ def test_mesh_single_process_distribution(partitioner):
     for conn in ((0, 1), (1, 0)):
         mesh.topology.create_connectivity(*conn)
         adj = mesh.topology.connectivity(*conn)
-        print(adj.array)
         for i in range(adj.num_nodes):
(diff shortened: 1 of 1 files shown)


python/test/unit/mesh/test_mesh.py
```diff
@@ -736,7 +737,19 @@ def test_mesh_create_cmap(dtype):
     assert msh.ufl_domain() is None
 
 
-def test_mesh_single_process_distribution():
+avail_partioners = []
+if dolfinx.has_ptscotch:
+    avail_partioners.append(
+        pytest.param(dolfinx.cpp.graph.partitioner_scotch, marks=pytest.mark.xfail(reason="Bug"))
```
drew-parsons: Give a bit more context in the skip message, a few more words than just "Bug".

drew-parsons (changes requested):
It'll keep CI passing until the underlying issue is found.
Give the message a bit more detail than just "Bug"

PR #4004, opened 2025-11-25
fix build with adios2 v2.11.0

qbisi:
The target cxx11_mpi has been renamed to cxx_mpi since adios2 v2.11.0.


schnellerhase:
New ADIOS version also breaks current spack `develop` builds.

PR #4005, opened 2025-11-25
Experiments with C++/C standards

jhale:


jhale:
`set(CMAKE_C_EXTENSIONS OFF)` and `set(CMAKE_CXX_EXTENSIONS OFF)` should be called extremely early, otherwise compiler extensions will be turned on, even when using `cxx_std_20 c_std_23` on the target. Not clear why. Lots of boiler plate to update.

PR #4012, opened 2025-11-29
Use `setup-spack` action in spack CIs

schnellerhase:


schnellerhase:
Mainly motivated by simpler setup and syntax in CIs, noted that the other spack CI already used it. Which shell type breaks?

PR #4013, opened 2025-12-01
Pass custom_data through assemblers to kernel functions.

sclaus2:
Add void* custom_data member to integral_data struct and pass it to assembly functions (matrix, vector, scalar, lifting). Custom_data defaults back to nullptr if no custom_data is set explicitly (with set_custom_data()). The required changes are minimal 

- Form.h: Add custom_data to integral_data, add accessors
- assemble_matrix_impl.h: Pass custom_data through all functions 
- assemble_vector_impl.h: Pass custom_data through all functions
- assemble_scalar_impl.h: Pass custom_data through all functions
- Python bindings: Expose set_custom_data() and custom_data()
- Add test_custom_data.py with comprehensive tests

This enables integration kernels to receive additional per-cell user data which is essential for example for runtime quadrature rules. With this extensions users are no longer required to completely rewrite all assembly routines for customised C-kernels. 

For an example library that relies on custom_data and its integration into the assembler see 

https://github.com/sclaus2/runintgen


jhale:
I say probably because it's a performance critical part of the code - might need a bit of thought.

python/dolfinx/wrappers/fem.cpp
```diff
@@ -761,6 +761,31 @@ void declare_form(nb::module_& m, std::string type)
...
+             int id, int kernel_idx, std::optional<std::uintptr_t> data)
+          {
+            self.set_custom_data(type, id, kernel_idx,
+                                 data ? std::optional<void*>((void*)*data)
+                                      : std::nullopt);
+          },
+          nb::arg("type"), nb::arg("id"), nb::arg("kernel_idx"),
+          nb::arg("data").none(),
+          "Set custom data pointer for an integral. The data pointer is "
+          "passed to the kernel. Pass None to clear.")
```
jhale: I think the last part "Pass None to clear" is unnecessary. Maybe instead explain that None means no data.

python/test/unit/fem/test_custom_data.py
```diff
@@ -0,0 +1,287 @@
...
+from dolfinx import la
+from dolfinx.fem import Form, IntegralType, form_cpp_class, functionspace
+from dolfinx.mesh import create_unit_square
+
+numba = pytest.importorskip("numba")
+ufcx_signature = ffcx.codegeneration.utils.numba_ufcx_kernel_signature
+
+
+# Helper intrinsic to cast void* to a typed pointer for custom_data
+@numba.extending.intrinsic
```
jhale: Do we already have this intrinsic in FFCx?

python/test/unit/fem/test_custom_data.py
```diff
@@ -0,0 +1,287 @@
...
+
+    Note: custom_data must be set to a valid pointer before assembly.
+    """
+
+    @numba.cfunc(ufcx_signature(dtype, xdtype), nopython=True)
+    def tabulate(b_, w_, c_, coords_, local_index, orientation, custom_data):
+        b = numba.carray(b_, (3), dtype=dtype)
+        coordinate_dofs = numba.carray(coords_, (3, 3), dtype=xdtype)
+
+        # Cast void* to float64* and read the scale value
```
jhale: Nice!

python/test/unit/fem/test_custom_data.py
```diff
@@ -0,0 +1,287 @@
...
+    Note: custom_data must be set to a valid pointer before assembly.
+    """
+
+    @numba.cfunc(ufcx_signature(dtype, xdtype), nopython=True)
+    def tabulate(A_, w_, c_, coords_, entity_local_index, cell_orientation, custom_data):
+        A = numba.carray(A_, (3, 3), dtype=dtype)
+        coordinate_dofs = numba.carray(coords_, (3, 3), dtype=xdtype)
+
+        # Cast void* to float64* and read the scale value
+        typed_ptr = voidptr_to_float64_ptr(custom_data)
```
jhale: Do we have any more complex examples? For example, an array of float64s index by the current cell. This is what @chrisrichardson mentioned before "a kernel might need to know which cell it's running on".

jhale: I saw the complex examples/tests below with a length-two region of memory, but they don't quite hit the point just made.

cpp/dolfinx/fem/Form.h
```diff
@@ -391,6 +399,41 @@ class Form
...
+                                   int kernel_idx) const
+  {
+    auto it = _integrals.find({type, id, kernel_idx});
+    if (it == _integrals.end())
+      throw std::runtime_error("Requested integral not found.");
+    return it->second.custom_data;
+  }
+
+  /// @brief Set the custom data pointer for an integral.
+  ///
```
jhale: I would like to add a warning here:

"void-pointers are inherently unsafe and cannot be type or bounds checked. Incorrect usage of this feature can and will lead to undefined behaviour and crashes."

python/dolfinx/wrappers/fem.cpp
```diff
@@ -761,6 +761,31 @@ void declare_form(nb::module_& m, std::string type)
       .def_prop_ro("integral_types", &dolfinx::fem::Form<T, U>::integral_types)
       .def_prop_ro("needs_facet_permutations",
                    &dolfinx::fem::Form<T, U>::needs_facet_permutations)
+      .def(
+          "set_custom_data",
+          [](dolfinx::fem::Form<T, U>& self, dolfinx::fem::IntegralType type,
+             int id, int kernel_idx, std::optional<std::uintptr_t> data)
+          {
+            self.set_custom_data(type, id, kernel_idx,
```
jhale: `set_custom_data` doesn't quite fit the design of the rest of the class - the custom data pointer should be passed up (or not) at creation time, and then it is immutable (at least, the pointer is - the user could modify the pointed memory of course).

python/dolfinx/wrappers/fem.cpp
```diff
@@ -761,6 +761,31 @@ void declare_form(nb::module_& m, std::string type)
...
+          },
+          nb::arg("type"), nb::arg("id"), nb::arg("kernel_idx"),
+          nb::arg("data").none(),
+          "Set custom data pointer for an integral. The data pointer is "
+          "passed to the kernel. Pass None to clear.")
+      .def(
+          "custom_data",
+          [](const dolfinx::fem::Form<T, U>& self,
+             dolfinx::fem::IntegralType type, int id,
+             int kernel_idx) -> std::optional<std::uintptr_t>
```
jhale: Why are we returning `std::uintptr_t` and not `void*`? If there is a good reason there needs to be a comment.

python/test/unit/fem/test_custom_data.py
```diff
@@ -0,0 +1,287 @@
...
+    # Create struct data: [scale=2.0, offset=0.5]
+    struct_data = np.array([2.0, 0.5], dtype=dtype)
+    L._cpp_object.set_custom_data(IntegralType.cell, 0, 0, struct_data.ctypes.data)
+
+    b = dolfinx.fem.assemble_vector(L)
+    b.scatter_reverse(la.InsertMode.add)
+
+    # Verify the assembly used our custom values
+    # The offset should contribute to each DOF
+    assert la.norm(b) > 0
```
jhale: I suspect this `assert` always comes up `true`. Can you do something more precise?

jhale (changes requested):
Looks pretty good - a few small changes to make.

python/dolfinx/wrappers/fem.cpp
sclaus2: I removed set_custom_data all together in favour of setting custom_data through integrals and Form constructor.

cpp/dolfinx/fem/Form.h
sclaus2: I added the warning to the custom_data method in C++ and python.

python/dolfinx/wrappers/fem.cpp
sclaus2: I removed it in favour of setting custom_data through the form constructor.

python/dolfinx/wrappers/fem.cpp
sclaus2: I added a comment : "We use std::uintptr_t to map void* into Python to allow seamless interop with numpy.ctypes.data."

python/test/unit/fem/test_custom_data.py
sclaus2: You and @chrisrichardson are absolutely right. For complexer per cell custom data a local index or entity index is required in the kernel. How this can be passed probably requires a more in-depth discussion.

python/test/unit/fem/test_custom_data.py
sclaus2: I changed the tests and added more comprehensive ones.

python/test/unit/fem/test_custom_data.py
sclaus2: FFCx doesn't have intrinsics for casting from void* to typed pointers. The FFCx tests use CFFI for that (C code).
However, FFCx has intrinsics for going to void* ([get_void_pointer], [empty_void_pointer]). 
We could add casting from void* to typed pointer  intrinsics to FFCx or keep it in the test (demonstration only).

sclaus2: Moved to ffcx see PR [https://github.com/FEniCS/ffcx/pull/804](url)

python/test/unit/fem/test_custom_data.py
sclaus2: Will be discussed in separate PR.

python/test/unit/fem/test_custom_data.py
```diff
@@ -0,0 +1,415 @@
...
+        Ae = abs((x0 - x1) * (y2 - y1) - (y0 - y1) * (x2 - x1))
+        B = np.array([y1 - y2, y2 - y0, y0 - y1, x2 - x1, x0 - x2, x1 - x0], dtype=dtype).reshape(
+            2, 3
+        )
+        A[:, :] = scale * np.dot(B.T, B) / (2 * Ae)
+
+    return tabulate
+
+
+@pytest.mark.parametrize("dtype", [np.float64])
```
jhale: Would be good to test both np.float64 and np.float32.

python/test/unit/fem/test_custom_data.py
```diff
@@ -0,0 +1,415 @@
...
+    # Pass None as custom_data (5th element) to get std::nullopt
+    integrals = {IntegralType.cell: [(0, tabulate_simple.address, cells, active_coeffs, None)]}
+    formtype = form_cpp_class(dtype)
+    L = Form(formtype([V._cpp_object], integrals, [], [], False, [], mesh=mesh._cpp_object))
+
+    # custom_data should be None (std::nullopt) when passed as None
+    assert L._cpp_object.custom_data(IntegralType.cell, 0, 0) is None
+
+
+@pytest.mark.parametrize("dtype", [np.float64])
```
jhale: Extend all tests to use float64 and float32.

jhale (changes requested):

python/test/unit/fem/test_custom_data.py
sclaus2: done

python/test/unit/fem/test_custom_data.py
sclaus2: done

PR #4015, opened 2025-12-08
Exclude nanobind 2.10.0

schnellerhase:
Ref. https://github.com/wjakob/nanobind/issues/1237. Together with https://github.com/FEniCS/basix/pull/980 should resolve build problems.


.github/workflows/fenicsx-refs.env
```diff
@@ -1,5 +1,5 @@
 basix_repository=FEniCS/basix
-basix_ref=main
+basix_ref=schnellerhase/nanobind-broken
```
schnellerhase, line 2:
```suggestion
basix_ref=main
```

PR #4025, opened 2025-12-19
Write cg1 and dg0 functions to VTKHDF

mleoni-pf:
This PR introduces a convenience function to write CG1 and DG0 functions to VTKHDF.

### DIFF
--- a/cpp/dolfinx/io/VTKHDF.h
+++ b/cpp/dolfinx/io/VTKHDF.h
@@ -6,9 +6,13 @@
 
 #include "HDF5Interface.h"
 #include <algorithm>
+#include <basix/element-families.h>
+#include <basix/finite-element.h>
 #include <concepts>
 #include <dolfinx/common/IndexMap.h>
+#include <dolfinx/fem/Function.h>
 #include <dolfinx/io/cells.h>
+#include <dolfinx/io/utils.h>
 #include <dolfinx/mesh/Mesh.h>
 #include <dolfinx/mesh/Topology.h>
 #include <dolfinx/mesh/utils.h>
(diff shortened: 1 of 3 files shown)


cpp/dolfinx/io/VTKHDF.h
```diff
@@ -298,6 +300,68 @@ void write_data(std::string point_or_cell, std::string filename,
...
+/// @param[in] u Function to write to file.
+/// @param[in] time Timestamp.
+///
+/// @note Mesh must be written to file first using `VTKHDF::write_mesh`.
+/// @note Only one dataset "u" can be written per file at present, with
+/// multiple timesteps.
+/// @note Limited support for floating point types at present (no
+/// complex number support).
+template <std::floating_point U>
+void write_CG1_function(std::string filename, const mesh::Mesh<U>& mesh,
```
schnellerhase: The decision to split up into different functions depending on function space looks problematic, it will not scale. A generic `write_function` would be favourable.

mleoni-pf: I was inspired by the Python interface having separate "write point data" and "write cell data" but I realize that that's not quite the same... I agree that a generic `write_function` would be better but I am not knowledgeable enough on the VTKHDF format to know how to write mixed or exotic elements...
An API-saving option could be to error out for unsupported function spaces; another one, used somewhere else in the repo though unsatisfactory, could be to interpolate whatever function is passed to CG1 or DG0 before saving. What are your thoughts?

schnellerhase: I think supporting subsets of elements is fine. Good to add tests, which check for a proper assertion being raised on such invocations.

chrisrichardson: I think in this case, raise an exception for unsupported function spaces.

schnellerhase:
> I would not go down the route of what we do in XDMFFile, which is restricting ourselves to writing to the mesh nodes (which involves a non-trivial interpolation step if function space!=geometry space).

`VTKHDF` supports arbitrary cell types - https://gitlab.kitware.com/keu-public/vtkhdf/vtkhdf-scripts/-/tree/main/tutorial?ref_type=heads#creating-a-basic-unstructured-grid. Therefore, if I understand correctly, we can aim for an actual exact high order output in the long run, i.e. supporting cell types such as [VTK_HIGHER_ORDER_TETRAHEDRON](https://vtk.org/doc/nightly/html/vtkCellType_8h.html#ab1d6fd1f3177b8a2a32bb018807151f8a13cec93631187d6be55347bd23f96552). If one aims for this, the dofmap based construction is not exactly what we need for the higher order case, as we do not want to rely on locally linear interpolations, right?

mleoni-pf:
Where do we stand on the discussion? Am I right just waiting for a review? :) @chrisrichardson @schnellerhase

jorgensd:
A key assumption in this PR as that one can only save a single function (time-dependent) per file.
I think one should aim for storing multiple functions, where the API would look like:
```python
with dolfinx.io.VTKHDFWriter(comm, filename, [list_of_P1_and_DG0-functions]) as vtkhdf:
    vtkhdf.write(1.1)
   # Update functions
   # ...
    vtkhdf.write(2.3)
```
as all functions in VTKHDF needs to have synchronized time-steps (For temporal datasets, all blocks should have the same number of time steps and time values.) (https://docs.vtk.org/en/latest/vtk_file_formats/vtkhdf_file_format/vtkhdf_specifications.html#partitioneddatasetcollection-and-multiblockdataset)
Even if we do not consider a multi-block problem, point data and cell data must update eachothers offsets if stored in the same file at different times, i.e. using the approach of writing (function, time_stamp) to file would result in:
```python
with dolfinx.io.VTKHDFWriter(comm, filename) as vtkhdf:
     vtkhdf.write_mesh(mesh)
     vtkhdf.write_function(u_P1, t0)
     # The following wont work as u_P1 also has to update offset counts.
     vtkhdf.write_function(p_DG0, t1)

PR #4029, opened 2025-12-23
Fix order of arguments in docstring of `dolfinx.mesh.create_mesh`

finsberg:
Make sure the order of the arguments in the docstring matches the order of the arguments to the function


schnellerhase (approved):
We should alter the settings of ruff, that should have been caught there.

PR #4033, opened 2026-01-05
Add better error handling for `locate_dofs_topological`

### LINKED ISSUE #4010: [BUG]: locate_dofs_topological segmentation (core dumped)
### Summarize the issue

I don't have time to write a MWE. I use FEniCSx v 0.1 (latest stable at the moment).
I have a mesh coming from Gmsh that I read with the dolfinx api.
I know how to fix the problem, so this isn't urgent, but still, I don't believe FEniCSx should core dump without leaving any trace of the error.

The code: 
```
        dim = mesh.topology.dim
        volt_boundary_dofs = locate_dofs_topological(ME.sub(1), dim-1, bottom_facets)
```
where bottom_facets is worth facet_markers.find(12). This code is fine. 

But if I accidentally put a left_dofs_temp = locate_dofs_topological(ME.sub(0), mesh.topology.dim-1, bottom_facets) instead of the bottom_facets... then the program core dumps.

### How to reproduce the bug

No MWE at the moment.

### Minimal Example (Python)

```Python

```

### Output (Python)

```bash
zsh: segmentation fault (core dumped)  /home/user/.conda/envs/fenicsx_0dot1_-env/bin/python
```

### Version

main branch

### DOLFINx git commit

_No response_

### Installation

_No response_

### Additional information

_No response_

jorgensd:
Resolves #4010

### DIFF
--- a/cpp/dolfinx/fem/DirichletBC.cpp
+++ b/cpp/dolfinx/fem/DirichletBC.cpp
@@ -23,13 +23,13 @@ using namespace dolfinx::fem;
 
 namespace
 {
-
 /// @brief Find the cell (local to process) and index of an entity
 /// (local to cell) for a list of entities.
+///
 /// @param[in] mesh The mesh
 /// @param[in] entities The list of entities
 /// @param[in] dim The dimension of the entities
-/// @returns A list of (cell_index, entity_index) pairs for each input
+/// @returns A list of `(cell_index, entity_index)` pairs for each input
 /// entity.
 std::vector<std::pair<std::int32_t, int>>
 find_local_entity_index(const mesh::Topology& topology,
@@ -59,7 +59,7 @@ find_local_entity_index(const mesh::Topology& topology,
   for (std::int32_t e : entities)
   {
     // Get first attached cell
-    if (num_entities_local <= e)
+    if (e >= num_entities_local)
     {
       throw std::out_of_range(
           "Input entity " + std::to_string(e)
@@ -67,11 +67,11 @@ find_local_entity_index(const mesh::Topology& topology,
           + std::to_string(num_entities_local) + ").");
     }
     assert(e_to_c->num_links(e) > 0);
-    const int cell = e_to_c->links(e).front();
 
     // Get local index of facet with respect to the cell
+    std::int32_t cell = e_to_c->links(e).front();
     auto entities_d = c_to_e->links(cell);
-    auto it = std::find(entities_d.begin(), entities_d.end(), e);
+    auto it = std::ranges::find(entities_d, e);
     assert(it != entities_d.end());
     std::size_t entity_local_index = std::distance(entities_d.begin(), it);
     entity_indices.emplace_back(cell, entity_local_index);
@@ -82,7 +82,7 @@ find_local_entity_index(const mesh::Topology& topology,
 //-----------------------------------------------------------------------------
 
 /// Find all DOFs on this process that have been detected on another
-/// process
+/// process.
 /// @param[in] comm A symmetric communicator
 /// @param[in] map The index map with the dof layout
 /// @param[in] bs_map The block size of the index map, i.e. the dof


cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -55,9 +55,17 @@ find_local_entity_index(const mesh::Topology& topology,
 
   std::vector<std::pair<std::int32_t, int>> entity_indices;
   entity_indices.reserve(entities.size());
+  std::int32_t num_entities_local = e_to_c->num_nodes();
   for (std::int32_t e : entities)
   {
     // Get first attached cell
+    if (num_entities_local <= e)
```
schnellerhase:
```suggestion
    if (num_entities_local <= e) [[unlikely]]
```
Maybe good to add here.

jorgensd: I don't think it will yield an optimization here, as there is no else-case, and there is no performance critical operation happening within the if test.

At least I get the same machine code for both GCC and clang with:
```c++
#include<vector>
#include<cstdint>
#include<stdexcept>

int test (std::int32_t max_size)
{
    std::vector<std::int32_t> a;
    for (auto ai: a)
    {
    if (ai >=max_size) [[likely]]
    // if (ai >=max_size) [[unlikely]]
    {
        throw std::runtime_error("error");
    }
    auto z = ai + 1;
    }
    return 0;

}
```

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -55,9 +55,17 @@ find_local_entity_index(const mesh::Topology& topology,
...
+    {
+      throw std::runtime_error(
+          "Input entity " + std::to_string(e)
+          + " is bigger than the number of entities on this process ("
```
jhale: larger than

jorgensd:
```suggestion
          + " is larger than the number of entities on this process ("
```

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -55,9 +55,17 @@ find_local_entity_index(const mesh::Topology& topology,
 
   std::vector<std::pair<std::int32_t, int>> entity_indices;
   entity_indices.reserve(entities.size());
+  std::int32_t num_entities_local = e_to_c->num_nodes();
   for (std::int32_t e : entities)
   {
     // Get first attached cell
+    if (num_entities_local <= e)
+    {
+      throw std::runtime_error(
```
jhale: std::out_of_range is more precise?

jorgensd:
```suggestion
      throw std::out_of_range(
```

python/test/unit/fem/test_bcs.py
```diff
@@ -406,3 +406,16 @@ def test_blocked_dof_ownership(shape):
...
+
+def test_bc_index_out_of_range():
+    mesh = create_unit_square(MPI.COMM_WORLD, 4, 4)
+    mesh.topology.create_connectivity(mesh.topology.dim - 1, mesh.topology.dim)
+    facet_map = mesh.topology.index_map(mesh.topology.dim - 1)
+    num_facets_on_proc = facet_map.size_local + facet_map.num_ghosts
+    facets = np.arange(num_facets_on_proc + 1, dtype=np.int32)
+    V = functionspace(mesh, ("Lagrange", 1))
+    with pytest.raises(
+        RuntimeError, match=r".*is bigger than the number of entities on this process.*"
```
jorgensd:
```suggestion
        IndexError, match=r".*is bigger than the number of entities on this process.*"
```

python/test/unit/fem/test_bcs.py
```diff
@@ -406,3 +406,16 @@ def test_blocked_dof_ownership(shape):
...
+
+def test_bc_index_out_of_range():
+    mesh = create_unit_square(MPI.COMM_WORLD, 4, 4)
+    mesh.topology.create_connectivity(mesh.topology.dim - 1, mesh.topology.dim)
+    facet_map = mesh.topology.index_map(mesh.topology.dim - 1)
+    num_facets_on_proc = facet_map.size_local + facet_map.num_ghosts
+    facets = np.arange(num_facets_on_proc + 1, dtype=np.int32)
+    V = functionspace(mesh, ("Lagrange", 1))
+    with pytest.raises(
+        IndexError, match=r".*is bigger than the number of entities on this process.*"
```
jorgensd:
```suggestion
        IndexError, match=r".*is larger than the number of entities on this process.*"
```

PR #4038, opened 2026-01-12
Fix interpolation onto submesh with maps requiring dof transformations

jorgensd:
Two bug fixes:

## Interpolation onto submesh with same element type

Currently, interpolation from space A to space B doesn't use permutation info if element_A == element_B.
However, this is not correct when working with submeshes, as the cell_permutation_data might be different between the meshes.

The initial issue was discovered by @afasil9 when working on TEAM30 (https://github.com/Wells-Group/TEAM30/).

This pull request add a test that fails on 2,3 and 4 processors on main and a fix to interpolate.

## Interpolation onto same function space (and mesh), with subset of cells
If the number of cells (`cells0.size()`) passed into `u.interpolate(v, cells0=cells0)` 
matches the number of cells on the process (including ghosts), we currently assume that `cells0=np.arange(num_cells_on_process)`, which is not True, as cells0 can contain duplicate cells.

This is fixed by adding a progressive set of checks to determine if this is True.

### DIFF
--- a/cpp/dolfinx/fem/interpolate.h
+++ b/cpp/dolfinx/fem/interpolate.h
@@ -1263,7 +1263,7 @@ void interpolate(Function<T, U>& u1, std::span<const std::int32_t> cells1,
   auto cell_map1 = mesh1->topology()->index_map(mesh1->topology()->dim());
   assert(cell_map1);
   std::size_t num_cells1 = cell_map1->size_local() + cell_map1->num_ghosts();
-  bool contains_all_cells = cells1.size() == num_cells1;
+  bool contains_all_cells = (std::size_t)cells1.size() == num_cells1;
   if (contains_all_cells)
   {
     // Check that cells0 is equal to cells1
@@ -1276,7 +1276,8 @@ void interpolate(Function<T, U>& u1, std::span<const std::int32_t> cells1,
       std::vector<std::int32_t> sorted_cells(cells0.begin(), cells0.end());
       std::ranges::sort(sorted_cells);
       auto [unique_end, range_end] = std::ranges::unique(sorted_cells);
-      auto num_unique_cells = std::distance(sorted_cells.begin(), unique_end);
+      std::size_t num_unique_cells
+          = std::distance(sorted_cells.begin(), unique_end);
       contains_all_cells = (num_unique_cells == num_cells1);
     }
   }


garth-wells:
Should there be changes in the core code too? I only see a new test in the PR.

jorgensd:
> Should there be changes in the core code too? I only see a new test in the PR.

Yes, just waiting for the CI to fail before I push the fix.

jorgensd:
> Should there be changes in the core code too? I only see a new test in the PR.

Fix pushed.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1283,10 +1283,11 @@ void interpolate(Function<T, U>& u1, std::span<const std::int32_t> cells1,
       throw std::runtime_error(
           "Interpolation: elements have different value dimensions");
     }
-
-    if (element1 == element0 or *element1 == *element0)
+    if ((element1 == element0 or *element1 == *element0)
```
garth-wells: Order change
```cpp
if ((mesh == u0.function_space()->mesh()) and (element1 == element0 or *element1 == *element0))
```
would be good because the pointer comparisons are cheap, so best that they come first.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1283,10 +1283,11 @@ void interpolate(Function<T, U>& u1, std::span<const std::int32_t> cells1,
...
+        and (mesh == u0.function_space()->mesh()))
```
jorgensd:
```suggestion
    if ((mesh == u0.function_space()->mesh())
        and (element1 == element0 or *element1 == *element0))
````

**Comment by jorgensd** [comment]
> Looking further up in the function, we have
> 
> ```c++
> auto mesh = u1.function_space()->mesh();
> ...
> auto cell_map0 = mesh->topology()->index_map(mesh->topology()->dim());
> ```
> 
> This looks a bit wrong (in a way related to what this PR fixes). We have `mesh` from `u1`, but then use `mesh` to extract `cell_map0`.
> 
> I'd suggest changing
> 
> ```c++
> auto mesh = u1.function_space()->mesh();
> ```
> 
> to
> 
> ```c++
> auto mesh1 = u1.function_space()->mesh();
> ```
> 
> and checking for `foo0` vs `foo1` consistency in the function.
> 
> This would also make the fix in this PR clearer with:
> 
> ```c++
>  if ((mesh1 == u0.function_space()->mesh()) and (element1 == element0 or *element1 == *element0))
> ```

Clarification pushed.
There is still the underlying assumption 
```c++
  auto mesh1 = u1.function_space()->mesh();
  assert(mesh1);
  assert(cells0.size() == cells1.size());

  auto cell_map1 = mesh1->topology()->index_map(mesh1->topology()->dim());
  assert(cell_map1);
  std::size_t num_cells1 = cell_map1->size_local() + cell_map1->num_ghosts();
  if (u1.function_space() == u0.function_space()
      and cells0.size() == num_cells1)
```
which in theory could be wrong, as `cells0 = {0, 1, 3, 3, 2, 2}` and `num_cells1 = 6`; would assume that `cells0 = {0,1,2,3,4,5};`.
It is not a very likely use-case, but an aweful one to debug. One would have to add the check
```c++
auto mismatch = std::mismatch(cells0.begin(),cells0.end(), cells1.begin());
if (u1.function_space() == u0.function_space()
    and cells0.size() == num_cells1 and mismatch.first()==cells0.end())
```

PR #4041, opened 2026-01-15
Fix exterior facet integrals from meshtags

jorgensd:
https://github.com/FEniCS/dolfinx/pull/3900 introduced a bug, in the sense that `ds(tag)` now doesn't discriminate between what is an exterior facet and what is an interior facet.

This PR corrects this behavior by adding a set intersection with the exterior facets while computing the integration domains.

First reported at: https://fenicsproject.discourse.group/t/inconsistent-behaviors-of-the-ds-measure-between-dolfinx-v0-9-0-and-v0-10-0/19352

### DIFF
--- a/python/test/unit/fem/test_element_integrals.py
+++ b/python/test/unit/fem/test_element_integrals.py
@@ -223,9 +223,11 @@ def test_facet_integral(cell_type, dtype):
         elif cell_type == CellType.tetrahedron:
             s = 2**0.5 * 3 ** (1 / 3)  # side length
             v.interpolate(
-                lambda x: (x[0] - s / 2) ** 2
-                + (x[1] - s / 2 / np.sqrt(3)) ** 2
-                + (x[2] - s * np.sqrt(2 / 3) / 4) ** 2
+                lambda x: (
+                    (x[0] - s / 2) ** 2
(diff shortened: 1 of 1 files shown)


jorgensd:
> Hmm, but the "simple" interfaces here work fine:
> 
> ```python
> value = dolfinx.fem.assemble_scalar(dolfinx.fem.form(dolfinx.fem.Constant(domain, 1.0)*ufl.ds))
> print(value) # 4
> 
> value = dolfinx.fem.assemble_scalar(dolfinx.fem.form(dolfinx.fem.Constant(domain, 1.0)*ufl.dx + dolfinx.fem.Constant(domain, 1.0)*ufl.ds))
> print(value) # 5
> ```
> 
> it only breaks down when one passes explicitly marked internal facets.

As I stated, it fails whenever you have a meshtags object as part of ds that tags interior (and exterior) facets. This is because compute integration data at the moment picks the first cell connected to each facet, irregardless of it being interior or exterior. 
this can happen in many cases when one reads in meshes, where an interior surface has the same tag as an exterior surface.

Your example also illustrates the inconsistency of `ds`, as one would expect `ds` to integrate more than `ds(tag)`, but if the subdomain data has all facets in the grid tagged with the value `tag` it would integrate over all interior and exterior facets.

PR #4049, opened 2026-01-24
Fix logic and uses ranges in interpolate

garth-wells:
The logic in interpolate functions was in a bit of a mess after the addition of interpolation on submeshes. This PR fixes the logic, preserving correctness and performance.

It also reduces the need to create 'iota' vectors for interpolation by using C++ ranges.

### DIFF
--- a/cpp/dolfinx/fem/sparsitybuild.h
+++ b/cpp/dolfinx/fem/sparsitybuild.h
@@ -34,7 +34,7 @@ namespace sparsitybuild
 /// @param dofmaps Dofmaps to used in building the sparsity pattern.
 /// @note The sparsity pattern is not finalised.
 template <std::ranges::input_range R0, std::ranges::input_range R1>
-void cells(la::SparsityPattern& pattern, std::pair<R0, R1>&& cells,
+void cells(la::SparsityPattern& pattern, const std::pair<R0, R1>& cells,
            std::array<std::reference_wrapper<const DofMap>, 2> dofmaps)
 {
   assert(cells.first.size() == cells.second.size());


cpp/dolfinx/fem/Function.h
```diff
@@ -178,7 +160,7 @@ class Function
           std::pair<std::vector<value_type>, std::vector<std::size_t>>(
               md::mdspan<const geometry_type,
                          md::extents<std::size_t, 3, md::dynamic_extent>>)>& f,
-      std::span<const std::int32_t> cells)
+      std::ranges::input_range auto&& cells)
```
schnellerhase:
```suggestion
  template <std::ranges::input_range R>
    requires std::same_as<std::remove_const<std::ranges::range_value_t<R>>,
                          std::int32_t>
  void interpolate(
      const std::function<
          std::pair<std::vector<value_type>, std::vector<std::size_t>>(
              md::mdspan<const geometry_type,
                         md::extents<std::size_t, 3, md::dynamic_extent>>)>& f,
      R&& cells)
```

We should constrain to `std::int32_t` valued iterators to avoid silent conversions throughout.

cpp/dolfinx/fem/Function.h
```diff
@@ -229,21 +211,21 @@ class Function
...
     assert(_function_space);
     assert(_function_space->mesh());
     int tdim = _function_space->mesh()->topology()->dim();
     auto cmap = _function_space->mesh()->topology()->index_map(tdim);
     assert(cmap);
-    std::vector<std::int32_t> cells(cmap->size_local() + cmap->num_ghosts(), 0);
-    std::iota(cells.begin(), cells.end(), 0);
-    interpolate(u, cells, cells);
+    interpolate(
+        f, std::ranges::iota_view(0, cmap->size_local() + cmap->num_ghosts()));
```
schnellerhase:
```suggestion
    std::int32_t num_cells = cmap->size_local() + cmap->num_ghosts();
    interpolate(f, std::ranges::iota_view(0, num_cells));
```

cpp/dolfinx/fem/Function.h
```diff
@@ -548,16 +564,14 @@ class Function
     std::vector<geometry_type> det_scratch(2 * gdim * tdim);
 
     // Prepare geometry data in each cell
-    for (std::size_t p = 0; p < cells.size(); ++p)
+    for (auto cell = cells.begin(); cell != cells.end(); ++cell)
```
schnellerhase: How about renaming iterators to `*_it`, so `cell_it` here? The usage`*cell` reads weirdly.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -733,10 +739,12 @@ void point_evaluation(const FiniteElement<U>& element, bool symmetric,
     while (matrix_size * matrix_size < fshape[0])
       ++matrix_size;
 
-    spdlog::info("Loop over cells");
     // Loop over cells
-    for (std::size_t c = 0; c < cells.size(); ++c)
+    spdlog::info("Loop over cells");
```
schnellerhase:
```suggestion
```
Debug output?

cpp/dolfinx/mesh/topologycomputation.cpp
```diff
@@ -206,15 +206,14 @@ get_local_indexing(MPI_Comm comm, const common::IndexMap& vertex_map,
...
-    std::iota(perm.begin(), perm.end(), 0);
-
     auto range_by_index = [&, shape = num_vertices_per_e + 1](auto e)
     {
       auto begin = std::next(entity_to_local_idx.begin(), e * shape);
       return std::ranges::subrange(begin, std::next(begin, shape));
     };
 
+    perm.resize(entity_to_local_idx.size() / (num_vertices_per_e + 1));
+    std::iota(perm.begin(), perm.end(), 0);
```
schnellerhase, line 216:
```suggestion
    std::ranges::iota(perm, 0);
```

garth-wells: `std::ranges::iota` is C++23, which is why we don't use it.

cpp/dolfinx/fem/Function.h
garth-wells: `cell` doesn't bother me.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1242,120 +1252,127 @@ void interpolate(Function<T, U>& u, const Function<T, U>& v,
...
+  }
+  else if (e1->map_type() == e0->map_type())
+  {
+    // Different elements, same basis function map type
+    impl::interpolate_same_map(u1, cells1, u0, cells0);
+  }
+  else
+  {
+    //  Different elements with different maps for basis functions
+    impl::interpolate_nonmatching_maps(u1, cells1, u0, cells0);
```
schnellerhase, line 1321: What happens if `cells` is of type `std::vector`, does this imply a copy here?

schnellerhase: This might be avoidable by restricting to [`std::ranges::view`](https://en.cppreference.com/w/cpp/ranges.html), ensuring we never get an owning iterator passed in the first place.

cpp/dolfinx/fem/Function.h
schnellerhase: Bad duplicate usage of `*`: I meant literal usage of `*cell`, i.e. `if (*cell < 0)`. The name `cell` already implies it is the cell index however dereferencing the cell index is not what we do here - hides it is just an iterator, `cell = *cell_it`.

cpp/dolfinx/fem/interpolate.h
garth-wells: Why would it copy? The functions take a (universal) reference.

schnellerhase: I think, declaring cells as `auto&&` only implies that the argument on call might be an rvalue. Inside the function it is an lvalue. Therefore we would need to move (or forward) here to avoid a copy making the lvalue an rvalue again to be passed back into the function without copy. Avoiding owning containers is what we had inplace before and should remove this problem.

cpp/dolfinx/fem/Function.h
garth-wells: I don't think so, I think we should allow any type that 'works'. I.e. impose no unnecessary constraints.

cpp/dolfinx/fem/interpolate.h
schnellerhase: https://godbolt.org/z/d4Tnnz4r6

garth-wells: > https://godbolt.org/z/d4Tnnz4r6

They're all references.

schnellerhase: > What happens if cells is of type std::vector, does this imply a copy here?

Right, so `cells` template type would be `std::vector&&` for `interpolate` and `std::vector&` for `interpolate_nonmatching_maps` thus not copying but different types. Since this is a dispatch this is a bit surprising, but no performance problem.

Personally I think the `std::forward(cells)` would then make it simpler to understand what is going on, but that's a style choice.

cpp/dolfinx/io/HDF5Interface.h
```diff
@@ -198,9 +198,9 @@ void write_dataset(hid_t file_handle, const std::string& dataset_path,
       if (chunk_size < 1024)
         chunk_size = 1024;
 
-      std::vector<hsize_t> chunk_dims = {chunk_size};
+      std::array<hsize_t, 2> chunk_dims = {chunk_size, hsize_t(0)};
       if (rank == 2)
-        chunk_dims.push_back(dimsf[1]);
+        chunk_dims[1] = dimsf[1];
```
schnellerhase:
```suggestion
      std::array<hsize_t, 2> chunk_dims = {chunk_size, rank == 2 ? dimsf[1] : hsize_t(0)};
```

jhale: This comes from another PR #4043 - would be best to change/suggest elsewhere.

python/dolfinx/wrappers/fem.cpp
```diff
@@ -444,39 +444,90 @@ void declare_objects(nb::module_& m, std::string type)
       .def("collapse", &dolfinx::fem::Function<T, U>::collapse,
            "Collapse sub-function view")
       .def(
-          "interpolate",
+          "interpolate_f",
```
jorgensd, line 447: Why was this renamed?

python/dolfinx/wrappers/fem.cpp
```diff
@@ -444,39 +444,90 @@ void declare_objects(nb::module_& m, std::string type)
...
+            }
+            else
+            {
+              dolfinx::fem::interpolate(
+                  self, std::span(f.data(), f.size()), {f.shape(0), f.shape(1)},
+                  std::span(cells->data(), cells->size()));
+            }
           },
-          nb::arg("f"), nb::arg("cells"), "Interpolate an expression function")
+          nb::arg("f"), nb::arg("cells").none(), "Interpolate an expression.")
```
jorgensd, line 504: This docstring is wrong, no? It should be "Interpolate an array of data (a function evaluated at interpolation points) into a function."

Issue #4047, opened 2026-01-24
[BUG]: test_symmetry_interior_facet_assembly failed with glibcxxassertions


qbisi:
### Summarize the issue

when hardening enable glibcxxassertions on fenics-{basix/dolfinx}, the test_symmetry_interior_facet_assembly run into TRAP when runing petsc_apply_lifting(b, a, bcs=bcs1) for unit_cube mesh with "Regge" FE space.

The SIGTRAP should live in basix::precompute::apply_matrix if debuged with gdb/lldb.

Related pr: https://github.com/NixOS/nixpkgs/pull/483329
Full build log: https://github.com/qbisi/nixpkgs-review-gha/actions/runs/21314500628/job/61355313001
### How to reproduce the bug

enable glibcxxassertions on both fenics-{basix/dolfinx} will reproduce this test failure

### Minimal Example (Python)

```Python
from mpi4py import MPI
import ufl
import numpy as np
from ufl import dS, dx, inner
from dolfinx.mesh import create_unit_cube, locate_entities_boundary, GhostMode
from dolfinx.fem import (
    Function,
    functionspace,
    form,
    locate_dofs_topological,
    dirichletbc,
    bcs_by_block,
    extract_function_spaces,
)

from petsc4py import PETSc
from dolfinx.fem.petsc import apply_lifting as petsc_apply_lifting
from dolfinx.fem.petsc import assemble_matrix as petsc_assemble_matrix
from dolfinx.fem.petsc import assemble_vector as petsc_assemble_vector
from dolfinx.fem.petsc import set_bc as petsc_set_bc


def bc(V):
    facetdim = mesh.topology.dim - 1
    bndry_facets = locate_entities_boundary(
        mesh, facetdim, lambda x: np.isclose(x[0], 0.0)
    )
    bdofsV = locate_dofs_topological(V, facetdim, bndry_facets)
    u_bc = Function(V)
    return dirichletbc(u_bc, bdofsV)


mesh = create_unit_cube(MPI.COMM_WORLD, 5, 5, 5, ghost_mode=GhostMode.shared_facet)

V0 = functionspace(mesh, ("N2E", 1))
V1 = functionspace(mesh, ("Regge", 1))
u0, v0 = ufl.TrialFunction(V0), ufl.TestFunction(V0)
u1, v1 = ufl.TrialFunction(V1), ufl.TestFunction(V1)
n = ufl.FacetNormal(mesh)
a00 = inner(u0, v0) * dx
[...]

PR #4051, opened 2026-01-25
Attempt to trigger SIGTRAP in CI

jhale:
Related to https://github.com/FEniCS/dolfinx/issues/4047 and https://github.com/FEniCS/dolfinx/pull/4043 which did not show up in CI.

### DIFF
--- a/.github/workflows/fenicsx-refs.env
+++ b/.github/workflows/fenicsx-refs.env
@@ -1,5 +1,5 @@
 basix_repository=FEniCS/basix
-basix_ref=jhale/cpp-hardening
+basix_ref=main
 ufl_repository=FEniCS/ufl
 ufl_ref=main
 ffcx_repository=FEniCS/ffcx


.github/workflows/fenicsx-refs.env
```diff
@@ -1,5 +1,5 @@
 basix_repository=FEniCS/basix
-basix_ref=main
+basix_ref=jhale/cpp-hardening
```
schnellerhase: Needs to be reverted before merge.

python/CMakeLists.txt
```diff
@@ -79,8 +80,41 @@ if(NANOBIND_SABI)
...
+check_cxx_compiler_flag(-g HAVE_DEBUG)
+if(HAVE_DEBUG)
+  list(APPEND DOLFINX_PY_CXX_DEVELOPER_FLAGS -g)
+endif()
+
+# Optimisation
+check_cxx_compiler_flag(-O2 HAVE_O2_OPTIMISATION)
+if(HAVE_O2_OPTIMISATION)
+  list(APPEND DOLFINX_PY_CXX_DEVELOPER_FLAGS -O2)
+endif()
```
schnellerhase, line 103: Why the new checks for options existing? It's very verbose. For developer mode it would be senseful to have cmake fail on not finding flags instead of silently changing the passed flags.

jhale: They've always been there.

schnellerhase: This is replacing `set(DOLFINX_PY_CXX_DEVELOPER_FLAGS -Wall;-Werror;-Wextra;-pedantic)`.

jhale: Yes - it is correcting a poor quality copy of what is done in DOLFINx C++.

schnellerhase: I see it aligns with the C++ cmake.

PR #4054, opened 2026-01-28
SuperLU_DIST solver interface

jhale:
Joint work with @chrisrichardson. 

This adds an optional SuperLU_DIST interface to DOLFINx for solving systems constructed using MatrixCSR. The purpose of this is to provide a reasonable MPI-capable LU solver for DOLFINx builds without PETSc on all platforms.

The implementation is pretty light, and does not include SuperLU headers into our headers.

In future PRs:

- `dolfinx.fem.LinearProblem` class for beginners.
- To discuss: `dolfinx.fem.NonlinearProblem` class for beginners.
- Enable user to select solver options

### DIFF
--- a/cpp/dolfinx/la/SuperLUDist.cpp
+++ b/cpp/dolfinx/la/SuperLUDist.cpp
@@ -185,7 +185,7 @@ int SuperLUDistSolver<T>::solve(const la::Vector<T>& b, la::Vector<T>& u) const
     dLUstructInit(m, &LUstruct);
     dSOLVEstruct_t SOLVEstruct;
 
-    spdlog::info("Call pdgssvx");
+    spdlog::info("Call SuperLU_DIST pdgssvx()");
     pdgssvx(&options, _supermatrix.get(), &ScalePermstruct, u.array().data(),
             ldb, nrhs, _gridinfo.get(), &LUstruct, &SOLVEstruct, berr.data(),
             &stat, &info);
@@ -204,7 +204,7 @@ int SuperLUDistSolver<T>::solve(const la::Vector<T>& b, la::Vector<T>& u) const
     sLUstructInit(m, &LUstruct);
     sSOLVEstruct_t SOLVEstruct;
 
-    spdlog::info("Call psgssvx");
+    spdlog::info("Call SuperLU_DIST psgssvx()");
     psgssvx(&options, _supermatrix.get(), &ScalePermstruct, u.array().data(),
             ldb, nrhs, _gridinfo.get(), &LUstruct, &SOLVEstruct, berr.data(),
             &stat, &info);
@@ -223,7 +223,7 @@ int SuperLUDistSolver<T>::solve(const la::Vector<T>& b, la::Vector<T>& u) const
     zLUstructInit(m, &LUstruct);
     zSOLVEstruct_t SOLVEstruct;
 
-    spdlog::info("Call pzgssvx");
+    spdlog::info("Call SuperLU_DIST pzgssvx()");
     pzgssvx(&options, _supermatrix.get(), &ScalePermstruct,
             reinterpret_cast<doublecomplex*>(u.array().data()), ldb, nrhs,
             _gridinfo.get(), &LUstruct, &SOLVEstruct, berr.data(), &stat,


cpp/dolfinx/la/SuperLUDist.cpp
```diff
@@ -0,0 +1,240 @@
...
+#include "superlu.h"
+extern "C"
+{
+#include "superlu_ddefs.h"
+#include "superlu_sdefs.h"
+#include "superlu_zdefs.h"
+}
+#include <dolfinx/la/MatrixCSR.h>
+#include <dolfinx/la/Vector.h>
+#include <iostream>
```
jorgensd: Remove?

python/test/unit/la/test_superlu.py
```diff
@@ -0,0 +1,71 @@
...
+@pytest.mark.skipif(not dolfinx.has_superlu_dist, reason="No SuperLU_dist")
+def test_superlu_solver(dtype):
+    """Manufactured Poisson problem solving u = x[1]**3 with P3 elements.
+    """
+    from dolfinx.la.superlu import superlu_solver
+   
+    mesh = create_unit_square(MPI.COMM_WORLD, 5, 5, dtype=np.float64)
+    V = functionspace(mesh, ("Lagrange", 3))
+    u, v = TrialFunction(V), TestFunction(V)
+    a = inner(grad(u), grad(v)) * dx
```
jorgensd:
```suggestion
```
remove repeat

python/test/unit/la/test_superlu.py
```diff
@@ -0,0 +1,71 @@
...
+    # Source term
+    x = SpatialCoordinate(mesh)
+    u_exact = x[1] ** 3
+    f = -div(grad(u_exact))
+
+    L = inner(f, v) * dx
+    L = form(L, dtype=dtype)
+
+    u_bc = Function(V, dtype=dtype)
+    u_bc.interpolate(lambda x: x[1]**3)
```
jorgensd: Wouldn't it be better to define
```
def u_ex(x):
    return x[1]**3

f = -div(grad(x))
u_bc.interpolate(u_ex)
```
so that if we ever change u_ex it is reflected everywhere at once.

cpp/dolfinx/la/SuperLUDist.cpp
chrisrichardson: used for `cout` for warning message.

python/dolfinx/wrappers/la.cpp
```diff
@@ -289,6 +308,12 @@ void la(nb::module_& m)
           },
           nb::rv_policy::reference_internal);
 
+#if defined(HAS_SUPERLU_DIST)
+  declare_superlu_solver<double>(m, "float64");
+  declare_superlu_solver<float>(m, "float32");
+  declare_superlu_solver<std::complex<double>>(m, "complex128");
+#endif // HAS_SUPERLU_DIST
```
chrisrichardson, line 325: I wonder if we should just put the `#ifdef` inside `declare_superlu_solver()` - it gets called, but does nothing - just looks neater with fewer `#ifdefs`.

python/test/unit/la/test_superlu.py
```diff
@@ -0,0 +1,71 @@
...
+    b.scatter_reverse(InsertMode.add)
+    bc.set(b.array)
+
+    a = form(a, dtype=dtype)
+    A = assemble_matrix(a, bcs=[bc])
+    A.scatter_reverse()
+
+    uh = Function(V, dtype=dtype)
+    solver = superlu_solver(A)
+    solver.solve(b._cpp_object, uh.x._cpp_object)
```
chrisrichardson: Isn't this supposed to work *without* `._cpp_object`?

python/dolfinx/wrappers/la.cpp
jhale: I don't mind the slight overuse of template ifs as long as they aren't nested.

python/dolfinx/la/superlu.py
```diff
@@ -0,0 +1,61 @@
...
+    dtype = A.data.dtype
+    if np.issubdtype(dtype, np.float32):
+        stype = _cpp.la.SuperLUSolver_float32
+    elif np.issubdtype(dtype, np.float64):
+        stype = _cpp.la.SuperLUSolver_float64
+    elif np.issubdtype(dtype, np.complex128):
+        stype = _cpp.la.SuperLUSolver_complex128
+    else:
+        raise NotImplementedError(f"Type {dtype} not supported.")
+    return stype(A._cpp_object)
```
chrisrichardson: I think this should be wrapped in `SuperLUSolver` above?

python/test/unit/la/test_superlu.py
jhale: Indeed, still TODO.

python/test/unit/la/test_superlu.py
```diff
@@ -0,0 +1,79 @@
...
+    # Source term
+    x = SpatialCoordinate(mesh)
+    u_exact = x[1] ** 3
+    f = -div(grad(u_exact))
+
+    L = inner(f, v) * dx
+    L = form(L, dtype=dtype)
+
+    u_bc = Function(V, dtype=dtype)
+    u_bc.interpolate(lambda x: x[1] ** 3)
```
jorgensd:
```suggestion
    def u_ex(x):
        return x[1] ** 3

    x = SpatialCoordinate(mesh)
    f = -div(grad(u_ex(x)))
    L = inner(f, v) * dx
    L = form(L, dtype=dtype)

    u_bc = Function(V, dtype=dtype)
    u_bc.interpolate(u_ex)
```

garth-wells:
We should change the naming - `SuperLU_dist \ne SuperLU`.

python/test/unit/la/test_superlu_dist.py
```diff
@@ -0,0 +1,81 @@
+# Copyright (C) 2026 Jack S. Hale, Chris Richardson
+#
+# This file is part of DOLFINx (https://www.fenicsproject.org)
+#
+# SPDX-License-Identifier:    LGPL-3.0-or-later
+"""Unit tests for MatrixCSR."""
```
chrisrichardson:
```suggestion
"""Unit tests for SuperLU_DIST."""
```

cpp/dolfinx/la/SuperLUDist.cpp
```diff
@@ -0,0 +1,257 @@
...
+  spdlog::info("Finished gridinit");
+  set_operator(*Amat);
+}
+//---------------------------------------------------------------------------------------
+template <typename T>
+void SuperLUDistSolver<T>::set_operator(const la::MatrixCSR<T>& Amat)
+{
+  spdlog::info("Start set_operator");
+  // Global size
+  int m = Amat.index_map(0)->size_global();
```
garth-wells: Throughout DOLFINx we favour fixed-width integers over `int` to be explicit.

cpp/dolfinx/la/SuperLUDist.h
```diff
@@ -0,0 +1,90 @@
...
+
+  // Pointer to struct gridinfo_t
+  std::unique_ptr<SuperLUDistStructs::gridinfo_t, GridInfoDeleter> _gridinfo;
+  // Pointer to SuperMatrix
+  std::unique_ptr<SuperLUDistStructs::SuperMatrix, SuperMatrixDeleter>
+      _supermatrix;
+
+  // Saved matrix operator with rows and cols in
+  // required integer type
+  std::shared_ptr<const la::MatrixCSR<T>> _Amat;
```
garth-wells: C++ has a quirk where underscore followed by a capital is reserved. Use something like `_matA`.

cpp/dolfinx/la/SuperLUDist.h
```diff
@@ -0,0 +1,90 @@
...
+  // cols and rowptr are required in opaque type "int_t" of SuperLU_DIST.
+  std::unique_ptr<SuperLUDistStructs::vec_int_t, VecIntDeleter> cols;
```
garth-wells: Use underscore as a prefix.

cpp/dolfinx/la/SuperLUDist.cpp
```diff
@@ -0,0 +1,257 @@
...
+  else if constexpr (std::is_same_v<T, std::complex<double>>)
+  {
+    zCreate_CompRowLoc_Matrix_dist(
+        _supermatrix.get(), m, n, nnz_loc, m_loc, first_row,
+        reinterpret_cast<doublecomplex*>(Amatdata), cols->vec.data(),
+        rowptr->vec.data(), SLU_NR_loc, SLU_Z, SLU_GE);
+  }
+  else
+  {
+    static_assert(dependent_false_v<T>, "Invalid scalar type");
```
garth-wells, line 125: No `std::complex<float>`?

cpp/dolfinx/la/SuperLUDist.cpp
```diff
@@ -0,0 +1,257 @@
...
+  // Local number of non-zeros
+  int nnz_loc = Amat.row_ptr()[m_loc];
+  cols->vec.resize(nnz_loc);
+  rowptr->vec.resize(m_loc + 1);
+
+  // Copy row_ptr from int64
+  std::copy(Amat.row_ptr().begin(),
+            std::next(Amat.row_ptr().begin(), m_loc + 1), rowptr->vec.begin());
+
+  // Convert local to global indices (and cast to int_t)
```
garth-wells: Is `int_t` defined by SuperLU? If yes, add this to the comment.

cpp/dolfinx/la/SuperLUDist.cpp
```diff
@@ -0,0 +1,257 @@
...
+    static_assert(dependent_false_v<T>, "Invalid scalar type");
+  }
+  spdlog::info("Finished set_operator");
+}
+//---------------------------------------------------------------------------------------
+template <typename T>
+int SuperLUDistSolver<T>::solve(const la::Vector<T>& bvec,
+                                la::Vector<T>& uvec) const
+{
+  int m = _Amat->index_map(0)->size_global();
```
garth-wells: Wrong `int` type. Be explicit.

cpp/dolfinx/la/SuperLUDist.h
garth-wells: Why is `_Amat` stored when it isn't used?

cpp/dolfinx/la/SuperLUDist.h
```diff
@@ -0,0 +1,90 @@
...
+/// Linear solver using SuperLU_DIST
+template <typename T>
+class SuperLUDistSolver
+{
+public:
+  /// @brief SuperLU_DIST solver wrapper
+  /// @param Amat Assembled matrix to invert
+  /// @param verbose Verbosity
+  /// @tparam T Scalar type
+  SuperLUDistSolver(std::shared_ptr<const MatrixCSR<T>> Amat,
```
garth-wells: No need to pass a `shared_ptr` and store the matrix when it's not used outside of the constructor.

cpp/dolfinx/la/SuperLUDist.cpp
```diff
@@ -0,0 +1,257 @@
...
+      rowptr(new SuperLUDistStructs::vec_int_t, VecIntDeleter{}),
+      _verbose(verbose)
+{
+  int size = dolfinx::MPI::size(Amat->comm());
+
+  int nprow = size;
+  int npcol = 1;
+
+  spdlog::info("Start gridinit");
+  superlu_gridinit(Amat->comm(), nprow, npcol, _gridinfo.get());
```
garth-wells: Copy into SuperLU data structure in the initializer list.

cpp/dolfinx/la/SuperLUDist.cpp
```diff
@@ -0,0 +1,257 @@
...
+            std::next(Amat.row_ptr().begin(), m_loc + 1), rowptr->vec.begin());
+
+  // Convert local to global indices (and cast to int_t)
+  std::vector<std::int64_t> global_col_indices(
+      Amat.index_map(1)->global_indices());
+  std::transform(Amat.cols().begin(), std::next(Amat.cols().begin(), nnz_loc),
+                 cols->vec.begin(), [&](std::int64_t local_index) -> int_t
+                 { return global_col_indices[local_index]; });
+
+  auto Amatdata = const_cast<T*>(Amat.values().data());
```
garth-wells: Don't use `auto` here - just obfuscates.

cpp/dolfinx/la/SuperLUDist.cpp
```diff
@@ -0,0 +1,257 @@
...
+  int m = _Amat->index_map(0)->size_global();
+  int m_loc = _Amat->num_owned_rows();
+
+  // RHS
+  int ldb = m_loc;
+  int nrhs = 1;
+
+  superlu_dist_options_t options;
+  set_default_options_dist(&options);
+  options.DiagInv = YES;
```
garth-wells, line 165: Caller should be able to control options.

cpp/dolfinx/la/SuperLUDist.cpp
```diff
@@ -0,0 +1,257 @@
...
+  if (!_verbose)
+    options.PrintStat = NO;
+
+  int info = 0;
+  SuperLUStat_t stat;
+  PStatInit(&stat);
+
+  // Copy b to u (SuperLU_DIST reads b from u and then overwrites u with
+  // solution)
+  std::copy(bvec.array().begin(), std::next(bvec.array().begin(), m_loc),
```
garth-wells: `std::copy_n` would be simpler.

cpp/dolfinx/la/SuperLUDist.cpp
```diff
@@ -0,0 +1,257 @@
...
+            uvec.array().begin());
+
+  if constexpr (std::is_same_v<T, double>)
+  {
+    std::vector<T> berr(nrhs);
```
garth-wells: `nrhs` is always so a simpler type can be used. And hoist outside of if statement.

cpp/dolfinx/la/SuperLUDist.cpp
```diff
@@ -0,0 +1,257 @@
...
+    spdlog::info("Start solve [float64]");
+    dScalePermstruct_t ScalePermstruct;
+    dLUstruct_t LUstruct;
+    dScalePermstructInit(m, m, &ScalePermstruct);
+    dLUstructInit(m, &LUstruct);
+    dSOLVEstruct_t SOLVEstruct;
+
+    spdlog::info("Call pdgssvx");
```
garth-wells: Use more descriptive log message.

garth-wells (changes requested):
Code can be simplified and improved in a number of places. See comments.

cpp/dolfinx/la/SuperLUDist.cpp
chrisrichardson: It wasn't available a couple of years ago when I started this, but things may have changed. Will check.

cpp/dolfinx/la/SuperLUDist.h
chrisrichardson: It shares data (matrix value entries) with SuperLU_DIST. The rows and columns can't be shared because the integer format varies, and it needs global indexing.

cpp/dolfinx/la/SuperLUDist.h
chrisrichardson: see above

cpp/dolfinx/la/SuperLUDist.cpp
chrisrichardson: Agree, but I would like to do that in a future PR.

cpp/dolfinx/la/SuperLUDist.cpp
chrisrichardson: Can't hoist because it is not always `vector<T>` - see the complex case.

cpp/dolfinx/la/SuperLUDist.cpp
jhale: It would be best to split this into two steps; `std::int64_t` out of the DOLFINx functions, and then a proper limits check before casting to `int`, or throwing an exception.

cpp/dolfinx/la/SuperLUDist.cpp
jhale: This can be hoisted but only with some type-trickery - not sure it's worth the additional complexity.

cpp/dolfinx/la/SuperLUDist.cpp
jhale: Boost has `numeric_cast` in `boost/cast.hpp`  or I can implement the specific cast just for this.

cpp/dolfinx/la/SuperLUDist.h
```diff
@@ -0,0 +1,91 @@
...
```
garth-wells: `MatrixCSR` has more than one template argument. Should probably expand to include them all, or template over the matrix. 

Related to this, any small tweaks that would allow us to use the SuperLU_DIST GPU support?

cpp/dolfinx/la/SuperLUDist.cpp
schnellerhase: We already have `dolfinx::scalar_value_t<T>` in place - `std::vector<dolfinx::scalar_value_t<T>>` should do what we need here.

cpp/dolfinx/la/SuperLUDist.h
jhale: What SuperLU_DIST actually accepts is not very general. We could probably template over MatrixCSR and then use concepts to constrain.

On the second point, on my reading you add `+cuda` or `+rocm` on the Spack build for `superlu-dist` and then (partial) GPU offload is enabled by default. I will try this another time.

https://dl.acm.org/doi/epdf/10.1145/3577197

cpp/dolfinx/la/SuperLUDist.cpp
```diff
@@ -0,0 +1,262 @@
...
+#include "SuperLUDist.h"
+extern "C"
+{
+#include "superlu_ddefs.h"
+#include "superlu_sdefs.h"
+#include "superlu_zdefs.h"
+}
+#include <dolfinx/la/MatrixCSR.h>
+#include <dolfinx/la/Vector.h>
+#include <iostream>
```
jhale:
```suggestion
```

cpp/dolfinx/la/SuperLUDist.cpp
```diff
@@ -0,0 +1,261 @@
...
+            int npcol = 1;
+
+            std::unique_ptr<SuperLUDistStructs::gridinfo_t, GridInfoDeleter> p(
+                new SuperLUDistStructs::gridinfo_t, GridInfoDeleter{});
+
+            superlu_gridinit(comm, nprow, npcol, p.get());
+            return p;
+          }()),
+      _supermatrix(new SuperLUDistStructs::SuperMatrix, SuperMatrixDeleter{}),
+      cols(new SuperLUDistStructs::vec_int_t, VecIntDeleter{}),
```
garth-wells: Do the vectors really need custom deleters?

cpp/dolfinx/la/SuperLUDist.h
```diff
@@ -9,12 +9,12 @@
...
 #include <dolfinx/la/MatrixCSR.h>
 #include <dolfinx/la/Vector.h>
+#include <memory>
 
 namespace dolfinx::la
 {
 
-// Trick for forward declaration of anonymous structs
-// Avoids including SuperLU_DIST headers in DOLFINx headers
+// Delcare structs to avoid exposing SuperLU_DIST headers in DOLFINx.
```
chrisrichardson: Spelling

cpp/dolfinx/la/SuperLUDist.cpp
jhale: I had a look earlier - the recent TOMS paper claims support but there are no headers on main.

cpp/dolfinx/la/SuperLUDist.cpp
jhale: Removed and works locally. Having some issues with headers in CI but will fix.

python/dolfinx/wrappers/la.cpp
```diff
@@ -24,6 +25,16 @@
 #include <nanobind/stl/vector.h>
 #include <span>
 
+#if defined(HAS_SUPERLU_DIST)
+#include <superlu_defs.h>
```
jhale, line 29: This will require SuperLU_DIST to be found in `python/CMakeLists.txt`. This does seem correct, given the current design.

Note a lot of the CIs are passing due to SuperLU_DIST's headers being in PETSc's paths - this is not the case for our Ubuntu PETSc-less install, where this issue shows up.

PR #4058, opened 2026-01-30
Use overlay on top of Spack upstream

jhale:
Tests have begun to fail using out-of-date Spack fork.

On a side note, PRs don't seem to be going through at spack-packages.


.github/workflows/ci-spack.yml
```diff
@@ -36,23 +36,33 @@ jobs:
...
-          repository: fenics/spack-packages
-          ref: fenics-testing
+          repository: spack/spack-packages
+          ref: develop
           path: spack-packages
+      - name: Get FEniCS Spack packages
+        uses: actions/checkout@v6
+        with:
+          repository: fenics/spack-fenics
+          ref: main
```
schnellerhase, line 56: Should we add those to `fenicsx-refs.env`?

PR #4060, opened 2026-02-01
Experimental threaded mesh construction

garth-wells:
High experimental and under development.

When building a mesh on the CPU and running assembly/solve on a GPU, it is natural to use one MPI rank per GPU. Typically, for each GPU on a node there will be many more CPU cores. Therefore, it is helpful to use threading to speed up steps that are best performed on the CPU.

The change starts to introduce threading in mesh construction.

### DIFF
--- a/python/dolfinx/wrappers/common.cpp
+++ b/python/dolfinx/wrappers/common.cpp
@@ -275,7 +275,8 @@ void common(nb::module_& m)
   m.def("timing", &dolfinx::timing);
   m.def("timings", &dolfinx::timings);
 
-  m.def("hardware_concurrency", &std::thread::hardware_concurrency);
+  m.def("hardware_concurrency",
+        []() { return std::max<int>(1, std::thread::hardware_concurrency()); });
 
   m.def(
       "list_timings",
--- a/python/dolfinx/wrappers/mesh.cpp
+++ b/python/dolfinx/wrappers/mesh.cpp
@@ -658,7 +658,7 @@ void mesh(nb::module_& m)
           nb::arg("cell_type"), nb::arg("vertex_map"), nb::arg("cell_map"),
           nb::arg("cells"), nb::arg("original_index").none())
       .def("create_entities", &dolfinx::mesh::Topology::create_entities,
-           nb::arg("dim"), nb::arg("num_threads"))
+           nb::arg("dim"), nb::arg("num_threads") = 1)
       .def("create_entity_permutations",
            &dolfinx::mesh::Topology::create_entity_permutations)
       .def("create_connectivity", &dolfinx::mesh::Topology::create_connectivity,


cpp/dolfinx/mesh/topologycomputation.cpp
```diff
@@ -478,115 +580,74 @@ compute_entities_by_key_matching(
...
 
-    const std::size_t num_cells = cells->num_nodes();
+    common::Timer t_thread("Threaded part");
+
+    const graph::AdjacencyList<std::int32_t>& cells
+        = std::get<1>(cell_lists[k]);
     int num_entities_per_cell = cell_type_entities[k].size();
-    for (std::size_t c = 0; c < num_cells; ++c)
+    // int num_threads = 2;
+    // int num_threads = std::thread::hardware_concurrency();
```
chrisrichardson: Why not use this (or default to it)?

cpp/dolfinx/mesh/topologycomputation.cpp
```diff
@@ -478,115 +580,74 @@ compute_entities_by_key_matching(
...
-              = entity_vertices[perm[j]];
-        }
-      }
+      auto [c0, c1]
+          = dolfinx::MPI::local_range(i, cells.num_nodes(), num_threads);
+      std::span<std::int32_t> _entity_list(
+          entity_list.data() + cell_type_offsets[k] * num_vertices_per_entity
+              + c0 * num_vertices_per_entity * num_entities_per_cell,
+          (c1 - c0) * num_vertices_per_entity * num_entities_per_cell);
+      threads[i] = std::jthread(
```
chrisrichardson, line 631: How do we know when all threads have completed? Where is the synchronisation point?

garth-wells: When `std::jthread` gets destroyed. This sets it apart form `std::thread`.

cpp/dolfinx/mesh/topologycomputation.cpp
garth-wells: Number of threads is now passed through function args. Can add a 'temporary' demo for testing.

python/dolfinx/wrappers/common.cpp
```diff
@@ -274,6 +275,8 @@ void common(nb::module_& m)
   m.def("timing", &dolfinx::timing);
   m.def("timings", &dolfinx::timings);
 
+  m.def("hardware_concurrency", &std::thread::hardware_concurrency);
```
chrisrichardson: I think we should modify this behaviour. The default, if undefined (according to https://en.cppreference.com/w/cpp/thread/thread/hardware_concurrency.html) is zero.
I think we should change it to be 1.

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -654,7 +658,7 @@ void mesh(nb::module_& m)
           nb::arg("cell_type"), nb::arg("vertex_map"), nb::arg("cell_map"),
           nb::arg("cells"), nb::arg("original_index").none())
       .def("create_entities", &dolfinx::mesh::Topology::create_entities,
-           nb::arg("dim"))
+           nb::arg("dim"), nb::arg("num_threads"))
```
chrisrichardson:
```suggestion
           nb::arg("dim"), nb::arg("num_threads") = 1)
```

cpp/dolfinx/mesh/topologycomputation.cpp
```diff
@@ -478,115 +584,69 @@ compute_entities_by_key_matching(
...
-
-        for (std::size_t j = 0; j < ev.size(); ++j)
-        {
-          entity_list[(cell_type_offsets[k] + idx) * num_vertices_per_entity
-                      + j]
-              = entity_vertices[perm[j]];
-        }
-      }
+      auto [c0, c1]
+          = dolfinx::MPI::local_range(i, cells.num_nodes(), num_threads);
```
chrisrichardson, line 624: We should move from `MPI::` to `common::`

cpp/dolfinx/mesh/topologycomputation.cpp
```diff
@@ -798,9 +863,10 @@ mesh::compute_entities(const Topology& topology, int dim, CellType entity_type)
   // Lists of all cells by cell type
   std::vector<CellType> cell_types = topology.entity_types(tdim);
   std::vector<std::tuple<
-      mesh::CellType, std::shared_ptr<const graph::AdjacencyList<std::int32_t>>,
-      std::shared_ptr<const common::IndexMap>>>
-      cell_lists(cell_types.size());
+      mesh::CellType,
+      std::reference_wrapper<const graph::AdjacencyList<std::int32_t>>,
```
jorgensd, line 867: Is the change to reference wrapper done for performance reasons?

michalhabera:
Any reason for merging "High experimental and under development" to main?

PR #4061, opened 2026-02-02
Split SuperLU solver and matrix

jhale:
This makes it possible to use the memory management of the SuperMatrix object separately from the solver in C++.

No changes in Python - doesn't seem worth the boiler plate at the minute.


cpp/dolfinx/la/superlu_dist.cpp
```diff
@@ -0,0 +1,289 @@
...
+          }()),
+      _verbose(verbose)
+{
+}
+
+//----------------------------------------------------------------------------
+template <typename T>
+int SuperLUDistSolver<T>::solve(const la::Vector<T>& b, la::Vector<T>& u) const
+{
+  std::int64_t m = _A_superlu_mat->Amat().index_map(0)->size_global();
```
jhale: Breaks encapsulation, get from native SuperMatrix?

PR #4074, opened 2026-02-06
Allow SuperLU_DIST to solve with repeated right-hand sides

jhale:
It turns out that SuperLU_DIST holds quite a bit of 'state' through repeated calls to its solve driver routines to allow for fast repeated solves with pre-computed factors. The previous design didn't work for repeated solves.

This PR:

1. Copies values from MatrixCSR into SuperLUDistMatrix, instead of sharing them. SuperLU_DIST modifies the values of the SuperLUDistMatrix in place during the solve, so the previous design led to MatrixCSR being modified.
2. Properly persists LU factors and other solve information across solves, allowing for repeated solves with different right-hand sides.
3. Wraps SuperLUDistMatrix into Python - this is necessary for advanced usage where new systems are solved with previously computed factors/permutations.

### DIFF
--- a/python/test/unit/la/test_superlu_dist.py
+++ b/python/test/unit/la/test_superlu_dist.py
@@ -113,23 +113,14 @@ def test_superlu_solver(dtype):
     a_2 = (inner(grad(u), grad(v)) + gamma * inner(u, v)) * dx
     a_2 = form(a_2, dtype=dtype)
 
-    # Exact solution
-    def u_ex_2(x):
-        return x[1] ** 2
-
-    f_2 = -div(grad(u_ex_2(x))) + gamma * u_ex_2(x)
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/la/superlu_dist.h
```diff
@@ -23,8 +24,64 @@ class SuperLUDistStructs
...
+  struct zLUstruct_t;
+
+  struct sSOLVEstruct_t;
+  struct dSOLVEstruct_t;
+  struct zSOLVEstruct_t;
+};
+
+// SuperLU_DIST has structs that are 'typed' with prefixes d, s, z. This allows
+// the solver class to select the typed set based on T.
+namespace detail
```
jhale: detail or impl here?

garth-wells: `impl`. `detail` is vague.

PR #4081, opened 2026-02-10
Fix dofmap builder performance fixes

garth-wells:
Call to `mesh::cell_entity_type` inside hot loop is expensive.

### DIFF
--- a/cpp/dolfinx/fem/dofmapbuilder.cpp
+++ b/cpp/dolfinx/fem/dofmapbuilder.cpp
@@ -262,13 +262,13 @@ build_basic_dofmaps(
     dofs[i].array.resize(num_cells * dofmap_width);
     spdlog::info("Cell type: {} dofmap: {}x{}", i, num_cells, dofmap_width);
 
-    std::vector<std::vector<mesh::CellType>> cell_entity_types;
+    std::vector<std::vector<mesh::CellType>> cell_entity_types(D + 1);
     for (std::size_t d = 0; d < D + 1; ++d)
     {
-      std::vector<mesh::CellType> tmp;
-      for (int e = 0; e < mesh::cell_num_entities(cell_type, d); ++e)
-        tmp.push_back(mesh::cell_entity_type(cell_type, d, e));
-      cell_entity_types.push_back(std::move(tmp));
+      int entities_d = mesh::cell_num_entities(cell_type, d);
+      cell_entity_types[d].reserve(entities_d);
+      for (int e = 0; e < entities_d; ++e)
+        cell_entity_types[d].push_back(mesh::cell_entity_type(cell_type, d, e));
     }
 
     std::int32_t dofmap_offset = 0;


cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -263,6 +262,15 @@ build_basic_dofmaps(
...
 
+    std::vector<std::vector<mesh::CellType>> cell_entity_types;
+    for (std::size_t d = 0; d < D + 1; ++d)
+    {
+      std::vector<mesh::CellType> tmp;
+      for (int e = 0; e < mesh::cell_num_entities(cell_type, d); ++e)
+        tmp.push_back(mesh::cell_entity_type(cell_type, d, e));
+      cell_entity_types.push_back(std::move(tmp));
+    }
+
```
schnellerhase:
```suggestion
    std::vector<std::vector<mesh::CellType>> cell_entity_types(D + 1);
    for (std::size_t d = 0; d < D + 1; ++d)
    {
      int entities_d = mesh::cell_num_entities(cell_type, d);
      cell_entity_types[d].reserve(entities_d);
      for (int e = 0; e < entities_d; ++e)
        cell_entity_types[d].push_back(mesh::cell_entity_type(cell_type, d, e));
    }

```

garth-wells: More lines, nor performance critical section.

cpp/dolfinx/mesh/cell_types.cpp
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2006-2019 Anders Logg and Garth N. Wells
+// Copyright (C) 2006-2016 Anders Logg and Garth N. Wells
```
jorgensd:
```suggestion
// Copyright (C) 2006-2026 Anders Logg and Garth N. Wells
```

cpp/dolfinx/mesh/cell_types.h
```diff
@@ -41,42 +41,110 @@ std::string to_string(CellType type);
...
 
-/// Return facet type of cell
-/// For simplex and hypercube cell types, this is independent of the facet
-/// index, but for prism and pyramid, it can be triangle or quadrilateral.
-/// @param[in] type The cell type
-/// @param[in] index The facet index
-/// @return The type of facet for this cell at this index
-CellType cell_facet_type(CellType type, int index);
+/// Return type of cell for entity of dimension d at given entity index.
+inline CellType cell_entity_type(CellType type, int d, int index)
```
jorgensd, line 111: We should start using `inline` more often on these kind of things.

python/dolfinx/wrappers/geometry.cpp
```diff
@@ -214,13 +214,16 @@ void declare_bbtree(nb::module_& m, const std::string& type)
...
         std::size_t p_s0 = points.ndim() == 1 ? 1 : points.shape(0);
         std::span<const T> _p(points.data(), 3 * p_s0);
-        std::optional<std::span<const std::int32_t>> _cells
-            = cells.has_value()
-                  ? std::span<const std::int32_t>(cells.value().data(),
-                                                  cells.value().size())
-                  : std::optional<std::span<const std::int32_t>>(std::nullopt);
-        return dolfinx::geometry::determine_point_ownership<T>(mesh, _p,
-                                                               padding, _cells);
+        if (cells.has_value())
```
jorgensd, line 217: This is solely done for the sake of readability, right?
I like the new readable version:)

cpp/dolfinx/mesh/cell_types.h
garth-wells: Inline is required here to avoid duplicate symbols.

python/dolfinx/wrappers/geometry.cpp
garth-wells: Compiler on one of the CI was failing on the old code.

PR #4083, opened 2026-02-11
Threaded mesh topology construction improvements

garth-wells:
`dolfinx::sort_perm` didn't respect `BITS` template parameter. PR fixes this, which gives a significant performance boost for some cases.

### DIFF
--- a/cpp/dolfinx/mesh/topologycomputation.cpp
+++ b/cpp/dolfinx/mesh/topologycomputation.cpp
@@ -648,7 +648,6 @@ compute_entities_by_key_matching(
   }
 
   // Start numbering entities
-  common::Timer timer0("Create entities 0");
   std::vector<std::int32_t> entity_index(cell_type_offsets.back());
   std::int32_t entity_count = 0;
   {
@@ -682,8 +681,6 @@ compute_entities_by_key_matching(
                                                         num_vertices_per_entity)
               : sort_threaded(entity_list_sorted, num_vertices_per_entity,
                               num_threads);
-    timer.stop();
-    timer.flush();
 
     std::vector<std::int32_t> entity(num_vertices_per_entity);
     std::vector<std::int32_t> entity0(num_vertices_per_entity);
@@ -713,15 +710,12 @@ compute_entities_by_key_matching(
       ++entity_count;
     }
   }
-  timer0.stop();
-  timer0.flush();
 
   //---------
   // Set ghost status array values
   // 0 = entities that are only in ghost cells (i.e. definitely not
   // owned) 1 = entities with local ownership or ownership that needs
   // deciding
-  common::Timer timer1("Create entities 1");
   std::vector<std::int8_t> ghost_status(entity_count, 1);
   for (std::size_t k = 0; k < cell_lists.size(); ++k)
   {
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/mesh/topologycomputation.cpp
```diff
@@ -51,14 +53,14 @@ namespace
...
-                       std::span<std::int32_t> entity_list_sorted,
-                       std::int32_t c0, std::int32_t num_cells,
-                       const graph::AdjacencyList<std::int32_t>& cells,
-                       const graph::AdjacencyList<std::int32_t>& e_vertices,
-                       mesh::CellType entity_type,
-                       const std::vector<std::int32_t>& cell_type_entities,
-                       const common::IndexMap& vertex_index_map)
+auto build_entity_list
+    = [](std::span<std::int32_t> entity_list,
+         std::span<std::int32_t> entity_list_sorted, auto&& cell_idx,
```
schnellerhase: Restrict to `std::int32_t` ranges.

cpp/dolfinx/mesh/topologycomputation.cpp
```diff
@@ -618,33 +622,61 @@ compute_entities_by_key_matching(
     const graph::AdjacencyList<std::int32_t>& cells
         = std::get<1>(cell_lists[k]);
     int num_entities_per_cell = cell_type_entities[k].size();
-    std::vector<std::jthread> threads(num_threads);
-    for (int i = 0; i < num_threads; ++i)
+    if (num_threads > 1)
```
schnellerhase: Should allow for compile time deduction for no threading configuration. Resolvable with https://github.com/FEniCS/dolfinx/pull/3716.

cpp/dolfinx/mesh/topologycomputation.cpp
```diff
@@ -618,33 +622,61 @@ compute_entities_by_key_matching(
...
+    common::Timer timer("Compute entities by key matching: number entities");
+
     // Sort the list and label uniquely
-    const std::vector<std::int32_t> sort_order
-        = dolfinx::sort_by_perm<std::int32_t>(entity_list_sorted,
-                                              num_vertices_per_entity);
+    std::vector<std::int32_t> sort_order(
+        entity_list_sorted.size() / num_vertices_per_entity, 0);
+    std::iota(sort_order.begin(), sort_order.end(), 0);
+    boost::sort::parallel_stable_sort(
```
schnellerhase: Adds new library dependency to `Boost.Sort`.`sort_by_perm` in particular avoids `lexicographical_compare` - performance impact?

cpp/dolfinx/mesh/topologycomputation.cpp
garth-wells: No need to overcomplicate - it will have no measurable runtime effect.

schnellerhase: Not about performance. It ensures optionality of the functionality. When passing `num_threads` around explicitly this needs to be deduced from the associated type.

garth-wells: I don't follow. The if/else switch isn't inside a hot loop, so what's the issue?

schnellerhase: It would be good to maintain optionality of the threading support. We can (should) not change function signatures with an altered configuration, so this would need to be indicated by the type of the argument.

cpp/dolfinx/mesh/topologycomputation.cpp
garth-wells: Significantly faster (several factors)

cpp/dolfinx/mesh/topologycomputation.cpp
garth-wells: It is optional now - '0' is no thread launch.

PR #4084, opened 2026-02-12
Move nanobind declarations to header files

michalhabera:
This PR allows users to declare templated binding for custom data types. For example,
```cpp
// my_custom_binding.cpp

#include <dolfinx_wrappers/fem.h>

//...

template <>
struct dolfinx::is_custom_scalar<my_scalar_t> : std::true_type {};

template <>
struct nanobind::ndarray_traits<my_scalar_t> {
  static constexpr bool is_complex = false;
  static constexpr bool is_float = true;
  static constexpr bool is_bool = false;
  static constexpr bool is_int = false;
  static constexpr bool is_signed = true;
};

template <>
struct dolfinx_wrappers::numpy_dtype_traits<my_scalar_t> {
  static constexpr char value = 'd';
};

NB_MODULE(_custom_ext, m) {
  // ...
  dolfinx_wrappers::declare_objects<my_scalar_t>(m, "my_scalar");
  dolfinx_wrappers::declare_form<my_scalar_t>(m, "my_scalar");
}
```

Also, naming is unified to follow `declare_foo` for all bindings.

### DIFF
--- a/cpp/dolfinx/fem/assemble_vector_impl.h
+++ b/cpp/dolfinx/fem/assemble_vector_impl.h
@@ -566,7 +566,7 @@ void _lift_bc_interior_facets(
     if (cells0[1] >= 0)
     {
       std::span sub_Ae0(Ae.data() + bs0 * num_dofs0 * num_cols,
-                        bs0 * num_dofs1 * num_cols);
+                        bs0 * num_dofs0 * num_cols);
       P0(sub_Ae0, cell_info0, cells0[1], num_cols);
     }
     if (cells1[0] >= 0)
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -84,9 +84,6 @@ markers = [
       "adios2: tests that require Adios2 (deselect with '-m \"not adios2\"').",
       "xfail_win32_complex: expected failures for complex numbers in Win32.",
 ]
-addopts = [
-  "-k", "not test_symmetry_interior_facet_assembly[mesh1]"
-]
 
 [tool.mypy]
 # Suggested at https://blog.wolt.com/engineering/2021/09/30/professional-grade-mypy-configuration/


python/dolfinx/__init__.py
```diff
@@ -67,6 +67,27 @@ def get_include(user=False):
...
+
+    try:
+        import mpi4py
+
+        dirs.append(mpi4py.get_include())
+    except ImportError:
+        pass
+
+    try:
+        import petsc4py
```
jhale: This isn't really safe - we allow petsc4py to be installed and DOLFINx to not be compiled against it. Use `has_petsc4py` instead.

python/dolfinx/__init__.py
```diff
@@ -67,6 +67,27 @@ def get_include(user=False):
         return os.path.join(os.path.dirname(d), "src")
 
 
+def get_include_dirs() -> list[str]:
+    """Return all include dirs needed to compile against dolfinx wrapper headers."""
+    dirs = [get_include()]
+
+    try:
+        import mpi4py
```
jhale: This is mandatory dependency, no need to be so polite with mpi4py.

python/dolfinx/__init__.py
```diff
@@ -67,6 +67,27 @@ def get_include(user=False):
         return os.path.join(os.path.dirname(d), "src")
 
 
+def get_include_dirs() -> list[str]:
```
jhale: Where is this used?

python/dolfinx/wrappers/common.cpp
```diff
@@ -16,6 +17,7 @@
 #include <dolfinx/common/log.h>
 #include <dolfinx/common/timing.h>
 #include <dolfinx/common/utils.h>
+#include <iostream>
```
jhale: Needed?

python/dolfinx/__init__.py
michalhabera: I've removed that for now. But it is good for libraries using dolfinx_wrappers to know what include dirs to add. In a custom nanobind project I want to get this systematically, not hardcode include dirs for mpi4py and petsc4py.

python/dolfinx/wrappers/dolfinx_wrappers/fem.h
```diff
@@ -0,0 +1,1244 @@
...
+            nb::arg("x"), nb::arg("cell_permutations"), nb::arg("dim"))
+        .def_prop_ro("needs_dof_transformations",
+                     &dolfinx::fem::FiniteElement<T>::needs_dof_transformations)
+        .def_prop_ro("signature", &dolfinx::fem::FiniteElement<T>::signature);
+  }
+}
+
+// Declare DirichletBC objects for type T
+template <typename T>
+void declare_objects(nb::module_& m, std::string type)
```
schnellerhase, line 332: Can we split this up? Into DBC, Function etc?

python/dolfinx/wrappers/dolfinx_wrappers/fem.h
```diff
@@ -0,0 +1,1244 @@
...
+
+  m.def("create_sparsity_pattern", &dolfinx::fem::create_sparsity_pattern<T, U>,
+        nb::arg("a"), "Create a sparsity pattern.");
+
+  m.def("build_sparsity_pattern", &dolfinx::fem::build_sparsity_pattern<T, U>,
+        nb::arg("pattern"), nb::arg("a"), "Build a sparsity pattern.");
+}
+
+template <typename T>
+void declare_cmap(nb::module_& m, const std::string& type)
```
schnellerhase:
```suggestion
void declare_coordinate_element(nb::module_& m, const std::string& type)
```
consistent naming scheme

python/dolfinx/wrappers/dolfinx_wrappers/fem.h
```diff
@@ -0,0 +1,1244 @@
...
+              self.pull_back_nonaffine(X, _x, g);
+
+            return dolfinx_wrappers::as_nbarray(std::move(Xb),
+                                                {num_points, tdim});
+          },
+          nb::arg("x"), nb::arg("cell_geometry"));
+}
+
+template <typename T>
+void declare_real_functions(nb::module_& m)
```
schnellerhase, line 1072: Naming

python/dolfinx/wrappers/dolfinx_wrappers/numpy_dtype.h
```diff
@@ -7,38 +7,72 @@
...
 
 namespace dolfinx_wrappers
 {
 
-/// @brief Return NumPy dtype char type for a C++ type
-/// @tparam T C++ type
-/// @return NumPy dtype
+// Primary template: unimplemented (will cause compile error if not specialized)
+template <typename T>
+struct numpy_dtype_traits;
```
schnellerhase: Nice, remove the `_traits` from the name? And introduce short notation `numpy_dtype_v` for simplified calling?

python/dolfinx/wrappers/dolfinx_wrappers/numpy_dtype.h
```diff
@@ -7,38 +7,72 @@
...
+  static constexpr char value = 'l';
+};
+template <>
+struct numpy_dtype_traits<std::uint64_t>
+{
+  static constexpr char value = 'L';
+};
+
 template <typename T>
 constexpr char numpy_dtype()
```
schnellerhase: Remove.

python/dolfinx/wrappers/dolfinx_wrappers/assemble.h
```diff
@@ -0,0 +1,576 @@
...
+#include <span>
+#include <string>
+#include <utility>
+
+namespace dolfinx_wrappers
+{
+
+namespace md = MDSPAN_IMPL_STANDARD_NAMESPACE;
+
+template <typename U>
```
schnellerhase, line 49: We should restrict this to admissible types, in particular as template types may have different meanings in some other cases, i.e. use `dolfinx::scalar` and `std::floating_point` where possible.

python/dolfinx/wrappers/dolfinx_wrappers/fem.h
michalhabera: I only intended to move the code and do very basic renaming here. Should we rather do clean-up and improvements in a separate PR?

python/dolfinx/wrappers/dolfinx_wrappers/assemble.h
michalhabera: I you agree I'd suggest to postpone these improvements to another PR.

jhale:
No looks good to me - you can do small improvements on other PRs.

Issue #4052, opened 2026-01-26
[BUG]: Fix errors from `-Wsign-compare` in Python bindings


jhale:
### Summarize the issue

Further use of strict compiler flags across the CI led to a number of `-Wsign-compare` errors being triggered when building the Python bindings.

These are currently being ignored:

https://github.com/FEniCS/dolfinx/pull/4051/changes#diff-005a07ca82fb46b215f58a3a8f4d6d6cc6ca088ade44e8f15cc2e6a51d1da6f4R87

and should be fixed.

### How to reproduce the bug

Remove `-Wno-sign-compare` from `python/CMakeLists.txt` and build DOLFINx Python bindings in Developer mode

### Minimal Example (Python)

```Python

```

### Output (Python)

```bash

```

### Version

main branch

schnellerhase:
Blocked by https://github.com/FEniCS/dolfinx/issues/4088.

PR #4085, opened 2026-02-12
Remove no-sign-compare

schnellerhase:
There is something bigger still wrong - this should scream on every platform when configured with `Developer`.

Fixes https://github.com/FEniCS/dolfinx/issues/4052


schnellerhase:
Tracked at https://github.com/FEniCS/dolfinx/issues/4088.

PR #4087, opened 2026-02-14
Fix Spack in install in CI

garth-wells:
Add `RPATH` to Basix install.

Removes `pyamg` from pyproject file. `pyamg` is problematic because it over-constrains dependency versions.

### DIFF
--- a/.github/workflows/ci-spack.yml
+++ b/.github/workflows/ci-spack.yml
@@ -50,18 +50,10 @@ jobs:
           ref: develop
           path: spack-packages
 
-      # - name: Get FEniCS Spack packages
-      #   uses: actions/checkout@v6
-      #   with:
-      #     repository: fenics/spack-fenics
-      #     ref: main
-      #     path: spack-fenics
-
       - name: Add Spack packages repo
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-spack.yml
```diff
@@ -109,8 +103,9 @@ jobs:
           . ./spack-src/share/spack/setup-env.sh
           spack env activate py
           pip install git+https://github.com/${{ env.ufl_repository }}.git@${{ env.ufl_ref }}
-          pip install -Ccmake.build-type="Developer" git+https://github.com/${{ env.basix_repository }}.git@${{ env.basix_ref }}
```
jhale, line 112: @garth-wells Was there a reason to remove this? When running DOLFINx tests this has picked up issues in the past as it enables a lot of C++ standard library asserts.

python/pyproject.toml
```diff
@@ -34,7 +34,6 @@ demo = [
   "numba",
   "matplotlib",
   "scipy",
-  "pyamg",
```
jhale, line 37: Helpful to keep this one as it documents what our demos are using.

PR #4091, opened 2026-02-15
Playing with Windows again

jhale:


jhale:
Windows builds and runs fine with `mpiexec -n 1` and seems to build and run fine on conga-forge - the MPI crashes for `mpiexec -n 2` are almost certainly environment/runner related.

PR #4092, opened 2026-02-16
Fix dtype in interpolation test

schnellerhase:
Caught in FFCx CI.

Fixes FFCx CI ref https://github.com/FEniCS/ffcx/pull/819.

### DIFF
--- a/python/test/unit/fem/test_interpolation.py
+++ b/python/test/unit/fem/test_interpolation.py
@@ -1158,7 +1158,7 @@ def test_submesh_interpolation_mapped(ghost_mode):
     comm = MPI.COMM_WORLD
 
     N = 8
-    domain = create_unit_cube(comm, N, N, N, ghost_mode=ghost_mode)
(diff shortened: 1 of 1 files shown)


python/test/unit/fem/test_interpolation.py
```diff
@@ -1173,7 +1173,7 @@ def test_submesh_interpolation_mapped(ghost_mode):
     parent_cells = sub_to_parent.sub_topology_to_topology(smsh_cells, inverse=False)
 
     degree = 1
-    el = element("N1curl", domain.basix_cell(), degree)
+    el = element("N1curl", domain.basix_cell(), degree, dtype=default_real_type)
```
jorgensd, line 1176: Then `dtype` should be set in the mesh constructor as well
```python
N = 8
    domain = create_unit_cube(comm, N, N, N, ghost_mode=ghost_mode, dtype=default_real_type)

python/test/unit/fem/test_interpolation.py
jorgensd, line 1184: `dtype` here should also be changed to match the `default_scalar_type`.

PR #4093, opened 2026-02-17
Add `nb::arg` to mixed topology mesh constructor

jorgensd:


garth-wells:
I don't quite follow the PR title - 'add nanobind arg'?

We haven't exposed all mixed topology interfaces to Python because it's still experimental and the interface still needs a thorough review. It's rough in parts and needs improvement.

PR #4094, opened 2026-02-17
Add overload type declaration for `AdjacencyList`

schnellerhase:
Ref https://github.com/scientificcomputing/networks_fenicsx/pull/18#discussion_r2514900046.


python/dolfinx/graph.py
```diff
@@ -69,7 +74,7 @@ def __repr__(self):
         """String representation of the adjacency list."""
         return self._cpp_object.__repr__()
 
-    def links(self, node: np.int32 | np.int64) -> npt.NDArray[np.int32 | np.int64]:
+    def links(self, node: _T) -> npt.NDArray[_T]:
```
garth-wells: Type of `node` is `int`.

garth-wells (approved):

PR #4099, opened 2026-02-23
Radix sort template argument fix, provides performance boost

garth-wells:
The `BITS` template argument was not being passed from `dolfinx::sort_perm` to `dolfinx::radix_sort`, and the ordering of the template arguments made it hard to set (by stopping helpful type deduction).

Observed to lead to a significant speed up of sort function calls in topology construction. 

Code has also been simplified.

### DIFF
--- a/python/CMakeLists.txt
+++ b/python/CMakeLists.txt
@@ -186,3 +186,28 @@ endif()
 set_target_properties(cpp PROPERTIES INSTALL_RPATH_USE_LINK_PATH TRUE)
 
 install(TARGETS cpp DESTINATION dolfinx)
+
+nanobind_add_stub(
+  cpp_stub
+  MODULE dolfinx.cpp
+  DEPENDS cpp
+  VERBOSE
+  RECURSIVE
+  INSTALL_TIME
+  OUTPUT_PATH dolfinx
+  OUTPUT
+    dolfinx/cpp/fem/__init__.pyi
+    dolfinx/cpp/fem/petsc.pyi
+    dolfinx/cpp/la/__init__.pyi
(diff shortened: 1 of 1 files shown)


cpp/dolfinx/common/sort.h
```diff
@@ -50,124 +49,119 @@ struct __unsigned_projection
...
+  for (I i = 0; i < its; i++)
+  {
+    // Zero counter array
+    std::ranges::fill(counter, 0);
 
-    // Adjacency list arrays for computing insertion position
-    std::array<I, bucket_size> counter;
-    std::array<I, bucket_size + 1> offset;
+    // Count number of elements per bucket
+    for (auto c : current_perm)
```
schnellerhase, line 141:
```suggestion
    for (const auto& c : current_perm)
```
`c` has generic type `T` not projected type `I` - it is non trivial in general and thus should never be copied.

cpp/dolfinx/common/sort.h
```diff
@@ -50,124 +49,119 @@ struct __unsigned_projection
...
 
-    uI mask_offset = 0;
-    std::vector<T> buffer(range.size());
-    std::span<T> current_perm = range;
-    std::span<T> next_perm = buffer;
-    for (I i = 0; i < its; i++)
+    // Prefix sum to get the inserting position
+    offset[0] = 0;
+    std::partial_sum(counter.begin(), counter.end(), std::next(offset.begin()));
+    for (auto c : current_perm)
```
schnellerhase, line 147:
```suggestion
    for (const auto& c : current_perm)
```

PR #4096, opened 2026-02-18
add cpp stub files

qbisi:
Hi, this patch generate stub files for dolfinx.cpp module using nanobind's builtin nanobind_add_stub macro.
An extra marker file py.typed will be generated under site-packages/dolfinx/cpp/py.typed. Not sure if this is needed as we have a top-level one in site-packages/dolfinx/py.typed.

PR #4101, opened 2026-02-24
CI coverage of auto generated stubs

schnellerhase:
Readds https://github.com/FEniCS/dolfinx/pull/4096 - now with cpp standalone mypy check for CI coverage, which catches https://github.com/FEniCS/dolfinx/pull/4096#issuecomment-3949804941.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -128,9 +128,15 @@ jobs:
         run: cmake --build . --target test
 
       - name: Build Python interface
-        run: |
-          pip install --check-build-dependencies --no-build-isolation --config-settings=cmake.build-type="Developer" 'python/[test]'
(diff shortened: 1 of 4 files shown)


jhale:
Would be good to get this substantial PR in before it gets stale. What needs sorting still?

python/dolfinx/fem/function.py
jhale, line 32: I do have my concerns here - if we add another type, we need to change this definition in 100 files. Can't it be in a library and imported?

schnellerhase:
Mainly we need to fix the mypy CI checks, they are non exhaustive in the sense that we currently do not run the checks against an installed dolfinx version. https://github.com/FEniCS/dolfinx/pull/4154 and https://github.com/FEniCS/dolfinx/pull/4131 are first steps to improve on this, so should also go in first.

jhale:
Merged into #4220

PR #4102, opened 2026-02-24
Fix wrong span extent in _lift_bc_interior_facets

michalhabera:
See e.g. https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/assemble_matrix_impl.h#L458


jhale:
Good catch. Are you sure you're running with all the standard library exceptions on? Writing past the bounds of span should cause an error.

jhale:
Fixes #4047

PR #4103, opened 2026-02-24
Pass correct cell/facet index in tabulate_expression

michalhabera:
See e.g. https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/fem/assemble_matrix_impl.h#L115


cpp/dolfinx/fem/assemble_expression_impl.h
```diff
@@ -107,7 +107,10 @@ void tabulate_expression(
          coord_dofs.data(), &entities(e, 1), nullptr, nullptr);
     }
 
-    P0(values_local, cell_info, e, size0);
+    if constexpr (entities.rank() == 1)
+      P0(values_local, cell_info, entities(e), size0);
+    else
+      P0(values_local, cell_info, entities(e, 0), size0);
```
schnellerhase: Move up into the other if-else block? There `entity` already has the correct signature.

schnellerhase (approved):

PR #4107, opened 2026-02-26
Expose JIT MPI communicator

jhale:
Allows the compilation of forms on a communicator different to the mesh. The hypothesis is that this may be useful for setting up JIT cache on HPC runs across multiple nodes with shared memory-backed `/tmp` - this is still to test.

Example of use:

```
from mpi4py import MPI
from dolfinx.fem import functionspace, form
import ufl

mesh = create_unit_square(MPI.COMM_WORLD, 10, 10)
V = functionspace(mesh, ("CG", 1))
u = ufl.TrialFunction(V)
v = ufl.TestFunction(V)
a = ufl.inner(u, v)*ufl.dx

# See https://docs.open-mpi.org/en/v5.0.x/man-openmpi/man3/MPI_Comm_split_type.3.html#split-types
shared_mem_comm = MPI.COMM_WORLD.Split_type(MPI.COMM_TYPE_SHARED, key=MPI.COMM_WORLD.rank)
a_dolfinx = form(a, jit_comm=shared_mem_comm)
```

The default behaviour (use the communicator from the form's mesh) is retained when `jit_comm=None`, which is also the default argument.


python/dolfinx/fem/forms.py
```diff
@@ -287,13 +290,19 @@ def mixed_topology_form(
         assert all([d is data[0] for d in data if d is not None])
 
     mesh = domain.ufl_cargo()
+    if mesh is None:
+        raise RuntimeError("Expecting to find a Mesh in the form.")
+    if jit_comm is None:
+        comm = mesh.comm
+    else:
+        comm = jit_comm
```
```suggestion
    comm = mesh.comm if jit_comm is None else jit_comm
```

PR #4108, opened 2026-02-27
Improve biharmonic demo

jorgensd:
Rework the biharmonic demo added by @juliusgh to include:
- Skip single precision runs. The matrix is very poorly conditioned and does not converge to the same solution for single and double precision (this also holds for the existing demo).
- Better choice of element (P2 yields sub-optimal convergence)
- Better choice of penalty parameter (thanks @schnellerhase)
- Usage of clamped boundary conditions, rather than simply supported bcs
- Explain the difference between the bcs.
- Use a method of manufactured solution as a verification mechanism
- Add relevant references to papers on discretization.

### DIFF
--- a/python/demo/demo_biharmonic.py
+++ b/python/demo/demo_biharmonic.py
@@ -322,7 +322,7 @@ ref_scale = fem.form(ufl.inner(u_ex, u_ex) * ufl.dx)
 local_scale = fem.assemble_scalar(ref_scale)
 scale = ref_scale.mesh.comm.allreduce(local_scale, op=MPI.SUM)
 print("||u_h-u_ex||_{L^2}^2/||u_ex||_L^2^2: ", f"{glob_error / scale:.5e}")
-assert default_real_type(glob_error) / default_real_type(scale) < 1e-6
+assert glob_error.real / scale.real < 1e-6
 
 # The solution can be written to a VTX-file using {py:class}`VTXWriter
 # <dolfinx.io.VTXWriter>` which can be opened with ParaView


python/demo/demo_biharmonic.py
```diff
@@ -173,33 +222,43 @@
...
+
+def u_manufactured(mod, x):
+    """Manufactured solution."""
+    return mod.sin(2 * mod.pi * x[0]) ** 2 * mod.sin(2 * mod.pi * x[1]) ** 2
+
 
-bc = fem.dirichletbc(value=ScalarType(0), dofs=dofs, V=V)
+g_D = fem.Function(V)
+g_D.interpolate(lambda x: u_manufactured(np, x))
+bc = fem.dirichletbc(value=g_D, dofs=dofs)
```
schnellerhase: Maybe simpler to read when expressed as interpolation of expression? Removes need for `u_manufactured` to support multiple modules.

python/demo/demo_biharmonic.py
```diff
@@ -119,35 +156,47 @@
...
+    exit(0)
+
 # We begin by using {py:func}`create_rectangle
 # <dolfinx.mesh.create_rectangle>` to create a rectangular
 # {py:class}`Mesh <dolfinx.mesh.Mesh>` of the domain, and creating a
 # finite element {py:class}`FunctionSpace <dolfinx.fem.FunctionSpace>`
 # $V$ on the mesh.
 
+N = 32
 msh = mesh.create_rectangle(
```
schnellerhase: `create_unit_square`?

python/demo/demo_biharmonic.py
```diff
@@ -225,28 +293,42 @@
...
+        "pc_type": "lu",
+    },
 )
 uh = problem.solve()
+
 assert isinstance(uh, fem.Function)
 assert problem.solver.getConvergedReason() > 0
+# -
+
+# We compute the error between the computed and exact solution:
```
schnellerhase:
```suggestion
# We compute the absolute $L^2$-error between the computed and exact solution:
```

python/demo/demo_biharmonic.py
```diff
@@ -24,6 +26,7 @@
 # - Solve a linear partial differential equation
 # - Use a discontinuous Galerkin method
 # - Solve a fourth-order differential equation
+# - Use the method of manufactured solutions
```
schnellerhase, line 29: Somewhere we should coin $C^0-$ *interior penalty method*. And maybe comment on connection to Nitsche's method with reference to your [demo](https://jsdokken.com/dolfinx-tutorial/chapter1/nitsche.html)?

python/demo/demo_biharmonic.py
```diff
@@ -84,31 +103,49 @@
...
+# L(v) = \int_{\Omega} fv ~{\rm d}x
+# +\sum_{E \in \mathcal{E}_h^{\rm ext}}\int_{E}\left(
+# -g_N \Delta v + \frac{\beta}{h_E} g_N \nabla v \cdot n
+# \right) ~{\rm d}s
 # $$
 #
 # Furthermore, $\left< u \right> = \frac{1}{2} (u_{+} + u_{-})$,
 # $[\!\![ w ]\!\!]  = w_{+} \cdot n_{+} + w_{-} \cdot n_{-}$,
-# $\alpha \ge 0$ is a penalty parameter and
+# $\beta \ge 0$ is a penalty parameter and
```
schnellerhase, line 131: We should demonstrate penalty methods not with a magical constant, at least dependency on $p$ and $h$ should be accounted for. Maybe best to adopt a more elaborate choice here, for example https://doi.org/10.1515/jnma-2023-0028 (A.1 holds plug&play `ufl` formula)

jorgensd: Nice find! Im adding it.

python/demo/demo_biharmonic.py
jorgensd: I'll add the reference, but I don't think there is a point to add cross references to my demo,
as we are dont doing Dirichlet through nitsche.

python/demo/demo_biharmonic.py
```diff
@@ -36,46 +39,62 @@
...
+# order equations (see the [Mixed Poisson demo](./demo_mixed-poisson)
+# for an example of a mixed method), or a variational
+# formulation can be constructed that imposes weak continuity of normal
+# derivatives between finite element cells. This demo uses a discontinuous
+# Galerkin approach to impose continuity of the normal derivative weakly,
+# see for instance {cite:t}`babuska1973penalty` or
+# {cite:t}`Georgoulis2009biharmonic`.
+#
+# In this demo, we consider equation 3.20-3.21 from
+# {cite:t}`Georgoulis2009biharmonic`, but instead of use a broken
```
michalhabera:
```suggestion
# {cite:t}`Georgoulis2009biharmonic`, but instead of using a broken
```

python/demo/demo_biharmonic.py
```diff
@@ -225,28 +293,42 @@
...
+
+# We compute the error between the computed and exact solution:
+
+# +
+error = fem.form(ufl.inner(uh - g_D, ufl.conj(uh - g_D)) * ufl.dx)
+local_error = fem.assemble_scalar(error)
+glob_error = np.sqrt(error.mesh.comm.allreduce(local_error, op=MPI.SUM))
+print(f"Global_error: {glob_error:.5e}")
+
+assert glob_error < 1e-3
```
michalhabera: Is it usually better to compare dimensionless quantities against each other, so I'd suggest to assemble the manufactured solution `ref = fem.form(ufl.inner(g_D, ufl.conj(g_D)) * ufl.dx)` and check for the relative error wrt. `ref`, if you have time.

jorgensd: The reason for choosing this comparison is that it makes it easier to set up a convergence study, which is how we realized that you get sub-optimal convergence for the degree=2 case.

michalhabera: I don't see the convergence study here and I don't understand how dividing with a reference constant changes the rate?

jorgensd: So I might have misunderstood your suggestion, as to me it read like you wanted to compare
$\int_\Omega u_h^2~\mathrm{d}x$ with $\int_\Omega u_{ex}^2~\mathrm{d}x$, while now I think you want to compute $\frac{L^2(\Omega)(u_h-u_{ex})}{\int_\Omega u_{ex}^2~\mathrm{d}x}$.
Is my new understanding what you initially meant?

The convergence study is not part of the code, but would be easy to get by adding a loop over the mesh creation (either by increasing N or refinement through `mesh.refine(..)`).

python/demo/demo_biharmonic.py
```diff
@@ -173,50 +223,75 @@
...
 
-bc = fem.dirichletbc(value=ScalarType(0), dofs=dofs, V=V)
+# +
+def u_manufactured(x):
+    """Manufactured solution."""
+    return ufl.sin(2 * ufl.pi * x[0]) ** 2 * ufl.sin(2 * ufl.pi * x[1]) ** 2
+
+
+x = ufl.SpatialCoordinate(msh)
+u_ex = u_manufactured(x)
```
schnellerhase:
```suggestion
x = ufl.SpatialCoordinate(msh)
u_ex = ufl.sin(2 * ufl.pi * x[0]) ** 2 * ufl.sin(2 * ufl.pi * x[1]) ** 2
```

python/demo/demo_biharmonic.py
schnellerhase: Did that improve the conditioning/results for the single precision run as well?

python/demo/demo_biharmonic.py
michalhabera: Yes, I meant $\lVert u_h - u_\text{ex}\rVert^2_{L^2(\Omega)} / \lVert u_\text{ex}\rVert^2_{L^2(\Omega_0)} $, where you can take the $\Omega_0$ be the initial mesh to avoid the interpolation error there.

This is a dimensionless number, so makes sense to compare to $0.3$ or whatever. Besides other arguments, even though not too important in this example, you'd notice that your error value depends on the mesh area, so with scaled mesh the assert is too loose/too tight.

python/demo/demo_biharmonic.py
jorgensd: Global error is still 2 orders of magnitude bigger than for double precision:
Global_error: 1.78900e-02
while double precision yields:
Global_error: 1.24028e-04

schnellerhase: Then we should think about preconditioning this at a later stage. BDDC could be a solution here https://arxiv.org/pdf/1811.06941 - would require https://github.com/FEniCS/dolfinx/pull/3885.

python/demo/demo_biharmonic.py
```diff
@@ -225,28 +295,43 @@
...
-with io.XDMFFile(msh.comm, out_folder / "biharmonic.xdmf", "w") as file:
-    V1 = fem.functionspace(msh, ("Lagrange", 1))
-    u1 = fem.Function(V1)
-    u1.interpolate(uh)
-    file.write_mesh(msh)
-    file.write_function(u1)
+error = fem.form(ufl.inner(uh - g_D, ufl.conj(uh - g_D)) * ufl.dx)
+local_error = fem.assemble_scalar(error)
+glob_error = error.mesh.comm.allreduce(local_error, op=MPI.SUM)
+ref_scale = fem.form(ufl.inner(u_ex, ufl.conj(u_ex)) * ufl.dx)
```
michalhabera: On a sidenote, I thought the `ufl.inner` applies `ufl.conj` on the second argument, is it needed?

python/demo/demo_biharmonic.py
```diff
@@ -225,28 +295,43 @@
...
 )
 uh = problem.solve()
+
 assert isinstance(uh, fem.Function)
 assert problem.solver.getConvergedReason() > 0
+# -
 
-# The solution can be written to a  {py:class}`XDMFFile
-# <dolfinx.io.XDMFFile>` file visualization with ParaView or VisIt
+# We compute the absolute $L^2$-error between the computed
```
schnellerhase:
```suggestion
# We compute the relative $L^2$-error between the computed
```

python/demo/demo_biharmonic.py
jorgensd:
```suggestion
error = fem.form(ufl.inner(uh - g_D, uh - g_D) * ufl.dx)
local_error = fem.assemble_scalar(error)
glob_error = error.mesh.comm.allreduce(local_error, op=MPI.SUM)
ref_scale = fem.form(ufl.inner(u_ex, u_ex) * ufl.dx)
```

python/demo/demo_biharmonic.py
```diff
@@ -225,28 +295,43 @@
...
-    file.write_mesh(msh)
-    file.write_function(u1)
+error = fem.form(ufl.inner(uh - g_D, uh - g_D) * ufl.dx)
+local_error = fem.assemble_scalar(error)
+glob_error = error.mesh.comm.allreduce(local_error, op=MPI.SUM)
+ref_scale = fem.form(ufl.inner(u_ex, u_ex) * ufl.dx)
+local_scale = fem.assemble_scalar(ref_scale)
+scale = ref_scale.mesh.comm.allreduce(local_scale, op=MPI.SUM)
+print("||u_h-u_ex||_{L^2}^2/||u_ex||_L^2^2: ", f"{glob_error / scale:.5e}")
+assert default_real_type(glob_error) / default_real_type(scale) < 1e-6
```
jorgensd:
```suggestion
assert glob_error.real / scale.real < 1e-6
```

PR #4114, opened 2026-03-01
Make partitioner options explicit

garth-wells:
Interface had default argument for max number of cells connected to a facet. This option can change fundamentally what is computed, so should be explicit. Will also allow thread option to be added.


jhale:
@garth-wells Can this be merged? Looks harmless enough.

PR #4117, opened 2026-03-02
Remove default behavior from non-affine pullback.

jorgensd:
Currently the non-affine pullback termination criterion for the Newton solver is not exposed, leading to several people having issues with convergence on non-affine geometries (https://fenicsproject.discourse.group/t/newton-method-failed-to-converge-for-non-affine-geometry-when-evaluating-the-function/11653/8?u=dokken).

In general, we would like people to be able to access these arguments when doing their stuff, such as `Function::eval` and `interpolate_nonmatching`.

This PR removes the default arguments from the C++ layer, and only introduce them at the top-level routines in the Python-API.

### DIFF
--- a/cpp/dolfinx/fem/Function.h
+++ b/cpp/dolfinx/fem/Function.h
@@ -425,7 +425,7 @@ public:
   /// @param[in] cells Cells in the mesh associated with `this` to
   /// interpolate into.
   /// @param[in] tol Tolerance for convergence in Newton method for non-affine
-  /// pullbacks. If the mesh geometr is affine this argument is ignored.
+  /// pullbacks. If the mesh geometry is affine this argument is ignored.
   /// @param[in] maxit Maximum number of Newton iterations in non-affine
   /// pull-back. If the mesh geometry is affine this argument is ignored.
   /// @param[in] interpolation_data Data required for associating the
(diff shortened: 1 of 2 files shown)


cpp/dolfinx/fem/Function.h
```diff
@@ -424,14 +424,18 @@ class Function
   /// @param[in] u Function to be interpolated.
   /// @param[in] cells Cells in the mesh associated with `this` to
   /// interpolate into.
+  /// @param[in] tol Tolerance for convergence in Newton method for non-affine
+  /// pullbacks. If the mesh geometr is affine this argument is ignored.
```
chrisrichardson:
```suggestion
  /// pullbacks. If the mesh geometry is affine this argument is ignored.
```

cpp/dolfinx/fem/Function.h
```diff
@@ -446,9 +450,13 @@ class Function
   /// points with a negative cell index. This argument must be passed
   /// with the correct size. Storage is row-major.
   /// @param[in] ushape Shape of `u`.
+  /// @param[in] tol Tolerance for convergence in Newton method for non-affine
+  /// pullbacks. If the mesh geometr is affine this argument is ignored.
```
chrisrichardson:
```suggestion
  /// pullbacks. If the mesh geometry is affine this argument is ignored.
```

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1169,12 +1169,16 @@ void interpolate(Function<T, U>& u, std::span<const T> f,
 /// @param u0 Function to interpolate from.
 /// @param cells Cells indices relative to the mesh associated with `u`
 /// that will be interpolated into.
-/// @param interpolation_data Data required for associating the
-/// interpolation points of `u` with cells in `v`. This is computed by
+/// @param[in] tol Tolerance for convergence in Newton method for non-affine
+/// pullbacks. If the mesh geometr is affine this argument is ignored.
```
chrisrichardson:
```suggestion
/// pullbacks. If the mesh geometry is affine this argument is ignored.
```

PR #4118, opened 2026-03-04
Extend derivative block to derive residuals from functionals

### LINKED ISSUE #3766: Extend `derivative_block()` to derive residuals from functionals

Many common variational problems (e.g. elasticity) are most naturally written as energy minimisation problems. The new 'derivative_block()` function takes block residuals to block Jacobians, but not functionals defined on multiple Functions to block residuals.

### Suggested user interface

No change to current interface to `derivate_block`, just extend it to handle the functional case.

Joseabcd:
Make `fem.forms.derivative_block` work on functionals to derive residuals, as described in #3766.

The following notes could be useful for the review:
1. The new logic is in a new code paragraph, independent of the original code. Not sure if a different style is preferred
2. I have used `ValueError` exceptions as opposed to `assert`s, assuming that the checks are on user input (as opposed to checks against bugs). The original code however has some `assert`s, so I’m not sure which is their right intent
3. I have not covered the case where `F` is a list of functionals
4. The original code creates a list of trial functions independently. Perhaps it should be updated to use `ufl.TrialFunctions(ufl.MixedFunctionSpace(…))`?
5. As in the original code, I have not checked that `du` is the correct type of argument (test vs trial), but possibly this might be helpful

CLOSES: #3766

### DIFF
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -702,12 +702,11 @@ def _derive_block_residual(
 ) -> Sequence[ufl.Form]:
     if du is None:
         du = ufl.TestFunctions(ufl.MixedFunctionSpace(*(u_i.function_space for u_i in u)))
-    elif (not isinstance(du, Sequence) 
-            or not len(u) == len(du) 
+    elif (not isinstance(du, Sequence)
             or not all(isinstance(du_i, ufl.Argument) and du_i.number() == 0 for du_i in du)):
             raise ValueError(
-                "When F is a functional of N functions, du must be a sequence "
-                "containing N test functions"
+                "When F is a functional of multiple functions, du must be a sequence of test "
+                "functions"
             )
     return ufl.extract_blocks(ufl.derivative(F, u, du))
 
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -793,7 +793,7 @@ def derivative_block(
         else:
             raise ValueError("u must be either a ufl.Function or a sequence of ufl.Function")
     elif isinstance(F, ufl.Form) and len(F.arguments()) == 1:
-        return _derive_univariate_jacobian(F, u, du)
+        return _derive_univariate_jacobian(F, u, du)  # type: ignore[arg-type]
     elif isinstance(F, Sequence):
         return _derive_block_jacobian(F, u, du)
     else:
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -757,22 +757,34 @@ def derivative_block(
     This is commonly used to derive a block residual from a functional, or to
     derive a block Jacobian from a block residual.
 
-    If ``F`` is a univariate rank zero form, returns the residual computed as
-    :math:`R = \\frac{\\partial F}{\\partial u}[\\delta u]`.
+    Four cases are supported:
 
-    If ``F`` is a rank one form, returns the Jacobian computed as :math:`J =
-    \\frac{\\partial F}{\\partial u}[\\delta u]`.
+    1. ``F`` is a rank-zero ``ufl.Form``, and ``u`` is a ``ufl.Function``.
+        Returns a ``ufl.Form`` representing the residual :math:`R =
+        \\frac{\\partial F}{\\partial u}[\\delta u]`. This is equivalent to
+        calling {py:func}`ufl.derivative` directly.
 
-    The two operations above are identical to calling {py:func}`ufl.derivative` directly.
+    2. ``F`` is a rank-zero `ufl.Form``, and ``u`` is a list of ``ufl.Function``.
+        Returns a list of ``ufl.Form`` representing the block residual :math:`R`,
+        with :math:`R_i = \\frac{\\partial F}{\\partial u_i}[\\delta u_i]`, where
+        :math:`\\delta u_i` is a test subfunction of the mixed space defined by
+        ``u``. This is equivalent to calling {py:func}`ufl.extract_blocks` on the
+        result from {py:func}`ufl.derivative`.
 
-    If ``F`` is a multivariate rank zero form, returns the block residual as a list with
-    :math:`R_i = \\frac{\\partial F}{\\partial u_i}[\\delta u_i]`, where 
-    :math:`\\delta u_i` is a test subfunction of the mixed space defined by u. This is
-    equivalent to calling {py:func}`ufl.extract_blocks` on the result from {py:func}`ufl.derivative`.
+    3. ``F`` is a rank-one `ufl.Form``, and ``u`` is a ``ufl.Function``.
+       Returns a ``ufl.Form`` representing the Jacobian :math:`J =
+       \\frac{\\partial F}{\\partial u}[\\delta u]`. This is equivalent to
+       calling {py:func}`ufl.derivative` directly.
 
-    If ``F`` is a list of rank one forms, returns the block Jacobian as a list of lists with
-    :math:`J_{ij} = \\frac{\\partial F_i}{u_j}[\\delta u_j]` using {py:func}`ufl.derivative`
-    called component-wise.
+    4. ``F`` is a list of rank-one `ufl.Form``, and ``u`` is a list of
+       ``ufl.Function``. Returns a list of lists representing the block Jacobian
+       :math:`J`, with :math:`J_{ij} = \\frac{\\partial F_i}{u_j}[\\delta u_j]`
+       using {py:func}`ufl.derivative` called component-wise.
+
+    Args:
+        F: UFL form(s) to be derived.
+        u: Function(s) with respect to the derivative is computed.
+        du: UFL argument(s) representing the direction of the derivative.
     """  # noqa: D301
     
     if isinstance(F, ufl.Form) and not F.arguments():
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -695,7 +695,7 @@ def _derive_univariate_residual(
     return ufl.derivative(F, u, du)
 
 
-def _derive_multivariate_residual(
+def _derive_block_residual(
     F: ufl.Form,
     u: Sequence[ufl.Form],
     du: Sequence[ufl.Argument] | None = None,
@@ -728,7 +728,7 @@ def _derive_univariate_jacobian(
     return ufl.derivative(F, u, du)
 
 
-def _derive_multivariate_jacobian(
+def _derive_block_jacobian(
     F: Sequence[ufl.Form],
     u: Sequence[ufl.Form],
     du: Sequence[ufl.Argument] | None = None,
@@ -783,13 +783,13 @@ def derivative_block(
         if isinstance(u, Function):
             return _derive_univariate_residual(F, u, du)
         elif isinstance(u, Sequence):
-            return _derive_multivariate_residual(F, u, du)
+            return _derive_block_residual(F, u, du)
         else:
             raise ValueError("u must be either a ufl.Function or a sequence of ufl.Function")
     elif isinstance(F, ufl.Form) and len(F.arguments()) == 1:
         return _derive_univariate_jacobian(F, u, du)
     elif isinstance(F, Sequence):
-        return _derive_multivariate_jacobian(F, u, du)
+        return _derive_block_jacobian(F, u, du)
     else:
         raise ValueError(
             "F must be either a UFL form (with rank zero or one), or a sequence of "
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -700,8 +700,6 @@ def _derive_block_residual(
     u: Sequence[ufl.Form],
     du: Sequence[ufl.Argument] | None = None,
 ) -> Sequence[ufl.Form]:
-    if not all(isinstance(u_i, Function) for u_i in u):
-        raise ValueError("u contains some non ufl.Function elements")
     if du is None:
         du = ufl.TestFunctions(ufl.MixedFunctionSpace(*(u_i.function_space for u_i in u)))
     elif (not isinstance(du, Sequence) 
@@ -737,8 +735,6 @@ def _derive_block_jacobian(
         raise ValueError("F contains some non rank-one ufl.Form elements")
     if not isinstance(u, Sequence):
         raise ValueError("When F is a sequence, u must be a sequence")
-    if not all(isinstance(u_i, Function) for u_i in u):
-        raise ValueError("u contains some non ufl.Function elements")
     if du is None:
         du = [ufl.TrialFunction(u_i.function_space) for u_i in u]
     elif (not isinstance(du, Sequence)
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -699,6 +699,23 @@ def derivative_block(
     \\frac{\\partial F}{\\partial u}[\\delta u]`. This is identical to
     calling ``ufl.derivative`` directly.
     """  # noqa: D301
+    
+    if isinstance(F, ufl.Form) and not F.arguments():  # if F is a functional
+        if isinstance(u, Function):
+            if du is None:
+                du = ufl.TestFunction(u.function_space)
+            elif not isinstance(du, ufl.Argument):
+                raise ValueError("When F is a functional of a single function, du must be a ufl.Argument")
+        elif all([isinstance(u_i, Function) for u_i in u]):
+            if du is None:
+                du = ufl.TestFunctions(ufl.MixedFunctionSpace(*(u_i.function_space for u_i in u)))
+            elif len(u) != len(du) or not all([isinstance(du_i, ufl.Argument) for du_i in du]):
+                raise ValueError("When F is a functional of N functions, du must be a sequence containing N ufl.Argument")
+        else:
+            raise ValueError("u must be a function or sequence of functions")
+        return ufl.extract_blocks(ufl.derivative(F, u, du))
+
+
     if isinstance(F, ufl.Form):
         if not isinstance(u, Function):
             raise ValueError("Must provide a single function when F is a UFL form")
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -730,13 +730,15 @@ def derivative_block(
             raise ValueError("Must provide a single function when F is a UFL form")
         if du is None:
             du = ufl.TrialFunction(u.function_space)
+        elif not (isinstance(du, ufl.Argument) and du.number() == 1):
+            raise ValueError("When F is a rank-one UFL form, du must be a trial function")
         return ufl.derivative(F, u, du)
     else:
         if not all([isinstance(Fi, ufl.Form) for Fi in F]):
             raise ValueError("F must be a sequence of UFL forms")
         if du is not None:
-            if len(F) != len(du)
-                raise ValueError("Number of forms and du must be equal")
+            if len(F) != len(du) or not all([isinstance(du_i, ufl.Argument) and du_i.number() == 1 for du_i in du]):
+                raise ValueError("Number of forms and du must be equal, and du must only contain trial functions")
         else:
             du = [ufl.TrialFunction(u_i.function_space) for u_i in u]
         return [[ufl.derivative(Fi, u_j, du_j) for u_j, du_j in zip(u, du)] for Fi in F]
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -706,8 +706,6 @@ def _derive_univariate_jacobian(
     u: Function,
     du: ufl.Argument | None = None,
 ) -> ufl.Form:
-    if not isinstance(u, Function):
-        raise ValueError("When F is a rank-one UFL form, u must be a ufl.Function")
     if du is None:
         du = ufl.TrialFunction(u.function_space)
     return ufl.derivative(F, u, du)
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -762,19 +762,20 @@ def derivative_block(
     This is commonly used to derive a block residual from a functional, or to
     derive a block Jacobian from a block residual.
 
-    If ``F`` is a univariate rank zero form, the residual is computed as
+    If ``F`` is a univariate rank zero form, returns the residual computed as
     :math:`R = \\frac{\\partial F}{\\partial u}[\\delta u]`.
 
-    If ``F`` is a multivariate rank zero form, the residual is a list with
-    :math:`R_i = \\frac{\\partial F}{\\partial u_i}[\\delta u_i]`, where 
-    :math:`\\delta u_i` is a test subfunction of the mixed space defined by u.
-
-    If ``F`` is a rank one form, the Jacobian is computed as :math:`J =
+    If ``F`` is a rank one form, returns the Jacobian computed as :math:`J =
     \\frac{\\partial F}{\\partial u}[\\delta u]`.
 
-    All three operations above are identical to calling {py:func}`ufl.derivative` directly.
+    The two operations above are identical to calling {py:func}`ufl.derivative` directly.
 
-    If ``F`` is a list of rank one forms, the Jacobian is a list of lists with
+    If ``F`` is a multivariate rank zero form, returns the block residual as a list with
+    :math:`R_i = \\frac{\\partial F}{\\partial u_i}[\\delta u_i]`, where 
+    :math:`\\delta u_i` is a test subfunction of the mixed space defined by u. This is
+    equivalent to calling {py:func}`ufl.extract_blocks` on the result from {py:func}`ufl.derivative`.
+
+    If ``F`` is a list of rank one forms, returns the block Jacobian as a list of lists with
     :math:`J_{ij} = \\frac{\\partial F_i}{u_j}[\\delta u_j]` using {py:func}`ufl.derivative`
     called component-wise.
     """  # noqa: D301
--- a/python/dolfinx/fem/forms.py
+++ b/python/dolfinx/fem/forms.py
@@ -691,7 +691,7 @@ def derivative_block(
     This is commonly used to derive a block Jacobian from a block
     residual.
 
-    If ``F_i`` is a list of forms, the Jacobian is a list of lists with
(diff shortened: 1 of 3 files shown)


python/dolfinx/fem/forms.py
```diff
@@ -686,19 +686,45 @@ def derivative_block(
...
-    \\frac{\\partial F}{\\partial u}[\\delta u]`. This is identical to
-    calling ``ufl.derivative`` directly.
+    If ``F`` is a multivariate rank zero form, the residual is a list with
+    :math:`R_i = \\frac{\\partial F}{\\partial u_i}[\\delta u_i]`, where 
+    :math:`\\delta u_i` is a test subfunction of the mixed space defined by u.
+
+    If ``F`` is a rank one form, the Jacobian is computed as :math:`J =
+    \\frac{\\partial F}{\\partial u}[\\delta u]`.
+
+    All three operations above are identical to calling ``ufl.derivative`` directly.
```
```suggestion
    All three operations above are identical to calling {py:func}`ufl.derivative` directly.
```

python/dolfinx/fem/forms.py
```diff
@@ -686,19 +686,45 @@ def derivative_block(
...
+
+    If ``F`` is a list of rank one forms, the Jacobian is a list of lists with
+    :math:`J_{ij} = \\frac{\\partial F_i}{u_j}[\\delta u_j]` using ``ufl.derivative``
```
```suggestion
    :math:`J_{ij} = \\frac{\\partial F_i}{u_j}[\\delta u_j]` using {py:func}`ufl.derivative`
```

python/dolfinx/fem/forms.py
```diff
@@ -686,19 +686,45 @@ def derivative_block(
...
+    called component-wise.
     """  # noqa: D301
+    
+    if isinstance(F, ufl.Form) and not F.arguments():  # if F is a functional
+        if isinstance(u, Function):
+            if du is None:
+                du = ufl.TestFunction(u.function_space)
+            elif not isinstance(du, ufl.Argument):
```
jorgensd: Should we also check that it is a `TestFunction`, i.e. `du.number() == 0`, i.e. we shouldn't allow a `TrialFunction` to be sent in.

python/dolfinx/fem/forms.py
```diff
@@ -686,19 +686,45 @@ def derivative_block(
...
+                raise ValueError("When F is a functional of a single function, du must be a ufl.Argument")
+        elif all([isinstance(u_i, Function) for u_i in u]):
+            if du is None:
+                du = ufl.TestFunctions(ufl.MixedFunctionSpace(*(u_i.function_space for u_i in u)))
+            elif len(u) != len(du) or not all([isinstance(du_i, ufl.Argument) for du_i in du]):
```
jorgensd: Same comment here.

python/test/unit/fem/test_forms.py
```diff
@@ -132,3 +134,78 @@ def test_multiple_measures_one_subdomain_data():
...
+    with pytest.raises(ValueError):
+        derivative_block(F, u0, v0)  # second argument not a ufl.Function
+
+    with pytest.raises(ValueError):
+        derivative_block(F, f0, u0)  # third argument not a test function
+
+    R = derivative_block(F, f0)
+    assert isinstance(R, ufl_form) and len(R.arguments()) == 1
+
+    R = derivative_block(F, f0, v0)
```
Joseabcd: In this case, where `u` and `du` are not sequences (what I called "univariate" functional), there is a subtle choice I made worth mentioning.

In the odd event that the user passes `v0` from a mixed space, applying `ufl.extract_blocks` to the output of `ufl.derivative` will be **a sequence** containing one `ufl.Form`. In other words, the result can be a `ufl.Form` or a `Sequence` depending on whether `v0` is part or not of a mixed space (!)

I found this potentially confusing, so I avoided applying `ufl.extract_blocks` in `derivative_block` when the input `du` is just a function

python/test/unit/fem/test_forms.py
```diff
@@ -132,3 +134,78 @@ def test_multiple_measures_one_subdomain_data():
...
+    V0 = functionspace(mesh, ("Lagrange", 1))
+    V1 = functionspace(mesh, ("Lagrange", 2))
+    V = MixedFunctionSpace(V0, V1)
+
+    f0, f1 = dolfinx.fem.Function(V0), dolfinx.fem.Function(V1)
+    v0, v1 = TestFunctions(V)
+    u0, u1 = TrialFunctions(V)
+
+
+    F = f0**2 * dx  # univariate functional
```
jhale: By convention we use `F` for residuals, i.e. rank one forms and `J` for Jacobians, i.e. rank two forms. Could you use something different for than `F` for functionals? Would make it easier to read the test.

python/dolfinx/fem/forms.py
```diff
@@ -681,35 +681,117 @@ def create_form(
...
+    if du is None:
+        du = ufl.TestFunction(u.function_space)
+    elif not isinstance(du, ufl.Argument) or du.number() != 0:
+        raise ValueError(
+            "When F is a functional of a single function, du must be a test function"
+        )
+    return ufl.derivative(F, u, du)
+
+
+def _derive_multivariate_residual(
```
jhale: On these method names, would e.g. `_derive_block_residual` be more consistent with naming elsewhere?

python/dolfinx/fem/forms.py
```diff
@@ -681,35 +681,117 @@ def create_form(
...
+def _derive_univariate_jacobian(
+    F: ufl.Form,
+    u: Function,
+    du: ufl.Argument | None = None,
+) -> ufl.Form:
+    if not isinstance(u, Function):
+        raise ValueError("When F is a rank-one UFL form, u must be a ufl.Function")
+    if du is None:
+        du = ufl.TrialFunction(u.function_space)
+    elif not isinstance(du, ufl.Argument) or du.number() != 1:
```
jhale: Shouldn't this be raised within `ufl.derivative` already?

python/dolfinx/fem/forms.py
```diff
@@ -681,35 +681,117 @@ def create_form(
...
+def _derive_multivariate_residual(
+    F: ufl.Form,
+    u: Sequence[ufl.Form],
+    du: Sequence[ufl.Argument] | None = None,
+) -> Sequence[ufl.Form]:
+    if not all(isinstance(u_i, Function) for u_i in u):
+        raise ValueError("u contains some non ufl.Function elements")
+    if du is None:
+        du = ufl.TestFunctions(ufl.MixedFunctionSpace(*(u_i.function_space for u_i in u)))
+    elif (not isinstance(du, Sequence)
```
jhale: Shouldn't this be raised within `ufl.derivative` or `ufl.extract_blocks` already?

python/dolfinx/fem/forms.py
```diff
@@ -681,35 +681,117 @@ def create_form(
...
+def _derive_multivariate_jacobian(
+    F: Sequence[ufl.Form],
+    u: Sequence[ufl.Form],
+    du: Sequence[ufl.Argument] | None = None,
+) -> Sequence[Sequence[ufl.Form]]:
+    if not all(isinstance(F_i, ufl.Form) and len(F_i.arguments()) == 1  for F_i in F):
+        raise ValueError("F contains some non rank-one ufl.Form elements")
+    if not isinstance(u, Sequence):
+        raise ValueError("When F is a sequence, u must be a sequence")
+    if not all(isinstance(u_i, Function) for u_i in u):
```
jhale: Too picky - let UFL handle it.

python/dolfinx/fem/forms.py
```diff
@@ -681,35 +681,117 @@ def create_form(
...
+        raise ValueError("When F is a rank-one UFL form, du must be a trial function")
+    return ufl.derivative(F, u, du)
+
+
+def _derive_multivariate_jacobian(
+    F: Sequence[ufl.Form],
+    u: Sequence[ufl.Form],
+    du: Sequence[ufl.Argument] | None = None,
+) -> Sequence[Sequence[ufl.Form]]:
+    if not all(isinstance(F_i, ufl.Form) and len(F_i.arguments()) == 1  for F_i in F):
```
jhale: Too picky, let UFL handle it.

python/dolfinx/fem/forms.py
```diff
@@ -681,35 +681,117 @@ def create_form(
...
+        raise ValueError("F contains some non rank-one ufl.Form elements")
+    if not isinstance(u, Sequence):
+        raise ValueError("When F is a sequence, u must be a sequence")
+    if not all(isinstance(u_i, Function) for u_i in u):
+        raise ValueError("u contains some non ufl.Function elements")
+    if du is None:
+        du = [ufl.TrialFunction(u_i.function_space) for u_i in u]
+    elif (not isinstance(du, Sequence)
```
jhale: Let UFL handle it.

python/dolfinx/fem/forms.py
```diff
@@ -681,35 +681,117 @@ def create_form(
...
-        if du is None:
-            du = ufl.TrialFunction(u.function_space)
-        return ufl.derivative(F, u, du)
-    else:
-        assert all([isinstance(Fi, ufl.Form) for Fi in F]), "F must be a sequence of UFL forms"
-        assert len(F) == len(u), "Number of forms and functions must be equal"
-        if du is not None:
-            assert len(F) == len(du), "Number of forms and du must be equal"
+    
+    if isinstance(F, ufl.Form) and not F.arguments():
```
jhale, line 792: What does `not F.arguments()` do/why does it work?

python/dolfinx/fem/forms.py
jorgensd: It is not raised, ref:
```python
from ufl import (
    FunctionSpace,
    Mesh,
    Coefficient,
    dx,
    TrialFunction,
    derivative
)
import basix.ufl
from mpi4py import MPI
import dolfinx
cell = "interval"
P1 = basix.ufl.element("Lagrange", cell, 1, shape=())
domain = Mesh(basix.ufl.element("Lagrange", cell, 1, shape=(1,)))
V = FunctionSpace(domain, P1)

u = Coefficient(V)
J = u ** 2 * dx
v = TrialFunction(V)
F = derivative(J, u, v)
F_compiled = dolfinx.fem.compile_form(MPI.COMM_WORLD, F)
```
which runs with no problem, but becomes a problem if you want to get the jacobian

python/dolfinx/fem/forms.py
jorgensd: `F.arguments()` returns a tuple of `ufl.Argument`. if a tuple is empty, its boolean state is `False`, thus the check says, if  we have no arguments (rank=0) go there.

python/test/unit/fem/test_forms.py
schnellerhase: This should be added to the documentation.

python/test/unit/fem/test_forms.py
```diff
@@ -132,3 +134,78 @@ def test_multiple_measures_one_subdomain_data():
     J_local = dolfinx.fem.assemble_scalar(J)
     J_global = comm.allreduce(J_local, op=MPI.SUM)
     assert np.isclose(J_global, 1 / 3 + 1 / 2)
+
+
+def test_derivative_block():
```
schnellerhase, line 148: Nice defensive programming! 

For easier readability I would rename
1)  `V -> VM`
2) `F -> F0` 
3) `R -> R0`
4) the 'multivariate functional' to `{}_block` (functional and variations)

python/test/unit/fem/test_forms.py
Joseabcd: I've now named the functionals in the test file as `M`, and the residuals as `F`. 

Let me know if `F` should also be good to avoid in the signatures of `derivative_block` and its small convenience functions. I avoided changing those because I can't find a good short generic name for the first parameter of `derivative_block`

python/dolfinx/fem/forms.py
Joseabcd: Should we then keep/remove this check? In my view, if the concept of deriving a rank-zero form w.r.t. a “trial” function never ever makes sense (that is, if it’s conceptually wrong at the UFL layer), then it’d be UFL’s responsibility to complain

python/dolfinx/fem/forms.py
Joseabcd: The UFL functions aren’t raising here either. For example, `ufl.derivative(F, [u0, u1, u2], [du0])` silently returns a sequence with just one form. But I agree this is UFL’s job, so I have for now removed the check `len(u) == len(du)`. 

I will wait to remove the other checks that remain in this `if`, depending on the answer to question in [the previous comment](https://github.com/FEniCS/dolfinx/pull/4118#discussion_r2921327214).

python/dolfinx/fem/forms.py
Joseabcd: I agree and I have removed it. UFL here raises something. For example, running `ufl.derivative(M, [f1, "bad function"], [v1, v2])` tells the user:
```
File "/Users/jose/Desktop/test.py", line 359, in <module>
    ufl.derivative(M, [f1, f2, "bad function"], [v1, v2, v3])
  File "/Users/jose/fenicsx-src/lib/python3.12/site-packages/ufl/formoperators.py", line 423, in derivative
    coefficients, arguments = _handle_derivative_arguments(form, coefficient, argument)
                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jose/fenicsx-src/lib/python3.12/site-packages/ufl/formoperators.py", line 339, in _handle_derivative_arguments
    if c.ufl_shape != a.ufl_shape:
       ^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'ufl_shape'
```
This can be helpful, even though ideally UFL should say something like: “ValueError: Got an invalid coefficient to derive form. The sequence of input coefficients should contain ufl.Function elements.”

python/dolfinx/fem/forms.py
Joseabcd: Before removing it, we need to answer the question in [the comment above](https://github.com/FEniCS/dolfinx/pull/4118#discussion_r2921327214): should we allow `derivative_block` to derive a functional using a “trial” function?

python/dolfinx/fem/forms.py
Joseabcd: The tests `isinstance(du, Sequence)`and `len(u) == len(du)` I believe are appropriate. Given that the concept of “component-wise” Jacobian is created here in the Python layer, isn't this layer responsible for making sure the result is a square “matrix”? UFL will never get to see anything other than single pairs of `(u_i, du_i)`, so will never get a chance to complain.

About the third check on `du_i.number()`, I will wait to remove it depending on answer to question in [this comment above](https://github.com/FEniCS/dolfinx/pull/4118#discussion_r2921327214)

python/test/unit/fem/test_forms.py
Joseabcd: > For easier readability I would rename
> 
> 1. `V -> VM`
> 2. `F -> F0`
> 3. `R -> R0`
> 4. the 'multivariate functional' to `{}_block` (functional and variations)

@schnellerhase, using @jhale's suggestion I'm now using `F` for residuals, and here I renamed the functional to `M`

Taking your 4th point, I've appended the suffix "block" to the "multivariate" variables. 

I think `F0` and `J0` could be confused as from the space `V0`, and `VM` for the functional `M`?  Do you think it's fine to leave as it is, or use something like `V_mixed`? I've seen typical space names are `V` `Q`, `W`, and typical function names `u`, `p`, but I think here the naming with subsequent numbers makes it more suitable to add more spaces/functions, if the tests had to be expanded

python/dolfinx/fem/forms.py
Joseabcd: I've renamed `_derive_multivariate_residual` to `_derive_block_residual`, and `_derive_multivariate_jacobian` to `_derive_block_jacobian`

Joseabcd:
> This looks pretty good but I think it's a bit 'strongly typed/overly checked' with if statements - shouldn't a lot of these already be picked up lower down in UFL?

I removed some of the checks suggested above. I'm happy to remove the rest of those suggested, but I left a question [in an inline comment](https://github.com/FEniCS/dolfinx/pull/4118#discussion_r2921327214) that I think needs to be resolved first

python/test/unit/fem/test_forms.py
Joseabcd: @schnellerhase I'll update the docstring, but could you tell me if I made the right choice? I'm still exploring the rest of the project and I lack the big picture of how this fits in the rest of the python layer.

I think that in essence, my dilemma is captured in the following question:

Given:
```
mesh = dolfinx.mesh.create_unit_interval(MPI.COMM_WORLD, 10)
V0 = functionspace(mesh, ("Lagrange", 1))
V1 = functionspace(mesh, ("Lagrange", 2))
V = MixedFunctionSpace(V0, V1)
f0, f1 = dolfinx.fem.Function(V0), dolfinx.fem.Function(V1)

M = f0**2 * dx
```
which of the following is the more accurate statement?
- `M` is a functional defined over a single function. We don't know about "blocks".
- `M` is a functional defined over `N` functions (where `N` happens to be 1). We work with "blocks".

python/test/unit/fem/test_forms.py
schnellerhase: There is a `_block` suffix to the (second) `M` declaration missing - overwriting the former. Otherwise I'm happy with this 😄 

> I think F0 and J0 could be confused as from the space V0, and VM for the functional M? Do you think it's fine to leave as it is, or use something like V_mixed?

The mixed function space for the variational forms does not imply anything mathematically - this becomes only relevant when assembly/discretisation is introduced. I.e. space `V0` and  `V[0]` are the same, so your functionals do live on `V0`. But I think it is clear enough now no need to alter.

python/test/unit/fem/test_forms.py
schnellerhase: Going along the argument in https://github.com/FEniCS/dolfinx/pull/4118#discussion_r2922751437 - I would say the first option is the favourable perspective. Going with the perspective: the block interpretation is at the symbolic level one of argument counting, not one of the setup of the argument's function spaces. What do you think @jorgensd - does that make any sense?

python/test/unit/fem/test_forms.py
Joseabcd: thank you, I amended `M` to `M_block` now

python/test/unit/fem/test_forms.py
Joseabcd: > This should be added to the documentation.

I implemented in the docstring a more structured format that I observed in other functions from the project. I hope the detail about the type of returned output depending on the input is clearer now

jorgensd: I also think the first one is the correct thing to consider here. As `f1` does not enter the functional, the construct of blocking them is of no interest to the functional itself.

python/dolfinx/fem/forms.py
jorgensd: `ufl` strictly speaking actually handles this correctly.
If we consider this minimal example using the Lagrange-element from the UFL test suite:
```python
from ufl import (
    FunctionSpace,
    Mesh,
    Coefficient,
    dx,
    TrialFunction,
    derivative,
    interval,
    TestFunction
)

from utils import LagrangeElement

cell = interval
P1 = LagrangeElement(cell, 1, shape=())
domain = Mesh(LagrangeElement(cell, 1, shape=(1,)))
V = FunctionSpace(domain, P1)

u = Coefficient(V)
J = u ** 2 * dx
v = TestFunction(V)
F = derivative(J, u, v)
N = derivative(F, u)
for arg in F.arguments():
    print("Test F", arg.number(), arg.part())
for arg in N.arguments():
    print("Test N", arg.number(), arg.part())

v = TrialFunction(V)
F_t = derivative(J, u, v)
N_t = derivative(F_t, u)
for arg in F_t.arguments():
    print("Trial F", arg.number(), arg.part())
for arg in N_t.arguments():
    print("Trial N", arg.number(), arg.part())
```
this returns:
```bash
Test F 0 None
Test N 0 None
Test N 1 None
Trial F 1 None
Trial N 1 None
Trial N 2 None
```
which means that if you do not supply the derivative direction of your second variation, you get an argument that has its number incremented by one, which is then correct in UFL terms, as the first variation is the one with the lowest numbering.

Joseabcd: @jorgensd was your comment in favor of removing or keeping the check? I was waiting for others to comment too.

I see UFL’s approach is to simply apply the math operation requested by the user (the variational problem is not UFL layer’s business). The question that remains is whether it is `derivative_block`’s business. If yes, then it should check that test/trial arguments are correct. 

I am myself not sure. On one hand it would be convenient if `derivative_block` complains about a poorly defined problem. On the other hand, removing the check gives the user more flexibility to define the problem in unconventional ways; for example, deriving a functional first w.r.t. a trial function, and then w.r.t. a test function (which results in a seemingly-valid rank-2 form, with argument numbers 0 and 1).

@jhale what do you think? I am ok to proceed either way. Could you let me know any concerns about this and the rest of changes? Happy to work on them

jorgensd: I think checking it becomes complicated, as seen by the example above, where `number!=1` if one uses `derivative` on a linear form (which has been defined with a trial function). 

So i am in favor of removing the check and leave it to the user to define their forms correctly.

Joseabcd: I have removed the check now

python/dolfinx/fem/forms.py
Joseabcd: Following the answer in linked comment, I’ve removed now the check that remained

python/dolfinx/fem/forms.py
Joseabcd: Following the answer in linked comment, I have removed it now

python/dolfinx/fem/forms.py
Joseabcd: Following the answer in linked comment, I’ve removed now the `du_i.number()` check

Joseabcd:
I think all comments above have been addressed now. The stringent checks on the inputs have mostly been removed, letting inputs reach UFL in most cases. Some checks remain inside `_derive_block_jacobian` (see [comment](https://github.com/FEniCS/dolfinx/pull/4118#discussion_r2921363634)), to prevent from breaking the code that here calculates the component-wise Jacobian.

Could you please have another look at it when you have time, and let me know if there are still any concerns?

jhale (approved):
Thanks this looks very nice, and super useful for people working with energy minimization principles.

python/test/unit/fem/test_forms.py
schnellerhase, line 3:
```suggestion
# Copyright (C) 2021-2026 Garth N. Wells and [...]
```

Add yourself to the copyright notice (if you want) - same for `forms.py`.

schnellerhase (approved):

schnellerhase:
Could you [link the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) so it auto closes/links to the PR?

Joseabcd:
> Could you [link the issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) so it auto closes/links to the PR?

done now

python/test/unit/fem/test_forms.py
Joseabcd: @schnellerhase thank you! but wouldn't adding more commits now require PR re-reviews? From my side there is no need

schnellerhase: No re-review needed for that, just don't force push at this point then it is easier to see what changed in the history - your call.

Joseabcd: ok done then thanks (I asked because some projects are configured to auto invalidate approvals if further commits pushed)

schnellerhase:
Can you fix up the linting issues caught in the CI?

Joseabcd:
> Can you fix up the linting issues caught in the CI?

Thanks for bringing it up. I've amended it and it's now passing those checks.

I've also verified that sphinx will generate the documentation nicely from `derivative_block`'s docstring. I did some extra checks adding it to a temporary local documentation build and noticed issues with code hyperlinks (they weren't working with the `{py:fun}` syntax, and switched to `:func:`).

schnellerhase:
There seem to be still ruff checks failing. Ensure you have the latest version installed when running the checks.

Joseabcd:
I've pushed some new fixes as `mypy` couldn’t resolve some types and the CI was failing again

In `derivative_block` I've used `cast`s to tell `mypy` the intended types explicitly. I see no perfect fix, but to choose this I thought:
- We wanted to let bad inputs reach the UFL layer, so I avoided extra `if`s or `assert`s (to avoid coming back to the approach discarded above with the excessive checks). With `cast`, the execution path is preserved
- We want to preserve the type annotations in the four small utility functions, as they serve as clear documentation about what inputs/outputs correspond to each case

In `python/dolfinx/fem/petsc.py`, I've also added an `assert` to narrow down the type of the output of `derivative_block`, and avoid a type mismatch. This is a consequence of using the same function for deriving functionals and residuals (as the return type now is a bit more generic).

Please could you let me know if these fixes look good, or if you would have preferred a different way? I am happy to amend again if so

Joseabcd:
The CI was failing again with a type error, so I pushed [f265dab](https://github.com/FEniCS/dolfinx/pull/4118/commits/f265dab66df2182a714f485ab12dea29d4c0d344) with a small fix.

I also applied and then reverted [500faaa](https://github.com/FEniCS/dolfinx/pull/4118/commits/500faaa5acd7a7493e556d697729734e92850e3f). This was to fix some additional errors `mypy` was reporting in my local run, but I reverted it when noticed that [GitHub's CI log](https://github.com/FEniCS/dolfinx/actions/runs/24996193085/job/73265646844?pr=4118#step:8:23) didn't complain about them. I'm not sure why I got more errors when running `mypy` locally in my computer.

I also pushed a small fix for a wrong type specification that my local run with `mypy` caught (see [5eca5e4](https://github.com/FEniCS/dolfinx/pull/4118/commits/5eca5e471802aa7392971b3727c53ec3a5c1695f)).

I didn't remove the reverted commit to avoid force-pushing, but let me know if the commit history is worth a clean up.

schnellerhase:
Ace on the git management 😉 - reproducing the CI mypy setup locally is close to impossible at the moment, ref. https://github.com/FEniCS/dolfinx/pull/4135.

PR #4123, opened 2026-03-12
Introduce threading for collision detection

jorgensd:
Following the way of implementation #4083, introduce threading for GJK, in the case where you have 
N bodies $b_0, b_1, \dots, b_N$, which you would like to compute the distance to another body `q` for all bodies `b_i`.

This PR makes it possible to control the threading for this new interface, `compute_distances_gjk`.

 **GJK improvements**:
Introduce monotonicity criteria for distance norm v. If v starts increasing we have hit machine tolerance and therefore exit.

**Test improvements**:
Removes alot of looping from the bounding box tests, in favor of a numpy/vectorized construction.
Tests `num_threads=1` and `num_threads=2` for the new algorithm.

**Bugs fixes**:
The tolerances used inside GJK was wrong <T> not <U>, causing the bounding box tests to fail if you refined the grid even slightly (the tolerance should use <U> as we have cast all data into that format).

### DIFF
--- a/cpp/dolfinx/geometry/gjk.h
+++ b/cpp/dolfinx/geometry/gjk.h
@@ -385,7 +385,7 @@ std::array<T, 3> compute_distance_gjk(std::span<const T> p0,
     // 2nd exit condition - strict monotonicity
     // Floating point can cause the algorithm to stagnate. Then we terminate.
     const U vn = impl_gjk::dot3(v, v);
-    if (vn > vnorm2)
+    if (vnorm2 <= vn)
       break;
 
     // 3nd exit condition - intersecting or touching


cpp/dolfinx/geometry/gjk.h
```diff
@@ -390,8 +390,66 @@ std::array<T, 3> compute_distance_gjk(std::span<const T> p0,
...
+std::vector<T>
+compute_distances_gjk(const std::vector<std::span<const T>>& bodies,
+                      std::span<const T> q, size_t num_threads)
+{
+  size_t total_size = bodies.size();
+  num_threads = std::min(num_threads, total_size);
+
+  std::vector<T> results(total_size * 3);
+  auto compute_chunk = [](std::span<T> out_data,
+                          std::span<const std::span<const T>> bodies_chunk,
```
chrisrichardson: I think it would be better to just pass `c0` and `c1` to the lambda, `span<span>` is a bit confusing.

cpp/dolfinx/geometry/gjk.h
```diff
@@ -318,7 +318,7 @@ std::array<T, 3> compute_distance_gjk(std::span<const T> p0,
   constexpr int maxk = 15; // Maximum number of iterations of the GJK algorithm
 
   // Tolerance
-  const U eps = 1.0e4 * std::numeric_limits<T>::epsilon();
+  const U eps = 1.0e4 * std::numeric_limits<U>::epsilon();
```
jorgensd:
```suggestion
  const U eps = 10 * std::sqrt(std::numeric_limits<U>::epsilon());
```

PR #4127, opened 2026-03-16
Improve non-matching interpolation for non-affine grids.

jorgensd:
For non-affine grids, the convex hull search often prove insufficient to get the correct colliding cell, for instance for non-matching interpolation (ref: https://fenicsproject.discourse.group/t/second-order-meshes-interpolation-error/19544/9?u=dokken).
This PR splits the non-affine (local) collision detection into two steps:
1. Check if point is within convex hull of cell.
2. If within convex hull, use the non-affine pullback to get point in reference space.

The logic for affine geometries is unchanged.
Requires #4126 as it showed up as a major performance bottleneck for this problem.

Test added to check the case where the convex hull of a cell contains a point, but the cell doesn't:
image

## Other new features
- Compute first colliding cell now supports mixed topology by passing in the cell_type index. For it to be truly useful with this feature, we need to rewrite `dolfinx::geometry::BoundingBoxTree` to support mixed topology meshes.
- Removed vector allocation from several hot-loops in the determine_point_ownership code, leveraging that each geometry dofmap has a fixed width (which wasn't assumed at the time of creation of this function).
## AI-usage
I used Gemini Pro to generate the ASCII figure that exaggerates the situation in the Paraview screenshot to illustrate the test.


schnellerhase:
For the approach

>     1. Check if point is within convex hull of cell.
> 
>     2. If within convex hull, use the non-affine pullback to get point in reference space.

we need that the convex hull property holds, so the higher order element is strictly contained in the convex hull of the support points, right? I'm having a hard time checking if we have a guarantee/assumption on this on general for higher order geometries in `dolfinx`, do we? For `gmsh` meshes this seems to be guaranteed https://gmsh.info/doc/preprints/gmsh_curved_preprint.pdf.

jorgensd:
> For the approach
> 
> > ```
> > 1. Check if point is within convex hull of cell.
> > 
> > 2. If within convex hull, use the non-affine pullback to get point in reference space.
> > ```
> 
> we need that the convex hull property holds, so the higher order element is strictly contained in the convex hull of the support points, right? I'm having a hard time checking if we have a guarantee/assumption on this on general for higher order geometries in `dolfinx`, do we? For `gmsh` meshes this seems to be guaranteed https://gmsh.info/doc/preprints/gmsh_curved_preprint.pdf.

If this doesn't hold, then we will get some minor interpolation errors as the points will not get picked up in the point ownership stage. The PR will make the interpolation error smaller, but we can still hit cases like the 1D plot sketched out in this post: https://fenicsproject.discourse.group/t/second-order-meshes-interpolation-error/19544/5?u=dokken, where any point $x\in [0, 0.5], y\in (2x^2-x, 0))$ is within the cell, but not the convex hull of the degrees of freedom.
The only alternative I can see to ensure correctness for this kind of operation is to make a convex hull based on points on the interface of the cells, similar to what is done in: https://a-latyshev.github.io/fenics26-tutorials/surrogate-grid-creation/#creating-the-real-boundary-gamma

PR #4128, opened 2026-03-17
Remove memory allocations from loops in GJK

jorgensd:
Major refactoring of GJK
- Remove memory allocations in various places within the code.
- Adjust the determinant computation to work in a fixed size array to avoid duplicate copies.
- Never copy the full body over to multi-precision/double from double/float. Only cast the simplex values that is required.
  - This introduces the need to compute the `support` in the normal precision, which can be achieved by normalizing the previous distance iterate `x_k` by its length.
  - Use const expr dispatch for the `nearest_simplex` computation, as it only has the runtime cases 2,3 and 4.

### DIFF
--- a/cpp/dolfinx/geometry/gjk.h
+++ b/cpp/dolfinx/geometry/gjk.h
@@ -397,8 +397,8 @@ std::array<T, 3> compute_distance_gjk(std::span<const T> p0,
     if (x_norm2 > eps * eps)
     {
       // ADL lookup:
-      // Will use std::sqrt for double/float, and
-      // boost::multiprecision::sqrt for U
+      // If U is double/float use std::sqrt
+      // If U is a boost::multiprecision member use boost::multiprecision::sqrt
       using std::sqrt;
       U inv_norm = U(1.0) / sqrt(x_norm2);
       x_k_normalized[0] *= inv_norm;
@@ -438,7 +438,7 @@ std::array<T, 3> compute_distance_gjk(std::span<const T> p0,
     if (xs_diff < (eps * x_norm2) or xs_diff < eps)
       break;
 
-    SPDLOG_DEBUG("GJK: vw={}/{}", static_cast<double>(xs_diff),
+    SPDLOG_DEBUG("GJK: xs_diff={}/{}", static_cast<double>(xs_diff),
                  static_cast<double>(eps));
 
     // Add new vertex to simplex


cpp/dolfinx/geometry/gjk.h
```diff
@@ -311,101 +365,139 @@ std::array<T, 3> compute_distance_gjk(std::span<const T> p0,
...
+
+    // Compute the squared norm of current iterate to normalize support search
+    // in original precision
+    const U x_norm2 = impl_gjk::dot3(x_k, x_k);
+    std::array<U, 3> x_k_normalized = x_k;
+    if (x_norm2 > eps * eps)
+    {
+      // ADL lookup:
+      // Will use std::sqrt for double/float, and
+      // boost::multiprecision::sqrt for U
```
chrisrichardson: What does this mean? Presumably `U` could be `double`?

cpp/dolfinx/geometry/gjk.h
```diff
@@ -311,101 +365,139 @@ std::array<T, 3> compute_distance_gjk(std::span<const T> p0,
...
-    const U vnorm2 = impl_gjk::dot3(v, v);
-    const U vw = vnorm2 - impl_gjk::dot3(v, w);
-    if (vw < (eps * vnorm2) or vw < eps)
+    // 1st exit condition: (x_k - s_k).x_k = 0
+    const U xs_diff = x_norm2 - impl_gjk::dot3(x_k, s_k);
+    if (xs_diff < (eps * x_norm2) or xs_diff < eps)
       break;
 
-    SPDLOG_DEBUG("GJK: vw={}/{}", static_cast<double>(vw),
+    SPDLOG_DEBUG("GJK: vw={}/{}", static_cast<double>(xs_diff),
```
chrisrichardson:
```suggestion
    SPDLOG_DEBUG("GJK: xs_diff={}/{}", static_cast<double>(xs_diff),
```

cpp/dolfinx/geometry/gjk.h
jorgensd: If U is a multiprecision number it will pick the `boost::multiprecision::sqrt`, while if it is a double/float it will use `std::sqrt`.

cpp/dolfinx/geometry/gjk.h
```diff
@@ -311,101 +370,139 @@ std::array<T, 3> compute_distance_gjk(std::span<const T> p0,
...
+
+    // Compute the squared norm of current iterate to normalize support search
+    // in original precision
+    const U x_norm2 = impl_gjk::dot3(x_k, x_k);
+    std::array<U, 3> x_k_normalized = x_k;
+    if (x_norm2 > eps * eps)
+    {
+      // ADL lookup:
+      // Will use std::sqrt for double/float, and
+      // boost::multiprecision::sqrt for U
```
jorgensd:
```suggestion
      // If U is double/float use std::sqrt
      // If U is a boost::multiprecision member use boost::multiprecision::sqrt
```

chrisrichardson (approved):

PR #4131, opened 2026-03-19
Add typing generics

schnellerhase:
Extends the approach of https://github.com/FEniCS/dolfinx/pull/4094 for multi export type wrappers to
- `Vector`
- `MatrixCSR`
- `SuperLUDistMatrix`
- `SuperLUDistSolver`
- [x] `DirichletBC`
- [x] `CoordinateElement`
- [x] `FiniteElement`
- [x] `Form`
- [x] `Constant`
- [x] `Expression`
- [x] `Function`
- [x] `FunctionSpace`
- [x] `Mesh`
- [x] `Geometry`
- [x] `CoordinateElement`
- [x] `BoundingBoxTree`
- [x] `PointOwnershipData`

### DIFF
--- a/python/demo/demo_lagrange_variants.py
+++ b/python/demo/demo_lagrange_variants.py
@@ -26,6 +26,7 @@
 from mpi4py import MPI
 
 import matplotlib.pylab as plt
+import numpy as np
 
 import basix
 import basix.ufl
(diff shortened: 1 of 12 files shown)


schnellerhase:
> Looks good - have you considered directly mimicking the naming/setup of the C++ concepts? Perhaps they could also be put somewhere more general for use across the whole interface.
> 
> https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/common/types.h#L26

AFAIK the python typing currently does not allow conditional type deductions, so we can not mimic any scalar/real type concepts.

jhale:
I think you're overcomplicating it a bit. My basic point is to add a `dolfinx.typing` module where you add meaningful definitions of e.g. `_Scalar` and `_Float` using `TypeVar` and then import them uniformly for use elsewhere. Perhaps some modules where the definition is isolated (e.g. MeshTags) the current design is appropriate.

jhale:
This should be merged when:

- `Makefile` is removed.
- Comment above on adding `dolfinx.typing` is responded to/addressed.

schnellerhase:
> This should be merged when:
> 
> * `Makefile` is removed.
> * Comment above on adding `dolfinx.typing` is responded to/addressed.

Has been rebased and Real/Scalar concepts centralised

PR #4135, opened 2026-03-20
Extend `mypy` checks to `test/`

schnellerhase:
- Extends checks to `python/test/`
- Enables (CI) checking of further type fixes to follow in https://github.com/FEniCS/dolfinx/pull/4131
- Fix: type of `dirichletbc` arg - `Number` is typing incompatible https://peps.python.org/pep-0484/#the-numeric-tower.

TODO in future
- Automatic stub generation
- Enable checks on `python/demo/`

Edited by @jhale


python/Makefile
jorgensd, line 1: I'm not sure I see the convenience of having this makefile, as it assumes:
- that there is a `python` executable, not `python3`.

python/demo/demo_pyamg.py
```diff
@@ -126,7 +126,7 @@ def poisson_problem(dtype: npt.DTypeLike, solver_type: str) -> None:
 
 
 # +
-def nullspace_elasticty(Q: fem.FunctionSpace) -> list[np.ndarray]:
+def nullspace_elasticty(Q: fem.FunctionSpace) -> npt.NDArray:
```
jorgensd, line 129: Should this then have a type, otherwise I don't see the point of changing it to `npt.NDArray`, could just be `np.ndarray`.

python/demo/demo_pyamg.py
```diff
@@ -84,7 +84,7 @@ def poisson_problem(dtype: npt.DTypeLike, solver_type: str) -> None:
 
     dofs = locate_dofs_topological(V=V, entity_dim=fdim, entities=facets)
 
-    bc = dirichletbc(value=dtype(0), dofs=dofs, V=V)
+    bc = dirichletbc(value=dtype(0.0), dofs=dofs, V=V)  # type: ignore
```
jorgensd, line 87: Doesn't this kind of show that the type hint for dirichletbc is wrong?

jorgensd (changes requested):

python/Makefile
schnellerhase: Added optional py variable, following https://gitlab.com/petsc/petsc/-/blob/main/src/binding/petsc4py/makefile?ref_type=heads#L9. 

Invoking with `make py=3 lint-mypy` will now use the `python3` executable. If not provided defaults to `python`.

python/demo/demo_pyamg.py
schnellerhase: Yes this should hold a type, but it needs the generic type of `FunctionSpace` - which will be introduced after this PR in https://github.com/FEniCS/dolfinx/pull/4131. So it is a first step into the right direction, dropping the `list`, but not quite perfect.

python/demo/demo_pyamg.py
schnellerhase: The problem arises from the explicit `dtype()` invocation:
```console
demo/demo_pyamg.py:87: error: "dtype[Any]" not callable  [operator]
```
I tried multiple fixes but I am not sure how to correctly create a scalar of a certain type in a typing compliant way. Arrays work fine, but constructor of a dtype seems to be a different case.

jhale:
This can be merged once the `Makefile` is removed. Please add instructions to the `python/README.md`.

schnellerhase:
> This can be merged once the `Makefile` is removed. Please add instructions to the `python/README.md`.

https://github.com/FEniCS/dolfinx/pull/4154 addresses this. 

I do think we need a way to allow for local replication of CI failures during development. If it is not makefiles then we should think about another way to do this.

jhale:
Please put developer documentation in python/README.md

schnellerhase:
Setup for https://github.com/FEniCS/dolfinx/pull/4216


```makefile
# Note: mypy is currenlty on CI tested without petsc4py!

PYTHON = python$(py)

lint-mypy:
	$(PYTHON) -m venv dolfinx-venv-mypy --system-site-packages
	. dolfinx-venv-mypy/bin/activate && pip install mypy types-cffi scipy-stubs scikit-build-core nanobind
	. dolfinx-venv-mypy/bin/activate && pip install git+https://github.com/fenics/ufl.git
	. dolfinx-venv-mypy/bin/activate && pip install git+https://github.com/fenics/basix.git --config-settings=build-dir=build/basix
	. dolfinx-venv-mypy/bin/activate && pip install git+https://github.com/fenics/ffcx.git
	. dolfinx-venv-mypy/bin/activate && pip install . --no-build-isolation --config-settings=build-dir=build/dolfinx -Ccmake.build-type="Developer"
	. dolfinx-venv-mypy/bin/activate && mypy -p dolfinx
	. dolfinx-venv-mypy/bin/activate && mypy test
	. dolfinx-venv-mypy/bin/activate && mypy demo

lint-mypy-clean:
	rm -rf dolfinx-venv-mypy

lint: lint-mypy

clean: lint-mypy-clean
```

jhale:
Remaining Spack error is pyamg problem, going to merge.

.github/workflows/ccpp.yml
```diff
@@ -129,24 +129,15 @@ jobs:
...
-          mypy -p dolfinx
-
-      - name: Run mypy  # Use a venv to avoid NumPy upgrades that are incompatible with numba
-        working-directory: python
         run: |
-          python -m venv mypy-env
-          . ./mypy-env/bin/activate
-          pip install mypy types-cffi scipy-stubs
-          mypy -p dolfinx
+          mypy --config-file python/pyproject.toml -p dolfinx # package mode, type checks installed dolfinx
```
```suggestion
          mypy --config-file python/pyproject.toml -p dolfinx # package mode, type checks installed dolfinx (working around name collision in python/ directory)
```

jhale: This issue would be fixed if we moved across to the now recommended `src/` directory layout.

https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

PR #4136, opened 2026-03-21
Simplifying lifting

chrisrichardson:
Simplifies `apply_lifting` by reusing the matrix assembler with a different inlined "mat_add" lambda kernel.
The main advantage is that we can delete some 600+ lines of code from `assemble_vector_impl.h` since this is mostly repeated code of the assembler, which should help with future maintenance.
Some changes are needed to `impl::assemble_matrix`, but by using templating (bool argument), the actual executed code for matrix assembly should be unchanged, and performance is unaffected.

### DIFF
--- a/cpp/dolfinx/fem/assemble_matrix_impl.h
+++ b/cpp/dolfinx/fem/assemble_matrix_impl.h
@@ -140,7 +140,7 @@ void assemble_cells_matrix(
     P0(Ae, cell_info0, cell0, ndim1);  // B = P0 \tilde{A}
     P1T(Ae, cell_info1, cell1, ndim0); // A =  B P1_T
 
-    // Do not clear BC rows/cols in "LiftingMode"
+    // In lifting mode only BC dofs are assembled, while in standard mode these row/column dofs are zeroed.
     if constexpr (!LiftingMode)
     {
       // Zero rows and columns for BCs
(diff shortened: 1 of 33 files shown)


cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -89,6 +91,7 @@ void assemble_cells_matrix(
   const int ndim1 = bs1 * num_dofs1;
   std::vector<T> Ae(ndim0 * ndim1);
   std::vector<scalar_value_t<T>> cdofs(3 * x_dofmap.extent(1));
+  std::vector<std::int32_t> bce0(ndim0), bce1(ndim1);
```
chrisrichardson:
```suggestion
  std::vector<std::int32_t> bce0(ndim0)
  std::vector<std::int32_t> bce1(ndim1);
```

jorgensd:
If this is ready for review, could you write a brief summary of the proposed changes, and show how/if it changes performance.

cpp/dolfinx/fem/assemble_matrix_impl.h
```diff
@@ -115,38 +140,38 @@ void assemble_cells_matrix(
     P0(Ae, cell_info0, cell0, ndim1);  // B = P0 \tilde{A}
     P1T(Ae, cell_info1, cell1, ndim0); // A =  B P1_T
 
-    // Zero rows/columns for essential bcs
-    std::span dofs0(dmap0.data_handle() + cell0 * num_dofs0, num_dofs0);
-    std::span dofs1(dmap1.data_handle() + cell1 * num_dofs1, num_dofs1);
-
-    if (!bc0.empty())
+    // Do not clear BC rows/cols in "LiftingMode"
```
jorgensd:
```suggestion
    // In lifting mode only BC dofs are assembled, while in standard mode these row/column dofs are zeroed.
```

jorgensd (approved):
LGTM

PR #4139, opened 2026-03-31
Use PETSc CTX for SNES solver

jorgensd:
Spawned by the discussion in https://gitlab.com/petsc/petsc/-/work_items/1883.
This is a better way to pass around the `_blocks` value if required (at least more PETSc like).
Maybe @dalcinl can comment on if this is better than our current approach.

### DIFF
--- a/python/dolfinx/fem/petsc.py
+++ b/python/dolfinx/fem/petsc.py
@@ -1270,7 +1270,7 @@ class NonlinearProblem:
         # Get potential attributes from the residual to pass to the
         # residual assembly function, e.g. block layout for block assembly.
         context = {}
-        if (_blocks=:self.b.getAttr("_blocks")) is not None:
+        if (_blocks:=self.b.getAttr("_blocks")) is not None:
             context["_blocks"] = _blocks
         context["u"] = u
         context["residual"] = self.F
--- a/python/dolfinx/fem/petsc.py
+++ b/python/dolfinx/fem/petsc.py
@@ -1014,9 +1014,9 @@ def assemble_residual(
     """Assemble the residual at ``x`` into the vector ``b``.
 
     A function conforming to the interface expected by ``SNES.setFunction``
-   A function conforming to the interface expected by ``SNES.setFunction``
-   by setting all arguments except `snes`, `x` and `b` through the `kargs`
-   keyword argument.
+    A function conforming to the interface expected by ``SNES.setFunction``
+    by setting all arguments except `snes`, `x` and `b` through the `kargs`
+    keyword argument.
 
     Example::
 
@@ -1086,10 +1086,10 @@ def assemble_jacobian(
 
     A function conforming to the interface expected by
     ``SNES.setJacobian`` can be created by fixing the first four
-   A function conforming to the interface expected by
-   ``SNES.setJacobian`` can be created by setting all
-   arguments except `_snes`, `x`, `J` and `P_mat`
-   through the `kargs` argument e.g.:
+    A function conforming to the interface expected by
+    ``SNES.setJacobian`` can be created by setting all
+    arguments except `_snes`, `x`, `J` and `P_mat`
+    through the `kargs` argument e.g.:
 
     Example::
 
@@ -1108,7 +1108,7 @@ def assemble_jacobian(
         jacobian: Compiled form of the Jacobian.
         preconditioner: Compiled form of the preconditioner.
         bcs: List of Dirichlet boundary conditions to apply to the Jacobian
-             and preconditioner matrices.
+            and preconditioner matrices.
     """
     # Copy existing soultion into the function used in the residual and
     # Jacobian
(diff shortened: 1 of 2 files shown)


python/dolfinx/fem/petsc.py
```diff
@@ -1280,10 +1265,16 @@ def __init__(
...
         self._snes = PETSc.SNES().create(self.A.comm)  # type: ignore[attr-defined]
-        self.solver.setJacobian(
-            partial(assemble_jacobian, u, self.J, self.preconditioner, bcs), self.A, self.P_mat
-        )
-        self.solver.setFunction(partial(assemble_residual, u, self.F, self.J, bcs), self.b)
+        cntx = {"u": self.u, "jacobian": self.J, "preconditioner": self.preconditioner, "bcs": bcs}
+        self.solver.setJacobian(assemble_jacobian, self.A, self.P_mat, kargs=cntx)
+        # Get potential attributes from the residual to pass to the
+        # residual assembly function, e.g. block layout for block assembly.
+        context = self.b.getDict().copy()
```
schnellerhase: Do we ever want something else than `_blocks` to be forwarded here?

jorgensd: In the context, yes?
As an attribute, unsure.

```suggestion
        context = {}
        if (_blocks=:self.b.getAttr("_blocks")) is not None:
            context["_blocks"] = _blocks
```
We could go for this, what do you think @schnellerhase ?:)

schnellerhase: I like it! that should also work for `b.getDict()` holding other attributes than `_blocks`.

python/dolfinx/fem/petsc.py
```diff
@@ -1280,10 +1265,18 @@ def __init__(
...
-        self.solver.setJacobian(
-            partial(assemble_jacobian, u, self.J, self.preconditioner, bcs), self.A, self.P_mat
-        )
-        self.solver.setFunction(partial(assemble_residual, u, self.F, self.J, bcs), self.b)
+        cntx = {"u": self.u, "jacobian": self.J, "preconditioner": self.preconditioner, "bcs": bcs}
+        self.solver.setJacobian(assemble_jacobian, self.A, self.P_mat, kargs=cntx)
+        # Get potential attributes from the residual to pass to the
+        # residual assembly function, e.g. block layout for block assembly.
+        context = {}
+        if (_blocks=:self.b.getAttr("_blocks")) is not None:
```
```suggestion
        if (_blocks:=self.b.getAttr("_blocks")) is not None:
```

schnellerhase (approved):

dalcinl:
Passing extra data via `kargs` seems reasonable. However, this approach will work as long as the SNES solver is tied to a specific instance of a problem, and not reused for another instance of problem where the `kargs` would not correspond. From my limited pure-human non-AI-assisted understanding of the code, things seem correct.

python/dolfinx/fem/petsc.py
```diff
@@ -1280,10 +1265,18 @@ def __init__(
         # Create the SNES solver and attach the corresponding Jacobian and
         # residual computation functions
         self._snes = PETSc.SNES().create(self.A.comm)  # type: ignore[attr-defined]
-        self.solver.setJacobian(
-            partial(assemble_jacobian, u, self.J, self.preconditioner, bcs), self.A, self.P_mat
-        )
-        self.solver.setFunction(partial(assemble_residual, u, self.F, self.J, bcs), self.b)
+        cntx = {"u": self.u, "jacobian": self.J, "preconditioner": self.preconditioner, "bcs": bcs}
```
jhale: Naming through here with cntx and context is confusing.

jhale (commented):
I do understand the issue and agree it needs fixing.

That said, I do prefer the more explicit approach of using partial.

It's not clear how kargs corresponds to the function arguments when reading our documentation.

python/dolfinx/fem/petsc.py
```diff
@@ -1280,10 +1265,18 @@ def __init__(
...
+        self.solver.setJacobian(assemble_jacobian, self.A, self.P_mat, kargs=cntx)
+        # Get potential attributes from the residual to pass to the
+        # residual assembly function, e.g. block layout for block assembly.
+        context = {}
+        if (_blocks := self.b.getAttr("_blocks")) is not None:
+            context["_blocks"] = _blocks
+        context["u"] = u
+        context["residual"] = self.F
```
jhale: Can't all this be done in a single line? Then the blocks.

jhale:
Final thing - there is a mistake in the docstring, it says that the first X arguments should be fixed, but this no longer applies, they should be set through the `kargs` context.

python/dolfinx/fem/petsc.py
```suggestion
   A function conforming to the interface expected by ``SNES.setFunction``
   by setting all arguments except `snes`, `x` and `b` through the `kargs`
   keyword argument.
```

```suggestion
   A function conforming to the interface expected by
   ``SNES.setJacobian`` can be created by setting all
   arguments except `_snes`, `x`, `J` and `P_mat`
   through the `kargs` argument e.g.:
```

PR #4140, opened 2026-04-01
Allow expressions with submeshes

### LINKED ISSUE #3207: Support `dolfinx::fem::Expression` with co-dim 0 submeshes

With the co-dim 0 support in assembly (and interpolation post #3177 ), it would be natural to extend `dolfinx::fem::Expression` to handle test/trial-functions and coefficients from different sub-meshes, linked by `entity_maps`.

There are two different ways this could be supported, either by passing in `entity_maps` at Expression construction
or through eval (which would be closer to what is done in #3111 and with #2730).
For either approach, we need to designate a "base"-mesh that we map all entities from.

### Suggested user interface

```python3
mesh = ...
u = dolfinx.fem.Function(V)
mesh0, map0, _, _ = dolfinx.mesh.create_submesh(mesh, ..., i)
mesh1, map1, _, _ = dolfinx.mesh.create_submesh(mesh, ..., j)
u0 = dolfinx.fem.functionspace(mesh, ("DG", 2, (gdim, )))
u1 = dolfinx.fem.functionspace(mesh, ("N1curl", 1))
# Interface 1, would work nicely with interpolation
expr = dolfinx.fem.Expression(ufl.inner(u0, u1)*u, V.element.interpolation_points(),
                              domain=mesh, entity_maps={mesh0:map0_inverse, mesh1:map1_inverse})
# Interface 2, would require additional information in interpolation wrapper regarding entity maps.
expr_2 = dolfinx.fem.Expression(ufl.inner(u0, u1)*u, V.element.interpolation_points()
expr_2.eval(mesh, np.arange(num_cells_in_mesh), entity_maps={mesh0:map0_inverse, mesh1:map1_inverse})

u_new = dolfinx.fem.Function(V)
# Interpolation with interface 1
u_new.interpolate(expr)

# Interpolation with interface 2
u_new.interpolate(expr_2, domain=mesh, entity_maps={mesh0:map0_inverse, mesh1:map1_inverse})
```

jorgensd:
With submesh support, it is quite common that one want to evaluate an expression that uses quantities from both the parent and submesh.

This PR extends expression to be able to support:
- Submesh of codim 0
- Submesh of codim 1 (one sided facet expressions)

# Other structural changes
- Refactors the `pack_coefficients` for `dolfinx::fem::Expression` into a multi-step approach:
   1. Before packing for a coefficient transform the integration entities input into a list of cells for the given coefficient.
   2. Pack entities for given coefficient.

# Other relevant changes
A fix to the DOLFINx Expression on facet code to ensure that it is evaluated based on the global facet orientation. (https://github.com/FEniCS/ffcx/pull/836)

Requires: https://github.com/FEniCS/ffcx/pull/828
Resolves: https://github.com/FEniCS/dolfinx/issues/3207


.github/workflows/fenicsx-refs.env
```diff
@@ -3,4 +3,4 @@ basix_ref=main
 ufl_repository=FEniCS/ufl
 ufl_ref=main
 ffcx_repository=FEniCS/ffcx
-ffcx_ref=main
+ffcx_ref=dokken/submesh-expression
```
```suggestion
ffcx_ref=main
```

cpp/demo/hyperelasticity/main.cpp
```diff
@@ -326,7 +326,7 @@ int main(int argc, char* argv[])
     // Compute Cauchy stress. Construct appropriate Basix element for
     // stress.
     fem::Expression sigma_expression = fem::create_expression<T, U>(
-        *expression_hyperelasticity_sigma, {{"u", u}}, {});
+        *expression_hyperelasticity_sigma, {{"u", u}}, {}, {});
```
garth-wells, line 329: Some of these factory functions are getting out-of-control with the number of arguments. Use more functions to improve readability and reduce number of redundant arguments.

cpp/dolfinx/fem/assembler.h
```diff
@@ -70,6 +71,12 @@ void tabulate_expression(
         std::pair<std::reference_wrapper<const FiniteElement<U>>, std::size_t>>
         element)
 {
+  // Check that domain is the same as mesh of the expression
+  if (e.coordinate_element_hash() != mesh.geometry().cmap(0).hash())
```
garth-wells: It would be good to avoid code that supports single cell types meshes only given the recent work in this area.

cpp/dolfinx/fem/Expression.h
```diff
@@ -62,38 +64,35 @@ class Expression
...
-             std::span<const geometry_type> X,
-             std::array<std::size_t, 2> Xshape,
-             std::function<void(scalar_type*, const scalar_type*,
-                                const scalar_type*, const geometry_type*,
-                                const int*, const uint8_t*, void*)>
-                 fn,
-             const std::vector<std::size_t>& value_shape,
-             std::shared_ptr<const FunctionSpace<geometry_type>> argument_space
-             = nullptr)
+  Expression(
```
garth-wells, line 74: Try to refactor to simplify the constructor args.

cpp/dolfinx/fem/Expression.h
```diff
@@ -189,5 +201,12 @@ class Expression
 
   // Evaluation points on reference cell
   std::pair<std::vector<geometry_type>, std::array<std::size_t, 2>> _x_ref;
+
+  // Map between different mesh topologies
```
garth-wells: 'Maps from X to Y'. "Maps" is too generic.

cpp/dolfinx/fem/pack.h
```diff
@@ -332,17 +333,23 @@ void pack_coefficients(const Form<T, U>& form,
 /// @brief Pack coefficient data over a list of cells or facets.
 ///
 /// Typically used to prepare coefficient data for an ::Expression.
-/// @tparam T
-/// @tparam U
+/// @tparam T data type of coefficients
```
garth-wells: Inconsistent capitalisation.

cpp/dolfinx/fem/pack.h
```diff
@@ -332,17 +333,23 @@ void pack_coefficients(const Form<T, U>& form,
...
 /// @param coeffs Coefficients to pack
-/// @param offsets Offsets
+/// @param mesh Mesh which the entities belong to
 /// @param entities Entities to pack over
-/// @param[out] c Packed coefficients.
+/// @param entity_maps Entity maps for the coefficient meshes
+/// @param offsets Offsets
+/// @param[in,out] c Packed coefficients.
 template <dolfinx::scalar T, std::floating_point U>
 void pack_coefficients(
```
garth-wells, line 445: We should refactor to avoid growing compexity.

cpp/dolfinx/fem/pack.h
```diff
@@ -332,17 +333,23 @@ void pack_coefficients(const Form<T, U>& form,
...
-/// @tparam U
+/// @tparam T data type of coefficients
+/// @tparam U floating point type of mesh geometry
 /// @param coeffs Coefficients to pack
-/// @param offsets Offsets
+/// @param mesh Mesh which the entities belong to
 /// @param entities Entities to pack over
-/// @param[out] c Packed coefficients.
+/// @param entity_maps Entity maps for the coefficient meshes
+/// @param offsets Offsets
```
garth-wells: Expand docstring.

cpp/dolfinx/fem/pack.h
```diff
@@ -354,17 +361,105 @@ void pack_coefficients(
...
-                                    cell_info, cells, offsets[coeff]);
+      // Helper function to get correct entity map
+      auto get_entity_map
+          = [mesh, &entity_maps](auto& mesh0) -> const mesh::EntityMap&
+      {
+        auto it = std::ranges::find_if(
+            entity_maps,
+            [mesh, mesh0](const mesh::EntityMap& em)
+            {
+              return ((em.topology() == mesh0->topology()
```
garth-wells, line 481: Are these cheap (memory comparison) or expensive (data compare) operations?

cpp/dolfinx/fem/pack.h
```diff
@@ -354,17 +361,105 @@ void pack_coefficients(
...
+      if constexpr (entities.rank() == 1)
+      {
+        e_b = emap.sub_topology_to_topology(
+            std::span(entities.data_handle(), entities.size()), inverse);
+        md::mdspan e(e_b.data(), e_b.size());
+        impl::pack_coefficient_entity(std::span(c), cstride,
+                                      coeffs[coeff].get(), cell_info, e,
+                                      offsets[coeff]);
+      }
+      else if (entities.rank() == 2)
```
garth-wells: This is wrong because `constexpr` is used above.

cpp/dolfinx/fem/pack.h
```diff
@@ -354,17 +361,105 @@ void pack_coefficients(
...
+      {
+        const mesh::Topology topology = *mesh.topology();
+        int tdim = topology.dim();
+        int codim = tdim - mesh_c->topology()->dim();
+        if (codim == 0)
```
garth-wells: Should be a switch.

cpp/dolfinx/fem/pack.h
```diff
@@ -354,17 +361,105 @@ void pack_coefficients(
...
+        {
+          // Codim 1 mesh, need to map (cell, local index) to facets and then
+          // to cells of the submesh
+          auto c_to_f = topology.connectivity(tdim, tdim - 1);
+          assert(c_to_f);
+          std::vector<std::int32_t> parent_entities;
+          parent_entities.reserve(entities.extent(0));
+          for (std::size_t e = 0; e < entities.extent(0); ++e)
+          {
+            auto pair = md::submdspan(entities, e, md::full_extent);
```
garth-wells: I think this reduces readability.

garth-wells (dismissed):
It would be good in this could be simplified. There is a lot going on in some constructors. We need to refactor before adding more complexity to already complex code.

cpp/dolfinx/fem/assembler.h
jorgensd: As you have stated in many places in other PRs, the multiple cell type work is still very much in progress. 
There has been no development whatsoever with respect to mixed topology and `dolfinx.fem.Expression` as far as I am aware.
We could add a if  `mesh.topology().cell_types().size() > 1` throw this is not supported, but I don't want to refactor all of expression to support mixed topology in this PR, as we do not have a mixed topology mesh in the pure UFL file used for FFCx).

cpp/dolfinx/fem/Expression.h
jorgensd:
```suggestion
  // Set of bidirectional maps of mesh entities (cells, facets, ridges or peaks) between meshes and submeshes of any codimension.
```

cpp/dolfinx/fem/pack.h
jorgensd:
```suggestion
/// @tparam T Data type of coefficients
```

cpp/dolfinx/fem/pack.h
```diff
@@ -332,17 +333,23 @@ void pack_coefficients(const Form<T, U>& form,
 /// @brief Pack coefficient data over a list of cells or facets.
 ///
 /// Typically used to prepare coefficient data for an ::Expression.
-/// @tparam T
-/// @tparam U
+/// @tparam T data type of coefficients
+/// @tparam U floating point type of mesh geometry
```
jorgensd:
```suggestion
/// @tparam U Floating point type of mesh geometry
```

cpp/dolfinx/fem/pack.h
```diff
@@ -332,17 +333,23 @@ void pack_coefficients(const Form<T, U>& form,
...
 /// @param coeffs Coefficients to pack
-/// @param offsets Offsets
+/// @param mesh Mesh which the entities belong to
 /// @param entities Entities to pack over
-/// @param[out] c Packed coefficients.
+/// @param entity_maps Entity maps for the coefficient meshes
```
jorgensd:
```suggestion
/// @param entity_maps Bidirectional maps between the entities of a parent mesh and a submesh in case of coefficients being defined on both.
```

cpp/dolfinx/fem/pack.h
jorgensd:
```suggestion
/// @param offsets Insertion offset for each of the `coeffs` when packed into `c`.
```

cpp/dolfinx/fem/pack.h
jorgensd: EntityMap.topology() is a shared_ptr, same as mesh0->topology(), so these are cheap
https://github.com/FEniCS/dolfinx/blob/de8a1aad37d9fbd8ad6d0f7061beef297f3dc7b1/cpp/dolfinx/mesh/EntityMap.h#L77-L81
https://github.com/FEniCS/dolfinx/blob/de8a1aad37d9fbd8ad6d0f7061beef297f3dc7b1/cpp/dolfinx/mesh/Mesh.h#L71-L73

cpp/dolfinx/fem/pack.h
jorgensd:
```suggestion
      else if constexpr (entities.rank() == 2)
```

cpp/dolfinx/fem/pack.h
jorgensd: Then my suggestion would be to split out the `entities, entity_maps` into an array of `mapped_entities`, and move the mapping logic to a pre-processing step.
i.e. replace:
```C++
 int cstride = coffsets.back();
  {
    std::vector<std::reference_wrapper<const Function<T, U>>> c;
    std::ranges::transform(coefficients, std::back_inserter(c),
                           [](auto c) -> const Function<T, U>& { return *c; });
    fem::pack_coefficients(c, mesh, entities, e.entity_maps(), coffsets,
                           std::span(coeffs));
  }
  std::vector<T> constants = fem::pack_constants(e);

  tabulate_expression<T, U>(
      values, e, md::mdspan(coeffs.data(), entities.size(), cstride),
      std::span<const T>(constants), mesh, entities, element);
```
with something like
```C++
 int cstride = coffsets.back();
  {
    std::vector<std::reference_wrapper<const Function<T, U>>> c;
    std::ranges::transform(coefficients, std::back_inserter(c),
                           [](auto c) -> const Function<T, U>& { return *c; });
    /// Extract entity maps for each of the coefficients here.
    entity_maps_as_array = ...
    std::vector<......>  fem::map_entities(entity_maps_as_array, entities);
    fem::pack_coefficients(c, mesh, entities, e.entity_maps(), coffsets,
                           std::span(coeffs));
  }
  std::vector<T> constants = fem::pack_constants(e);

  tabulate_expression<T, U>(
      values, e, md::mdspan(coeffs.data(), entities.size(), cstride),
      std::span<const T>(constants), mesh, entities, element);
```

cpp/demo/hyperelasticity/main.cpp
jorgensd: Not sure how I'm supposed to refactor this when we  do not attach entity maps to coefficients. As stated further down, we can refactor pack_coefficients to work on mapped entities, but you still need to pass the maps here.

cpp/dolfinx/fem/pack.h
jorgensd: @garth-wells I've now refactored this.

cpp/dolfinx/fem/Expression.h
jorgensd: Not sure how this refactor should be done. Please have a look at the current, refactored PR that at least simplifies the logic in `pack_coefficients`.

jorgensd:
CI failing due to global Github issues:
Update - Actions is experiencing degraded performance. We are continuing to investigate.
May 26, 2026 - 12:17 UTC
Update - We are investigating authentication issues leading to failure in starting Actions runs and downloading actions. At this time the majority of Actions runs is impacted.
May 26, 2026 - 11:53 UTC
Update - Actions is experiencing degraded availability. We are continuing to investigate.
May 26, 2026 - 11:19 UTC

cpp/dolfinx/fem/Expression.h
```diff
@@ -189,5 +201,13 @@ class Expression
 
   // Evaluation points on reference cell
   std::pair<std::vector<geometry_type>, std::array<std::size_t, 2>> _x_ref;
+
+  // Set of bidirectional maps of mesh entities (cells, facets, ridges or peaks)
+  // between meshes and submeshes of any codimension.
```
jorgensd:
```suggestion
  // between meshes and submeshes of any codimension
```

cpp/dolfinx/fem/pack.h
```diff
@@ -329,20 +330,131 @@ void pack_coefficients(const Form<T, U>& form,
...
+      std::vector<std::int32_t> contiguous_cells(cells.extent(0));
+      for (std::size_t i = 0; i < cells.extent(0); ++i)
+        contiguous_cells[i] = cells(i);
+      return contiguous_cells;
+    }
+  }
+  else
+  {
+    assert(entity_map.has_value());
+    const mesh::Topology topology = *mesh.topology();
```
schnellerhase:
```suggestion
    const mesh::Topology& topology = *mesh.topology();
```
Won't this copy?

jhale (approved):

cpp/dolfinx/fem/pack.h
```diff
@@ -329,20 +330,131 @@ void pack_coefficients(const Form<T, U>& form,
...
+  auto mesh_c = coeff.function_space()->mesh();
+  assert(mesh_c);
+
+  if (mesh_c->topology() == mesh.topology())
+  {
+    // If same mesh no mapping is needed
+    if constexpr (entities.rank() == 1)
+    {
+      return std::vector<std::int32_t>(
+          entities.data_handle(), entities.data_handle() + entities.size());
```
schnellerhase: Is this stride safe, i.e. what happens if entities is already a `submdspan`?

cpp/dolfinx/fem/pack.h
```diff
@@ -329,20 +330,131 @@ void pack_coefficients(const Form<T, U>& form,
...
+    std::vector<std::int32_t> e_b;
+    const dolfinx::mesh::EntityMap& emap = entity_map.value().get();
+    bool inverse = emap.sub_topology() == mesh_c->topology();
+    // If cells are supplied on the parent mesh, we can directly map them to
+    // cells on the coefficient mesh.
+    if constexpr (entities.rank() == 1)
+    {
+      assert(codim == 0);
+      e_b = emap.sub_topology_to_topology(
+          std::span(entities.data_handle(), entities.size()), inverse);
```
schnellerhase:
```suggestion
      return emap.sub_topology_to_topology(
          std::span(entities.data_handle(), entities.size()), inverse);
```

The control flow is hard to read. Can we change to `if () {return ... }` and avoid redundant `else` case indentations throughout to simplify? Exceptions already follow this approach.

python/test/unit/fem/test_expression.py
```diff
@@ -526,3 +526,112 @@ def test_rank1_blocked():
...
+    V = dolfinx.fem.functionspace(mesh, el)
+    u = dolfinx.fem.Function(V, dtype=dtype)
+    u.interpolate(lambda x: x[0] + 2.0 * x[1])
+
+    tol = 50 * np.finfo(xtype).resolution
+
+    def mark_left_facets(x):
+        return np.isclose(x[0], 1.0, atol=tol)
+
+    left_facets = dolfinx.mesh.locate_entities(mesh, mesh.topology.dim - 1, mark_left_facets)
```
schnellerhase: This is an empty list. That is not desired for the test, right? Should the marker before be `np.isclose(x[0], 0.0, atol=tol)`?

cpp/dolfinx/fem/assembler.h
```diff
@@ -135,7 +135,7 @@ void tabulate_expression(std::span<T> values, const fem::Expression<T, U>& e,
   std::vector<T> constants = fem::pack_constants(e);
 
   tabulate_expression<T, U>(
-      values, e, md::mdspan(coeffs.data(), entities.size(), cstride),
+      values, e, md::mdspan(coeffs.data(), entities.extent(0), cstride),
```
schnellerhase, line 138: @jorgensd need some 👀 . Appears to be unrelated to the test issues though.

jorgensd:
@schnellerhase  and I got further. Now we are missing a permutation to compare the output of the two approaches, as they produce the same output, but in a permuted order..

jorgensd:
Related issue: https://github.com/FEniCS/ffcx/pull/836

.github/workflows/fenicsx-refs.env
```diff
@@ -3,4 +3,4 @@ basix_ref=main
 ufl_repository=FEniCS/ufl
 ufl_ref=main
 ffcx_repository=FEniCS/ffcx
-ffcx_ref=main
+ffcx_ref=dokken/facet_expression_parent_fix
```
```suggestion
ffcx_ref=main
```

schnellerhase (approved):

PR #4149, opened 2026-04-14
Remove `column_index_map` from SparsityPattern

chrisrichardson:
Remove `column_index_map` from SparsityPattern, just use `index_map(1)` instead after finalisation.


jorgensd (approved):
This seems sensible to me, as I see no use for the original index map that is lacking the additional ghosts.

chrisrichardson:
If user needs the original indexmap for some reason, it is up to them to copy it beforehand.

PR #4152, opened 2026-04-17
Draft PR Add Python demo for smoothed TV inpainting

### LINKED ISSUE #4137: Add Python demo for variational image inpainting with smoothed TV regularization on irregular masks
Hi all, 

I would like to propose a new Python demo for DOLFINx showing variational image inpainting on an irregular mask using a smoothed total variation regularization. This demo would set up the inpainting problem as a nonlinear variational problem. The fidelity term on the known part of the image together with a smoothed TV regularization term on the full domain. It would derive the week form and implement the residual and Jacobian in UFL. For geometry, the plan is to use the Gmsh Python interface to build a conforming unstructured mesh of the full image domain, with the irregular mask boundary represented inside the domain. That would make it possible to tag the known and unknown regions and apply the fidelity term only where image data is available. The solve would use dolfinx.fem.petsc.NonlinearProblem together with PETSc SNES / Newton iteration. I would aim to keep the example self-contained, probably using a synthetic image and mask generated directly in code. I think this could be a useful demo because it combines several pieces of DOLFINx functionality in one example. 

If it fits for this project, my assumption is that it would go in the Python demo collection under python/demo/ as a new demo file. A likely scope would be:

- formulate the smoothed TV inpainting problem
- derive the weak form
- generate a conforming mesh with an irregular interior mask/interface
- tag the relevant regions
- implement the nonlinear problem in UFL
- solve it with NonlinearProblem and SNES
- show basic reconstruction results and diagnostics
- have documentation

I have tried to use the Slack link mentioned in the project resources, but it said the link was no longer active. I'm opening this issue first to check whether this would be a welcome contribution and whether you would prefer it as a core demo or as an external tutorial. 

Sincerely,

Jack Brockett

jb2178-star:
## Summary

This PR adds a minimal Python demo for for variational image inpainting with smoothed total variation regularization on a synthetic image with an irregular interior mask, as this demonstrates the use of FEM over Finite Difference methods.

Starting with a variational inpainting model with the spirit of [Chan, Shen](https://epubs.siam.org/doi/abs/10.1137/S0036139900368844?) we have J(u)= data fit + regularization term, we use the TV term from [Rudin, Osher, Fatemi](https://www.sciencedirect.com/science/article/abs/pii/016727899290242F) :

$$ \int_{\Omega}|\nabla u| \mathrm{d}x$$

And make it smooth:

$$ \int_{\Omega}\sqrt{|\nabla u|^2 +\varepsilon^2}$$

and the data fitted term, from standard masked least squares (m=1 on known, m=0 on missing data):

$${1\over 2}\int_{\Omega}m(x)(u-f)^2\mathrm{d}x $$

Due to the fact $|\nabla u|$ is not differentiable when $\nabla u=0$, and its hard to use newton's method, so we introduce a small $\varepsilon$ term. 

When applying Euler-Lagrange to minimize $J(u)$ we get:

$$ m(u-f )-\alpha \nabla \cdot({\nabla u\over \sqrt{|\nabla u|^2 +\varepsilon^2}})=0$$

And with our test function $v$ we can arrive at the weak form:

$$\int_{\Omega}m(u-f)v\mathrm{d}x+\alpha \int_{\Omega}{\nabla u \cdot \nabla v \over \sqrt{|\nabla u|^2+\varepsilon^2}}\mathrm{d}x=0 $$

this current demo adds a unit-square mesh generated with DOLFINx, a synthetic scalar image, an irregular mask, a nonlinear variational formulation in UFL, a Jacobian generated with ufl.derivative and a nonlinear solve using dolfinx.fem.petsc.NonlinearProblem

## Notes

This is intended as a rough draft of the demo. I do intend to include plotting/diagnostics and more advanced geometry in a follow-up revision if that makes sense. Additionally more defined comments and documentation.

### DIFF
--- a/python/demo/demo_smoothed_tv_inpainting.py
+++ b/python/demo/demo_smoothed_tv_inpainting.py
@@ -15,7 +15,7 @@
 # * {download}`Python script <./demo_smoothed_tv_inpainting.py>`
 # * {download}`Jupyter notebook <./demo_smoothed_tv_inpainting.ipynb>`
 # ```
-#
+# Rough draft notes
 # This demo solves a variational image inpainting problem on the unit square.
 # A synthetic image is masked on an irregular interior region, and the missing
 # values are reconstructed using smoothed total variation regularization.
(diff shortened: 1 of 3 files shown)


jb2178-star:
Added plotting with matplotlib for comparison of fields (u_true, mask, f, u, and difference). Matplotlib was chosen here for clarity in comparing multiple scalar fields. PyVista could be used for alternative visualization if preferred. Additionally added diagnostics for data fidelity, TV seminorm and reconstruction error.

jb2178-star:
Hi all,

I believe this PR is now ready for review. I have added clearer and more detailed documentation to better align with the style used in dolfinx. I can also include a more complete derivation of the weak form if that would be helpful. I have also added the nonlinear solver metrics such as:

- initial vs final objective (optimization progress)
- number of iterations of nonlinear iterations
- final residual norm

As well as reconstruction metrics:

- L2 error
- hole error (error only in the missing regions)
- PSNR (peak signal to noise ratio)
- H1 seminorm error
- 
I have also improved the visualization by including both global error and the hole only error plots.

This example demonstrates how to use dolfinx for nonlinear variational problems that may arise in image processing. In particular it demonstrates:

 - formulation of an energy function using UFL
 - differentiation of the jacobian via ufl.derviative
 - solution of the resulting nonlinear system using the integrated PETSc SNES
 - FEM assembly and discretization on a triangular mesh

This highlights dolfinx's ability to handle nonlinear, non-quadratic variational problems in a clear and concise way. 

Thanks to @mscroggs for linking this PR to the relevant issue, and to @jorgensd for reviewing the original issue and providing feedback. 

If there are any further suggestions or improvements, please let me know. Otherwise I believe this is now a complete documented inpainting demo ready for review.

mscroggs:
The CI lint checks are picking up some ruff warnings. You can fix many of these by running `ruff check --fix` - this command should also display the other warnings that need fixing to get the green tick

jorgensd:
@jb2178-star I've tried to do a major refactoring of the text (to make sure the docs render) and simplify the text a bit.
A follow-up question is: why not use `ufl.derivative` for getting the `F`?

jb2178-star:
Hi @jorgensd ,

I really appreciate you for refactoring , simplifying the text and taking the time to review this. It has been a really valuable learning experience for me, as this is my first time contributing to an open source project.

I initially wrote F with the intention to show the weak form explicitly. However I understand that using ufl.derivative to obtain the weak form directly is cleaner and more concise, and I'm happy to refactor it if you think that would be preferable.

PR #4153, opened 2026-04-17
MatrixCSR enhancements

chrisrichardson:
Adds the following to `MatrixCSR`:

* transpose multiply `y = A^T x`
* matrix multiply `C = A * B`

Matrix multiply is in a separate header file, as a free function, so that it can be included only when needed.

Prepared with assistance from Claude AI, but all code also checked manually.

### DIFF
--- a/python/dolfinx/la/__init__.py
+++ b/python/dolfinx/la/__init__.py
@@ -177,7 +177,7 @@ class MatrixCSR:
         """
         if self.index_map(1).size_local != B.index_map(0).size_local:
             raise RuntimeError("Invalid matrix sizes for matmult.")
-        return self._cpp_object.mult(B._cpp_object)
+        return MatrixCSR(self._cpp_object.mult(B._cpp_object))
 
     @property
     def block_size(self) -> list:
--- a/python/dolfinx/la/__init__.py
+++ b/python/dolfinx/la/__init__.py
@@ -168,6 +168,17 @@ class MatrixCSR:
         else:
             self._cpp_object.mult(x._cpp_object, y._cpp_object)
 
+    def matmul(self, B):
+        """Compute ``C = A * B``, where `A` is this matrix.
+
+        Args:
+            B: Input Matrix
+            C: Output Matrix
+        """
+        if self.index_map(1).size_local != B.index_map(0).size_local:
+            raise RuntimeError("Invalid matrix sizes for matmult.")
+        return self._cpp_object.mult(B._cpp_object)
+
     @property
     def block_size(self) -> list:
         """Block sizes for the matrix."""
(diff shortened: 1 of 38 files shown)


python/dolfinx/la/__init__.py
```diff
@@ -155,14 +155,29 @@ def index_map(self, i: int) -> IndexMap:
         """
         return self._cpp_object.index_map(i)
 
-    def mult(self, x: Vector, y: Vector) -> None:
-        """Compute ``y += Ax``.
+    def mult(self, x: Vector, y: Vector, transpose: bool = False) -> None:
```
jhale, line 158: Might just want to go with two methods mult and mult_trans?

python/test/unit/la/test_matmul.py
```diff
@@ -0,0 +1,60 @@
...
+def test_matmul(dtype, mat_gather):
+    mesh = create_unit_square(MPI.COMM_WORLD, 3, 3)
+    imap = mesh.topology.index_map(0)
+    sp = _cpp.la.SparsityPattern(mesh.comm, [imap, imap], [1, 1])
+    rows = np.arange(0, imap.size_local)
+    cols = np.arange(0, imap.size_local + imap.num_ghosts)
+    sp.insert(rows, cols)
+    sp.finalize()
+
+    # Identity
```
jhale: Do these methods check functioning of the off-diagonals if only the Identity matrix is being used?

If you used a tridiagonal you can also have an explicit expression to check against.

python/dolfinx/la/__init__.py
```diff
@@ -155,14 +155,29 @@ def index_map(self, i: int) -> IndexMap:
...
+            self._cpp_object.multT(x._cpp_object, y._cpp_object)
+        else:
+            self._cpp_object.mult(x._cpp_object, y._cpp_object)
+
+    def matmul(self, B):
+        """Compute ``C = A * B``, where `A` is this matrix.
+
+        Args:
+            B: Input Matrix
+            C: Output Matrix
```
jhale: Missing?

python/test/unit/la/test_matrix_vector.py
```diff
@@ -53,27 +53,78 @@ def test_create_matrix_csr():
...
+    "dtype",
+    [
+        np.float32,
+        np.float64,
+        np.complex64,
+        np.complex128,
+    ],
+)
+def test_matvec_transpose(dtype, mat_gather):
+    mesh = create_unit_square(MPI.COMM_WORLD, 3, 3)
```
jhale: Place mat generation into private function for reuse?

python/test/unit/la/test_matmul.py
chrisrichardson: To be honest, I have no idea why the comment `# Identity` is there. The tests use full (dense) matrices.

python/dolfinx/la/__init__.py
chrisrichardson: leaving for now

Issue #4141, opened 2026-04-01
Add Python demo for adaptive finite element method for linear problems


brodbeck-m:
## Aims and Scope
FEniCSx by default supports most of the relevant parts for the design of adaptive finite element algorithms, such as mesh-independent form compilation, evaluating the (cell-wise) errors, refining the mesh and resolving the problem. Nevertheless, a comprehensive demo combining these features, as this was done in the [legacy version](https://olddocs.fenicsproject.org/dolfin/latest/python/demos/auto-adaptive-poisson/demo_auto-adaptive-poisson.py.html) as well as performant implementations of cell marking and interaction with (facet-based) refinement routines, is missing. This is also not covered by external demos including [fenicsx-error-estimation](https://github.com/jhale/fenicsx-error-estimation) or the [FEniCSx tutorials](https://jsdokken.com/dolfinx-tutorial/chapter2/amr.html)

## Planed features
As nearly every relevant error estimator is initially introduced for the Poisson problem, the demo will focus on the adaptive solution of a Poisson problem in both primal as well as mixed form. In order to  keep the complexity of the actual error estimation low, residual-type error estimators (see e.g. Carstensen [1](https://doi.org/10.1016/j.camwa.2013.12.003)) are considered. The classical L-shaped domain or a unit square with varying coefficient [2](https://doi.org/10.1016/S0898-1221(03)90086-1) could serve as test problem.

## Prerequisites
For the marking routines we suggest to implement in C++:
- ~Convenience functionality to identify the closure of topological entities based on their connectivity to neighboring entities of different topological dimensions. This allows the conversion from cell-wise markers to an edge list or alike.~
- A scalable implementation of the classical maximum strategy [3](https://doi.org/10.1007/BF01389757) and Doerfler marking [4](https://doi.org/10.1137/0733054) to demonstrate their different behavior

jorgensd:
> Convenience functionality to identify the closure of topological entities based on their connectivity to neighboring entities of different topological dimensions. This allows the conversion from cell-wise markers to an edge list or alike.

Isn't this exactly what `dolfinx.mesh.compute_incident_entities` does for a collection of entities (https://docs.fenicsproject.org/dolfinx/v0.10.0.post5/python/generated/dolfinx.mesh.html#dolfinx.mesh.compute_incident_entities)
[...]

PR #4156, opened 2026-04-18
Add maximum marking

schnellerhase:
Implements a maximum marking criterion for AFEM schemes, i.e. for $\theta \in (0, 1)$ and a marker $\eta \in \mathbb{R}^N$, computes

$$
  \mathcal{I} = \left\\{ i \ | \ \eta_i > \theta \max_j \eta_j \right\\}.
$$

In particular this introduces/defines the interface for other marking routines (to follow in the future): based on a process local list of (entity-)indicators the (local) marked index set of it is computed.

<details>

<summary>
Test program to visualise results (parallel ready).
</summary>

```python
from mpi4py import MPI

import numpy as np

from dolfinx import mesh, plot
import pyvista as pv

mesh_0 = mesh.create_unit_square(comm := MPI.COMM_WORLD, n := 10, n)

cell_im = mesh_0.topology.index_map(tdim := mesh_0.topology.dim)
cell_count = cell_im.size_local + cell_im.num_ghosts

marker = np.random.random(cell_count)
marked_cells = mesh.mark_maximum(marker, theta := 0.5, comm)

mesh_0.topology.create_entities(1)
marked_edges = mesh.compute_incident_entities(mesh_0.topology, marked_cells, tdim, 1)

mesh_1, _, _ = mesh.refine(mesh_0, marked_edges)

grid_0 = pv.UnstructuredGrid(*plot.vtk_mesh(mesh_0))
grid_1 = pv.UnstructuredGrid(*plot.vtk_mesh(mesh_1))

plotter = pv.Plotter(shape=(1, 2))
plotter.subplot(0, 0)
plotter.add_mesh(grid_0, show_edges=True)
plotter.view_xy()
plotter.add_text("mesh_0", position="upper_left")

plotter.subplot(0, 1)
plotter.add_mesh(grid_1, show_edges=True)
plotter.view_xy()
plotter.add_text("mesh_1", position="upper_left")

plotter.link_views()
plotter.show()

```
</details>

Work towards https://github.com/FEniCS/dolfinx/issues/4141.

### DIFF
--- a/cpp/dolfinx/refinement/mark.h
+++ b/cpp/dolfinx/refinement/mark.h
@@ -1,0 +1,59 @@
+// Copyright (C) 2026 Paul T. Kühner
+//
+// This file is part of DOLFINX (https://www.fenicsproject.org)
+//
+// SPDX-License-Identifier:    LGPL-3.0-or-later
+
+#pragma once
+
+#include <algorithm>
+#include <cassert>
+#include <concepts>
+#include <cstdint>
+#include <iterator>
+#include <limits>
+#include <mpi.h>
+#include <spdlog/spdlog.h>
+#include <vector>
+
+#include "dolfinx/common/MPI.h"
+
+namespace dolfinx::refinement
+{
+
+/// @brief Maximum marking of a marker.
+///
+/// @param[in] marker Input marker (local) - usually an error indicator per
+/// entity
+/// @param[in] theta Cut off parameter, 0 ≤ θ ≤ 1
+/// @param[in] comm Communicator over which the maximum is computed.
+/// @return Indices (local) of marker elements, which satisfy: marker_i ≥ θ
+/// max(marker).
+template <std::floating_point T>
+std::vector<std::int32_t> mark_maximum(std::span<const T> marker, T theta,
+                                       MPI_Comm comm)
+{
+  assert((0 <= theta) && (theta <= 1));
+
+  T max = marker.empty() ? std::numeric_limits<T>::min()
+                         : std::ranges::max(marker);
+  MPI_Allreduce(MPI_IN_PLACE, &max, 1, dolfinx::MPI::mpi_t<T>, MPI_MAX, comm);
+
+  auto mark = [=](auto e) { return e >= theta * max; };
+
+  std::vector<std::int32_t> indices;
+  indices.reserve(std::ranges::count_if(marker, mark));
+
+  for (std::int32_t i = 0; i < static_cast<std::int32_t>(marker.size()); ++i)
+  {
+    if (mark(marker[i]))
+      indices.push_back(i);
+  }
+
+  spdlog::info("Marking (max) {} / {} (local) entities.", indices.size(),
+               marker.size());
+
+  return indices;
+}
+
+} // namespace dolfinx::refinement
--- a/cpp/dolfinx/refinement/mark.h
+++ b/cpp/dolfinx/refinement/mark.h
@@ -33,7 +33,8 @@ template <std::floating_point T>
 std::vector<std::int32_t> mark_maximum(std::span<const T> marker, T theta,
                                        MPI_Comm comm)
 {
-  assert((0 <= theta) && (theta <= 1));
+  if ((theta < 0) || (theta > 1))
+    throw std::invalid_argument("Theta needs to fullfill 0 ≤ θ ≤ 1.");
 
   T max = marker.empty() ? std::numeric_limits<T>::min()
                          : std::ranges::max(marker);
--- a/cpp/dolfinx/refinement/mark.h
+++ b/cpp/dolfinx/refinement/mark.h
@@ -10,7 +10,6 @@
 #include <cassert>
 #include <concepts>
 #include <cstdint>
-#include <iterator>
 #include <limits>
 #include <mpi.h>
 #include <spdlog/spdlog.h>
@@ -25,7 +24,7 @@ namespace dolfinx::refinement
 ///
 /// @param[in] marker Input marker (local) - usually an error indicator per
 /// entity
-/// @param[in] theta Cut off parameter, 0 ≤ θ ≤ 1
+/// @param[in] theta Cut off parameter, 0 < θ < 1
 /// @param[in] comm Communicator over which the maximum is computed.
 /// @return Indices (local) of marker elements, which satisfy: marker_i ≥ θ
 /// max(marker).
@@ -33,15 +32,14 @@ template <std::floating_point T>
 std::vector<std::int32_t> mark_maximum(std::span<const T> marker, T theta,
                                        MPI_Comm comm)
 {
-  if ((theta < 0) || (theta > 1))
-    throw std::invalid_argument("Theta needs to fullfill 0 ≤ θ ≤ 1.");
+  if ((theta <= 0) || (theta >= 1))
+    throw std::invalid_argument("Theta needs to fullfill 0 < θ < 1.");
 
   T max = marker.empty() ? std::numeric_limits<T>::lowest()
                          : std::ranges::max(marker);
   MPI_Allreduce(MPI_IN_PLACE, &max, 1, dolfinx::MPI::mpi_t<T>, MPI_MAX, comm);
 
-  auto mark = [=](T e)
-  { return e + std::numeric_limits<T>::epsilon() * 1e2 > theta * max; };
+  auto mark = [=](T e) { return e > theta * max; };
 
   std::vector<std::int32_t> indices;
   indices.reserve(std::ranges::count_if(marker, mark));
--- a/cpp/dolfinx/refinement/mark.h
+++ b/cpp/dolfinx/refinement/mark.h
@@ -40,7 +40,8 @@ std::vector<std::int32_t> mark_maximum(std::span<const T> marker, T theta,
                          : std::ranges::max(marker);
   MPI_Allreduce(MPI_IN_PLACE, &max, 1, dolfinx::MPI::mpi_t<T>, MPI_MAX, comm);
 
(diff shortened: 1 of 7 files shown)


cpp/dolfinx/refinement/mark.h
```diff
@@ -0,0 +1,59 @@
...
+/// entity
+/// @param[in] theta Cut off parameter, 0 ≤ θ ≤ 1
+/// @param[in] comm Communicator over which the maximum is computed.
+/// @return Indices (local) of marker elements, which satisfy: marker_i ≥ θ
+/// max(marker).
+template <std::floating_point T>
+std::vector<std::int32_t> mark_maximum(std::span<const T> marker, T theta,
+                                       MPI_Comm comm)
+{
+  assert((0 <= theta) && (theta <= 1));
```
jhale: Cheap and useful check - should always throw std::value_error

cpp/dolfinx/refinement/mark.h
```diff
@@ -0,0 +1,59 @@
...
+  T max = marker.empty() ? std::numeric_limits<T>::min()
+                         : std::ranges::max(marker);
+  MPI_Allreduce(MPI_IN_PLACE, &max, 1, dolfinx::MPI::mpi_t<T>, MPI_MAX, comm);
+
+  auto mark = [=](auto e) { return e >= theta * max; };
+
+  std::vector<std::int32_t> indices;
+  indices.reserve(std::ranges::count_if(marker, mark));
+
+  for (std::int32_t i = 0; i < static_cast<std::int32_t>(marker.size()); ++i)
```
jhale, line 48: Is there a std library way to do this loop?

cpp/dolfinx/refinement/mark.h
```diff
@@ -0,0 +1,59 @@
...
+std::vector<std::int32_t> mark_maximum(std::span<const T> marker, T theta,
+                                       MPI_Comm comm)
+{
+  assert((0 <= theta) && (theta <= 1));
+
+  T max = marker.empty() ? std::numeric_limits<T>::min()
+                         : std::ranges::max(marker);
+  MPI_Allreduce(MPI_IN_PLACE, &max, 1, dolfinx::MPI::mpi_t<T>, MPI_MAX, comm);
+
+  auto mark = [=](auto e) { return e >= theta * max; };
```
jhale: Use const T instead of auto e?

Also remove inequality, pointless for floating point arithmetic.

jhale (dismissed):

cpp/dolfinx/refinement/mark.h
schnellerhase: Yes, but only in C++23 https://en.cppreference.com/cpp/ranges/enumerate_view.

cpp/dolfinx/refinement/mark.h
schnellerhase: Changed to `T` - think we do not follow const correctness for trivial types.

schnellerhase: > Also remove inequality, pointless for floating point arithmetic.

Don't think this is correct: for $\theta=1.0$ and $\eta = [ 1, \dots, 1]$ that would then mark nothing (tested).

cpp/dolfinx/refinement/mark.h
```diff
@@ -0,0 +1,60 @@
...
+/// @return Indices (local) of marker elements, which satisfy: marker_i ≥ θ
+/// max(marker).
+template <std::floating_point T>
+std::vector<std::int32_t> mark_maximum(std::span<const T> marker, T theta,
+                                       MPI_Comm comm)
+{
+  if ((theta < 0) || (theta > 1))
+    throw std::invalid_argument("Theta needs to fullfill 0 ≤ θ ≤ 1.");
+
+  T max = marker.empty() ? std::numeric_limits<T>::min()
```
chrisrichardson: Does this make sense? Shouldn't it just be zero?

cpp/dolfinx/refinement/mark.h
```diff
@@ -0,0 +1,60 @@
...
+                         : std::ranges::max(marker);
+  MPI_Allreduce(MPI_IN_PLACE, &max, 1, dolfinx::MPI::mpi_t<T>, MPI_MAX, comm);
+
+  auto mark = [=](T e) { return e >= theta * max; };
```
chrisrichardson, line 43:
```suggestion
  auto mark = [=](T e) { return e > theta * max; };
```

cpp/dolfinx/refinement/mark.h
schnellerhase: We do not assume the marker to be positive, so smaller than zero values are admissible in the marker, therefore 0 would not allow for the correct max value being identified in the all reduce.

cpp/dolfinx/refinement/mark.h
schnellerhase: Ref https://github.com/FEniCS/dolfinx/pull/4156#discussion_r3109060529.

cpp/dolfinx/refinement/mark.h
chrisrichardson: It depends what you are using for `T` - but for most floating point types, this will just be a small positive number:
https://en.cppreference.com/cpp/types/numeric_limits/min

schnellerhase: Right, should be `lowest` - thanks!

chrisrichardson (approved):

jhale:
My comments on the floating point comparisons are not resolved.

cpp/dolfinx/refinement/mark.h
jhale: And through what astronomically unlikely event would eta be binary 1.0 for every cell?

chrisrichardson:
@jhale - I think @schnellerhase did address it. Did you mean that it should be a simple `>` rather than `>=`?

jhale:
Yes - the equality is completely redundant in a floating point context and the example given is a contrived one.

cpp/dolfinx/refinement/mark.h
schnellerhase: Unlikely but valid input to the function. We can either leave as is or change to $e + \epsilon > \theta \max \eta$ for some close to machine epsilon $\epsilon$, which should ensure this edge case passes. Fine with both.

python/test/unit/refinement/test_mark.py
```diff
@@ -0,0 +1,33 @@
...
+    marked_cells = mesh.mark_maximum(marker, theta, comm)
+
+    assert np.allclose(
+        marked_cells,
+        np.argwhere(marker >= theta * comm.allreduce(np.max(marker), MPI.MAX)).flatten(),
+    )
+
+    msh.topology.create_entities(1)
+    marked_edges = mesh.compute_incident_entities(msh.topology, marked_cells, tdim, 1)
+    mesh.refine(msh, marked_edges)
```
jorgensd, line 33: Should there be something more at the end of this test ?

schnellerhase: I just wanted to check that the input works as expected with the refine calls - so the test being "refine works". Not sure what to exactly test for on the refined mesh, up to it being a mesh.

cpp/dolfinx/refinement/mark.h
```diff
@@ -0,0 +1,61 @@
...
+{
+  if ((theta < 0) || (theta > 1))
+    throw std::invalid_argument("Theta needs to fullfill 0 ≤ θ ≤ 1.");
+
+  T max = marker.empty() ? std::numeric_limits<T>::lowest()
+                         : std::ranges::max(marker);
+  MPI_Allreduce(MPI_IN_PLACE, &max, 1, dolfinx::MPI::mpi_t<T>, MPI_MAX, comm);
+
+  auto mark = [=](T e)
+  { return e + std::numeric_limits<T>::epsilon() * 1e2 > theta * max; };
```
jhale: You're overthinking this.

Why use epsilon around 1 when e and max can vary across many orders of magnitude? Why 1e-2? Why systematically favor inclusion over exclusion?

Why does this matter in the context of mesh refinement where small perturbations in the marked set make no difference?

jhale (changes requested):

cpp/dolfinx/refinement/mark.h
schnellerhase: Since the maximum is attained by a value of the marker, we need inclusion. Otherwise there is no guarantee that an element gets marked for $\theta = 1$.

This is now a less than check with abs. tol of 1e2 * machine eps tolerance to make this check as tight as possible while allowing some reasonable leeway for range differences. Can adjust tolerances or change to relative check.

schnellerhase: I think if we restrict to $\theta \in (0,1)$ we can stick to strict inequality checks.

jhale: I agree - the endpoint at 1.0 is best done through the obvious approach (refine everything).

python/test/unit/refinement/test_mark.py
```diff
@@ -0,0 +1,33 @@
...
+
+    tdim = msh.topology.dim
+    cell_count = (cell_im := msh.topology.index_map(tdim)).size_local + cell_im.num_ghosts
+    marker = np.random.default_rng(0).random(cell_count)
+
+    marked_cells = mesh.mark_maximum(marker, theta, comm)
+
+    assert np.allclose(
+        marked_cells,
+        np.argwhere(marker >= theta * comm.allreduce(np.max(marker), MPI.MAX)).flatten(),
```
schnellerhase:
```suggestion
        np.argwhere(marker > theta * comm.allreduce(np.max(marker), MPI.MAX)).flatten(),
```

jhale (approved):

PR #4158, opened 2026-04-20
Fix: adios2 bug

### LINKED ISSUE #4157: [BUG]: VTX write list of DG0 functions + other spaces does not work
### Summarize the issue

The VTX writer of `dolfinx` writes a list of DG0 + other functions to a .bp folder without errors but ParaView errors when it tries to read it.

ParaView believes the DG0 functions are point data when they should be cell data.

The feature was broken by afec7b1680d7454cfc9daebd88979a447d115b40 (2026-01-05).
It didn't break the test `test_dg_0_data`, as the test only writes the data but does
not verify it can be read.

TODO:
- Fix bug
- Add reading to the test.

I will make a PR later today if it hasn't been addressed.

### How to reproduce the bug

Add these two lines to `dolfinx/python/test/unit/io/test_adios2.py`:
```
if __name__=="__main__":
    TestVTX().test_dg_0_data(tempdir="out")
```
Run the file and attempt to read the bp folder in `out` with ParaView.

### Minimal Example (Python)

```Python

```

### Output (Python)

```bash
ParaView error:

ERROR: In vtkDataSet.cxx, line 708
vtkUnstructuredGrid (0x3615bb60): Point array q with 3 components, only has 50 tuples but there are 231 points

ERROR: In vtkPVGeometryFilter.cxx, line 877
vtkPVGeometryFilter (0x3911fa20): Input is not a vtkDataObjectTree.
```

### Version

main branch

### DOLFINx git commit

_No response_

### Installation

Install from source on Linux machine.

### Additional information

_No response_

schnellerhase:
https://github.com/FEniCS/dolfinx/pull/4034 messed up taking a copy over a reference. Made clearer with unique var naming now.

Fixes https://github.com/FEniCS/dolfinx/issues/4157.

### DIFF
--- a/cpp/dolfinx/io/ADIOS2Writers.h
+++ b/cpp/dolfinx/io/ADIOS2Writers.h
@@ -183,6 +183,7 @@ extract_function_names(const typename adios2_writer::U<T>& u)
               using U = std::decay_t<decltype(v)>;
               using X = typename U::element_type;
 
+              // append function names either to dg0_names or names
               auto& fnames
                   = impl::is_cellwise(*(v->function_space()->element()))
                         ? dg0_names
--- a/cpp/dolfinx/io/ADIOS2Writers.h
+++ b/cpp/dolfinx/io/ADIOS2Writers.h
@@ -183,15 +183,17 @@ extract_function_names(const typename adios2_writer::U<T>& u)
               using U = std::decay_t<decltype(v)>;
               using X = typename U::element_type;
 
-              if (impl::is_cellwise(*(v->function_space()->element())))
-                names = dg0_names;
+              auto& fnames
+                  = impl::is_cellwise(*(v->function_space()->element()))
+                        ? dg0_names
+                        : names;
 
               if constexpr (std::is_floating_point_v<typename X::value_type>)
-                names.push_back(v->name);
+                fnames.push_back(v->name);
               else
               {
-                names.push_back(v->name + impl_adios2::field_ext[0]);
-                names.push_back(v->name + impl_adios2::field_ext[1]);
+                fnames.push_back(v->name + impl_adios2::field_ext[0]);
+                fnames.push_back(v->name + impl_adios2::field_ext[1]);
               }
             },
             v);
--- a/python/test/unit/io/test_adios2.py
+++ b/python/test/unit/io/test_adios2.py
@@ -293,4 +293,4 @@ class TestVTX:
             assert "v" in point_arrays or ("v_imag" in point_arrays and "v_real" in point_arrays)
 
             cell_arrays = get_data_array_names(xml_cell_data)
-            assert "q" in cell_arrays or ("q_imag" in point_arrays and "q_real" in point_arrays)
+            assert "q" in cell_arrays or ("q_imag" in cell_arrays and "q_real" in cell_arrays)
--- a/python/test/unit/io/test_adios2.py
+++ b/python/test/unit/io/test_adios2.py
@@ -5,6 +5,7 @@
 # SPDX-License-Identifier:    LGPL-3.0-or-later
 
 from pathlib import Path
+from xml.etree import ElementTree as ET
 
 from mpi4py import MPI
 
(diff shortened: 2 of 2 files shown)


ordinary-slim:
The fix looks good to me. I was working on my side and I wrote it exactly the same way.

We can't properly read back the ouput but we can check that the xml has the correct array names in `PointData` and `CellData`. I would add this at the bottom of `test_adios2.py`

```python3
        # Check that q and v were written as point and cell data respectively
        with adios2.FileReader(str(filename), mesh.comm) as bpFile:
            vtk_data = bpFile.read_attribute_string("vtk.xml")

            from xml.etree import ElementTree as ET

            vtk_data = ET.fromstring(vtk_data).find("UnstructuredGrid/Piece")
            assert vtk_data is not None

            xml_point_data = vtk_data.find("PointData")
            xml_cell_data = vtk_data.find("CellData")
            assert xml_point_data is not None
            assert xml_cell_data is not None

            def get_data_array_names(data):
                return set(da.get("Name") for da in data.findall("DataArray"))

            point_arrays = get_data_array_names(xml_point_data)
            assert "v" in point_arrays

            cell_arrays = get_data_array_names(xml_cell_data)
            assert "q" in cell_arrays
```

It adds 20 lines of code to cover a corner case but I think it's still worth it.

schnellerhase:
Nice extension - can you push it into a branch (in a fork is fine)? Then I will happily merge this into this PR.

ordinary-slim:
> Nice extension - can you push it into a branch (in a fork is fine)? Then I will happily merge this into this PR.

You can find it in commit 3c553e0eedf1a6e8feaedaa8ed2416d65e71c980 of https://github.com/ordinary-slim/dolfinx/tree/main
EDIT: Wrote wrong commit hash first time

python/test/unit/io/test_adios2.py
```diff
@@ -272,3 +271,26 @@ def test_dg_0_data(self, tempdir):
...
+            assert xml_cell_data is not None
+
+            def get_data_array_names(data):
+                return set(da.get("Name") for da in data.findall("DataArray"))
+
+            point_arrays = get_data_array_names(xml_point_data)
+            assert "v" in point_arrays or ("v_imag" in point_arrays and "v_real" in point_arrays)
+
+            cell_arrays = get_data_array_names(xml_cell_data)
+            assert "q" in cell_arrays or ("q_imag" in point_arrays and "q_real" in point_arrays)
```
schnellerhase:
```suggestion
            assert "q" in cell_arrays or ("q_imag" in cell_arrays and "q_real" in cell_arrays)
```

python/test/unit/io/test_adios2.py
```diff
@@ -272,3 +271,26 @@ def test_dg_0_data(self, tempdir):
         q.interpolate(lambda x: (x[1], x[0]))
         writer.write(2)
         writer.close()
+
+        # Check that q and v were written as point and cell data respectively
+        with adios2.FileReader(str(filename), mesh.comm) as bpFile:
+            vtk_data = bpFile.read_attribute_string("vtk.xml")
+
+            from xml.etree import ElementTree as ET
```
jorgensd: Can we move this import to the top of the file?

cpp/dolfinx/io/ADIOS2Writers.h
```diff
@@ -183,15 +183,17 @@ extract_function_names(const typename adios2_writer::U<T>& u)
               using U = std::decay_t<decltype(v)>;
               using X = typename U::element_type;
 
-              if (impl::is_cellwise(*(v->function_space()->element())))
-                names = dg0_names;
+              auto& fnames
```
jorgensd, line 187: Maybe add a comment here about the fact that `fnames` now references either names or dg0_names which is used later to check for uniquenesss, as this code here is a bit hard on the eyes on first glance.

jorgensd (approved):
Only minor comment on import in test and regarding some extra comments

jorgensd (approved):

Issue #4045, opened 2026-01-23
PETSc error code checking


schnellerhase:
We do not perform consistent error code checking of the `PETSc` error codes returned by function invocations. Not dealing with the returned error code will become a compiler warning in the future, ref https://gitlab.com/petsc/petsc/-/blob/main/include/petscsystypes.h#L54-57.

Verifiable by configuring `PETSc` with `--with-strict-petscerrorcode` (`[[no-discard]]` marking of `PetscErrorCode`) and building with `-Werror`.

Siya-05:
Hello, Can I work on this?

schnellerhase:
Yes this has not been picked up yet, thanks!

PR #4160, opened 2026-04-20
Add missing PETSc/SLEPc error checks

Siya-05:
## Summary
Refers to #4045.

Add missing PETSc/SLEPc error checks for previously unchecked calls in `fem`, `la`, and `nls`.

## Changes
- add checks for `VecCreateNest` in `cpp/dolfinx/fem/petsc.cpp`
- add missing PETSc error checks in `cpp/dolfinx/la/petsc.cpp` using the existing `CHECK_ERROR(...)` pattern
- add checks for selected SLEPc/PETSc calls in `cpp/dolfinx/la/slepc.cpp`
- add checks for `VecNorm`, `VecAXPY`, and `MatCreateVecs` in `cpp/dolfinx/nls/NewtonSolver.cpp`

## Notes
- follow the existing per-file error-handling style
- do not change behavior except to check and report PETSc/SLEPc errors consistently


jhale:
This needs doing, but it would be good to drop into Slack #development to discuss precisely how it should be done.

Siya-05:
Sorry for being late on this,

While looking through the remaining PETSc-related files, I noticed that NewtonSolver.cpp and slepc.cpp also have a few unchecked PETSc calls. Unlike la/petsc.cpp, they don't have access to the local CHECK_ERROR macro.

So I updated the PETSc error handling in la/petsc.cpp by replacing remaining manual if (ierr != 0) petsc::error(...).

jhale:
I still see a number of `if (ierr != 0) petsc::error(...)` which need to be replaced with `CHECK_ERROR`.

jhale:
If the macro is an issue, just reimplement the macro where you need it?

Siya-05:
@jhale I've reimplemented the CHECK_ERROR macro where needed and updated the remaining manual PETSc/SLEPc error checks to use it consistently.

When you have a chance, could you please take another look? Thanks!

Siya-05:
@jhale @schnellerhase , I fixed all the errors. Can you check if I have to make any further changes?

PR #4167, opened 2026-04-21
Add mixed-topology support to `locate_dofs_topological`

jpdean:
This PR adds mixed-topology support to`locate_dofs_topological`.


cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -26,55 +26,107 @@ namespace
...
+    {
+      c_to_e_connectivities[i]
+          = topology.connectivity({tdim, i}, {dim, entity_type_index});
+
+      // NOTE: If e_to_c exists, c_to_e should also exist.
+      if (!c_to_e_connectivities[i])
+      {
+        throw std::runtime_error("Cell-to-entity connectivity of dimension "
+                                 + std::to_string(tdim) + "->"
+                                 + std::to_string(dim) + " doesn't exist");
```
jorgensd, line 74:
```suggestion
        throw std::runtime_error(std::format("Cell-to-entity connectivity of dimension {}->{} doesn't exist", tdim, dim));
```

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -26,55 +26,107 @@ namespace
...
 /// @param[in] entities The list of entities
 /// @param[in] dim The dimension of the entities
-/// @returns A list of `(cell_index, entity_index)` pairs for each input
-/// entity.
-std::vector<std::pair<std::int32_t, int>>
+/// @param[in] entity_type_index The index of the entity type in
+/// `topology.entity_types(dim)`
+/// @returns A list of `(cell_index, entity_index, cell_type_index)` tuples
+/// for each input entity. `cell_type_index` is the index of the cell type in
+/// `topology.cell_types()`.
```
jorgensd, line 36: Heads up, @chrisrichardson is rewriting all of these to have `i` inputs, as in: https://github.com/FEniCS/dolfinx/pull/4169/

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -26,55 +26,107 @@ namespace
...
 find_local_entity_index(const mesh::Topology& topology,
-                        std::span<const std::int32_t> entities, int dim)
+                        std::span<const std::int32_t> entities, int dim,
+                        int entity_type_index)
 {
   // Initialise entity-cell connectivity
   const int tdim = topology.dim();
-  auto e_to_c = topology.connectivity(dim, tdim);
-  if (!e_to_c)
+  std::vector<mesh::CellType> cell_types = topology.cell_types();
```
jorgensd, line 46:
```suggestion
  const std::vector<mesh::CellType>& cell_types = topology.cell_types();
```

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -26,55 +26,107 @@ namespace
...
+    // was a triangle
+    e_to_c_connectivities[i]
+        = topology.connectivity({dim, entity_type_index}, {tdim, i});
+
+    if (e_to_c_connectivities[i])
+    {
+      c_to_e_connectivities[i]
+          = topology.connectivity({tdim, i}, {dim, entity_type_index});
+
+      // NOTE: If e_to_c exists, c_to_e should also exist.
```
jorgensd, line 69: Is this assumption always true? I didn't think that was the case, i.e. (tdim, 0) exists at construction of a topology, but (0, tdim) does not.

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -26,55 +26,107 @@ namespace
...
   const int tdim = topology.dim();
-  auto e_to_c = topology.connectivity(dim, tdim);
-  if (!e_to_c)
+  std::vector<mesh::CellType> cell_types = topology.cell_types();
+  const int num_cell_types = cell_types.size();
+
+  // Get connectivities for each cell type.
+  std::int32_t num_entities_local = 0;
+  bool any_connectivity = false;
+  std::vector<std::shared_ptr<const dolfinx::graph::AdjacencyList<int32_t>>>
```
jorgensd, line 52: Should we rather use a `reserve` and pushback?

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -26,55 +26,107 @@ namespace
...
-    assert(it != entities_d.end());
-    std::size_t entity_local_index = std::distance(entities_d.begin(), it);
-    entity_indices.emplace_back(cell, entity_local_index);
+
+    bool entity_found = false;
+    for (int i = 0; i < num_cell_types; ++i)
+    {
+      auto e_to_c = e_to_c_connectivities[i];
+
+      if (e_to_c and e_to_c->num_links(e) > 0)
```
jorgensd, line 106: Can it happen that an entity is not connected to a cell?, i.e. `e_to_c->num_links(e)  == 0`?

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -26,55 +26,107 @@ namespace
...
+        entity_found = true;
+        break;
+      }
+    }
+
+    if (!entity_found)
+    {
+      throw std::runtime_error(
+          "Entity " + std::to_string(e)
+          + " is not connected to any cells or is out of bounds.");
```
jorgensd, line 128: Out of bounds is covered in  Line 93 to 99?

jorgensd, line 128:
```suggestion
      throw std::runtime_error(std::format("Entity {} is not connected to any cells or is out of bounds.", e));
```

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -200,54 +252,103 @@ std::vector<std::int32_t> fem::locate_dofs_topological(
     const mesh::Topology& topology, const DofMap& dofmap, int dim,
     std::span<const std::int32_t> entities, bool remote)
 {
-  mesh::CellType cell_type = topology.cell_type();
+  const std::size_t num_cell_types = topology.cell_types().size();
+  if (num_cell_types != 1)
```
jorgensd, line 256:
```suggestion
  if (topology.cell_types().size() != 1)
```

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -26,55 +26,107 @@ namespace
...
   }
 
-  auto c_to_e = topology.connectivity(tdim, dim);
-  if (!c_to_e)
+  if (!any_connectivity)
   {
-    throw std::runtime_error(
-        "Cell-to-entity connectivity has not been computed. Missing dims "
-        + std::to_string(tdim) + "->" + std::to_string(dim));
+    throw std::runtime_error("No connectivity from entities of dimension "
```
jorgensd, line 84: Use std::format here as well.

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -200,54 +252,103 @@ std::vector<std::int32_t> fem::locate_dofs_topological(
...
-  dofs.reserve(
-      entities.size()
-      * dofmap.element_dof_layout().entity_closure_dofs(dim, 0).size());
+  dofs.reserve(entities.size() * max_closure_dofs);
 
   // V is a sub space we need to take the block size of the dofmap and
   // the index map into account as they can differ
-  const int bs = dofmap.bs();
-  const int element_bs = dofmap.element_dof_layout().block_size();
+  const int bs = dofmaps.front().get().bs();
```
jorgensd, line 312: We should probably add a comment here that the block size of either dofmap should be the same, as pointed out by @chrisrichardson  in #4165

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -200,54 +252,103 @@ std::vector<std::int32_t> fem::locate_dofs_topological(
...
+                                 0, remote);
+}
+//-----------------------------------------------------------------------------
+std::vector<std::int32_t> fem::locate_dofs_topological(
+    const mesh::Topology& topology,
+    const std::vector<std::reference_wrapper<const DofMap>>& dofmaps, int dim,
+    std::span<const std::int32_t> entities, int entity_type_index, bool remote)
+{
+  std::vector<mesh::CellType> cell_types = topology.cell_types();
+  const int num_cell_types = cell_types.size();
```
jorgensd, line 282: See warning from sonarcloud.

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -200,54 +252,103 @@ std::vector<std::int32_t> fem::locate_dofs_topological(
...
+  const int bs = dofmaps.front().get().bs();
+  const int element_bs
+      = dofmaps.front().get().element_dof_layout().block_size();
 
   // Iterate over marked facets
   if (element_bs == bs)
   {
     // Work with blocks
-    for (auto [cell, entity_local_index] : entity_indices)
+    for (auto [cell, entity_local_index, cell_type_idx] : entity_indices)
```
jorgensd, line 320: Should we use c rather than cell to avoid the shadowing warning?

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -200,54 +252,103 @@ std::vector<std::int32_t> fem::locate_dofs_topological(
...
   {
     // Space is not blocked, unroll dofs
-    for (auto [cell, entity_local_index] : entity_indices)
+    for (auto [cell, entity_local_index, cell_type_idx] : entity_indices)
     {
       // Get cell dofmap and loop over facet dofs and 'unpack' blocked
       // dofs
-      std::span<const std::int32_t> cell_dofs = dofmap.cell_dofs(cell);
-      for (int index : entity_dofs[entity_local_index])
+      std::span<const std::int32_t> cell_dofs
```
jorgensd, line 343: Seems like we should call `dofmaps[cell_type+idx].get()` once rather than twice.

cpp/dolfinx/fem/DirichletBC.cpp
```diff
@@ -271,7 +372,7 @@ std::vector<std::int32_t> fem::locate_dofs_topological(
     // Get bc dof indices (local) in V spaces on this process that were
     // found by other processes, e.g. a vertex dof on this process that
     // has no connected facets on the boundary.
-    auto map = dofmap.index_map;
+    auto map = dofmaps.front().get().index_map;
```
jorgensd, line 375: Same here and below, call `dofmaps.front().get()` once as a separate variable?

jorgensd (changes requested):
Some minor things. In general it looks good!:)

cpp/dolfinx/fem/DirichletBC.cpp
chrisrichardson: but not in `Topology` just yet, so get this merged first!

cpp/dolfinx/fem/DirichletBC.cpp
chrisrichardson: true - maybe just reword this comment to state that both are required.

jpdean: I think, with the current connectivity implementation in FEniCSx, this statement is always true because of the direction I stated it in. Your example, Jørgen, is assuming that the reverse statement also holds, which is not the case. However, I am happy to clarify the comment! The code doesn't actually rely on it anyway, I've included a check just in case!

PR #4168, opened 2026-04-21
Improved Mixed-topology interface

jorgensd:
Suggested improved interface


python/dolfinx/fem/function.py
```diff
@@ -367,6 +363,34 @@ def functiontype(dtype):
             else:
                 raise NotImplementedError(f"Type {dtype} not supported.")
 
+        if isinstance(V, ufl.MixedFunctionSpace):
```
jorgensd, line 366: This feels a bit hacky, but I don't have a better way of doing this.

PR #4169, opened 2026-04-21
Use `cmap(i)` in `Geometry`

chrisrichardson:
Replaces previous convoluted access with `geometry.cmaps()[i]`

### DIFF
--- a/python/dolfinx/wrappers/dolfinx_wrappers/mesh.h
+++ b/python/dolfinx/wrappers/dolfinx_wrappers/mesh.h
@@ -215,7 +215,7 @@ void declare_mesh(nb::module_& m, std::string type)
       .def(
           "cmap", [](dolfinx::mesh::Geometry<T>& self, std::optional<int> i)
           { return self.cmap(i); }, "The ith coordinate map",
-        nb::arg("i").none())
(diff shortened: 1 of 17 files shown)


python/dolfinx/wrappers/dolfinx_wrappers/mesh.h
```diff
@@ -209,12 +209,13 @@ void declare_mesh(nb::module_& m, std::string type)
           nb::rv_policy::reference_internal,
           "Return coordinates of all geometry points. Each row is the "
           "coordinate of a point.")
-      .def_prop_ro(
+      .def(
```
michalhabera, line 212: @chrisrichardson Could this have been kept as `def_prop_ro`?

michalhabera: Well, not as far as Python allows. But what I mean is that you've broken a very common public API here, for renaming a very corner-case and experimental API.

chrisrichardson: @michalhabera - I am not sure what is a better interface. Happy to receive suggestions. I know this is an API change, but not too major.

Issue #225, opened 2018-10-28
Missing Real finite element 


michalhabera:
Snippet
```python
import dolfin
mesh = dolfin.UnitSquareMesh(dolfin.MPI.comm_world, 2, 2)
dolfin.FunctionSpace(mesh, ("Real", 0)).dofmap().dofs(mesh, 2)
```
returns
```python
array([0, 1, 2, 3, 4, 5, 6, 7], dtype=int32)
```
but should return (as in 2018.1.0)
```python
array([0, 0, 0, 0, 0, 0, 0, 0], dtype=int32)
```

In other words, `Real` works now as `DG0`, having one scalar per each cell, instead of one scalar per whole domain.

michalhabera:
Real element is not like `QuadratureElement` or any `BaseElement` subclass because it does not have local basis. In old dolfin such element would export `num_global_dofs` in the generated ffc code and this was then used in dolfin to construct dofmap.

As we do not want to interfere with dofmap construction in dolfinx, these globally supported spaces must be handled separately, yet allow to be assembled into block matrices with dense rows/columns.

RemDelaporteMathurin:
Hi team! Is this still planned for dolfinx?

RemDelaporteMathurin:
Hi @michalhabera @garth-wells , I looked amongst the various repos for Real elements and found a couple of issues/PRs to add a Real space.

Is it now supported? I found mentions of real elements in FFCX
https://github.com/FEniCS/ffcx/blob/06957dbc7428256760a52a2a543eaca37a80bebf/ffcx/codegeneration/ufcx.h#L69

jorgensd:
There is a pure C++ implementation of this at: https://gist.github.com/jorgensd/7b0ed8e55b8446fac0eff4b821dad67a
@finsberg and I will make a suggested Python interface for this in a few days.

jorgensd:
> > There is a pure C++ implementation of this at: https://gist.github.com/jorgensd/7b0ed8e55b8446fac0eff4b821dad67a @finsberg and I will make a suggested Python interface for this in a few days.
> 
> Can I help with this in any way? Also pinging @KulaginVladimir

A Python interface for this feature is available in: https://github.com/scientificcomputing/scifem
and should work for DOLFINx>=v0.8.0. Any feedback is appreciated.

jorgensd:
Implementation at: https://scientificcomputing.github.io/scifem/examples/real_function_space.html

[...]

PR #4170, opened 2026-04-21
Add real element to DOLFINx

jorgensd:
Add the notion of real element to DOLFINx. Implemented based on https://github.com/scientificcomputing/scifem/blob/main/src/scifem.cpp
Relies on https://github.com/FEniCS/basix/pull/1001

One should not be able to make a `mixed_element` with real element. One should use the block constructor (i.e. MixedFunctionSpace or manually block it).

The real element is now:
- A DG-0 element in basix, gives the exact same kernel as with a DG-0/Constant.
- Gives the ownership of the DOF to the first process that has a cell (sometimes rank 0 has no cell).

### DIFF
--- a/cpp/dolfinx/fem/dofmapbuilder.cpp
+++ b/cpp/dolfinx/fem/dofmapbuilder.cpp
@@ -1,4 +1,5 @@
-// Copyright (C) 2008-2018 Anders Logg, Ola Skavhaug and Garth N. Wells
+// Copyright (C) 2008-2026 Anders Logg, Ola Skavhaug, Garth N. Wells and Jørgen
+// S. Dokken
 //
 // This file is part of DOLFINx (https://www.fenicsproject.org)
 //
@@ -12,6 +13,7 @@
 #include <dolfinx/common/IndexMap.h>
 #include <dolfinx/common/MPI.h>
 #include <dolfinx/common/Timer.h>
+#include <dolfinx/fem/DofMap.h>
 #include <dolfinx/graph/AdjacencyList.h>
 #include <dolfinx/mesh/Topology.h>
 #include <dolfinx/mesh/cell_types.h>
@@ -689,4 +691,51 @@ fem::build_dofmap_data(
   return {std::move(index_map), element_dof_layouts.front().block_size(),
           std::move(dofmaps)};
 }
+
+//-----------------------------------------------------------------------------
+fem::DofMap fem::build_real_element_dofmap(
+    const mesh::Topology& topology,
+    const std::vector<std::vector<std::vector<int>>>& entity_dofs,
+    const std::vector<std::vector<std::vector<int>>>& entity_closure_dofs,
+    std::size_t value_size)
+{
+  // We select the process that owns cell 0 to have all dofs and all other
+  // processes ghost them
+  std::shared_ptr<const dolfinx::common::IndexMap> cell_map
+      = topology.index_map(topology.dim());
+  const int is_owner
+      = (cell_map->local_range()[0] == 0) and cell_map->size_local() > 0;
+
+  std::int32_t num_dofs = (is_owner) ? 1 : 0;
+  std::int32_t num_ghosts = (is_owner) ? 0 : 1;
+  std::vector<std::int64_t> ghosts(num_ghosts, 0);
+  ghosts.reserve(1);
+
+  // Send owning rank to all processes so that they can set owner of ghost dofs
+  int rank = dolfinx::MPI::rank(topology.comm());
+  std::array<int, 2> send_owner_pair = {is_owner, rank};
+  std::array<int, 2> recv_owner_pair;
+  MPI_Allreduce(&send_owner_pair, &recv_owner_pair, 1, MPI_2INT, MPI_MAXLOC,
+                topology.comm());
+  std::vector<int> owners(num_ghosts, recv_owner_pair[1]);
+  owners.reserve(1);
+
+  // Create index map
+  std::shared_ptr<const dolfinx::common::IndexMap> imap
+      = std::make_shared<const dolfinx::common::IndexMap>(
+          topology.comm(), num_dofs, ghosts, owners);
+
+  // Create element dof layout
+  dolfinx::fem::ElementDofLayout dof_layout(value_size, entity_dofs,
+                                            entity_closure_dofs, {}, {});
+
+  // Create dofmap array
+  std::size_t num_cells_on_process
+      = topology.index_map(topology.dim())->size_local()
+        + topology.index_map(topology.dim())->num_ghosts();
+
+  std::vector<std::int32_t> dofmap(num_cells_on_process, 0);
+  dofmap.reserve(1);
+  return dolfinx::fem::DofMap(dof_layout, imap, value_size, dofmap, value_size);
+};
 //-----------------------------------------------------------------------------
--- a/cpp/dolfinx/fem/dofmapbuilder.cpp
+++ b/cpp/dolfinx/fem/dofmapbuilder.cpp
@@ -721,9 +721,8 @@ fem::DofMap fem::build_real_element_dofmap(
   owners.reserve(1);
 
   // Create index map
-  auto imap
-      = std::make_shared<const dolfinx::common::IndexMap>(
-          topology.comm(), num_dofs, ghosts, owners);
+  auto imap = std::make_shared<const dolfinx::common::IndexMap>(
+      topology.comm(), num_dofs, ghosts, owners);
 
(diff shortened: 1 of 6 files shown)


jorgensd:
Resolves: https://github.com/FEniCS/dolfinx/issues/225

mscroggs (approved):

michalhabera:
> One should not be able to make a mixed_element with real element. One should use the block constructor (i.e. MixedFunctionSpace or manually block it).

You mean that you cannot create a Real element with nontrivial shape, but should work in MFS?

jorgensd:
> > One should not be able to make a mixed_element with real element. One should use the block constructor (i.e. MixedFunctionSpace or manually block it).
> 
> You mean that you cannot create a Real element with nontrivial shape, but should work in MFS?

No, one should **not** be able to do:
```python
el = basix.ufl.element("Lagrange", mesh.basix_cell(), 2)
r_el = basix.ufl.real_element(mesh.basix_cell(), shape=(2,))
me = basix.ufl.mixed_element([el, r_el])
```
but one can do:
```python
el = basix.ufl.element("Lagrange", mesh.basix_cell(), 2)
r_el = basix.ufl.real_element(mesh.basix_cell(), shape=(2,))
R = dolfinx.fem.functionspace(mesh, r_el)
W = ufl.MixedFunctionSpace(V, R)
u, lmbd = ufl.TrialFunctions(W)
v, mu = ufl.TestFunctions(W)
```

michalhabera (approved):
This change is minimal and nicely isolated. I like you've added custom dofmap builder and did not try to tweak the existing one -- which was the bad design in legacy FEniCS if I remember well.

I still believe we should follow matrix-free/MATSHELL approach for all these extremely skew spaces, but this is a good solution until we have more time to work on the alternative.

jorgensd:
> This change is minimal and nicely isolated. I like you've added custom dofmap builder and did not try to tweak the existing one -- which was the bad design in legacy FEniCS if I remember well.
> 
> I still believe we should follow matrix-free/MATSHELL approach for all these extremely skew spaces, but this is a good solution until we have more time to work on the alternative.

In combination with:
https://github.com/FEniCS/dolfinx/pull/3938
we should be able to use matrix free approach with this element  with no changes to the core.

chrisrichardson (approved):

mscroggs (approved):

cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -689,4 +691,51 @@ fem::build_dofmap_data(
...
+    const std::vector<std::vector<std::vector<int>>>& entity_dofs,
+    const std::vector<std::vector<std::vector<int>>>& entity_closure_dofs,
+    std::size_t value_size)
+{
+  // We select the process that owns cell 0 to have all dofs and all other
+  // processes ghost them
+  std::shared_ptr<const dolfinx::common::IndexMap> cell_map
+      = topology.index_map(topology.dim());
+  const int is_owner
+      = (cell_map->local_range()[0] == 0) and cell_map->size_local() > 0;
```
mscroggs: SonarCloud suggest this. I'm not sure which is better...

```suggestion
      = (cell_map->local_range()[0] == 0) && cell_map->size_local() > 0;
```

cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -689,4 +691,51 @@ fem::build_dofmap_data(
...
+
+  std::int32_t num_dofs = (is_owner) ? 1 : 0;
```
mscroggs:
```suggestion
  std::int32_t num_dofs = is_owner ? 1 : 0;
```

cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -689,4 +691,51 @@ fem::build_dofmap_data(
...
+  std::int32_t num_ghosts = (is_owner) ? 0 : 1;
```
mscroggs:
```suggestion
  std::int32_t num_ghosts = is_owner ? 0 : 1;
```

python/dolfinx/fem/element.py
```diff
@@ -368,6 +368,10 @@ def finiteelement(
                 ufl_e.is_symmetric,
             )
         )
+    elif ufl_e.is_real:
+        basix_e = ufl_e.basix_element._e
+        value_shape = ufl_e.reference_value_shape if ufl_e.block_size > 1 else None
+        return FiniteElement(CppElement(basix_e, value_shape, ufl_e.is_symmetric))
```
mscroggs: SonarCloud correct spotted that these two branches are the same
```suggestion
```

cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -689,4 +691,51 @@ fem::build_dofmap_data(
...
+  int rank = dolfinx::MPI::rank(topology.comm());
+  std::array<int, 2> send_owner_pair = {is_owner, rank};
+  std::array<int, 2> recv_owner_pair;
+  MPI_Allreduce(&send_owner_pair, &recv_owner_pair, 1, MPI_2INT, MPI_MAXLOC,
+                topology.comm());
+  std::vector<int> owners(num_ghosts, recv_owner_pair[1]);
+  owners.reserve(1);
+
+  // Create index map
+  std::shared_ptr<const dolfinx::common::IndexMap> imap
```
jhale:
```suggestion
  auto imap
```

Typing is explicit when using `make_shared`

cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -689,4 +691,51 @@ fem::build_dofmap_data(
   return {std::move(index_map), element_dof_layouts.front().block_size(),
           std::move(dofmaps)};
 }
+
+//-----------------------------------------------------------------------------
+fem::DofMap fem::build_real_element_dofmap(
+    const mesh::Topology& topology,
+    const std::vector<std::vector<std::vector<int>>>& entity_dofs,
+    const std::vector<std::vector<std::vector<int>>>& entity_closure_dofs,
+    std::size_t value_size)
```
jhale: Should be `int` to be consistent with DofMap constructor.

jhale:
```suggestion
    int value_size)
```

cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -689,4 +691,51 @@ fem::build_dofmap_data(
...
+  std::vector<int> owners(num_ghosts, recv_owner_pair[1]);
+  owners.reserve(1);
+
+  // Create index map
+  std::shared_ptr<const dolfinx::common::IndexMap> imap
+      = std::make_shared<const dolfinx::common::IndexMap>(
+          topology.comm(), num_dofs, ghosts, owners);
+
+  // Create element dof layout
+  dolfinx::fem::ElementDofLayout dof_layout(value_size, entity_dofs,
```
jhale, line 728: Can use `auto`, type explicit in use of `make_shared`

cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -689,4 +691,51 @@ fem::build_dofmap_data(
...
+                                            entity_closure_dofs, {}, {});
+
+  // Create dofmap array
+  std::size_t num_cells_on_process
```
jhale: Cannot be more than `std::int32_t`

jhale (approved):

cpp/dolfinx/fem/dofmapbuilder.cpp
jorgensd:
```suggestion
  std::int32_t num_cells_on_process
```

cpp/dolfinx/fem/dofmapbuilder.cpp
jorgensd: But we are not using a make_shared pointer here as it is not used further down the line?

jhale (approved):
Looks good. 

An aside, Real is a misnomer now we have complex numbers... anyway!

jorgensd:
> Looks good.
> 
> An aside, Real is a misnomer now we have complex numbers... anyway!

Are you saying that a real complex space is confusing? 😂 or that a real element with data type complex128;)
I kind of agree, but for historical reasons let’s keep the name for now and revise it later. 

What I really want is to have a space of simple functions (patchwise real elements), which exists in scifem :)

PR #4171, opened 2026-04-22
Add point mesh constructor

### LINKED ISSUE #2593: Support vertex-only 'meshes'

For mixed dimensional problems, it would be sensible to have 0D (vertex meshes).
However, in the current implementation, there are many assumptions (such as the `build_local_dual_graph`, not clear what we should do in this case) that does not work with vertex meshes.

Changes that is required (and clear):
`dolfinx/cpp/dolfinx/mesh/graphbuild.cpp`
```c++
constexpr mesh::CellType get_cell_type(int num_vertices, int tdim)
{
  switch (tdim)
  {
  case 0:
    return mesh::CellType::point;
```
`dolfinx/python/dolfinx/mesh.py`
```python
_uflcell_to_dolfinxcell = {
    "vertex": CellType.point,
```

### Suggestion user interface

```python3
from mpi4py import MPI
import ufl
import dolfinx
import numpy as np
points = np.array([[0, 0], [0, 1], [1, 0], [1, 1]], dtype=np.float64)
vertex = np.array([[0], [1], [2], [3]], dtype=np.int64)
ufl_tri = ufl.Mesh(ufl.VectorElement("DG", ufl.vertex, 0, dim=2))


def serial_partitioner(comm, n, m, topo):
    dest = np.zeros(topo.num_nodes, dtype=np.int32)

    return dolfinx.cpp.graph.AdjacencyList_int32(dest)


dolfinx.log.set_log_level(dolfinx.log.LogLevel.INFO)
tri_mesh = dolfinx.mesh.create_mesh(
    MPI.COMM_WORLD, vertex, points, ufl_tri, partitioner=serial_partitioner)
```

jorgensd:
Resolve https://github.com/FEniCS/dolfinx/issues/2593.

Point cloud meshes are for instance very useful for visualizing data at quadrature points, or for reading in point data for inverse modelling.


chrisrichardson (approved):

python/dolfinx/mesh.py
```diff
@@ -1268,3 +1269,44 @@ def create_geometry(
     if (dtype := np.dtype(element.dtype)) != x.dtype:
         raise ValueError(f"Mismatch in x dtype ({x.dtype}) and coordinate element ({dtype})")
     return Geometry(ftype(index_map, dofmap, element._cpp_object, x, input_global_indices))
+
+
+def create_point_mesh(comm: _MPI.Intracomm, points: npt.NDArray[np.float32 | np.float64]) -> Mesh:
+    """Create a mesh consisting of points only.
+
+    Note:
+        No nodes are shared between processes.
```
jhale: Nodes or vertices?

python/dolfinx/mesh.py
```diff
@@ -1268,3 +1269,44 @@ def create_geometry(
...
+
+    Args:
+        comm: MPI communicator to create the mesh on.
+        points: Points local to the process in the mesh.
+    """
+    # Create mesh topology
```
jhale: This comment is trivial - instead, what exactly is in a topology-less mesh that still has a topology?

python/dolfinx/mesh.py
jorgensd: I guess for a 1D mesh we can call it vertex?

python/dolfinx/mesh.py
jorgensd: Well, it has a topology, but the only connectivity that can exist is 0->0 being an identity.

PR #4173, opened 2026-04-23
Geometry access to `dofmap(i)` with mixed topology 

chrisrichardson:
Makes the access to `dofmap` the same as for `cmap`, with an access function `dofmap(i)` where `i` is optional.


cpp/dolfinx/mesh/Geometry.h
```diff
@@ -108,28 +108,22 @@ class Geometry
   int dim() const { return _dim; }
 
   /// @brief DofMap for the geometry.
```
garth-wells, line 110: Expand docs. 

How to I get `i` for a given cell type?

PR #4176, opened 2026-04-24
Fix set_bc

jorgensd:
`bcs` are reference wrappers, thus, no set option.

Found while updating asimov contact and discussing compilation errors in that library with Gemini 3.1 Pro

### DIFF
--- a/cpp/dolfinx/fem/petsc.h
+++ b/cpp/dolfinx/fem/petsc.h
@@ -454,28 +454,34 @@ void apply_lifting(
 
 // FIXME: Move these function elsewhere?
 
-// FIXME: clarify x0
-// FIXME: clarify what happens with ghosts
-
-/// Set bc values in owned (local) part of the PETSc vector, multiplied
-/// by 'alpha'. The vectors b and x0 must have the same local size. The
-/// bcs should be on (sub-)spaces of the form L that b represents.
+/// Entries in `b` that are constrained by a Dirichlet boundary
+/// conditions are set to `alpha * (x_bc - x0)`, where `x_bc` is the
+/// (interpolated) boundary condition value.
+///
+/// @param[in] b The vector to apply the boundary condition to. The local
+/// (owned) part of this vector is modified. The user is reponsible of
+/// scattering the changes to the ghost part of the vector if necessary.
+/// @param[in] bcs The boundary conditions to apply.
+/// @param[in] x0 Optional vector used in computing the value to set. If
+/// not provided it is treated as zero. The local (owned) part of this vector is
+/// used.
+/// @param[in] alpha Scaling to apply.
 template <std::floating_point T>
 void set_bc(
     Vec b,
     const std::vector<std::reference_wrapper<const DirichletBC<PetscScalar, T>>>
         bcs,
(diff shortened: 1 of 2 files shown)


chrisrichardson (approved):

cpp/dolfinx/fem/petsc.h
```diff
@@ -482,14 +482,14 @@ void set_bc(
     VecGetArrayRead(x0_local, &array);
     std::span<const PetscScalar> _x0(array, n);
     for (auto& bc : bcs)
-      bc.set(_b, _x0, alpha);
+      bc.get().set(_b, _x0, alpha);
```
schnellerhase, line 491:
```suggestion
      bc.set(_b, _x0, alpha);
```

cpp/dolfinx/fem/petsc.h
```diff
@@ -482,14 +482,14 @@ void set_bc(
...
     VecRestoreArrayRead(x0_local, &array);
     VecGhostRestoreLocalForm(x0, &x0_local);
   }
   else
   {
     for (auto& bc : bcs)
-      bc->set(_b, std::nullopt, alpha);
+      bc.get().set(_b, std::nullopt, alpha);
```
schnellerhase, line 498:
```suggestion
      bc.set(_b, std::nullopt, alpha);
```

schnellerhase (commented):
`std::reference_wrapper` implicitly converts to `T&` and can be used as a 'normal' ref.

`bc->set` not caught, because untested (never compiled), right? C++ hyperelasticity demo duplicates the logic (in `F` callback) if I am not mistaken, can we use `set_bc` there to ensure it is covered?

chrisrichardson:
So can we just close this?

jorgensd:
> So can we just close this?

No, there is still a bug, i.e. `bc->set` is not valid.

jorgensd:
> `std::reference_wrapper` implicitly converts to `T&` and can be used as a 'normal' ref.
> 
> `bc->set` not caught, because untested (never compiled), right? C++ hyperelasticity demo duplicates the logic (in `F` callback) if I am not mistaken, can we use `set_bc` there to ensure it is covered?

As far as I can tell, this is not true, as I get:
```bash
/root/shared/dolfinx/cpp/demo/hyperelasticity/main.cpp:172:23:   required from here
/usr/local/dolfinx-real/include/dolfinx/fem/petsc.h:493:10: error: 'const class std::reference_wrapper<const dolfinx::fem::DirichletBC<double, double> >' has no member named 'set'; did you mean 'get'?
  493 |       bc.set(_b, _x0, alpha);
      |       ~~~^~~
      |       get
/usr/local/dolfinx-real/include/dolfinx/fem/petsc.h:500:7: error: base operand of '->' has non-pointer type 'const std::reference_wrapper<const dolfinx::fem::DirichletBC<double, double> >'
  500 |       bc->set(_b, std::nullopt, alpha);
      |       ^~
```

schnellerhase:
Right, c++'s dot operator being weird https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2014/n4173.pdf.

schnellerhase (approved):

PR #4177, opened 2026-04-24
`VTXWriter` add support for append file mode

schnellerhase:
Adds support to `ADIOS2Writer` and `VTXWriter` to open files in append mode (following interface of `VTKFile`). Enabling usage such as

```python
import dolfinx
from mpi4py import MPI

mesh = dolfinx.mesh.create_unit_square(comm := MPI.COMM_WORLD, n := 10, n)
V = dolfinx.fem.functionspace(mesh, ("CG", 1))
f = dolfinx.fem.Function(V)

# this usage previously only held the 9th iterate in the output, i.e. was not supported
for i in range(10):
    f.x.array[:] = i
    with dolfinx.io.VTXWriter(
        comm,
        "out_a.bp",
        [f],
        mode="a",
    ) as writer:
        writer.write(i)

# now holds equivalent output to:
with dolfinx.io.VTXWriter(
    comm,
    "out_w.bp",
    [f],
) as writer:
    for i in range(10):
        f.x.array[:] = i
        writer.write(i)
```

- Also `VTXMeshPolicy.reuse` is supported (with compatibility of meshes being checked)
- Demonstrates the new feature, with simplified I/o logic in the naiver stokes demo.


python/dolfinx/io/utils.py
```diff
@@ -56,6 +57,8 @@ def __init__(
             Args:
                 comm: The MPI communicator
                 filename: The output filename
+                mode: The filemode to open the file in, one of 'a' (append)
```
jhale: Remove "one of", automatic when using "or".

cpp/dolfinx/io/ADIOS2Writers.cpp
```diff
@@ -38,6 +38,24 @@ void ADIOS2Writer::close()
     _engine->Close();
 }
 //-----------------------------------------------------------------------------
+adios2::Mode impl_adios2::mode(const std::string& mode)
+{
+  if (mode == "a")
+    return adios2::Mode::Append;
+
+  if (mode == "w")
```
jhale: else if and else?

cpp/dolfinx/io/ADIOS2Writers.cpp
```diff
@@ -38,6 +38,24 @@ void ADIOS2Writer::close()
     _engine->Close();
 }
 //-----------------------------------------------------------------------------
+adios2::Mode impl_adios2::mode(const std::string& mode)
```
jhale: Did you consider just wrapping `adios2::Mode` directly as an enum?

jhale: Use non-owning `std::string_view`

cpp/dolfinx/io/ADIOS2Writers.h
```diff
@@ -157,6 +159,11 @@ extract_common_mesh(const typename adios2_writer::U<T>& u)
   return mesh;
 }
 
+/// Convert string to corresponding adios2 mode.
+/// @param[in] mode Mode in string representation, either "w" or "a".
```
jhale, line 166: Leaving aside Enum vs no Enum, it's generally most transparent to map modes directly e.g. "Read" -> "adios2::Mode::Read". Then I have a clear idea what "Read" means in the context of the underlying library.

jhale (changes requested):

jorgensd:
I'm a bit sceptical about adding an "append" mode, as the VTXScheme shouldn't (or at least couldn't previously) be modified at runtime with ADIOS2, which is why the initializer takes in the functions/meshes to store at each time-step, instead of following the XDMF file version where we write when we can.

adding an append-mode needs us to revise the internal logic to ensure for consistent input at multiple steps.
We would also have to thoroughly check which variables ADIOS2 now allows to be time-dependent (last section of https://adios2.readthedocs.io/en/latest/ecosystem/visualization.html)

cpp/dolfinx/io/ADIOS2Writers.cpp
schnellerhase: The `ADIOS2Writer` operates directly on `adios::Mode` only the `VTXWriter` on the string. I understood the second as being the no longer adios2 interface specific implementation of dolfinx's interface, therefore not using the enum there.

cpp/dolfinx/io/ADIOS2Writers.h
schnellerhase: `r`, `w` are the unix names for the operation and also what we use for the `VTKFile` interface.

schnellerhase:
> I'm a bit sceptical about adding an "append" mode, as the VTXScheme shouldn't (or at least couldn't previously) be modified at runtime with ADIOS2, which is why the initializer takes in the functions/meshes to store at each time-step, instead of following the XDMF file version where we write when we can.
> 
> adding an append-mode needs us to revise the internal logic to ensure for consistent input at multiple steps. We would also have to thoroughly check which variables ADIOS2 now allows to be time-dependent (last section of https://adios2.readthedocs.io/en/latest/ecosystem/visualization.html)

I checked with the following script, which seems to correctly append to the xml contents (while currently duplicating everything though) of the VTXScheme. That should allow us to enable append mode writing in principle, right?

```python
import dolfinx
from mpi4py import MPI

mesh = dolfinx.mesh.create_unit_square(comm := MPI.COMM_WORLD, n := 10, n)
V = dolfinx.fem.functionspace(mesh, ("CG", 1))
V2 = dolfinx.fem.functionspace(mesh, ("CG", 2))
f = dolfinx.fem.Function(V)
f.x.array[:] = 1
f2 = dolfinx.fem.Function(V2)
f2.x.array[:] = 2

with dolfinx.io.VTXWriter(
    comm,
    "out.bp",
    "w",
    [f],
) as writer:
    writer.write(0)


with dolfinx.io.VTXWriter(
    comm,
    "out.bp",
    "a",
    [f2],
) as writer:
    writer.write(1)

from adios2 import Stream

with Stream("out.bp", "r") as s:
    # steps comes from the stream
    for _ in s.steps():
        # track current step
        print(f"Current step is {s.current_step()}")

        # inspect variables in current step
        # for name, info in s.available_variables().items():
        #     print("variable_name: " + name, end=" ")
        #     for key, value in info.items():
        #         print("\t" + key + ": " + value, end=" ")
        #     print()

        xml = s.read_attribute("vtk.xml")
        print(xml)
```

Output:
```bash
Current step is 0
<?xml version="1.0"?>
<VTKFile type="UnstructuredGrid" version="0.1">
  <UnstructuredGrid>
    <Piece NumberOfPoints="NumberOfNodes" NumberOfCells="NumberOfCells">
      <Points>
        <DataArray Name="geometry" />
      </Points>
      <Cells>
        <DataArray Name="connectivity" />
        <DataArray Name="types" />
      </Cells>
      <PointData>
        <DataArray Name="TIME">step</DataArray>
        <DataArray Name="vtkOriginalPointIds" />
        <DataArray Name="vtkGhostType" />
        <DataArray Name="f" />
      </PointData>
    </Piece>
  </UnstructuredGrid>
</VTKFile>

Current step is 1
<?xml version="1.0"?>
<VTKFile type="UnstructuredGrid" version="0.1">
  <UnstructuredGrid>
    <Piece NumberOfPoints="NumberOfNodes" NumberOfCells="NumberOfCells">
      <Points>
        <DataArray Name="geometry" />
      </Points>
      <Cells>
        <DataArray Name="connectivity" />
        <DataArray Name="types" />
      </Cells>
      <PointData>
        <DataArray Name="TIME">step</DataArray>
        <DataArray Name="vtkOriginalPointIds" />
        <DataArray Name="vtkGhostType" />
        <DataArray Name="f" />
      </PointData>
    </Piece>
  </UnstructuredGrid>
</VTKFile>
```

jorgensd:
@schnellerhase They did change how attributes work in adios2 a while ago (https://github.com/ornladios/ADIOS2/issues/4471) which blurred the lines between variables and attributes, which I guess it makes it work.

cpp/dolfinx/io/ADIOS2Writers.h
garth-wells: I prefer enums (tighter typing avoid bugs). Move the enums to the `dolfinx::io` namespace - can be shared across backends.

schnellerhase:
Verified the sample https://github.com/FEniCS/dolfinx/pull/4177#issuecomment-4363574641 visualizes as time series in Paraview.

cpp/dolfinx/io/ADIOS2Writers.h
schnellerhase: Filed issue https://github.com/FEniCS/dolfinx/issues/4366 to address - this requires changes in other back ends as well.

jorgensd:
> Verified the sample [#4177 (comment)](https://github.com/FEniCS/dolfinx/pull/4177#issuecomment-4363574641) visualizes as time series in Paraview.

It only works if the function name is the same for both time steps, i.e.
```python
import dolfinx
from mpi4py import MPI

mesh = dolfinx.mesh.create_unit_square(comm := MPI.COMM_WORLD, n := 10, n)
V = dolfinx.fem.functionspace(mesh, ("CG", 1))
V2 = dolfinx.fem.functionspace(mesh, ("CG", 2))
f = dolfinx.fem.Function(V)
f.interpolate(lambda x: x[0] + x[1])
f2 = dolfinx.fem.Function(V2)
f2.interpolate(lambda x: x[0] + 2*x[1])
f2.name = "p"
with dolfinx.io.VTXWriter(
    comm,
    "out.bp",
    "w",
    [f],
) as writer:
    writer.write(0)


with dolfinx.io.VTXWriter(
    comm,
    "out.bp",
    "a",
    [f2],
) as writer:
    writer.write(1)

from adios2 import Stream

with Stream("out.bp", "r") as s:
    # steps comes from the stream
    for _ in s.steps():
        # track current step
        print(f"Current step is {s.current_step()}")

        # inspect variables in current step
        # for name, info in s.available_variables().items():
        #     print("variable_name: " + name, end=" ")
        #     for key, value in info.items():
        #         print("\t" + key + ": " + value, end=" ")
        #     print()

        xml = s.read_attribute("vtk.xml")
        print(xml)
  ```
  visualizes as
  
image
but works if you remove:
```python
f2.name = "p"
```
Furthermore, what do we expect if you append to an existing time step.
Currently I can toggle this in Paraview through the time stepper, even if there is no time step:

```python
import dolfinx
from mpi4py import MPI
comm = MPI.COMM_WORLD
mesh = dolfinx.mesh.create_unit_square(comm, 5, 5, dolfinx.mesh.CellType.triangle)
V = dolfinx.fem.functionspace(mesh, ("CG", 1))
f = dolfinx.fem.Function(V)
f.interpolate(lambda x: x[0] + x[1])

with dolfinx.io.VTXWriter(
    comm,
    "out.bp",
    "w",
    [f],
) as writer:
    writer.write(0)

mesh.topology.create_connectivity(mesh.topology.dim-1, mesh.topology.dim)
exterior_facets = dolfinx.mesh.exterior_facet_indices(mesh.topology)
submesh, _, _, _ = dolfinx.mesh.create_submesh(mesh, mesh.topology.dim - 1, exterior_facets)
Q = dolfinx.fem.functionspace(submesh, ("CG", 2))
q = dolfinx.fem.Function(Q)
q.interpolate(lambda x: x[0] + x[1])

with dolfinx.io.VTXWriter(
    submesh.comm,
    "out.bp",
    "a",
    [q],
) as writer:
    writer.write(0)

from adios2 import Stream

with Stream("out.bp", "r") as s:
    # steps comes from the stream
    for _ in s.steps():
        # track current step
        print(f"Current step is {s.current_step()}")

        # inspect variables in current step
        # for name, info in s.available_variables().items():
        #     print("variable_name: " + name, end=" ")
        #     for key, value in info.items():
        #         print("\t" + key + ": " + value, end=" ")
        #     print()

        xml = s.read_attribute("vtk.xml")
        print(xml)
```

cpp/dolfinx/io/ADIOS2Writers.h
```diff
@@ -461,21 +475,33 @@ class VTXWriter : public ADIOS2Writer
...
   /// @param[in] filename Name of output file.
+  /// @param[in] mode Mode to open file with
   /// @param[in] mesh Mesh to write.
   /// @param[in] engine ADIOS2 engine type.
   /// @note This format supports arbitrary degree meshes.
   /// @note The mesh geometry can be updated between write steps but the
   /// topology should not be changed between write steps.
   VTXWriter(MPI_Comm comm, const std::filesystem::path& filename,
-            std::shared_ptr<const mesh::Mesh<T>> mesh,
+            const std::string& mode, std::shared_ptr<const mesh::Mesh<T>> mesh,
```
jorgensd: Let's put mode at the end, and make it optional (with default "w") as it's silly to break API of all IO for this new argument.

schnellerhase:
With different names now raises `RuntimeError: VTK scheme not compatible.` - we compare the `vtk.xml` of the previous step in append mode to align.

schnellerhase:
If we re-/overwrite a time step the latest write will persist in the output (2nd example of https://github.com/FEniCS/dolfinx/pull/4177#issuecomment-5191205419) - it does not show multiple time steps in the time series for me, `.bp.` reload issues, @jorgensd ?

schnellerhase:
One more concern could be the mesh reuse policy - probably we should extend the check of alignment to `NumberOfNodes` and `NumberOfCells` for write compatibility (the xml structure only holds the adios2 variable names not the actual values)?

jorgensd:
> One more concern could be the mesh reuse policy - probably we should extend the check of alignment to `NumberOfNodes` and `NumberOfCells` for write compatibility (the xml structure only holds the adios2 variable names not the actual values)?

Yes, the xml structure just points to the variable, that might change over steps. So we need some checks if we use append mode.

python/test/unit/io/test_adios2.py
```diff
@@ -293,3 +293,48 @@ def get_data_array_names(data):
 
             cell_arrays = get_data_array_names(xml_cell_data)
             assert "q" in cell_arrays or ("q_imag" in cell_arrays and "q_real" in cell_arrays)
+
+    def test_append_compat_func(self, tempdir):
```
schnellerhase:
```suggestion
    def test_append_compat_schema(self, tempdir):
```

cpp/dolfinx/io/ADIOS2Writers.h
```diff
@@ -1,20 +1,23 @@
-// Copyright (C) 2021-2023 Jørgen S. Dokken and Garth N. Wells
+// Copyright (C) 2021-2026 Jørgen S. Dokken, Garth N. Wells and Paul T. Kühner
 //
 // This file is part of DOLFINX (https://www.fenicsproject.org)
 //
 // SPDX-License-Identifier:    LGPL-3.0-or-later
 
 #pragma once
 
+#include <cstdint>
```
schnellerhase:
```suggestion
```

cpp/dolfinx/io/ADIOS2Writers.h
```diff
@@ -461,21 +471,26 @@ class VTXWriter : public ADIOS2Writer
   ///
   /// @param[in] comm MPI communicator to open the file on.
   /// @param[in] filename Name of output file.
+  /// @param[in] mode Mode to open file with
   /// @param[in] mesh Mesh to write.
   /// @param[in] engine ADIOS2 engine type.
```
schnellerhase:
```suggestion
/// @param[in] mesh Mesh to write.
/// @param[in] engine ADIOS2 engine type.
/// @param[in] mode Mode to open file with
```

cpp/dolfinx/io/ADIOS2Writers.h
```diff
@@ -495,8 +511,10 @@ class VTXWriter : public ADIOS2Writer
   /// @note This format supports arbitrary degree meshes.
   VTXWriter(MPI_Comm comm, const std::filesystem::path& filename,
```
schnellerhase:
```suggestion
/// @param[in] u List of functions. The functions must (1) share the
/// same mesh and (2) be (discontinuous) Lagrange functions. The
/// element family and degree, and degree-of-freedom map (up to the
/// blocksize) must be the same for all functions.
/// @param[in] engine ADIOS2 engine type.
/// @param[in] mesh_policy Controls if the mesh is written to file at
/// the first time step only or is re-written (updated) at each time
/// step.
/// @param[in] mode Mode to open file with
/// @note This format supports arbitrary degree meshes.
VTXWriter(MPI_Comm comm, const std::filesystem::path& filename,
```

cpp/dolfinx/io/ADIOS2Writers.h
schnellerhase:
```suggestion
/// @param[in] u List of functions. The functions must (1) share the
/// same mesh and (2) be (discontinuous) Lagrange functions. The
/// element family and degree must be the same for all functions.
/// @param[in] mesh_policy Controls if the mesh is written to file at
/// the first time step only or is re-written (updated) at each time
/// step.
/// @param[in] mode Mode to open file with
```

PR #4184, opened 2026-04-27
Ignore: Add SuperLU_DIST LinearProblem

jhale:
Very rough - just started by copying across PETSc LinearProblem and sketching in some notes.


jhale:
This was a mistake using GitHub, can be ignored.

PR #4185, opened 2026-04-27
Add SuperLU_DIST LinearProblem class

jhale:
This proposes a very simple SuperLU_DIST `LinearProblem` class which has a similar interface to the PETSc one.

Points for discussion:
- (Done). Do we want to expose the SuperLU_DIST solvers underlying options? Symmetry might be interesting, although @chrisrichardson couldn't find any performance differences when he tried it.
- Currently there is no way to reuse the decomposition of `A` for repeated solves - that said, the same applies to our PETSc `LinearProblem`. This is possible through the low-level class, so 'advanced' users can just use that.


jorgensd:
> * Currently there is no way to reuse the decomposition of `A` for repeated solves - that said, the same applies to our PETSc `LinearProblem`. This is possible through the low-level class, so 'advanced' users can just use that

PETSc reuses the LU decomposition under the hood if the matrix doesn't change, we don't have to set any options for this to happen.

jhale:
> > * Currently there is no way to reuse the decomposition of `A` for repeated solves - that said, the same applies to our PETSc `LinearProblem`. This is possible through the low-level class, so 'advanced' users can just use that
> 
> PETSc reuses the LU decomposition under the hood if the matrix doesn't change, we don't have to set any options for this to happen.

Indeed, but I'm not totally convinced our code allows for that:

https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/fem/petsc.py#L900

If A is zeroed and re-assembled, even with the same form + data, then  PETSc probably detects this as a 'change'?

jhale:
I experimented with factor re-use on `jhale/superlu-linear-problem-2-factor`. On balance I think users who need to re-use factors should use the linear solver directly.

jorgensd:
> > > * Currently there is no way to reuse the decomposition of `A` for repeated solves - that said, the same applies to our PETSc `LinearProblem`. This is possible through the low-level class, so 'advanced' users can just use that
> > 
> > 
> > PETSc reuses the LU decomposition under the hood if the matrix doesn't change, we don't have to set any options for this to happen.
> 
> Indeed, but I'm not totally convinced our code allows for that:
> 
> https://github.com/FEniCS/dolfinx/blob/main/python/dolfinx/fem/petsc.py#L900
> 
> If A is zeroed and re-assembled, even with the same form + data, then PETSc probably detects this as a 'change'?

Indeed, then you don't get the re-usage:
```python
from mpi4py import MPI
import dolfinx.fem.petsc
import ufl
import time
N = 25
mesh = dolfinx.mesh.create_unit_cube(MPI.COMM_WORLD, N, N, N)
u, v= ufl.TrialFunction(V), ufl.TestFunction(V)

a = ufl.inner(u, v) * ufl.dx
x = ufl.SpatialCoordinate(mesh)
L = ufl.inner(ufl.sin(5 * ufl.pi * x[0]) * ufl.sin(5 * ufl.pi * x[1]), v) * ufl.dx

problem = dolfinx.fem.petsc.LinearProblem(a, L, petsc_options={"ksp_type": "preonly", "pc_type": "lu", "pc_factor_mat_solver_type": "mumps"}, petsc_options_prefix="pp_")
start = time.perf_counter()
problem.solve()
end = time.perf_counter()
print(f"Time to solve: {end - start:.2e} seconds")
start_mid = time.perf_counter()
problem.solver.solve(problem.b, problem.x)
end_mid = time.perf_counter()
print(f"Time to solve (mid): {end_mid - start_mid:.2e}")
start2 = time.perf_counter()
problem.solve()
end2 = time.perf_counter()
print(f"Time to solve (second time): {end2 - start2:.2e} seconds")
```
yields:
```
Time to solve: 2.22e-01 seconds
Time to solve (mid): 3.80e-03
Time to solve (second time): 1.56e-01 seconds
```
However, if you change the options to:
```python
problem = dolfinx.fem.petsc.LinearProblem(a, L, petsc_options={"ksp_type": "preonly", "pc_type": "lu", "pc_factor_mat_solver_type": "mumps","ksp_reuse_preconditioner": True}, petsc_options_prefix="pp_")
```
yields the speedup we would like.
```
Time to solve: 4.89e-01 seconds
Time to solve (mid): 8.83e-03
Time to solve (second time): 9.68e-02 seconds
```
Of course the latter option is for experts only as, this would give the wrong result if the matrix actually changed.

jhale:
Thanks that's useful to know. Nonetheless, that method still  reassembles the matrix A (which is unused and thus a bit wasteful), even if the factors are re-used from the previous A.

Edit: Indeed, I see your timings show the overhead of this path.

chrisrichardson (approved):

jhale:
An aside, but mypy really doesn't play nicely with `functools.singledispatch`.

jhale:
This is now merged. Implementation aside, it would also be possible to place this at `dolfinx.fem.superlu_dist` which mirrors `dolfinx.fem.petsc` - however it's not clear to me this would be so good in terms of discoverability.

PR #4187, opened 2026-04-27
Add equidistribution marking

### LINKED ISSUE #3133: Parallel marking algorithm

The current mesh refinement module does not contain any algorithms for calculating the set of marked cells for subsequent
mesh refinement. In parallel this is non-trivial, and so I think it should be part of the library. Additionally, marking is a generic algorithm independent of the underlying PDE and error estimation strategy.

There is an interesting paper by Pfeiler and Praetorius on Dorfler marking in O(N) time with a C++ implementation which would work for us if it was refactored https://arxiv.org/pdf/1907.13078.pdf - however, I cannot find much in the literature about the parallel context where the naive algorithm would require a single rank having knowledge of all estimators.

### Suggested user interface

The estimators are usually DG_0 so the interface could be ndarray/std::vector based.

schnellerhase:
Implements an (squared) equidistribution marking criterion for AFEM schemes, i.e. for $\theta \in (0, 1)$ and a marker $\eta \in \mathbb{R}^N$, computes

$$
  \mathcal{I} = \left\\{ i \ | \ \eta_i > \theta \frac{|\eta|_2^2}{N} \right\\}.
$$

Work towards https://github.com/FEniCS/dolfinx/issues/4141.

Fixes https://github.com/FEniCS/dolfinx/issues/3133.


cpp/dolfinx/refinement/mark.h
```diff
@@ -39,18 +67,45 @@ std::vector<std::int32_t> mark_maximum(std::span<const T> marker, T theta,
...
+/// @return Indices (local) of marker elements, which satisfy: marker_i ≥ θ
+/// (sum_i marker_i^2)^1/2 / N^1/2.
+template <std::floating_point T>
+std::vector<std::int32_t> mark_equidistribution(std::span<const T> marker,
+                                                T theta, MPI_Comm comm)
+{
+  if ((theta <= 0) || (theta >= 1))
+    throw std::invalid_argument("Theta needs to fullfill 0 < θ < 1.");
+
+  T norm{0};
```
jhale: `T norm = std::inner_product(marker.begin(), marker.end(), marker.begin(), T{0});`

cpp/dolfinx/refinement/mark.h
```diff
@@ -39,18 +67,45 @@ std::vector<std::int32_t> mark_maximum(std::span<const T> marker, T theta,
...
 
-  spdlog::info("Marking (max) {} / {} (local) entities.", indices.size(),
+/// @brief Equidistribution marking of a marker.
+///
+/// @param[in] marker Input marker (local) - usually an error indicator per
+/// entity
+/// @param[in] theta Parameter, 0 < θ < 1
+/// @param[in] comm Communicator over which the total marker is computed.
+/// @return Indices (local) of marker elements, which satisfy: marker_i ≥ θ
+/// (sum_i marker_i^2)^1/2 / N^1/2.
```
jhale: Please correct me if I'm wrong, but there is probably no need to do so many square root operations - assuming markers/indicators and theta are always positive, you can square both sides:

marker_i^2 ≥ θ^2 (sum_i marker_i^2) / N.

python/test/unit/refinement/test_mark.py
```diff
@@ -31,3 +31,26 @@ def test_mark_maximum(theta: float, dtype: np.dtype) -> None:
...
+    cell_count = (cell_im := msh.topology.index_map(tdim)).size_local + cell_im.num_ghosts
+    marker = np.random.default_rng(0).random(cell_count)
+
+    marked_cells = mesh.mark_equidistribution(marker, theta, comm)
+
+    norm = np.sqrt(comm.allreduce(np.sum(marker**2), MPI.SUM))
+    count = comm.allreduce(marker.size)
+    assert np.allclose(
+        marked_cells,
+        np.argwhere(marker > theta * norm / np.sqrt(count)).flatten(),
```
jhale: If my square-rootless version is correct, still keep this alternative version as-is in the test.

jhale (changes requested):

cpp/dolfinx/refinement/mark.h
```diff
@@ -39,18 +67,45 @@ std::vector<std::int32_t> mark_maximum(std::span<const T> marker, T theta,
...
+}
 
-  spdlog::info("Marking (max) {} / {} (local) entities.", indices.size(),
+/// @brief Equidistribution marking of a marker.
+///
+/// @param[in] marker Input marker (local) - usually an error indicator per
+/// entity
+/// @param[in] theta Parameter, 0 < θ < 1
+/// @param[in] comm Communicator over which the total marker is computed.
+/// @return Indices (local) of marker elements, which satisfy: marker_i ≥ θ
```
jhale: Local indices of marked entities.

cpp/dolfinx/refinement/mark.h
```diff
@@ -39,18 +67,45 @@ std::vector<std::int32_t> mark_maximum(std::span<const T> marker, T theta,
...
-    if (mark(marker[i]))
-      indices.push_back(i);
-  }
+  return indices;
+}
 
-  spdlog::info("Marking (max) {} / {} (local) entities.", indices.size(),
+/// @brief Equidistribution marking of a marker.
+///
+/// @param[in] marker Input marker (local) - usually an error indicator per
```
jhale: The terminology is non-standard. Marker is normally denoted eta and called an error indicator per entity.

Furthermore, and see my comment below, it is usually most efficient to take eta_i^2 - this often saves a square root in user code.

cpp/dolfinx/refinement/mark.h
jhale: Can't we use LaTeX in our doxygen? Not a big fan of unicode maths, typesets poorly.

cpp/dolfinx/refinement/mark.h
schnellerhase: That's true but we would sacrifice the shared code path of `mark_threshold`. Since it's 2x sqrt's not critical?

jhale: Not really - in the vast majority of cases I'm familiar with, one naturally assembles the square of the entity estimator, so your approach effectively requires the user to do many square roots outside this function.

jhale: See e.g. https://doi.org/10.1016/j.camwa.2022.11.009 eq A.1, A.4, 3.15 (which should really be squared on both sides).

schnellerhase: That would require changing the input to be interpreted as squared already which complicates the interface: the equidistribution property is based on $\eta$ (not $\eta^2$) so invoking with $\eta^2$ and equilibrating wrt. to $\eta$ is easy to mess up.

~~If one wants to avoid the sqrt's (per entity) on caller side for performance optimisations, one can replace `mark_equidistribution(marker, theta, ...)` with `mark_equidistribution(marker/sqrt(N), theta, ...)`~~

jhale: We only added the other PR yesterday! It would make sense to change the other interface to accept squared entity estimators as well.

schnellerhase: I am concerned about the functionality being correctly interpreted by the user with an invocation with $\eta^2$, not about altering the new interface.

Maximum marking has a similar (simpler) relation allowing the use case already: replacing `mark_maximumum(marker, theta, ...)` with `mark_maximumum(marker**2, theta**2, ...)` since

$$
  \eta_i > \theta \max_j \eta_j 
  \iff 
  \eta_i^2 > \theta^2 \left( \max_j \eta_j  \right)^2 = \theta^2 \max_j \eta_j^2.
$$

I am hesitant to restrict the general interface of markers, which does not only apply to error estimators, but also any other kind of heuristic or indicator function for the sake of eliminating a possible sqrt on the caller side. This feels like premature optimisation to me.

jhale: Could you give a concrete example of a heuristic or other indicator function? Or a reference to the literature? It would help me understand a bit more the use case you have in mind.

I don't follow the arguments on my interface being confusing - the argument should be called eta_squared and what the function expects and does would be fully documented.

jhale: Let's chat about this in person next week - there are quite a few issues in this PR with terminology, even putting aside this squaring debate.

schnellerhase: Examples are refinement for interfaces (shock fronts, parameter discontinuities, cracks, ...) which could be localised through an aux. field that encodes its location (phase field, levelest, ...). Also mesh properties (size of entities, skewness, ...) could be used to mark a selection of elements for further inspection or possible refinement.

jhale: These are not error estimates in the sense of classical AFEM and squaring them, inside or outside the routine, is fundamentally flawed.

schnellerhase: I agree, but they are valid indicators for selecting a region of interest. That's why we should not have an interface which requires squared quantities as input, right?

jhale: Then why are you squaring them on the inside? Both my design and your design imply the suitability of the l^2 norm - in the proper AFEM setting this suitability is entirely justified. For your 'indicators' it is a completely unjustified imposition.

schnellerhase: That would be a users responsibility to ensure that this checks out (if there is any theory on that). The maximum marking is also not assuming an underlying l2 structure and gets heavily used in AFEM schemes (for noisy indicators, I think). The interface should not assume structure on the input data. Any other norm based, or not norm based, measure could be supported.

brodbeck-m: What if we added something like a marker_is_squared flag? While this would have no effect for maximum marking or Dörfler marking, where it is the user’s responsibility to provide either squared or non-squared markers, as well as the correct marking parameter, this could solve the issue with the equidistribution strategy.

schnellerhase: I like the idea - feels like a practical resolution to the problem honouring that the l2 case is the 95% use case and allowing for general interface usage in the meantime. Only I would prefer to provide the additional argument to all functions if we enable it (so also for max marking for now), for a consistent user interface - even though it is not strictly functionally required. What do you think @jhale?

jhale:
Just write two functions, one that takes squared markers, and one that doesn't - writing the docstring for a single function and explaining the differences/behaviour will just add to the confusion. I'm also not a big fan of optional args, easy to overlook.

There is no need to adjust the Dörfler, or the existing maximum marking - the user can pass the p-th power and the correct definition of theta, as @brodbeck-m says. This can be explained in the docstrings.

A further comment:

The terminology 'marker' is not really right for the incoming argument. I would prefer something like "local indicators" or just "indicators" - that would encompass the adhoc approaches mentioned before, and proper AFEM error estimators.

Marking is associated with the selection decision (which is why you use it in your function names, quite correctly).

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,124 @@
 namespace dolfinx::refinement
 {
 
+namespace impl
+{
+
+/// @brief Threshold marking helper
+///
+/// @param[in] marker Indicator \f$ \eta \f$
```
schnellerhase:
```suggestion
/// @param[in] indicator Indicator \f$ \eta \f$
```

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,124 @@
...
+}
+
+} // namespace impl
+
 /// @brief Maximum marking of a marker.
 ///
-/// @param[in] marker Input marker (local) - usually an error indicator per
-/// entity
-/// @param[in] theta Cut off parameter, 0 < θ < 1
+/// @param[in] marker Indicator (local) \f$ \eta \f$ - usually an error
```
schnellerhase:
```suggestion
/// @param[in] indicator Indicator (local) \f$ \eta \f$ - usually an error
```

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,124 @@
...
-    if (mark(marker[i]))
-      indices.push_back(i);
-  }
+  return indices;
+}
 
-  spdlog::info("Marking (max) {} / {} (local) entities.", indices.size(),
+/// @brief Equidistribution marking of a marker.
+///
+/// @param[in] marker Indicator (local) \f$ \eta \f$ - usually an error
```
schnellerhase:
```suggestion
/// @param[in] indicator Indicator (local) \f$ \eta \f$ - usually an error
```

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,124 @@
...
+    if (mark(indicator[i]))
+      indices.push_back(i);
+  }
+
+  return indices;
+}
+
+} // namespace impl
+
 /// @brief Maximum marking of a marker.
```
garth-wells: Not a change in this PR, but that's a really bad docstring.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,124 @@
...
 ///
-/// @param[in] marker Input marker (local) - usually an error indicator per
-/// entity
-/// @param[in] theta Cut off parameter, 0 < θ < 1
+/// @param[in] indicator Indicator (local) \f$ \eta \f$ - usually an error
```
garth-wells: 'error' is a heavily overloaded term, expand comment.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,124 @@
 namespace dolfinx::refinement
 {
 
+namespace impl
+{
+
+/// @brief Threshold marking helper
```
garth-wells: Add more text on what it does.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,124 @@
...
+///
+/// @param[in] indicator Indicator \f$ \eta \f$
```
garth-wells: Doc doesn't say anything. Expand.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,124 @@
...
+/// @param[in] threshold Lower bound for values to mark
+///
+/// @returns indices \f$ i \f$ which satisfy \f$ \eta_i > \text{threshold} \f$.
```
garth-wells: The maths is missing context.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,124 @@
...
+template <std::floating_point T>
+std::vector<std::int32_t> mark_threshold(std::span<const T> indicator,
+                                         T threshold)
+{
+  auto mark = [=](T e) { return e > threshold; };
```
garth-wells: `[=]` is nasty. Use `[threshold]`.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,124 @@
...
-  {
-    if (mark(marker[i]))
-      indices.push_back(i);
-  }
+  return indices;
+}
 
-  spdlog::info("Marking (max) {} / {} (local) entities.", indices.size(),
+/// @brief Equidistribution marking of a marker.
+///
```
garth-wells, line 94: Explanation is docs needed.

garth-wells (dismissed):
Could you take a careful look at the docs? There is very little context to help reader understand.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
-  for (std::int32_t i = 0; i < static_cast<std::int32_t>(marker.size()); ++i)
-  {
-    if (mark(marker[i]))
-      indices.push_back(i);
-  }
+  return indices;
+}
 
-  spdlog::info("Marking (max) {} / {} (local) entities.", indices.size(),
+/// @brief Equidistribution marking of a marker.
```
jhale: Computes equidistribution threshold marking of an indicator.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
+///
+/// @param[in] indicator Indicator (local) \f$ \eta \f$ - usually an error
+/// indicator per entity
+/// @param[in] theta Parameter, \f$ 0 < \theta < 1 \f$
+/// @param[in] comm Communicator over which the total marker is computed.
+/// @return Local indices of marked entities, which satisfy: \f$
+/// \eta_i \geq \theta \frac{|\eta|_2}{\sqrt{N}} \f$.
```
jhale: I would use double || \cdot || for norm, more recognisable for for non-mathematicians.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
+
+  spdlog::info("Marking (equi) {} / {} (local) entities.", indices.size(),
+               indicator.size());
+
+  return indices;
+}
+
+/// @brief Equidistribution marking of a marker with squared input.
+///
+/// @param[in] marker Input marker (local) \f$ \eta^2 \f$ - usually an error
```
jhale: usually a squared error indicator per entity.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
+  norm = std::sqrt(norm);
+
+  std::int32_t count = indicator.size();
+  MPI_Allreduce(MPI_IN_PLACE, &count, 1, dolfinx::MPI::mpi_t<std::int32_t>,
+                MPI_SUM, comm);
+
+  auto indices
+      = impl::mark_threshold<T>(indicator, theta * norm / std::sqrt(count));
+
+  spdlog::info("Marking (equi) {} / {} (local) entities.", indices.size(),
```
jhale: equi -> equidistribution

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
+  MPI_Allreduce(MPI_IN_PLACE, &norm, 1, dolfinx::MPI::mpi_t<T>, MPI_SUM, comm);
+
+  std::int32_t count = marker.size();
+  MPI_Allreduce(MPI_IN_PLACE, &count, 1, dolfinx::MPI::mpi_t<std::int32_t>,
+                MPI_SUM, comm);
+
+  auto indices
+      = impl::mark_threshold<T>(marker, std::pow(theta, 2) * norm / count);
+
+  spdlog::info("Marking (equi) {} / {} (local) entities.", indices.size(),
```
jhale: equi - equidistribution

cpp/test/mesh/refinement/mark.cpp
```diff
@@ -51,3 +51,51 @@ TEMPLATE_TEST_CASE("Mark maximum", "[refinement][mark][maximum]", double, float)
...
+
+TEMPLATE_TEST_CASE("Mark equidistribution empty",
+                   "[refinement][mark][equidistribution]", double, float)
+{
+  std::vector<TestType> marker;
+  auto indices = mark_equidistribution<TestType>(marker, .5, MPI_COMM_WORLD);
+  CHECK(indices.size() == 0);
+}
+
+TEMPLATE_TEST_CASE("Mark equidistribution",
```
jhale, line 68: We don't have an explicit policy on Python vs C++ tests, but for here my feeling is that Python is most appropriate.

python/test/unit/refinement/test_mark.py
```diff
@@ -31,3 +31,28 @@ def test_mark_maximum(theta: float, dtype: np.dtype) -> None:
     msh.topology.create_entities(1)
     marked_edges = mesh.compute_incident_entities(msh.topology, marked_cells, tdim, 1)
     mesh.refine(msh, marked_edges)
+
+
+@pytest.mark.parametrize("theta", [0.2, 0.4, 0.6, 0.8])
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_mark_equidistribution(theta: float, dtype: np.dtype) -> None:
+    msh = mesh.create_unit_square(comm := MPI.COMM_WORLD, n := 10, n, dtype=dtype)
```
jhale: I don't see the need to use the Walrus here, and SonarCloud agrees "":=" operator is confusing in this context".

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
 namespace dolfinx::refinement
 {
 
-/// @brief Maximum marking of a marker.
+namespace impl
```
jhale, line 27: Put this in the public interface? Seems generally useful if there are thresholds we cannot imagine.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
+/// marking routines.
 ///
-/// @param[in] marker Input marker (local) - usually an error indicator per
-/// entity
-/// @param[in] theta Cut off parameter, 0 < θ < 1
+/// @param[in] indicator Entity wise indicator \f$ \eta \f$ to compute threshold
+/// marking of.
+/// @param[in] threshold Lower bound for values to mark
+///
+/// @returns list of indices \f$ i \f$ of indicator entries which satisfy
```
jhale: which -> that

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
+namespace impl
+{
+
+/// @brief Computes a local marking based on thresholding - helper for other
+/// marking routines.
 ///
-/// @param[in] marker Input marker (local) - usually an error indicator per
-/// entity
-/// @param[in] theta Cut off parameter, 0 < θ < 1
+/// @param[in] indicator Entity wise indicator \f$ \eta \f$ to compute threshold
```
jhale: Entity-wise indicator \f$ \eta \f$ used for marking.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
+/// marking of.
+/// @param[in] threshold Lower bound for values to mark
```
jhale: /// @param[in] threshold Threshold value; indicators greater
/// than this are marked.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
 namespace dolfinx::refinement
 {
 
-/// @brief Maximum marking of a marker.
+namespace impl
+{
+
+/// @brief Computes a local marking based on thresholding - helper for other
```
jhale: /// @brief Computes local threshold-based marking.
///
/// Helper for other marking routines.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
+    if (mark(indicator[i]))
+      indices.push_back(i);
+  }
+
+  return indices;
+}
+
+} // namespace impl
+
+/// @brief Compute maximum marking of a marker, i.e. all elements which are
```
jhale: @brief Computes maximum-based marking of an indicator.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
+/// greater equal the markers maximum times a cut off parameter.
```
jhale: Split to new line:

/// Marks all local entries whose indicators are greater than or equal to
/// \f$ \theta \f$ times the maximum indicator value across `MPI_Comm`.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
-    if (mark(marker[i]))
-      indices.push_back(i);
-  }
+  return indices;
+}
 
-  spdlog::info("Marking (max) {} / {} (local) entities.", indices.size(),
+/// @brief Equidistribution marking of a marker.
+///
+/// @param[in] indicator Indicator (local) \f$ \eta \f$ - usually an error
```
jhale: Punctuation throughout.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
+  auto indices
+      = impl::mark_threshold<T>(indicator, theta * norm / std::sqrt(count));
+
+  spdlog::info("Marking (equi) {} / {} (local) entities.", indices.size(),
+               indicator.size());
+
+  return indices;
+}
+
+/// @brief Equidistribution marking of a marker with squared input.
```
jhale: Computes equidistribution threshold marking of a squared indicator.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
+  return indices;
+}
 
-  spdlog::info("Marking (max) {} / {} (local) entities.", indices.size(),
+/// @brief Equidistribution marking of a marker.
+///
+/// @param[in] indicator Indicator (local) \f$ \eta \f$ - usually an error
+/// indicator per entity
+/// @param[in] theta Parameter, \f$ 0 < \theta < 1 \f$
+/// @param[in] comm Communicator over which the total marker is computed.
```
jhale: Communicator over which the equidistribution threshold is computed.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
+
+  return indices;
+}
+
+/// @brief Equidistribution marking of a marker with squared input.
+///
+/// @param[in] marker Input marker (local) \f$ \eta^2 \f$ - usually an error
+/// indicator per entity
+/// @param[in] theta Parameter, \f$ 0 < \theta < 1 \f$
+/// @param[in] comm Communicator over which the total marker is computed.
```
jhale: Communicator over which the equidistribution threshold is computed.

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
+  T max = indicator.empty() ? std::numeric_limits<T>::lowest()
+                            : std::ranges::max(indicator);
   MPI_Allreduce(MPI_IN_PLACE, &max, 1, dolfinx::MPI::mpi_t<T>, MPI_MAX, comm);
 
-  auto mark = [=](T e) { return e > theta * max; };
+  auto indices = impl::mark_threshold<T>(indicator, theta * max);
 
-  std::vector<std::int32_t> indices;
-  indices.reserve(std::ranges::count_if(marker, mark));
+  spdlog::info("Marking (max) {} / {} (local) entities.", indices.size(),
```
jhale: max -> maximum

jhale (changes requested):
Comments for improved docstring clarity.

python/dolfinx/mesh.py
```diff
@@ -35,6 +35,8 @@
 from dolfinx.cpp.refinement import (
     IdentityPartitionerPlaceholder,
     RefinementOption,
+    mark_equidistribution,
```
jhale: What do the docstrings look like on these?

cpp/dolfinx/refinement/mark.h
jhale: which -> that

cpp/dolfinx/refinement/mark.h
```diff
@@ -20,37 +20,128 @@
...
+
+} // namespace impl
+
+/// @brief Compute maximum marking of a marker, i.e. all elements which are
+/// greater equal the markers maximum times a cut off parameter.
+///
+/// @param[in] indicator Indicator (local) \f$ \eta \f$ - usually an entity wise
+/// error indicator.
+/// @param[in] theta Cut off parameter, \f$ 0 < \theta < 1 \f$.
 /// @param[in] comm Communicator over which the maximum is computed.
```
jhale: Communicator to compute the maximum over.

jhale (dismissed):

jhale:
Another comment - I could not find another method in DOLFINx where the MPI Communicator is last, it is always first.

cpp/test/mesh/refinement/mark.cpp
schnellerhase: The Python ones do the big parameter sweep, but C++ tests allow for faster dev turn around times - would really want to keep them.

python/dolfinx/mesh.py
schnellerhase: As long as we have no stubs exported, there are none.

cpp/dolfinx/refinement/mark.h
schnellerhase: `impl::mark_threshold` is also an implementation detail of a possible user facing `mark_threshold`, i.e. with input checks and logging. `mark_threshold` can be added in another PR if desired.

jhale:
Did some small revisions and proofing this morning - happy for this to go in now @schnellerhase

cpp/dolfinx/refinement/mark.h
```diff
@@ -8,50 +8,162 @@
...
+/// usually an error indicator associated with mesh entity \f$ i \f$.
+/// @param[in] threshold Threshold value; indicators greater than this are
+/// marked.
+/// @return Local indices of marked entities.
+template <std::floating_point T>
+std::vector<std::int32_t> mark_threshold(std::span<const T> indicators,
+                                         T threshold)
+{
+  std::vector<std::int32_t> indices;
+  indices.reserve(indicators.size());
```
schnellerhase: This will over allocate. Why was the `count_if` removed?

jhale: Your version moves across the data twice - once for the `count_if` and once for the index push back. For large meshes memory-bandwidth is the bottleneck, and the extra memory usage is a rounding error for any 'normal' usage of theta (0.3-0.5 is typical). Furthermore the proposed version is simpler.

schnellerhase: We do not expect the marked elements to be fairly distributed across processes. The decision of over estimating memory should not be done silently under the hood - there is no way of turning it off for a caller of this code.

Its also how we do it for similar setups in https://github.com/FEniCS/dolfinx/blob/8c8f7d6efebbf42507b2fa7d55e1ff5ed683b18f/cpp/dolfinx/mesh/MeshTags.h#L88 and https://github.com/FEniCS/dolfinx/blob/8c8f7d6efebbf42507b2fa7d55e1ff5ed683b18f/cpp/dolfinx/fem/DirichletBC.h#L131.

jhale: Fair point - I'll change it back.

jhale (changes requested):
Threshold computation (usually, a global step on owned indicators) and marking (local, on owned and shared indicators) should be split into two separate steps.

garth-wells:
I'm struggling with the PR description and the code doc strings due to the limited context provided.

cpp/dolfinx/refinement/mark.h
schnellerhase: Removed the non squared version for now to simplify. Can be re-added in the future if necessary.

schnellerhase:
This is good to go for another round of reviews after the fix of the ghost issues is now resolved with the merge of https://github.com/FEniCS/dolfinx/pull/4476. The code paths of maximum and equidistribution share a lot - to be improved on in the future, once the API is finalized.

cpp/dolfinx/refinement/mark.h
```diff
@@ -97,4 +98,69 @@ std::vector<std::int32_t> mark_maximum(std::span<const T> values,
...
 }
 
+/// @brief Computes equidistribution threshold marking of a squared indicator.
+///
+/// Returns the indices \f$i\f$ of the squared indicators \f$ \eta_i^2 \f$ that
+/// satisfy the equidistribution threshold: \f$ \eta_i^2 > \theta^2
+/// \frac{||\eta||^2}{N} \f$ where \f$ N \f$ is the (global) number of
+/// indicators.
+///
+/// @warning `squared_indicators` must contain owned entities only. Ghost
```
jhale: Remove?

cpp/dolfinx/refinement/mark.h
```diff
@@ -97,4 +98,69 @@ std::vector<std::int32_t> mark_maximum(std::span<const T> values,
...
+    throw std::invalid_argument(
+        std::format("values must have size index_map.size_local() + "
+                    "index_map.num_ghosts() = {}, got {}.",
+                    size, n));
+  }
+
+  T norm = std::accumulate(values.begin(),
+                           values.begin() + index_map.size_local(), T{0});
+
+  MPI_Allreduce(MPI_IN_PLACE, &norm, 1, dolfinx::MPI::mpi_t<T>, MPI_SUM,
```
jhale: Remove in place reduction.

cpp/dolfinx/refinement/mark.h
```diff
@@ -97,4 +98,69 @@ std::vector<std::int32_t> mark_maximum(std::span<const T> values,
   return indices;
 }
 
+/// @brief Computes equidistribution threshold marking of a squared indicator.
```
jhale: I can harmonize the docstrings.

cpp/dolfinx/refinement/mark.h
```diff
@@ -97,4 +98,69 @@ std::vector<std::int32_t> mark_maximum(std::span<const T> values,
...
+  std::vector<std::int32_t> indices;
+  indices.reserve(std::ranges::count_if(values, mark));
+  for (std::int32_t i = 0; i < n; ++i)
+  {
+    if (mark(values[i]))
+      indices.push_back(i);
+  }
+
+  spdlog::info(
+      "Marking (equidistribution): marked {} of {} local entries (owned + "
```
jhale, line 173: Harmonise with maximum wording - unclear as is.

python/dolfinx/mesh.py
```diff
@@ -975,6 +977,42 @@ def mark_maximum(
     return _mark_maximum(values, index_map, theta)
 
 
+def mark_equidistribution(
+    values: npt.NDArray[Real],
+    index_map: _IndexMap,
+    theta: float,
+) -> npt.NDArray[np.int32]:
+    r"""Return local indices of values exceeding a fraction of the MS.
```
jhale: What is MS? Avoid acronym.

jhale (commented):
Deleted comment, misled by wrong docstring.

cpp/dolfinx/refinement/mark.h
```diff
@@ -97,4 +98,69 @@ std::vector<std::int32_t> mark_maximum(std::span<const T> values,
...
+  std::int32_t size = index_map.size_local() + index_map.num_ghosts();
+  if (n != size)
+  {
+    throw std::invalid_argument(
+        std::format("values must have size index_map.size_local() + "
+                    "index_map.num_ghosts() = {}, got {}.",
+                    size, n));
+  }
+
+  T norm = std::accumulate(values.begin(),
```
jhale:
```
auto owned = values.first(index_map.size_local());
T local_squared_norm = std::accumulate(owned.begin(), owned.end(), T{0});
```

cpp/dolfinx/refinement/mark.h
jhale: squared_norm

cpp/dolfinx/refinement/mark.h
```diff
@@ -97,4 +98,69 @@ std::vector<std::int32_t> mark_maximum(std::span<const T> values,
   return indices;
 }
 
+/// @brief Computes equidistribution threshold marking of a squared indicator.
+///
+/// Returns the indices \f$i\f$ of the squared indicators \f$ \eta_i^2 \f$ that
+/// satisfy the equidistribution threshold: \f$ \eta_i^2 > \theta^2
+/// \frac{||\eta||^2}{N} \f$ where \f$ N \f$ is the (global) number of
+/// indicators.
```
jhale: I think this needs the same note about values being scatter forwarded.

python/dolfinx/mesh.py
```diff
@@ -975,6 +977,42 @@ def mark_maximum(
...
+
+    Args:
+        values: Values, often with each entry associated with a mesh
+            entity, e.g. an error indicator (squared per entity).
+        index_map: Index map describing the parallel layout of ``values``.
+        theta: Cut-off parameter, :math:`0 < \theta \leq 1`.
+
+    Returns:
+        Local indices, ascending and including ghosts, of the entries
+        satisfying :math:`v_i > \theta^2 \frac{||v||_2}{N}`.
```
jhale: Inconsistency

cpp/test/mesh/refinement/mark.cpp
```diff
@@ -76,3 +76,66 @@ TEMPLATE_TEST_CASE("Mark maximum", "[refinement][mark][maximum]", double, float)
     CHECK(expect_marked == marked);
   }
 }
+
+TEMPLATE_TEST_CASE("Mark equidistribution empty",
```
jhale: Shall I quickly port these to Python? Don't see much that needs C++

schnellerhase: They could equally be in Python, but having them in C++ allows for faster dev cycles.

jhale: Ok then I will port to Python.

jhale (approved):
Found a logic bug in the C++ test while porting it across. The magnitude of the ghost entry must be much larger than the owned entry for the test to be effective.

PR #4191, opened 2026-04-30
Fix conda weekly run

jhale:


jhale:
Not working, will ask around for help.

Issue #4098, opened 2026-02-20
Parametrise mixed topology/geometry tests over datatypes


schnellerhase:
Ref https://github.com/FEniCS/dolfinx/pull/4097.

Siya-05:
Can I work on this issue?

jorgensd:
@Siya-05, sure feel free to work on this. Just bear in mind that the mixed topology interface changes rapidly, so we should aim for small PRs that can be merged quickly.

PR #4194, opened 2026-04-30
dtype coverage for mixed topology tests

Siya-05:
## Summary

Add dtype coverage (`float32` and `float64`) to selected mixed-topology mesh tests.

## Changes

- Parametrized:
  - `test_create_mixed_mesh`
  - `test_mixed_topology_mesh`

- Updated:
  - Geometry arrays to use `dtype`
  - `coordinate_element(..., dtype=dtype)`

- Kept:
  - Topology/index arrays as `int64` (unchanged)

## Rationale

- These tests operate at the mesh/geometry level and are naturally dtype-agnostic, making them suitable for dtype parametrization.
- FEM-level tests (e.g., dofmap tests) were explored but not included due to current limitations:
  Basix UFL elements default to `float64`, leading to dtype mismatches when using `float32`.

### DIFF
--- a/python/test/unit/mesh/test_create_mixed_mesh.py
+++ b/python/test/unit/mesh/test_create_mixed_mesh.py
@@ -1,6 +1,7 @@
 from mpi4py import MPI
 
 import numpy as np
+import pytest
 
 from dolfinx.cpp.mesh import (
     GhostMode,
@@ -10,7 +11,8 @@ from dolfinx.fem import coordinate_element
 from dolfinx.mesh import CellType, create_cell_partitioner
 
 
-def test_create_mixed_mesh():
+@pytest.mark.parametrize("dtype", [np.float32, np.float64])
+def test_create_mixed_mesh(dtype):
     nx = 7
     ny = 11
     nz = 8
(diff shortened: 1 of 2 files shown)


schnellerhase:
Ref https://github.com/FEniCS/dolfinx/issues/4098.

python/test/unit/mesh/test_create_mixed_mesh.py
```diff
@@ -70,15 +72,15 @@ def test_create_mixed_mesh():
         geom += [[ix / nx, iy / ny, iz / nz]]
 
     if MPI.COMM_WORLD.rank == 0:
-        cells_np = [np.array(c, dtype=np.int64) for c in cells]
-        geomx = np.array(geom, dtype=np.float64)
+        cells_np = [np.array(c) for c in cells]
```
schnellerhase:
```suggestion
        cells_np = [np.array(c, dtype=np.int64) for c in cells]
```

python/test/unit/mesh/test_create_mixed_mesh.py
```diff
@@ -70,15 +72,15 @@ def test_create_mixed_mesh():
...
+        geomx = np.array(geom, dtype=dtype)
     else:
-        cells_np = [np.zeros(0, dtype=np.int64) for c in cells]
-        geomx = np.zeros((0, 3), dtype=np.float64)
+        cells_np = [np.zeros(0) for c in cells]
```
schnellerhase:
```suggestion
        cells_np = [np.zeros(0, dtype=np.int64) for c in cells]
```

schnellerhase (approved):

schnellerhase:
Thanks, almost all done 😄

Siya-05:
Thanks for the quick review! Really appreciate it.

PR #4195, opened 2026-04-30
conda: avoid defaults channel

minrk:
updates setup-conda on ci to use miniforge instead of miniconda, which avoids getting anything from the defaults (Anaconda) channel. Also switches the installer to mamba>=2.6, which enables sharded repo data (less downloads, should be faster solves).

I'm not sure what's going on in #4191, but it's always a good idea to make sure channels aren't mixed before diving too deep into debugging conda installs.

closes #4191 if it works

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -1,10 +1,16 @@
 name: Conda install
 
 on:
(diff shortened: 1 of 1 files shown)


minrk:
seems to be working

PR #4197, opened 2026-05-05
Fix Spack builds and Clang 22 with older version of Catch2

jhale:
Issue 1:

See https://github.com/catchorg/Catch2/blob/devel/docs/release-notes.md#3140.

"Suppressed the new Clang warning about __COUNTER__ usage. (#3076)"

This raises a warning which results in an error due to our strict compiler flags. Can be removed once Catch 3.14 is available in Spack.

Issue 2:

Current `develop` branch of Spack tools cannot correctly push/pull from the GitHub OCI registry. Pinning to v1.1.1 fixes this. It would probably be best to use stable releases of Spack tooling

### DIFF
--- a/.github/workflows/ci-spack.yml
+++ b/.github/workflows/ci-spack.yml
@@ -40,7 +40,7 @@ jobs:
         uses: actions/checkout@v6
         with:
           repository: spack/spack
-          ref: develop
+          ref: v1.1.1
           path: spack-src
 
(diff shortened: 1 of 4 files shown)


jhale:
This was picked up by wiping the RedHat Spack build cache and forcing a run - that led to this issue being discovered (i.e. running on a clean cache). However, the Spack build cache is now failing again.

PR #4203, opened 2026-05-07
Allow built-in meshes to be embedded in gdim >= tdim

jhale:
This is useful for working on manifold meshes without having to bring in an external mesh generator. Users can push geometry through an appropriate mapping. A next obvious step would be to create these "flat" meshes with higher-order coordinate elements, but that is a more substantial refactor.

### DIFF
--- a/cpp/dolfinx/mesh/generation.h
+++ b/cpp/dolfinx/mesh/generation.h
@@ -291,12 +291,14 @@ create_interval(MPI_Comm comm, std::int64_t n, std::array<T, 2> p,
     for (std::size_t i = 0; i < npts; i++)
       x[i * gdim] = x1d[i];
     return create_mesh(comm, MPI_COMM_SELF, cells, element, MPI_COMM_SELF, x,
-                       {npts, gdim}, partitioner, 2, reorder_fn);
+                       {npts, static_cast<std::size_t>(gdim)}, partitioner, 2,
+                       reorder_fn);
   }
   else
   {
     return create_mesh(comm, MPI_COMM_NULL, {}, element, MPI_COMM_NULL,
-                       std::vector<T>{}, {0, gdim}, partitioner, 2, reorder_fn);
+                       std::vector<T>{}, {0, static_cast<std::size_t>(gdim)},
+                       partitioner, 2, reorder_fn);
   }
 }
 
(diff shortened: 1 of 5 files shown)


cpp/dolfinx/mesh/generation.h
```diff
@@ -45,13 +45,14 @@ template <std::floating_point T>
 Mesh<T> build_tri(MPI_Comm comm, std::array<std::array<T, 2>, 2> p,
                   std::array<std::int64_t, 2> n,
                   const CellPartitionFunction& partitioner,
-                  DiagonalType diagonal, const CellReorderFunction& reorder_fn);
+                  DiagonalType diagonal, const CellReorderFunction& reorder_fn,
+                  std::size_t gdim);
```
jhale: Tacking this on the end seemed like the least API breaking way to proceed given that the majority of people won't be using this `gdim` argument.

cpp/dolfinx/mesh/generation.h
```diff
@@ -241,14 +253,19 @@ Mesh<T> create_rectangle(MPI_Comm comm, std::array<std::array<T, 2>, 2> p,
 /// @param[in] partitioner Partitioning function for distributing cells
 /// across MPI ranks.
 /// @param[in] reorder_fn Function for (locally) reordering cells
+/// @param[in] gdim Geometric dimension. Must be >= 1. The interval lies
```
jhale, line 252: This is a slight simplification, but I consider the geometry always being in a three-dimensional space an implementation detail.

cpp/dolfinx/mesh/generation.h
jhale: Actually I'm going to change this - `reorder_fn` will be tricky to use...

jorgensd:
> This is useful for working on manifold meshes without having to bring in an external mesh generator. Users can push geometry through an appropriate mapping. A next obvious step would be to create these "flat" meshes with higher-order coordinate elements, but that is a more substantial refactor.

To make them higher order is quite easy. We could borrow the functionality from:
https://github.com/scientificcomputing/io4dolfinx/blob/27d97a70dcd1bfd04fa5d2dfe9ef8a660b9f2028/src/io4dolfinx/utils.py#L204-L260
where I already push forward linear grids to curved grids and vice-versa.

jorgensd (approved):

michalhabera (commented):
I'd personally:
1. use `int` for `gdim`. The `std::size_t` is more to store a size of a container.
2. have an if-elseif-else logic for gdim where the padding is done. It currently allows to call with gdim=4.

Other than that looks ok and useful.

PR #4205, opened 2026-05-07
Create new mesh with interpolation of existing mesh geometry

jhale:
Adds a single free function that can create a new mesh by raising or lowering the polynomial order of an existing mesh. Useful for elevating the output of built-in mesh creation functions, or for lowering to allow output on backends that only support low-order geometry.

The topology is shared at the memory level between the new and old mesh.

In terms of header spread, this function sits much happier in `dolfinx.fem` than `dolfinx.mesh` as it largely uses `dolfinx.fem` functionality.

Claude Sonnet was used for some large refactorings at the end.

To avoid a circular import I also refactored various free functions into `fem/utils.py` from `fem/__init__.py` file, which seems cleaner in any case.


cpp/dolfinx/mesh/utils.cpp
```diff
@@ -168,3 +177,70 @@ mesh::compute_incident_entities(const Topology& topology,
...
+        "Cell shape of new coordinate element must match input mesh.");
+  }
+
+  const int gdim = mesh->geometry().dim();
+
+  // Build a vector-valued Lagrange FiniteElement from the coordinate element.
+  basix::FiniteElement<T> b_element = basix::create_element<T>(
+      basix::element::family::P,
+      mesh::cell_type_to_basix_type(new_cmap.cell_shape()), new_cmap.degree(),
+      new_cmap.variant(), basix::element::dpc_variant::unset, false);
```
jorgensd: I was thinking about this on the way home, and could we potentially here make a DG geometry if we would like to, i.e. have an input option for setting discontinuous to True here?

jorgensd:
> Adds a single free function that can p-refine (and possibly coarsen?) an existing mesh.
> 
> Testing a bit limited: P1 to P2 and P1 to P1. Ideas welcome.

See for instance the tests in IO4Dolfinx that goes from 1->3->1 and 1->2->1:
https://github.com/scientificcomputing/io4dolfinx/blob/main/tests/test_mesh.py#L11-L56

jorgensd:
It's useful for IO as well, as it would allow you to read/write point data on higher order grids (and then linearize the grid after the fact).

Additionally, as the meshes share topology, it is easy to do interpolation between the two grids.

PR #4211, opened 2026-05-13
Add AI contribution guide

michalhabera:
AI assistance:
I used ChatGPT 5.5 to draft parts of this PR. I reviewed, edited, tested, and take responsibility for the final contribution.


francesco-ballarin (approved):

jorgensd (approved):
LGTM!

cdaversin (approved):

jhale:
Looks good; let's see how it goes over the next 6 months and then we can update.

jhale (approved):

michalhabera:
> Looks good; let's see how it goes over the next 6 months and then we can update.

Very true. Even if all contributions will be AI assisted in the near future, is would be good to keep the responsibility and model disclosure.

IgorBaratta:
One small suggestion, I think it would be helpful to explicitly mention the licensing risks associated with AI coding assistants. The main concern is that an AI might inadvertently reproduce restrictively licensed code (like GPL) into the repo. Perhaps making it explicit that the contributor is responsible for verifying this?

mscroggs (approved):

nate-sime (approved):

michalhabera:
> One small suggestion, I think it would be helpful to explicitly mention the licensing risks associated with AI coding assistants. The main concern is that an AI might inadvertently reproduce restrictively licensed code (like GPL) into the repo. Perhaps making it explicit that the contributor is responsible for verifying this?

There is the statement:
> The contributor must:
> ...
> - ensure that licensing requirements are satisfied,

Is this not enough?

IgorBaratta:
> > One small suggestion, I think it would be helpful to explicitly mention the licensing risks associated with AI coding assistants. The main concern is that an AI might inadvertently reproduce restrictively licensed code (like GPL) into the repo. Perhaps making it explicit that the contributor is responsible for verifying this?
> 
> There is the statement:
> 
> > The contributor must:
> > ...
> > 
> > * ensure that licensing requirements are satisfied,
> 
> Is this not enough?

You're right, I missed that. 
I still think there's value in being explicitly clear about AI in this specific context, but happy if you merge without.

jhale:
I agree @IgorBaratta, but how should a contributor practically verify this?

IgorBaratta:
Good point, @jhale.  
I think it's hard to verify unless the contributor actively runs their own plagiarism checks or relies on filters (like [Copilot's public code blocker](https://docs.github.com/en/copilot/how-tos/get-code-suggestions/find-matching-code?tool=webui)), which are not available in most AI assistants.
Maybe "ensure licensing requirements are satisfied" is the best we can do for now without overcomplicating and can always revisit it in 6 months as you suggested!

PR #4214, opened 2026-05-14
Mixed topology assemble scalar and gmsh api + fix GMSH tests on AlmaLinux

### LINKED ISSUE #4213: [BUG]: Non-reproducible gmsh recombination issue in `test_higher_order.py`
### Summarize the issue

This seems to appear only on ArchLinux/Spack using `pip install gmsh`.

The recombination appears to fail; this leads to tets and hexes, which the test doesn't support. The 'fix' is to throw an xfail on more than one cell type appearing in the mesh. This of course could be removed once the multiple cell type support is finished (infact, this seems like an ideal test case).

https://github.com/FEniCS/dolfinx/blob/main/python/test/unit/mesh/test_higher_order_mesh.py#L770
https://github.com/FEniCS/dolfinx/blob/main/python/test/unit/mesh/test_higher_order_mesh.py#L849

### How to reproduce the bug

Difficult - only way to recreate at the moment is to remove the xfail and run CI.

### Minimal Example (Python)

```Python

```

### Output (Python)

```bash

```

### Version

main branch

### DOLFINx git commit

_No response_

### Installation

_No response_

### Additional information

_No response_

jorgensd:
Attempt to resolve: https://github.com/FEniCS/dolfinx/issues/4213.
Seems like running pytests in xdist mode requires all gmsh models to have unique names. This has now been added.

Adds support for:
- `dolfinx.fem.assemble_scalar` on mixed topology meshes.
- Second order prism and pyramids (the ones currently supported are not the ones we use for Lagrange elements, as they are missing the middle node of the quadrilateral faces, similar to serendipity).
- Split tests in two:
   - simplex meshes tested for order 1,2,3
   - mixed meshes tested for order 1 and 2

- Supports reading in meshes with model_to_mesh for mixed meshes. No meshtags are read in as we do not have a way of constructing these yet.

- Mixed meshes with order > 2 is not supported, as create_entity_permutations doesn't work on mixed meshes.

Gemini Pro was used to create the 3D mixed mesh with two cylinders on top of each other. All other coding was done by me with no LLM.

The test now uses all 4 cell types for the 3D test (tetra green, pyramid orange, hex red, wedge blue):
image


python/dolfinx/io/gmsh.py
```diff
@@ -361,73 +357,106 @@ def model_to_mesh(
...
+                coordinate_element(
+                    dolfinx_ct,
+                    ufl_domain.ufl_coordinate_element().degree,
+                    variant=basix.LagrangeVariant.equispaced,
+                    dtype=dtype,
+                )._cpp_object
+            )
+        # The mixed topology constructor is not great at the moment
+        # import dolfinx
+        # dolfinx.log.set_log_level(dolfinx.log.LogLevel.DEBUG)
```
nate-sime:
```suggestion
```

Old remnant? I'm surprised ruff didn't pick this up.

jorgensd: Yes.

jorgensd: Partially, the debug bit should go, but the mixed topology dolfinx constructor is not great. That is due to the fundamental issues that ufl doesn't support a `ufl.Mesh` with more than one coordinate element.

jorgensd:
```suggestion
```

jorgensd:
@jhale this still occasionally crashes on alma linux. Now I'm not sure why, as our API should support the meshes generated:
https://github.com/FEniCS/dolfinx/actions/runs/25905433471/job/76137968460

jorgensd:
> @jhale this still occasionally crashes on alma linux. Now I'm not sure why, as our API should support the meshes generated: https://github.com/FEniCS/dolfinx/actions/runs/25905433471/job/76137968460

Resolved by giving each gmsh model a name that is unique for parametrized tests.

jhale:
Thanks for fixing this and also finishing a cool new feature!

My view is that `gmsh.initialize()` and `gmsh.finalize()` are not properly cleaning up global state between pytest function calls on the same process and so we should probably add this to *all* gmsh-related tests.

Issue #4059, opened 2026-02-01
Adapt to availability of `petsc4py` stubs


schnellerhase:
Upstream stub support for `petsc4py` has been added, [ref](https://gitlab.com/petsc/petsc/-/merge_requests/8366).  With the next release `v3.25.0` we will therefore have access to stubs rendering the related `type: ignore`'s superfluous.

(`slepc4py` support [pending](https://gitlab.com/slepc/slepc/-/merge_requests/934))

Would be good to check these before the release happens.

As part of resolution this should revert https://github.com/FEniCS/basix/pull/1004.

jhale:
Ah this is great news.

Our stub workflow also needs looking at (checking that we are using auto-generated stubs, and that the output of those stubs is good).

mscroggs:
The Basix CI (see eg https://github.com/FEniCS/basix/actions/runs/24659086230/job/72100042177) is picking up a lot of mypy errors coming from dolfinx/fem/petsc.py

PR #4216, opened 2026-05-15
Mypy check on non-isolated install

schnellerhase:
Adds mypy check to run on full fledged (non-PETSc) install. This allows dependency lookups and runtime awareness which enriches coverage of the `mypy` invocation.

This was previously done in the BASIx CI already, which caught in multiple occasions more strict typing errors https://github.com/FEniCS/basix/blob/main/.github/workflows/dolfinx-tests.yml#L86.

This should be extended to also check the `test/` and `demo/` source trees (and alter to the PETSc aware runs), to follow in another PR as this hot-fixes the BASIx CI.

Ref https://github.com/FEniCS/dolfinx/issues/4059#issuecomment-4287461055, accompanies https://github.com/FEniCS/basix/pull/1004.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -2,10 +2,10 @@ name: CI
 
 on:
   push:
-    branches:
-      - release
-    tags:
-      - "v*"
+    # branches:
+    #   - release
+    # tags:
+    #   - "v*"
   pull_request:
     branches:
       - main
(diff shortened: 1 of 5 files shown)


.github/workflows/ccpp.yml
```diff
@@ -2,10 +2,10 @@ name: CI
...
 on:
   push:
-    branches:
-      - release
-    tags:
-      - "v*"
+    # branches:
+    #   - release
+    # tags:
+    #   - "v*"
```
```suggestion
    branches:
      - release
    tags:
      - "v*"
```

jorgensd (approved):
I'm fine with this. Its a shame that we can't separate typing from a full-blown installation.

schnellerhase:
Agreed, it would maybe make sense to move these install dependant linting checks outside of the general CI into an independent job. But that's something for the CI redesign if at all.

PR #4220, opened 2026-05-26
Add nanobind stubgen

jhale:
Adds automatic nanobind stubgen for `dolfinx.cpp` and adds a way to disable automatic generation - I predict stubgen could cause some problems in particular build systems, so good to be able to disable it.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -128,9 +128,15 @@ jobs:
         run: cmake --build . --target test
 
       - name: Build Python interface
-        run: |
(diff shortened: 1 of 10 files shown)


jhale:
Merged and supersedes #4101

jhale:
Unsurprisingly `Found 365 errors in 18 files (checked 29 source files)` - unlikely to be merged before release of 0.11.

jhale:
The stubgen fails silently on Windows as it is unable to import `dolfinx.cpp`. The tests run fine and the same strategy works with Basix. Difficult to debug without local environment.

garth-wells:
Superseded by #4342.

PR #4231, opened 2026-06-03
Align Functionspace.dofmap(s) interface

schnellerhase:
Aligns `Functionspace.dofmaps` interface to geometry access patter, ref https://github.com/FEniCS/dolfinx/pull/4227. Otherwise current access patterns of `mesh.geometry.dofmaps` and `functionspace.dofmaps` do confusingly not align anymore.

- does not deprecate (python) `FunctionSpace.dofmap`, since also c++ layer still relies on it.

### DIFF
--- a/cpp/dolfinx/fem/DirichletBC.h
+++ b/cpp/dolfinx/fem/DirichletBC.h
@@ -339,7 +339,7 @@ public:
           "Rank mismatch between Constant and function space in DirichletBC");
     }
 
-    if (g->value.size() != (std::size_t)_function_space->dofmaps(0)->bs())
+    if (g->value.size() != (std::size_t)_function_space->dofmaps()[0]->bs())
     {
       throw std::runtime_error(
           "Creating a DirichletBC using a Constant is not supported when the "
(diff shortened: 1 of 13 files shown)


chrisrichardson (commented):
We should use .front and .at instead of []...

schnellerhase:
I understand the `.at`. Why `.front` over `.at(0)`?

jhale:
`.front()` is semantically correct from an STL perspective, works across all container types, and is cheaper (no bounds check), although this latter point is of course marginal in this non-hot loop setting.

schnellerhase:
The question is whether we want to use `.at(0)` (bound-checked) or `.front` (not bound checked). Since every other access (and current usage) is bound checked through `.at` for consistency I think we should then stick to `.at(0)` unless there is another concern here?

chrisrichardson:
We have used front elsewhere, so to be consistent I guess we should stick with that. It's not likely (or possible?) to have no dofmap.

jhale:
As an aside, it's better/idiomatic to guard `v.front()` with `!v.empty()`, if needed.

PR #4242, opened 2026-06-15
Replace `cmake-format` with `gersemi`

schnellerhase:
`cmake-format`  has been unmaintained for 5 years, see https://github.com/cheshirekow/cmake_format. [`gersemi`](https://github.com/BlankSpruce/gersemi) is a newer (maintained) tool. We have already been using this tool for the last year without complaints.

- formatting applied throughout
- `gersemi` check is now also blocking


jhale:
Indeed cmake-format needs to go... Could you do a run for us on this branch so we can see what the output is like? Also, maybe do Basix first?

schnellerhase:
You mean a `gersemi` run? I fully applied the changes here.

schnellerhase:
Running `gersemi --check . ` does seem to recurse the directory structure for me, what is the issue in the CI? Also tested it for basix at https://github.com/FEniCS/basix/tree/schnellerhase/gersemi.

PR #4248, opened 2026-06-26
Build Linux and macOS binary wheels

jhale:
These wheels are compatible with Python 3.12+ via abi3 and contain SuperLU_DIST and Scotch (no PETSc). 

No MPI libraries or executables are included - on Linux users can use any MPICH-like ABI and on macOS any OpenMPI ABI.

Example of use:

```
# download the platform wheelhouse artifact and unzip
pip install fenics_*.whl
pip install mpich # openmpi on macOS
```

All tests pass. 

This can be streamlined further once MPI5 ABI is available and we can produce one wheel per platform.

In the short term we can setup our own repo at fenicsproject.org and advertise through the README, and then push to pypi.org for 0.12 if we're happy.

Probably won't go for Windows too much maintenance - let's improve the conda packages further.

Artifacts here:

https://github.com/FEniCS/dolfinx/actions/runs/28248343437


jhale:
These recipes are now sufficiently good for a public beta distributed from `pub.fenicsproject.org` without any long-term commitment on our part.

For Windows it will still be necessary to write a vcpkg file for SuperLU_DIST, and modify the upstream SCOTCH vcpkg file to build the Parmetis compatibility layer.

PR #4249, opened 2026-06-28
Check PETSc errors and validate BC assembly

masaqif:
Summary:
- Check return codes for additional PETSc/SLEPc calls in NewtonSolver, la::petsc wrappers, SLEPcEigenSolver, and fem PETSc vector helpers.
- Reject Dirichlet boundary conditions for same-mesh single bilinear form assembly when trial/test spaces differ, while preserving mixed-domain/submesh assembly.
- Let the ADIOS2 optional VTX test skip before importing VTXWriter.

Related issues: #3995, #3988, #4045.

Testing:
- git diff --check
- /opt/homebrew/bin/python3.11 -m py_compile python/dolfinx/fem/assemble.py python/dolfinx/fem/petsc.py python/test/unit/fem/test_assembler.py python/test/unit/io/test_adios2.py
- PYTHONPATH=/tmp/dolfinx-pr-ruff /opt/homebrew/bin/python3.11 -m ruff check .  (from python/)
- PYTHONPATH=/tmp/dolfinx-pr-ruff /opt/homebrew/bin/python3.11 -m ruff format --check .  (from python/)
- Not run: focused pytest, because pytest is not installed for the available Python 3.11.
- Not run: C++ build/clang tooling, because PETSc/SLEPc, clang-tidy, and clang-format are not available locally.

Issue #3995, opened 2025-11-18
[BUG]: test_dg_0_data fails when VTXWriter (adios2) is not available


drew-parsons:
### Summarize the issue

VTXWriter is optional (adios2 does not support 32-bit systems for instance).
PR #3375 introduced DG0 support tested by `test_dg_0_data`.

The test checks if adios2 is available, skips if it is not, but only after trying to import VTXWriter. i.e. the test [fails](https://buildd.debian.org/status/fetch.php?pkg=fenics-dolfinx&arch=armhf&ver=1%3A0.10.0.post2-1exp2&stamp=1763482798&raw=0) (is not skipped) when adios2 (and therefore VTXWriter) is not available.

The skip check for adios2 should be performed before importing VTXWriter.

### How to reproduce the bug

pytest-3 -k test_dg_0_data test/unit/io/test_adios2.py

### Minimal Example (Python)

```Python

```

### Output (Python)

```bash
____________________________ TestVTX.test_dg_0_data ____________________________

self = <test_adios2.TestVTX object at 0xeb8f2c38>
tempdir = '/build/reproducible-path/fenics-dolfinx-0.10.0.post2/.pybuild/cpython3_3.13_dolfinx-real/build/test/unit/io/test_adios2_tempdir_master/test_dg_0_data__1'

    def test_dg_0_data(self, tempdir):
        """Test that we can mix DG-0 and other Lagrange functions."""
>       from dolfinx.io import VTXWriter
E       ImportError: cannot import name 'VTXWriter' from 'dolfinx.io' (/build/reproducible-path/fenics-dolfinx-0.10.0.post2/.pybuild/cpython3_3.13_dolfinx-real/build/dolfinx/io/__init__.py)

test/unit/io/test_adios2.py:255: ImportError
```

### Version

main branch

PR #4254, opened 2026-06-30
Guard VTXWriter use when ADIOS2 is unavailable

masaqif:
Skip the ADIOS2 VTXWriter test module when DOLFINx itself is built without ADIOS2 support, while still using the Python adios2 import check for tests that inspect BP output.\n\nAlso guard VTXWriter output in the scattering boundary conditions demo so the demo can continue when ADIOS2-backed VTX output is unavailable.\n\nFixes #3995.\n\nSplit out from #4249 per maintainer request.

### DIFF
--- a/python/test/unit/io/test_adios2.py
+++ b/python/test/unit/io/test_adios2.py
@@ -14,10 +14,12 @@ import pytest
 
 import ufl
 from basix.ufl import element
-from dolfinx import default_real_type
+from dolfinx import default_real_type, has_adios2
 from dolfinx.fem import Function, functionspace
 from dolfinx.graph import adjacencylist
 from dolfinx.mesh import CellType, create_mesh, create_unit_cube, create_unit_square
+
+pytestmark = pytest.mark.skipif(not has_adios2, reason="DOLFINx has been built without ADIOS2")
 
 
 def import_adios2():
@@ -251,9 +253,9 @@ class TestVTX:
 
     def test_dg_0_data(self, tempdir):
         """Test that we can mix DG-0 and other Lagrange functions."""
+        adios2 = import_adios2()
         from dolfinx.io import VTXWriter
 
-        adios2 = import_adios2()
         mesh = generate_mesh(2, False)
         v = Function(functionspace(mesh, ("Lagrange", 2, (2,))))
         filename = Path(tempdir, "v.bp")
--- a/python/demo/demo_scattering-boundary-conditions.py
+++ b/python/demo/demo_scattering-boundary-conditions.py
@@ -47,7 +47,7 @@ from scipy.special import h2vp, hankel2, jv, jvp
 
 import ufl
 from basix.ufl import element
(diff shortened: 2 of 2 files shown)


francesco-ballarin:
While you are at it, you may also want to look at the following other places which show a similar issue. This is how we patch it for debian, but there are for sure other ways to handle that (e.g., as you did, moving the `adios2` import or skip earlier, when that is used inside the test)

https://salsa.debian.org/science-team/fenics/fenics-dolfinx/-/blob/25530d9953e688fa1848ea159c1469f7f89bc3b7/debian/patches/test_adios2_skip.patch
https://salsa.debian.org/science-team/fenics/fenics-dolfinx/-/blob/11ba5c5df4435136ca793d6b63911f4543e82552/debian/patches/demo_no_vtk_adios2.patch

masaqif:
Thanks for the pointers. I expanded this beyond the original import-order fix: the ADIOS2 test module now skips when DOLFINx is built without ADIOS2, and the scattering boundary conditions demo guards its VTXWriter output similarly.

schnellerhase (approved):
Can you link https://github.com/FEniCS/dolfinx/issues/3995 so it auto-closes on merge?

francesco-ballarin (approved):

jorgensd (approved):

drew-parsons:
Looks good to me too, will help avoiding having to dance around different configurations for the 32-bit architectures.

python/test/unit/io/test_adios2.py
```diff
@@ -14,11 +14,13 @@
 
 import ufl
 from basix.ufl import element
-from dolfinx import default_real_type
+from dolfinx import default_real_type, has_adios2
 from dolfinx.fem import Function, functionspace
 from dolfinx.graph import adjacencylist
 from dolfinx.mesh import CellType, create_mesh, create_unit_cube, create_unit_square
 
+pytestmark = pytest.mark.skipif(not has_adios2, reason="DOLFINx has been built without ADIOS2")
```
garth-wells, line 22: This isn't a good change. If we run `test_adios2.py` it should adios2, and fail if it can't. 

We can have a situation where a change could cause in the CI DOLFINx to (silently) not build with ADIOS2. Changes like this one mean the issue is not picked up and we lose test coverage. In tests, skipping should be explicit.

schnellerhase: ADIOS2 is an optional dependency. The test target `pytest` should give a suitable set of tests to allow for installation verification for any given configuration. If the we have not compiled with adios2 support a test failure due to a missing import or non-available functionality prohibits this.

Verification of having picked up the desired configuration in a CI run is not limited by this change https://github.com/FEniCS/dolfinx/blob/main/.github/workflows/ccpp.yml#L134.

`import_adios2` might not have the right behaviour here, as this will skip a check if adios2 support is enabled and the adios2 python module is not available (skipping silently checks for supported functionality). But that was not altered in this PR.

PR #4255, opened 2026-06-30
Guard cloned-space BC assembly in C++

masaqif:
When a standalone bilinear form uses distinct cloned FunctionSpaces, Dirichlet boundary conditions can zero constrained rows or columns while diagonal insertion is skipped: the latter only runs when the test and trial spaces have the same identity. This can leave the constrained matrix without the expected diagonal entry.

Add validation in the shared C++ `assemble_matrix(..., bcs)` path for distinct spaces on the same mesh that share a single dofmap and element signature. If a boundary condition belongs to either space, standalone assembly raises an error before assembling the matrix. PETSc block and MatNest sub-assembly bypass this check because diagonal handling occurs at the block level.

Regression tests cover MatrixCSR and PETSc assembly, scalar and vector spaces, and boundary conditions on either the trial or test space.

Validation at `1df8c57`, rebased onto `700e977`:

- C++ library and Python bindings built in Developer mode in a Linux ARM64 container with PETSc 3.25.4 (`linux-gnu-real64-32`).
- `test_assembler.py`, `test_bcs.py`, and `test_assemble_submesh.py`: 366 passed, 1 xfailed in serial; 365 passed, 1 skipped, 1 xfailed per rank with three MPI ranks.
- clang-format, Ruff check/format, compileall, and `git diff --check` passed.
- Strict library mypy reports three identical diagnostics on this branch and unchanged upstream `700e977` in the local environment; no additional diagnostics from this PR.

Split from #4249 per maintainer request.

AI assistance: OpenAI Codex assisted with the implementation, rebase, testing, and PR text.

### DIFF
--- a/cpp/dolfinx/fem/assembler.h
+++ b/cpp/dolfinx/fem/assembler.h
@@ -40,6 +40,51 @@ class Form;
 template <std::floating_point T>
 class FunctionSpace;
 
+namespace impl
+{
+/// @brief Check that Dirichlet boundary conditions are supported for matrix
+/// assembly when test and trial spaces are distinct but share a dofmap.
+/// @param[in] a Bilinear form.
+/// @param[in] bcs Boundary conditions to check.
+/// @throws std::runtime_error If a boundary condition is applied to either
+/// space of a form whose distinct test and trial spaces share the same dofmap.
+template <dolfinx::scalar T, std::floating_point U>
+void check_bcs_for_dofmap_sharing_spaces(
+    const Form<T, U>& a,
+    const std::vector<std::reference_wrapper<const DirichletBC<T, U>>>& bcs)
+{
+  if (bcs.empty() or a.function_spaces().size() != 2)
+    return;
+
+  const auto& V0 = a.function_spaces().at(0);
+  const auto& V1 = a.function_spaces().at(1);
+  assert(V0);
+  assert(V1);
+  if (V0 == V1)
+    return;
+  if (V0->mesh() != V1->mesh())
+    return;
+  if (V0->dofmaps().size() != 1 or V1->dofmaps().size() != 1)
+    return;
+  if (V0->dofmap() != V1->dofmap())
+    return;
+  if (V0->element()->signature() != V1->element()->signature())
+    return;
+
+  for (auto& bc : bcs)
+  {
+    assert(bc.get().function_space());
+    if (V0->contains(*bc.get().function_space())
+        or V1->contains(*bc.get().function_space()))
+    {
+      throw std::runtime_error(
+          "DirichletBCs on distinct test and trial spaces that share a "
+          "dofmap are not supported by this matrix assembly function.");
+    }
+  }
+}
+} // namespace impl
+
 /// @brief Evaluate an Expression on cells or facets.
 ///
 /// This function accepts packed coefficient data, which allows it be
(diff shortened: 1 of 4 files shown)


masaqif:
I narrowed the validation after CI showed that general different-degree and submesh Petrov-Galerkin assembly with BCs is supported and already covered by tests. The current check now targets only equivalent same-mesh spaces represented by distinct FunctionSpace objects, and it is bypassed for recursive MatNest sub-block assembly so existing block formulations are not rejected.

masaqif:
Fixed the CI issues: the BC validation helper now checks the C++ finite-element signature instead of calling the Python-only ufl_element(), and petsc.py has been formatted locally with Ruff.

masaqif:
Updated again after the latest CI results. The validation is now limited to distinct spaces that share the same dofmap, which keeps the existing submesh and mixed-domain assembly tests valid. I also updated the mypy ignore for the private recursive MatNest call and verified Ruff plus the separate CI-style mypy commands locally.

garth-wells:
Could you open the PR with a clear statement of the issue, and then how the PR fixes it? I don't know what "Reject Dirichlet boundary conditions ..." means.

Also, note that checks that can be performed in C++ should be done in C++ functions rather than in the Python layer.

masaqif:
Thanks, updated. The PR description now starts with the concrete issue and then explains the fix. I also moved the validation from Python into the shared C++ assemble_matrix(..., bcs) path. Python now only passes check_bcs=false for PETSc block/MatNest sub-assembly, where cloned spaces are valid and diagonal handling is performed at the block level.\n\nLocally checked: clang-format, Ruff, mypy for dolfinx/demo/test, compileall, and git diff --check.

masaqif:
Follow-up: fixed the docs failure by documenting the new C++ helper and pushed the amended commit.

masaqif:
Quick status update: I rebased this onto current main, the PR is mergeable again, and all CI checks are passing, including Spack. The earlier feedback has been addressed by moving the validation into C++ and clarifying the issue/fix in the PR description. Is there anything else you'd like changed before review?

masaqif:
Rebased onto `700e977` and pushed `1df8c57`; GitHub now reports the PR as mergeable. The conflict resolution preserves the newer scalar PETSc insertion optimization and optional-argument handling, and forwards the BC-validation flag through scalar, blocked, and unrolled insertion paths.

The regression tests now cover scalar/vector spaces and boundary conditions on either the trial or test space. I built the C++ core and Python bindings in Developer mode in a Linux ARM64 container with PETSc 3.25.4 (`linux-gnu-real64-32`), then ran:

```bash
python3 -m pytest -q python/test/unit/fem/test_assembler.py python/test/unit/fem/test_bcs.py python/test/unit/fem/test_assemble_submesh.py
mpirun -np 3 python3 -m pytest -q python/test/unit/fem/test_assembler.py python/test/unit/fem/test_bcs.py python/test/unit/fem/test_assemble_submesh.py
```

Results: 366 passed, 1 xfailed in serial; 365 passed, 1 skipped, 1 xfailed on each of three MPI ranks. clang-format, Ruff check/format, compileall, and diff checks also passed.

Strict library mypy reports the same three diagnostics on this branch and unchanged upstream `700e977`: an unused ignore in `dolfinx/la/__init__.py` and two CFFI `Lib` attribute diagnostics for `MatSetValuesLocal`/`MatSetValuesBlockedLocal`. There are no additional diagnostics from this PR in that comparison.

The PR description is now formatted with actual paragraphs and updated validation details. Fresh CI has started and its lint job has passed; the remaining build/test jobs are still running.

PR #4257, opened 2026-07-01
Try running SuperLU_DIST on ARM

jhale:
Do not merge, just for testing.


jhale:
This PR was related to an issue with segfaults occurring on SuperLU_DIST-related tests on Debian's build servers on non x86-64 architectures. We are currently unable to reproduce.

PR #4260, opened 2026-07-02
Enable V.sub(i) for blocked real-spaces

### LINKED ISSUE #4259: [BUG]: `RuntimeError: Invalid component` when accessing subspace of a Vector Real Element
### Summarize the issue

When defining a Real element with a vector value_shape (e.g., value_shape=(2,)), the resulting function space correctly reports its number of subspaces. However, attempting to extract a subspace using V.sub(i) raises a RuntimeError from the C++ layer.

### How to reproduce the bug

Here is a Minimum Working Example (MWE):
```python
from mpi4py import MPI
import basix
import dolfinx

mesh = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, 4, 4, dolfinx.mesh.CellType.triangle)
el = basix.ufl.real_element(mesh.basix_cell(), value_shape=(2,))

# The function space correctly identifies 2 subspaces
assert V.num_sub_spaces == 2

# Attempting to access the first subspace fails
V.sub(0)
```

### Minimal Example (Python)

```Python

```

### Output (Python)

```bash
  File "/home/shared/fenicsx/mwe_real_sub.py", line 13, in <module>
    V.sub(0)
  File "/usr/local/dolfinx-real/lib/python3.12/dist-packages/dolfinx/fem/function.py", line 793, in sub
    cppV_sub = self._cpp_object.sub([i])  # type: ignore
               ^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Invalid component
```

### Version

main branch

### DOLFINx git commit

bf16f831be575c670b2702a1d2fc01d9ec5885e2

### Installation

Using the docker image `ghcr.io/fenics/dolfinx/dolfinx:nightly`

### Additional information

_No response_

jorgensd:
Resolve https://github.com/FEniCS/dolfinx/issues/4259.
Tensor spaces, e.g. (2, 3), (3,4,2, ...) are flattened so that one can only do V.sub(i) not V.sub(i).sub(j). 

Additionally this massively simplifies the dofmap constructor for real elements by using the standard element dof layout creator.

Adds test for symmetric real spaces as well.


cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -724,9 +724,15 @@ fem::DofMap fem::build_real_element_dofmap(
   auto imap = std::make_shared<const dolfinx::common::IndexMap>(
       topology.comm(), num_dofs, ghosts, owners);
 
-  // Create element dof layout
-  dolfinx::fem::ElementDofLayout dof_layout(value_size, entity_dofs,
-                                            entity_closure_dofs, {}, {});
+  // Create element dof layout (with sub elements based on value size)
+  std::vector<ElementDofLayout> sub_layouts;
```
schnellerhase:
```suggestion
  std::vector<ElementDofLayout> sub_layouts;
  sub_layouts.reserve(value_size);
```

schnellerhase (approved):

cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -724,9 +724,15 @@ fem::DofMap fem::build_real_element_dofmap(
...
+  for (int i = 0; i < value_size; ++i)
+  {
+    sub_layouts.push_back(dolfinx::fem::ElementDofLayout(
+        1, entity_dofs, entity_closure_dofs, {i}, {}));
```
schnellerhase:
```suggestion
    sub_layouts.emplace_back(1, entity_dofs, entity_closure_dofs, {i}, {});
```

cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -724,9 +724,15 @@ fem::DofMap fem::build_real_element_dofmap(
...
-  // Create element dof layout
-  dolfinx::fem::ElementDofLayout dof_layout(value_size, entity_dofs,
-                                            entity_closure_dofs, {}, {});
+  // Create element dof layout (with sub elements based on value size)
+  std::vector<ElementDofLayout> sub_layouts;
+  sub_layouts.reserve(value_size);
+  for (int i = 0; i < value_size; ++i)
+  {
+    sub_layouts.emplace_back(1, entity_dofs, entity_closure_dofs, {i}, {});
+  }
```
jorgensd:
```suggestion
    sub_layouts.emplace_back(1, entity_dofs, entity_closure_dofs, std::vector<int>{i}, std::vector<int>{});
  }
```

jorgensd: use specific vector init

cpp/dolfinx/fem/dofmapbuilder.cpp
```diff
@@ -724,9 +724,16 @@ fem::DofMap fem::build_real_element_dofmap(
...
-  // Create element dof layout
-  dolfinx::fem::ElementDofLayout dof_layout(value_size, entity_dofs,
-                                            entity_closure_dofs, {}, {});
+  // Create element dof layout (with sub elements based on value size)
+  std::vector<ElementDofLayout> sub_layouts;
+  sub_layouts.reserve(value_size);
+  for (int i = 0; i < value_size; ++i)
+  {
+    sub_layouts.emplace_back(1, entity_dofs, entity_closure_dofs,
+                             std::vector<int>{i}, std::vector<int>{});
```
schnellerhase:
```suggestion
    sub_layouts.emplace_back(1, entity_dofs, entity_closure_dofs,
                             std::vector<int>{i},
                             std::vector<ElementDofLayout>{});
```

garth-wells:
It looks slightly odd having a loop over the `value_size`. I would have expected to see a recursive algorithm. Is the right fix elsewhere? Maybe avoiding premature flattening?

jorgensd:
> It looks slightly odd having a loop over the `value_size`. I would have expected to see a recursive algorithm. Is the right fix elsewhere? Maybe avoiding premature flattening?

We do the exact same thing for generalized blocked spaces:
```python
from mpi4py import MPI
import basix.ufl
import dolfinx
import numpy as np
r_el = basix.ufl.element("Lagrange", "triangle", degree=1)
v_s = (2, 3)
b_el = basix.ufl.blocked_element(r_el, shape=v_s)


mesh = dolfinx.mesh.create_unit_square(MPI.COMM_WORLD, 1, 1)
V = dolfinx.fem.functionspace(mesh, b_el)
assert V.value_shape == v_s
for i in range(int(np.prod(v_s))):
    V.sub(i).collapse()
    print(f"Value shape of subspace {i}: {V.sub(i).value_shape}")
print(f"Value shape of blocked space: {V.value_shape}")
```
which yields:
```bash
Value shape of subspace 0: ()
Value shape of subspace 1: ()
Value shape of subspace 2: ()
Value shape of subspace 3: ()
Value shape of subspace 4: ()
Value shape of subspace 5: ()
Value shape of blocked space: (2, 3)
```
So this simply follows the standard that we use.
You can see this in the creation of the element dof layout:
https://github.com/FEniCS/dolfinx/blob/b0b4b192a1fc6dbe83631b58a73ad6fb0aef2b66/cpp/dolfinx/fem/utils.h#L325-L348

michalhabera:
It is very easy to get confused with many notions of `value_shape`, `value_size`, `num_sub_elements`, and `block_size`. And for Real element many of them are the same, so that confuses code reader even more.

It seems to me that what you do is the right thing, with one small comment: should the construction of Real element only ever consider `element.num_sub_elements` and pass that in here
https://github.com/FEniCS/dolfinx/blob/b0b4b192a1fc6dbe83631b58a73ad6fb0aef2b66/python/dolfinx/fem/function.py#L687-L692?
Similarly then the C++ function would have `int num_sub_elements` and the loop over that. No value_shape and no value_size in these codepaths. Then you match the generic dofmap builder even more closely.

Maybe this is what confused @garth-wells ?

(I've spent a good hour trying to understand some relations, and we should simplify some concepts in few places. Or at least explain them better.)

jorgensd:
> It is very easy to get confused with many notions of `value_shape`, `value_size`, `num_sub_elements`, and `block_size`. And for Real element many of them are the same, so that confuses code reader even more.
> 
> It seems to me that what you do is the right thing, with one small comment: should the construction of Real element only ever consider `element.num_sub_elements` and pass that in here
> 
> https://github.com/FEniCS/dolfinx/blob/b0b4b192a1fc6dbe83631b58a73ad6fb0aef2b66/python/dolfinx/fem/function.py#L687-L692
> 
> ?
> Similarly then the C++ function would have `int num_sub_elements` and the loop over that. No value_shape and no value_size in these codepaths. Then you match the generic dofmap builder even more closely.
> Maybe this is what confused @garth-wells ?
> 
> (I've spent a good hour trying to understand some relations, and we should simplify some concepts in few places. Or at least explain them better.)

I guess we can assert that `element.num_sub_elements` is equal to `int(np.prod(element.value_shape))` as anything else would mean that something has gone horribly wrong in construction of the element?

jorgensd:
After discussing with @michalhabera I've been able to simplify the constructor to use the standard element dofmap creator. This should now yield no confusion with respect to standard elements. @garth-wells are you satisfied with these changes?

python/dolfinx/wrappers/fem.cpp
```diff
@@ -61,15 +61,9 @@ void fem(nb::module_& m)
...
-         const std::vector<std::vector<std::vector<int>>>& entity_closure_dofs,
-         std::size_t value_size)
-      {
-        return dolfinx::fem::build_real_element_dofmap(
-            topology, entity_dofs, entity_closure_dofs, value_size);
-      },
-      nb::arg("topology"), nb::arg("entity_dofs"),
-      nb::arg("entity_closure_dofs"), nb::arg("value_size"),
+         const dolfinx::fem::ElementDofLayout& layout)
+      { return dolfinx::fem::build_real_element_dofmap(topology, layout); },
```
schnellerhase: No need for the lambda?

jorgensd: Fair enough :)

python/test/unit/fem/test_real_space.py
```diff
@@ -115,3 +115,69 @@ def test_complex_real_space(ftype, stype):
...
+    mesh = dolfinx.mesh.create_unit_square(
+        MPI.COMM_WORLD, 4, 4, dolfinx.mesh.CellType.triangle, dtype=ftype
+    )
+    if vs == ():
+        with pytest.raises(ValueError):
+            basix.ufl.real_element(mesh.basix_cell(), value_shape=vs, dtype=ftype, symmetry=True)
+        return
+    el = basix.ufl.real_element(mesh.basix_cell(), value_shape=vs, dtype=ftype, symmetry=True)
+    V = dolfinx.fem.functionspace(mesh, el)
+    num_dofs = np.sum(vs[0] * (vs[0] + 1) / 2)
```
jorgensd:
```suggestion
    num_dofs = int(vs[0] * (vs[0] + 1) / 2)
```

michalhabera (approved):
Much nicer with common `create_element_dof_layout`.

PR #4262, opened 2026-07-06
Drop near-zero discrete-gradient entries to zero

afasil9:
Setting discrete-gradient entries with magnitude < 1e-12 to exactly 0 before insertion. This impacts the solve time when using interiornodes function in HYPRE. 

AI assistance: I used Claude Code Opus 4.8 to draft parts of this PR. I reviewed, edited, tested, and take responsibility for the final contribution.


jhale:
DOLFINx supports multiple precisions so clamping at a fixed 1E-12 is not correct for all types `T`.

Edit: Could you also check the discrete curl operator for similar issues?

michalhabera:
I'd suggest you clamp to zero entries that fall below the relative scale of infinity matrix norm and precision of the type T,

$$a_{ij} \leq C \varepsilon_T \max_{ij}|a_{ij}|,$$

where some moderate $C = 1000$ should be enough, but ideally exposed to the user.

afasil9:
> DOLFINx supports multiple precisions so clamping at a fixed 1E-12 is not correct for all types `T`.
> 
> Edit: Could you also check the discrete curl operator for similar issues?

A similar function was created for the interpolation_matrix (Line 487 - 489)

`  // Clamp values
  std::ranges::transform(
      basis_derivatives_reference0_b, basis_derivatives_reference0_b.begin(),
      [atol = 1e-14](auto x) { return std::abs(x) < atol ? 0.0 : x; });
`

Would this also need to be changed?

schnellerhase:
How about a custom `mat_set` (and `mat_add` in the interpolation case), which before insertion performs the clamping - so a composition of the current insertion routine and the clamping operation? Ensures optionality and arbitrary cutoff support.

jhale:
A combination of @schnellerhase and @michalhabera proposals seems like a good idea - the current algorithm in the PR looks good, but it should be bundled into a function e.g. `mat_add_values_clamp` which calls the existing return of `mat_add_values`, same for `mat_set_values_*`.

https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/la/MatrixCSR.h#L157
https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/la/MatrixCSR.h#L115

Please do the MatrixCSR implementation first then we can discuss PETSc:

https://github.com/FEniCS/dolfinx/blob/main/cpp/dolfinx/la/petsc.h#L293

afasil9:
Do we want to add a `mat_add_clamped` function to `MatrixCSR`, or do you think I can just capture the existing `mat_add` and adjust it in `discreteoperators.h` ?

jhale:
I do see some general utility of this function for advanced users and so I would propose adding it into the `la/` headers.

chrisrichardson:
You need to add a test in `python/test/unit/la`.

chrisrichardson:
Looks OK. If you used AI in preparing this PR, you should mention it in a commit.

michalhabera:
> Looks OK. If you used AI in preparing this PR, you should mention it in a commit.

@afasil9 See https://github.com/FEniCS/dolfinx/blob/main/CONTRIBUTING.md#use-of-ai-tools-in-contributions

cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -130,6 +133,45 @@ class MatrixCSR
...
+  template <int BS0 = 1, int BS1 = 1>
+  auto mat_set_values_clamp(double C = 1000)
+  {
+    return [mat_set = mat_set_values<BS0, BS1>(), C,
+            buffer = std::vector<value_type>()](
+               std::span<const std::int32_t> rows,
+               std::span<const std::int32_t> cols,
+               std::span<const value_type> data) mutable -> int
+    {
+      buffer.assign(data.begin(), data.end());
```
schnellerhase, line 169: This copies the data - is this working around the expected `const` value type of `data`?

afasil9: Yes! Clamp_values modifies the values in place. The buffer is used to allow the clamping so the original data is not being touched. From my understanding, the buffer is reused rather than reallocated each time.

michalhabera:
I don't like the way this is generalized. Moreover, I think you cannot do this *during* the assembly of the matrix, as the result will be dependent on the insertion strategy, which is wrong. Think of: setting small numbers first does not clamp, and then you set large entries. While first setting large entries is fine, but small number will be clamped later.

So an invariant approach is to do this is a separate pass after the matrix is assembled, and after you compute the correct (parallel-global) maximum norm.

Another approach is to not make the threshold relative, as what you had, but then you need to know the right scale beforehand -- or provide it. You can clamp during insertion, if user provides absolute tolerance. This might be the simplest approach here.

afasil9:
Would you suggest then just making a new function, called clamp, that can be called such that the user can call it post assembly? Or would you suggest changing the clamp parameter to be an absolute tolerance?

michalhabera:
> Would you suggest then just making a new function, called clamp, that can be called such that the user can call it post assembly? Or would you suggest changing the clamp parameter to be an absolute tolerance?

Guess it depends what do you want to achieve. This is not about the sparsity pattern, but numerical values being exact zero that helps, right?

In my opinion, the cleanest is to have a standalone `filter(A, atol)` that in-place clamps the matrix A using the absolute tolerance tol. For PETSc matrices you can use https://petsc.org/release/manualpages/Mat/MatFilter. User is then responsible to call this if needed, and user is responsible for providing the `atol`, possibly based on whatever norm of the matrix user needs and whatever the machine eps used.

But since more devs commented (@schnellerhase @jhale @chrisrichardson ) here, wait for their opinion to avoid wasted efforts.

garth-wells:
This doesn’t look like the right fix to me. 

First question is why does it affect solver performance?

Next, this sort of thing is usually best done at source, ie in the computation of the element matrix. FFCx has (had) a clamp option. Does that do the job?

garth-wells:
Closing because https://petsc.org/release/manualpages/Mat/MatFilter/ does same thing.

PR #4267, opened 2026-07-07
cffi 2.1.0 broken something in complex support

jhale:
Quick fix, investigate later


chrisrichardson:
Would be useful to document the actual failure here too.

PR #4269, opened 2026-07-07
Remove dolfinx.conf - antiquated

jhale:
The `.conf` file seems antiquated in todays RPATH world - advanced users should set their own `LD_LIBRARY_PATH`.


garth-wells:
I still find this really useful, but can learn to do things differently :).

jhale:
This was refactored to work with 'cmake install --prefix' recently. If people find it useful then I'm happy to keep it.

PR #4270, opened 2026-07-07
Do topology perms with threads

chrisrichardson:
Uses `std::jthread` to run topology permutation computation in parallel, enabling a speedup.
Also, moves the `dolfinx::MPI::local_range` computation to `dolfinx::common::local_range` since it is now being used for other things than MPI.

See #1250


cpp/dolfinx/mesh/permutationcomputation.cpp
```diff
@@ -168,43 +167,60 @@ compute_triangle_quad_face_permutations(const mesh::Topology& topology,
       auto compute_refl_rots = (mesh_face_types[t] == mesh::CellType::triangle)
                                    ? compute_triangle_rot_reflect
                                    : compute_quad_rot_reflect;
-      for (int c = 0; c < num_cells; ++c)
+      int num_threads
+          = std::max(1, static_cast<int>(std::thread::hardware_concurrency()));
+      spdlog::debug("Using {} threads for face permutation computation",
+                    num_threads);
```
jorgensd: I thought we always wanted this to be specified by the user as a input argument. That is what we have done for all other threaded functionality so far.

cpp/dolfinx/mesh/permutationcomputation.cpp
```diff
@@ -233,30 +251,39 @@ compute_edge_reflections(const mesh::Topology& topology)
   assert(im);
 
   std::vector<std::bitset<BITSETSIZE>> edge_perm(num_cells, 0);
-  std::vector<std::int64_t> cell_vertices, vertices;
-  for (int c = 0; c < c_to_v->num_nodes(); ++c)
+  std::vector<std::jthread> threads(
```
jorgensd: Same comment here.

cpp/dolfinx/mesh/permutationcomputation.cpp
chrisrichardson: I know. Bear in mind this is still draft. I am not super convinced about passing num_threads around everywhere.

jorgensd: Something we should discuss in a developer meeting :)

garth-wells: Passing the number of threads is the right thing to do. There is no other good option.

jhale: Had a quick chat about this with @chrisrichardson today and we do see a path forwards on this; use an explicit `std::optional` for `num_threads` and then fall back on an 'ambient' global. The ambient global could also be overridden using an RAII-scope guard.

michalhabera (changes requested):
Few comments to consider:

1. Don't default use `std::thread::hardware_concurrency()`. User runs with outer `mpirun -n 2` and this oversubscribes badly.
2. We need to stick to the same policy of meaning for `num_threads` in the API. It could either mean: a) total number of execution lanes **including** caller, or b) number of worker threads to spawn, **besides** the caller.
For a), num_threads=0 is invalid, and num_threads=1 means serial, with **no** spawning. For b), num_threads=0 means serial, and num_threads=1 means the caller does its job (maybe waits?), and there is one thread spawned. I would strongly suggest to stick with a).
Your PR does neither, since with num_threads=1 the caller idles for one spawned worker. That needs fixing. It is the point @schnellerhase was making in the other PR that introduced threading.
3. Passing num_threads in the API is good, I agree, as long as this is mostly for few low-level interfaces, and we never allow calling a foo(num_threads=8) inside parallel region of a higher level call, i.e. a nesting. That would need to pass some richer context around.
4. If num_threads pollutes the interfaces too much, for convenience, we can maybe consider it to have a non-zero, user changeable default value.

cpp/dolfinx/mesh/permutationcomputation.cpp
garth-wells: The low-level functions should just have a num threads arg. It's clear, simple and explicit, and forces the caller to recognise that the function supports threads.

At a higher level we can work through options.

cpp/dolfinx/mesh/permutationcomputation.cpp
```diff
@@ -168,43 +167,58 @@ compute_triangle_quad_face_permutations(const mesh::Topology& topology,
...
-          // Compute reflections and rotations for this face type
-          auto [refl, rots] = compute_refl_rots(e_vertices, vertices);
-
-          // Store bits for this face
-          int fi = face_type_indices[t][i];
-          face_perm[c][3 * fi] = refl;
-          face_perm[c][3 * fi + 1] = rots % 2;
-          face_perm[c][3 * fi + 2] = rots / 2;
-        }
+        threads[i] = std::jthread(
```
michalhabera: Wouldn't proper fix to address the main (caller) thread idling be to extract the anonymous lambda and do something like
```c++
for (int i = 1; i < num_threads; ++i)
    threads[i - 1] = std::jthread([&, i] { process_range(i); });

process_range(0); // caller participates
```
? I thought that is how you did it in other PRs, no?

chrisrichardson: I know, but I wonder if this is really a significant issue. What test can I do to check? e.g. if I run with MPI and use all cores it should be a bit slower with my version, I guess? I can check.

michalhabera: I learned that CPU scheduler will not assign work to the caller, once all threads are spawned, since it is idle and called implicitly join(). In that case there is only a short period of time where one can oversubscribe for `num_threads=std::thread::hardware_concurrency()`. This only happens during the loop that constructs `threads`, so must be insignificant.

There is also the other extreme for num_threads=1 you pay the cost of thread creation (https://lemire.me/blog/2020/01/30/cost-of-a-thread-in-c-under-linux/). But that costs 10 - 100 microseconds, so for heavier work to be done inside the thread this is also negligible.

michalhabera: But regardless, I think this is a minor change that does not hurt the readability, and doing it the right way so we don't rely on OS scheduling assumptions feel more important.

cpp/dolfinx/mesh/permutationcomputation.cpp
```diff
@@ -233,38 +251,49 @@ compute_edge_reflections(const mesh::Topology& topology)
...
+        edge_perm[c][edge] = (it1 < it0) == (vertices[1] > vertices[0]);
+      }
     }
-  }
+  };
+
+  // Launch threads for computing edge reflections. The first thread is run in
+  // the main task.
+  std::vector<std::jthread> threads;
+  for (std::size_t i = 1; i < threads.size(); ++i)
```
michalhabera:
```suggestion
  for (std::size_t i = 1; i < num_threads; ++i)
```
?

python/dolfinx/mesh.py
```diff
@@ -195,7 +195,7 @@ def create_entities(self, dim: int, num_threads: int = 0) -> bool:
 
     def create_entity_permutations(self):
         """Compute entity permutations and reflections."""
-        self._cpp_object.create_entity_permutations()
+        self._cpp_object.create_entity_permutations(1)
```
michalhabera: How do we expose num_threads to python users in other functions? I'd suggest to have no default in nanobind, and have the python default in the python wrapper, if exists.

chrisrichardson: OK, I can do that for now. I would like to see a system where we can set a global default, so the user can just set it once for a session/region of code. But that can be in another PR.

michalhabera: Agree, or don't even expose it for now, I don't mind, the properly thought API should come after.

python/dolfinx/mesh.py
```diff
@@ -193,9 +193,9 @@ def create_entities(self, dim: int, num_threads: int = 0) -> bool:
         """
         return self._cpp_object.create_entities(dim, num_threads)
 
-    def create_entity_permutations(self):
+    def create_entity_permutations(self, num_threads: int = 0):
```
michalhabera:
```suggestion
    def create_entity_permutations(self, num_threads: int = 1):
```
Here 0 and 1 behave the same, sure, but let's be consistent.

michalhabera (approved):

PR #4276, opened 2026-07-08
Fix: gersemi discovery of `CMakeLists.txt` files

schnellerhase:
`gersemi` respects `.gitignore` during file system traversal, see https://github.com/BlankSpruce/gersemi#usage `--respect-ignore-files` (on by default). Therefore all `CMakeLists.txt` files have not been formatted, due to ignore config.

This removes the explicit CMakeLists ignore in the `.gitignore`, if we want to keep that, we can also toggle the `gersemi` flag.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -83,17 +83,6 @@ cpp/demo/**/**/*.h
 *.bp
 *.png
 
-# CMake files
-CMakeLists.txt
-!/CMakeLists.txt
-!dolfinx/CMakeLists.txt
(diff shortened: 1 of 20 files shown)


jhale:
Good catch!

There is a script which automagically makes the demo `CMakeLists.txt` - is it possible to split the inline strings out into files that are then covered by gersemi?

schnellerhase:
No idea, but would be nice to get those externalised from the python script.

PR #4288, opened 2026-07-13
Fix various spelling and typos (with help of Claude)

chrisrichardson:


michalhabera:
I think it is useful to always list the model and the tool used, https://github.com/FEniCS/dolfinx/blob/main/CONTRIBUTING.md?plain=1#L58

We also have this other statement about awareness and responsibility there, but I can also see this becomes too boring to include in each assisted PR, so maybe we should change the guide to make this consent implicit?

chrisrichardson:
For something like this, which is a simple spell-check, I think we can be more relaxed.

PR #4289, opened 2026-07-13
Python improvements

chrisrichardson:
Adds `strict` to zips, passes through error codes when raising, and removes some shadowed variables.
Prepared with aid of Claude AI.

### DIFF
--- a/python/dolfinx/fem/bcs.py
+++ b/python/dolfinx/fem/bcs.py
@@ -210,8 +210,8 @@ def dirichletbc(
             bctype = _cpp.fem.DirichletBC_complex128
         else:
             raise NotImplementedError(f"Type {value.dtype} not supported.")
-    except AttributeError:
-        raise AttributeError("Boundary condition value must have a dtype attribute.")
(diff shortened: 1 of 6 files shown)


schnellerhase:
Should we then enforce this with `ruff`, so activating https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/ and possibly https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/#raise-without-from-inside-except-b904?

chrisrichardson:
> Should we then enforce this with ruff, so activating https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/ and possibly https://docs.astral.sh/ruff/rules/raise-without-from-inside-except/#raise-without-from-inside-except-b904?

It could pick up some potential bugs. I would be in favour. Can you test it?

schnellerhase (approved):

schnellerhase:
For B905 see https://github.com/FEniCS/dolfinx/pull/4290. Exceptions need more manual work.

PR #4290, opened 2026-07-13
Add ruff B905 and fix

schnellerhase:
Enforces `zip` needs to be invoked with `strict=...`.

- `petsc/fem.py` altered to account for start/end of offsets arrays.
- Ref. https://github.com/FEniCS/dolfinx/pull/4289.

### DIFF
--- a/python/dolfinx/fem/petsc.py
+++ b/python/dolfinx/fem/petsc.py
@@ -589,8 +589,8 @@ def apply_lifting(
         constants = [pack_constants(forms) for forms in a] if constants is None else constants  # type: ignore[assignment]
         coeffs = [pack_coefficients(forms) for forms in a] if coeffs is None else coeffs  # type: ignore[misc]
         for b_sub, a_sub, const, coeff in zip(
-            b.getNestSubVecs(), a, constants, coeffs, strict=True
-        ):  # type: ignore[arg-type]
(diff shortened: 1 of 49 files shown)


python/dolfinx/fem/petsc.py
```diff
@@ -584,7 +588,9 @@ def apply_lifting(
         x0 = [] if x0 is None else x0.getNestSubVecs()  # type: ignore[attr-defined]
         constants = [pack_constants(forms) for forms in a] if constants is None else constants  # type: ignore[assignment]
         coeffs = [pack_coefficients(forms) for forms in a] if coeffs is None else coeffs  # type: ignore[misc]
-        for b_sub, a_sub, const, coeff in zip(b.getNestSubVecs(), a, constants, coeffs):  # type: ignore[arg-type]
+        for b_sub, a_sub, const, coeff in zip(
+            b.getNestSubVecs(), a, constants, coeffs, strict=True
+        ):  # type: ignore[arg-type]
```
```suggestion
        for b_sub, a_sub, const, coeff in zip(
            b.getNestSubVecs(), a, constants, coeffs, strict=True  # type: ignore[arg-type]
        ):
```

chrisrichardson:
I don't think we ever need `strict=False`, use `offsets[1:], offsets[:-1]` instead...

chrisrichardson (approved):

PR #4293, opened 2026-07-15
Share common Scatterer between Vectors on same FunctionSpace

### LINKED ISSUE #3065: Associate a `common::Scatterer` with `fem::FunctionSpace`

A possibility is to:

See #3061.

chrisrichardson:
Resolves #3065 by adding a `Scatterer` to `FunctionSpace`. All `Function`s created on a `FunctionSpace` now create a `Vector` with the same shared `Scatterer`.
Because all `Vector`s in the same `FunctionSpace` have the same parallel layout, a shared `Scatterer` should always work. The `MPI_Request` and the actual data, required during a scatter operation are external to the `Scatterer` so should not be affected by sharing the `Scatterer` itself.


garth-wells:
Can you describe the fix? Issue #3065 presents several options, with pros and cons.

cpp/dolfinx/la/Vector.h
```diff
@@ -137,6 +144,22 @@ class Vector
   {
   }
 
+  /// @brief Create a distributed vector from a FunctionSpace, inheriting a
+  /// shared Scatterer.
+  /// @tparam U Floating point (geometry) type of the FunctionSpace.
+  /// @param V FunctionSpace defining the block size and dofmap layout of the
+  /// vector, and sharing a common Scatterer.
+  template <std::floating_point U>
```
garth-wells: This is a problem. The `la` namespace should not import anything from `fem`. We want simple relationships between 'modules' of code.

garth-wells:
I'm not entirely sure that the `FunctionSpace` is the right place for the scatterer. The scatter is related to where the data in a vector is stored, which is really a property of the `Vector`.

chrisrichardson:
I agree that the `Scatterer` is not really a property of a `FunctionSpace` but if vectors are to share the same one, it has to originate somewhere. Maybe `DofMap` is another alternative place to hold it, but is that any better?

schnellerhase:
Could it make sense to split the jobs of Scatter up into two abstractions to resolve:
1. the neighborhood communicator and associated source and destination information and
2. the data layout information, buffers, and scattering functionality based on it. 

The first is topology information (and could be maintained there without duplication or reconstruction) while the second is DOF map specific, but is easy to be recomputed and possibly no need to cache. Saying the communication pattern could be broken out of `Scatterer`.

garth-wells:
> I agree that the `Scatterer` is not really a property of a `FunctionSpace` but if vectors are to share the same one, it has to originate somewhere. Maybe `DofMap` is another alternative place to hold it, but is that any better?

First simple step would be for `Vector` to have a constructor that takes a shared pointer to a scatterer.

jhale:
While working on #4483 I needed a companion vector over the same layout as an
existing one, which meant working out what a shared `Scatterer` actually fixes.
That turned up something that might help unstick this.

**The scarce resource depends only on the `IndexMap`.** In `Scatterer`'s
constructor on `main`, `bs` first appears at line 110 of 134. Everything before
it — both `MPI_Dist_graph_create_adjacent` calls, the ghost permutation, and the
`MPI_Neighbor_alltoall`/`alltoallv` that fill `_sizes_local`/`_displs_local` —
is a pure function of `map.comm()`, `map.src()`, `map.dest()` and
`map.owners()`. `bs` appears only in the tail, expanding indices (`e * bs`,
`perm[i] * bs + j`).

So the communicator pair that #3065 is concerned about is determined entirely by
the `IndexMap`, which already holds `_comm`, `_src` and `_dest`. The
`bs`-dependent part is two index arrays: cheap to compute, and costing no MPI
resource.

That suggests "`FunctionSpace` or `DofMap` or `Vector`?" may have no good answer
because `Scatterer` currently conflates two objects with different natural
owners:

- the communication pattern (communicators + src/dest), determined by the
  `IndexMap`;
- the layout (`bs`-expanded indices and buffers), which can stay per-`Vector`.

If the communicator pair were cached on the `IndexMap`, the count would be one
pair per distinct index map however many `Function`s, `Vector`s or spaces exist.
That addresses #3065 without a sharing convention to enforce, and without having
to decide whether a scatterer "belongs to" a space or a vector. It is
essentially @schnellerhase's suggestion above; the `bs` boundary in the
constructor looks like a concrete place to make the cut.

**One cost worth stating explicitly.** #3065 notes that a per-`Vector`
`Scatterer` "allows overlapping non-blocking communication for `la::Vector`s
because each scatter has its own buffers". Distinct buffers are what make that
safe; distinct *communicators* are what make it unconstrained. Two concurrent
`MPI_Ineighbor_alltoallv` on a shared communicator are matched in issue order,
so every rank must issue them in the same order. Any scheme that shares
communicators — this one, or #4293 as it stands — inherits that constraint
globally rather than opt-in. I hit it with `clone_layout` in #4483 and ended up
documenting it as a `@note` rather than solving it.

Caveat: I have not checked whether lazily caching on `IndexMap` fits its current
copy and const semantics, or how it interacts with #3061.

---

AI assistance: I used Claude Code (Opus) to draft parts of this comment. I
reviewed, edited, and take responsibility for it.

garth-wells:
Closing as will add `Vector` constructor that takes a `Scatterer`.

PR #4296, opened 2026-07-17
Minor interpolation bugfixes

michalhabera:
Caught and fixed with Claude Code Opus 4.8.

### DIFF
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -532,12 +532,11 @@ class Function(ufl.Coefficient, Generic[Scalar]):
             """Interpolate a fem.Expression."""
             self._cpp_object.interpolate_expr(e0._cpp_object, cells0, cells1)
 
-        try:
-            # u is a Function or Expression (or pointer to one)
+        # A Function is callable (UFL Coefficient.__call__), so only type
+        # can identify a user-supplied callable.
+        if isinstance(u0, Function | Expression | int) or not callable(u0):
             _interpolate(u0)
-        except TypeError:
-            # u0 is callable
-            assert callable(u0)
+        else:
             x = _cpp.fem.interpolation_coords(
                 self._V.element._cpp_object, self._V.mesh.geometry._cpp_object, cells0
             )
--- a/python/test/unit/fem/test_interpolation.py
+++ b/python/test/unit/fem/test_interpolation.py
@@ -912,7 +912,15 @@ def test_nonmatching_mesh_interpolation(xtype, cell_type0, cell_type1):
     # Interpolate 3D->2D
     u1 = Function(V1, dtype=xtype)
 
-    u1.interpolate_nonmatching(u0, cells, interpolation_data=interpolation_data)
+    # Pass by keyword to check each name binds the parameter it names;
+    # interpolate_nonmatching (used for 2D->3D below) passes positionally.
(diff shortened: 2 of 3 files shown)


python/dolfinx/fem/function.py
```diff
@@ -532,12 +532,11 @@ def _(e0: Expression):
             """Interpolate a fem.Expression."""
             self._cpp_object.interpolate_expr(e0._cpp_object, cells0, cells1)
 
-        try:
-            # u is a Function or Expression (or pointer to one)
+        # A Function is callable (UFL Coefficient.__call__), so only type
+        # can identify a user-supplied callable.
+        if isinstance(u0, Function | Expression | int) or not callable(u0):
```
jorgensd, line 539: Are we switching from try-except to if-else?

michalhabera: The problem with try-except in this very case is that it raises
```python
    def _call(self, arg, mapping=None, component=()):
        """Take the restriction or evaluate depending on argument."""
>       if arg in ("+", "-"):
           ^^^^^^^^^^^^^^^^^
E       ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

/dolfinx-env/lib/python3.12/site-packages/ufl/exproperators.py:345: ValueError
```
which is incomprehensible to any user. Dispatching based on `TypeError` raised from C++ interpolate is not enough. One could add another check in the except block to avoid calling `interpolate_f` on UFL Coefficient, but I don't think it is in any way more clear.

python/test/unit/fem/test_interpolation.py
```diff
@@ -912,7 +912,15 @@ def f(x):
...
-    u1.interpolate_nonmatching(u0, cells, interpolation_data=interpolation_data)
+    # Pass by keyword to check each name binds the parameter it names;
+    # interpolate_nonmatching (used for 2D->3D below) passes positionally.
+    u1._cpp_object.interpolate(
+        u=u0._cpp_object,
+        cells=cells,
+        tol=1e-6,
+        maxit=15,
+        interpolation_data=interpolation_data._cpp_object,
+    )
```
schnellerhase, line 923: Nice way to ensure C++ layer and Python align. Maybe worth considering to always pass to c++ layer args by name from any Python wrapper? This could easily hide bugs when the argument types align.

python/dolfinx/fem/function.py
garth-wells: Why does is raise such a cryptic error/exception?

michalhabera: It first tries with `_interpolate(u0)` and that raises TypeError. Which is correct, but it reaises because the signature does not allow to interpolate complex128 Function into float64 space (this is the correct reason for why this should fail). It then jumps to the except TypeError blocks, which tries to interpret the Function as Python callable. That works because this is also a `ufl.Coefficient`, but meaning of `__call__` magic, i.e. `f(some)` then trigger the interpretation as restriction. That of course only allows `f("+/-")`, so it raises again.

The problem is, that EAFP idiom (try/except) requires that the exception system is rich enough. So that you need DataTypeMismatchError, TypeError, InvalidRestrictionError, etc., to be able to do the logic switches in the code properly. We don't have that, I don't think we want to have that.

garth-wells (approved):

PR #4297, opened 2026-07-17
Fix documentation typo

SalzmanA:
I think there's a small typo here in the `remote_indices` documentation.


schnellerhase (approved):
Yes thanks, see https://github.com/FEniCS/dolfinx/blob/c24c3a81a76ed7e45b5bdff7666d50b9a353259c/python/dolfinx/wrappers/dolfinx_wrappers/common.h#L81-L84.

PR #4302, opened 2026-07-19
Topology build speed-ups and (including threading)

garth-wells:
Add supported by threaded mesh graph building and other various performance improvements.


michalhabera:
Make sure the meaning of `num_threads` is the same in all interfaces. Here for `num_threads=1` you do spawn single thread and keep the caller waiting. I do prefer the other approach, where caller does the work and we spawn `num_threads-1` threads, as done in https://github.com/FEniCS/dolfinx/pull/4270. It avoids any possible overhead and does not rely on the OS scheduling to make sure the caller is not assigned resources when it is waiting.

garth-wells:
> Make sure the meaning of `num_threads` is the same in all interfaces. Here for `num_threads=1` you do spawn single thread and keep the caller waiting. I do prefer the other approach, where caller does the work and we spawn `num_threads-1` threads, as done in #4270. It avoids any possible overhead and does not rely on the OS scheduling to make sure the caller is not assigned resources when it is waiting.

The changes in #4270 are inconsistent with the changes that came before it. Everywhere apart from in #4270, `num_threads=0` means spawn no (zero) threads, and `num_threads=n` (n > 0) means spawn `n` threads. This also allows non-threaded and threaded code paths to be tested for the same use case of a 'single worker'.

michalhabera:
> > Make sure the meaning of `num_threads` is the same in all interfaces. Here for `num_threads=1` you do spawn single thread and keep the caller waiting. I do prefer the other approach, where caller does the work and we spawn `num_threads-1` threads, as done in #4270. It avoids any possible overhead and does not rely on the OS scheduling to make sure the caller is not assigned resources when it is waiting.
> 
> The changes in #4270 are inconsistent with the changes that came before it. Everywhere apart from in #4270, `num_threads=0` means spawn no (zero) threads, and `num_threads=n` (n > 0) means spawn `n` threads. This also allows non-threaded and threaded code paths to be tested for the same use case of a 'single worker'.

That's not true, see what GJK does, https://github.com/FEniCS/dolfinx/blob/02343da3754f693b142aa7dc3a45bbd01aae8d2d/cpp/dolfinx/geometry/gjk.h#L539-L542
or also what ordering does, https://github.com/FEniCS/dolfinx/blob/02343da3754f693b142aa7dc3a45bbd01aae8d2d/cpp/dolfinx/graph/ordering.cpp#L194-L198

Both have for num_threads=1 serial, non-threaded codepath. This has to be unified, thats a primary concern. Other issues we can benchmark and fix uniformly later.

garth-wells:
Update for `num_thread` consistency.

python/dolfinx/mesh.py
```diff
@@ -807,6 +808,8 @@ def create_mesh(
             cells across MPI ranks.
         max_facet_to_cell_links: Maximum number of cells a facet can
             be connected to.
+        num_threads: Number of threads to use tp build mesh. Must be
```
chrisrichardson:
```suggestion
        num_threads: Number of threads to use to build mesh. Must be
```

PR #4303, opened 2026-07-19
Unify num_threads meaning

michalhabera:
@garth-wells This unifies meaning of all num_threads to
- num_threads=0: serial codepath, no spawning,
- num_threads=n: there are n threads spawned, while main, caller "thread" waits for the `std::jthread` block to join.

I've also run a quick benchmark using 
[num_threads_speedup.py](https://github.com/user-attachments/files/30169779/num_threads_speedup.py)

`python3 num_threads_speedup.py --repeats 3 --mesh-n 60 --num-bodies 50000 --max-threads 10 --out num_threads_speedup_test.png`

My M4 CPU only has 4 performance cores, so the scaling of entity permutations and GJK looks good. But also, this was run inside the docker container and not the best build flags, so take with a pinch of salt.

num_threads_speedup_test

Used Claude Code Opus 4.8.

### DIFF
--- a/python/dolfinx/wrappers/mesh.cpp
+++ b/python/dolfinx/wrappers/mesh.cpp
@@ -149,7 +149,7 @@ void mesh(nb::module_& m)
                                                num_threads);
       },
       nb::arg("topology"), nb::arg("dim"), nb::arg("entity_type"),
-      nb::arg("num_threads") = 1);
+      nb::arg("num_threads") = 0);
   m.def("compute_connectivity", &dolfinx::mesh::compute_connectivity,
         nb::arg("topology"), nb::arg("d0"), nb::arg("d1"));
 
@@ -217,10 +217,10 @@ void mesh(nb::module_& m)
           nb::arg("cell_type"), nb::arg("vertex_map"), nb::arg("cell_map"),
           nb::arg("cells"), nb::arg("original_index").none())
       .def("create_entities", &dolfinx::mesh::Topology::create_entities,
-           nb::arg("dim"), nb::arg("num_threads") = 1)
+           nb::arg("dim"), nb::arg("num_threads") = 0)
       .def("create_entity_permutations",
            &dolfinx::mesh::Topology::create_entity_permutations,
-           nb::arg("num_threads"))
+           nb::arg("num_threads") = 0)
       .def("create_connectivity", &dolfinx::mesh::Topology::create_connectivity,
            nb::arg("d0"), nb::arg("d1"))
       .def(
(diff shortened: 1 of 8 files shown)


python/dolfinx/wrappers/mesh.cpp
```diff
@@ -217,10 +217,10 @@ void mesh(nb::module_& m)
           nb::arg("cell_type"), nb::arg("vertex_map"), nb::arg("cell_map"),
           nb::arg("cells"), nb::arg("original_index").none())
       .def("create_entities", &dolfinx::mesh::Topology::create_entities,
-           nb::arg("dim"), nb::arg("num_threads") = 1)
+           nb::arg("dim"), nb::arg("num_threads") = 0)
       .def("create_entity_permutations",
            &dolfinx::mesh::Topology::create_entity_permutations,
-           nb::arg("num_threads"))
+           nb::arg("num_threads") = 0)
```
michalhabera, line 223: I am not really sure what defaults policy we have: default in python layer, default in nanobind and defaults in C++. We need to stick to a decision, then a quick AI pass will unify all for us.

garth-wells: We avoid defaults in the nanobind layer, and handle any defaults in the Python later.

garth-wells:
In #4304 I've made consistency fixes such that:
- num_threads=1: serial codepath, no spawning,
- num_threads=n (n > 1): there are n-1 threads spawned, and the main, caller "thread" participates
- num_threads=0: error

Turns out that the code for the above is very simple. Asking Claude, it seems that Intel TBB and the parallel Boost.Sort algorithms do the above.

Do we want this, or the change in this PR (which is what I did in my previous changes):
- num_threads=0: serial codepath, no spawning,
- num_threads=n: there are n threads spawned, while main, caller "thread" waits for the std::jthread block to join.

Pros and cons?

garth-wells:
From a web search, consensus is that the former (spawn n-1 threads) is the standard patter.

It avoids case where on a CPU with 4 cores we spawn 4 threads in addition to the working process ('thread'), giving 5 in total.

michalhabera:
Yes, I prefer that, and it is what I was suggesting in https://github.com/FEniCS/dolfinx/pull/4270. Num_threads=N means there are N execution lanes and the working process participates.

michalhabera:
> From a web search, consensus is that the former (spawn n-1 threads) is the standard patter.
> 
> It avoids case where on a CPU with 4 cores we spawn 4 threads in addition to the working process ('thread'), giving 5 in total.

In relation to oversubscribing by a single, idling main thread, see my comment https://github.com/FEniCS/dolfinx/pull/4270#discussion_r3578200132

PR #4304, opened 2026-07-19
Threading consistency fixes

garth-wells:
- `num_threads == 1`: Run on calling process
- `num_threads == n` (`n>1`): Spawn `n-1` threads *and* run on calling process
- `num_threads < 1`: Raise exception

And avoid clobbering by capturing reference to all objects.

### DIFF
--- a/cpp/dolfinx/geometry/gjk.h
+++ b/cpp/dolfinx/geometry/gjk.h
@@ -520,6 +520,7 @@ compute_distances_gjk(const std::vector<std::span<const T>>& bodies,
                       std::span<const T> q, size_t num_threads)
 {
   size_t total_size = bodies.size();
+  assert(num_threads > 0);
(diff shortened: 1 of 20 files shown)


cpp/dolfinx/geometry/gjk.h
```diff
@@ -517,16 +518,21 @@ template <std::floating_point T,
           typename U = boost::multiprecision::cpp_bin_float_double_extended>
 std::vector<T>
 compute_distances_gjk(const std::vector<std::span<const T>>& bodies,
-                      std::span<const T> q, size_t num_threads)
+                      std::span<const T> q, int num_threads)
 {
-  size_t total_size = bodies.size();
-  num_threads = std::min(num_threads, total_size);
+  if (num_threads <= 0)
```
michalhabera, line 523: You throw here, but assert in other places.

PR #4312, opened 2026-07-21
Fix use of assert/throw

michalhabera:
Asserts must be used if and only if a failing assert means a bug in the code.

Used Claude Code Sonnet 5.

### DIFF
--- a/cpp/dolfinx/fem/DirichletBC.h
+++ b/cpp/dolfinx/fem/DirichletBC.h
@@ -587,7 +587,9 @@ public:
   {
     for (std::int32_t idx : _dofs0)
     {
-      assert(idx < (std::int32_t)markers.size());
+      if (idx >= (std::int32_t)markers.size())
+        throw std::runtime_error("Marker array is too short for the "
+                                 "boundary condition dofs.");
       markers[idx] = true;
     }
   }
--- a/cpp/dolfinx/fem/DirichletBC.h
+++ b/cpp/dolfinx/fem/DirichletBC.h
@@ -585,13 +585,14 @@ public:
   /// unchanged.
   void mark_dofs(std::span<std::int8_t> markers) const
   {
-    for (std::int32_t idx : _dofs0)
+    if (!_dofs0.empty()
+        and *std::ranges::max_element(_dofs0) >= (std::int32_t)markers.size())
     {
-      if (idx >= (std::int32_t)markers.size())
-        throw std::runtime_error("Marker array is too short for the "
-                                 "boundary condition dofs.");
-      markers[idx] = true;
+      throw std::runtime_error("Marker array is too short for the boundary "
+                               "condition dofs.");
     }
+    for (std::int32_t idx : _dofs0)
+      markers[idx] = true;
   }
 
 private:
--- a/cpp/dolfinx/fem/DirichletBC.h
+++ b/cpp/dolfinx/fem/DirichletBC.h
@@ -585,12 +585,14 @@ public:
   /// unchanged.
   void mark_dofs(std::span<std::int8_t> markers) const
   {
+#ifndef NDEBUG
     if (!_dofs0.empty()
         and *std::ranges::max_element(_dofs0) >= (std::int32_t)markers.size())
     {
       throw std::runtime_error("Marker array is too short for the boundary "
                                "condition dofs.");
     }
+#endif
     for (std::int32_t idx : _dofs0)
       markers[idx] = true;
   }
--- a/cpp/dolfinx/graph/AdjacencyList.h
+++ b/cpp/dolfinx/graph/AdjacencyList.h
@@ -14,6 +14,7 @@
 #include <optional>
 #include <span>
 #include <sstream>
+#include <stdexcept>
 #include <utility>
 #include <vector>
 
@@ -67,7 +68,9 @@ public:
   AdjacencyList(U&& data, V&& offsets)
       : _array(std::forward<U>(data)), _offsets(std::forward<V>(offsets))
   {
-    assert(_offsets.back() == (std::int32_t)_array.size());
+    if (_offsets.back() != (std::int32_t)_array.size())
+      throw std::runtime_error("Last offset must equal the size of the data "
+                               "array.");
   }
 
   /// @brief Construct adjacency list from arrays of link (edge) data,
@@ -89,10 +92,13 @@ public:
       : _array(std::forward<U>(data)), _offsets(std::forward<V>(offsets)),
         _node_data(std::forward<W>(node_data))
   {
-    assert(_node_data.has_value()
-           and _node_data->size() == _offsets.size() - 1);
+    if (!_node_data.has_value() or _node_data->size() != _offsets.size() - 1)
+      throw std::runtime_error("Node data size must equal the number of "
+                               "nodes.");
     _array.reserve(_offsets.back());
-    assert(_offsets.back() == (std::int32_t)_array.size());
+    if (_offsets.back() != (std::int32_t)_array.size())
+      throw std::runtime_error("Last offset must equal the size of the data "
+                               "array.");
   }
 
   /// Set all connections for all entities (T is a '2D' container, e.g.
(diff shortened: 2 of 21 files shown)


garth-wells:
> Asserts must be used if and only if a failing assert means a bug in the code.
> 
> Used Claude Code Sonnet 5.

There is a caveat; don't use throw inside hot loops.

We should use throw when checking index bounds to vectors are and spans. If we really want index check (which for performance we often don't), use `std::vector::at`.

michalhabera:
> > Asserts must be used if and only if a failing assert means a bug in the code.
> > Used Claude Code Sonnet 5.
> 
> There is a caveat; don't use throw inside hot loops.
> 
> We should use throw when checking index bounds to vectors are and spans. If we really want index check (which for performance we often don't), use `std::vector::at`.

True, but if there is a throw in a hot loop, then it should not have been there in the first place. In many cases it can be moved outside the loop, and if it cannot be moved, then for maximizing throughput we should have a separate validation and evaluation passes.

I'll iterate this PR more to eliminate hot loops exceptions.

michalhabera:
Fixed hot loop exceptions. Backed two places from assert to exceptions which are potentially dangerous -- as per Copilot's suggestion. Ready for review.

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -585,11 +585,14 @@ class DirichletBC
   /// unchanged.
   void mark_dofs(std::span<std::int8_t> markers) const
   {
-    for (std::int32_t idx : _dofs0)
+    if (!_dofs0.empty()
```
garth-wells, line 589: We've followed that plain exceptions are used for O(1) checks. For higher complexity checks we should use only in DEBUG mode and document the requiments.

cpp/dolfinx/graph/AdjacencyList.h
```diff
@@ -67,7 +68,9 @@ class AdjacencyList
   AdjacencyList(U&& data, V&& offsets)
       : _array(std::forward<U>(data)), _offsets(std::forward<V>(offsets))
   {
-    assert(_offsets.back() == (std::int32_t)_array.size());
+    if (_offsets.back() != (std::int32_t)_array.size())
+      throw std::runtime_error("Last offset must equal the size of the data "
```
garth-wells, line 73: Multiline exception - use curly braces.

cpp/dolfinx/graph/AdjacencyList.h
```diff
@@ -89,10 +92,13 @@ class AdjacencyList
       : _array(std::forward<U>(data)), _offsets(std::forward<V>(offsets)),
         _node_data(std::forward<W>(node_data))
   {
-    assert(_node_data.has_value()
-           and _node_data->size() == _offsets.size() - 1);
+    if (!_node_data.has_value() or _node_data->size() != _offsets.size() - 1)
```
garth-wells, line 97: Curly braces.

cpp/dolfinx/graph/AdjacencyList.h
```diff
@@ -89,10 +92,13 @@ class AdjacencyList
...
+      throw std::runtime_error("Node data size must equal the number of "
+                               "nodes.");
     _array.reserve(_offsets.back());
-    assert(_offsets.back() == (std::int32_t)_array.size());
+    if (_offsets.back() != (std::int32_t)_array.size())
```
garth-wells, line 103: Curly braces.

cpp/dolfinx/io/VTKHDF.h
```diff
@@ -419,7 +420,9 @@ mesh::Mesh<U> read_mesh(MPI_Comm comm, const std::filesystem::path& filename,
   H5Dclose(dset_id);
 
   // Remove coordinates if gdim != 3
-  assert(gdim <= 3);
+  if (gdim > 3)
+    throw std::runtime_error("Geometric dimension must be less than or equal "
```
garth-wells, line 425: Curly braces.

cpp/dolfinx/io/xdmf_utils.cpp
```diff
@@ -29,7 +29,9 @@ xdmf_utils::get_cell_type(const pugi::xml_node& topology_node)
 {
   assert(topology_node);
   pugi::xml_attribute type_attr = topology_node.attribute("TopologyType");
-  assert(type_attr);
+  if (!type_attr)
+    throw std::runtime_error("XDMF topology node has no \"TopologyType\" "
```
garth-wells, line 34: Curly braces.

cpp/dolfinx/io/xdmf_utils.cpp
```diff
@@ -95,7 +97,9 @@ xdmf_utils::get_hdf5_paths(const pugi::xml_node& dataitem_node)
   // Split string into file path and HD5 internal path
   std::vector<std::string> paths;
   boost::split(paths, path, boost::is_any_of(":"));
-  assert(paths.size() == 2);
+  if (paths.size() != 2)
+    throw std::runtime_error("XDMF DataItem path \"" + path
```
garth-wells: Use std::format  and add curly braces.

cpp/dolfinx/mesh/utils.h
```diff
@@ -1100,7 +1101,9 @@ Mesh<typename std::remove_reference_t<typename U::value_type>> create_mesh(
     for (std::int32_t i = 0; i < num_cell_types; ++i)
     {
       std::size_t num_cell_nodes = doflayouts[i].num_dofs();
-      assert(cells[i].size() % num_cell_nodes == 0);
+      if (cells[i].size() % num_cell_nodes != 0)
+        throw std::runtime_error("Cell array size is not a multiple of the "
```
garth-wells, line 1106: Curly braces.

cpp/dolfinx/mesh/utils.h
```diff
@@ -1133,7 +1136,9 @@ Mesh<typename std::remove_reference_t<typename U::value_type>> create_mesh(
     {
       cells1[i] = std::vector<std::int64_t>(cells[i].begin(), cells[i].end());
       std::int32_t num_cell_nodes = doflayouts[i].num_dofs();
-      assert(cells1[i].size() % num_cell_nodes == 0);
+      if (cells1[i].size() % num_cell_nodes != 0)
+        throw std::runtime_error("Cell array size is not a multiple of the "
```
garth-wells, line 1143: Curly braces

python/dolfinx/fem/forms.py
```diff
@@ -288,7 +288,8 @@ def mixed_topology_form(
 
     # Check that subdomain data for each integral type is the same
     for data in sd.get(domain).values():
-        assert all([d is data[0] for d in data if d is not None])
```
garth-wells, line 291: Does assert vs raise make any practical different in Python? Support try/except?

garth-wells (changes requested):

python/dolfinx/fem/forms.py
schnellerhase: Assertions are not guaranteed to persist the python build/runtime https://docs.astral.sh/ruff/rules/assert/.

michalhabera: Also, users can potentially disable asserts with `python3 -O script.py`, see https://docs.python.org/3/using/cmdline.html#cmdoption-O

cpp/dolfinx/fem/DirichletBC.h
```diff
@@ -585,11 +585,16 @@ class DirichletBC
   /// unchanged.
   void mark_dofs(std::span<std::int8_t> markers) const
   {
-    for (std::int32_t idx : _dofs0)
+#ifndef NDEBUG
+    if (!_dofs0.empty()
+        and *std::ranges::max_element(_dofs0) >= (std::int32_t)markers.size())
```
schnellerhase, line 590: Isn't a `NDEBUG` marked raise just an assertion?

garth-wells: It can be caught, but in practice there isn't much of a difference.

garth-wells (approved):

cpp/dolfinx/fem/DirichletBC.h
michalhabera: @schnellerhase Semantically, not really, because exceptions are about user errors. This is a user error. Performance implications could be thought of as an independent aspect.

PR #4314, opened 2026-07-23
Use std::format to format strings

garth-wells:
Simplify string formatting by using `std::format`

Use Claude Sonnet 5.


francesco-ballarin (approved):
I manually checked that all affected files had `#include <format>` (not needed by CI and any recent compiler, but required by older compilers), because in the past this often required a fixup PR.

PR #4323, opened 2026-07-27
math.h constexpr and mdspan2 concept

jhale:
Adds global mdspan2 concept and constrains math.h functions more tightly.


cpp/dolfinx/common/types.h
```diff
@@ -50,4 +51,17 @@ using scalar_value_t = typename scalar_value<T>::type;
 /// @private mdspan/mdarray namespace
 namespace md = MDSPAN_IMPL_STANDARD_NAMESPACE;
 
+/// @private Concept for a rank-2 mdspan-like type: exposes a
+/// `value_type`, has static `rank() == 2`, and supports two-index access
+/// and per-dimension `extent` queries. cv- and reference-qualifiers on
+/// `T` are stripped, so it can also constrain forwarding-reference
+/// parameters.
+template <typename T>
+concept mdspan2 = requires(std::remove_cvref_t<T> x, std::size_t i) {
```
garth-wells: I think `mdspan2` is a confusing name since we often use it as a typedef. Can it be made more distinctive?

jhale: Changed to `MDSpanRank2`.

PR #4328, opened 2026-07-27
Wait after non-blocking send before function return

jhale:
The function posts `MPI_Isends` that read from `send_buffer`, a local vector destroyed on return. MPI requires the send buffer stay valid until the request completes. `MPI_Waitall` enforces local send completion before `return`.


garth-wells (commented):
We don't use PCX - shall we just remove it?

PR #4329, opened 2026-07-27
Harden IndexMap communication and modernise C++ usage

jhale:
Correctness/robustness:
- compute_owned_indices/shared_indices: pack the ghost send buffer from a single sorted (owner, index) pair array instead of sorting the index and owner arrays independently. This groups the buffer by owning rank by construction rather than relying on the rank-order == global-index- order partitioning invariant holding.
- IndexMap constructors: issue MPI_Iallreduce on the duplicated _comm.comm() rather than the caller's raw comm, so it cannot match unrelated user collectives on the original communicator.

Modern C++20 / style:
- Replace legacy std::find/std::upper_bound with std::ranges equivalents using pair::first projections.
- Replace C-style (int) casts with static_cast.
- Make size/global-index-difference narrowings to int32_t explicit with static_cast (num_ghosts(), size_local() and internal locals).

Last point on `size_t` to `int32_t` explicit cast open to discussion.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -1338,7 +1363,7 @@ std::vector<std::int32_t> IndexMap::weights_dest() const
   std::vector<std::int32_t> w_dest(_dest.size());
   for (std::size_t i = 0; i < _dest.size(); ++i)
   {
-    ierr = MPI_Irecv(w_dest.data() + i, 1, MPI_INT32_T, _dest[i], MPI_ANY_TAG,
+    ierr = MPI_Irecv(w_dest.data() + i, 1, MPI_INT32_T, _dest[i], 0,
```
jhale: The tag now matches the call to `MPI_Isend` below.

cpp/dolfinx/common/IndexMap.cpp
IgorBaratta: This `MPI_Iexscan` still uses the caller’s raw comm.
Should it also use `_comm.comm()` here, as it already does in the other constructor?

cpp/dolfinx/common/IndexMap.cpp
```diff
@@ -300,8 +301,8 @@ compute_submap_indices(const IndexMap& imap,
             = std::ranges::unique(dest_begin, new_owner_dest_ranks.end());
         new_owner_dest_ranks.erase(unique_end, range_end);
 
-        std::size_t num_unique_dest_ranks
-            = std::distance(dest_begin, unique_end);
+        std::int32_t num_unique_dest_ranks
```
garth-wells: This look like noise to me - what's wrong with `std::size_t`?

cpp/dolfinx/common/IndexMap.cpp
garth-wells: Can the MPI calls be replaced by a neighbourhood collective?

cpp/dolfinx/common/IndexMap.cpp
jhale: The LLM wants to make it explicit that you cast a std::size_t into a std::int32_t vector on the following line.

jhale: Open to discuss these things as then we can add guidance into `AGENTS.md` on `std::int32_t` vs `std::int64_t`.

PR #4336, opened 2026-07-28
Build vcpkg SCOTCH and SuperLU-DIST

jhale:
This is an attempt to build SCOTCH (rather than Parmetis which is non-FOSS) and SuperLU_DIST on Windows to enable a Windows wheel build. The SCOTCH and SuperLU_DIST patches should be moved upstream.


jhale:
I'm going to port these back upstream, leaving this here so I remember to do it.

jhale:
The Windows build patches are now in forks.

PR #4341, opened 2026-07-30
Deduplicate compile-time/runtime block-size loops in assembly and CSR spmv

garth-wells:
## Summary

An alternative, smaller-footprint way to address the code-duplication objective of #3716 (`Introduce BlockSize`), without introducing the generic `ConstexprType`/`BlockSize` concept machinery.

The `if constexpr (_bs > 0) { ... } else { ... }` branches in `assemble_cells`, `assemble_entities`, `assemble_interior_facets` (`fem/assemble_vector_impl.h`) and `spmv`/`spmvT` (`la/matrix_csr_impl.h`) duplicated the entire scatter/accumulate loop body for the compile-time and runtime block-size paths.

Rather than retaining the `if constexpr` dispatch inside each function (factoring only the loop body into a shared lambda), the block size is now carried as a single value that is either a runtime `int` or a compile-time `std::integral_constant<int, N>`, and the scatter/accumulate loop is written once, unconditionally, in terms of that value:

```cpp
for (std::size_t i = 0; i < dofs.size(); ++i)
  for (int k = 0; k < bs; ++k)
    b[bs * dofs[i] + k] += be[bs * i + k];
```

`std::integral_constant<int, N>` converts implicitly to `int`, so this single loop is valid for both cases, and the compiler can still constant-fold/unroll it for the `bs = 1, 3` fast paths since `bs` is a literal `std::integral_constant` there — the unrolling guarantee is preserved by the type system, not left to the optimizer to reconstruct from a plain `int`, which was the concern that motivated moving away from #3716's `consteval`-accessor approach in the first place (see [my review comment on #3716](https://github.com/FEniCS/dolfinx/pull/3716#issuecomment-5124163026)). The dispatch between the two now happens once, at the call site (`if (bs == 1) ... else if (bs == 3) ... else ...`), rather than being re-checked with `if constexpr` inside every function.

The same treatment is applied to `la::impl::spmv`/`spmvT`: the `template <typename T, int BS1>` non-type template parameter and its `if constexpr (BS1 == -1)` branch are replaced by plain `auto bs0, auto bs1` parameters, with `MatrixCSR::mult`/`multT` passing either `std::integral_constant<int, 1>{}` or the runtime `_bs[1]`.

Since `assemble_cells`/`assemble_entities`/`assemble_interior_facets` now take the dofmap/block-size/index data as an unconstrained `const auto& dofmap`, three concepts (`DofMapPackCells`, `DofMapPackEntities`, `DofMapPackFacets`) were added to `fem/traits.h` to give this parameter a real, checked type constraint again. They share a common base (`DofMapPackBase`: a 3-tuple whose first entry is a rank-2 `const std::int32_t` dofmap mdspan and second entry is convertible to `int`) and differ only in how the third entry (cell/entity/facet indices) is indexed, since that shape genuinely differs between the three assembly kernels — a flat span for cells, a rank-2 mdspan for entities, a rank-3 mdspan for interior facets.

This does not address #3716's secondary objective (block size specified twice — once as a template parameter, once via the runtime dofmap tuple, checked only by `assert`); that's a separate, independent concern.

## Test plan
- [x] `clang-format --dry-run --Werror` clean on all touched files
- [x] Full `libdolfinx` rebuild (`-Werror`) clean
- [x] `cpp/test` unittests: 68/68 passed, 27210/27210 assertions, including `[la_matrix]` (exercises `spmv`/`spmvT` directly)
- [x] Python bindings rebuilt; smoke-tested vector assembly and matrix-vector product (`A.mult`) at bs=1, bs=2, bs=3 — results match expected sums in each case

### DIFF
--- a/cpp/demo/custom_kernel/main.cpp
+++ b/cpp/demo/custom_kernel/main.cpp
@@ -173,10 +173,10 @@ double assemble_vector1(const mesh::Geometry<T>& g, const fem::DofMap& dofmap,
   common::Timer timer("Assembler1 lambda (vector)");
   std::vector<T> cdofs_b(3 * g.dofmaps().front().extent(1));
   std::vector<T> be_b(dofmap.map().extent(1));
-  fem::impl::assemble_cells<1>([](auto, auto, auto, auto) {}, b.array(),
-                               g.dofmaps().front(), x, cells,
-                               {dofmap.map(), 1, cells}, kernel, {}, {}, {},
-                               std::span(be_b), std::span(cdofs_b));
+  fem::impl::assemble_cells(
+      [](auto, auto, auto, auto) {}, b.array(), g.dofmaps().front(), x, cells,
+      std::tuple{dofmap.map(), std::integral_constant<int, 1>{}, cells}, kernel,
+      {}, {}, {}, std::span(be_b), std::span(cdofs_b));
   b.scatter_rev(std::plus<T>());
   return la::squared_norm(b);
 }
--- a/cpp/dolfinx/fem/assemble_vector_impl.h
+++ b/cpp/dolfinx/fem/assemble_vector_impl.h
@@ -74,7 +74,7 @@ template <typename V, std::floating_point U,
 void assemble_cells(
     const fem::DofTransformKernel<T> auto& P0, V&& b, mdspan2_t x_dofmap,
     md::mdspan<const U, md::extents<std::size_t, md::dynamic_extent, 3>> x,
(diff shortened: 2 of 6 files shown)


jorgensd:
In general I'm fine with the PR (up to some documentation issues highlighted by copilot above).
The only thing I'd like to discuss is the naming of `DofMapPackCells`, `DofMapPackEntities` and `DofMapPackFacets` which I don't think are meaningful names for a general trait.
These should probably just be PackedCells, PackedEntities, PackedInteriorEntities as that is what the trait describes (nothing dofmap specific).

PR #4342, opened 2026-07-30
Add nanobind stubgen (fixes mypy failures from #4220)

garth-wells:
## Summary

Continues #4220 (`jhale/nanobind-autogen-stubs`) with the mypy fix that PR needed: the auto-generated `dolfinx.cpp` stubs give mypy real, precise types for the compiled extension for the first time, which surfaced ~446 errors in the "Build and test" CI job (the only mypy invocation that actually installs dolfinx before running mypy, so the only one exercising the generated stubs).

- Fixed a real bug in the stub-generation CMake setup: on Unix the compiled module was imported as bare `cpp` instead of `dolfinx.cpp`, so nanobind wrote cross-submodule references like `import cpp.la`, which doesn't exist and silently resolved to `Any` under mypy — masking real errors and producing bogus "overload can never match" diagnostics. Now stages the built module under a throwaway `dolfinx/` package dir before invoking `nanobind_add_stub` so the module's real `__name__` is `dolfinx.cpp`.
- Fixed `_IntegralType`/`MPICommWrapper` bindings that used names/`const_name()` values with no resolvable Python type, which broke stub cross-references.
- Fixed `FiniteElement`/`AdjacencyList` `__eq__` bindings that exposed the raw C++ `operator==`, violating `object.__eq__`'s Liskov contract.
- Removed dead, unreachable overloads (complex instantiations of `interpolation_matrix`/`discrete_curl`/`discrete_gradient` that are always shadowed by the real ones, and a redundant `read_geometry_data` registration).
- Added missing `nanobind/stl/map.h` and `.../string.h` includes that were causing stubgen to fall back to invalid raw C++ type strings.
- ~250 Python-side errors are the same runtime-safe-but-statically-unverifiable scalar-type dispatch pattern already used elsewhere in this PR (`fem/petsc.py`); fixed with matching `# type: ignore`. A handful were real bugs instead (wrong return-type annotations in `forms.py`, dispatch-table locals missing an explicit `Union` annotation).
- The ~24 remaining errors are genuine ecosystem-level gaps (mypy can't disambiguate numpy dtype/rank in `NDArray` annotations; nanobind has no `std::reference_wrapper` caster; basix's C++ types aren't stub-resolvable across the package boundary), confirmed via isolated repros rather than dolfinx bugs. Suppressed per-file via a new nanobind stub pattern file (`stub_patterns.txt`) that injects a scoped `# mypy: disable-error-code=...` into just the affected generated stubs.

Rebased on current `main` (merge commit included) and re-verified after the merge.

## Test plan

- [x] `mypy --config-file python/pyproject.toml -p dolfinx` → clean (matching the failing job's exact PETSc/ADIOS2-disabled config)
- [x] `ruff check` / `ruff format --check` clean
- [x] `clang-format --dry-run --Werror` clean on touched C++ files
- [x] Full C++ build with PETSc+SLEPc+ADIOS2+ParMETIS+SuperLU_DIST (`-Werror`) compiles cleanly
- [x] Runtime sanity checks (`CellType.name`, `__eq__` against unrelated objects, MPI communicator casting) pass


jhale:
Just to note that Windows stubgen should be fixed in Basix first before we try and fix it here, we can then port what we understand to here.

python/CMakeLists.txt
```diff
@@ -31,6 +31,13 @@ add_feature_info(
 	to set FALSE."
 )
 
+option(ENABLE_NANOBIND_STUBGEN "Run nanobind stub generation" ON)
```
jhale: Should probably make this DOLFINX_ENABLE...

python/CMakeLists.txt
```diff
@@ -190,4 +197,103 @@ set_target_properties(
...
+        dolfinx/cpp/la/petsc.pyi
+        dolfinx/cpp/log.pyi
+        dolfinx/cpp/mesh.pyi
+        dolfinx/cpp/nls.pyi
+        dolfinx/cpp/nls/__init__.pyi
+        dolfinx/cpp/nls/petsc.pyi
+        dolfinx/cpp/refinement.pyi
+    )
+  else()
+    # On UNIX-like systems we can import the cpp compiled module before
```
jhale: Seems like something we should ask about upstream, not very pretty.

garth-wells: Add link to docs, https://nanobind.readthedocs.io/en/latest/api_cmake.html#command:nanobind_add_stub

garth-wells:
> Just to note that Windows stubgen should be fixed in Basix first before we try and fix it here, we can then port what we understand to here.

Basix working with Windows now.

PR #4349, opened 2026-08-01
Add note on PETSc build in rigid body mode null space construction

schnellerhase:


garth-wells:
Building the nullspace (without PETSc) is an important illustrative example. It should be kept.

jorgensd:
I'm not fully decided on what is better here.. 
As this null-space is directly used for the PETSc operators, I think it would be nice to simplify it to the proposed demo.
Of course, I can see the point that if you have slightly different null spaces (such as two disjoint bodies in a single `dolfinx.mesh.Mesh`), the construction we have in our code is more easily extendible to cover this case than using the PETSc library function.

jhale:
I'd leave it as was with a quick comment that it's possible to achieve the same thing for 3D elasticity using PETSc; if you use rst links it should even resolve to the method in the petsc4py docs.

schnellerhase:
I have no strong preference on this one; the reason I stumbled into this was because I wanted to extend an elasticity solver elsewhere taking this demo as baseline. Working with the NS construction from PETSc is then just more concise and easier to read so I opted for that. Guess its a question on what this demo should highlight.

garth-wells:
The demos are for 'demonstrating'; the objective is not the shortest possible code. Building the nullspace demonstrates a number of features. 

Smoothed aggregation solvers need the near nullspace and there are solvers other than PETSC GAMG. This demo shows how one would build the near nullspace for other solvers.

jorgensd (approved):

python/demo/demo_elasticity.py
```diff
@@ -202,7 +211,7 @@ def σ(v):
 opts["mg_levels_ksp_chebyshev_esteig_steps"] = 10
 
 # Create PETSc Krylov solver and turn convergence monitoring on
-solver = PETSc.KSP().create(msh.comm)
+solver = PETSc.KSP().create(comm)
```
jorgensd: I think it is safer if all the below operators (including XDMFFile) uses mesh.comm.

jorgensd (approved):

PR #4350, opened 2026-08-01
Wrapping more return types as 'Python' types rather than nanobind types

garth-wells:
There are places where methods return 'raw' nanobind wrapped types when we have a Python class that holds the nanobind wrapped object. Returning raw types leads to some inconsistencies in behaviour, missing docs and less effective type hints.

This PR tidies up a number of these return types, and adds Python wrappers where one was missing. All changes are in the Python interface; the C++ library is untouched.

### Return types now wrapped

- `Form.mesh` returns a `mesh.Mesh`, and `Form.function_spaces` a tuple of `fem.FunctionSpace`. Both were annotated as the Python types but returned the C++ objects. The Python objects are now stored on the `Form`, as is done for the mesh, so identity is preserved (`a.mesh is msh`).
- `Topology.connectivity` and `MeshTags.topology` return `graph.AdjacencyList` and `mesh.Topology`. `connectivity` also accepts the mixed-topology `(dim, entity type index)` pairs that the C++ method already supported. A dimension is anything integer-like, a NumPy integer included; the two arguments must be both dimensions or both pairs.
- `DofMap.dof_layout` and `CoordinateElement.create_dof_layout` return the new `fem.ElementDofLayout`.
- `fem.create_sparsity_pattern` returns the new `la.SparsityPattern`.
- `mesh.build_dual_graph` and `fem.transpose_dofmap` were bare `dolfinx.cpp` re-exports returning a C++ adjacency list. Both are now Python functions returning `graph.AdjacencyList`.

### New Python wrappers

- `la.SparsityPattern`, wrapping `index_map`, `num_nonzeros`, `insert`, `insert_diagonal`, `finalize` and `graph`, with `la.sparsity_pattern` and `la.sparsity_pattern_blocked` factories covering the two C++ constructors. This was the most visible gap: the standard create/insert/finalize/`matrix_csr` sequence called C++ methods directly, and constructing a pattern at all required importing from `dolfinx.cpp.la`.
- `fem.ElementDofLayout`, wrapping `num_dofs`, `block_size`, `entity_dofs` and `entity_closure_dofs`.

### Arguments that required `dolfinx.cpp` objects

- `graph.distribute`'s `destinations` argument and `mesh.create_mesh`'s `reorder_fn` callback took/passed raw C++ adjacency lists, so user code could not avoid `dolfinx.cpp`. Both now use `graph.AdjacencyList`.
- `graph.reorder_rcm` gains a Python counterpart (the default cell reordering, and the natural thing to pass as `reorder_fn`), and `mesh.cell_num_entities` is re-exported.
- `la.matrix_csr` and `fem.build_sparsity_pattern` take a `la.SparsityPattern`.

### Wrapper identity and caching

The wrappers are thin handles around a shared C++ object, and several are rebuilt on each access, so comparing two of them fell back to identity and silently returned `False` — including the 'both dofmaps have the same index map' check that `demo_mixed-topology.py` documents.

- `AdjacencyList`, `DofMap`, `ElementDofLayout`, `Topology` and `Geometry` define `__eq__`/`__hash__` delegating to the wrapped object. nanobind hands back the same Python object for a given C++ pointer, so delegating is exact and the hash stays consistent with equality.
- `DofMap.dof_layout` and `Geometry.cmaps` are `functools.cached_property`, next to the existing `FunctionSpace.element`/`.dofmap`, rather than building a wrapper per access.
- Collections that are fixed by the object they are derived from are returned as tuples: `FunctionSpace.dofmaps`, `Geometry.cmaps` and `Form.function_spaces`. Each was a list copied out of a cached tuple on every access, purely to stop callers mutating the cache; a tuple says that directly and drops the copy. Every caller in the library and the tests only indexes them.

### API changes for callers

- `Form.__init__` takes the mesh and the argument function spaces (`Form(form, msh, spaces, ...)`). Forms are normally built with `fem.form`; the custom-kernel route in `demo_static-condensation.py` shows the updated call.
- Callers that relied on `Form.function_spaces` returning C++ spaces now take `._cpp_object` explicitly. The normalise-either-type branch in `bcs_by_block`, added for that leak, is removed.
- `FunctionSpace.dofmaps`, `Geometry.cmaps` and `Form.function_spaces` return tuples rather than lists.
- `Mesh.__init__` attaches the Python `Mesh` to the UFL domain rather than the C++ mesh, so `ufl.Mesh.ufl_cargo` returns the Python object. `Mesh` and `ufl.Mesh` then reference each other. The link has to be strong — `fem.form` needs `ufl_cargo` to give back that same `Mesh` — so a mesh is now reclaimed by the cyclic garbage collector rather than by reference counting. Where a mesh, and the MPI communicator it holds, must be released at a known point, `gc.collect()` is the lever; it matters because `~Comm` frees the communicator collectively.

### Fixes

- `get_integration_domains` mixed `subdomain.topology` and `subdomain._cpp_object.topology`.
- The docstrings of `create_sparsity_pattern` and `build_sparsity_pattern` told callers to call `assemble` on the pattern; the method is `finalize`.

### Still to do

- `SparsityPattern.index_map` returns a C++ `IndexMap`, which has no Python wrapper yet. #4496, stacked on this PR, adds one and closes the gap.
- `demo_mixed-topology.py` still imports the mixed-topology `create_mesh` and `FunctionSpace_float64` from `dolfinx.cpp`, which have no pure-Python equivalent yet. Other demos no longer import from `dolfinx.cpp`.

### AGENTS.md

Two conventions that came up while doing this: Python demos must not import from `dolfinx.cpp`, and Python tests needing PETSc must live in a file with `petsc` in the name.

---

### DIFF
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -701,6 +701,7 @@ def functionspace(
     # Check that element and mesh cell types match
     if ((domain := mesh.ufl_domain()) is None) or ufl_e.cell != domain.ufl_cell():
         raise ValueError("Non-matching UFL cell and mesh cell shapes.")
+
     # Create DOLFINx objects
     dolfinx_element = finiteelement(mesh.topology.cell_type, ufl_e, dtype)
 
--- a/python/dolfinx/fem/function.py
+++ b/python/dolfinx/fem/function.py
@@ -848,15 +848,15 @@ class FunctionSpace(ufl.FunctionSpace, Generic[Real]):
         """Function space finite element."""
         return FiniteElement(self._cpp_object.element)
 
-    @property
+    @cached_property
     def dofmap(self) -> DofMap:
         """Degree-of-freedom map associated with the function space."""
         return DofMap(self._cpp_object.dofmap)
 
-    @property
+    @cached_property
     def dofmaps(self) -> list[DofMap]:
         """The geometry dofmaps, one per cell type."""
-        return [DofMap(_o) for _o in self._cpp_object.dofmaps]
+        return [DofMap(map) for map in self._cpp_object.dofmaps]
 
     @property
     def mesh(self) -> Mesh[Real]:
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -127,6 +127,21 @@ disclosure process.
   already matches a callback/`std::function` parameter exactly, pass
   the function directly (e.g. `graph::reorder_rcm`) rather than
   wrapping it in a trivial forwarding lambda.
+- **Lambdas**: no `[=]`/`[&]` — list captures explicitly, e.g.
+  `[&v]`. Capture by reference for lambdas invoked in
+  place; by value (cheap scalars, or
+  `[v = std::move(v)]` to move a container) for lambdas that escape the
+  scope, where a captured reference or `span` into a local dangles
+  silently. `[this]` only if the lambda cannot outlive the object.
+  Never capture a container or `shared_ptr` by value for convenience:
+  the copy happens at capture and again whenever the lambda or its
+  `std::function` is copied.
+  Prefer explicit parameter types over `auto` (`auto&&` in generic
+  code); add an explicit return type when the deduced one is non-obvious
+  or must not decay. Avoid `mutable`. Promote long or reused lambdas to
+  a free function in an anonymous namespace. A by-reference capture is
+  `const` only if the captured variable is, so declare read-only locals
+  `const`.
 - **Algorithms**: prefer `<algorithm>`/`<ranges>` (`std::ranges::...`)
   over hand-written loops where it doesn't hurt clarity or
   performance; flattened row-major storage is the default convention
--- a/cpp/dolfinx/common/sort.h
+++ b/cpp/dolfinx/common/sort.h
@@ -64,7 +64,7 @@ inline constexpr _unsigned_projection unsigned_projection{};
 /// @code
 /// std::array<std::int16_t, 3> a{2, 3, 1};
 /// std::array<std::int16_t, 3> i{0, 1, 2};
-/// dolfinx::radix_sort(i, [&](auto i){ return a[i]; }); // yields i = {2, 0,
+/// dolfinx::radix_sort(i, [&a](auto i){ return a[i]; }); // yields i = {2, 0,
 /// 1} and a[i] = {1, 2, 3};
 /// @endcode
 /// @tparam BITS The number of bits to sort at a time.
@@ -90,7 +90,7 @@ constexpr void radix_sort(R&& range, P proj = {})
 
   if constexpr (!std::is_same_v<uI, I>)
   {
-    radix_sort<_BITS>(std::forward<R>(range), [&](const T& e) -> uI
+    radix_sort<_BITS>(std::forward<R>(range), [&proj](const T& e) -> uI
                       { return unsigned_projection(proj(e)); });
     return;
   }
--- a/cpp/dolfinx/fem/DirichletBC.h
+++ b/cpp/dolfinx/fem/DirichletBC.h
@@ -497,7 +497,7 @@ public:
   {
     // set_fn is a lambda which gets evaluated for every index in [0,
     // _dofs0.size()) and its result is assigned to x[_dofs0[i]].
-    auto apply = [&](std::invocable<std::int32_t> auto set_fn)
+    auto apply = [this, &x](std::invocable<std::int32_t> auto set_fn)
     {
       static_assert(
           std::is_same_v<std::invoke_result_t<decltype(set_fn), std::int32_t>,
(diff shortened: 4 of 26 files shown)


python/dolfinx/fem/function.py
```diff
@@ -855,9 +855,13 @@ def dofmap(self) -> DofMap:
...
 
     @cached_property
+    def _dofmaps(self) -> tuple[DofMap, ...]:
+        return tuple(DofMap(map) for map in self._cpp_object.dofmaps)
+
+    @property
     def dofmaps(self) -> list[DofMap]:
         """The geometry dofmaps, one per cell type."""
-        return [DofMap(map) for map in self._cpp_object.dofmaps]
+        return list(self._dofmaps)
```
jorgensd: Don't we want this to be a tuple as well?

python/test/unit/fem/test_dofmap.py
```diff
@@ -112,6 +112,14 @@ def test_entity_dofs(mesh):
         assert all(d == cd for d, cd in zip(dofs, cdofs, strict=True))
 
 
+def test_dofmaps_is_immutable(mesh):
+    """Test that the returned dofmaps cannot be used to modify the space."""
```
jhale, line 116: Not immediately clear how the test implements the description, comment?

python/test/unit/la/test_sparsity_pattern.py
```diff
@@ -28,3 +29,17 @@ def test_add_diagonal():
...
     pattern.finalize()
     assert len(blocks) == pattern.num_nonzeros
+
+
+def test_blocked_pattern_with_empty_blocks():
+    """Test creation of a blocked pattern with structural zero blocks."""
+    index_map = IndexMap(MPI.COMM_SELF, 2)
+    pattern = sparsity_pattern(MPI.COMM_SELF, [index_map, index_map], [1, 1])
+    blocked_pattern = sparsity_pattern_blocked(
+        MPI.COMM_SELF,
```
jhale, line 42: Any reason to avoid to construct on `COMM_SELF`? Add comment.

jhale (approved):

python/dolfinx/fem/utils.py
```diff
@@ -24,18 +24,22 @@
...
 from dolfinx.fem.function import FunctionSpace
 from dolfinx.geometry import PointOwnershipData as _PointOwnershipData
 from dolfinx.la import MatrixCSR as _MatrixCSR
+from dolfinx.la import SparsityPattern as _SparsityPattern
 
 if typing.TYPE_CHECKING:
+    # 'dolfinx.la.SparsityPattern' is spelled out in the annotations
+    # below because a bare 'SparsityPattern' is ambiguous in the
+    # generated docs, matching 'dolfinx.cpp.la.SparsityPattern' too.
+    import dolfinx.la
```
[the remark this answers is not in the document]
garth-wells: Fixed in 5cdd400308: dropped the redundant plain `import dolfinx.la`. `import dolfinx.mesh` already binds the `dolfinx` package name, and `dolfinx/__init__.py` imports `la`, so `dolfinx.la.SparsityPattern` still resolves in the annotation without a second, conflicting import style for the same module.

python/dolfinx/fem/function.py
garth-wells: Already a tuple as of db7393c47f (`FunctionSpace.dofmaps` returns `tuple[DofMap, ...]`, cached). Also fixed a leftover copy-paste in its docstring in 5cdd400308 ("geometry dofmaps" -> "function space dofmaps").

python/test/unit/mesh/test_mesh.py
```diff
@@ -684,14 +684,44 @@ def reorder(dual_graph):
...
+    with pytest.raises(TypeError):
+        msh.topology.connectivity(2, (0, 0))
+
+
+def test_wrappers_compare_equal():
+    """Wrappers built around the same C++ object compare equal."""
+    msh = create_unit_square(MPI.COMM_WORLD, 3, 3)
+    msh.topology.create_connectivity(2, 0)
+    assert msh.topology.connectivity(2, 0) == msh.topology.connectivity(2, 0)
+    assert msh.geometry == msh.geometry
```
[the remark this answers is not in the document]
garth-wells: Fixed in 5cdd400308: rewrote the wrapper-equality assertions to compare two independently bound variables instead of the literal `x == x` form, so the false positive goes away while the test still exercises independently-built wrappers around the same C++ object comparing equal.

python/test/unit/fem/test_dofmap.py
garth-wells: Added a comment in 5cdd400308 explaining the assertions: a tuple can't be appended/assigned to, and identity across repeated access confirms it's the same cached tuple rather than a fresh copy a caller could otherwise mutate without effect.

python/test/unit/la/test_sparsity_pattern.py
garth-wells: Added a comment in 5cdd400308: COMM_SELF is used because the block structure under test is process-local and involves no cross-rank communication, so the test runs unmodified under any number of MPI ranks.

PR #4351, opened 2026-08-02
Fix O(n^2) Debug-build regression in interpolate/eval cell loops

### LINKED ISSUE #4347: Regression in unoptimised interpolate performance
Consider the following interpolate only example, identified by @jorgensd:

```cpp
// Copyright (C) 2026 Jørgen S. Dokken and Paul T. Kühner
//
// SPDX-License-Identifier:    MIT

#include <basix/finite-element.h>
#include <chrono>
#include <dolfinx.h>
#include <dolfinx/fem/CoordinateElement.h>
#include <dolfinx/io/XDMFFile.h>
#include <dolfinx/mesh/Mesh.h>
#include <dolfinx/mesh/cell_types.h>
#include <dolfinx/mesh/utils.h>

using T = double;

int main(int argc, char *argv[]) {
  MPI_Init(&argc, &argv);

  spdlog::set_level(spdlog::level::info);

  int n = 30;
  auto mesh =
      std::make_shared<dolfinx::mesh::Mesh<T>>(dolfinx::mesh::create_box(
          MPI_COMM_WORLD, {{{0, 0, 0}, {1, 1, 1}}}, {n, n, n},
          dolfinx::mesh::CellType::tetrahedron));

  auto cell_map = mesh->topology()->index_map(mesh->topology()->dim());
  spdlog::warn("Cells (local + ghosts): {}",
               cell_map->size_local() + cell_map->num_ghosts());

  auto element = std::make_shared<const dolfinx::fem::FiniteElement<T>>(
      basix::create_element<T>(
          basix::element::family::P,
          dolfinx::mesh::cell_type_to_basix_type(mesh->topology()->cell_type()),
          0, basix::element::lagrange_variant::unset,
          basix::element::dpc_variant::unset, true));
  auto V0 = std::make_shared<dolfinx::fem::FunctionSpace<T>>(
      dolfinx::fem::create_functionspace(mesh, element));
  auto lmbda = std::make_shared<dolfinx::fem::Function<T>>(V0);

  spdlog::warn("Starting interpolate for constant DG0 field");

  auto t0 = std::chrono::steady_clock::now();
  lmbda->interpolate(
      [](auto x) -> std::pair<std::vector<T>, std::vector<std::size_t>> {
        spdlog::warn("Entered lambda. x shape: ({}, {})", x.extent(0),
                     x.extent(1));
        std::vector<T> values(x.extent(1));
        std::fill(values.begin(), values.end(), 1.0);
        return {values, {values.size()}};
      });

  auto t1 = std::chrono::steady_clock::now();

  const double t_interpolate =
      std::chrono::duration_cast<std::chrono::duration<double>>(t1 - t0)
          .count();
  spdlog::warn("Interpolate finished in {:.6f} s", t_interpolate);
  
  MPI_Finalize();

  return 0;
}
```

[CMakeLists.txt](https://github.com/user-attachments/files/30581948/CMakeLists.txt)

This yields for an (unoptimised) `Debug` build:

| tag | runtime (s) |
| - | - |
| `main` | 52 |
| `v0.11.0.post0` | 53 |
| `v0.10.0.post5` | 0.2 |

This is only an issue for unoptimised builds - the huge discrepancy disappears from `-O1` onwards.

garth-wells:
## Summary

PR #4049 rewrote the cell loops in `fem::interpolate` (and `Function::eval`) to
iterate generically over any `mesh::CellRange`, computing a per-cell offset via
`std::size_t offset = std::distance(cells.begin(), cell_it);`.

`Function::interpolate()`'s whole-domain overload passes
`std::ranges::iota_view(0, num_cells)` as the cell range. `iota_view`'s
iterator satisfies the C++20 `std::random_access_iterator` *concept*, but not
the legacy `LegacyRandomAccessIterator` named requirement, because
`operator*` returns a prvalue rather than a genuine reference. As a result
`std::iterator_traits<It>::iterator_category` resolves to `input_iterator_tag`,
and `std::distance` (which dispatches on that legacy tag, not the C++20
concept) silently falls back to its O(n) increment-and-count implementation
instead of an O(1) subtraction.

Since this call sits inside the per-cell loop, the whole interpolation
becomes **O(num_cells²)** instead of O(num_cells) in unoptimised (Debug)
builds — exactly the regression reported in #4347 (52s vs 0.2s for a
constant-field interpolation over ~162k cells). It disappears from `-O1`
onward because the compiler can inline through the loop and collapse the
trivial counting loop via strength reduction.

`std::ranges::distance` dispatches on the C++20 iterator concept rather than
the legacy category tag, and stays O(1) for `iota_view` and friends, so this
swaps all affected call sites to it.

Locally reproducing the issue's demo confirms the fix: n=30 (162,000 cells)
drops from 52s to 0.25s, matching the pre-regression `v0.10.0.post5` baseline
of ~0.2s, with clean linear scaling at smaller n.

Fixes #4347.

## Changes

- `cpp/dolfinx/fem/interpolate.h`: 5 call sites in `interpolation_coords`,
  `impl::point_evaluation` (x2), `impl::identity_mapped_evaluation`,
  `impl::piola_mapped_evaluation`.
- `cpp/dolfinx/fem/Function.h`: 2 call sites in `Function::eval`, which takes
  the same generic `mesh::CellRange auto&&` parameter and is public API, so
  it's exposed to the same pitfall if ever called with a view-typed range.

## Test plan

- [x] `clang-format --dry-run --Werror` clean on both files
- [x] `ninja` rebuild of `libdolfinx` (Developer/Debug build type) clean under `-Werror`
- [x] Full C++ Catch2 suite: 68 test cases / 27,210 assertions passing
- [x] Local reproducer from #4347 confirms O(n²) → O(n) and 52s → 0.25s at n=30

### DIFF
--- a/cpp/dolfinx/fem/Function.h
+++ b/cpp/dolfinx/fem/Function.h
@@ -588,7 +588,7 @@ public:
           coord_dofs(i, j) = x_g[pos + j];
       }
 
-      std::size_t p = std::distance(cells.begin(), cell_it);
+      std::size_t p = std::ranges::distance(cells.begin(), cell_it);
       for (std::size_t j = 0; j < gdim; ++j)
         xp(0, j) = x[p * xshape[1] + j];
 
@@ -674,7 +674,7 @@ public:
 
       // Permute the reference basis function values to account for the
       // cell's orientation
-      std::size_t p = std::distance(cells.begin(), cell_it);
+      std::size_t p = std::ranges::distance(cells.begin(), cell_it);
       apply_dof_transformation(
           std::span(basis_derivatives_reference_values_b.data()
                         + p * num_basis_values,
--- a/cpp/dolfinx/fem/interpolate.h
+++ b/cpp/dolfinx/fem/interpolate.h
@@ -92,7 +92,7 @@ std::vector<T> interpolation_coords(const fem::FiniteElement<T>& element,
     }
 
     // Push forward coordinates (X -> x)
-    std::size_t offset = std::distance(cells.begin(), cell_it);
+    std::size_t offset = std::ranges::distance(cells.begin(), cell_it);
     for (std::size_t p = 0; p < Xshape[0]; ++p)
     {
       for (std::size_t j = 0; j < gdim; ++j)
@@ -753,7 +753,7 @@ void point_evaluation(const FiniteElement<U>& element, bool symmetric,
       std::size_t row = 0;
       std::size_t rowstart = 0;
       std::span<const std::int32_t> dofs = dofmap.cell_dofs(*cell_it);
-      std::size_t offset = std::distance(cells.begin(), cell_it);
+      std::size_t offset = std::ranges::distance(cells.begin(), cell_it);
       for (int k = 0; k < element_bs; ++k)
       {
         if (k - rowstart > row)
@@ -791,7 +791,7 @@ void point_evaluation(const FiniteElement<U>& element, bool symmetric,
     // Loop over cells
     for (auto cell_it = cells.begin(); cell_it != cells.end(); ++cell_it)
     {
-      std::size_t offset = std::distance(cells.begin(), cell_it);
+      std::size_t offset = std::ranges::distance(cells.begin(), cell_it);
       std::span<const std::int32_t> dofs = dofmap.cell_dofs(*cell_it);
       for (int k = 0; k < element_bs; ++k)
       {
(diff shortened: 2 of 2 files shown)


garth-wells:
Closed automatically by GitHub when the source branch was renamed (`garth/rangess-speed` → `garth/ranges-speed`) to fix a typo — no code changes. Superseded by #4352.

PR #4354, opened 2026-08-03
Fix circular namespaces with `common` and `graph`

garth-wells:
Code in the namespace `common` should not depend on code in `graph`. The relationship should be strictly `graph <- common`.

### DIFF
--- a/python/dolfinx/wrappers/common.cpp
+++ b/python/dolfinx/wrappers/common.cpp
@@ -129,8 +129,14 @@ void common(nb::module_& m)
       .def_prop_ro("num_ghosts", &dolfinx::common::IndexMap::num_ghosts)
       .def_prop_ro("local_range", &dolfinx::common::IndexMap::local_range,
                    "Range of indices owned by this map")
-      .def("index_to_dest_ranks",
-           &dolfinx::common::IndexMap::index_to_dest_ranks)
+      .def(
+          "index_to_dest_ranks",
+          [](const dolfinx::common::IndexMap& self, int tag)
+          {
+            auto [data, offsets] = self.index_to_dest_ranks(tag);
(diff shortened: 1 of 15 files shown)


python/dolfinx/wrappers/common.cpp
```diff
@@ -135,8 +135,15 @@ void common(nb::module_& m)
...
-           &dolfinx::common::IndexMap::index_to_dest_ranks)
+      .def(
+          "index_to_dest_ranks",
+          [](const dolfinx::common::IndexMap& self, int tag)
+          {
+            auto [data, offsets] = self.index_to_dest_ranks(tag);
+            return std::pair{dolfinx_wrappers::as_nbarray(std::move(data)),
+                             dolfinx_wrappers::as_nbarray(std::move(offsets))};
+          },
+          nb::arg("tag"))
```
jorgensd, line 146: Could the python wrapper wrap this as a `dolfinx::graph::AdjacencyList` to avoid API breakage?
If not, i guess it is time to make `dolfinx.common.IndexMap` that overloads the C++ one that could do this conversion (once) as a cached property.

garth-wells: It could return an `AdjacencyList`, but that brings in a circular dependency in the binding later.

The C++ function has comment that we should consider removing `index_to_dest_ranks` and the function is low level, so I'm not bothered by it returning just he data and offset arrays.

jorgensd: Removing it is something I'm strongly against, as I rely on it in several libraries:
- io4dolfinx/adios4dolfinx
- dolfinx_mpc

I can live with the changed signature.

garth-wells: There has been a comment on the doc string that we consider removing it for a long time.

PR #4360, opened 2026-08-04
Improve type-hint of local_range so that it is `tuple[int, int]` rather than `list[int...]`

jorgensd:


python/dolfinx/wrappers/common.cpp
```diff
@@ -127,8 +127,14 @@ void common(nb::module_& m)
...
-      .def_prop_ro("local_range", &dolfinx::common::IndexMap::local_range,
-                   "Range of indices owned by this map")
+      .def_prop_ro(
+          "local_range",
+          [](const dolfinx::common::IndexMap& self)
+          {
+            std::array<std::int64_t, 2> range = self.local_range();
+            return std::make_pair(range[0], range[1]);
+          },
+          "Range of indices owned by this map")
```
schnellerhase, line 137:
```suggestion
      .def_prop_ro("local_range", &dolfinx::common::IndexMap::local_range,
                   "Range of indices owned by this map",
                   nb::sig("def local_range(self) -> [int, int]"))
```

jorgensd: Not sure I'm convinced letting `nanobind` do the conversion is a better way to go than being explicit..
At least, shouldn't it be:
```
 nb::sig("def local_range(self) -> tuple[int, int]"
```
what about:
```
.def_prop_ro(
    "local_range",
    [](const dolfinx::common::IndexMap& self) {
        auto range = self.local_range();
        return nb::make_tuple(range[0], range[1]); 
    },
    "Range of indices owned by this map"
)
```
which I guess is a compromise.

schnellerhase: Fair, I missed the exposed Python type is altered not just the signature.

schnellerhase (approved):

PR #4361, opened 2026-08-04
Add bcs to P11 nested preconditioner of `demo_stokes.py` to make it a more extendible example

### LINKED ISSUE #4246: Inconsistency of different solvers in demos under certain conditions
### Intro
When adapting the `demo_stokes.py` demo to a mixed-formulation Poisson solver as a check, I noticed that one of the iterative solver schemes yields results that are significantly different from the other solution methods. The fix in the demo is easy enough, just passing the BCs to a preconditioner, but I don't know whether that's the end of the story, or it is indicative of something else.

### Steps to reproduce
Unfortunately I don't have a MWE , since I would like to change the original as little as possible. So, there's a short list of steps.

1. Download latest `demo_stokes.py` from [https://github.com/FEniCS/dolfinx/blob/main/python/demo/demo_stokes.py](url) .
2. Apply the attached patch `change2poisson.patch` to convert the Stokes problem into a heat conduction (Poisson) problem fixing the temperature at the left (hot) and right (cold) edges of the unit square, with no explicit BCs applied to the sidewalls (y=0 and y=1).
3. Execute the resulting file. One should get something like (note how the 2nd (A) is inconsistent with all the other results) :
```
(A) Norm of velocity coefficient vector (high-level nested, iterative): 129.99999438549634
(A) Norm of pressure coefficient vector (high-level nested, iterative): 9900.019439007961
(A) Norm of velocity coefficient vector (low-level nested, iterative): 129.99524174925395
(A) Norm of pressure coefficient vector (low-level nested, iterative): 9900.019843920487
(B) Norm of velocity coefficient vector (blocked, iterative): 129.99999519209598
(B) Norm of pressure coefficient vector (blocked, iterative): 9900.01944457705
(C) Norm of velocity coefficient vector (blocked, direct): 129.9999999995917
(C) Norm of pressure coefficient vector (blocked, direct): 9900.019479146438
(D) Norm of velocity coefficient vector (monolithic, direct): 130.00000000170624
(D) Norm of pressure coefficient vector (monolithic, direct): 9900.019479151191
```
4. Change line 325 of the patched file , from 
    `P11 = assemble_matrix(a_p11, [])`
to
    `P11 = assemble_matrix(a_p11, bcs=bcs)`
5. Execute again. Now the two (A)'s should be identical, and consistent with the results from the other methods; most notably with (B), since it's the other iterative method.

### Comments & Questions
- The regular `demo_stokes.py` case does not seem to be affected by this issue
- Is this a lucky fix, and the root cause of the inconsistency lies somewhere else?
- If it is indeed determined that, when building the preconditioner, the BCs must be taken into account as in step 5 above, could it be added to the `demo_stokes.py` demo and anywhere else where it might make sense?
- And related to above, could a short explanation be added to the documentation as to why this is needed?

Thanks

[change2poisson.patch](https://github.com/user-attachments/files/29241229/change2poisson.patch)

jorgensd:
As seen in: #4246 the fact that we didn't pass in the bcs (which in turn is ignored internally if the function spaces of the bc doesn't align with the form spaces) confuses users, leading to the wrong behaviour in demos extending the functionality presented in `demo_stokes.py`.

This PR adds the simple patch of passing in the `bcs`, even if they have no effect on the P11 block.

### DIFF
--- a/python/demo/demo_stokes.py
+++ b/python/demo/demo_stokes.py
@@ -301,7 +301,7 @@ def nested_iterative_solver_low_level():
     # Create a nested matrix P to use as the preconditioner. The
     # top-left block of P is shared with the top-left block of A. The
     # bottom-right diagonal entry is assembled from the form a_p11:
-    P11 = assemble_matrix(a_p11, [])
+    P11 = assemble_matrix(a_p11, bcs=bcs)
     P = PETSc.Mat().createNest([[A.getNestSubMatrix(0, 0), None], [None, P11]])
     P.assemble()
 
--- a/python/demo/demo_stokes.py
+++ b/python/demo/demo_stokes.py
@@ -301,6 +301,8 @@ def nested_iterative_solver_low_level():
     # Create a nested matrix P to use as the preconditioner. The
     # top-left block of P is shared with the top-left block of A. The
     # bottom-right diagonal entry is assembled from the form a_p11:
+    # Even if the Dirichlet conditions are only enforced on the velocity
+    # space, we pass it to the pressure assembler for consistency.
     P11 = assemble_matrix(a_p11, bcs=bcs)
     P = PETSc.Mat().createNest([[A.getNestSubMatrix(0, 0), None], [None, P11]])
     P.assemble()


nate-sime (approved):

jhale:
Looks good, but probably worth a quick comment that this is a nullop, and why.

jhale (commented):
As in main thread

python/demo/demo_stokes.py
```diff
@@ -301,7 +301,7 @@ def nested_iterative_solver_low_level():
     # Create a nested matrix P to use as the preconditioner. The
     # top-left block of P is shared with the top-left block of A. The
     # bottom-right diagonal entry is assembled from the form a_p11:
-    P11 = assemble_matrix(a_p11, [])
+    P11 = assemble_matrix(a_p11, bcs=bcs)
```
jorgensd, line 306:
```suggestion
    # Even if the Dirichlet conditions are only enforced on the velocity
    # space, we pass it to the pressure assembler for consistency.
    P11 = assemble_matrix(a_p11, bcs=bcs)
```

PR #4365, opened 2026-08-04
Modular restructuring of pml demo

HawraaAl:
Created a modular version of the `demo_pml.py` 

The goal :
- Helps separate specific parts of the problem to make it easier to navigate
- Introduces the idea to of using modules to solve complicated problems similar to the pml demo

The modules are as follows :

- `main.py`: main execution script
- `config.py`: parameters and physical tags
- `mesh_wire.py`: Gmsh geometry and DOLFINx mesh conversion
- `fields.py`: analytical field formulas and `curl_2d`
- `pml.py`: PML coordinate stretching and transformed tensors
- `solve.py`: function spaces, permittivity, weak form, and solver
- `visualize.py`: VTX writing and optional PyVista plotting
- `postprocess.py`: absorption/scattering/extinction efficiencies


HawraaAl:
@jorgensd

jorgensd (commented):
My personal opinion is that this is too many files for the demo.
By adding new stand alone functions for all these options, one should have full documentation of each function (Input arguments, output arguments, functionality),
as done in the main documentation. Furthermore, I wonder how we would render this nicely as an online documentation.

We have discussed modularizing demos for a while, so what is your opinion @jhale, @chrisrichardson, @garth-wells, @cdaversin, @francesco-ballarin, @nate-sime

nate-sime:
I quite like this, but perhaps I'd call it `demo_pml-project` or something to signify we have project-scoped demos. Unfortunately I doubt there will be consensus; a verdict on this PR will likely arrive from subjective opinion.

cdaversin:
I agree with @nate-sime's point on differentiating this kind of demo (project-scoped) from the simpler/shorter ones — maybe in a dedicated directory? Splitting such a demo into several files makes sense to me. Though I agree with @jorgensd on the documentation, considering the number of functions to document.

An option would be to standardize these project-scoped demos, with a standard list of files (for example: main.py, config.py, mesh.py, solve.py, postprocessing.py), each with a few required functions having standardized names and signatures to ease the documentation. These required functions could obviously involve additional functions to be documented.

On the other hand, such standardization would be worth it only if we have enough demos of this type. It also entails a consensus on which files and which functions are required, with the risk of this being a never-ending discussion.

PR #4369, opened 2026-08-06
Change args of `create_topology` to NDArray and introduce `dolfinx_wrappers::vec_of_spans`

schnellerhase:
Caught by `mypy` check on downstream usage of `create_topology`, while possible due to implicit conversions, we should stick to the `NDArray` in the interface.

Previously this implied the (non numpy compatible signature):
```python
def create_topology(arg0: mpi4py.MPI.Comm, arg1: Sequence[CellType], arg2: Sequence[Sequence[int]], arg3: Sequence[Sequence[int]], arg4: Sequence[Sequence[int]], arg5: Sequence[int], arg6: int, /) -> Topology:
    """Create a Topology object."""
```
now
```python
def create_topology(arg0: mpi4py.MPI.Comm, arg1: Sequence[CellType], arg2: Sequence[Annotated[NDArray[numpy.int64], dict(shape=(None,), order='C', writable=False)]], arg3: Sequence[Annotated[NDArray[numpy.int64], dict(shape=(None,), order='C', writable=False)]], arg4: Sequence[Annotated[NDArray[numpy.int32], dict(shape=(None,), order='C', writable=False)]], arg5: Annotated[NDArray[numpy.int64], dict(shape=(None,), order='C', writable=False)], arg6: int, /) -> Topology:
    """Create a Topology object."""
```

- Introduced common wrapper functionality `dolfinx_wrappers::vec_of_spans` to facilitate `std::vector<nb::ndarray>` to `std::vector<std::span>` conversions.

### DIFF
--- a/python/dolfinx/wrappers/mesh.cpp
+++ b/python/dolfinx/wrappers/mesh.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2017-2025 Chris N. Richardson, Garth N. Wells, Jørgen S. Dokken
+// Copyright (C) 2017-2026 Chris N. Richardson, Garth N. Wells, Jørgen S. Dokken
 // and Paul T. Kühner
 //
 // This file is part of DOLFINx (https://www.fenicsproject.org)
--- a/python/dolfinx/wrappers/mesh.cpp
+++ b/python/dolfinx/wrappers/mesh.cpp
@@ -333,7 +333,6 @@ void mesh(nb::module_& m)
              ghost_owners,
          nb::ndarray<const std::int64_t, nb::ndim<1>, nb::c_contig>
              boundary_vertices,
-
          int num_threads)
       {
         std::vector<std::span<const std::int64_t>> cells_span
--- a/python/dolfinx/wrappers/mesh.cpp
+++ b/python/dolfinx/wrappers/mesh.cpp
@@ -340,25 +340,12 @@ void mesh(nb::module_& m)
 
          int num_threads)
       {
-        auto to_vec_of_spans
-            = []<typename T>(
-                  const std::vector<nb::ndarray<T, nb::ndim<1>, nb::c_contig>>&
-                      vec)
-        {
-          std::vector<std::span<T>> vec_span;
-          vec_span.reserve(vec.size());
-          std::ranges::transform(
-              vec, std::back_inserter(vec_span),
-              [](auto& nd) { return std::span<T>(nd.data(), nd.size()); });
-          return vec_span;
-        };
-
         std::vector<std::span<const std::int64_t>> cells_span
-            = to_vec_of_spans(cells);
+            = vec_of_spans(cells);
         std::vector<std::span<const std::int64_t>> original_cell_index_span
-            = to_vec_of_spans(original_cell_index);
+            = vec_of_spans(original_cell_index);
         std::vector<std::span<const int>> ghost_owners_span
-            = to_vec_of_spans(ghost_owners);
+            = vec_of_spans(ghost_owners);
 
         std::span<const std::int64_t> boundary_vertices_span(
             boundary_vertices.data(), boundary_vertices.size());
--- a/python/dolfinx/wrappers/mesh.cpp
+++ b/python/dolfinx/wrappers/mesh.cpp
@@ -128,12 +128,8 @@ void mesh(nb::module_& m)
              nb::ndarray<const std::int64_t, nb::ndim<1>, nb::c_contig>>& cells,
          std::optional<std::int32_t> max_facet_to_cell_links, int num_threads)
       {
-        std::vector<std::span<const std::int64_t>> cell_span(cells.size());
-        for (std::size_t i = 0; i < cells.size(); ++i)
-        {
-          cell_span[i]
-              = std::span<const std::int64_t>(cells[i].data(), cells[i].size());
-        }
+        std::vector<std::span<const std::int64_t>> cell_span
+            = vec_of_spans(cells);
         return dolfinx::mesh::build_dual_graph(
             comm.get(), cell_types, cell_span, max_facet_to_cell_links,
             num_threads);
--- a/python/dolfinx/wrappers/mesh.cpp
+++ b/python/dolfinx/wrappers/mesh.cpp
@@ -366,6 +366,10 @@ void mesh(nb::module_& m)
             comm.get(), cell_type, cells_span, original_cell_index_span,
             ghost_owners_span, boundary_vertices_span, num_threads);
       },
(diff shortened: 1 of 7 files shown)


python/dolfinx/wrappers/mesh.cpp
```diff
@@ -328,19 +329,39 @@ void mesh(nb::module_& m)
...
+                  const std::vector<nb::ndarray<T, nb::ndim<1>, nb::c_contig>>&
+                      vec)
+        {
+          std::vector<std::span<T>> vec_span;
+          vec_span.reserve(vec.size());
+          std::ranges::transform(
+              vec, std::back_inserter(vec_span),
+              [](auto& nd) { return std::span<T>(nd.data(), nd.size()); });
+          return vec_span;
+        };
```
schnellerhase: Shall this be moved to a shared implementation in `array.h`? Other exports (mostly mixed topology related) have similar code paths.

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -1,5 +1,5 @@
-// Copyright (C) 2017-2025 Chris N. Richardson, Garth N. Wells and Jørgen S.
-// Dokken
+// Copyright (C) 2017-2025 Chris N. Richardson, Garth N. Wells, Jørgen S. Dokken
```
jorgensd:
```suggestion
// Copyright (C) 2017-2026 Chris N. Richardson, Garth N. Wells, Jørgen S. Dokken
```

python/dolfinx/wrappers/mesh.cpp
jorgensd: It is also in a few other wrapper places:
https://github.com/search?q=repo%3AFEniCS%2Fdolfinx+std%3A%3Avector%3Cstd%3A%3Aspan+path%3A%2F%5Epython%5C%2Fdolfinx%5C%2Fwrappers%5C%2F%2F&type=code

python/test/unit/mesh/test_mixed_topology.py
```diff
@@ -32,10 +32,10 @@
...
     set_log_level(LogLevel.INFO)
 
-    cells = [[0, 1, 2, 1, 2, 3], [2, 3, 4, 5]]
-    orig_index = [[0, 1], [2]]
-    ghost_owners = [[], []]
-    boundary_vertices = []
+    cells = [np.array([0, 1, 2, 1, 2, 3]), np.array([2, 3, 4, 5])]
+    orig_index = [np.array([0, 1]), np.array([2])]
+    ghost_owners = [np.array([]), np.array([])]
+    boundary_vertices = np.array([])
```
jorgensd: Set dtype argumenens for teh various errors here to ensure correctness?

python/test/unit/mesh/test_mixed_topology.py
```diff
@@ -98,10 +98,14 @@ def test_mixed_topology_mesh(dtype):
...
-    ghost_owners = [[], [], []]
-    boundary_vertices = []
+    cells = [
+        np.array([0, 1, 2, 3, 1, 2, 3, 4]),
+        np.array([2, 3, 4, 5, 6, 7]),
+        np.array([3, 4, 6, 7, 8, 9, 10, 11]),
+    ]
+    orig_index = [np.array([0, 1]), np.array([2]), np.array([3])]
+    ghost_owners = [np.array([]), np.array([]), np.array([])]
+    boundary_vertices = np.array([])
```
jorgensd: Same here?

python/test/unit/mesh/test_mixed_topology.py
```diff
@@ -211,12 +215,12 @@ def test_parallel_mixed_mesh(dtype):
...
-    cells = [[t + 3 * rank for t in tri], [q + 3 * rank for q in quad]]
-    orig_index = [[3 * rank, 1 + 3 * rank], [2 + 3 * rank]]
+    cells = [np.array([t + 3 * rank for t in tri]), np.array([q + 3 * rank for q in quad])]
+    orig_index = [np.array([3 * rank, 1 + 3 * rank]), np.array([2 + 3 * rank])]
     # No ghosting
-    ghost_owners = [[], []]
+    ghost_owners = [np.array([]), np.array([])]
     # All vertices are on boundary
-    boundary_vertices = [3 * rank + i for i in range(6)]
+    boundary_vertices = np.array([3 * rank + i for i in range(6)])
```
jorgensd: Same here.

python/dolfinx/wrappers/mesh.cpp
garth-wells: This code looks weird and over-complicated. What does nanobind even do with spans?

schnellerhase: In our C++ interfaces we rely on spans; so any input from the wrapper layer coming in form of `ndarray`s needs to be converted at some point. For (Python) lists of arrays (as here), these turn into `std::vector<ndarray>` on C++ side and need to be converted to `std::vector<std::span>` before calling the C++ library. This function facilitates this transform. This could be generalized further (in particular in shape) and moved into `array.h` since this is a common wrapper pattern.

garth-wells:
```c++
auto to_vec_of_spans = [](const auto& vec)
{
  using X = std::remove_cvref_t<decltype(vec)>::value_type::Scalar;
  std::vector<std::span<X>> vec_span;
  for (auto& x : vec)
    vec_span.emplace_back(x.data(), x.size());
  return vec_span;
};
```

schnellerhase: Moved to `array.h`, replaced implementation to be based on loop + `emplace_back`, other use cases adapted.

python/dolfinx/wrappers/mesh.cpp
```diff
@@ -328,23 +325,34 @@ void mesh(nb::module_& m)
...
-         const std::vector<std::int64_t>& boundary_vertices, int num_threads)
+         std::vector<nb::ndarray<const std::int64_t, nb::ndim<1>, nb::c_contig>>
+             cells,
+         std::vector<nb::ndarray<const std::int64_t, nb::ndim<1>, nb::c_contig>>
+             original_cell_index,
+         std::vector<nb::ndarray<const int, nb::ndim<1>, nb::c_contig>>
+             ghost_owners,
+         nb::ndarray<const std::int64_t, nb::ndim<1>, nb::c_contig>
+             boundary_vertices,
+
```
garth-wells: Stray line?

schnellerhase: Suprising that this passes `clang-format` - removed.

garth-wells (approved):

PR #4375, opened 2026-08-08
Stricter mypy type checking

### LINKED ISSUE #2297: Progressively turn on stricter mypy checking

garth-wells:
Resolves #2297.


schnellerhase:
Can we get https://github.com/FEniCS/dolfinx/pull/4364 in before increasing strictness (possibly extending it also to the demos)? That should increase coverage - we have still not all configs type checked in the post installation mode, so increasing strictness will make accomplishing this even harder.

PR #4383, opened 2026-08-09
Add type hints to all Python functions

garth-wells:
## Summary

Enables `disallow_untyped_defs = true` for the `dolfinx` Python library in `python/pyproject.toml`, and adds the missing type annotations needed to satisfy it (~120 previously-untyped functions/methods).

Test and demo code are exempted via explicit mypy overrides rather than being annotated:
- `conftest`, `unit.*`, and demo's `test` module — test bodies are unannotated pytest functions by convention (see AGENTS.md).
- Each `demo_*` module — demos are intentionally lightly typed. (mypy's per-module glob only matches whole dotted components, so `demo_*` can't be expressed as a single wildcard pattern and each demo module is listed explicitly.)

## Notes

- Annotations follow existing conventions in the codebase: PEP 604 unions, `npt.NDArray[...]` for array-likes, cpp binding return types (via the nanobind stubs) for thin wrapper methods, and `# type: ignore[arg-type]`/`[return-value]` at the small number of sites with known mpi4py/petsc4py/cpp-union type gaps that already appear elsewhere in the library.
- Correctly typing library internals surfaced a couple of pre-existing, previously-hidden issues in demos, fixed here:
  - `demo_pml.py` compared an `MPI.Comm` to an `int` instead of `.comm.size`.
  - `demo_navier-stokes.py` accumulated a `float` with a `Constant.value` (which returns an `ndarray`).
- A handful of other demos pass an `mpi4py.MPI.Comm` directly to PETSc APIs typed as expecting `petsc4py.PETSc.Comm`; suppressed with `# type: ignore[arg-type]`, matching the existing pattern used elsewhere in the library for this same mpi4py/petsc4py stub mismatch.

## Test plan

- `mypy --config-file python/pyproject.toml -p dolfinx`: 0 `no-untyped-def` errors (only pre-existing `no-any-unimported` errors remain, from untyped `ufl`/`ffcx`/`gmsh`).
- `mypy demo` / `mypy test`: error output unchanged from the pre-change baseline.
- `ruff check` / `ruff format --check`: clean on all touched files.
- Full Python test suite run locally: passes (one pre-existing, unrelated `distutils` failure in `test_cffi_assembly` reproduces identically on `main`).

## AI assistance

I used Claude Code to make the changes in this PR (adding the type annotations and the `pyproject.toml` mypy configuration). I reviewed, tested, and take responsibility for the final contribution.


python/dolfinx/common.py
```diff
@@ -131,12 +135,12 @@ def __init__(self, name: str | None = None):
         """
         self._cpp_object = _cpp.common.Timer(name)
 
-    def __enter__(self):
+    def __enter__(self) -> "Timer":
```
jorgensd: Should we use typing.Self?

python/dolfinx/fem/dofmap.py
```diff
@@ -44,27 +48,27 @@ def cell_dofs(self, cell_index: int):
...
 
     @property
-    def bs(self):
+    def bs(self) -> int:
         """Block size of the dofmap."""
         return self._cpp_object.bs
 
     @property
-    def dof_layout(self):
+    def dof_layout(self) -> _cpp.fem.ElementDofLayout:
```
jorgensd, line 56: Should we make a Python wrapper for element dof layout?

python/dolfinx/fem/dofmap.py
```diff
@@ -44,27 +48,27 @@ def cell_dofs(self, cell_index: int):
...
         """Layout of dofs on an element."""
         return self._cpp_object.dof_layout
 
     @property
-    def index_map(self):
+    def index_map(self) -> _cpp.common.IndexMap:
```
jorgensd, line 61: Should we have a Python wrapper for indexmap rather than returning the C++ binding?

python/dolfinx/io/utils.py
```diff
@@ -97,19 +98,24 @@ def __init__(
                 )
                 self._cpp_object = _vtxwriter(comm, filename, cpp_objects, engine, mesh_policy)  # type: ignore[arg-type]
 
-        def __enter__(self):
+        def __enter__(self) -> "VTXWriter":
```
jorgensd: Typing self?

python/dolfinx/io/utils.py
```diff
@@ -134,15 +140,20 @@ def __init__(self, comm: _MPI.Comm, filename: str | Path, mode: str):
         """
         self._cpp_object = _cpp.io.VTKFile(comm, filename, mode)
 
-    def __enter__(self):
+    def __enter__(self) -> "VTKFile":
```
jorgensd: Typing self?

python/dolfinx/io/utils.py
```diff
@@ -181,15 +192,20 @@ def __init__(
         """
         self._cpp_object = _cpp.io.XDMFFile(comm, filename, file_mode, encoding)
 
-    def __enter__(self):
+    def __enter__(self) -> "XDMFFile":
```
jorgensd: Typing self?

python/dolfinx/mesh.py
```diff
@@ -1047,8 +1051,8 @@ def create_interval(
     nx: int,
     points: npt.ArrayLike,
     dtype: npt.DTypeLike = default_real_type,
-    ghost_mode=GhostMode.shared_facet,
-    partitioner=None,
+    ghost_mode: GhostMode = GhostMode.shared_facet,
+    partitioner: Callable | None = None,
```
jorgensd, line 1055: Should we have a more detailed type hint than callable here?

python/dolfinx/mesh.py
```diff
@@ -1096,8 +1100,8 @@ def create_unit_interval(
     comm: _MPI.Comm,
     nx: int,
     dtype: npt.DTypeLike = default_real_type,
-    ghost_mode=GhostMode.shared_facet,
-    partitioner=None,
+    ghost_mode: GhostMode = GhostMode.shared_facet,
+    partitioner: Callable | None = None,
```
jorgensd, line 1104: Same here

jorgensd (changes requested):

python/dolfinx/common.py
garth-wells: In which version of Python did typing.Self become available?

python/dolfinx/fem/dofmap.py
garth-wells: Yes, and some other classes.

python/dolfinx/fem/dofmap.py
garth-wells: Yes

python/dolfinx/common.py
jorgensd: 3.11, which fits with our current lower bound: https://docs.python.org/3/library/typing.html#typing.Self

python/dolfinx/fem/element.py
```diff
@@ -220,9 +220,10 @@ def __init__(
         """
         self._cpp_object = cpp_object
 
-    def __eq__(self, other):
+    def __eq__(self, other: object) -> bool:
         """Check equality with another finite element."""
-        return self._cpp_object == other._cpp_object
+        _other = cast("FiniteElement", other)
+        return self._cpp_object == _other._cpp_object
```
schnellerhase, line 226:
```suggestion
        if not isinstance(other, FiniteElement):
            return NotImplemented
        return self._cpp_object == other._cpp_object
```

Same pattern for the other eq checks.

jhale (commented):
Aside from the valid and easy points from the others on e.g. `typing.Self`, I would get this in quickly and then remove e.g. `_cpp` classes from the public API and tighter `Callable` definitions in some other PRs.

garth-wells:
`typing.Self` is done, will merge shortly.

jorgensd (approved):

PR #4387, opened 2026-08-09
Fix script-injection risk and use HTTPS in CI workflows

garth-wells:
## Summary
- `docker-update-stable.yml`: pass `workflow_dispatch` inputs (`end-user-tag`, `dev-env-tag`) via `env:` and reference them as quoted shell variables instead of interpolating `${{ inputs.* }}` directly into `run:` blocks, closing a script-injection vector (SonarCloud `S7630`).
- `build-wheels.yml`: switch the `simple503 --base-url` values for `packages.fenicsproject.org` from `http://` to `https://` (SonarCloud `S5332`).

Addresses a subset of the open SonarCloud code-scanning alerts on this repo; the remaining alerts (pip `--no-build-isolation` / dependency-locking warnings across CI/Docker/`pyproject.toml`) are intentional given DOLFINx's need to build against system `petsc4py`/`mpi4py` and were left as-is.

## Test plan
- [ ] CI runs green on this branch (workflow YAML changes only, no code paths affected)
- [ ] Manually trigger `docker-update-stable.yml` via `workflow_dispatch` with a normal tag to confirm `regctl image copy` still works with the env-var substitution
- [ ] Confirm `packages.fenicsproject.org` serves HTTPS before merging (not verifiable from this environment)

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -336,7 +336,7 @@ jobs:
 
       - name: Update FEniCS Project simple503 repository
         run: |
-          python -m simple503 --base-url https://packages.fenicsproject.org/simple simple
+          python -m simple503 --base-url http://packages.fenicsproject.org/simple simple
 
       - name: Upload simple503 artifact
         uses: actions/upload-artifact@v7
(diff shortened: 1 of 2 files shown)


.github/workflows/build-wheels.yml
jhale, line 1: Slight misinterpretation from the tool here. The URL is an output related to where the wheels will be downloaded from, not an input.

This change won't work - I use http://pub.fenicsproject.org rather than because https://pub.fenicsproject.org gives a certificate error as we manage our certificates using GitHub Pages.

jhale (dismissed):
Revert [.github/workflows/build-wheels.yml‎](https://github.com/FEniCS/dolfinx/pull/4387/changes#diff-14b51e9e6321caf5d23b2253f2415daf475fdaa417c0aa8fe896c3a460f2d023).

Issue #4273, opened 2026-07-07
[BUG]: cffi 2.1.0 has complex-related type bug


jhale:
### Summarize the issue

Triggers some tests related to quadrature with complex types.

### How to reproduce the bug

Run tests with cffi 2.1.0.

### Minimal Example (Python)

```Python

```

### Output (Python)

```bash

```

### Version

main branch

francesco-ballarin:
The upstream issue is still present in `cffi` 2.1.1, released Aug 3, 2026.

See for instance [this CI run](https://github.com/FEniCS/dolfinx/actions/runs/31266696985/job/93125935845?pr=4287) within #4287

garth-wells:
Has anyone reported upstream?

jhale:
This turns out to be a numba issue, in that Numba cannot type CFFI complex arguments `_cffi_{float,double}_complex_t` which have been present since cffi 1.17.

I've put a PR in #4397 but the real fix will have to come upstream in a PR to Numba.

garth-wells:
See https://github.com/numba/numba/pull/10770 for Numba issue.

PR #4397, opened 2026-08-10
Unpin cffi and fix stale complex skip condition

jhale:
Related to https://github.com/FEniCS/dolfinx/issues/4273.

This is not a fix, but does apply and give a reason for the skip. The fix will need to be done in Numba, rather than cffi.

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -28,7 +28,7 @@ authors = [
 ]
 dependencies = [
       "numpy>=2",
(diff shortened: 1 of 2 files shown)


jhale:
I have submitted a PR to Numba https://github.com/numba/numba/pull/10770, obviously it will be quite some time until we see that in a built version, so I propose we proceed with this workaround for the failing test.

garth-wells (approved):

Issue #4402, opened 2026-08-11
Mixed-topology meshes redistribute cells with one NBX consensus round per cell type


garth-wells:
`mesh::impl::partition_cells` (in `cpp/dolfinx/mesh/utils.h`) redistributes cells by calling `graph::build::distribute` once per cell type:

```cpp
for (std::int32_t i = 0; i < num_cell_types; ++i)
{
  ...
  std::tie(cells1[i], src_ranks, original_idx1[i], ghost_owners[i])
      = graph::build::distribute(comm, cells[i], {num_cells, num_cell_nodes}, dest_i);
}
```

Each call to `graph::build::distribute` independently discovers its neighbourhood communicator using the NBX consensus algorithm (`MPI::compute_graph_edges_nbx`, see `cpp/dolfinx/graph/partition.cpp`). This is scalable (no arrays the size of the communicator are built, and the communication pattern stays sparse), but it is not free: each call costs at least one non-blocking consensus round.

For the common case of a single-cell-type mesh this is a non-issue (one cell type, one call, one round). For a genuinely mixed-topology mesh (e.g. tetrahedra + hexahedra), each cell type currently pays for its own independent consensus round, even though the destination rank sets for different cell types typically overlap substantially, since they come from the same overall partitioning decision over the same physical domain.

Possible directions:
- Discover the neighbourhood once (e.g. from the union of destination ranks across all cell types, or from one cell type as an over-approximation of the others) and reuse it for the remaining cell types' exchanges.
- Restructure so a single interleaved payload spanning all cell types is redistributed in one `distribute` call.

This only matters for mixed-topology meshes, and only at large enough rank counts that the NBX consensus-round cost is non-negligible relative to the actual data exchange. For the common single-cell-type case there is nothing to fix.

PR #4403, opened 2026-08-11
Add space filling curve (SFC) and hybrid SFC/graph partitioning algorithms for building parallel meshes 

garth-wells:
## Summary

Adds space-filling curve (SFC) and hybrid SFC/graph partitioning, so that mesh construction cost can stay roughly flat as the rank count grows, instead of scaling with the (increasingly expensive) cost of graph partitioning. Also includes a general mesh-construction performance pass, a ParMETIS re-partitioner for rebalancing, and `demo_partition.py`, a new demo comparing all available partitioners on the same input under varying degrees of input-distribution randomness.

## What's new

*(This section was substantially reworked on 2026-08-28 to track a round of internal simplification/renaming on the branch since the previous write-up — see the note after each changed bullet.)*

**Geometric and hybrid partitioning types** (`graph/partition.h`)
- `graph::partition_fn` — graph-only partitioners (required dual graph, no coordinates).
- `graph::geom_partition_fn` — partitioners that use node positions only, with no access to the graph at all (so they can never ghost).
- `graph::hybrid_partition_fn` — partitioners that need *both* the graph and node positions unconditionally, because the graph edges are part of the partitioning decision itself, not only used for ghosting.
- All three now take optional node weights (`hybrid_partition_fn`/`partition_fn` also take optional edge weights); a `graph::Partitioner` struct bundles one of the three (`AnyPartitionFunction`, renamed from `AnyCellPartitionFunction` and moved from `dolfinx::mesh` to `dolfinx::graph`) together with the node weights it should be called with. `graph::partition_graph` is a new default graph partitioner that dispatches to whichever of ParMETIS/PT-SCOTCH/KaHIP the build has enabled, in that preference order.
  <br>*(Update: `geom_partition_fn` gained node-weight support and `AnyCellPartitionFunction` was renamed/relocated since the original write-up; both are typical of the ongoing internal tidy-up on this branch rather than user-facing behaviour changes.)*

**Space-filling curve partitioning** (`graph/sfc.h{cpp}`)
- `graph::partition_sfc_morton` / `partition_sfc_hilbert`: partition points along a Morton or Hilbert curve, now with optional node weights (partitions aim for equal weight sums along the curve rather than equal counts). No external library required. Cost is nearly independent of rank count, at the price of a larger edge cut than graph partitioning.
- The internal step that routes each point to the rank that owns its curve position was reworked twice since the original write-up: first via a "post-office" neighbourhood exchange instead of an `MPI_Allgatherv`-based scheme, then simplified/further sped up again. See performance data below — this is a real, measurable speed-up, not just a refactor.
  <br>*(Update: the earlier `graph::sfc::partitioner(curve)` factory and its `curve` enum, described in the original write-up, have been removed. `partition_sfc_morton`/`partition_sfc_hilbert` are used directly now; on the Python side they're exposed as ready-to-use `dolfinx.graph.partition_morton` / `partition_hilbert` objects rather than via a curve-selecting factory function.)*

**ParMETIS additions** (`graph/partitioners.h{cpp}`)
- `graph::parmetis::geom_partitioner` (`ParMETIS_V3_PartGeom`) and `geom_partitioner_kway` (`ParMETIS_V3_PartGeomKway`, a `hybrid_partition_fn` since it uses the graph edges as well as the coordinates). These C++ names are unchanged; the Python bindings for them were renamed for consistency with the rest of the module (now `dolfinx.graph.partitioner_parmetis_geom` and `partitioner_parmetis_hybrid`).
- `graph::parmetis::repartitioner` (`ParMETIS_V3_AdaptiveRepart`): treats the current data distribution as the current partition and computes a new one that balances load while limiting how many nodes move — suited to rebalancing after non-uniform refinement, as opposed to partitioning from scratch. Still not exposed to Python; re-measured below via a throwaway C++ driver.
- Default PT-SCOTCH strategy changed to `strategy::speed` (see performance data below).

**Python-layer partitioner wrapping** (was: "Mesh-level wrappers", `mesh/partition.h{cpp}`)
<br>*(Update: `mesh/partition.h{cpp}`, the file this bullet originally pointed at, no longer exists — it was folded into `graph/partition.h` (the `AnyPartitionFunction`/`Partitioner` types above) and `mesh/utils.h` (the cell-centroid computation, still there, just relocated). More importantly, `create_geometric_cell_partitioner`/`create_hybrid_cell_partitioner` are no longer C++ library functions that return a `dolfinx::mesh::` type — they're thin nanobind-layer helpers (`dolfinx.mesh.create_geometric_cell_partitioner`/`create_hybrid_cell_partitioner`, implemented via `_cpp.graph`) that wrap a plain Python callable into the opaque `GeometricPartitioner`/`HybridPartitioner` handle `create_mesh` dispatches on. The behaviour they described is unchanged, just relocated and re-described below.)*
- `dolfinx.mesh.create_geometric_cell_partitioner`: wraps a Python callable taking cell centroids into a `geom_partition_fn`-shaped handle. **The result never ghosts** — it has no access to the mesh dual graph, so a mesh built with it always has `GhostMode.none`, regardless of what `create_mesh`'s `ghost_mode` argument requests. Use `create_hybrid_cell_partitioner` for a partitioner that also needs the dual graph (e.g. for ghosting). Only needed for a *custom* geometric partitioner — the built-ins (`partition_morton`, `partition_hilbert`, `partitioner_parmetis_geom`) already come pre-wrapped.
- `dolfinx.mesh.create_hybrid_cell_partitioner`: as above but for a hybrid partitioner — the dual graph is always computed and supplied, regardless of `ghost_mode`, since these partitioners need it unconditionally. Also only needed for a custom hybrid partitioner; ParMETIS `GeomKway` (`partitioner_parmetis_hybrid`) already comes pre-wrapped.

**`create_mesh` performance and structure**
- Removed redundant copies/sorts in the serial path (re-measured: 3.47 s for 6M tetrahedra, Release build, single rank, against 4.22 s before this change — see below), avoided a second copy of the concatenated cell array, reused the sorted unique vertex indices that topology construction already computes instead of re-deriving them, and skipped the (identity) topology-extraction step for 'P1 geometry'.
- Extracted the cell-partitioning/redistribution logic into `mesh::impl::partition_cells` (still accurate — this function is unaffected by the restructuring above), shrinking `create_mesh` to a clearer sequence of named steps.

**`demo_partition.py`**
- Compares every available partitioner (ParMETIS Kway/GeomKway/Geom, PT-SCOTCH, SFC Morton/Hilbert) on the same distributed input, reporting cell imbalance and edge cut, across input distributions ranging from fully locality-preserving to fully randomised.
- A "two-stage partitioning" comparison: partition with the space-filling curve first (cheap, insensitive to input distribution), then re-partition the result with a graph partitioner (now running on a well-distributed input) — the general idea ParMETIS `GeomKway` embodies, but applicable around any graph partitioner.
- New since the original write-up: a `slab_partitioner` example showing how little code a custom geometric partitioner needs (cut the domain into `nparts` slabs along one axis), demonstrating `create_geometric_cell_partitioner` on a partitioner that isn't one of the built-ins.

## Performance data

All numbers below were **re-measured on 2026-08-28 against the current tip of this branch** (`a5e21446f3`), on the same 20-core / 40-thread Intel Xeon w7-3445 workstation used for both previous write-ups. This branch picked up a further round of SFC-specific performance work since the last re-run (2026-08-27) — most visibly "Simplify and speed up SFC post-office routing" — so several SFC-touching numbers below moved measurably, not just within noise; those are called out explicitly. ParMETIS and PT-SCOTCH figures are single runs (PT-SCOTCH's result varies a little run to run — see the demo's own notes), so treat small differences elsewhere as noise.

- **`create_mesh` copy/sort removal**: 4.22 s → 3.47 s, 6M tetrahedra, serial, Release build (best of 3 runs; unaffected by the SFC work below since this is a purely serial, non-partitioning code path — figure unchanged from the previous re-run).
- **SFC vs. graph partitioning for mesh construction**: now 1.97x faster at 40 ranks for 15.4M tetrahedra (SFC Hilbert 0.91 s vs. ParMETIS Kway 1.78 s, `ghost_mode=none`) — up from 1.48x on the previous re-run (SFC Hilbert dropped from 1.20 s to 0.91 s; `Kway` is unchanged), directly attributable to the SFC routing speed-up. As before, this is a timing-only comparison: `create_geometric_cell_partitioner`-based partitioners (SFC included) never ghost through the public API, so there's no ghost-cell count to compare for pure SFC. For a curve-informed partitioner that *can* ghost, ParMETIS `GeomKway` (hybrid) still gives modestly fewer ghost cells than plain `Kway` at the same scale (629,921 vs. 646,922, about 3% fewer) at close to the same cost now (2.20 s vs. 2.04 s, versus a bigger gap previously); PT-SCOTCH gives the fewest ghost cells (599,829) at roughly 2.3x `Kway`'s cost (4.76 s).
- **PT-SCOTCH `strategy::speed` vs. default**: `speed` measured 4–13% faster than SCOTCH's own default (`strategy::none`) this time (20 ranks/3.1M cells: 2.50 s → 2.22 s, 152.1k → 148.4k cut edges; 40 ranks/6M cells: 5.75 s → 5.58 s, 326.7k → 329.8k cut edges — a smaller gap than the 10–14% seen on the previous re-run, itself already below the original write-up's ~20%; PT-SCOTCH is not affected by the SFC changes, so this drift is run-to-run noise, not a trend). `strategy::quality` again matched `strategy::none`'s cut exactly and was never faster. `speed` remaining the best of the three by both cut and time is the finding that has held up across all three measurements.
- **ParMETIS Kway vs. GeomKway crossover** (n=128, 12.6M tetrahedra, 20 ranks): with locality-preserving input, Kway is faster (1.49 s vs. 1.59 s); with fully randomised input, GeomKway wins, and by more than before (2.96 s vs. 4.36 s) — Kway degrades 2.93x under randomisation against GeomKway's 1.86x. `GeomKway` uses the SFC internally to redistribute the graph before the k-way phase, so its improvement here (3.32 s → 2.96 s on the randomised case) is the same routing speed-up showing through a hybrid partitioner rather than a pure geometric one; `Kway` (no SFC involved) is essentially unchanged.
- **ParMETIS re-partitioner vs. partitioning from scratch** (20 ranks, 3.1M cells, deliberately skewed starting distribution, imbalance 3.5 — via the same throwaway C++ driver as the previous re-run, since this operation isn't exposed to Python and doesn't touch the SFC code at all): re-partitioning rebalances to 1.021 while moving 73.7% of cells, against 80.8% moved for partitioning from scratch, at the cost of a 10% larger edge cut (93,447 vs. 84,930) — identical to the previous re-run, as expected.
- **Two-stage SFC-then-PT-SCOTCH vs. PT-SCOTCH alone**, fully randomised input:
  - 12.6M tetrahedra, 20 ranks: time inside SCOTCH falls from 8.74 s to 3.59 s (2.4x), edge cut essentially unchanged (378456 → 380726); end-to-end mesh creation 10.29 s → 5.24 s. Stage 1 (the SFC pass itself) dropped from 0.72 s to 0.69 s — a small, expected change since the SFC-vs-Kway comparison above showed the bigger routing win at higher rank/data counts than this stage 1 call reaches on its own.
  - 42.5M tetrahedra, 20 ranks (measured separately): SCOTCH alone 40.94 s total (34.56 s inside SCOTCH) vs. two-stage 21.03 s total (13.93 s inside SCOTCH) — SCOTCH itself is ~2.5x faster once the input has been pre-localised, edge cut a wash in the other direction (854418 vs. 869254, within PT-SCOTCH's own run-to-run variability).
  - At 50% (rather than fully) randomised input, 42.5M tetrahedra: SCOTCH alone 31.99 s vs. two-stage 18.80 s; the time advantage shrinks with less randomness, as expected, since there is less locality to lose in the first place.

### Rank-count scaling, fixed mesh, fully randomised input

82,944 tetrahedra, `fraction=1.0` (fully randomised). Time in seconds; edge cut in parentheses. Unchanged from the previous re-run within noise — this mesh is too small for the SFC routing change to show up against fixed per-call overhead:

| Ranks | ParMETIS Kway | ParMETIS GeomKway | ParMETIS Geom | PT-SCOTCH | SFC Morton | SFC Hilbert |
|---|---|---|---|---|---|---|
| 1 | 0.039 (0) | 0.032 (0) | 0.021 (0) | 0.037 (0) | 0.023 (0) | 0.023 (0) |
| 2 | 0.066 (3060) | 0.053 (2484) | 0.020 (2390) | 0.066 (2344) | 0.015 (2364) | 0.024 (2464) |
| 3 | 0.063 (4214) | 0.048 (4072) | 0.017 (5812) | 0.094 (4052) | 0.012 (5828) | 0.017 (5570) |
| 5 | 0.048 (6418) | 0.034 (6494) | 0.010 (8672) | 0.092 (5754) | 0.007 (8520) | 0.012 (7240) |

(At 1 rank there is nothing to cut, hence 0.) Same conclusion as both previous write-ups: at this mesh size, absolute times are dominated by fixed per-call overhead rather than by the cost of partitioning itself, so this mainly demonstrates correct behaviour at low rank counts, not asymptotic scaling — see the two-stage numbers above for where the scaling advantage of the SFC/hybrid approach actually shows up.

### Problem-size scaling at fixed rank count (20 ranks, fully randomised input)

| Tetrahedra | ParMETIS Kway | ParMETIS GeomKway | ParMETIS Geom | PT-SCOTCH | SFC→PT-SCOTCH (two-stage) | SFC Morton | SFC Hilbert |
|---|---|---|---|---|---|---|---|
| 12.6M | 4.36 s (439644) | 2.96 s (432946) | 1.21 s (509284) | 11.91 s (378456) | 5.24 s (380726) | 1.06 s (510252) | 1.29 s (457634) |
| 42.5M | 17.78 s (1000720) | 12.23 s (974632) | 4.56 s (1144506) | 39.31 s (854418) | 21.03 s (869254) | 4.06 s (1149220) | 4.60 s (1028176) |
| time ratio for 3.4x more cells | 4.08x | 4.13x | 3.77x | 3.30x | 4.01x | 3.83x | 3.57x |

The overall picture is the same as both previous write-ups — PT-SCOTCH remains the most expensive in absolute terms at both sizes by a wide margin, and the two-stage route is consistently faster than PT-SCOTCH alone at both sizes (5.24 s vs. 11.91 s, and 21.03 s vs. 39.31 s). Every geometric/hybrid column (GeomKway, Geom, Morton, Hilbert, two-stage) is faster in absolute terms than on the previous re-run, consistent with the SFC routing speed-up; `Kway` and `PT-SCOTCH`, which don't touch the SFC code, are essentially unchanged. The time-ratio spread (3.30x–4.13x for a 3.4x increase in cell count) is similar in shape to the previous re-run's (3.21x–4.30x) — still more spread than the original write-up's tight 3.3–3.65x cluster, and still flagged as possibly including single-run noise at the larger size rather than a confirmed change in asymptotic scaling. The core motivation for the SFC/hybrid approach — a rank-count-independent partitioning cost — is still not fully visible in a same-rank-count comparison like this one, and would be expected to show up more clearly at higher rank counts than tested here.

## Known follow-up

Redistributing cells for the two-stage comparison currently goes through building a full intermediate mesh, most of whose cost (topology, geometry, ghost cells) is wasted since only the cell destinations are needed before the second stage. A `redistribute_by_partitioner` helper that calls the partitioner directly and exchanges only the cell rows was prototyped in the demo and measured at roughly half the cost of the full-mesh route for this step, but is not yet part of the public API (confirmed still the case).

Filed [#4402](https://github.com/FEniCS/dolfinx/issues/4402) (still open): for mixed-topology meshes, cell redistribution currently pays for one NBX neighbourhood-discovery round per cell type; the destination rank sets for different cell types typically overlap substantially, so this is a candidate for future optimisation. Does not affect single-cell-type meshes, which are unaffected (one cell type, one round).

`graph::parmetis::repartitioner` (see "What's new" above) is still not exposed to Python — the re-partitioner benchmark above was re-measured with a throwaway C++ driver rather than through `demo_partition.py`.

## Testing

C++ unit tests pass on 1, 2 and 3 ranks; Python test suite passes (re-checked `test_mesh_partitioners.py` on this pass); `demo_partition.py` runs via the demo test harness across multiple rank counts and mesh sizes, including a 42.5M-tetrahedron mesh on 20 ranks.


cpp/dolfinx/graph/sfc.h
```diff
@@ -0,0 +1,116 @@
+// Copyright (C) 2020-2026 Garth N. Wells
```
jorgensd:
```suggestion
// Copyright (C) 2026 Garth N. Wells
```

cpp/dolfinx/graph/sfc.cpp
```diff
@@ -0,0 +1,330 @@
+// Copyright (C) 2020-2026 Garth N. Wells
```
jorgensd:
```suggestion
// Copyright (C) 2026 Garth N. Wells
```

python/demo/demo_partition.py
```diff
@@ -0,0 +1,513 @@
...
+        point with global index ``j + offset``, where ``offset`` is the
+        number of points held by all lower ranks.
+
+        The two are not aligned with one another: the vertex indices in
+        ``cells`` refer to the global point numbering, and the
+        corresponding coordinates are in general held by other ranks.
+    """
+    # Cells: this rank's slice of the n^3 hexahedra
+    c0 = (n**3 * comm.rank) // comm.size
+    c1 = (n**3 * (comm.rank + 1)) // comm.size
```
jorgensd: Could we use `dolfinx::common::local_range` here instead to document/expose it, as it is a really neat feature: https://github.com/FEniCS/dolfinx/blob/bc3300a791cff2ef7616c37d1c97285b4d0632ad/cpp/dolfinx/common/local_range.h#L16-L27

python/demo/demo_partition.py
```diff
@@ -0,0 +1,513 @@
...
+        for dz in (0, 1)
+        for dy in (0, 1)
+        for dx in (0, 1)
+    ]
+    tets = [(0, 1, 3, 7), (0, 1, 7, 5), (0, 5, 7, 4), (0, 3, 2, 7), (0, 6, 4, 7), (0, 2, 6, 7)]
+    cells = np.stack([v[k] for t in tets for k in t], axis=1).reshape(-1, 4)
+
+    # Points: this rank's slice of the (n + 1)^3 grid points
+    p0 = ((n + 1) ** 3 * comm.rank) // comm.size
+    p1 = ((n + 1) ** 3 * (comm.rank + 1)) // comm.size
```
jorgensd: Same here.

python/demo/demo_partition.py
```diff
@@ -0,0 +1,513 @@
...
+    return np.ascontiguousarray(np.concatenate(recv), dtype=np.int64)
+
+
+# ## Partition quality
+#
+# The imbalance is computed from the number of cells owned by each rank.
+# The edge cut is the global number of inter-process facets: a facet is
+# listed as inter-process on the rank that owns it when its two cells are
+# owned by different ranks, so summing the local counts counts each cut
+# facet once.
```
jorgensd: True for most meshes, but there are meshes with more than two cells per facet (T-joints, 1D graph structures etc.). Maybe worth  mentioning?

python/demo/demo_partition.py
```diff
@@ -0,0 +1,513 @@
...
+    msh.topology.create_entities(tdim - 1)
+    msh.topology.create_connectivity(tdim - 1, tdim)
+    cut = comm.allreduce(len(msh.topology.interprocess_facets()), MPI.SUM)
+
+    return imbalance, cut
+
+
+# ## Creating a mesh with each partitioner
+#
+# The coordinate element is the same in each case; only the partitioner and
```
jorgensd: Reference coordinate element class?

python/demo/demo_partition.py
```diff
@@ -0,0 +1,513 @@
...
+comm = MPI.COMM_WORLD
+
+# Number of divisions in each direction of the cube, giving 6 n^3
+# tetrahedra. The effect of the input cell distribution on partitioning
+# cost is visible only on a large mesh, so `n` is large by default. It is
+# reduced when there is nothing to partition (a single rank), and in CI,
+# where the demo must run quickly. Set DEMO_PARTITION_N to override, e.g.
+# when running on few ranks.
+_small = comm.size == 1 or "CI" in os.environ or "GITHUB_ACTIONS" in os.environ
+n = int(os.environ.get("DEMO_PARTITION_N", 24 if _small else 128))
```
jorgensd: We usually don't use environmental variables in demos in this way. Do we want to introduce it here? Or should one have an arg-parser?

garth-wells:
# Migration guide: user-supplied mesh partitioners (#4403)

PR #4403 adds space-filling-curve (SFC) and hybrid SFC/graph partitioners
and, along the way, simplifies the partitioner interface used by
`create_mesh` and friends. **If you only use the built-in partitioners
(`dolfinx.graph.partitioner()`, `partitioner_scotch()`,
`partitioner_parmetis()`, `partitioner_kahip()`) you don't need to do
anything** — the defaults are unchanged. This guide is for anyone who
wrote a *custom* partitioner function, or called `create_cell_partitioner`
explicitly.

## Summary of the change

Previously, the function passed as `create_mesh`'s `partitioner`
argument operated on **raw cells** (`cell_types` + cell connectivity)
and was responsible for building the dual graph itself if it needed
one (`mesh::CellPartitionFunction` in C++). Ghost mode and
`max_facet_to_cell_links` were baked into the partitioner at
construction time via `create_cell_partitioner(part, ghost_mode,
max_facet_to_cell_links)`.

Now, `create_mesh` builds the dual graph (and, for geometric/hybrid
partitioners, the cell centroids) itself, once, and calls the
partitioner directly with that data. `ghost_mode` is a new, standalone
`create_mesh` argument, applied at call time, independent of which
partitioner is in use. `create_cell_partitioner` and
`CellPartitionFunction` are gone, in both Python and C++.

## Python

### Before

```python
def my_partitioner(comm, nparts, local_graph, node_weights, edge_weights, ghosting):
    ...
    return dolfinx.graph.adjacencylist(dest, offsets)

partitioner = dolfinx.mesh.create_cell_partitioner(
    my_partitioner, dolfinx.mesh.GhostMode.shared_facet, max_facet_to_cell_links=2
)
msh = dolfinx.mesh.create_mesh(comm, cells, e, x, partitioner=partitioner)
```

### After

```python
def my_partitioner(comm, nparts, local_graph, node_weights, edge_weights, ghosting):
    ...
    return dolfinx.graph.adjacencylist(dest, offsets)

msh = dolfinx.mesh.create_mesh(
    comm, cells, e, x,
    partitioner=my_partitioner,
    ghost_mode=dolfinx.mesh.GhostMode.shared_facet,
    max_facet_to_cell_links=2,
)
```

Key points:

- **`dolfinx.mesh.create_cell_partitioner` no longer exists.** Pass your
  partitioning function straight through as `create_mesh`'s
  `partitioner` argument (also accepted by `create_box`,
  `create_rectangle`, `create_interval` and
  `dolfinx.io.gmsh.model_to_mesh`).
- **`ghost_mode` is a new, standalone `create_mesh` argument** (it did
  not exist as a `create_mesh` parameter before at all). It now applies
  regardless of which partitioner kind you use — with one exception,
  see the geometric caveat below — instead of being fixed when the
  partitioner was constructed.
- **The function shape passed to `create_mesh` is now the dual-graph
  shape**, i.e. exactly `dolfinx.mesh.PartitioningFunc`: `(comm,
  nparts, local_graph, node_weights, edge_weights, ghosting) ->
  AdjacencyList_int32`, where `local_graph` is the mesh dual graph
  `create_mesh` builds for you. If your old function already matched
  this shape (i.e. it was written to be passed *into*
  `create_cell_partitioner`, not the cell-level function
  `create_cell_partitioner` produced), no changes are needed beyond
  removing the wrapper call, as in the example above.
- **Weights are now `None`, not an empty array, when absent.**
  `node_weights`/`edge_weights` are typed `npt.NDArray[np.int32] |
  None`. If your function branched on `len(node_weights) == 0`, change
  it to check `node_weights is None`.
- If your old function instead matched the *cell-level* shape (took
  `cell_types`/`cells` directly, e.g. because it was written against
  the `partitioner` type hint of the old `dolfinx.io.gmsh.model_to_mesh`,
  or against `mesh::CellPartitionFunction` at the C++ level), and it
  used cell/node position rather than the dual graph, see the geometric
  partitioner section below. If it genuinely needs both raw cell data
  and to build its own dual graph, it will need to be rewritten against
  the new dual-graph-based shape above.

### New: geometric and hybrid partitioners

If your custom partitioner used cell/node position rather than graph
edges (e.g. a Morton/Hilbert curve, or a wrapper that computed
centroids and delegated to `ParMETIS_V3_PartGeom`), there is now a
dedicated, cheaper path that skips dual-graph construction entirely:

```python
def my_geometric_partitioner(comm, nparts, x, node_weights):
    # x: cell centroids, shape (num_cells, gdim)
    ...
    return dest  # np.ndarray[int32], one destination rank per cell

partitioner = dolfinx.mesh.create_geometric_cell_partitioner(my_geometric_partitioner)
msh = dolfinx.mesh.create_mesh(comm, cells, e, x, partitioner=partitioner)
```

Built-in geometric partitioners — `dolfinx.graph.partition_morton`,
`partition_hilbert`, and (with ParMETIS) `partitioner_parmetis_geom` —
can be passed directly as `create_mesh`'s `partitioner`, no wrapping
needed.

**Caveat:** a geometric partitioner has no access to the dual graph, so
it can never ghost — the resulting mesh always has `GhostMode.none`
regardless of the `ghost_mode` argument you pass to `create_mesh`. If
you need both position-based partitioning *and* ghosting, use
`create_hybrid_cell_partitioner` instead, which is always given the
dual graph as well as cell centroids (e.g. ParMETIS `GeomKway`, exposed
as `dolfinx.graph.partitioner_parmetis_hybrid`).

## C++

### Before

```cpp
mesh::CellPartitionFunction partitioner = mesh::create_cell_partitioner(
    mesh::GhostMode::shared_facet, graph::partition_graph,
    /*max_facet_to_cell_links=*/2);

auto mesh = mesh::create_mesh(comm, comm, cells, cell_weights, elements, commg,
                               x, xshape, partitioner,
                               /*max_facet_to_cell_links=*/2, /*num_threads=*/1);
```

### After

```cpp
graph::Partitioner partitioner{.fn = graph::partition_graph,
                               .node_weights = cell_weights};

auto mesh = mesh::create_mesh(comm, comm, cells, elements, commg, x, xshape,
                               partitioner, mesh::GhostMode::shared_facet,
                               /*max_facet_to_cell_links=*/2, /*num_threads=*/1);
```

Key points:

- **`mesh::CellPartitionFunction` and `mesh::create_cell_partitioner`
  are removed.** `create_mesh` now takes a `const graph::Partitioner&`
  — a small struct bundling a `graph::AnyPartitionFunction fn` (a
  `std::variant` of `graph::partition_fn`, `graph::geom_partition_fn`
  and `graph::hybrid_partition_fn`) with the node weights it should be
  called with.
- **`cell_weights` moves from its own `create_mesh` parameter into
  `graph::Partitioner::node_weights`**, and changes from a plain,
  non-optional `std::span<const std::int32_t>` (empty span meant
  unweighted) to `std::optional<std::span<const std::int32_t>>`
  (`std::nullopt` means unweighted).
- **`graph::partition_fn`'s signature is unchanged**: `(MPI_Comm comm,
  int nparts, const AdjacencyList<std::int64_t>& local_graph,
  optional<span<const int32_t>> node_weights,
  optional<span<const int32_t>> edge_weights, bool ghosting) ->
  AdjacencyList<int32_t>`. If you previously wrote a function of this
  shape to hand to `create_cell_partitioner`, assign it directly as
  `graph::Partitioner::fn` — it converts implicitly into
  `AnyPartitionFunction`; no other changes needed.
- **`ghost_mode` and `max_facet_to_cell_links` become `create_mesh`'s
  own, separate parameters**, applied at call time rather than baked
  into the partitioner. `ghost_mode` has no effect when `fn` holds a
  `geom_partition_fn` (see the geometric caveat above).
- **The dual graph is now built once by `create_mesh` itself** and
  handed to `partition_fn`/`hybrid_partition_fn`; a
  `CellPartitionFunction` that used to call `build_dual_graph` (or
  equivalent) internally on every invocation should drop that step and
  just consume the `AdjacencyList` it's given.
- **New: `graph::geom_partition_fn`** — `(MPI_Comm, int nparts,
  span<const double> x, int gdim, optional<span<const int32_t>>
  node_weights) -> std::vector<int>` — for position-only partitioners
  (`graph::partition_sfc_morton`, `graph::partition_sfc_hilbert`,
  `graph::parmetis::geom_partitioner`). `create_mesh` computes cell
  centroids for you and supplies them as `x`. Assign such a function
  directly to `graph::Partitioner::fn`, e.g.
  `graph::Partitioner{.fn = my_geom_partition_fn}` — no separate
  wrapping helper is needed in C++ (unlike Python, where
  `create_geometric_cell_partitioner`/`create_hybrid_cell_partitioner`
  exist only because nanobind cannot tell the three function shapes
  apart from a plain Python callable's signature the way `std::variant`
  construction can from a C++ function's type).
- **New: `graph::hybrid_partition_fn`** — `partition_fn` plus a
  centroid per node — for partitioners that need both the dual graph
  and positions (`graph::parmetis::geom_partitioner_kway`, i.e.
  ParMETIS `GeomKway`).
- **ABI note**: this refactor changes the mangled names of `create_mesh`
  and related functions in this area (a defaulted `num_threads`
  parameter was also added to the dual-graph build path along the way).
  Reinstall/rebuild against the new headers rather than relying on an
  incremental/binary-compatible update.
- `create_box`, `create_rectangle` and `create_interval` move from
  `mesh::CellPartitionFunction` to `graph::AnyPartitionFunction` for
  their `partitioner` parameter (note: unlike `create_mesh`, these take
  the bare `AnyPartitionFunction`, not the weight-bundling
  `graph::Partitioner`, since generated cells carry no weights), so a
  geometric or hybrid partitioner can now be used with the built-in
  mesh generators too, not just `create_mesh` directly.

## Quick reference

| Before | After |
|---|---|
| `dolfinx.mesh.create_cell_partitioner(fn, ghost_mode, max_facet_to_cell_links)` | *(removed)* — pass `fn` directly as `create_mesh(..., partitioner=fn, ghost_mode=..., max_facet_to_cell_links=...)` |
| `mesh::create_cell_partitioner(ghost_mode, partfn, max_facet_to_cell_links)` | *(removed)* — assign `partfn` directly as `graph::Partitioner::fn`; `ghost_mode`/`max_facet_to_cell_links` are `create_mesh` parameters |
| `mesh::CellPartitionFunction` (operates on raw cells, builds its own dual graph) | `graph::partition_fn` (operates on the dual graph `create_mesh` builds for you), via `graph::Partitioner`/`graph::AnyPartitionFunction` |
| C++ `cell_weights` as a separate, non-optional `create_mesh` span parameter | `graph::Partitioner::node_weights`, `std::optional<std::span<const std::int32_t>>` |
| Python/C++ partitioner-callback node/edge weights: empty array/span = "none" | `None`/`std::nullopt` = "none" |
| Custom position-based partitioner hand-rolled around `CellPartitionFunction` | `graph::geom_partition_fn` / `dolfinx.mesh.create_geometric_cell_partitioner` |

PR #4408, opened 2026-08-14
Spack CI simplification

garth-wells:
## Summary

Rebuilds the Spack CI workflow around a **single Spack environment** driven by a
**build matrix**, replacing the previous two-environment (`cxx` + `py`) layout.

The old workflow created two environments and concretised and installed DOLFINx
in each. The two resolve to an identical runtime spec and differ only in
build-time dependencies, so the second install rebuilt the same library under a
different hash. There is now one `ci` environment that both the C++ and the
Python test steps activate.

## Test matrix

`fail-fast: false`, nine jobs:

| MPI | PETSc scalar | PETSc index |
|---|---|---|
| OpenMPI | float32 | int32 |
| OpenMPI | float64 | int32 |
| OpenMPI | complex64 | int32 |
| OpenMPI | complex128 | int32 |
| OpenMPI | float64 | int64 |
| OpenMPI | complex128 | int64 |
| MPICH | float64 | int32 |
| MPICH | complex128 | int32 |
| OpenMPI | *no PETSc* | — |

Scalar/index coverage sits on OpenMPI; MPICH repeats the two most common
combinations, since it exercises the MPI layer rather than the scalar type. The
PETSc-free leg covers the optional-PETSc code paths and needs only one MPI.

The per-job `petsc_dep` solver sets mirror the per-arch `--download-` packages in
`docker/Dockerfile.test-env`, so this workflow and `ccpp.yml` build PETSc the same
way. Two of those have no Spack variant in `petsc` (`spai` on real64-32, serial
`superlu` on the int32 arches); everything else matches.

## Container and Mesa

- `container:` moves from `ubuntu:24.04` to `ubuntu:26.04`, matching
  `runs-on: ubuntu-26.04`. Buildcache pushes use the same base image.
- Mesa is now an external in `gh-actions-env-test.yml` rather than a Spack build.
  `mesa@25.0.5` — the newest the pinned `packages_ref` knows — does not compile
  against Ubuntu 26.04's glibc (its bundled C11 threads polyfill redeclares
  `once_flag`/`call_once`), and newer releases reject meson options the recipe
  still passes. The distro build matches the container's glibc by construction.

## Other workflow changes

- Job names render the matrix entry, e.g. `ci-spack-build (openmpi, petsc complex64/int32)`.
- New guard step asserts `dolfinx.has_petsc` matches the matrix, so a
  mis-specified variant cannot silently build PETSc into the PETSc-free leg and
  leave it testing nothing new.
- The PETSc-free leg skips the C++ demos (several demo `CMakeLists` reference
  `PETSC_INCLUDE_DIRS` unconditionally, which `--warn-uninitialized` turns into an
  error, and the top-level demo `CMakeLists` adds every subdirectory) and passes
  `-m "not petsc4py"` to the Python unit tests.
- Buildcache push moved ahead of the test steps so artifacts survive a failing
  test; only the final push rebuilds the index, which is the expensive part.

## Supporting source changes

**`python/test/unit/fem/test_quadrature_elements.py`** — the new `complex64` leg
exposed a latent tolerance bug in `test_quadrature_assembly`. The test compared a
quadrature-element assembly against a `SpatialCoordinate` assembly with
`assert_allclose`'s default `rtol=1e-7`, which is *below* the `float32` epsilon of
`1.1920929e-07`. On `complex64` the two paths differ by under one ULP — the
observed failure was 1 element of 33 at a relative difference of `1.1649415e-07`
— so the assertion could never be robust there. The tolerance is now scaled from
the scalar type (`100 * eps`, matching the idiom in `test_dofmap.py` and
`test_dof_permuting.py`). It failed only in the MPI pass, where the ghost scatter
changes the summation order.

**`python/demo/demo_stokes.py`** — `block_direct_solver()` now uses MUMPS
unconditionally instead of falling back to SuperLU_DIST on `int64`, and the
`mixed_direct()` velocity-norm check is skipped on `int64`.

> ⚠️ The skip is a workaround and wants a proper fix before merge. The evidence
> points at SuperLU_DIST rather than MUMPS: on `complex128/int64`, `mixed_direct()`
> (still on SuperLU_DIST) returned velocity 17.5434 vs 17.4784 and pressure
> **76450 vs 311**, while `block_direct_solver()` on forced MUMPS returned the
> correct 17.478370157081073 / 311.44253797758 in the same run. The C++
> `demo_mixed_poisson` at np=2 also routes to SuperLU_DIST on `int64` and is the
> only demo in the matrix reporting `DIVERGED_PC_FAILED`. Switching
> `mixed_direct()` to MUMPS as well would likely remove the need for the skip.

## Known remaining CI failures

Two failures in this workflow are not caused by the changes here and do not
reproduce outside the Spack environment (checked against the same source in
`ghcr.io/fenics/dolfinx/dolfinx:nightly`, serial and with 4 xdist workers):

- **8 SIGSEGVs inside `basix::create_element`**, always the same high-degree
  hexahedral H(curl)/H(div) tests (`test_evaluation[3-hexahedron-AAE/AAF]`,
  `test_AA_hex[3-AAE/AAF]`, `test_discrete_curl[...-4-...]`,
  `test_NCE_interpolation[4-hex]`, `test_interpolation_non_affine_nonmatching_maps`).
  Intermittent per job, identical set whenever it strikes. Not memory pressure —
  that whole set peaks at 208 MB RSS with only 4 workers.
- **`demo_half-loaded-waveguide.py` SIGSEGV** under `mpiexec -np 3` on the
  MPICH `float64/int32` leg.

Both look like the spack toolchain (gcc 15.2, `-O3`, `target=x86_64_v3`) rather
than DOLFINx.

---

AI assistance: I used Claude Code to draft parts of this PR. I reviewed, edited,
tested, and take responsibility for the final contribution.


python/demo/demo_stokes.py
```diff
@@ -682,4 +678,5 @@ def mixed_direct():
 # Solve using a non-blocked matrix and an LU solver
 
 norm_u_4, norm_p_4 = mixed_direct()
-np.testing.assert_allclose(norm_u_4, norm_u_0, rtol=1e-4)
+if PETSc.IntType != np.int64:
```
michalhabera, line 681: What is this condition supposed to assert?

PR #4410, opened 2026-08-14
Spack CI: run the petsc4py tests, and pin/harden the workflow

garth-wells:
Two test fixes and some hardening of the Spack CI workflow. Unrelated to
each other except that the first is what exposed the second.

## Run the petsc4py tests

The pytest marker expression was written

```yaml
${{ matrix.petsc && '' || '-m "not petsc4py"' }}
```

GitHub expressions treat the empty string as false, so on the PETSc legs
`true && ''` fell through to the fallback and *every* leg ran with `-m
"not petsc4py"`. The eight petsc4py-marked test modules were skipped
everywhere, not just on the PETSc-free leg.

Inverting the condition so the non-empty branch carries the marker
restores 174 tests per PETSc leg. On `openmpi, petsc float64/int32` the
serial run goes from `2983 passed, 174 deselected` to `3157 passed`, and
the `np=3` run from 2101 to 2269.

## Scale a nonlinear assembly tolerance with the working precision

One of the newly-restored tests then failed, on the single-precision legs
only:

```
test_matrix_assembly_block_nl
assert 5.832494375549596 == 5.832486629486084 ± 5.8e-06
```

`test_matrix_assembly_block_nl` compares the operator norms from the
blocked, nested and monolithic assembly paths against fixed relative
tolerances of 1e-6/1e-5. The three paths accumulate in different orders,
so at `np=3` in single precision the blocked and nested norms differ by
1.33e-6 relative — about 11 ulp, and a real failure on `float32` and
`complex64`. Both tolerances are now floored at `500 * eps` of the
working precision: 1.1e-13 in double, so the existing limits are
unchanged there, and 6.0e-5 in single.

## Workflow hardening

- **Pin `fenics/spack-fenics`.** Spack and its packages repo were pinned;
  this one tracked `main`, so it could change what CI builds with no
  commit here — a red run that cannot be attributed to the pull request.
  The pin names the current `main`, so nothing about the build changes.
  The UFL/FFCx/Basix checkouts stay on `main` deliberately: testing
  against the development branches of the stack is the point of this
  workflow, and those SHAs are recorded and passed to spack as `commit=`.
- **Skip the buildcache pushes on fork pull requests.** They run with a
  read-only `GITHUB_TOKEN` whatever `permissions` asks for, so the push
  would 401 as soon as it had anything to upload, failing a contributor's
  job on a step that tests nothing of theirs. Registry reads are
  anonymous, so those builds still take from the cache.
- **Pin `actions/checkout` to a commit**, matching `spack/setup-spack`,
  and state `contents: read` alongside `packages: write`.
- **`-j $(nproc)`** instead of a hardcoded `-j 4`, so the cap is not
  invisible if the job moves to a larger runner.
- **Drop `-V` from ctest**: it prints output unconditionally, which makes
  the neighbouring `--output-on-failure` inert.
- `--no-install-recommends` on the apt install.

## Known issue, not addressed here

The Spack CI legs segfault intermittently — 0-3 of the 9 legs, varying
run to run — inside basix element creation for high-degree hexahedral
H(curl)/H(div) elements. It predates this branch. Ruled out so far:
OpenBLAS kernel dispatch (a non-dispatching build whose compiled-in core
matched the CPU crashes identically), lazy symbol binding, and anything
reproducible on an Intel host; every occurrence so far has been on an AMD
EPYC 9V74 runner, where it reproduces from a one-line element
construction with no pytest involved. The instrumentation used to
establish that is in `169cf98c54` and was removed again in `c2cc5eabff`,
since it did not identify the fault.

Note that restoring the petsc4py tests makes one more test
(`test_petsc_discrete_operators.py`) eligible to hit the crash.

---

reviewed, edited, tested, and take responsibility for the final
contribution.


.github/workflows/ci.yml
```diff
@@ -75,11 +76,14 @@ jobs:
           packages_ref: d342cee94fb90fb020776c8bd8500b54a19c63cf
           # packages_ref: v2026.06.0
 
+      # Pinned, like spack itself and its packages repo above: an unpinned
+      # 'main' changes what CI builds with no commit here, which makes a
+      # failure impossible to attribute. Bump deliberately.
       - name: Get FEniCS Spack packages
-        uses: actions/checkout@v7
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
```
jhale, line 83: There is no need to do this for GitHub provided actions, tags are stable and well maintained.

.github/workflows/ci.yml
```diff
@@ -75,11 +76,14 @@ jobs:
...
         with:
           repository: fenics/spack-fenics
-          ref: main
+          ref: 05ea6cd78f5fde9c580fe29cb6dbb040d05f74ee # main, 2026-08-12
```
jhale, line 86: This repo moves pretty slowly, don't see the necessity.

python/test/unit/fem/test_petsc_nonlinear_assembler.py
```diff
@@ -199,14 +199,21 @@ def monolithic():
             b.destroy()
             return Anorm, bnorm
 
+        # The blocked, nested and monolithic operators are assembled in
```
jhale, line 202: Trivial comment.

Issue #2700, opened 2023-06-30
Add function to remove (explicit) zero entries from a CSR matrix


IgorBaratta:
### Describe new/missing feature

Explicit zero values can be added to the entries of CSR matrix.
For some types of elements, the element tensor is not dense, but the sparsity pattern doesn't have this information.

Example using scipy:
``` python3
V = dolfinx.fem.FunctionSpace(mesh, element)
u, v = ufl.TrialFunction(V), ufl.TestFunction(V)
a = ufl.inner(ufl.grad(u), ufl.grad(v)) * ufl.dx + ufl.inner(u, v) * ufl.dx
a = dolfinx.fem.form(a, jit_options=jit_options)
A = dolfinx.fem.assemble_matrix(a)


# Convert to scipy sparse matrix
from scipy.sparse import csr_matrix
D = csr_matrix((A.data, A.indices, A.indptr))


print(D.shape, D.nnz, D.nnz/(D.shape[0]*D.shape[1]) * 100, "%")

D.data[D.data < 1e-15] = 0
D.eliminate_zeros()
D.prune()

print(D.shape, D.nnz, D.nnz/(D.shape[0]*D.shape[1]) * 100, "%")
```
Results
```
# Before
(64, 64) 512 12.5 %
# After
(64, 64) 256 6.25 %
```

Maybe a similar function should be implemented for `MatrixCSR.h`.

### Suggestion user interface

```python3
class MatrixCSR
{
...
...

void eliminate_zeros(Scalar tol)
}
```

IgorBaratta:
> This will require `MatrixCSR` to recompute its sparsity and some of its internal data (insert position for finalize for example).

Yes, I think that's the main idea of the issue.
Maybe instead of changing the matrix in place, we could create a new matrix with a new sparsity pattern.

> We also should check (in debug mode) that insertion to indices which are not in the sparsity cause appropriate exceptions.

This something that we want to do anyway, right? Even without removing zeros, since we call `std::lower_bound`.

garth-wells:
PETSc now has a function that does this for its matrix types, https://petsc.org/release/manualpages/Mat/MatFilter/.

PR #4411, opened 2026-08-14
Eliminate zeros from MatrixCSR

chrisrichardson:
Similar to MatFilter in Petsc, this removes explicit zeros (or near zeros) from MatrixCSR data.
Because it affects the structure, no add, insert or scatter operations can be performed afterwards, and a `_finalized` lock is set.

Fixes Issue #2700 

PR assistance with Claude Sonnet 5


cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -581,6 +586,77 @@ class MatrixCSR
...
+  void eliminate_zeros(value_type tol = 0)
+  {
+    // Remove any zero entries (blocks, where all entries in the block
+    // are within tolerance of zero) in data, and update the column
+    // indices and row pointers accordingly.
+    const std::size_t bs2 = _bs[0] * _bs[1];
+
+    // True if every entry of the block starting at block index j is
+    // within tolerance of zero, i.e. the whole block can be dropped.
+    auto is_zero_block = [&](std::int64_t j)
```
garth-wells: We avoid `[&]`.

PR #4423, opened 2026-08-19
Remove reassignment of variable to itself

schnellerhase:
Code paths dealing with (mutable) optionals looked like 

```python
a = DEFAULT if a is None else a
```
however, the reassignment `a=a` in the non-default case should not happen. Replaced throughout with

```python
if a is None:
  a = DEFAULT
```


jhale:
FWIW I always preferred the plain `if` statement! Are there more of these to change?

schnellerhase:
Checked for all `* is None else *` occurrences.

PR #4425, opened 2026-08-19
Fix unchecked errors in SLEPcEigenSolver and trim its API

jhale:
`la::SLEPcEigenSolver` dropped the return code of nearly every SLEPc call. SLEPc
signals failure through the return code and leaves its output arguments
untouched, so each dropped code turned a diagnosable error into an indeterminate
value:

- `get_eigenvalue`, `get_eigenpair` and `get_number_converged` read an
  uninitialised `PetscInt` when `EPSGetConverged` failed, which it does via
  `EPSCheckSolved` if `EPSSolve` has not been called.
- `solve()` called `MatGetSize` on the null `Mat` that `EPSGetOperators` returns
  when no operators have been set (`if (k<1) *A = NULL;`).
  `PetscValidHeaderSpecific` compiles to a no-op in optimised PETSc builds, so
  this was a null dereference rather than a clean error.
- The `EPSType` from `EPSGetType` was formatted without a null check; it is null
  until `EPSSetUp` has run.
- `get_options_prefix` constructed a `std::string` from the null pointer
  `PetscObjectGetOptionsPrefix` returns when no prefix is set (`*prefix =
  obj->prefix;`, null by default).

Every return code is now checked with the `CHECK_ERROR` macro already used in
`petsc.cpp`, which routes the code to `petsc::error` and throws
`std::runtime_error`. The null-prefix bug is fixed in `petsc::Vector`,
`petsc::Matrix` and `petsc::KrylovSolver` too.

Checking the return code is sufficient for the argument errors SLEPc already
detects, so the class does not add its own guards:
`EPSGetEigenvalue`/`EPSGetEigenpair` call `EPSCheckSolved` and bounds check the
index, and `EPSSetUp` errors with "EPSSetOperators must be called first". These
use `PetscCheck`, which — unlike `PetscValidHeaderSpecific` — is unconditional,
so they hold in optimised builds.

### Behaviour change

`solve()` no longer calls `EPSSetFromOptions`. It ran *after* all user
configuration, so the options database overrode anything applied between
construction and the solve — and because `EPSSetFromOptions` applies `-eps_nev`
via `EPSSetDimensions`, `-eps_nev 10` silently overrode a requested three
eigenpairs. Callers wanting database options call `set_from_options()`
themselves, after `set_options_prefix()` and after any settings the database
should override.

### API changes

Following review, the class is trimmed to what actually earns a wrapper: `EPS`
lifetime management, hiding `char*` behind `std::string`, and normalising the
real/complex split in `get_eigenvalue`.

- `get_eigenvalue` and `get_eigenpair` take `PetscInt` rather than `int`, and
  `solve` no longer takes a count. `PetscInt` removes a `static_cast` at every
  SLEPc call and matches the `PetscScalar`/`Mat`/`Vec` already in these
  signatures. The fixed-width rule in `AGENTS.md` governs DOLFINx entity indices,
  not arguments handed straight to PETSc; `AGENTS.md` now records the
  distinction.
- `get_iteration_number` is removed — a passthrough to `EPSGetIterationNumber`
  that manages nothing and saves no boilerplate. The iteration count is still
  logged after a solve.
- `get_number_converged` is removed. It existed to bound a loop over
  `get_eigenvalue`; `EPSGetConverged(solver.eps(), &nconv)` does that directly,
  and the accessor documentation points at it.
- `solve(n)` is removed in favour of a bare `solve()`. Setting `nev` behind the
  caller's back clobbered a programmatic `EPSSetDimensions` on `eps()`, and the
  no-argument overload requested as many eigenpairs as the global dimension of
  `A`, which sized SLEPc's working subspace to match. Dimensions are now set on
  `eps()` like every other solver option:

  ```cpp
  SLEPcEigenSolver solver(comm);
  solver.set_operators(A, nullptr);
  EPSSetDimensions(solver.eps(), 5, PETSC_DETERMINE, PETSC_DETERMINE);
  EPSSetWhichEigenpairs(solver.eps(), EPS_SMALLEST_REAL);
  solver.solve();
  ```

The class has no in-tree callers, no Python binding and no demo, so the practical
impact of the removals is nil. Exposing it through nanobind is not proposed —
`slepc4py` covers Python.

### Testing

Adds `cpp/test/slepc.cpp`, the first test coverage for this class: a Hermitian
problem with a known spectrum, a non-symmetric operator with a known complex
conjugate pair (checking that `get_eigenvalue` is independent of the PETSc scalar
type while raw `get_eigenpair` storage is not), and the error paths above. Passes
at np = 1, 2, 3, and on the real and complex CI legs.

Reverting `slepc.h`/`slepc.cpp` to their previous state fails six assertions and
aborts in `get_options_prefix` with libc++'s
`basic_string(const char*) detected nullptr` hardening assertion, so the test does
exercise the fixes.

---

AI assistance: I used Claude Opus 5 (Claude Code) to draft parts of this PR. I

### DIFF
--- a/cpp/dolfinx/la/slepc.cpp
+++ b/cpp/dolfinx/la/slepc.cpp
@@ -8,30 +8,39 @@
 
 #include "slepc.h"
 #include "petsc.h"
-#include "utils.h"
-#include <dolfinx/common/MPI.h>
+#include <cassert>
 #include <dolfinx/common/log.h>
 #include <format>
-#include <petscmat.h>
-#include <slepcversion.h>
+#include <stdexcept>
+#include <utility>
 
 using namespace dolfinx;
 using namespace dolfinx::la;
 
 //-----------------------------------------------------------------------------
-SLEPcEigenSolver::SLEPcEigenSolver(MPI_Comm comm) { EPSCreate(comm, &_eps); }
+#define CHECK_ERROR(NAME)                                                      \
+  do                                                                           \
+  {                                                                            \
+    if (ierr != 0)                                                             \
+      petsc::error(ierr, __FILE__, NAME);                                      \
+  } while (0)
+
+//-----------------------------------------------------------------------------
+SLEPcEigenSolver::SLEPcEigenSolver(MPI_Comm comm) : _eps(nullptr)
+{
+  PetscErrorCode ierr = EPSCreate(comm, &_eps);
+  CHECK_ERROR("EPSCreate");
+}
 //-----------------------------------------------------------------------------
 SLEPcEigenSolver::SLEPcEigenSolver(EPS eps, bool inc_ref_count) : _eps(eps)
 {
   if (!eps)
     throw std::runtime_error("SLEPc EPS must be initialised before wrapping");
 
-  PetscErrorCode ierr;
   if (inc_ref_count)
   {
-    ierr = PetscObjectReference((PetscObject)_eps);
-    if (ierr != 0)
-      petsc::error(ierr, __FILE__, "PetscObjectReference");
+    PetscErrorCode ierr = PetscObjectReference((PetscObject)_eps);
+    CHECK_ERROR("PetscObjectReference");
   }
 }
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


jorgensd:
Is there any value in exposing the `slepceigensolver` to Python?

schnellerhase:
> Is there any value in exposing the slepceigensolver to Python?

Not really that should just be done through `slepc4py`, I think. 

On that note why are we maintaining for `SLEPc` a custom wrapper? That is not following the design idioms of how we interact with 3rd party code elsewhere. Would it not make sense to remove and add a (C++) demo on correct usage of `SLEPc` instead?

cpp/dolfinx/la/slepc.cpp
```diff
@@ -56,111 +65,137 @@ SLEPcEigenSolver::operator=(SLEPcEigenSolver&& solver) noexcept
...
   assert(n <= _n);
 #endif
 
   // Set number of eigenpairs to compute
-  assert(_eps);
-  EPSSetDimensions(_eps, n, PETSC_DECIDE, PETSC_DECIDE);
-
-  // Set any options from the PETSc database
-  EPSSetFromOptions(_eps);
+  ierr = EPSSetDimensions(_eps, static_cast<PetscInt>(n), PETSC_DECIDE,
```
garth-wells: I find `static_cast<PetscInt>` to be noise. Can we get rid of it? The function argument should probably be `PetscInt` in this function.

cpp/dolfinx/la/slepc.cpp
```diff
@@ -56,111 +65,137 @@ SLEPcEigenSolver::operator=(SLEPcEigenSolver&& solver) noexcept
...
-  spdlog::info("Eigenvalue solver ({}) converged in {} iterations.", eps_type,
-               num_iterations);
+  ierr = EPSGetType(_eps, &eps_type);
+  CHECK_ERROR("EPSGetType");
+  spdlog::info("Eigenvalue solver ({}) converged in {} iterations.",
+               eps_type ? eps_type : "unknown", num_iterations);
 }
 //-----------------------------------------------------------------------------
-std::complex<PetscReal> SLEPcEigenSolver::get_eigenvalue(int i) const
+std::complex<PetscReal> SLEPcEigenSolver::get_eigenvalue(std::int64_t i) const
```
garth-wells: `PetscInt`?

garth-wells (commented):
Nice, just a few inline comments on integer types.

jorgensd:
> > Is there any value in exposing the slepceigensolver to Python?
> 
> Not really that should just be done through `slepc4py`, I think.
> 
> On that note why are we maintaining for `SLEPc` a custom wrapper? That is not following the design idioms of how we interact with 3rd party code elsewhere. Would it not make sense to remove and add a (C++) demo on correct usage of `SLEPc` instead?

PETSc and SLEPC from C++ is always a bit more verbose, and would make sense to have some kind of scaffolding for the C++ folks to piggy-back off. For that reason we do also maintain a PETSc::KrylovSolver: https://github.com/FEniCS/dolfinx/blob/8d913e913dc81268310adda9ad96ee5dd3493591/cpp/dolfinx/la/petsc.h#L251
which is used in the NewtonSolver (until @jhale makes a snes example).

I think having them as part of the C++ library (not exposed through nanobind) is better than having them floating around in demos.

garth-wells:
> > Is there any value in exposing the slepceigensolver to Python?
> 
> Not really that should just be done through `slepc4py`, I think.
> 
> On that note why are we maintaining for `SLEPc` a custom wrapper? That is not following the design idioms of how we interact with 3rd party code elsewhere. Would it not make sense to remove and add a (C++) demo on correct usage of `SLEPc` instead?

Good point. We compromise on the design to help with memory management of PETSc and reduce boilerplate for typical solvers. We do it for SLEPc too for 'symmetry'.

cpp/dolfinx/la/slepc.cpp
jhale: I had the same thought and did some back and forth with the LLM - its justification is that we've specifically asked it to use `std::int64_t` for global indexes in the public API.

jhale: I'm inclined to agree with you though, after all we use PetscReal on the return. Perhaps an exception to the rule?

jhale:
> On that note why are we maintaining for SLEPc a custom wrapper? That is not following the design idioms of how we interact with 3rd party code elsewhere.

Well, because we are lucky to have an excellent Python interface to PETSc, but not a C++ one.

That said, I do think we could strip this `SLEPcEigenSolver` class back a bit to be more in the spirit of DOLFINx. It's not clear why there are methods `get_iteration_number` and `get_number_converged`, or why we have our interface for passing `nev` which has 'strange' interactions with options and programmatically setting `nev` via the first-party interface.

Edit: I'm OK with `*_options_prefix`, don't really want to deal with pointers to char in C++.

jhale:
Thanks for the valuable comments. This is now really minimal - memory management and boiler plate reduction.

It would probably be worth re-reviewing the `PETScKrylovSolver`.

garth-wells (approved):

PR #4427, opened 2026-08-20
Use system MPI in Dockerfile

garth-wells:
Also reduces the number of PETSc builds.

Needs #4426.


docker/Dockerfile.test-env
```diff
@@ -82,9 +73,11 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
     g++ \
     gfortran \
     libboost-dev \
+    libhdf5-mpi-dev \
     liblapack-dev \
     libopenblas-dev \
     libpugixml-dev \
+    mpi-default-dev \
```
jhale, line 81: We should try `--no-install-recommends` here, could reduce the size of the final image.

docker/Dockerfile.test-env
```diff
@@ -82,9 +73,11 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
     g++ \
     gfortran \
     libboost-dev \
+    libhdf5-mpi-dev \
```
jhale, line 77: I had success with system ADIOS2 as well, worth considering.

docker/README.md
```diff
@@ -8,19 +8,19 @@ they are built into, and how they are built.
...
 
-The following images are pushed to both `docker.io/dolfinx` and
+The following image is pushed to both `docker.io/dolfinx` and
 `ghcr.io/fenics/dolfinx` and can be used by end-users to build FEniCS from
 source:
 
-* `dev-env:current-mpich` -  MPICH, debugging off.
-* `dev-env:current-openmpi` - OpenMPI, debugging off.
-* `dev-env:current` - Points at `dev-env:current-mpich`.
+* `dev-env:current` - OpenMPI, debugging off.
```
jhale, line 20: Should document the `:stable` tag here.

docker/README.md
```diff
@@ -8,19 +8,19 @@ they are built into, and how they are built.
...
 DOLFINx based on Ubuntu. It does not contain the FEniCS components.
 
-The following images are produced and pushed to both `docker.io/fenicsproject`
-and `ghcr.io/fenics` and are used in our testing infrastructure:
+The following image is produced and pushed to both `docker.io/fenicsproject`
+and `ghcr.io/fenics` and is used in our testing infrastructure:
 
-* `test-env:current-mpich` - MPICH, debugging on. Used on CircleCI.
-* `test-env:current-openmpi` - OpenMPI, debugging on. Used on GitHub Actions.
+* `test-env:current` - OpenMPI, debugging on. Used on GitHub Actions.
```
jhale: PETSc debugging on? The utility of this image seems pretty minimal now.

garth-wells:
@jhale view on the number of PETSc builds?

docker/README.md
garth-wells: What should it say? I haven't kept track of all the tags.

docker/Dockerfile.test-env
```diff
@@ -268,7 +190,7 @@ RUN apt-get -qq update && \
     --download-superlu_dist \
     --with-scalar-type=real \
     --with-precision=single && \
-    make PETSC_ARCH=linux-gnu-real32-32 ${MAKEFLAGS} all && \
+    make PETSC_ARCH=linux-gnu-real32-32 -j${BUILD_NP} all && \
```
jhale: I saw your question but it disappeared; I don't see any reason to keep 32-bit real/complex builds in these images. Disable? We have test coverage from Spack now.

PR #4433, opened 2026-08-20
Add nls::petsc::SNESSolver, a C++ wrapper for PETSc SNES

jhale:
Wraps PETSc `SNES` in `dolfinx::nls::petsc::SNESSolver`, adapting C++ callables to the `SNES` callback interface. It is intended to supersede the deprecated `nls::petsc::NewtonSolver`.

The class owns the `SNES` object and holds references to the vector and matrices that define the layout of the objects the callbacks assemble into. Configuration of the solve is left to the user, via the options database or the `SNES` object returned by `snes()`. Lifecycle follows `la::petsc::KrylovSolver`: an `MPI_Comm` constructor, a `(SNES, bool inc_ref_count)` wrapping constructor, deleted copies, and `noexcept` moves.

```cpp
nls::petsc::SNESSolver solver(mesh.comm());
solver.set_F([&](const Vec x, Vec b) { ... }, b);
solver.set_J([&](const Vec x, Mat A, Mat) { ... }, A);
solver.set_options_prefix("my_problem_");
solver.set_from_options();
solver.solve(x);
```

The name deliberately differs from the Python `dolfinx.fem.petsc.NonlinearProblem`. That class takes UFL forms, `u` and `bcs`, and creates its own matrices, vectors and `SNES`; this one takes callables and objects the caller owns. `NonlinearProblem` stays free for a form-based C++ class, should one be added.

Also included:

- `fem::petsc::assemble_residual` and `fem::petsc::assemble_jacobian`, C++ counterparts of the functions of the same name in `dolfinx.fem.petsc`, for use as the bodies of the callbacks. Their semantics were checked step by step against the Python versions: ghost updates, zeroing, lifting with `alpha = -1`, `set_bc`, and the guard that inserts a unit diagonal only for square forms.
- The C++ hyperelasticity demo uses both, replacing its hand-written Newton loop and its `HyperElasticProblem` class. It gains a line search, and converges in three Newton iterations rather than four because the hand-written residual did not lift the Dirichlet conditions.
- `set_update`, exposing `SNESSetUpdate`. That PETSc callback takes no context argument, so the solver is recovered from a `PetscContainer` composed on the `SNES`.
- Exceptions thrown by a callback are stored and re-thrown by `solve` rather than being reduced to a PETSc error code. PETSc does not restore its state as the aborted solve unwinds (`x` is left locked for read-only access), so this is documented as terminal for the solver.
- `KrylovSolver::solve` and `SNESSolver::solve` mutate solver state, so neither is `const`. `KrylovSolver::solve` losing `const` is the only change visible to existing code; the two call sites in the tree are unaffected.
- Fix for undefined behaviour when reading an unset options prefix: PETSc reports it as a null pointer, which was passed straight to the `std::string` constructor. Affected `get_options_prefix` on `la::petsc::Vector`, `la::petsc::Matrix` and `la::petsc::KrylovSolver` as well.
- C-style casts of PETSc handles in `la/petsc.cpp` replaced with `reinterpret_cast`.

Tested with `cpp/test/nls/snessolver.cpp` on one, two and three ranks: solve, solve after a move, wrapping an existing `SNES`, a separate preconditioner matrix, the update hook, exceptions from each of the three callbacks, non-convergence, the initial guess being respected, repeated solves, and PETSc reference counts before and after destruction.

The solver only ever handles `Vec` and `Mat`, so `MATNEST` and `VECNEST` need no support and pass through untouched. There is a test for this: two decoupled blocks with different roots, `MPIAIJ` under MPI, preconditioned with `fieldsplit` since a nest matrix cannot be factored directly.

Known gaps, for reviewers:

- The two new `fem::petsc` free functions have no unit test of their own; the hyperelasticity demo is their only coverage.
- The `fem::petsc` assembly functions handle single forms only. Blocked and nest assembly, which the Python class supports, are not implemented: they are built on `VecGhostGetLocalForm`, which does not apply to a `VECNEST`. This is assembly-helper work, not solver work.
- No Python bindings: Python users have petsc4py.
- `assemble_jacobian` takes the preconditioner form `P` as a trailing defaulted pointer, so it is separated from `Pmat` in the argument list. Passing `nullptr` in the middle of the list breaks template argument deduction.
- A `SNES` that outlives its `SNESSolver` carries callbacks pointing at a destroyed object. This is documented on the wrapping constructor rather than defended against: the function and Jacobian contexts cannot be cleared, as `SNESSetFunction` ignores null arguments.

AI assistance: I used Claude Code (Opus 5) to draft parts of this PR. I reviewed, edited, tested, and take responsibility for the final contribution.

### DIFF
--- a/cpp/dolfinx/fem/petsc.h
+++ b/cpp/dolfinx/fem/petsc.h
@@ -181,32 +181,32 @@ Mat create_matrix_block(
 
   // Create PETSc local-to-global map/index sets and attach to matrix
   ISLocalToGlobalMapping petsc_local_to_global0;
-  PetscErrorCode ierr = ISLocalToGlobalMappingCreate(
-      MPI_COMM_SELF, 1, _maps[0].size(), _maps[0].data(), PETSC_COPY_VALUES,
-      &petsc_local_to_global0);
-  common::petsc::check(ierr, "ISLocalToGlobalMappingCreate");
+  common::petsc::check(ISLocalToGlobalMappingCreate(
+                           MPI_COMM_SELF, 1, _maps[0].size(), _maps[0].data(),
+                           PETSC_COPY_VALUES, &petsc_local_to_global0),
+                       "ISLocalToGlobalMappingCreate");
   if (V[0] == V[1])
   {
-    ierr = MatSetLocalToGlobalMapping(A, petsc_local_to_global0,
-                                      petsc_local_to_global0);
-    common::petsc::check(ierr, "MatSetLocalToGlobalMapping");
-    ierr = ISLocalToGlobalMappingDestroy(&petsc_local_to_global0);
-    common::petsc::check(ierr, "ISLocalToGlobalMappingDestroy");
+    common::petsc::check(MatSetLocalToGlobalMapping(A, petsc_local_to_global0,
+                                                    petsc_local_to_global0),
+                         "MatSetLocalToGlobalMapping");
+    common::petsc::check(ISLocalToGlobalMappingDestroy(&petsc_local_to_global0),
+                         "ISLocalToGlobalMappingDestroy");
   }
   else
   {
     ISLocalToGlobalMapping petsc_local_to_global1;
-    ierr = ISLocalToGlobalMappingCreate(MPI_COMM_SELF, 1, _maps[1].size(),
-                                        _maps[1].data(), PETSC_COPY_VALUES,
-                                        &petsc_local_to_global1);
-    common::petsc::check(ierr, "ISLocalToGlobalMappingCreate");
-    ierr = MatSetLocalToGlobalMapping(A, petsc_local_to_global0,
-                                      petsc_local_to_global1);
-    common::petsc::check(ierr, "MatSetLocalToGlobalMapping");
-    ierr = ISLocalToGlobalMappingDestroy(&petsc_local_to_global0);
-    common::petsc::check(ierr, "ISLocalToGlobalMappingDestroy");
-    ierr = ISLocalToGlobalMappingDestroy(&petsc_local_to_global1);
-    common::petsc::check(ierr, "ISLocalToGlobalMappingDestroy");
+    common::petsc::check(ISLocalToGlobalMappingCreate(
+                             MPI_COMM_SELF, 1, _maps[1].size(), _maps[1].data(),
+                             PETSC_COPY_VALUES, &petsc_local_to_global1),
+                         "ISLocalToGlobalMappingCreate");
+    common::petsc::check(MatSetLocalToGlobalMapping(A, petsc_local_to_global0,
+                                                    petsc_local_to_global1),
+                         "MatSetLocalToGlobalMapping");
+    common::petsc::check(ISLocalToGlobalMappingDestroy(&petsc_local_to_global0),
+                         "ISLocalToGlobalMappingDestroy");
+    common::petsc::check(ISLocalToGlobalMappingDestroy(&petsc_local_to_global1),
+                         "ISLocalToGlobalMappingDestroy");
   }
 
   return A;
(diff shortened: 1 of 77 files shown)


jhale:
This was essentially created by Opus by being told to follow our Python `NonlinearProblem` and the style/intent of the two other PETSc solver wrappers.

I think `NonlinearProblem` should be in `fem::petsc` - this matches the Python interface too.

jhale:
This is ready for review now - might take some small lessons back to the Python version once this is merged.

cpp/dolfinx/fem/petsc.h
```diff
@@ -612,6 +614,195 @@ void set_bc(Vec b,
...
+  CHECK_ERROR("VecGhostRestoreLocalForm");
+}
+
+/// @brief Zero `A`, assemble `a` into it with `bcs` applied, set the
+/// unit diagonal on rows constrained by `bcs`, and finalise assembly.
+/// @param[out] A Matrix to assemble into.
+/// @param[in] a Bilinear form to assemble.
+/// @param[in] bcs Dirichlet boundary conditions.
+template <std::floating_point T>
+void assemble_operator(
```
jorgensd, line 627: This function can replace the verbose part of assembling the matrix in:
https://github.com/FEniCS/dolfinx/blob/8045231002333b34f160aa6c8e15a4257857816f/cpp/demo/poisson/main.cpp#L212-L220

jorgensd: Similar comment for:
https://github.com/FEniCS/dolfinx/blob/8045231002333b34f160aa6c8e15a4257857816f/cpp/demo/biharmonic/main.cpp#L231-L239
and 
https://github.com/FEniCS/dolfinx/blob/8045231002333b34f160aa6c8e15a4257857816f/cpp/demo/mixed_poisson/main.cpp#L317-L327
Of course we should probably keep one of them to show people how to do it manually.

jorgensd (approved):
Only a minor comment about reusing `petsc::assemble_operator` in some of the linear demos. Other than that it looks ok to me.

jhale:
Seems sensible for at least the Poisson C++ demo which is introductory. I'll submit in another PR though.

cpp/dolfinx/fem/petsc.h
jhale: Will do in another PR.

cpp/dolfinx/fem/petsc.h
```diff
@@ -612,6 +614,195 @@ void set_bc(Vec b,
...
+  ierr = VecGhostGetLocalForm(b, &b_local);
+  CHECK_ERROR("VecGhostGetLocalForm");
+  ierr = VecZeroEntries(b_local);
+  CHECK_ERROR("VecZeroEntries");
+  ierr = VecGhostRestoreLocalForm(b, &b_local);
+  CHECK_ERROR("VecGhostRestoreLocalForm");
+
+  assemble_vector(b, F);
+
+  std::vector<std::optional<std::reference_wrapper<const Form<PetscScalar, T>>>>
```
jhale, line 703: Not so nice...

cpp/dolfinx/fem/petsc.h
```diff
@@ -612,6 +614,195 @@ void set_bc(Vec b,
...
+      a{J};
+  std::vector<
```
jhale, line 705: Is incoming `bcs` wrong in the first place?

cpp/dolfinx/fem/petsc.h
```diff
@@ -612,6 +614,195 @@ void set_bc(Vec b,
...
+      std::vector<std::reference_wrapper<const DirichletBC<PetscScalar, T>>>>
+      bcs1{bcs};
+  apply_lifting(b, a, bcs1, std::vector<Vec>{x}, -1);
```
jhale, line 708: Check `std::vector<Vec>{x}` looks odd.

cpp/dolfinx/nls/SNESSolver.cpp
```diff
@@ -0,0 +1,303 @@
...
+  {
+    const char* reason_str;
+    common::petsc::check(SNESGetConvergedReasonString(_snes, &reason_str),
+                         "SNESGetConvergedReasonString");
+    spdlog::warn("PETSc SNES solver did not converge in {} iterations "
+                 "(PETSc reason: {}).",
+                 num_iterations, reason_str);
+  }
+
+  return num_iterations;
```
jhale: My view - we should return the converged reason, not the number of iterations. Same for the KSP wrapper.

jhale, line 213: It's unclear to me the utility of returning this iteration count, same comment for the KrylovSolver.

garth-wells: Fine to change.

garth-wells (approved):

PR #4434, opened 2026-08-21
Remove ADIOS2 build from Dockerfile

garth-wells:


jorgensd:
Is the plan then to use the apt-get install of ADIOS2?
Note that we have promoted ADIOS2 as the preferred format for a few years now (even if that will gradually shift towards VTKHDF once there are more features in there).
Removing it from end-user images is not something I think is a good idea.

jhale:
`libadios2-mpi-c++-dev` worked for me and seems like the minimal choice (there are many adios2 packages!).

jhale:
@jorgensd A full review of the file show ADIOS2 is already added.

jorgensd:
@jhale This change is unfortunate for my checkpointing libraries, as it requires the Python bindings of ADIOS2, which are no longer present in our nightly images.
I can of course ship my own build of DOLFINx with ADIOS2 Python bindings.

jhale:
This isn't an issue; it just needs a tweak to the venv and an install of the Ubuntu adios2 Python bindings.

PR #4438, opened 2026-08-22
Assembly performance optimisations

garth-wells:
## Summary

Performance optimisations for per-cell/facet/entity matrix and vector
assembly, targeting `std::function` indirection for DOF transformations and
geometry/coefficient-fetch overhead. No public API or behavioural changes.

### DOF transformations: skip the closure entirely when none is needed

- `FiniteElement::dof_transformation_fn`/`dof_transformation_right_fn` now
  return `nullptr` instead of a no-op closure when an element needs no DOF
  transformation, rather than allocating and invoking an empty
  `std::function` on every cell.
- All assembly, interpolation and packing call sites guard on this before
  invoking, via a new `fem::is_transform_set()` helper (`traits.h`): it
  checks truthiness for a nullable `std::function`, and always returns
  `true` for a non-nullable callable such as the plain lambda the
  `custom_kernel` demo passes when calling `impl::assemble_*` directly. The
  guard is hoisted out of the per-cell/per-facet loop wherever the
  transform is loop-invariant.
- An earlier version of this PR also added a directly-callable
  (non-type-erased) `DirectDofTransform` path for elements that genuinely
  need a transformation applied at assembly time (H(curl)/H(div) families
  such as N1curl/RT — scalar Lagrange elements of any degree never do,
  since their permutations are resolved once at dofmap-construction time,
  not per cell; see `needs_dof_transformations()`). Benchmarking against
  N1curl/RT mass-matrix assembly (isolating just this path, holding every
  other change constant) showed it only wins for vector assembly of
  degree-1 elements (~4-7%) and is flat-to-negative for matrix assembly at
  any degree and for vector assembly at degree 2 — not worth ~200 extra
  lines in `FiniteElement.h` and 16 call-site conversions, so it was
  dropped. See the Benchmarks section below.

### Geometry/coefficient fetch across cell, facet, and entity integrals

`assemble_cells`/`assemble_cells_matrix`, `assemble_entities`, and
`assemble_interior_facets` (in both `assemble_matrix_impl.h` and
`assemble_vector_impl.h`) fetch each cell's coordinate DOFs via raw-pointer
indexing into the geometry dofmap and coordinate array (avoiding
`md::submdspan` construction overhead in the innermost loop), then copy the
coordinates with `std::copy_n`. The same loop-invariant transform check
(`is_transform_set`, hoisted out of the loop) and `std::copy_n` gather-loop
cleanup is applied to `fem::impl::pack_impl`/`pack_coefficient_entity` in
`pack.h`. An earlier version of the geometry copy used a manual
per-component copy loop instead of `std::copy_n`, which regressed vector
assembly of P1 Lagrange forms by ~7.6%: `perf`-annotated, unstripped
production binaries showed it compiles to three separate 8-byte scalar
loads/stores per vertex, vs. `std::copy_n`'s vectorised 16+8-byte copy for
the same 24 bytes — GCC 13 doesn't give a hand-written copy loop the same
treatment as `std::copy_n`'s specialised small-copy path. Fixed by
restoring `std::copy_n` for the copy itself while keeping the raw-pointer
index computation.

### Compile-time block size for matrix assembly

`impl::assemble_matrix`'s cell-integral, interior-facet, and
exterior-facet/entity dispatch now branches on the test/trial block size
(`bs0`/`bs1`) at compile time for the common `bs == 1` and `bs == 3` cases
(via `std::integral_constant<int, 1/3>`, threaded through new
`DofMapPackCells`/`DofMapPackEntities`/`DofMapPackFacets`-constrained
dofmap parameters), falling back to the runtime `int` block size
otherwise — mirroring the dispatch already used in `impl::assemble_vector`.

### Other fixes found along the way

- A mismatched DOF-transformation guard in `assemble_matrix_impl.h`'s
  interior-facet code (checked one transform but applied the other).
- Two GCC/portability failures not visible to clang locally: an
  `if constexpr` on `entities.rank()` inside a lambda that captures
  `entities` by reference (rejected by GCC as not a constant expression;
  fixed via `remove_cvref_t<decltype(entities)>::rank()`), and the
  sign-compare/template-deduction fallout from the
  `std::integral_constant`-based block-size dispatch above (fixed to use
  `std::integral_constant<int, N>` and explicit `std::tuple{...}` wrapping
  at the `custom_kernel` demo's direct call sites).

## Benchmarks

Serial (1 rank), Release build, unit-cube tetrahedral mesh, median of 15
repetitions, this branch vs. current `main`. Correctness cross-checked on
every run: matrix nnz/Frobenius norm identical between `main` and this
branch in every case below.

**Poisson (`-div(grad(u)) = f`), scalar Lagrange:**

| case | main | branch | Δ |
|---|---|---|---|
| P1 matrix reassemble (N=50, 750k cells) | 112.1 ms | 105.1 ms | **+6.2%** |
| P1 vector assemble | 23.4 ms | 23.4 ms | +0.1% |
| P3 matrix reassemble (N=20, 48k cells) | 403.6 ms | 332.6 ms | **+17.6%** |
| P3 vector assemble | 10.0 ms | 10.0 ms | +0.1% |

**H(curl)/H(div) mass matrix (`inner(u, v)*dx`)** — the element families
that actually exercise the DOF-transformation-guard change, since Lagrange
elements never need a transformation at assembly time:

| case | main | branch | Δ |
|---|---|---|---|
| N1curl deg1 matrix reassemble (N=24, 102k dofs) | 41.2 ms | 39.6 ms | +3.9% |
| N1curl deg1 vector assemble | 12.2 ms | 11.9 ms | +1.9% |
| N1curl deg2 matrix reassemble (N=16, 163k dofs) | 145.0 ms | 135.5 ms | +6.5% |
| N1curl deg2 vector assemble | 5.9 ms | 6.1 ms | −2.0% |
| RT deg1 matrix reassemble (N=24, 169k dofs) | 29.8 ms | 29.1 ms | +2.4% |
| RT deg1 vector assemble | 12.8 ms | 12.6 ms | +1.4% |

Vector-assembly deltas above are all within typical run-to-run noise for
this benchmark harness (±1-2%) except where noted; the matrix-assembly wins
(driven by the geometry-fetch fix and, for P1/P3, the new `bs` dispatch)
are the consistent, reproducible result of this PR. The interior-facet and
exterior-facet/entity paths (DG/discontinuous forms, Neumann/Robin
boundary integrals) receive the same geometry-fetch and `bs`-dispatch
treatment but were not separately re-benchmarked; the numbers above are
representative of the cell-integral case, which dominates assembly cost
for the forms exercised here.

## Testing

- Full local C++ test suite (Catch2, `cpp/test`): 27476 assertions pass (70
  test cases pass, 3 skipped as parallel-only when run serially).
- Local Python `fem` (2384 passed) and `io` (198 passed; ADIOS2-dependent
  tests skip in this environment, unrelated to this change) test suites
  pass.
- `custom_kernel` demo (which calls `impl::assemble_*` directly with a
  non-nullable lambda kernel, exercising the `is_transform_set` fallback
  path) builds and runs correctly, with `Assembler0` (`std::function`-based)
  and `Assembler1` (direct-lambda) assembly producing matching norms.


cpp/dolfinx/fem/FiniteElement.cpp
```diff
@@ -282,7 +283,7 @@ mesh::CellType FiniteElement<T>::cell_type() const noexcept
 }
 //-----------------------------------------------------------------------------
 template <std::floating_point T>
-std::string FiniteElement<T>::signature() const noexcept
+const std::string& FiniteElement<T>::signature() const noexcept
```
jhale, line 286: High potential for dangling reference.

cpp/dolfinx/la/matrix_csr_impl.h
```diff
@@ -35,6 +36,55 @@ concept BlockSizeArg = std::integral<std::remove_cvref_t<T>>
...
+/// @brief Position of `col` within the sorted column-index range
+/// `cols[begin, end)` of one CSR row.
+///
+/// Matches `std::lower_bound(cols.begin() + begin, cols.begin() + end,
+/// col)` in semantics -- returns `end` if `col` is greater than every
+/// entry in the range, otherwise the position of the first entry
+/// `>= col`. Callers must check `cols[pos] == col` themselves to
+/// detect a genuinely missing entry; this function does not throw.
+///
+/// A single shared search used by `insert_csr`, `insert_blocked_csr`,
```
jhale: Results, copy into PR text as comment.

jhale (approved):

PR #4439, opened 2026-08-22
Add weight options to mesh partitioners

chrisrichardson:
Adds a `cell_weights` argument when creating meshes from scratch, which is passed through to the cell partitioner (ParMETIS, SCOTCH, KaHiP etc.)

Currently, there is no defined way to load cell weights from file. I temporarily added a way of doing this for [XDMFFile.cpp](https://github.com/FEniCS/dolfinx/pull/4439/changes#diff-6d996a20b980b667c6520ba09ac82e6e0e6a39dc5305011684a7eeb4742dd4ec), but we need some thought about how it should work in the future, especially for e.g. mixed topology.

Support for `edge_weights`, i.e. the strength of connection between cells, is partially completed, but not wired through yet to top-level. This will be a future PR.

Prepared with help from Claude Sonnet 5.

### DIFF
--- a/cpp/dolfinx/graph/partition.cpp
+++ b/cpp/dolfinx/graph/partition.cpp
@@ -22,12 +22,15 @@ using namespace dolfinx;
 graph::AdjacencyList<std::int32_t>
 graph::partition_graph(MPI_Comm comm, int nparts,
                        const AdjacencyList<std::int64_t>& local_graph,
-                       bool ghosting)
+                       std::span<std::int32_t> node_weights,
+                       std::span<std::int32_t> edge_weights, bool ghosting)
 {
 #if HAS_PARMETIS
-  return graph::parmetis::partitioner()(comm, nparts, local_graph, ghosting);
+  return graph::parmetis::partitioner()(comm, nparts, local_graph, node_weights,
+                                        edge_weights, ghosting);
 #elif HAS_PTSCOTCH
-  return graph::scotch::partitioner()(comm, nparts, local_graph, ghosting);
+  return graph::scotch::partitioner()(comm, nparts, local_graph, node_weights,
+                                      edge_weights, ghosting);
 #elif HAS_KAHIP
   return graph::kahip::partitioner()(comm, nparts, local_graph, ghosting);
 #else
--- a/cpp/dolfinx/graph/partition.cpp
+++ b/cpp/dolfinx/graph/partition.cpp
@@ -22,8 +22,9 @@ using namespace dolfinx;
 graph::AdjacencyList<std::int32_t>
 graph::partition_graph(MPI_Comm comm, int nparts,
                        const AdjacencyList<std::int64_t>& local_graph,
-                       std::span<std::int32_t> node_weights,
-                       std::span<std::int32_t> edge_weights, bool ghosting)
+                       std::span<const std::int32_t> node_weights,
+                       std::span<const std::int32_t> edge_weights,
+                       bool ghosting)
 {
 #if HAS_PARMETIS
   return graph::parmetis::partitioner()(comm, nparts, local_graph, node_weights,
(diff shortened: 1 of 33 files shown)


jorgensd:
> Adds a `cell_weights` argument when creating meshes from scratch, which is passed through to the cell partitioner (ParMETIS, SCOTCH, KaHiP etc.)
> 
> 
> 
> Currently, there is no defined way to load cell weights from file. I temporarily added a way of doing this for [XDMFFile.cpp](https://github.com/FEniCS/dolfinx/pull/4439/changes#diff-6d996a20b980b667c6520ba09ac82e6e0e6a39dc5305011684a7eeb4742dd4ec), but we need some thought about how it should work in the future, especially for e.g. mixed topology.
> 
> 
> 
> Support for `edge_weights`, i.e. the strength of connection between cells, is partially completed, but not wired through yet to top-level. This will be a future PR.
> 
> 
> 
> Prepared with help from Claude Sonnet 5.

So for cell_weights, you want to have a function that just reads in a flat array of some dtype with the same striding as the reading of rows of the connectivity array?

chrisrichardson:
> > Adds a `cell_weights` argument when creating meshes from scratch, which is passed through to the cell partitioner (ParMETIS, SCOTCH, KaHiP etc.)
> > Currently, there is no defined way to load cell weights from file. I temporarily added a way of doing this for [XDMFFile.cpp](https://github.com/FEniCS/dolfinx/pull/4439/changes#diff-6d996a20b980b667c6520ba09ac82e6e0e6a39dc5305011684a7eeb4742dd4ec), but we need some thought about how it should work in the future, especially for e.g. mixed topology.
> > Support for `edge_weights`, i.e. the strength of connection between cells, is partially completed, but not wired through yet to top-level. This will be a future PR.
> > Prepared with help from Claude Sonnet 5.
> 
> So for cell_weights, you want to have a function that just reads in a flat array of some dtype with the same striding as the reading of rows of the connectivity array?

Yes. Something like that. But it needs to be before or during reading the mesh itself.

jorgensd:
> > > Adds a `cell_weights` argument when creating meshes from scratch, which is passed through to the cell partitioner (ParMETIS, SCOTCH, KaHiP etc.)
> > > Currently, there is no defined way to load cell weights from file. I temporarily added a way of doing this for [XDMFFile.cpp](https://github.com/FEniCS/dolfinx/pull/4439/changes#diff-6d996a20b980b667c6520ba09ac82e6e0e6a39dc5305011684a7eeb4742dd4ec), but we need some thought about how it should work in the future, especially for e.g. mixed topology.
> > > Support for `edge_weights`, i.e. the strength of connection between cells, is partially completed, but not wired through yet to top-level. This will be a future PR.
> > > Prepared with help from Claude Sonnet 5.
> > 
> > 
> > So for cell_weights, you want to have a function that just reads in a flat array of some dtype with the same striding as the reading of rows of the connectivity array?
> 
> Yes. Something like that. But it needs to be before or during reading the mesh itself.

That isn't a problem though, as it should only need to know the global number of cells to get the striding. I use something similar in: https://github.com/scientificcomputing/io4dolfinx/blob/d82ad2ad75fb32d741aaed4e72d9f606e45af311/src/io4dolfinx/backends/xdmf/backend.py#L455-L542
It should more or less call the same as read_cell_tags, but instead of calling distribute entity data, it should just return that data. I guess the general function should be `read_partitioned_data(name_of_array)`, which would do the compute_local_range striding across the first axis of the data?

chrisrichardson:
@jorgensd - it is not a problem of how to read the data, more a question of how to add it to XDMF (or other formats). Or should it be provided by the user as extra data? e.g. the user obtains it somehow beforehand, from another file or something, and it is then an extra input to `read_mesh()`...

jorgensd:
> @jorgensd - it is not a problem of how to read the data, more a question of how to add it to XDMF (or other formats). Or should it be provided by the user as extra data? e.g. the user obtains it somehow beforehand, from another file or something, and it is then an extra input to `read_mesh()`...

We already have `xdmf_utils::get_dataset` that reads the data in this strided fashion.
Exposing this would make your life easier: https://github.com/FEniCS/dolfinx/blob/52a37cb90e6285a0709fb6130f73ea01d41c9127/cpp/dolfinx/io/xdmf_utils.h#L147-L285

jhale (commented):
I would remove the io changes and split that out into another PR, if they are really needed. Rest of PR looks good.

PR #4440, opened 2026-08-22
Build gmsh headless, OpenCASCADE from source, and trim dev-env bloat

jhale:
## Summary

Bottom line: 4GB (!) to 2GB.

- Build gmsh with `-DENABLE_FLTK=0`, dropping its FLTK/OpenGL/X11 GUI
  dependencies (`libglu1`, `libxcursor-dev`, `libxft2`, `libxinerama1`,
  `libfltk1.3-dev`, `libfreetype6-dev`, `libgl1-mesa-dev`) since none of
  it is needed for headless meshing.
- Build OpenCASCADE (OCCT), gmsh's CAD kernel, from source instead of via
  apt: Ubuntu's `libocct-data-exchange-dev` unconditionally pulls in the
  full `libvtk9-dev` toolchain (GDAL, MySQL/Postgres client libs, ffmpeg,
  `default-jdk`, X11/GLU, Boost, ...), none of which the headless build
  uses. Only `FoundationClasses`/`ModelingData`/`ModelingAlgorithms`/
  `DataExchange` are enabled.
- Add `--no-install-recommends` to apt installs that were missing it
  (~128 fewer Recommends packages: fonts, gdb, openssh-client, etc.).
- Swap `libgl1-mesa-dev` + `mesa-utils` for runtime-only `libgl1` +
  `libosmesa6` in the `lab` image: nothing in that stage compiles
  against OpenGL (vtk/pyvista are pip wheels), but pyvista's headless
  render window needs OSMesa's software backend or it segfaults (there's
  no X server/EGL in the container).
- Drop the unused third PETSc/SLEPc build configuration
  (`linux-gnu-real64-64`, 64-bit indices) — DOLFINx's own build never
  targets it, and it was tripling the footprint of HYPRE, MUMPS,
  SuiteSparse, SuperLU, ScaLAPACK and (PT)SCOTCH for no benefit.
- Drop `libeigen3-dev` and `valgrind`: no C++ source in this repo uses
  Eigen (DOLFINx moved off it years ago), and valgrind isn't invoked by
  any test/CI script here.
- Fix two build regressions introduced by the `--no-install-recommends`
  tightening above: add back `make` (silently dropped, broke PETSc's
  configure step) and pin `cython<3.3` (3.3 tightened pointer-indexing
  type checks that break petsc4py 3.25.4's generated `PC.pyx`).
- Update the CI matrix (`ccpp.yml`) to match the dropped PETSc
  configuration.

Net effect: `dev-env` shrinks from ~4.0 GB to ~2.0 GB (arm64).

## Test plan

- [x] Built `dev-env` locally with `podman build --format docker`;
      verified both remaining PETSc/SLEPc configurations
      (`linux-gnu-real64-32`, `linux-gnu-complex128-32`) import and
      report the correct scalar type.
- [x] Verified gmsh imports, initializes, and meshes an OCC box via its
      Python API.
- [x] Built the full `dolfinx-onbuild` → `dolfinx` → `lab` chain against
      the trimmed `dev-env`.
- [x] Ran the full `python/demo` suite (25 demos) against the built
      `lab` image: 24/25 pass; the one failure (`demo_comm-pattern.py`,
      missing `networkx`) is a pre-existing gap unrelated to this PR.
- [x] Confirmed `demo_pyvista.py` renders successfully (only benign
      EGL/Xcursor warnings, handled by the OSMesa fallback).

### DIFF
--- a/docker/Dockerfile.test-env
+++ b/docker/Dockerfile.test-env
@@ -63,12 +63,9 @@ ENV OPENBLAS_NUM_THREADS=1 \
 # - First set of packages are required to build and run FEniCS.
 # - Second set of packages are recommended and/or required to build
 #   documentation or tests.
-# NOTE: OpenCASCADE (OCCT), the CAD kernel used by gmsh, is built from
-# source below rather than installed via apt: Ubuntu's
-# libocct-data-exchange-dev unconditionally pulls in the full
-# libvtk9-dev toolchain, which gmsh's headless build doesn't need.
-# libx11-dev below is OCCT's own dependency (Xlib.h headers only, no
-# runtime GUI), not related to gmsh's GUI, which stays disabled.
+# OCCT (gmsh's CAD kernel) is built from source below since Ubuntu's
+# libocct-data-exchange-dev drags in the full libvtk9-dev toolchain.
+# libx11-dev is OCCT's own header-only dependency, unrelated to gmsh's GUI.
 RUN export DEBIAN_FRONTEND=noninteractive && \
     apt-get -qq update && \
     apt-get -yq --with-new-pkgs -o Dpkg::Options::="--force-confold" upgrade && \
@@ -128,9 +125,8 @@ RUN python3 -m venv ${VIRTUAL_ENV}
 
 # Install Python packages (via pip)
 RUN pip install --no-cache-dir --upgrade pip setuptools wheel && \
-    # Cython 3.3 tightened pointer-indexing type checks in a way that
-    # breaks petsc4py 3.25.4's generated Cython source (PC.pyx); pin
-    # below that until petsc4py catches up.
+    # Cython 3.3 tightened pointer-indexing checks that break
+    # petsc4py 3.25.4's generated PC.pyx; pin below that for now.
     pip install --no-cache-dir "cython<3.3" numpy==${NUMPY_VERSION} && \
     pip install --no-cache-dir mpi4py
 
@@ -142,17 +138,9 @@ RUN wget -nc --quiet https://github.com/kahip/kahip/archive/v${KAHIP_VERSION}.ta
     cmake --install build-dir && \
     rm -rf /tmp/*
 
-# Install OpenCASCADE (OCCT), the CAD kernel used by gmsh, from source,
-# headless: only the FoundationClasses/ModelingData/ModelingAlgorithms/
-# DataExchange modules gmsh needs for STEP/IGES/meshing are built;
-# Visualization, ApplicationFramework and Draw are disabled, and
-# FreeType is turned off explicitly (DataExchange still probes for it
-# otherwise). TKService, a DataExchange dependency, still needs
-# Xlib.h (libx11-dev) even with Visualization off, but nothing beyond
-# the header is used headless. Installs headers under
-# include/opencascade to match gmsh's find_path() search. ldconfig is
-# required since a manual cmake --install doesn't trigger the postinst
-# hook that updates the linker cache for an apt package.
+# Headless OCCT: only the modules gmsh's STEP/IGES/meshing needs are
+# built. FreeType is force-disabled since DataExchange still probes for
+# it with Visualization off. ldconfig covers what a manual install skips.
 RUN git clone -b V${OCCT_VERSION} --single-branch --depth 1 https://github.com/Open-Cascade-SAS/OCCT.git && \
     cmake -G Ninja -DCMAKE_BUILD_TYPE=Release \
     -DBUILD_LIBRARY_TYPE=Shared \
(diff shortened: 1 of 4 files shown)


jhale:
@jorgensd A comment on new gmsh build suitability would be appreciated, thanks!

jorgensd:
> @jorgensd A comment on new gmsh build suitability would be appreciated, thanks!

I'd need to test it. I know there are users that call `gmsh.fltk.run()` to launch the GUI even when using docker (as you can easily et up GUI support). Similarly for pyvista, people don't run headless (especially not in the lab environments).

garth-wells (approved):

docker/Dockerfile.test-env
```diff
@@ -76,64 +76,33 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
...
-    rm -rf /tmp/*
-
 ENV VIRTUAL_ENV=/dolfinx-env
 ENV PATH=/dolfinx-env/bin:$PATH
 RUN python3 -m venv ${VIRTUAL_ENV}
 
 # Install Python packages (via pip)
 RUN pip install --no-cache-dir --upgrade pip setuptools wheel && \
-    pip install --no-cache-dir cython numpy==${NUMPY_VERSION} && \
+    # Cython 3.3 tightened pointer-indexing checks that break
```
francesco-ballarin, line 107: maybe let's open an issue to remind ourselves to unpin it when PETSc 3.26 gets released, otherwise I am afraid that this pin will be forever forgotten ;)

jhale:
> > @jorgensd A comment on new gmsh build suitability would be appreciated, thanks!
> 
> 
> 
> I'd need to test it. I know there are users that call `gmsh.fltk.run()` to launch the GUI even when using docker (as you can easily et up GUI support). Similarly for pyvista, people don't run headless (especially not in the lab environments).

Docker is a bad tool for GUI stuff, and x11 even worse for forwarding 3D... I always have gmsh installed on my system.

That said, I'll rebuild with fltk and see if there is a substantial difference in size.

jhale:
> **Note:** this comment is entirely LLM-generated content (Claude Opus 5, via Claude Code) — the measurements below were run by the model and the text drafted by it.

@jorgensd Measured the cost of putting the FLTK GUI back. Short version: **~70 MiB on `dev-env`, ~3.5% of the 2 GB image.** Happy to re-enable it.

**apt closure** (ubuntu:26.04, `--no-install-recommends`, `Installed-Size` sums; BASE = this PR's `dev-env` set, 320 pkgs / 1277 MiB):

| variant | pkgs | Δ |
|---|---|---|
| `libfltk1.3-dev` alone | 355 | +64 MiB |
| main's 7-package FLTK set | 376 | +72 MiB |
| `libfltk1.4-dev` | 403 | +101 MiB |

**gmsh itself** (4.15.2, `ENABLE_OCC=0`, dynamic, aarch64): `libgmsh.so` 22.08 → 23.97 MiB, i.e. **+1.9 MiB**. The GUI code is nearly free; the cost is all apt, and 47 MiB of it is a single package, `mesa-libgallium`.

**Proposal — add just two packages to the existing first apt block, and `-DENABLE_FLTK=1`:**

```
libfltk1.3-dev
libglu1-mesa-dev
```

- `libfltk1.3-dev`'s closure already provides `libfltk-gl`, `libgl1`, `libglx-mesa0`, `libxcursor1`, `libxft2`, `libxinerama1`, xcb, fontconfig/freetype/dejavu fonts. Main's other six packages were redundant (+8 MiB for nothing).
- `libglu1-mesa-dev` is *not* in that closure and gmsh's GUI links `libGLU.so.1` — the build fails to configure without it.
- `ldd` on the FLTK build shows 27 extra libs, all satisfied by those two. Nothing further is needed for a forwarded X11 session.
- Stay on 1.3; 1.4 costs ~37 MiB more for no benefit.

Everything downstream is `FROM dev-env`, so this is one line and the runtime libs propagate. For `lab` specifically the marginal cost is only **~11 MiB**, since `libosmesa6` already pulls `mesa-libgallium` there — so `gmsh.fltk.run()` in the lab image is essentially free.

jhale:
> **Note:** this comment is entirely LLM-generated content (Claude Opus 5, via Claude Code) — the measurements below were run by the model and the text drafted by it.

Followed up with real `dev-env` builds both ways (`podman build --format docker`, aarch64), from two Dockerfiles differing only in the FLTK lines:

| image | size | Δ |
|---|---|---|
| `ENABLE_FLTK=0` | 2,060,353,405 B (1965 MiB) | — |
| `ENABLE_FLTK=1` + `libfltk1.3-dev` `libglu1-mesa-dev` | 2,129,311,653 B (2031 MiB) | **+65.8 MiB (+3.3%)** |

So the estimate above (~70 MiB) was close; the measured cost is **65.8 MiB**. Both builds exit 0, gmsh reports `Fltk` in its build options, `libgmsh.so` links FLTK/GLU with no unresolved libs, `gmsh.fltk.run` is callable, and OCC meshing is unaffected (1096 tets on a unit box). Not tested: an actual forwarded X11 session, which needs a real `DISPLAY`.

Pushing the change now — `dev-env` lands at ~2.13 GB instead of ~2.06 GB, versus 4 GB before this PR.

jhale:
Merged into #4449

PR #4442, opened 2026-08-22
Simplify la::petsc/fem::petsc/la::slepc error handling and remove unused Operator base class

jhale:
## Summary

- Replace three independent copies of the `CHECK_ERROR` macro (in `la/petsc.{h,cpp}`, `fem/petsc.h`, `la/slepc.cpp`) with a single `check()`/`error()` function pair, usable from header-inline templates/lambdas where a macro couldn't be. `error()`/`check()` take a `std::source_location` defaulted to the call site instead of a caller-supplied `__FILE__`, so messages report the exact call-site line.
- Move `error()`/`check()` out of `la::petsc` into a new `dolfinx::common::petsc` namespace, since they are used well beyond `la` — by `la::slepc`, `fem::petsc`, `nls::NewtonSolver`, and the petsc4py wrapper — and don't depend on any `la`-specific type.
- Convert every `PetscErrorCode ierr = PetscCall(...); check(ierr, "PetscCall");` two-line pattern to a single inline `check(PetscCall(...), "PetscCall")` call, across `la::petsc`, `la::slepc`, `fem::petsc`, and the petsc4py wrapper, dropping the intermediate `ierr` variable wherever it wasn't needed afterward.
- Add `assert(error_code)` to `common::petsc::error()`, documenting that it is only ever reached from `check()` after a nonzero code.
- Move the PETSc options-database helpers (previously `la::petsc::options::set`/`clear`) into `common::petsc` for the same reason — already called from `nls::NewtonSolver` and several demos, not `la`-specific — flattening `options::set`/`options::clear` into plain `common::petsc::set_option`/`set_option<T>`/`clear_option`/`clear_options` functions rather than keeping a third-level namespace, matching the rest of the library's `dolfinx::<module>::<submodule>` convention.
- Fix three `Matrix` option-prefix methods (`set_options_prefix`, `get_options_prefix`, `set_from_options`) that were previously calling PETSc functions without checking the returned error code at all.
- Remove `la::petsc::Operator`, merging its members (`mat()`, `size()`, `create_vector(dim)`, `Mat` lifetime/ref-counting) directly into `la::petsc::Matrix`. `Matrix` was `Operator`'s only subclass and nothing in the codebase used `Operator` polymorphically, so the base class was pure indirection.
- Remove `Matrix::apply(AssemblyType)` and `Matrix::norm(Norm)`, confirmed unused everywhere in the codebase (demos and the Python layer call `MatAssemblyBegin`/`End` and `MatNorm` directly on the raw `Mat`).

## Why

`la::petsc` and its siblings had accumulated three independent, near-identical copies of the same error-checking macro; error-handling and options-database code that lived under `la::petsc` purely for historical reasons despite being used by `nls`, `fem`, `slepc`, and demos; an inconsistency in which methods actually checked PETSc return codes; and a base class with no remaining reason to exist. None of this changes observable behaviour except that a few previously-silent PETSc failures now correctly throw.

## Test plan

- [x] `cpp/test/petsc.cpp` and `cpp/test/slepc.cpp` pass (`Operator` reference-counting test sections renamed to exercise `Matrix` instead)
- [x] Full C++ library builds clean with `-Werror`; `ninja install` and the Python nanobind wrapper rebuild clean
- [x] All four affected demos (`poisson`, `biharmonic`, `hyperelasticity`, `mixed_poisson`) build and link clean
- [x] `clang-format --dry-run --Werror` clean on all touched files; `gersemi --check` clean on touched `CMakeLists.txt`

AI assistance: I used Claude to draft parts of this PR. I reviewed, edited, tested, and take responsibility for the final contribution.

### DIFF
--- a/cpp/dolfinx/la/petsc.cpp
+++ b/cpp/dolfinx/la/petsc.cpp
@@ -33,15 +33,13 @@ la::petsc::create_vectors(MPI_Comm comm,
   std::vector<Vec> v(x.size());
   for (std::size_t i = 0; i < v.size(); ++i)
   {
-    PetscErrorCode ierr;
-    ierr = VecCreateMPI(comm, x[i].size(), PETSC_DETERMINE, &v[i]);
-    common::petsc::check(ierr, "VecCreateMPI");
+    common::petsc::check(
+        VecCreateMPI(comm, x[i].size(), PETSC_DETERMINE, &v[i]),
+        "VecCreateMPI");
     PetscScalar* data;
-    ierr = VecGetArray(v[i], &data);
-    common::petsc::check(ierr, "VecGetArray");
+    common::petsc::check(VecGetArray(v[i], &data), "VecGetArray");
     std::ranges::copy(x[i], data);
-    ierr = VecRestoreArray(v[i], &data);
-    common::petsc::check(ierr, "VecRestoreArray");
+    common::petsc::check(VecRestoreArray(v[i], &data), "VecRestoreArray");
   }
 
   return v;
@@ -56,8 +54,6 @@ Vec la::petsc::create_vector(const common::IndexMap& map, int bs)
 Vec la::petsc::create_vector(MPI_Comm comm, std::array<std::int64_t, 2> range,
                              std::span<const std::int64_t> ghosts, int bs)
 {
-  PetscErrorCode ierr;
-
   // Get local size
   assert(range[1] >= range[0]);
   std::int32_t local_size = range[1] - range[0];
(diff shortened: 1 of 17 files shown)


cpp/dolfinx/la/petsc.cpp
```diff
@@ -27,23 +26,16 @@ using namespace dolfinx;
...
-      petsc::error(ierr, __FILE__, NAME);                                      \
-  } while (0)
-
-//-----------------------------------------------------------------------------
-void la::petsc::error(PetscErrorCode error_code, std::string_view filename,
-                      std::string_view petsc_function)
+void la::petsc::error(PetscErrorCode error_code,
+                      std::string_view petsc_function, std::source_location loc)
 {
   // Fetch PETSc error description
```
schnellerhase:
```suggestion
  if (!error_code)
    return;
  // Fetch PETSc error description
```

cpp/dolfinx/la/petsc.h
schnellerhase, line 1: Don't like that the error handling code resides in namespace `la`. How about `common::petsc`?

schnellerhase (commented):
Whats the benefit of passing the function name to the `check`? Are we sure this adds any additional information to the log that PETSc does not already provide? Dropping it would put us to the desirable `petsc::check(PetscF(...))` code style.

cpp/dolfinx/la/petsc.cpp
schnellerhase: Or better `assert`?

jhale:
> Whats the benefit of passing the function name to the `check`? Are we sure this adds any additional information to the log that PETSc does not already provide? Dropping it would put us to the desirable `petsc::check(PetscF(...))` code style.

Only PETSc's internally created errors have the nice info + trace. That was the entire point of the original macro CHECK_ERROR which this modernises.

schnellerhase:
With the (great) change to `std::source_location::current()`, the output now reads for example:
```bash
1: [2026-08-23 13:43:54.417] [error] PETSc error in '⬛⬛⬛/dolfinx/cpp/dolfinx/la/slepc.cpp:121', 'EPSGetEigenpair'
1: [2026-08-23 13:43:54.417] [error] PETSc error code '63' 'Argument out of range'
```

If we stick to my previously outlined style the provided line will contain already the respective PETSc function that will have caused the error - therefore it would remove the additional verbose passing of the name and implies the same diagnostic.

jhale:
The user would still need to lookup the source code to know the name of the function - it's pretty handy to have the name of the call.

jhale:
Had a look at this - we should enable '--with-strict-petscerrorcode' in PETSc and always swallow the error code directly into our new 'check' function. This will avoid captured but not checked mistakes.

PR #4446, opened 2026-08-25
Avoid undefined behaviour in kernel-invocation coefficient pointer for empty coefficient arrays

jhale:
`&coeffs(idx, 0[, 0])` dereferences before taking the address, which is UB when `coeffs` is empty (a form with no `Function` coefficients, e.g. plain Poisson) — not an edge case, since it fires on ordinary `fem::assemble_matrix`. Caught by UBSan (`-fsanitize=undefined`).

This is a hot loop (once per cell/facet per assembly), so the fix was checked for overhead rather than assumed free. `submdspan(coeffs, idx, md::full_extent).data_handle()` is UB-free but not free: at `-O2` its generic offset computation adds a compare + conditional select per call versus the original. Instead, the base pointer and per-cell/per-facet stride are hoisted out of each loop once, and the pointer is computed with plain arithmetic: `coeffs.data_handle() + idx * coeffs.extent(1)` (and the two-term form for interior-facet integrals, where `coeffs` is indexed `(f, side, cstride)`). This is well-defined even when `coeffs` is empty — `cstride` is then `0`, so every offset collapses to `nullptr + 0`, which the standard explicitly permits (unlike a nonzero offset on a null pointer) — and compiles to *less* code than the original `&coeffs(idx, 0)`, since the per-iteration multiply is also eliminated.

Verified under `-fsanitize=address,undefined`: `cpp/test/` passes at np=1/2/3 (previously aborted on the first assembly test), and `python/test/unit` passes in full (3486 passed).

**Why this was never observed at runtime:** the loaded value from `&coeffs(idx, 0)` is discarded — only its address is used — so any optimizing compiler reduces it to plain pointer arithmetic with no actual memory access, identical to the fix in this PR. It's UBSan's job specifically to flag the abstract-machine violation regardless of generated-code behavior; no build type here previously enabled it.

### DIFF
--- a/cpp/dolfinx/fem/assemble_expression_impl.h
+++ b/cpp/dolfinx/fem/assemble_expression_impl.h
@@ -82,9 +82,6 @@ void tabulate_expression(
   std::vector<T> values_local(size0 * num_argument_dofs, 0);
   std::size_t offset = values_local.size();
 
-  // Base pointer and per-entity stride into coeffs, computed once rather
-  // than per entity; coeffs_data + e * cstride is well-defined even when
-  // coeffs is empty (cstride == 0), unlike &coeffs(e, 0).
   const T* coeffs_data = coeffs.data_handle();
   const std::size_t cstride = coeffs.extent(1);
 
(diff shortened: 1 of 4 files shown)


jhale:
Pleased to say, this was the only bug surfaced using UBSan and AddressSan!

chrisrichardson (approved):

PR #4448, opened 2026-08-26
Make SNES/KSP/EPS solve() return [[nodiscard]] convergence reason

jhale:
This is a breaking API change.

SNESSolver::solve, KrylovSolver::solve, and SLEPcEigenSolver::solve now return a [[nodiscard]] PETSc convergence reason instead of silently letting non-convergence go unchecked. Iteration counts are no longer returned; callers who need them fetch them via the raw PETSc API (SNESGetIterationNumber/KSPGetIterationNumber), as shown in the updated demos. All callers in the library, tests, and demos have been updated accordingly.


cpp/dolfinx/nls/NewtonSolver.cpp
```diff
@@ -207,8 +209,14 @@ std::pair<int, bool> nls::petsc::NewtonSolver::solve(Vec x)
     if (_fnP)
       _fnP(x, _matP);
 
-    // Perform linear solve and update total number of Krylov iterations
-    _krylov_iterations += _solver.solve(_dx, _b);
+    // Perform linear solve
+    static_cast<void>(_solver.solve(_dx, _b));
```
jhale, line 213: Eep. Glad we're getting rid of this!

garth-wells (approved):

PR #4449, opened 2026-08-27
Revise Docker development image

jhale:
## Changes

**`--system-site-packages` venv.** The venv is created with system site
packages visible, so apt's Python modules are importable from `/dolfinx-env`.

**ADIOS2 Python bindings from apt.** `python3-adios2` carries the Python API
and `python3-adios2-mpi` only the compiled `.so`, so both are named; without
the second, apt satisfies the first's serial-or-MPI dependency with the serial
build. `ADIOS2_ALWAYS_USE_MPI=1` stops `adios2/bindings/__init__.py` falling
back to serial bindings that are not installed.

**NumPy, mpi4py, cffi, scipy, pytest, pytest-xdist and packaging from apt.**
These either have to match a system library ABI or are stable enough that
Ubuntu's version is fine, and CI installs them on every job today.
`python3-numpy-dev` is included because it holds the headers that
`numpy.get_include()` resolves to. `ARG NUMPY_VERSION` and its
Numba-compatibility note are dropped, since NumPy is now Ubuntu's 2.3.5.

**matplotlib, networkx, numba and pyamg from pip, in dev-env.** These move
faster than an LTS release, and Ubuntu's matplotlib is also the larger option:
`python3-fonttools` depends on `sympy` and `unicode-data`, making the apt
closure ~151 MiB against ~100 MiB of wheels.

**The build group is read from `python/pyproject.toml`.** The Dockerfile no
longer duplicates `[build-system].requires`; it copies the file in and uses
pip's `--group`. That requires the repository root as build context, so
`docker-dev-test-env.yml` changes `context: docker/` to `context: .`, and the
`COPY` is the last layer so editing `pyproject.toml` does not invalidate the
PETSc/SLEPc layers above.

**An artefact-only `.dockerignore`.** A dirty working tree would otherwise ship
its build output — 1146 MB here against 4 MB of tracked files, now 234 MB. Only
artefacts are excluded, and only extensions that `.gitignore` also ignores, so
no tracked file can be dropped; `.git` is kept because `cpp/CMakeLists.txt`
reads the commit hash into `DOLFINX_GIT_COMMIT_HASH`.

**`HDF5_ROOT` pinned to the MPI build.** CMake's `FindHDF5` picks Ubuntu's
serial HDF5 whenever both flavours are installed, and the `hdf5.pc` alternative
does not prevent it. DOLFINx sets `HDF5_PREFER_PARALLEL` for its own build and
in its installed config, but that only covers projects reaching
`find_package(DOLFINX)` first; pinning `HDF5_ROOT` fixes it for any consumer.

**`Dockerfile.end-user` shrinks.** The onbuild stage installs no packages at
all now — no build group, no `pyamg pytest scipy matplotlib numba` — it only
compiles. The `lab` stage's separate `pip install matplotlib` is redundant
against the base and is removed.

Resulting versions: NumPy 2.3.5, mpi4py 4.1.1 (Open MPI 5.0.10), ADIOS2 2.11.0,
scipy 1.16.3, pytest 9.0.2, matplotlib 3.11.1, numba 0.67.0, Python 3.14.4. The
image is 2.56 GB, up from ~2.09 GB for the ADIOS2 change alone; the difference
is numba, matplotlib and the build tooling moving into the base from every
downstream build and CI job.

## Testing

Built with `podman 5.7.0` on aarch64, `--target dev-env`, from the repository
root. In the resulting image:

- ADIOS2 resolves to `adios2_bindings_mpi...so` serially and under `mpirun`,
  and a 2-rank `.bp` write/read round-trips.
- petsc4py 3.25.4 imports in both arches and slepc4py 3.25.1 imports; both
  still build from source against the system NumPy/mpi4py.
- basix, UFL, FFCx and DOLFINx build with no pip install of build or test
  dependencies, mirroring the emptied onbuild stage and ending in
  `--no-build-isolation --check-build-dependencies`. `test_adios2.py` passes 42
  tests and the `static-condensation`, `lagrange-variants` and `tnt-elements`
  demos pass, exercising numba and matplotlib.
- With serial `libhdf5-dev` installed alongside, a bare `find_package(HDF5)`
  reports `IS_PARALLEL=TRUE` and `/usr/lib/hdf5-mpi/include`.
- A `COPY .` probe diffed against `git ls-files` confirms `.dockerignore` drops
  no tracked file.

`docker-dev-test-env.yml` passed on this branch — all four builds (test-env and
dev-env, amd64 and arm64) and the multiarch manifest job — covering amd64 and
the `context:` change. The nightly `docker-end-user.yml` on top of this base is
still worth running before merge, as the onbuild stage is only exercised
locally in imitation, not in its ONBUILD form.

---


garth-wells:
If we take too much, e.g. numba, from the package manager I think the image will go stale very quickly. Some packages move much faster than Ubuntu LTS releases.

docker/Dockerfile.test-env
```diff
@@ -85,8 +83,21 @@ RUN export DEBIAN_FRONTEND=noninteractive && \
     mpi-default-dev \
     ninja-build \
     pkg-config \
+    python3-adios2 \
```
jorgensd, line 87: Its suprising one needs both

jhale: _This reply was drafted by Claude Code (Opus 5); I have checked it and take responsibility for it._

It surprised me too, but the package split is inverted from what the names suggest: `python3-adios2-mpi` is bindings-only. In a clean `ubuntu:26.04` it ships exactly one file under `dist-packages`:

```
/usr/lib/python3/dist-packages/adios2/bindings/adios2_bindings_mpi.cpython-314-aarch64-linux-gnu.so
```

No `__init__.py` anywhere. Installing it alone does not give an `ImportError` — Python treats the directory as an implicit namespace package, so `import adios2` "succeeds" and you get a hollow module:

```
namespace pkg, __file__ = None
has Stream: False
```

`python3-adios2` is the front end that carries the actual Python API: 13 `.py` files, including `__init__.py`, `stream.py`, `engine.py`, and `bindings/__init__.py`, the selector that chooses between the serial and MPI `.so` at import time. Adding it flips the same check to `has Stream: True 2.11.0`.

The dependency directions match that: `python3-adios2` declares `Depends: python3-adios2-serial | python3-adios2-mpi`, while `python3-adios2-mpi` only `Recommends: python3-adios2`, which `--no-install-recommends` discards. So both names are needed — `python3-adios2` for the API, and `python3-adios2-mpi` spelled out so apt does not satisfy the alternative with the serial build (and pull a second, serial ADIOS2 stack into an image whose C++ half is MPI-only).

I will expand the comment in the Dockerfile, which currently explains only why the MPI variant is named explicitly and not why both are listed.

docker/Dockerfile.test-env
```diff
@@ -144,14 +132,39 @@ RUN wget -nc --quiet https://github.com/kahip/kahip/archive/v${KAHIP_VERSION}.ta
     cmake --install build-dir && \
     rm -rf /tmp/*
 
-# Install GMSH
+# Headless OCCT: only the modules gmsh's STEP/IGES/meshing needs are
```
michalhabera, line 135: GMSH/OCCT changes are unrelated here. And we need some of the modules, for STEP files with boundary tagging. Please make that a separate PR so it is easier to track countless issues that will appear.

michalhabera: @jhale ?

jhale: It's no problem, I will fix in another PR.

Issue #4450, opened 2026-08-27
[BUG]: Wrong collision detected in `dolfinx.geometry.determine_point_ownership`


MathPhy42:
### Summarize the issue

Hi,

 I found a bug using `dolfinx.geometry.determine_point_ownership`, since a collision is detected for a point that lies outside the mesh. I think this is possible, if a point lies outside the mesh, but still inside the bounding box of the cell. 

### How to reproduce the bug

I ran the following code below using `dolfinx 0.11.0` on a single MPI-Process. It creates a mesh consisting of a single tetrahedron and check collision of 3 different points using `dolfinx.geometry.determine_point_ownership`. The first point is contained inside the tetrahedron, 
the second point lies outside the tetrahedron but inside its bounding box and the third point lies outside this bounding box. 
As expected a collision is found for the first point and no collision is found for the third point, but for the second point there's also a collision detected, which is wrong. See output messages below.
I believe this is a bug in the implementation of `dolfinx.geometry.determine_point_ownership` (probably on the C++-side), where the collisions with bounding boxes were checked correctly, but not the actual collision with the cells.

### Minimal Example (Python)

```Python
from mpi4py import MPI
import numpy as np
import ufl
import basix

import dolfinx
import pyvista as pv

comm = MPI.COMM_WORLD
mpi_size = comm.size

# Define mesh consiting of a single tetrahedron
rank = comm.rank
points = np.array([
  [0.0, 0.0, 0.0],
  [1.0, 0.0, 0.0],
  [0.0, 1.0, 0.0],
  [0.0, 0.0, 1.0]
], dtype=np.float64)

cells = np.array([[0, 1, 2, 3]], dtype=np.int64)

coord_element = basix.ufl.element("Lagrange", "tetrahedron", 1, shape=(3,))
ufl_mesh = ufl.Mesh(coord_element)
mesh = dolfinx.mesh.create_mesh(comm, cells=cells, e=ufl_mesh, x=points)
tdim = mesh.topology.dim

# Define points, for which collisions with the mesh will be detected
if comm.rank == 0:
  points_ext = np.array([
    [0.25, 0.25, 0.25], # Point inside the mesh
    [0.5, 0.5, 0.5], # Point outside the mesh, but inside the bounding box
    [1.5, 0.5, 0.5], # Point outside the bounding box
  ], dtype=np.float64)
else:
  points_ext = np.empty((0, 3), dtype=np.float64)

# Check, which rank contains the point, if any does
[...]

PR #4453, opened 2026-08-28
Add option to allow/disallow extrapolation in PointCollisionData.

jorgensd:
Addresses #4450.

Makes it optional to perform extrapolation within cell bounding box for closest point detection (and in turn non-matching interpolation).

Default behavior of allowing extrapolation is kept to avoid breaking user code.

C++ changes have been co-authored with Claude. Test is based on code from @MathPhy42. I have reviewed the contribution.

## Other fixes
- **Buffer-hoisting**: In two places in `determine_point_ownership` there were arrays containing local coordinate_dofs geometry data being made in hot (and very hot) loops. Resolved by pre-allocating a buffer and adding scratch memory to `compute_first_colliding_cell`.
- **Improved documentation**: More explicit documentation as requested by reviewers
- **Subtle bug fix** a  dist>0 is replaced by dist>=0. Very unlikely to trigger with GJK, as we rarely get machine 0, but better than not having it.


jhale (approved):

cpp/dolfinx/geometry/utils.h
```diff
@@ -844,87 +850,92 @@ determine_point_ownership(const mesh::Mesh<T>& mesh, std::span<const T> points,
...
-        }
-        const std::array<T, 3> d = compute_distance_gjk<T>(
-            std::span<const T>(point.data(), point.size()), nodes);
-        if (T current_distance = d[0] * d[0] + d[1] * d[1] + d[2] * d[2];
-            current_distance < shortest_distance)
-        {
-          shortest_distance = current_distance;
-          closest_cell = cell;
+          auto dofs = md::submdspan(x_dofmap, cell, md::full_extent);
+          std::vector<T> nodes(3 * dofs.size());
```
jhale: New dynamic memory allocation inside hotter loop?

python/dolfinx/fem/utils.py
```diff
@@ -83,6 +84,10 @@ def create_interpolation_data(
             interpolate into.
         padding: Absolute padding of bounding boxes of all entities on
             mesh_to.
+        allow_extrapolation: If ``True`` (default), points in `V_to`'s
+            mesh not overlapped by any cell in `V_from`'s mesh are
+            assigned the closest cell (within `padding`). If ``False``,
+            such points are left unowned.
```
jhale: Could you add the '-1' comment to all the docstrings? Most useful piece of information.

cpp/dolfinx/fem/interpolate.h
```diff
@@ -1103,14 +1103,19 @@ void piola_mapped_evaluation(const FiniteElement<U>& element, bool symmetric,
 /// an interpolation point from `mesh0` is on the surface of a cell in
 /// `mesh1`. This parameter can also be used for extrapolation, i.e. if
 /// cells in `mesh0` is not overlapped by `mesh1`.
+/// @param[in] allow_extrapolation If `true` (default), points in `mesh0`
```
garth-wells: This is quite hard to read, and I don't think 'extrapolate' is the right word.

The docstring for the `padding` argument implies that the same can be achieved by adjusting the padding.

garth-wells (commented):
It would be good to run an LLM over the docstrings for the affected functions. They can be made easier to read.

cpp/dolfinx/geometry/utils.h
jorgensd: This is a remainder from the time the geometry dofmap was an adjacency list. Will move out.

cpp/dolfinx/fem/interpolate.h
jorgensd: Extrapolate is the right word here, I believe, as the points found here is pulled back to the relevant reference space and then evaluated with those basis functions.
I've renamed the argument name in `determine_closest_point`, as that function does not relate to extrapolation. There it is now named `find_closest_cell`.

jorgensd:
Further fixes made by Claude and me:
- Sentinel bug: `current_dist > 0` → `current_dist >= 0` in the closest-cell aggregation loop. -1 is the "not computed" sentinel; a genuinely closest candidate can legitimately have squared distance 0 (point exactly on a cell boundary), and the old check rejected that valid answer the same way it rejected the sentinel. 
- Buffer hoisting: `compute_first_colliding_cell` now takes a coordinate_dofs scratch-buffer span (sized to fit, oversized buffers explicitly allowed for future mixed-topology support) instead of allocating internally. `determine_point_ownership` hoists one buffer before its point loop and reuses it for both the initial collision loop and the closest-cell fallback loop (previously only the fallback loop had been hoisted).
- Added a note documenting that find_closest_cell must be uniform across all ranks of `mesh.comm()`, since it gates collective MPI calls.
- Renamed send_extrapolate/dest_extrapolate → send_needs_closest_cell/dest_needs_closest_cell for consistency with the renamed public parameter.
Added missing direct includes: <cassert>, <limits>, <optional>, <utility>.
- Replaced auto with explicit types on plain std::int32_t/T values (loop variables over collisions.links(...), and current_dist/d in the aggregation loop), per AGENTS.md.

garth-wells (approved):

PR #4454, opened 2026-08-28
Enable OpenCASCADE CAF in the test environment for STEP attributes

jhale:
gmsh is built without `OpenCASCADE-CAF`, so STEP names and colours are unavailable for boundary marking. The OCCT build is fine — the XCAF toolkits already come in via `TKDESTEP`'s dependency closure. The missing piece is `libfreetype-dev`: gmsh wraps its whole CAF search in `if(FREETYPE_FOUND)` and silently drops CAF without it.

Still headless. Pulls in `libpng-dev`/`libbz2-dev`; gmsh gains `Png` as a side effect.

Verified with podman: `gmsh -info` gains `OpenCASCADE-CAF`, and a STEP import with `Geometry.OCCImportLabels=1` reads labels (1 vs 0 before).

Stacked on #4449; retarget to `main` once that merges.

AI assistance: I used Claude Code (Opus 5) to draft parts of this PR. I reviewed, edited, tested, and take responsibility for the final contribution.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4456, opened 2026-08-29
Move declaration of OpenMPI-related environment variables to `docker/Dockerfile.test-env`

francesco-ballarin:
Move setting the variables
```
OMPI_ALLOW_RUN_AS_ROOT=1
OMPI_ALLOW_RUN_AS_ROOT_CONFIRM=1
PRTE_MCA_rmaps_default_mapping_policy=":oversubscribe"
```
to `docker/Dockerfile.test-env`, instead of setting them in CI workflow.

The nightly image has been broken in parallel for around a week, probably since #4427, with errors like
```
prterun has detected an attempt to run as root.

Running as root is *strongly* discouraged as any mistake (e.g., in
defining TMPDIR) or bug can result in catastrophic damage to the OS
file system, leaving your system in an unusable state.

We strongly suggest that you run prterun as a non-root user.

You can override this protection by adding the --allow-run-as-root
option to your command line.  However, we reiterate our strong advice
against doing so - please do so at your own risk.
```
when trying to run with `mpirun`. Since a end user is not expected to use a non-root user, it makes sense that this variables should be part of the docker image, not added later on by the CI workflow.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -184,9 +184,6 @@ jobs:
     container: "ghcr.io/fenics/test-env:current"
     env:
       PETSC_ARCH: ${{ matrix.petsc_arch }}
-      OMPI_ALLOW_RUN_AS_ROOT: 1
-      OMPI_ALLOW_RUN_AS_ROOT_CONFIRM: 1
-      PRTE_MCA_rmaps_default_mapping_policy: :oversubscribe
       PYTEST_ADDOPTS: "-W error"
 
     name: Build and test (${{ matrix.petsc_arch }})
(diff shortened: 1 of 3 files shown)


francesco-ballarin:
CI will now fail, because I removed the environment variables from the yaml file, but not rebuilt the docker image yet. @jhale (#4454) and myself have both triggered a rebuild, and we should probably try not to step on each other toes when doing it ;)

jhale:
Wish OpenMPI would remove this slightly mad user hand-holding check...

Can you cancel your runs and make a PR? I would like to check the gmsh STEP fix first.

Also, we don't want to override the PRTE setting subscription setting, from my memory this is GitHub/CI related hack.

jhale:
Do you understand why the MPI basic test runs successfully in the end user image build? It would be good to work out why this test didn't pick up the issue you mention.

francesco-ballarin:
> Can you cancel your runs and make a PR? I would like to check the gmsh STEP fix first.

I did cancel my rebuild as soon as I realized you were already building yours.
The CI runs currently in progress in this PR will not interfere with yours

> Also, we don't want to override the PRTE setting subscription setting, from my memory this is GitHub/CI related hack.

OK, moved back into the CI workflow

> Do you understand why the MPI basic test runs successfully in the end user image build? It would be good to work out why this test didn't pick up the issue you mention.

Good point. Running under `mpirun -n ...` wasn't tested. I added one now.

PR #4457, opened 2026-08-30
Relax mixed Poisson tolerance for single precision

garth-wells:
## Summary

- use a relative tolerance of 1e-6 for 32-bit scalar types in the mixed-Poisson demo
- retain the 1e-8 tolerance for 64-bit scalar types

The previous 1e-8 tolerance is not robust for the fp32 FFCx integration jobs: small, valid code-generation round-off differences can cause GMRES to break down before reaching it.

## Testing

- ruff check python/demo/demo_mixed-poisson.py
- ruff format --check python/demo/demo_mixed-poisson.py
- serial mixed-Poisson demo with the local PETSc build
- verified tolerance selection for float32, complex64, float64, and complex128

The local PETSc installation is fp64, so the fp32 demo execution is covered by CI.

AI assistance: I used Codex to draft parts of this PR. I reviewed, edited, tested, and take responsibility for the final contribution.

### DIFF
--- a/python/demo/demo_mixed-poisson.py
+++ b/python/demo/demo_mixed-poisson.py
@@ -249,7 +249,7 @@ problem = fem.petsc.LinearProblem(
         "ksp_type": "gmres",
         "pc_type": "fieldsplit",
         "pc_fieldsplit_type": "additive",
-        "ksp_rtol": 1e-8,
+        "ksp_rtol": 1e-6 if np.finfo(dtype).bits == 32 else 1e-8,
         "ksp_gmres_restart": 100,
         "ksp_view": "",
     },


python/demo/demo_mixed-poisson.py
```diff
@@ -249,7 +249,7 @@
         "ksp_type": "gmres",
         "pc_type": "fieldsplit",
         "pc_fieldsplit_type": "additive",
-        "ksp_rtol": 1e-8,
+        "ksp_rtol": 1e-6 if np.finfo(dtype).bits == 32 else 1e-8,
```
michalhabera, line 252:
```suggestion
        "ksp_rtol": max(1e-8, 10 * np.finfo(dtype).eps),
```
Btw., this looks terrible. Plus the width of the type is not strictly related to its precision. It works in the limited case of fp32 and fp64 being all that is allowed, but it communicates a completely wrong intent.

PR #4458, opened 2026-08-30
Improve mixed Poisson demo

garth-wells:
## Summary

- Refactor the mixed Poisson demo into a degree-parameterised solver and run degrees 1 and 2.
- Make Hypre AMS versus LU selection explicit and clarify the solver setup.
- Write scalar and flux VTX output for each degree, including the required RT-to-DG flux interpolation.
- Use PETSc scalar and real types directly and simplify the demo documentation and element definitions.

## Testing

- ruff check python/demo/demo_mixed-poisson.py
- ruff format --check python/demo/demo_mixed-poisson.py
- Ran the demo with Hypre AMS for k=1 and k=2; verified all scalar and flux VTX outputs.
- Ran solve(1, False) to exercise the LU path.

AI assistance: I used Codex to draft parts of this PR. I reviewed, edited, tested, and take responsibility for the final contribution.

### DIFF
--- a/python/demo/demo_mixed-poisson.py
+++ b/python/demo/demo_mixed-poisson.py
@@ -32,9 +32,8 @@
 # ## Equation and problem definition
 #
 # An alternative formulation of Poisson equation can be formulated by
-# introducing an additional vector variable, the flux $\sigma = \nabla u$.
-# The partial differential equations
-# then read
+# introducing an additional vector variable, the flux $\sigma = \nabla
+# u$. The partial differential equations then read
 #
 # $$
 # \begin{aligned}
@@ -110,10 +109,10 @@ dtype = PETSc.ScalarType
 xdtype = PETSc.RealType
 # -
 
-# Create a two-dimensional mesh. The iterative solver constructed
-# later requires special construction that is specific to two
-# dimensions. Application in three-dimensions would require a number of
-# changes to the linear solver.
+# Create a two-dimensional mesh. The iterative solver constructed later
+# requires special construction that is specific to two dimensions.
+# Application in three-dimensions would require a number of changes to
+# the linear solver.
 
 # +
 msh = create_unit_square(MPI.COMM_WORLD, 96, 96, CellType.triangle, dtype=xdtype)
@@ -252,7 +251,6 @@ def solve(k: int, use_hypre: bool) -> tuple[fem.Function, fem.Function]:
             opts[f"{ksp_sigma.prefix}pc_hypre_ams_tol"] = 1e-12  # type: ignore[index]
             opts[f"{ksp_sigma.prefix}pc_hypre_ams_max_iter"] = 3  # type: ignore[index]
 
-        ksp_sigma.setFromOptions()
     else:
         pc_sigma.setType("lu")
         use_superlu = PETSc.IntType == np.int64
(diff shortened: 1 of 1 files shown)


jhale:
@garth-wells I'm without a proper computer. As this failure is a bit mysterious, could you disable the 32-bit tests on the merge queue so we can move forwards with some other PRs? It's somewhere in the settings...

python/demo/demo_mixed-poisson.py
```diff
@@ -100,259 +99,198 @@
...
+from dolfinx import fem, has_adios2, mesh
 from dolfinx.fem.petsc import discrete_gradient, interpolation_matrix
 from dolfinx.mesh import CellType, create_unit_square
 
 # Solution scalar (e.g., float32, complex128) and geometry (float32/64)
 # types
-dtype = default_scalar_type
-xdtype = default_real_type
+dtype = PETSc.ScalarType
+xdtype = PETSc.RealType
```
jorgensd, line 109: This change doesn't seem sensible to me, as default_scalar_type and real_scalar_type is equal to the petsc ones if we have petsc4py.

python/demo/demo_mixed-poisson.py
```diff
@@ -100,259 +99,198 @@
...
 # +
 msh = create_unit_square(MPI.COMM_WORLD, 96, 96, CellType.triangle, dtype=xdtype)
+gdim = msh.geometry.dim
+fdim = msh.topology.dim - 1
+facets_top = mesh.locate_entities_boundary(msh, fdim, lambda x: np.isclose(x[1], 1.0))
+facets_bottom = mesh.locate_entities_boundary(msh, fdim, lambda x: np.isclose(x[1], 0.0))
+cells_top = mesh.compute_incident_entities(msh.topology, facets_top, fdim, fdim + 1)
+cells_bottom = mesh.compute_incident_entities(msh.topology, facets_bottom, fdim, fdim + 1)
+has_hypre = PETSc.Sys().hasExternalPackage("hypre")
+hypre_ams_compatible = not np.issubdtype(dtype, np.complexfloating)
```
jorgensd, line 126: Wouldn't it be more natural to define this after the solver blocks, rather than here?

python/demo/demo_mixed-poisson.py
```diff
@@ -100,259 +99,198 @@
...
+            pc_sigma.setHYPRESetEdgeConstantVectors(cvec0.x.petsc_vec, cvec1.x.petsc_vec, None)
+        else:
+            V_H1d = fem.functionspace(msh, ("Lagrange", k, (msh.geometry.dim,)))
+            Pi = interpolation_matrix(V_H1d, V)
+            Pi.assemble()
+            pc_sigma.setHYPRESetInterpolations(msh.geometry.dim, None, None, Pi, None)
+            opts[f"{ksp_sigma.prefix}pc_hypre_ams_tol"] = 1e-12  # type: ignore[index]
+            opts[f"{ksp_sigma.prefix}pc_hypre_ams_max_iter"] = 3  # type: ignore[index]
 
-    # Construct and set the 'discrete gradient' operator, which maps
```
jorgensd, line 291: I think it is not great that text like this is removed in the new demo. Now there are tons of operations that are not explained in any detail.

PR #4461, opened 2026-08-31
Deduplicate CSR column search in MatrixCSR

garth-wells:
## Summary

Split out from #4438: `insert_csr`, `insert_blocked_csr`, `insert_nonblocked_csr`,
and `MatrixCSR`'s ghost-row unpacking each duplicated the same
`std::lower_bound`-based column search. This factors it into a shared
`la::impl::csr_position()` helper, and fixes a drive-by bug found while
doing so.

This was dropped from #4438 during rebasing/scope-narrowing of that PR
(now assembly-focused only) — re-proposing standalone since it's an
independent, self-contained improvement to `la::MatrixCSR`.

### Changes

- Factor the repeated column-search logic into `la::impl::csr_position()`.
- A hand-rolled "branchless" (conditional-move) rewrite of this search was
  tried first, on the theory that the column found is unpredictable from
  one call to the next so a branchy `std::lower_bound` mispredicts. It
  measured as a reproducible ~13% regression in
  `scripts/bench_backend.py`'s `csr_reasm` (P1, N=100): GCC 13 did not turn
  the idiom into a cmov, confirmed by disassembly — it kept a real
  conditional branch, structurally near-identical to libstdc++'s own
  `lower_bound`, just without libstdc++'s tuning. Reverted to a thin
  wrapper around `std::lower_bound` instead; the deduplication alone
  measured a genuine, reproducible +6.2% on P1/N=100 `csr_reasm` (no
  significant change on P3/N=64, where kernel cost dominates insertion
  cost).
- clang if-converts `csr_position`'s `std::lower_bound` into a branchless
  cmov chain, ~35-50% slower here than GCC's branchy codegen (the cmov
  serializes each iteration's load behind the prior comparison, killing
  speculation on this short, cache-resident row). Confirmed via a
  standalone reproducer and disassembly; a hand-rolled loop with
  `__builtin_expect` recovers full speed, but reimplementing `lower_bound`
  over a compiler codegen quirk isn't worth it, so this is left as-is and
  documented in a code comment so it isn't re-tried without new evidence.
- Drive-by: `MatrixCSR::add()` passed `_row_ptr.size()` as its debug-only
  row-range bound, one row looser than `set()`'s correct
  `size_local()+num_ghosts()`. Aligned `add()` to match, closing a
  theoretical one-row out-of-bounds read for `row == num_all_rows()` in
  Developer builds.

### Simplify block-size handling with `auto`/`std::integral_constant`

`insert_csr`/`insert_blocked_csr` took their block sizes as non-type
template parameters (`template <int BS0, int BS1, ...>`, always
compile-time), while `insert_nonblocked_csr` took plain runtime `int
bs0, int bs1` (via `std::div`). Converted all three to the
`BlockSizeArg auto` convention already used by `spmv`/`spmvT` in this
same file (a parameter that accepts either a runtime integral or a
`std::integral_constant<U, N>`), with `decltype(+BS0)`-typed loop
counters. `insert_nonblocked_csr`'s `std::div` calls become plain `/`,
`%`, since `std::div` requires concrete arithmetic types incompatible
with a generic `auto` parameter. `MatrixCSR::set`/`add`'s public
`template <int BS0, int BS1>` API is unchanged; they now pass
`std::integral_constant<int, BS0>{}` into `insert_csr`/
`insert_blocked_csr` instead of explicit template arguments.

## Testing

- Formatting: `clang-format --dry-run --Werror` passes on all touched
  files.
- The `csr_position` dedup + row-bound fix commits previously passed the
  full local C++ test suite (27476 assertions) and `python/test`
  `unit/la` + `unit/fem` (2594 tests) while part of #4438, before being
  reverted there purely to narrow that PR's scope to assembly
  performance.
- The block-size-handling simplification could not be exercised through
  the full local test suite in this environment (local MPI toolchain is
  currently broken here — a Homebrew `open-mpi` upgrade invalidated the
  cached PETSc/MPI paths — unrelated to this change). Instead verified
  standalone: compiled `matrix_csr_impl.h` directly with both GCC 16 and
  Apple clang under `-std=c++20 -Wall -Wextra -Wsign-compare` (clean, no
  warnings), and cross-checked `insert_csr`, `insert_blocked_csr`, and
  `insert_nonblocked_csr` — each called with both a runtime `int` and a
  `std::integral_constant` block size — against an independently
  constructed dense reference matrix; all six combinations reproduce the
  same matrix.

### DIFF
--- a/cpp/dolfinx/la/MatrixCSR.h
+++ b/cpp/dolfinx/la/MatrixCSR.h
@@ -16,6 +16,7 @@
 #include <mpi.h>
 #include <numeric>
 #include <span>
+#include <type_traits>
 #include <utility>
 #include <vector>
 
@@ -275,15 +276,17 @@ public:
     assert(x.size() == rows.size() * cols.size() * BS0 * BS1);
     if (_bs[0] == BS0 and _bs[1] == BS1)
     {
-      impl::insert_csr<BS0, BS1>(_data, _cols, _row_ptr, x, rows, cols, set_fn,
-                                 num_rows);
+      impl::insert_csr(_data, _cols, _row_ptr, x, rows, cols, set_fn, num_rows,
+                       std::integral_constant<int, BS0>{},
+                       std::integral_constant<int, BS1>{});
     }
     else if (_bs[0] == 1 and _bs[1] == 1)
     {
       // Set blocked data in a regular CSR matrix (_bs[0]=1, _bs[1]=1)
       // with correct sparsity
-      impl::insert_blocked_csr<BS0, BS1>(_data, _cols, _row_ptr, x, rows, cols,
-                                         set_fn, num_rows);
+      impl::insert_blocked_csr(_data, _cols, _row_ptr, x, rows, cols, set_fn,
+                               num_rows, std::integral_constant<int, BS0>{},
+                               std::integral_constant<int, BS1>{});
     }
     else
     {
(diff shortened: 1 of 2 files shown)


jhale:
"a drive-by bug" - nice to see Claude getting into stopping organised crime.

PR #4464, opened 2026-08-31
Fix right DOF transformation for mixed elements with block size > 1

### LINKED ISSUE #4294: [BUG]: dof_transformation_right_fn mis-slices data for mixed elements with block_size > 1
### Summarize the issue

This is a bug found by Claude AI - which may or may not be accurate. I don't know if this is correct or not, but just pasting here as an issue, in case anyone can look into it.

`cpp/dolfinx/fem/FiniteElement.h`
Function: `FiniteElement<T>::dof_transformation_right_fn` (mixed-element branch)

### Summary

In the mixed-element branch of dof_transformation_right_fn, each sub-element's DOF-transformation function is called with a data span that is not correctly restricted to that sub-element's own columns when block_size > 1. This silently produces incorrect data for any mixed (heterogeneous) element where a sub-element needs a DOF transformation and the transformation is applied from the right with more than one row of data — the situation that arises during bilinear-form matrix assembly.

### Current code

```
cpp// cpp/dolfinx/fem/FiniteElement.h
else if (!_sub_elements.empty())
{
  if (!_reference_value_shape) // Mixed element
  {
    std::vector<std::function<void(
        std::span<U>, std::span<const std::uint32_t>, std::int32_t, int)>>
        sub_element_fns;
    for (std::size_t i = 0; i < _sub_elements.size(); ++i)
    {
      sub_element_fns.push_back(
          _sub_elements[i]->template dof_transformation_right_fn<U>(ttype));
    }

    return [this, sub_element_fns](std::span<U> data,
                                   std::span<const std::uint32_t> cell_info,
                                   std::int32_t cell, int block_size)
    {
      std::size_t offset = 0;
      for (std::size_t e = 0; e < sub_element_fns.size(); ++e)
      {
        sub_element_fns[e](data.subspan(offset, data.size() - offset),
                           cell_info, cell, block_size);
        offset += _sub_elements[e]->space_dimension();
      }
    };
  }
  ...
```

### Root cause

block_size here is a row count, not a component count. This is confirmed by the call site in assemble_matrix_impl.h:

```
cpp// cpp/dolfinx/fem/assemble_matrix_impl.h
P0(Ae, cell_info0, cell0, ndim1);  // left-apply,  ndim1 rows
P1T(Ae, cell_info1, cell1, ndim0); // right-apply, ndim0 rows
```

data is therefore a row-major buffer of shape (block_size rows) x (total dofs). For sub-element e, occupying dof-columns [offset, offset + dim_e), the correct data for that sub-element is a strided slice — the same column range repeated in every row — not a single contiguous run.

### How to reproduce the bug

N/A

### Minimal Example (Python)

```Python

```

### Output (Python)

```bash

```

### Version

main branch

### DOLFINx git commit

_No response_

### Installation

_No response_

### Additional information

_No response_

PaulXiCao:
Fixes #4294.

In the mixed-element branch of `FiniteElement::dof_transformation_right_fn`,
the data passed to each sub-element was sliced as

```cpp
data.subspan(offset, data.size() - offset)
```

which treats the buffer as a single contiguous run. For a right-applied transformation the layout is `(block_size rows) x (ndofs columns)`, so sub-element `e` owns columns `[offset, offset + dims[e])` of *every* row. Basix derives the row stride as `data.size() / n`, so trimming the front of the span shortened the stride by `offset / n` and misaligned every row after the first. Only `block_size == 1`, where the trimmed prefix is absorbed exactly, was correct.

Each row is now transformed separately over the sub-element's own column range. This affects bilinear-form matrix assembly for mixed elements needing DOF transformations, where the transposed transformation is applied with one row per test-function DOF.

## Tests

`cpp/test/fem/finite_element.cpp` compares a multi-row application against a row-by-row application through the `block_size == 1` path, for a mixed element with the transforming sub-element at a non-zero DOF offset (mixed(P1 Lagrange, N1curl degree 2) on a triangle). Two control cases cover the same element at offset 0 and the non-mixed leaf path, both of which were already correct. All cases are parametrised over the four `doftransform` values, each of which failed independently before the fix.

## Note on performance

The sub-element transformation is now invoked once per row rather than once per sub-element, which loses the batching Basix does internally over `n`. Basix's right-apply already supports a column offset internally (`apply_tranpose_matrix_right` indexes `data[data_size * b + offset + i]`), so exposing that in the public `*_apply_right` signature would let this be a single call per sub-element again. Happy to follow up upstream if wanted.

---

AI assistance: I used Claude Code (Opus 5) to draft parts of this PR. I reviewed, edited, tested, and take responsibility for the final contribution.

### DIFF
--- a/cpp/test/fem/finite_element.cpp
+++ b/cpp/test/fem/finite_element.cpp
@@ -1,0 +1,171 @@
+// Copyright (C) 2026 Paul T. Kuehner
+//
+// This file is part of DOLFINX (https://www.fenicsproject.org)
+//
+// SPDX-License-Identifier:    LGPL-3.0-or-later
+
+#include <catch2/catch_test_macros.hpp>
+
+#include <basix/finite-element.h>
+
+#include <dolfinx/fem/FiniteElement.h>
+
+#include <array>
+#include <cstdint>
+#include <functional>
+#include <memory>
+#include <numeric>
+#include <span>
+#include <vector>
+
+using namespace dolfinx;
+
+namespace
+{
+using Func = std::function<void(
+    std::span<double>, std::span<const std::uint32_t>, std::int32_t, int)>;
+
+/// Nedelec (first kind), degree 2, on a triangle. Two DOFs per edge, so
+/// the DOF transformations are non-trivial (and not permutations).
+basix::FiniteElement<double> nedelec()
+{
+  return basix::create_element<double>(
+      basix::element::family::N1E, basix::cell::type::triangle, 2,
+      basix::element::lagrange_variant::legendre,
+      basix::element::dpc_variant::unset, false);
+}
+
+/// P1 Lagrange on a triangle. DOF transformations are the identity, so
+/// the only role of this sub-element is to shift the offset of the
+/// element it is mixed with.
+basix::FiniteElement<double> lagrange()
+{
+  return basix::create_element<double>(
+      basix::element::family::P, basix::cell::type::triangle, 1,
+      basix::element::lagrange_variant::gll_isaac,
+      basix::element::dpc_variant::unset, false);
+}
+
+/// Apply `fn` to `A` (shape `(nrows, ncols)`, row-major) one row at a
+/// time, i.e. exclusively through the `block_size == 1` code path.
+std::vector<double> apply_row_by_row(const Func& fn, std::span<const double> A,
+                                     int nrows, int ncols,
+                                     std::span<const std::uint32_t> cell_info)
+{
+  std::vector<double> out(A.begin(), A.end());
+  for (int i = 0; i < nrows; ++i)
+  {
+    const std::uint32_t cell = 0;
+    const int block_size = 1;
+    fn(std::span(out).subspan(i * ncols, ncols), cell_info, cell, block_size);
+  }
+  return out;
+}
+
+/// Row-major matrix of distinct, well-separated entries.
+std::vector<double> test_data(int nrows, int ncols)
+{
+  std::vector<double> A(nrows * ncols);
+  std::iota(A.begin(), A.end(), 1.0);
+  return A;
+}
+} // namespace
+
+TEST_CASE("Mixed element DOF transformation from the right, block size > 1",
+          "[fem][finite_element]")
+{
+  // Cell permutation with edges 0 and 1 reflected.
+  const std::array<std::uint32_t, 1> cell_info{0b011};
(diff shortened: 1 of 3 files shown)


jhale:
Could you write the tests in the Python at python/test/unit/fem? Perhaps there is an appropriate file with tests already related to transforms. We prefer Python for test writing unless there is a good specific reason to use C++.

cpp/test/fem/finite_element.cpp
```diff
@@ -0,0 +1,204 @@
+// Copyright (C) 2026 Paul T. Kuehner
```
schnellerhase: Not my code.

PaulXiCao: Wasnt sure what to put there, e.g. 
* owner of repo
* ower of component
* creator of the file

How does it change over time when someone else makes changes to the file?

I am putting my name for now..

PaulXiCao:
> Could you write the tests in the Python at python/test/unit/fem? Perhaps there is an appropriate file with tests already related to transforms. We prefer Python for test writing unless there is a good specific reason to use C++.

Hi @jhale . I am a new contributor and might need some guidance, but AFAI can tell there is no wrapper for either `dof_transformation_fn` function. I checked `python/dolfinx/wrappers/dolfinx_wrappers/fem.h:ll.132-345` and also just found these occurrences:
```bash
$ rg "dof_transformation" python 
python/dolfinx/fem/element.py
301:    def needs_dof_transformations(self) -> bool:
314:        return self._cpp_object.needs_dof_transformations

python/dolfinx/wrappers/dolfinx_wrappers/fem.h
341:        .def_prop_ro("needs_dof_transformations",
342:                     &dolfinx::fem::FiniteElement<T>::needs_dof_transformations)
```
Do you know if these were deliberately left out of the python wrapper? Otherwise i could look into adding them to the wrapper in a follow-up issue.

Or was your idea to write a python test where we directly call down into the c++ layer (similar to writing a temporary wrapper just for the test)?

jhale:
They were probably left out because they are performance critical functions that should not be called from Python.

What I would propose is that they are wrapped into dolfinx.cpp as _function_name, indicating that they are private, but not wrapped into the Python layer, which defines our public API. We can then unit test them from Python.

jhale (approved):
Thanks for this @PaulXiCao. I ported your work with a few minor tweaks using Claude Sonnet 5.

Issue #4451, opened 2026-08-27
[Bug]: Wrong integer type in dolfinx::geometry::impl::is_leaf


MathPhy42:
Hi,

I think the argument type of `bbox` in the function `dolfinx::geometry::impl::is_leaf`:

https://github.com/FEniCS/dolfinx/blob/5c1c561ff2eee35f081b63bc5a17d85f64bd50cf/cpp/dolfinx/geometry/utils.h#L187-L192

should be `std::array<std::int32_t, 2>` instead of `std::array<int, 2>` corresponding to the return type of `dolfinx::geometry::BoundingBoxTree::bbox`:

https://github.com/FEniCS/dolfinx/blob/5c1c561ff2eee35f081b63bc5a17d85f64bd50cf/cpp/dolfinx/geometry/BoundingBoxTree.h#L400-L411

I think most compilers did not care about this issue since their types `int` and `std::int32_t` are the same, but in general there's no guarantee for that. 
Furthermore, wouldn't it be better making `is_leaf` a (static) function of the class `dolfinx::geometry::BoundingBoxTree`?

jhale:
There are probably a few stray ints through this code, rather than std::int32_t.

I don't see a need to promote the free function to a public static member. Can you explain why you would want that?

MathPhy42:
The promotion to a member function was just a suggestion. There's no need to do it. 
I just thought the `is_leaf` function would fit better in the `dolfinx::geometry::BoundingBoxTree` class since there's a close relation between the class and this function.

jhale:
Indeed, we often put functions exclusively related to a class in an anonymous namespace next to the class. I will check if it's being used elsewhere first.

PaulXiCao:
AFAI can tell `is_leaf` is only being used inside this file, e.g.
```bash
$ rg is_leaf                                                
cpp/dolfinx/geometry/utils.h
188:constexpr bool is_leaf(std::array<int, 2> bbox)
249:  if (is_leaf(bbox))
318:        is_leaf(bbox) and point_in_bbox(view_bbox(next), p))
379:  const bool is_leaf_A = is_leaf(bbox_A);
380:  const bool is_leaf_B = is_leaf(bbox_B);
381:  if (is_leaf_A and is_leaf_B)
388:  else if (is_leaf_A)
394:  else if (is_leaf_B)
585:    assert(impl::is_leaf(leaf0));
```

I'll try to take a look into cleaning up the stray `int`s in this file.

[...]

PR #4467, opened 2026-09-01
Use `std::int32_t` for around `geometry/utils.h:is_leaf()`

PaulXiCao:
Partly of #4451 .

Uses `std::int32_t` instead of `int` for `impl::is_leaf()` inside `geometry/utils.h`. This is founded in the return type of `geometry::BoundingBoxTree::bbox()`.

Also fixes some more stray `int`s inside that file.


schnellerhase (approved):

PaulXiCao:
@schnellerhase Thanks for the quick feedback/merging here! First merge for me in this repo :D

PR #4468, opened 2026-09-02
Fix get_petsc_lib() for a pip-installed PETSc

shimwell:
## Summary

`import dolfinx.fem.petsc` fails outright against a PETSc installed with pip:

```
OSError: .../site-packages/petsc/lib/libpetsc.so: file too short
```

`get_petsc_lib()` only looked for the unversioned `libpetsc.so`, which in a pip-installed PETSc is a 24 byte GNU ld linker script rather than a library, because wheels cannot contain symbolic links:

```
$ cat site-packages/petsc/lib/libpetsc.so
INPUT(libpetsc.so.3.25)
```

`os.path.exists()` accepts it, so we returned a path `dlopen` cannot open. Linking is unaffected because `ld` follows the script, so only the runtime load fails. It takes out the whole module rather than one function, since `ctypes_utils`, `cffi_utils` and `numba_utils` all call `get_petsc_lib()` at class body scope.

Fix is to put the versioned name ahead of the unversioned one, keeping the latter as a fallback. Nothing changes for a source built PETSc, where the versioned library is what `libpetsc.so` points at anyway.

The "more than one library found" error has to go, since both names now normally exist. Deduplicating instead does not work: the linker script is a regular file, so `realpath` returns the script rather than the versioned library and the check fires on every pip install. Happy to add a different guard if you would rather keep one.

Relevant now because [petsc!9488](https://gitlab.com/petsc/petsc/-/merge_requests/9488) landed in PETSc 3.25.5, which keeps the pkg-config files in the Python package tree and makes a pip-installed PETSc usable by DOLFINx at all.

## Testing

Against `pip install petsc==3.25.5 petsc4py==3.25.5` (Linux, OpenMPI, Python 3.14), with DOLFINx built against it:

- before: `import dolfinx.fem.petsc` raises `OSError: ... file too short`; after: `get_petsc_lib()` returns `libpetsc.so.3.25` and all three utils classes load
- `pytest -m petsc4py` over `fem/test_assembler.py` and `la/`, 30 passed
- `fem/test_custom_assembler.py` and `fem/test_petsc_custom_assembler.py`, 5 passed 2 skipped (pre-existing)
- a `LinearProblem` Poisson solve converges to `max(u) = 0.073446`
- `ruff check` and `ruff format --check` pass, and a source built PETSc resolves to the same library as before

#4469 adds CMake discovery so this is reachable without setting `PETSC_DIR` by hand. The two are independent.


shimwell:
CI is failing for unrelated reasons to this PR. So I've put in a separate PR to fix the CI and if https://github.com/FEniCS/dolfinx/pull/4470 is merged then I think I can get this PR to pass all the CI

python/dolfinx/fem/petsc.py
```diff
@@ -1735,14 +1735,18 @@ def get_petsc_lib() -> pathlib.Path:
...
     Raises:
-        RuntimeError: If PETSc library cannot be found for if more than
-            one library is found.
+        RuntimeError: If PETSc library cannot be found.
     """
     import petsc4py as _petsc4py
 
     petsc_dir = _petsc4py.get_config()["PETSC_DIR"]
     petsc_arch = _petsc4py.lib.getPathArchPETSc()[1]  # type: ignore
+    # The unversioned libpetsc.so is a GNU ld linker script, not a
```
jhale: Not universally true, remove. Logical to go from specific lib version to general in all cases.

python/dolfinx/fem/petsc.py
```diff
@@ -1735,14 +1735,18 @@ def get_petsc_lib() -> pathlib.Path:
...
+    # shared library, in a pip-installed PETSc.
+    version = ".".join(str(v) for v in PETSc.Sys.getVersion()[:2])
     candidate_paths = [
+        os.path.join(petsc_dir, petsc_arch, "lib", f"libpetsc.so.{version}"),
```
jhale: Is it worth it/possible to add the patch version i.e most specific version first?

python/dolfinx/fem/petsc.py
```diff
@@ -1735,14 +1735,18 @@ def get_petsc_lib() -> pathlib.Path:
...
-            one library is found.
+        RuntimeError: If PETSc library cannot be found.
     """
     import petsc4py as _petsc4py
 
     petsc_dir = _petsc4py.get_config()["PETSC_DIR"]
     petsc_arch = _petsc4py.lib.getPathArchPETSc()[1]  # type: ignore
+    # The unversioned libpetsc.so is a GNU ld linker script, not a
+    # shared library, in a pip-installed PETSc.
+    version = ".".join(str(v) for v in PETSc.Sys.getVersion()[:2])
```
jhale: Consider major_minor_version for variable name

python/dolfinx/fem/petsc.py
shimwell: Ah yes sorry this is just for linux I forgot about mac and windows. I've removed this code comment now, thanks for spotting that

python/dolfinx/fem/petsc.py
shimwell: Yes I can add the patch version, I agree it is worth having, sorry for missing it the first time.

jhale (approved):

python/dolfinx/fem/petsc.py
shimwell: Yes I like it, I can do this rename here and elsewhere to be consistent.

shimwell:
Many many thanks for the speedy review @jhale I've done as you recommend with the review comments

python/dolfinx/fem/petsc.py
```suggestion
exists_paths = [path for path in candidate_paths if os.path.exists(path)]
```
Not altered here but can be simplified.

shimwell: would this exceed the 100 line char limit that the ruff check imposes

schnellerhase: I'm bad in counting 😅  - possibly needs a reformat.

shimwell: Perhaps something like this, but I might be creeping a bit far from the PR objective
image

schnellerhase: even better,  no list ever constructed.

PR #4469, opened 2026-09-02
Find a pip-installed PETSc/SLEPc via the Python interpreter

shimwell:
## Summary

[petsc!9488](https://gitlab.com/petsc/petsc/-/merge_requests/9488) landed and is included in PETSc 3.25.5 on PyPI just a few days ago. This change in PETSC is one I added to keep the pkg-config files in the Python package tree, with a relocatable prefix:

```
$ head -1 site-packages/petsc/lib/pkgconfig/PETSc.pc
prefix=${pcfiledir}/../..
$ pkg-config --modversion PETSc
3.25.5
```

`cpp/CMakeLists.txt` already prepends `$PETSC_DIR/lib/pkgconfig` to `PKG_CONFIG_PATH`, which happens to match the pip layout exactly, so this works today if you know to go and export `PETSC_DIR` yourself. This asks the Python interpreter for `petsc.get_petsc_dir()` when `PETSC_DIR` is unset.

The same hint goes into `DOLFINXConfig.cmake.in`, which repeats the `pkg_search_module()` call for consumers. Without it, configuring against an installed DOLFINx, including the DOLFINx Python interface, fails with `None of the required 'PETSc;petsc' found` even though the C++ library was just linked against that PETSc. The Basix hint is already duplicated there for the same reason.

No new CMake option: `PETSC_DIR`/`SLEPC_DIR` is already the override, and the hint only gets used when it is unset. 

## Testing

Built DOLFINx C++ and Python end to end against a pip PETSc with no environment variables set (Linux, OpenMPI, Python 3.14, under `env -u PETSC_DIR -u PKG_CONFIG_PATH`), after:

```
PETSC_CONFIGURE_OPTIONS="--download-parmetis --download-metis" pip install petsc petsc4py
```

```
-- Found petsc Python module at .../site-packages/petsc
-- Found PETSc version 3.25.5, prefix: .../site-packages/petsc/lib/pkgconfig/../..
-- Found ParMETIS: .../site-packages/petsc/lib/libparmetis.so (found suitable version "4.0.3")
```

- `ldd libdolfinx.so` resolves `libpetsc.so.3.25`, `libparmetis.so` and `libbasix.so.0` out of `site-packages`
- the Python interface then builds against the installed `DOLFINXConfig.cmake`, where the hint fires again
- a `LinearProblem` Poisson solve converges to `max(u) = 0.073446`; `pytest -m petsc4py`, 36 passed
- unchanged with an explicit `PETSC_DIR`, and with a system PETSc on `PKG_CONFIG_PATH`; `gersemi --check .` passes

`--download-parmetis` also satisfies the mandatory graph partitioner from the same prefix, which is already appended to `CMAKE_PREFIX_PATH`. That is what footnote [2] in `installation.rst` recommends, so the pip route needs no special case.

PETSc publishes sdists only, so `pip install petsc` still compiles from source against whatever `mpicc` resolves to, which is the right ABI for DOLFINx. [petsc!9516](https://gitlab.com/petsc/petsc/-/merge_requests/9516) has also landed, due next release, and is what would make binary wheels viable later.

Independent of #4468, which fixes `get_petsc_lib()` for a pip PETSc. Both are needed for the pip route to work end to end; neither depends on the other.


shimwell:
CI is failing for unrelated reasons to this PR. So I've put in a separate PR to fix the CI and if https://github.com/FEniCS/dolfinx/pull/4470 is merged then I think I can get this PR to pass all the CI

jhale (approved):
Thanks for this useful PR.

I made some small additions so that this functionality can be explicitly disabled, which we prefer for Spack and Easybuild packaging.

As OpenMPI 6 is going to be released soon, we can try building some DOLFINx MPI ABI 5 wheels without PETSc, in the first instance.

PR #4476, opened 2026-09-05
Rework marking interface to respect ghosts

schnellerhase:
https://github.com/FEniCS/dolfinx/pull/4156 missed the handling of ghost entries; caught by @jhale in extending upon PR https://github.com/FEniCS/dolfinx/pull/4187.

The underlying issue is that marker computation needs to happen on locally owned entities; while the marked entities (output) need to include ghost members. The dolfinx native way to achieve this is to just rely on a vector which carries exactly the infrastructure to differentiate between ghosted and non-ghosted contributions. Also this makes the caller interface simpler; as usually markers are $\mathcal{DG}_0$ functions anyhow.

- `mark_maximum` now computes the max over the locally owned entries (no change here, max. is insensitive to duplicates - but delicate for other markers)
- and returns the marked ghost entities as well
- tests extended to include non-trivial ghost setups
- Python wrapper pulled in from https://github.com/FEniCS/dolfinx/pull/4187
- supports both vector and array based interfaces.

Part of https://github.com/FEniCS/dolfinx/issues/4141.

### DIFF
--- a/cpp/test/mesh/refinement/mark.cpp
+++ b/cpp/test/mesh/refinement/mark.cpp
@@ -6,8 +6,11 @@
 
 #include <algorithm>
 #include <catch2/catch_template_test_macros.hpp>
+#include <cstddef>
+#include <cstdint>
 #include <dolfinx/common/MPI.h>
 #include <dolfinx/refinement/mark.h>
+#include <memory>
 #include <mpi.h>
 #include <vector>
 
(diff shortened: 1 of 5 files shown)


jhale:
Looks good.

I have a few small suggestions to make, but away from PC until Wednesday.

garth-wells (commented):
Does this use anywhere `la::Vector` functionality, e.g. fwd/rev scatter?

schnellerhase:
It does use its layout (returning indices of possible ghost elements), but it does not scatter itself - updated vector on input is assumed.

garth-wells:
> It does use its layout (returning indices of possible ghost elements), but it does not scatter itself - updated vector on input is assumed.

Is an IndexMap enough? Best to use the lightest possible concept. `la::Vector` feels heavyweight.

jhale:
I've pushed a new version here:

https://github.com/FEniCS/dolfinx/compare/schnellerhase/fix-marking...jhale/fix-marking-tweak

I'm still a bit unsure on how best to handle the end points theta = 0 and theta = 1.

The docstrings have been carefully re-written to use Vector terminology, rather than 'entity' and 'marker/marked' terminology - we are simply asking users to walk into the mesh entity numbering == dof numbering trap written as it is (your tests correctly construct v on the cell dofmap, but I guarantee users will start throwing in DG0 functions!).

Once we're a bit further ahead with this effort we should add a proper AMR demo in another PR.

jhale:
> > It does use its layout (returning indices of possible ghost elements), but it does not scatter itself - updated vector on input is assumed.
> 
> 
> 
> Is an IndexMap enough? Best to use the lightest possible concept. `la::Vector` feels heavyweight.

It is strictly enough. The only downside is that the interface for the other marker we have in mind may require Vector.

schnellerhase:
But we do need the values as well; so either ghosted span and index map or vector. But the former is just a vector right?

jhale:
Indeed, the suggestion is a span +  index map. If you look at the definition of Vector it does have all sorts of fancy stuff we don't use here.

jhale:
Here's a diff proposing `mark_maximum` take a `span`/`ndarray` + `IndexMap` instead of `la::Vector`, per the discussion above: https://github.com/FEniCS/dolfinx/compare/schnellerhase/fix-marking...jhale/mark-take-span-indexmap

schnellerhase:
Okay, let's switch to span + im for the core implementation and provide a wrapped callback based on a vector (which is the default use case), this allows for no vector based C++ implementation and the user facing interface may remain `mark_maximum(v, theta)` over `mark_maximum(v.array, v.index_map, theta)`, if vector based.

jhale:
Is that necessary? Seems trivial from the examples.

The less trivial case is a user wanting to pass a DG0 Function - that requires a mapping which is easy to forget to do, or possibly annoying boiler plate, and perhaps warrants a wrapper.

cpp/test/mesh/refinement/mark.cpp
```diff
@@ -17,36 +20,60 @@ using namespace dolfinx::refinement;
...
+    for (int i = 0; i < size; i++)
+      v[i] = i;
+  }
+  else
+  {
+    CHECK(v.size() == 1);
+    // Poison the ghost slot: if the local max were wrongly computed over
+    // ghosts too, this would inflate the (globally reduced) threshold and
+    // the checks below would fail.
+    v[0] = static_cast<TestType>(1000);
```
schnellerhase:
```suggestion
    // Check max reduction ignores ghosts.
    v[0] = static_cast<TestType>(size + 1);
```

schnellerhase:
API path `f(a.x, a.y)` forces bad style on caller and asks for a `f(a)`path. Also, I'd like to retain compatibility with PETSc's [`VecTagger`](https://petsc.org/main/manualpages/Vec/VecTagger/), which is vector based. The mapping to dofs is a separate issue, but input to marking is not necessarily function space information.

jhale:
No member of DOLFINx's public API takes `la::Vector` as an input argument. For PETSc, the `std::span` interface is ideal - get the raw contiguous array from PETSc Vec and pass it as a span.

schnellerhase:
The other usages do not require both the index map and the values. Having them side by side and not treat as vector feels contradictory to me. But we can revert here - this is not about finalizing the interface. Regarding PETSc compatibility I was thinking it should be easy to replace our markers for a PETSc one. So the code path for assemble + mark should look as close as possible between the two, so not a conversion, but how the same estimate + mark steps look like.

PR #4477, opened 2026-09-08
Additional tests for codim-1 manifolds

jorgensd:
Companion tests for: https://github.com/FEniCS/ffcx/pull/879

Let $\Omega$ be the parent mesh and $\Lambda$ a codim-1 submesh of it, $g \in G(\Lambda)$ a
degree- $k$ Lagrange function on $\Lambda$, and $\bar x$ the spatial coordinate of $\Lambda$.
Write

- `dxL = ufl.dx(domain=Lambda)` — $\Lambda$ integrated as a mesh in its own right;
- `dsO = ufl.ds(domain=Omega)`, `dSO = ufl.dS(domain=Omega)` — the same entities
  integrated as facets of $\Omega$, assembled with `entity_maps=[entity_map]`.

Each test asserts that a quantity belonging to $\Lambda$ is unchanged by which of the two
measures carries it:

- **Manifold gradient**, $\Lambda = \partial\Omega$:
  `g.dx(0)*dxL == g.dx(0)*dsO`, on affine and on higher-order (curved) $\Omega$.
- **Spatial coordinate**, $\Lambda = \partial\Omega$:
  `h(x_bar)*dxL == h(x_bar)*dsO`. Here $\Lambda$ carries no coefficient or argument, so it
  enters only through geometry — this covers the coordinate-dof gather alone, and checks
  that a form which is mixed-dimensional only through a geometric terminal still gets its
  facet permutations computed. $h$ is nonlinear ($h(\bar x) = \bar x_0^2$); a linear one is
  orientation-invariant and is reproduced even by a wrong gather.
- **Restrictions**, $\Lambda =$ the interior facets:
  `g.dx(0)*dxL == g('+').dx(0)*dSO == g('-').dx(0)*dSO`. The `"-"` side's coordinate dofs
  start one *parent* cell into `coordinate_dofs`; the submesh's smaller stride gave 0.
  `test_interior_facet_codim_1` checks values only, and those are right on both
  restrictions even when the gradient is not.

$g$ interpolates $x_0 + 2x_1 + 3x_2$, which is linear (hence exact at every $k$) with a
distinct coefficient per axis (so a wrong tangential gradient does not cancel around the
boundary). $\Omega$ is a deliberately non-cubic box for the same reason. Each case is
parametrised over `triangle`, `quadrilateral`, `tetrahedron` and `hexahedron`; the gradient
test additionally over $k \in \{1, 3\}$, and the higher-order case over geometry degree
$\in \{2, 3\}$ — a unit square/cube mapped onto an ellipse/ellipsoid via
`interpolate_geometry`. The affine cases assert exact values (6/18 for the gradient,
27/72 for the coordinate); the curved ones have no closed form, so only the two paths are
compared.

Verified against FFCx `main` and against FFCx before #879, with a fresh JIT cache each time:
all 24 cases pass with #879 and all 24 fail without it, serially and under `mpirun -n 3`.

### DIFF
--- a/.github/workflows/fenicsx-refs.env
+++ b/.github/workflows/fenicsx-refs.env
@@ -1,6 +1,6 @@
 basix_repository=FEniCS/basix
 basix_ref=main
 ufl_repository=FEniCS/ufl
-ufl_ref=dokken/manifold-fixes
+ufl_ref=main
 ffcx_repository=FEniCS/ffcx
-ffcx_ref=main
+ffcx_ref=dokken/manifold-fixes
--- a/.github/workflows/fenicsx-refs.env
+++ b/.github/workflows/fenicsx-refs.env
@@ -1,6 +1,6 @@
 basix_repository=FEniCS/basix
 basix_ref=main
 ufl_repository=FEniCS/ufl
-ufl_ref=main
+ufl_ref=dokken/manifold-fixes
 ffcx_repository=FEniCS/ffcx
 ffcx_ref=main
--- a/python/test/unit/fem/test_assemble_submesh.py
+++ b/python/test/unit/fem/test_assemble_submesh.py
@@ -422,6 +422,46 @@ def test_mixed_dom_codim_1_gradient(n, k):
     assert np.isclose(c_parent, c_submesh)
 
 
+@pytest.mark.parametrize("cell_type", [CellType.tetrahedron, CellType.hexahedron])
+def test_mixed_dom_codim_1_gradient_3d(cell_type):
+    """3D analogue of test_mixed_dom_codim_1_gradient.
+
+    Grad of a coefficient defined on a codimension-1 (2D facet) submesh
+    of a 3D mesh must agree whether integrated via an exterior facet
+    measure on the parent mesh or directly on the submesh's own cell
+    measure, for both a simplex parent (tetrahedron -> triangle facets)
+    and a tensor-product parent (hexahedron -> quadrilateral facets).
+    """
+    msh = create_box(
+        MPI.COMM_WORLD, [[0.0, 0.0, 0.0], [3.0, 1.0, 2.0]], (6, 2, 4), cell_type=cell_type
+    )
+
+    tdim = msh.topology.dim
+    msh.topology.create_connectivity(tdim - 1, tdim)
+    boundary_facets = exterior_facet_indices(msh.topology)
+    smsh, entity_map = create_submesh(msh, tdim - 1, boundary_facets)[:2]
+
+    Vbar = fem.functionspace(smsh, ("Lagrange", 1))
+    g = fem.Function(Vbar)
+    g.interpolate(lambda x: x[0] + 2.0 * x[1] + 3.0 * x[2])
+
+    ds = ufl.Measure("ds", domain=msh)
+    dx_smsh = ufl.Measure("dx", domain=smsh)
+
+    entity_maps = [entity_map]
+    M_parent = fem.form(g.dx(0) * ds, entity_maps=entity_maps)
+    M_submesh = fem.form(g.dx(0) * dx_smsh)
+
+    c_parent = msh.comm.allreduce(fem.assemble_scalar(M_parent), op=MPI.SUM)
+    c_submesh = msh.comm.allreduce(fem.assemble_scalar(M_submesh), op=MPI.SUM)
+
+    # Guard against a vacuous test (a too-symmetric domain/coefficient can
+    # let this bug's error cancel to ~0 across the boundary, as happened
+    # with an earlier version of test_mixed_dom_codim_1_gradient).
+    assert not np.isclose(c_submesh, 0.0)
+    assert np.isclose(c_parent, c_submesh)
+
+
 def test_disjoint_submeshes():
     # FIXME Simplify this test
     """Test assembly with multiple disjoint submeshes in same variational form."""
--- a/python/test/unit/fem/test_assemble_submesh.py
+++ b/python/test/unit/fem/test_assemble_submesh.py
@@ -455,10 +455,14 @@ def test_mixed_dom_codim_1_gradient_3d(cell_type):
     c_parent = msh.comm.allreduce(fem.assemble_scalar(M_parent), op=MPI.SUM)
     c_submesh = msh.comm.allreduce(fem.assemble_scalar(M_submesh), op=MPI.SUM)
 
-    # Guard against a vacuous test (a too-symmetric domain/coefficient can
-    # let this bug's error cancel to ~0 across the boundary, as happened
-    # with an earlier version of test_mixed_dom_codim_1_gradient).
-    assert not np.isclose(c_submesh, 0.0)
+    # g.dx(0) is the x-component of the *manifold* (tangential) gradient
+    # of g, i.e. of (1, 2, 3) projected onto each face's own tangent
+    # plane. On the x=const faces (normal (1, 0, 0)) that projection
+    # removes the x-component entirely, giving 0; on the y=const and
+    # z=const faces it is 1. Weighted by each pair of faces' area:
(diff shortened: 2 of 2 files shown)


jhale (approved):

.github/workflows/fenicsx-refs.env
```diff
@@ -3,4 +3,4 @@ basix_ref=main
 ufl_repository=FEniCS/ufl
 ufl_ref=main
 ffcx_repository=FEniCS/ffcx
-ffcx_ref=main
+ffcx_ref=dokken/manifold-fixes
```
garth-wells: Blocker for merge: needs to go back to `ffcx_ref=main` once FFCx #879 lands.

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -377,6 +377,95 @@ def coeff_expr(x):
...
+    Regression test: the parent-mesh facet integral used to silently use
+    the submesh coefficient's own (unavailable) Jacobian instead of the
+    parent's FacetJacobian, giving a facet-independent, wrong result. A
+    non-square rectangle and a coefficient linear in x (whose derivative
+    is a nonzero constant, exactly reproducible at every degree k) are
+    used deliberately: on a symmetric domain with a smooth/periodic
+    coefficient this bug's error can cancel to ~0 across the boundary,
+    masking the regression.
+    """
+    msh = create_rectangle(MPI.COMM_WORLD, ((0.0, 0.0), (3.0, 1.0)), n)
```
garth-wells, line 397: Two coverage gaps anchor here, both cheap to close and both verified locally.

**Only affine (P1) parent geometry is tested.** This is the biggest one: the heart of #879 is the per-orientation closure-dof tables for *arbitrary-order* coordinate elements, but with P1 geometry a facet's closure dofs are just its vertices, so the orientation tables are barely exercised. I built P2 and P3 parent meshes (triangle, quadrilateral, tetrahedron, hexahedron) and ran the same comparison: all fail on FFCx `main` (P2 triangle gives 0.0335 against the exact 6.0) and all pass with #879. At least one higher-order-geometry case belongs here.

**Only triangles in 2D**, while the 3D test deliberately covers simplex + tensor-product. A quadrilateral parent gives exactly 0.0 on FFCx `main` and passes with #879, so adding `cell_type` to the parametrisation is nearly free and restores the symmetry with the 3D test.

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -377,6 +377,95 @@ def coeff_expr(x):
...
+
+    Vbar = fem.functionspace(smsh, ("Lagrange", k))
+    g = fem.Function(Vbar)
+    g.interpolate(lambda x: x[0] + 2.0 * x[1])
+
+    ds = ufl.Measure("ds", domain=msh)
+    dx_smsh = ufl.Measure("dx", domain=smsh)
+
+    entity_maps = [entity_map]
+    M_parent = fem.form(g.dx(0) * ds, entity_maps=entity_maps)
```
garth-wells, line 412: `SpatialCoordinate` on the submesh is untested. #879 gates the coordinate-dofs gather on `t in (ufl.geometry.SpatialCoordinate, ufl.geometry.Jacobian)`, but only the Jacobian half is exercised here. `ufl.SpatialCoordinate(smsh)[0] * ds` (with `entity_maps`) gives 11.0 on FFCx `main` against the correct 12.0 for this mesh, and is fixed by #879 — so it's a one-line addition that covers the other half of the gate.

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -377,6 +377,95 @@ def coeff_expr(x):
...
+def test_mixed_dom_codim_1_gradient(n, k):
+    """Test that Grad of a coefficient defined on a codimension-1 submesh
+    (the boundary of a mesh) gives the same result whether integrated via
+    an exterior facet measure on the parent mesh, or directly via a cell
+    measure on the submesh itself.
+
+    Regression test: the parent-mesh facet integral used to silently use
+    the submesh coefficient's own (unavailable) Jacobian instead of the
+    parent's FacetJacobian, giving a facet-independent, wrong result. A
+    non-square rectangle and a coefficient linear in x (whose derivative
```
garth-wells, line 391: Small inaccuracy: the coefficient is linear in x *and* y (`x[0] + 2.0 * x[1]`), not just x — and it's the y-dependence that makes the vertical edges contribute 0 rather than something wrong-but-cancelling, so it's worth stating.

More generally, both docstrings are long by the repo convention (AGENTS.md: "compress comments to their essence"). The regression rationale is worth keeping — the point about cancellation on a symmetric domain is a real trap for whoever touches this next — but it fits in two or three lines.

python/test/unit/fem/test_assemble_submesh.py
```diff
@@ -377,6 +377,95 @@ def coeff_expr(x):
...
+
+    # g.dx(0) is the *manifold* (tangential) gradient's x-component: 1 on
+    # the horizontal edges (tangent (1, 0), total length 6), 0 on the
+    # vertical edges (tangent (0, 1), orthogonal to x).
+    assert np.isclose(c_submesh, 6.0)
+    assert np.isclose(c_parent, c_submesh)
+
+
+@pytest.mark.parametrize("cell_type", [CellType.tetrahedron, CellType.hexahedron])
+def test_mixed_dom_codim_1_gradient_3d(cell_type):
```
garth-wells, line 426: This and `test_mixed_dom_codim_1_gradient` are ~90% identical — only the mesh factory, the interpolant and the expected value differ. Parametrising a single test over `(mesh factory, expected value)` would collapse them and remove the current asymmetry where the 2D test parametrises `k in {1, 3}` but the 3D one fixes `k=1`.

garth-wells (commented):
Reviewed the tests by running them locally against both FFCx `main` and the FFCx #879 branch, with a fresh JIT cache each time (a stale cache silently masks the difference).

| | FFCx `main` | FFCx #879 |
|---|---|---|
| the 6 new cases | **all fail** | **all pass** |
| same, `mpirun -n 3` | — | all pass |
| full `test_assemble_submesh.py` | — | 82 passed, 1 xfailed |

`ruff check` / `ruff format --check` clean. I re-derived 6.0 and 18.0 independently — both correct — and the `c_submesh` assertions hold even on FFCx `main`, so the tests correctly isolate the parent-facet path. These are genuine, well-targeted regression tests. Comments inline; two things that don't anchor to a line:

### A real bug survives the fix: the `"-"` restriction of a gradient in `dS`

On a unit square with an interior-facet submesh and `g = x + 2y`, **with #879 applied**:

```
submesh dx  g.dx(0)        -> 23.970563   (reference)
dS          g('+').dx(0)   -> 23.970563   OK
dS          g('-').dx(0)   ->  0.000000   WRONG
dS          avg(g.dx(0))   -> 11.985281   (silently halved)
```

Values are fine on both restrictions (`avg(g)`, `g('+')`, `g('-')` all give 37.9706); only the gradient's `-` side is wrong. `test_interior_facet_codim_1` checks values only, so nothing catches this. Since `definitions.py` in #879 explicitly handles `mt.restriction == "-"` (separate offset and `quadrature_permutation[1]`), a `dS` + gradient test belongs in this PR — it will fail today, so it needs either the FFCx side extended or a strict `xfail` plus a tracking issue.

### FYI: the ridge/peak branches in #879 are unreachable from DOLFINx

A codim-2 submesh coefficient under `dr` throws `Integral type not supported.` from `cpp/dolfinx/fem/Form.h:295`, so `ridge_closure_dofs` / `peak_closure_dofs` are dead code for now. Nothing to add here, just worth knowing they ship untested.

.github/workflows/fenicsx-refs.env
```suggestion
ffcx_ref=main
```

jorgensd:
All tests have been improved/condensed/parametrized and documented.

PR #4483, opened 2026-09-10
la::Vector: add a constructor taking a shared Scatterer

### LINKED ISSUE #3065: Associate a `common::Scatterer` with `fem::FunctionSpace`

jhale:
## Summary

Adds one constructor to `la::Vector`, taking a
`std::shared_ptr<const common::Scatterer<...>>` alongside the index map:

```cpp
Vector(std::shared_ptr<const common::IndexMap> map,
       std::shared_ptr<const common::Scatterer<ScatterContainer>> scatterer);
```

The `(map, bs)` constructor delegates to it. `common::Scatterer` gains a
`bs()` accessor so the block size can be taken from the scatterer rather than
passed again — see "Why the block size is not a parameter" below.

This is the "first simple step" @garth-wells asked for on #4293:

> First simple step would be for `Vector` to have a constructor that takes a
> shared pointer to a scatterer.

@chrisrichardson's constraint on #3065 — "We still would want to be able to
create a `Vector` just from an `IndexMap` too" — is preserved by the delegating
form.

## What this is for

The constructor lets a caller decide which `Scatterer` a `Vector` uses, rather
than having one built for it. Two things follow from that.

**A caller can give a `Vector` a scatterer built on its own communication
pattern.** In #4484, scatterers over one `IndexMap` share a cached pattern, and
therefore a pair of neighbourhood communicators. Concurrent scatters in the
same direction on shared communicators are matched in start order, so every
rank must start them in the same order. A caller that cannot guarantee that
opts out by building its own pattern — and this constructor is the only way to
hand the result to a `Vector`:

```cpp
auto pattern = std::make_shared<const common::ScatterPattern>(*map);
auto sct = std::make_shared<const common::Scatterer<>>(pattern, bs);
la::Vector<double> v(map, sct);  // its own communicators
```

**Several vectors can share one scatterer**, avoiding the per-`Vector` rebuild
of the block size-expanded scatter indices.

## What changed since this PR was opened

It originally also added `la::Vector::clone_layout`, and argued the
communicator cost of `Scatterer` as its motivation. Both have been dropped.

**The motivation moved.** The original argument was that every `la::Vector`
built from an `IndexMap` costs two neighbourhood communicators, and that
sharing a scatterer avoids them. #4484 fixes that at the source, by caching the
communication pattern on the `IndexMap`, so `Vector(map, bs)` no longer creates
communicators at all. Reviewing this PR on the communicator argument would be
reviewing it for a reason that no longer holds.

**`clone_layout` went with it.** Its purpose was a vector over the same layout
with a different scalar type and no data copied — the copy-converting
constructor copies, and narrowing a value that does not fit the target type is
undefined behaviour, which `1e30` into an `std::int8_t` is not a hypothetical
for. With the pattern cached, that is simply:

```cpp
la::Vector<std::int8_t> marks(v.index_map(), v.bs());
```

which value-initialises rather than copying, and creates no communicator.
Verified equivalent at np = 3: same index map by pointer, same block size, same
size, same contents, and it scatters correctly.

What `clone_layout` still offered was inferring `Container` and
`ScatterContainer` from the source vector, so a device vector clones to a
device vector, and saving the index-array rebuild — measured at 16.9 µs versus
11.5 µs for a 2000-ghost, `bs = 3` vector. Ergonomics and microseconds, not
correctness, and not enough to carry the API plus its `rebind_container`
machinery. It is gone, along with `impl::rebind_container` and its tests.

## Why the block size is not a parameter

`Scatterer` took `bs`, expanded its indices for it, and discarded it — no
member, no accessor. So a `Vector` sharing a scatterer had to be told the block
size separately, making `(map, bs, scatterer)` a triple the caller must keep
consistent. Passing a scatterer built for a different block size compiled,
allocated the wrong storage and corrupted every scatter, with nothing checking
it. That is the same hazard this PR originally cited as its reason *not* to
expose a `scatterer()` accessor:

> The accessor would only make the consistency of the (map, bs, scatterer)
> triple a caller obligation, where passing a scatterer built for a different
> block size compiles and silently corrupts every scatter.

The argument applies just as well to a three-argument constructor. `Scatterer`
now stores the block size and reports it through `bs()`, and the constructor
takes `(map, scatterer)`. There is no second block size to disagree with.

A `Scatterer` is a communication pattern plus a block size — that is exactly
what #4484's split makes explicit — so being able to report which block size it
was built for is a property it should have had.

The index map stays a parameter: `Scatterer` does not hold one, and it must
not. #4484 has `IndexMap` own the communication pattern, so a pointer back
would be an uncollectable cycle.

## Relationship to #4293, #3065 and #4484

#4293 proposed associating a `Scatterer` with `fem::FunctionSpace`, and did not
converge: @garth-wells noted the scatterer "is really a property of the
`Vector`", @chrisrichardson raised `DofMap` without being convinced it was
better, and @schnellerhase suggested splitting `Scatterer` into a
topology-level communication pattern and a dofmap-level layout.

#4484 implements that split and closes #3065. This PR is deliberately
independent of where a shared scatterer should live: it only lets a caller
supply one.

## Tests

`test_vector_shared_scatterer` in `cpp/test/vector.cpp` builds two vectors of
different scalar types on one `Scatterer` and checks that the scatterer is
shared by reference count rather than copied, that the index map is carried
through, that both vectors report the scatterer's block size, and that both
scatter correctly on it.

Verified the reference count check is not vacuous: expecting 2 references
instead of 3 fails at np = 2.

Passing at np = 1, 2 and 3, with zero failures in both suites:

| --- | --- | --- |
| np = 1 | 83 cases | 3492 passed, 110 skipped, 21 xfailed |
| np = 2 | 83 cases | 2431 passed, 1089 skipped, 21 xfailed |
| np = 3 | 83 cases | 2432 passed, 1088 skipped, 21 xfailed |

---

### DIFF
--- a/cpp/dolfinx/common/Scatterer.h
+++ b/cpp/dolfinx/common/Scatterer.h
@@ -62,7 +62,7 @@ public:
   /// @param[in] bs Number of values associated with each `map` index
   /// (the block size).
   Scatterer(const IndexMap& map, int bs)
-      : _sizes_remote(map.src().size(), 0),
+      : _bs(bs), _sizes_remote(map.src().size(), 0),
         _displs_remote(map.src().size() + 1), _sizes_local(map.dest().size()),
         _displs_local(map.dest().size() + 1)
   {
@@ -215,7 +215,7 @@ public:
   /// @param s Scatterer to copy
   template <class U>
   Scatterer(const Scatterer<U>& s)
-      : _comm0(s._comm0), _comm1(s._comm1),
+      : _comm0(s._comm0), _comm1(s._comm1), _bs(s._bs),
         _remote_inds(s._remote_inds.begin(), s._remote_inds.end()),
         _sizes_remote(s._sizes_remote), _displs_remote(s._displs_remote),
         _local_inds(s._local_inds.begin(), s._local_inds.end()),
@@ -403,6 +403,15 @@ public:
   /// @return Indices container.
   const container_type& remote_indices() const noexcept { return _remote_inds; }
 
+  /// @brief Number of values associated with each index map index that
+  /// this scatterer was built for (the block size).
+  ///
+  /// The scatter indices are expanded for this block size, so a
+  /// scatterer can only be used with data of this block size.
+  ///
+  /// @return Block size.
+  int bs() const noexcept { return _bs; }
+
 private:
   // False only on a single rank, where _comm0/_comm1 stay MPI_COMM_NULL
   bool has_neighbours() const noexcept
@@ -430,6 +439,9 @@ private:
   // - in-edges (src) are from ranks that 'ghost' my owned indices
   // - out-edges (dest) are to the owning ranks of my ghost indices
   dolfinx::MPI::Comm _comm1{MPI_COMM_NULL};
+
+  // Block size the scatter indices are expanded for
+  int _bs;
 
   // Permutation indices used to pack and unpack ghost data (remote)
   container_type _remote_inds;
--- a/cpp/dolfinx/la/Vector.h
+++ b/cpp/dolfinx/la/Vector.h
@@ -123,17 +123,34 @@ public:
   static_assert(std::is_same_v<value_type, typename container_type::value_type>,
                 "Scalar type and container value type must be the same.");
 
+  /// @brief Create a distributed vector with a given communication
+  /// pattern.
+  ///
+  /// Sharing a scatterer avoids rebuilding a communication pattern, and
+  /// with it the two neighbourhood communicators that each Scatterer
+  /// creates.
+  ///
+  /// @param map Index map that describes the parallel layout of
+  /// the data.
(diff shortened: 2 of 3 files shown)


garth-wells (commented):
Could we get rid of bs in Scatter by using MPI_Type_contiguous (https://rookiehpc.org/mpi/docs/mpi_type_contiguous/index.html)? It would be nice if one Scatter could support different block sizes.

jorgensd:
This fix is also relevant for DOLFINx-adjoint, as one needs placeholder functions proper replays, and a ufl.replace keeps these Function's alive longer than needed (resolved by calling an intrusive gc.collect()) in the code. This PR would resolve this issue (https://github.com/scientificcomputing/dolfinx-adjoint/pull/88)

jhale:
Will be recreated after https://github.com/FEniCS/dolfinx/pull/4487.

PR #4484, opened 2026-09-10
Cache the scatter communication pattern on IndexMap

### LINKED ISSUE #3065: Associate a `common::Scatterer` with `fem::FunctionSpace`

jhale:
# Cache the scatter communication pattern on `IndexMap`

## Summary

Splits `common::Scatterer` at the block size boundary. Everything that needs
MPI — the two neighbourhood communicators, the source/destination ranks, the
neighbourhood sizes and displacements, the ghost permutation — moves into a new
`common::ScatterPattern`, which is built once per `IndexMap` and shared. What
remains in `Scatterer` is the block size-dependent part: two scaled size arrays
and two expanded index arrays, all cheap and none of it an MPI resource.

Result: **one communicator pair per distinct `IndexMap`**, however many
`Function`s, `Vector`s, block sizes or scatter container types exist.

Stacked on #4483, and targeted at its branch so the diff shows only this
change; retarget to `main` once #4483 merges. Merged with `main` after #4478
landed — see "Relationship to #4478" below.

Closes #3065, and addresses the neighbourhood-communicator half of #3061.

## Why

Every `Scatterer` constructor calls `MPI_Dist_graph_create_adjacent` twice, so
every `la::Vector` — and hence every `fem::Function` — costs two neighbourhood
communicators. As #3065 puts it, a code that creates a lot of `Function`s "can
create a lot of (neighbourhood) communicators, and exhaust the available
communicators". #3061 is a user hitting exactly that from hIPPYlib.

## What is split, and why here

`Scatterer`'s constructor on `main` spans 140 lines, and `bs` first appears in
the last 30 of them. Everything before that is a pure function of `map.comm()`, `map.src()`,
`map.dest()`, `map.owners()` and `map.ghosts()`:

- both `MPI_Dist_graph_create_adjacent` calls,
- the ghost permutation (radix sort by owning rank),
- the `MPI_Neighbor_alltoall`/`alltoallv` that fill the size and displacement
  arrays.

`bs` appears only in the tail, scaling sizes/displacements and expanding two
index arrays. So the scarce resource is determined entirely by the `IndexMap`,
and the `bs`-dependent remainder costs nothing scarce.

One small change makes the split clean: `ScatterPattern` stores the owned
shared indices as `recv_buffer[i] - range[0]`, i.e. local numbering at `bs = 1`.
The block size expansion of *both* index arrays is then the same
`idx[i] * bs + j`, and the pattern needs no index map after construction.

## Relationship to #4478

#4478 ("`common::Scatterer`: use neighbourhood collectives only") landed on
`main` while this was in review, and the two overlap heavily. They compose
cleanly, and the merge made this change *smaller*:

- `Scatterer` keeps everything #4478 gave it — no point-to-point path, separate
  `scatter_fwd_end`/`scatter_rev_end`, the `has_neighbours()` and `wait()`
  helpers. Only the constructor is replaced, and `has_neighbours()` now tests
  the pattern's communicator.
- #4478 removed `Scatterer`'s `_src`/`_dest` members, which were the only
  consumers of `ScatterPattern::src()`/`dest()`. Those accessors and members
  are gone here too, and the ranks are taken as local spans in the
  constructor — exactly what #4478 did to `Scatterer`. `ScatterPattern`'s
  public surface is smaller than first proposed as a result.

## Why the `IndexMap` and not the `FunctionSpace`, `DofMap` or `Vector`

This is the question #4293 stalled on. @garth-wells noted the scatterer "is
really a property of the `Vector`"; @chrisrichardson raised `DofMap` without
being convinced; @schnellerhase suggested splitting `Scatterer` into a
communication pattern and a layout. This PR is that suggestion, and it lets the
ownership question go unanswered: the pattern is a function of the `IndexMap`
alone, so the `IndexMap` is the only object that can own it without a
convention anyone has to enforce.

`IndexMap` deletes both copy operations, so a cache on it has no
copy-duplication or invalidation problem.

## Why the pattern is built on demand and not in the constructor

DOLFINx prefers fully-constructed objects, and this is the one piece of lazy
state in the library, so it needs justifying. I instrumented `IndexMap`'s and
`ScatterPattern`'s constructors and counted, at np = 3:

| Workload | `IndexMap`s built | patterns needed |
| --- | --- | --- |
| Poisson solve, 21 `Function`s, 20 forward scatters | 7 | 1 |
| 3x refinement, 3 spaces, 15 `Function`s, 1 matrix | 20 | 3 |

Most index maps are mesh and topology maps that are never scattered. Since
every `IndexMap` already dups its own communicator in its constructor
(`_comm(comm, true)`), the live communicator count for the second workload
works out as (measured map and pattern counts, arithmetic at 2 communicators
per `Scatterer` or per pattern):

| | live communicators |
| --- | --- |
| `main` | 20 + 2 x 15 = 50 |
| this PR | 20 + 2 x 3 = **26** |
| eager in the `IndexMap` constructor | 20 + 2 x 20 = 60 |

Building eagerly would consume *more* communicators than the code it is meant
to fix, on exactly the mesh-heavy codes that have the most index maps. So the
pattern is built on the first call to `IndexMap::scatter_pattern()`.

Laziness adds no collective-ordering requirement that does not already exist.
The collective fires at `Scatterer` construction, which is where one fires on
`main`; ranks that build scatterers for different maps in different orders
already deadlock today. The rationale is recorded in the header so it does not
have to be rediscovered.

The cache is not thread safe. There is no `std::mutex`, `once_flag` or
`MPI_THREAD` anywhere in `cpp/` or `python/`, so this matches existing practice;
a `once_flag` member would also break `IndexMap`'s defaulted move operations,
which `create_sub_index_map` (returns `IndexMap` by value) relies on.

## What this does not fix

The remaining 20 communicators above are the `MPI_Comm_dup` that each
`IndexMap` performs on its own communicator. That is the other half of #3061
and a separate axis; deliberately out of scope here.

Stripping the point-to-point scatter path was listed as out of scope here; it
landed independently as #4478 and this branch is merged with it.

## Overlapping non-blocking communication

#3065 records that each `Vector` having its own `Scatterer` "allows overlapping
non-blocking communication". Sharing the communicators keeps that, with one
documented constraint.

What still holds: every `Vector` owns its buffers and its `MPI_Request`, so
there is no shared state to corrupt, and forward and reverse scatters use
different communicators (`comm0` and `comm1`), so a forward and a reverse
scatter overlap with no ordering constraint between them.

What changes: concurrent scatters *in the same direction* now share a
communicator, and MPI matches nonblocking collectives on a communicator in the
order they are started. Every rank must therefore start them in the same order.
Previously the per-`Vector` communicators made the order irrelevant.

`test_scatter_overlap` puts four forward scatters in flight at once over one
index map and checks each vector receives its own data, then overlaps a forward
with a reverse scatter. Verified that issuing the four in a rank-dependent
order makes the test fail — the mismatch corrupts silently rather than
deadlocking, which is why the constraint is a `@note` on `Scatterer` rather
than left to be discovered.

In DOLFINx this order is fixed by the code path, which is identical on every
rank; the pattern that would break is issuing scatters in a data-dependent
order that varies by rank. No code in the library has two scatters in flight on
one index map: `MatrixCSR::mult` overlaps a *single* vector's scatter with
computation, which is unaffected. So the constraint is only reachable from user
code. It is the same constraint #4293 would introduce, and
`la::Vector::clone_layout` already carries a note to this effect.

### Opting out

A caller that cannot guarantee a rank-consistent order can take a private pair
of communicators back, recovering the pre-PR behaviour. This needs no new API —
`ScatterPattern`'s constructor is public, and #4483's `(map, scatterer)`
constructor accepts the result:

```cpp
auto pattern = std::make_shared<const common::ScatterPattern>(*map);
auto sct = std::make_shared<const common::Scatterer<>>(pattern, bs);
la::Vector<double> v(map, sct);  // its own communicators
```

`test_private_scatter_pattern` checks that the private pattern is a distinct
object on distinct communicators, that the vector takes no reference to the
map's shared pattern, and that it scatters correctly. So sharing is the
default, not the only option.

## Incidental fix

`MPI::Comm`'s copy constructor duplicates the communicator, so `Scatterer`'s
copy and cast-copy constructors each cost two `MPI_Comm_dup`s on `main`. They
now copy a `shared_ptr` instead. This is on the CPU to GPU path used by
`la::Vector::scatter_ptr`, where a copy-converted vector was paying for a
second pair of communicators describing the same graph.

## API

```cpp
// common::IndexMap — the shared pattern, built on first call
std::shared_ptr<const ScatterPattern> scatter_pattern() const;

// common::Scatterer, new primary constructor; performs no MPI.
// Throws std::invalid_argument on a null pattern or a block size < 1.
Scatterer(std::shared_ptr<const ScatterPattern> pattern, int bs);
Scatterer(const IndexMap& map, int bs);  // unchanged, now delegates
int bs() const noexcept;                 // added on #4483
```

`la::Vector` and the nanobind wrappers are unchanged, which was the check on
the design: if preserving the `Scatterer` interface had required touching
`Vector`, the split would have been in the wrong place. `ScatterPattern` is not
bound to Python. No `nb::keep_alive` is needed on the existing
`Scatterer(index_map, block_size)` binding, because the pattern is
heap-allocated and co-owned by the `Scatterer`, so a Python-owned `IndexMap`
collected first cannot dangle it.

## Tests

`test_scatter_pattern_shared` in `cpp/test/common/index_map.cpp` checks that
the pattern is returned by pointer identity rather than rebuilt, that
scatterers of different block size and different index container type share one
pattern, that the cast-copy constructor shares it rather than duplicating
communicators, and that 16 `la::Vector`s over one index map add exactly 16
references and no communicators — the last being the regression test for #3065,
expressed through `use_count()` so it needs no non-portable live-communicator
count.

Verified that the test fails if the implementation is wrong: making
`scatter_pattern()` rebuild on every call produces 10 failed assertions at
np = 2.

`test_scatter_rev` additionally covers `expand_indices` on a wider index
container directly, rather than only through the cast-copy constructor, and
covers the new block size and null pattern preconditions.

The existing coverage is otherwise unmodified and must stay that way —
`test_scatter_fwd`/`test_scatter_rev` (`GENERATE(1, 5, 10)` block sizes and the
`Scatterer<int32> -> Scatterer<int64>` cast-copy),
`test_scatter_with_isolated_rank` and `test_vector_scatter_rev` from #4478,
`test_vector_cast`, `test_vector_clone_layout`, and
`python/test/unit/common/test_scatterer.py`.

Passing at np = 1, 2 and 3, with zero failures in both suites:

| --- | --- | --- |
| np = 1 | 86 cases | 3492 passed, 110 skipped, 21 xfailed |
| np = 2 | 86 cases | 2431 passed, 1089 skipped, 21 xfailed |
| np = 3 | 86 cases | 2432 passed, 1088 skipped, 21 xfailed |

## Cost

Constructing a `Scatterer` on an index map, np = 3, through the Python binding:

| | per construct |
| --- | --- |
| first, builds the pattern | 67.5 us |
| subsequent, shares it | 0.5 us |

The communicator count, not the microseconds, is the point.

---

### DIFF
--- a/cpp/dolfinx/common/CMakeLists.txt
+++ b/cpp/dolfinx/common/CMakeLists.txt
@@ -11,6 +11,7 @@ set(
   ${CMAKE_CURRENT_SOURCE_DIR}/math.h
   ${CMAKE_CURRENT_SOURCE_DIR}/MPI.h
   ${CMAKE_CURRENT_SOURCE_DIR}/petsc.h
+  ${CMAKE_CURRENT_SOURCE_DIR}/ScatterPattern.h
   ${CMAKE_CURRENT_SOURCE_DIR}/Scatterer.h
   ${CMAKE_CURRENT_SOURCE_DIR}/Table.h
   ${CMAKE_CURRENT_SOURCE_DIR}/Timer.h
@@ -27,6 +28,7 @@ target_sources(
     log.cpp
     MPI.cpp
     petsc.cpp
+    ScatterPattern.cpp
     Table.cpp
     TimeLogger.cpp
     timing.cpp
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -1343,7 +1343,7 @@ std::span<const int> IndexMap::dest() const noexcept { return _dest; }
 std::shared_ptr<const ScatterPattern> IndexMap::scatter_pattern() const
 {
   if (!_scatter_pattern)
-    _scatter_pattern = std::make_shared<const ScatterPattern>(*this);
+    _scatter_pattern = create_scatter_pattern(*this);
   return _scatter_pattern;
 }
 //-----------------------------------------------------------------------------
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -1343,7 +1343,8 @@ std::span<const int> IndexMap::dest() const noexcept { return _dest; }
 std::shared_ptr<const ScatterPattern> IndexMap::scatter_pattern() const
 {
   if (!_scatter_pattern)
-    _scatter_pattern = create_scatter_pattern(*this);
+    _scatter_pattern
+        = std::make_shared<const ScatterPattern>(create_scatter_pattern(*this));
   return _scatter_pattern;
 }
 //-----------------------------------------------------------------------------
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -6,10 +6,12 @@
 // SPDX-License-Identifier:    LGPL-3.0-or-later
 
 #include "IndexMap.h"
+#include "ScatterPattern.h"
 #include "sort.h"
 #include <algorithm>
 #include <cstdint>
 #include <functional>
+#include <memory>
 #include <numeric>
 #include <ranges>
 #include <set>
@@ -1337,6 +1339,13 @@ std::vector<std::int32_t> IndexMap::shared_indices() const
 std::span<const int> IndexMap::src() const noexcept { return _src; }
 //-----------------------------------------------------------------------------
 std::span<const int> IndexMap::dest() const noexcept { return _dest; }
+//-----------------------------------------------------------------------------
+std::shared_ptr<const ScatterPattern> IndexMap::scatter_pattern() const
+{
+  if (!_scatter_pattern)
+    _scatter_pattern = std::make_shared<const ScatterPattern>(*this);
+  return _scatter_pattern;
+}
 //-----------------------------------------------------------------------------
 std::vector<std::int32_t> IndexMap::weights_src() const
 {
--- a/cpp/dolfinx/common/IndexMap.cpp
+++ b/cpp/dolfinx/common/IndexMap.cpp
@@ -1343,8 +1343,7 @@ std::span<const int> IndexMap::dest() const noexcept { return _dest; }
 std::shared_ptr<const ScatterPattern> IndexMap::scatter_pattern() const
 {
   if (!_scatter_pattern)
-    _scatter_pattern
-        = std::make_shared<const ScatterPattern>(create_scatter_pattern(*this));
+    _scatter_pattern = std::make_shared<const ScatterPattern>(*this);
   return _scatter_pattern;
 }
 //-----------------------------------------------------------------------------
--- a/cpp/dolfinx/common/IndexMap.h
+++ b/cpp/dolfinx/common/IndexMap.h
@@ -7,6 +7,7 @@
 #pragma once
 
 #include "MPI.h"
+#include "ScatterPattern.h"
 #include <cstdint>
 #include <memory>
 #include <span>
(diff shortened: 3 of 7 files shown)


jhale:
This is ready to review at a conceptual level - clearly some of the code and docstrings needs tidying.

jhale:
A much simpler version will be created after https://github.com/FEniCS/dolfinx/pull/4487.

PR #4486, opened 2026-09-11
Fix norm comparisons in Stokes demo

jorgensd:
Bug pointed out by: Léo Costa <leocosta@poli.ufrj.br>.
We returned the wrong norms for the mixed_direct_solver.
Furthermore, the variable naming was dubious.

For int32 builds, with mumps, we should remove the pressure nullspace for the comparison.

For int64 builds, with superlu_dist, which doesn't support null pivot detection, we do not assert anything.

### DIFF
--- a/python/demo/demo_stokes.py
+++ b/python/demo/demo_stokes.py
@@ -573,17 +573,17 @@ def mixed_direct():
 
     # No slip boundary condition
     W0 = W.sub(0)
-    Q, _ = W0.collapse()
-    noslip = Function(Q)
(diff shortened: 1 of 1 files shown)


jhale (commented):
Nothing wrong with the PR, looks good.

My broader concern is educational; should we be advising users to use superlu_dist at all on problems with a nullspace?

A better option might be to use the new Real space functionality to pin it. For an issue?

garth-wells (approved):

PR #4487, opened 2026-09-11
Move the block size out of common::Scatterer

jhale:
## Summary

Takes the block size out of `common::Scatterer`. A `Scatterer` currently
expands its indices, sizes and displacements for a block size fixed at
construction, so it can serve exactly one block size. Send blocks rather than
scalars, using a contiguous derived datatype, and that dependence goes away:

- The constructor takes only an `IndexMap`. Sizes, displacements, the ghost
  permutation and the owned shared indices are all in blocks.
- `scatter_fwd_begin` / `scatter_rev_begin` take a block size and build an
  `MPI_Type_contiguous` for the transfer. For `bs == 1` no type is created.
- `la::Vector` does the blocking in its pack/unpack, and sizes its buffers as
  `bs * blocks`.

This is @garth-wells' suggestion from review of #4484:

> I think the Scatterer PRs are getting too elaborate by overlooking a core
> issue: the scatter depends on the block size. If we use `MPI_Type_contiguous`
> the Scatterer becomes independent of the block size and simplifications
> follow.

## Why

A block size-free `Scatterer` is reusable across every `la::Vector` over an
index map, whatever their block sizes. That is the property #3065 and #3061
need — they are about a code creating many `Function`s and exhausting the
neighbourhood communicators, two of which every `Scatterer` creates.

This PR deliberately stops at making the `Scatterer` reusable. **It does not
decide who should reuse it**, and adds no cache or sharing of its own, so it
is independent of the open question on #4293 and #4484. `la::Vector` still
builds its own `Scatterer` per construction, exactly as on `main`.

## The datatype costs nothing

Measured at np = 3, 5000 blocks of 3 doubles:

| | derived type | scalar type |
| --- | --- | --- |
| `MPI_Neighbor_alltoallv` | 11.29 us | 11.56 us |
| `MPI_Ineighbor_alltoallv` (what `Scatterer` uses) | 12.12 us | 12.93 us |
| `MPI_Type_contiguous` + commit + free | 0.153 us | — |

The derived type is marginally faster in both, and creation is noise against a
~20 us scatter, so the type is built per call under RAII rather than cached. A
cached `static MPI_Datatype` was tried and rejected: it is never freed before
`MPI_Finalize` and is not thread safe, and it measured no faster.

## The pack/unpack dispatch is not optional

With block indices, the pack/unpack inner loop runs over the block size. Left
as a runtime value the compiler cannot unroll it, and that alone makes this
change a 2x regression. On the pack kernel in isolation, 5000 blocks:

| | |
| --- | --- |
| blocked, block size a compile-time constant | 3.57 us |
| blocked, block size a runtime value | 17.41 us |
| flat, pre-expanded indices (what this replaces) | 3.72 us |

`Vector::dispatch_bs` therefore calls the kernels with the block size as an
`std::integral_constant` for sizes 1, 2 and 3, and as a plain `int` otherwise.
That range matches the rest of the library — `la::MatrixCSR::mult` and
`fem/pack.h` dispatch 1, 2 and 3, and the assemblers dispatch 1 and 3. The
kernels are written once and take the block size as a value, so the same source
serves both paths.

Reviewers should know this is load-bearing: it looks like an optimisation and
is actually what makes the approach viable.

## Performance

Release build, np = 3, 100k owned and 5000 ghost indices, microseconds per
scatter:

| bs | fwd, this branch | fwd, main | rev, this branch | rev, main |
| --- | --- | --- | --- | --- |
| 1 | 10.4 | 10.1 | 10.7 | 10.2 |
| 2 | **15.0** | 15.9 | 17.1 | 17.4 |
| 3 | **20.1** | 22.4 | **22.7** | 24.5 |
| 4 | 32.1 | 28.3 | 34.0 | 32.5 |
| 9 | **52.4** | 64.5 | **56.0** | 71.5 |

`bs = 1` is a wash within run-to-run variation. Block sizes 2 and 3 are faster,
because one blocked index replaces `bs` expanded ones in the pack/unpack
gather.

**Block sizes 4 to 8 are the cost of this change**: they fall outside the
dispatched range, and `bs = 4` is about 14% slower on a forward scatter.
Extending the dispatch to 9 removes it entirely (`bs = 4` becomes 23.5 us,
faster than main), but no other dispatch site in the library goes past 3, so I
have kept the convention rather than the microseconds. Say if you would rather
have it the other way round. Note that `bs = 9` is already well ahead without
dispatch, because at large block sizes main's gather over `bs * N` expanded
indices is itself the bottleneck — so the regression window is only 4 to 8.

Benchmarking found two defects before these numbers were reached, both fixed
here: `scatter_rev` was slower than `main` at every block size, because
`get_unpack_op` had been left with a runtime trip count while the forward
kernels were dispatched; and block sizes outside the dispatched set fell off a
cliff (`bs = 4` at 38 us against main's 28 us) because the fallback used a
per-index `std::copy_n`, i.e. a `memmove` call per index.

## API changes

```cpp
// common::Scatterer
Scatterer(const IndexMap& map);                     // was (map, bs)
void scatter_fwd_begin(const T*, T*, int bs, MPI_Request&) const;
void scatter_rev_begin(const T*, T*, int bs, MPI_Request&) const;
const container_type& local_block_indices() const;  // was local_indices()
const container_type& remote_block_indices() const; // was remote_indices()
```

**The buffer size contract changes**, and the rename is there to enforce it.
The accessors return block indices now, so a buffer is `bs * size()` long
rather than `size()`. Adding the `bs` argument makes stale *calls* fail to
compile, but it would not have caught stale *sizing*:

```cpp
std::vector<T> send_buffer(sct.local_indices().size());   // bs times too small
```

That compiles, and MPI writes past the end of it — and the sizing usually sits
several lines from the call the compiler does flag. Renaming turns that silent
heap overflow into a compile error. No runtime check is possible: the scatter
functions take raw pointers so that device pointers work, so there is no size
to assert against.

Python-facing: `_cpp.common.Scatterer(index_map)`, and `scatter_fwd` /
`scatter_rev` take a block size.

`la::Vector`'s public interface is unchanged. It is also, on this branch, not
possible to give a `Vector` a scatterer whose block size disagrees with its
own: the `Scatterer` has no block size, `Vector` has no scatterer-taking
constructor and no `scatterer()` accessor, and the one `_bs` is used both to
size the buffers and to make the scatter call. On `main` the `Scatterer`
carries a block size that has to match the `Vector`'s, with nothing enforcing
it.

## Tests

`cpp/test/common/index_map.cpp` and `python/test/unit/common/test_scatterer.py`
are updated for the block indices and the block size argument; their coverage
is otherwise unchanged, including `GENERATE(1, 5, 10)` block sizes — note that
5 and 10 exercise the dispatched and fallback paths respectively.

Passing at np = 1, 2 and 3 in both Developer and Release builds, with zero
failures:

| --- | --- | --- |
| np = 1 | 82 cases | 3492 passed, 110 skipped, 21 xfailed |
| np = 2 | 82 cases | — |
| np = 3 | 82 cases | 2432 passed, 1088 skipped, 21 xfailed |

The Doxygen build is also clean (`cpp/doc/Doxyfile`, which treats warnings as
errors).

## Relationship to #4483 and #4484

#4484 caches a communication pattern on `IndexMap`, splitting `Scatterer` in
two along the block size boundary. If this lands, that split is unnecessary:
there is nothing left to separate, and the thing to cache is simply the
`Scatterer`. I would close #4484 in favour of a much smaller follow-up.

#4483 adds a constructor letting a caller supply a `Vector`'s `Scatterer`. It
survives, but its `Scatterer::bs()` accessor does not, and its constructor
reverts to `(map, bs, scatterer)` since the scatterer no longer knows the
block size.

One thing this PR does **not** address: a block size-free `Scatterer` is still
templated on the index container, so a GPU `Scatterer<device_vector>` cannot
share a CPU one and still duplicates the communicators. The block size axis and
the container axis are separate problems.

## Incidental cleanups

Two things this change made possible, in separate commits:

- **`dolfinx::MPI::Datatype<T>`.** The contiguous-datatype wrapper this PR
  needs already existed hand-rolled in `la/mattrans.h`: a raw `MPI_Datatype`, a
  conditional create and commit, and a conditional free 25 lines later, each
  guarded by its own copy of the `bs[0] * bs[1] == 1` test. Promoting the
  wrapper to `common/MPI.h`, beside `MPI::Comm` and following the same
  handle-owner pattern, lets `mattrans.h` drop both guards and the raw handle.
  It is templated on the value type rather than taking a base datatype,
  because every site that builds one builds it over something `mpi_t` already
  names. Like every other datatype creation in the library it is not error
  checked: these are local calls, there is no communicator to abort on, and
  under the default `MPI_ERRORS_ARE_FATAL` handler a failure aborts before a
  return code is visible. Four further sites create contiguous types by hand
  and could adopt it (`graph/partition.cpp` twice, `mesh/graphbuild.cpp`,
  `io/utils.h`, `common/MPI.h`); they are unconditional so they gain less, and
  are left alone.
- **`get_unpack` is now `get_unpack_op`** with an operation that keeps the
  received value. After the block size dispatch the two kernels were identical
  apart from that, so one of three near-identical kernels goes.

---

### DIFF
--- a/cpp/dolfinx/common/MPI.cpp
+++ b/cpp/dolfinx/common/MPI.cpp
@@ -70,45 +70,6 @@ dolfinx::MPI::Comm::operator=(dolfinx::MPI::Comm&& comm) noexcept
 //-----------------------------------------------------------------------------
 MPI_Comm dolfinx::MPI::Comm::comm() const noexcept { return _comm; }
 //-----------------------------------------------------------------------------
-dolfinx::MPI::Datatype::Datatype(int count, MPI_Datatype base) : _base(base)
-{
-  if (count > 1)
-  {
-    int err = MPI_Type_contiguous(count, base, &_type);
-    dolfinx::MPI::check_error(MPI_COMM_SELF, err);
-    err = MPI_Type_commit(&_type);
-    dolfinx::MPI::check_error(MPI_COMM_SELF, err);
-  }
-}
-//-----------------------------------------------------------------------------
-dolfinx::MPI::Datatype::Datatype(Datatype&& type) noexcept
-    : _type(type._type), _base(type._base)
-{
-  type._type = MPI_DATATYPE_NULL;
-}
-//-----------------------------------------------------------------------------
-dolfinx::MPI::Datatype::~Datatype()
-{
-  if (_type != MPI_DATATYPE_NULL)
-    MPI_Type_free(&_type);
-}
-//-----------------------------------------------------------------------------
-dolfinx::MPI::Datatype&
-dolfinx::MPI::Datatype::operator=(Datatype&& type) noexcept
-{
-  if (_type != MPI_DATATYPE_NULL)
-    MPI_Type_free(&_type);
-  _type = type._type;
-  _base = type._base;
-  type._type = MPI_DATATYPE_NULL;
-  return *this;
-}
-//-----------------------------------------------------------------------------
-MPI_Datatype dolfinx::MPI::Datatype::type() const noexcept
-{
-  return _type == MPI_DATATYPE_NULL ? _base : _type;
-}
-//-----------------------------------------------------------------------------
 int dolfinx::MPI::rank(const MPI_Comm comm)
 {
   int rank;
--- a/cpp/dolfinx/common/MPI.cpp
+++ b/cpp/dolfinx/common/MPI.cpp
@@ -70,6 +70,45 @@ dolfinx::MPI::Comm::operator=(dolfinx::MPI::Comm&& comm) noexcept
 //-----------------------------------------------------------------------------
 MPI_Comm dolfinx::MPI::Comm::comm() const noexcept { return _comm; }
 //-----------------------------------------------------------------------------
+dolfinx::MPI::Datatype::Datatype(int count, MPI_Datatype base) : _base(base)
+{
+  if (count > 1)
+  {
+    int err = MPI_Type_contiguous(count, base, &_type);
+    dolfinx::MPI::check_error(MPI_COMM_SELF, err);
+    err = MPI_Type_commit(&_type);
+    dolfinx::MPI::check_error(MPI_COMM_SELF, err);
+  }
+}
+//-----------------------------------------------------------------------------
+dolfinx::MPI::Datatype::Datatype(Datatype&& type) noexcept
+    : _type(type._type), _base(type._base)
+{
+  type._type = MPI_DATATYPE_NULL;
+}
+//-----------------------------------------------------------------------------
+dolfinx::MPI::Datatype::~Datatype()
+{
+  if (_type != MPI_DATATYPE_NULL)
+    MPI_Type_free(&_type);
+}
+//-----------------------------------------------------------------------------
+dolfinx::MPI::Datatype&
+dolfinx::MPI::Datatype::operator=(Datatype&& type) noexcept
+{
+  if (_type != MPI_DATATYPE_NULL)
+    MPI_Type_free(&_type);
+  _type = type._type;
+  _base = type._base;
+  type._type = MPI_DATATYPE_NULL;
+  return *this;
+}
+//-----------------------------------------------------------------------------
+MPI_Datatype dolfinx::MPI::Datatype::type() const noexcept
+{
+  return _type == MPI_DATATYPE_NULL ? _base : _type;
+}
+//-----------------------------------------------------------------------------
 int dolfinx::MPI::rank(const MPI_Comm comm)
 {
   int rank;
--- a/cpp/dolfinx/common/Scatterer.h
+++ b/cpp/dolfinx/common/Scatterer.h
@@ -240,8 +240,8 @@ public:
     if (!has_neighbours())
       return;
 
-    const block_type<T> type(bs);
-    MPI_Datatype dt = type.get();
+    const dolfinx::MPI::Datatype type(bs, dolfinx::MPI::mpi_t<T>);
+    MPI_Datatype dt = type.type();
     int ierr = MPI_Ineighbor_alltoallv(
         send_buffer, _sizes_local.data(), _displs_local.data(), dt, recv_buffer,
         _sizes_remote.data(), _displs_remote.data(), dt, _comm0.comm(),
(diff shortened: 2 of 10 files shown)


cpp/dolfinx/common/MPI.cpp
```diff
@@ -70,6 +70,45 @@ dolfinx::MPI::Comm::operator=(dolfinx::MPI::Comm&& comm) noexcept
 //-----------------------------------------------------------------------------
 MPI_Comm dolfinx::MPI::Comm::comm() const noexcept { return _comm; }
 //-----------------------------------------------------------------------------
+dolfinx::MPI::Datatype::Datatype(int count, MPI_Datatype base) : _base(base)
```
chrisrichardson: Could be a template, since second arg is pretty much always `dolfinx::MPI::mpi_t<T>`...
e.g. `dolfinx::MPI::DataType<T>(int count)` ?

chrisrichardson (approved):

cpp/dolfinx/common/Scatterer.h
```diff
@@ -247,20 +227,25 @@ class Scatterer
...
   /// passed to Scatterer::scatter_fwd_end to complete the communication.
   template <typename T>
-  void scatter_fwd_begin(const T* send_buffer, T* recv_buffer,
+  void scatter_fwd_begin(const T* send_buffer, T* recv_buffer, int bs,
                          MPI_Request& request) const
   {
     if (!has_neighbours())
       return;
 
+    const dolfinx::MPI::Datatype<T> type(bs);
```
schnellerhase, line 244: This will call `MPI_Type_contiguous` and thus register a new MPI type on every scatter operation.

chrisrichardson: Yes but it is very cheap.

jhale: I tried caching but the time difference with creating + RAII is noise vs the actual work.

jhale:
This looks pretty good to me so I will re-write the doc strings and comments for brevity/clarity, before moving on to rebasing the other ideas.

jhale:
To note that I deliberately renamed the method to '*_block_*' so that any downstream callers are aware of the API change.

cpp/dolfinx/common/Scatterer.h
schnellerhase: I am a bit worried about cases for when a lot of scattering happens, say during non-linear solves. Then everything depends on the specific MPI implementation on when types are cleaned up and properly discarded/recycled. Also it shares the pattern of registering a large number of MPI entities, which caused problems in  https://github.com/FEniCS/dolfinx/issues/3065 for communicators.

garth-wells (approved):

cpp/dolfinx/common/Scatterer.h
garth-wells: I've added interfaces where the caller supplies the `MPI_Datatype`. This can be used to avoid creating and destroying MPI data types. The block size version creates a `MPI_Datatype` and calls the version that takes a `MPI_Datatype`.

PR #4489, opened 2026-09-12
Reuse common::Scatterer in la::Vector and expose it to Python

### LINKED ISSUE #3061: Reopening the discussion regarding `MPI_Comm_dup`

garth-wells:
## Summary

- C++: `la::Vector` gains a constructor overload that reuses an existing `common::Scatterer`, so applications creating many vectors with the same parallel layout don't exhaust MPI communicators; the original constructor delegates to it. Documents the copy-conversion behavior.
- Python: `dolfinx.common.Scatterer` is exposed as a first-class Python object (a `scatterer()` factory) instead of requiring `dolfinx.cpp.common.Scatterer`, modeled directly on `common::Scatterer`:
  - `local_indices_block`/`remote_indices_block` properties expose the packing indices, mirroring the C++ methods of the same name.
  - `scatter_fwd_begin`/`scatter_fwd_end` and `scatter_rev_begin`/`scatter_rev_end` are thin, stateless methods that take/return a real `mpi4py.MPI.Request`; the caller packs/unpacks the buffers itself using the index properties, exactly as shown in `Scatterer.h`'s own worked example (also added to the C++ docs here).
  - A new `MPIRequestWrapper`, alongside the existing `MPICommWrapper` (now combined into a shared `mpi_wrappers.h`), lets an `MPI_Request` cross the Python/C++ boundary as a genuine `mpi4py.MPI.Request` rather than an opaque handle.
- `Vector.scatterer` and `vector(..., scatterer=...)` wire a shared scatterer through to `la::Vector`; `Vector.scatterer`/`Vector.petsc_vec` are now `functools.cached_property`, so the wrapped scatterer/PETSc vector are built once and stay the same object on repeat access.
- Fixed along the way (all caught before merge): a `vector()` positional-argument compatibility break flagged by CI (`dtype` is now keyword-only), a correctness bug where the reverse scatter's accumulation step silently dropped contributions when an owned entry was ghosted by more than one rank (only visible with ≥3 MPI ranks — `np.add.at` fixes it), and mypy overload-resolution errors on the new API.

Fixes #3061

## Testing

- C++: built with CMake+Ninja (Developer mode) and `ninja install`; `clang-format --dry-run --Werror` on changed C++ files.
- Python: built and installed against the above (`pip install --no-build-isolation -Ccmake.build-type=Developer`); ran `python/test/unit/common/test_scatterer.py` and `python/test/unit/la/test_vector_scatter.py` both serially and under `mpirun -np 3`/`-np 4` (the repeated-index accumulation bug only reproduces with ≥3 ranks); `ruff check`/`ruff format --check` and `mypy` on changed Python files.

AI assistance: OpenAI Codex and Claude Code were used to draft parts of this PR. I reviewed, edited, tested, and take responsibility for the final contribution.

### DIFF
--- a/cpp/dolfinx/la/Vector.h
+++ b/cpp/dolfinx/la/Vector.h
@@ -153,6 +153,13 @@ public:
 
   /// @brief Create a distributed vector.
   ///
+  /// This constructor creates a new Scatterer for the Vector. For
+  /// applications that create many Vectors with the same parallel layout,
+  /// constructing a distinct Scatterer for each Vector can exhaust the
+  /// available MPI communicators. This can be avoided by creating one
+  /// Scatterer and sharing it among those Vectors using the constructor that
+  /// takes a shared pointer to an existing Scatterer.
+  ///
   /// @param map Index map that describes the parallel layout of
   /// the data.
   /// @param bs Number of entries per index map 'index' (block size).
--- a/cpp/dolfinx/la/Vector.h
+++ b/cpp/dolfinx/la/Vector.h
@@ -450,31 +450,36 @@ public:
   }
 
   /// Get IndexMap
-  std::shared_ptr<const common::IndexMap> index_map() const { return _map; }
+  std::shared_ptr<const common::IndexMap> index_map() const noexcept
+  {
+    return _map;
+  }
 
   /// @brief Get the scatterer used for halo communication.
   /// @return The scatterer.
-  std::shared_ptr<const common::Scatterer<ScatterContainer>> scatterer() const
+  std::shared_ptr<const common::Scatterer<ScatterContainer>>
+  scatterer() const noexcept
   {
     return _scatterer;
   }
 
   /// Get block size
-  constexpr int bs() const { return _bs; }
+  constexpr int bs() const noexcept { return _bs; }
 
   /// @brief Get the process-local part of the vector.
   ///
   /// Owned entries appear first, followed by ghosted entries.
-  container_type& array() { return _x; }
+  container_type& array() noexcept { return _x; }
 
   /// @brief Get the process-local part of the vector (const version).
   ///
   /// Owned entries appear first, followed by ghosted entries.
-  const container_type& array() const { return _x; }
+  const container_type& array() const noexcept { return _x; }
 
   /// @deprecated Use ::array instead.
   /// @brief Get local part of the vector
-  [[deprecated("Use array() instead.")]] container_type& mutable_array()
+  [[deprecated("Use array() instead.")]] container_type&
+  mutable_array() noexcept
   {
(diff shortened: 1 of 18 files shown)


cpp/dolfinx/la/Vector.h
```diff
@@ -152,13 +153,33 @@ class Vector
...
+  /// @param[in] bs Number of entries per index map 'index' (block size).
+  /// @param[in] scatterer Scatterer compatible with `map`.
+  Vector(std::shared_ptr<const common::IndexMap> map, int bs,
+         std::shared_ptr<const common::Scatterer<ScatterContainer>> scatterer)
+      : _map(std::move(map)), _bs(bs),
+        _x(bs * (_map->size_local() + _map->num_ghosts())),
+        _scatterer(std::move(scatterer)),
         _buffer_local(bs * _scatterer->local_indices_block().size()),
         _buffer_remote(bs * _scatterer->remote_indices_block().size())
   {
```
jhale, line 185: Any O(1) checks on indexmap and scatterer compatibility that could go in here? Going to be easy to throw in a mismatched pair.

cpp/dolfinx/la/Vector.h
```diff
@@ -422,24 +450,36 @@ class Vector
...
   /// @brief Get the process-local part of the vector (const version).
   ///
   /// Owned entries appear first, followed by ghosted entries.
-  const container_type& array() const { return _x; }
+  const container_type& array() const noexcept { return _x; }
 
   /// @deprecated Use ::array instead.
   /// @brief Get local part of the vector
-  [[deprecated("Use array() instead.")]] container_type& mutable_array()
+  [[deprecated("Use array() instead.")]] container_type&
```
jhale, line 481: Good time to remove?

PR #4490, opened 2026-09-12
Add one-shot common::scatter_fwd/scatter_rev helpers

jhale:
## Summary

The pack → `scatter_fwd_begin` → `scatter_fwd_end` → unpack sequence around
`common::Scatterer` is written out by hand at every call site. This adds two
free functions in `dolfinx::common` that do it in one call for contiguous host
data, and uses them where the hand-rolled version bought nothing:

- `refinement/uniform.cpp` — fwd (`std::int64_t`), 18 lines → 3.
- `wrappers/dolfinx_wrappers/common.h` — fwd + rev (`int64`/`float`/`double`),
  both lambda bodies now delegate.

```cpp
template <typename T, class Container>
  requires requires(Container c) {
    { c.data() } -> std::same_as<typename Container::value_type*>;
  }
void scatter_fwd(const Scatterer<Container>& sc, std::span<const T> local_data,
                 std::span<T> remote_data, int bs);

template <typename T, class Container, typename BinaryOperation>
  requires requires(Container c) {
    { c.data() } -> std::same_as<typename Container::value_type*>;
  }
void scatter_rev(const Scatterer<Container>& sc, std::span<T> local_data,
                 std::span<const T> remote_data, int bs, BinaryOperation op);
```

This supersedes #4482, which was written before #4487 moved the block size out
of `common::Scatterer`.

## Design notes

- **Free functions, not members.** A one-shot member must allocate buffers
  internally, contradicting the documented contract that a `Scatterer` is
  stateless. A free function taking `const Scatterer&` leaves the caller owning
  the `Scatterer`, so its construction can still be amortised.
- **Two spans, not array + offset.** `Scatterer` stores no `IndexMap`, so an
  offset would have to be recomputed by every caller. This also matches the
  existing Python binding signature exactly, so `test_scatterer.py` passes
  unchanged. A caller with one contiguous array splits it with `x.first(n)` /
  `x.subspan(n)`.
- **The `Container` constraint** is the one `la::Vector` uses to select its CPU
  overloads: raw-pointer storage, which GPU device containers lack, so a
  `thrust::device_vector` is rejected at the constraint rather than deep in the
  body. It is a capability check, not a memory-space check — host residency of
  the spans stays a documented caller precondition.
- **`la::Vector` is untouched**, deliberately: it allocates its buffers once in
  the constructor (a per-call allocation would land on every solver iteration),
  exposes `begin`/`end` separately for comm/compute overlap, is generic over
  device storage that `std::span` cannot express, and unpacks at
  `_bs * size_local()`, which it knows and the helper does not.
- **No regression.** Both converted sites already allocated exactly these two
  buffers per call; the allocation moved rather than appeared.

## Tests

- `test_scatterer.py` passes unchanged — its calls now route through the
  helpers, covering `int64`/`float32`/`float64`, both directions, serial and
  parallel, with no new test code.
- One new `[index_map_scatter_helpers]` case in `cpp/test/common/index_map.cpp`
  for `std::int8_t` with `bs = 2` — a narrow signed type, and the only host type
  the bindings do not instantiate.

Verified: full C++ suite on 1, 2 and 3 ranks; `pytest -k "scatterer or refine"`
(81 tests) serial and on 3 ranks; `clang-format`, `ruff`, `ruff format`,
`gersemi` clean. mypy matches its pre-existing baseline — no Python source is
touched.

---

AI assistance: I used Claude Code to draft parts of this PR. I reviewed,

### DIFF
--- a/cpp/dolfinx/common/Scatterer.h
+++ b/cpp/dolfinx/common/Scatterer.h
@@ -1,4 +1,4 @@
-// Copyright (C) 2022-2026 Igor Baratta and Garth N. Wells
+// Copyright (C) 2022-2026 Igor Baratta, Garth N. Wells and Jack S. Hale
 //
 // This file is part of DOLFINx (https://www.fenicsproject.org)
 //
@@ -12,6 +12,8 @@
 #include <algorithm>
 #include <array>
 #include <cassert>
+#include <concepts>
+#include <cstddef>
 #include <cstdint>
 #include <mpi.h>
 #include <numeric>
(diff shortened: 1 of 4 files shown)


jhale:
Closing. #4491 replaced the one-shot Python `scatter_fwd`/`scatter_rev` bindings with split `scatter_*_begin`/`scatter_*_end` taking caller-packed buffers, so that call site is gone. With only `refinement::uniform_refine` left, the helpers don't earn their place.

PR #4493, opened 2026-09-13
common: convert O(1) debug asserts to unconditional throws

jhale:
local_range, MPI::distribute_data and sort_by_perm validated cheap,
O(1) preconditions (positive size/shape1, size relationships) only via
assert, which compiles out under NDEBUG. Bad arguments then caused a
division by zero or an out-of-bounds read in Release builds instead of
a diagnostic. Per AGENTS.md, an O(1) check should be an unconditional
throw rather than assert-only, so convert these to
std::invalid_argument and add Catch2 coverage.

MPI::index_owner and math::det/inv/pinv were also surveyed but are
called per-element inside hot loops (mesh partitioning, per-quadrature
-point Jacobian computation), so they are intentionally left as
assert-only per the "no exceptions in hot loops" rule. IndexMap.cpp is
excluded from this pass as it is being hardened separately.

AI assistance: I used Claude to draft this PR. I reviewed, edited,
tested, and take responsibility for the final contribution.

---


cpp/dolfinx/common/local_range.h
```diff
@@ -26,9 +26,12 @@ namespace dolfinx::common
 constexpr std::array<std::int64_t, 2> local_range(int index, std::int64_t N,
                                                   int size)
 {
-  assert(index >= 0);
-  assert(N >= 0);
-  assert(size > 0);
+  if (index < 0)
```
jhale, line 29: Changes justifiable only if not called in tight loop, need to check.

cpp/test/common/utils.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include <catch2/catch_test_macros.hpp>
+#include <dolfinx/common/MPI.h>
+#include <dolfinx/common/local_range.h>
+#include <mpi.h>
+#include <stdexcept>
+#include <vector>
+
+using namespace dolfinx;
+
+TEST_CASE("local_range bad arguments", "[local_range]")
```
jhale: Certainly over testing, will remove.

garth-wells (approved):

cpp/dolfinx/common/local_range.h
jhale: Checked, not called in tight loop.

PR #4496, opened 2026-09-13
Wrap IndexMap in a Python class

garth-wells:
Stacked on #4350, and continuing the same theme: `common.IndexMap` was a bare `dolfinx.cpp.common` re-export, so every index map handed out by the Python interface was a raw nanobind object, and building one meant calling the C++ constructor overloads directly.

### `common.IndexMap`

A Python class wrapping `comm`, `size_local`, `size_global`, `num_ghosts`, `local_range`, `ghosts`, `owners`, `index_to_dest_ranks`, `local_to_global` and `global_to_local`.

It is returned from `Topology.index_map`, `Topology.index_maps`, `Geometry.index_map`, `DofMap.index_map`, `Vector.index_map`, `MatrixCSR.index_map` and `SparsityPattern.index_map` — the last of these being the gap noted in #4350.

Like the other wrappers, it is rebuilt on each access, so it defines `__eq__`/`__hash__` delegating to the wrapped object; `Scatterer` gains the same. Without it, `V.dofmap.index_map == W.dofmap.index_map` compares two fresh Python objects and is always `False`.

### `common.index_map`

The C++ class has three constructors distinguished by arity, which do not map onto a single Python signature cleanly. They are exposed as

```python
index_map(comm, local_size, ghosts=None, *, dest_src=None, tag=...)
```

`ghosts` is a `(ghost_indices, owners)` pair rather than two arguments: the two arrays are meaningless apart and must have the same length, so pairing them removes the 'one given without the other' error case instead of checking for it. `dest_src` and `tag` are keyword-only — a positional list-of-ranks and a positional tag are easy to confuse at a call site, and the old C++ signatures put them in different positions. Passing `dest_src` skips the consensus exchange that otherwise discovers which ranks ghost the caller's owned indices.

All three constructors are collective, which shapes the contract: `ghosts` must be `None` on every process (a non-overlapping map) or given on every process, a process with no ghosts passing empty arrays, and `dest_src` likewise all-or-nothing. I first enforced that with an `allreduce` over each flag, and backed it out: AGENTS.md requires validation in Release builds to be local, so a collective check here would add communication to every call to catch a caller error. The requirement is documented on the arguments instead, and mismatched participation is a precondition violation, as it is on the C++ side.

`tag` defaults to `dolfinx::MPI::tag::consensus_nbx`. The wrapper exports that constant as `dolfinx.cpp.common.consensus_nbx_tag` — a six-line addition to `wrappers/common.cpp`, and the only C++ change in this PR — rather than repeating `1202` in Python where nothing would catch the two drifting apart.

### `common.create_sub_index_map`

Was only reachable as `dolfinx.cpp.common.create_sub_index_map`; now has a Python counterpart taking and returning the wrapper. It follows the contract introduced in #4479: no `allow_owner_change` argument, and a third return value reporting whether an index acquired a new owner, documented as rank-local and needing a reduction before it can drive a collective decision.

### Call sites

`la.sparsity_pattern`, `la.sparsity_pattern_blocked`, `la.vector`, `la.petsc.create_vector`, `graph.comm_graph`, `mesh.mark_maximum`, `mesh.mark_equidistribution`, `mesh.create_geometry`, `mesh.create_point_mesh`, `common.scatterer` and the blocked PETSc index-set construction in `fem/petsc.py` take or pass `_cpp_object` as needed. The tests that constructed index maps directly use `common.index_map`.

One caller-visible consequence: `fem.petsc.create_vector` accepted a raw `dolfinx.cpp.fem.FunctionSpace_float32/64` alongside the Python `FunctionSpace`. Its dofmap yields a raw C++ index map, which `la.petsc.create_vector` can no longer unwrap, so it now accepts only the Python type — as its type hint, and the guidance against using `dolfinx.cpp` directly, already said.

### Testing

`python/test/unit` passes in serial (3563 passed, 81 skipped, 24 xfailed) and under `mpirun -n 2` and `-n 3`. The three `cffi` tests error in my checkout because setuptools refuses to read `../README.md` from outside `python/`; that is unrelated to this PR and reproduces on `main`. The demos that exercise these paths — `demo_elasticity`, `demo_poisson-matrix-free`, `demo_mixed-topology`, `demo_matrix-free-petsc`, `demo_static-condensation`, `demo_hdg`, `demo_partition` — run.

---

Review the base PR #4350 first; this PR's diff against `main` includes it.


python/dolfinx/la/__init__.py
```diff
@@ -88,10 +87,10 @@ def __del__(self) -> None:
         if (petsc_x := self.__dict__.get("petsc_vec")) is not None:
             petsc_x.destroy()
 
-    @property
+    @functools.cached_property
```
jhale, line 90: Add note on cached property to docstring, has state and easy to miss.

python/dolfinx/common.py
```diff
@@ -61,11 +62,131 @@
...
+        """
+        self._cpp_object = imap
+
+    def __eq__(self, other: object) -> bool:
+        """Check that two wrappers hold the same index map."""
+        if not isinstance(other, IndexMap):
+            return NotImplemented
+        return self._cpp_object == other._cpp_object
+
+    def __hash__(self) -> int:
```
jhale: What's going into this hash? Not completely clear.

python/test/unit/common/test_index_map.py
```diff
@@ -270,14 +268,27 @@ def test_sub_index_map_multiple_possible_owners():
...
     # `owners_changed` is rank-local; reduce as a caller would.
     assert comm.allreduce(owners_changed, op=MPI.LOR)
 
     assert sub_imap.size_global == 3
     assert sub_imap.size_local == submap_size_local_expected
     assert sub_imap.num_ghosts == submap_num_ghosts_expected
+
+
+def test_index_map_equality():
+    """Index maps for the same C++ object compare (and hash) equal."""
```
jhale: I don't see an explicit check on hashing `__hash__`, looks like `__eq__` check?

python/dolfinx/common.py
```diff
@@ -61,11 +62,131 @@
...
+            User code should call :func:`index_map` to create an index
+            map.
+
+        Args:
+            imap: C++ IndexMap object.
+        """
+        self._cpp_object = imap
+
+    def __eq__(self, other: object) -> bool:
+        """Check that two wrappers hold the same index map."""
```
jhale: Reword, perhaps "Check that two wrappers hold the same underlying C++ index map". Otherwise equality might mean "two separately-constructed maps with identical meaning compare equal".

python/test/unit/common/test_index_map.py
```diff
@@ -270,14 +268,27 @@ def test_sub_index_map_multiple_possible_owners():
...
+    msh = create_unit_square(MPI.COMM_WORLD, 3, 3)
+    V = functionspace(msh, ("Lagrange", 1))
+    imap = V.dofmap.index_map
+    assert imap == V.dofmap.index_map
+    assert len({imap, V.dofmap.index_map}) == 1
```
jhale, line 293: Add an explicit comment here that this is the line checking the hash function (by constructing a dictionary) works? It's easy to miss as is.

python/dolfinx/common.py
```diff
@@ -241,6 +372,85 @@ def scatter_rev_end(self, request: _MPI.Request) -> None:
...
+    """Create an index map.
+
+    Args:
+        comm: MPI communicator to distribute the indices over.
+        local_size: Number of indices owned by the calling process.
+        ghosts: Tuple ``(ghost_indices, owners)`` of global ghost
+            indices and their owning ranks. If ``None``, the index map
+            is non-overlapping and ``ghosts`` must be ``None`` on every
+            process. For an overlapping map, a process with no ghosts
+            must pass empty arrays.
```
jhale, line 404: Presumably the homogeneity across ranks (None or Array) cannot be checked as it requires communication - might be worth mentioning its not explicitly checked?

python/dolfinx/fem/dofmap.py
```diff
@@ -69,10 +70,10 @@ def dof_layout(self) -> ElementDofLayout:
         """Layout of dofs on an element."""
         return ElementDofLayout(self._cpp_object.dof_layout)
 
-    @property
-    def index_map(self) -> _cpp.common.IndexMap:
+    @cached_property
```
jhale: `functools.cached_property` easier to read.

jhale: Add note to docstring on it being a cached property, easy to miss.

python/dolfinx/la/__init__.py
garth-wells: Done — added a `Note:` that the property caches, in 891162ce62. Did the same for `Vector.scatterer` just below, which was already a cached property without a note.

python/dolfinx/common.py
garth-wells: Used your wording, and added a note making the other reading explicit: "This is identity of the wrapped object, not equivalence of the distribution it describes. Two separately constructed index maps do not compare equal, even if identical."

python/test/unit/common/test_index_map.py
garth-wells: Added: "Building a set exercises `__hash__`: equal wrappers must hash equal and collapse to a single entry."

python/dolfinx/common.py
garth-wells: Right, it cannot be checked locally. Added a `Note:` at the top of the docstring: collective, `ghosts` is `None` on every process or given on every process (likewise `dest_src`), and this is a precondition that is not checked since checking it would require communication on every call.

python/dolfinx/fem/dofmap.py
garth-wells: Changed to `import functools` / `functools.cached_property`, matching `la/__init__.py`. Also changed the pre-existing `dof_layout` in the same file so the two are not spelled differently.

garth-wells: Added, same wording as the one in `la/__init__.py`.

jhale (approved):

PR #4500, opened 2026-09-13
Switch CI type checking from mypy to Pyrefly

garth-wells:
## Summary

- replace the CI mypy job and dependency with Pyrefly
- add Pyrefly configuration for the Python package, demos, and tests
- baseline the 58 existing migration diagnostics so new diagnostics still fail CI
- update type-checking documentation and repository guidance
- preserve the generated-stub mypy suppressions for downstream consumers

## Validation

- Pyrefly checks pass in the openmpi / PETSc float64 / int32 CI leg
- loaded the configuration with Pyrefly 1.3.0
- validated the baseline JSON
- parsed .github/workflows/ci.yml as YAML
- ran Ruff check and format checks on the touched Python source
- ran Gersemi on python/CMakeLists.txt (only the existing unknown nanobind command warnings)
- ran git diff --check

## AI assistance

AI assistance: OpenAI Codex drafted parts of this PR. Human review is pending before marking it ready.


jhale:
Looks good, mypy was missing some odd things. Is it possible to put the config largely back into pyproject? Prefer the single file approach.

jhale (commented):
Looks good, very close to DOLFINx being fully typed.

PR #4503, opened 2026-09-13
Add -Wshadow to Developer compiler flags

jhale:
Enables `-Wshadow` for the `Developer` build type, guarded by
`check_cxx_compiler_flag` like the other flags in
`DolfinxDeveloperCompilerFlags.cmake`. Since `-Werror` is already on, this
required fixing the 41 shadowing sites it reports.

Most are redundant re-declarations, removed rather than renamed:

- inner `int err = MPI_...` re-declaring the enclosing `err` (`common/MPI.cpp`)
- a second, identical `MPI_Dist_graph_neighbors_count` block in `refinement/utils.h`
- an unused `std::vector simplex_set` in `refinement/plaza.h`
- a duplicate `cell_type` in `io/xdmf_mesh.cpp`

Nested repetitions of `auto [unique_end, range_end] = std::ranges::unique(x); x.erase(...)`
become the equivalent one-liner `x.erase(std::ranges::unique(x).begin(), x.end())`.

The rest, where both names are live, are renamed: a local mdspan named `_x`
shadowing the `Function::_x` member, lambda parameters in `io/utils.h` shadowing
the outer `[nodes_g_p, recv_disp, src, dest]` binding, nested
`timer`/`ranks`/`vtu`/`piece_node`/`u`/`n`/`i` locals, a `[type, kernels]`
binding shadowing `declare_form`'s `type` parameter in the nanobind layer, and a
lambda-local `using U` shadowing the file-scope `using U` in three C++ demos.

No functional change intended.

## Testing

Built in `Developer` mode on macOS/AppleClang 17: C++ library, nanobind
wrappers, `cpp/test` and all `cpp/demo` targets compile clean with the new flag.
C++ unit tests pass (92 cases, 29845 assertions, 3 skipped).
`clang-format --dry-run --Werror` and `gersemi --check` are clean.


jhale:
Think this is worth having on, picks up near-miss programmer mistakes like this automatically:

```
    int tdim = mesh->topology()->dim();
    int fdim = tdim - 1;

    auto submesh_data = [](auto& mesh, int tdim, auto&& dfacets)
    {
      auto [submesh, e_map, v_map, g_map]
          = mesh::create_submesh(mesh, tdim, dfacets);
      return std::pair(std::make_shared<mesh::Mesh<U>>(std::move(submesh)),
                       std::move(e_map));
    };
```

PR #4508, opened 2026-09-14
python: remove dolfinx.cpp usage from the test suite

garth-wells:
## Summary

Extends the existing demo convention (`AGENTS.md`: "Users and developers are discouraged from using `dolfinx.cpp` directly") to `python/test/`. Before this PR, 20 test files imported `dolfinx.cpp` directly or reached into it via `._cpp_object`; this brings that down to 6, each with a documented reason.

**Pure-Python API additions** (`python/dolfinx/`), added to close real gaps that were forcing tests into the nanobind layer:
- `mesh._wrap_partitioner`: a user-supplied partitioner callback may now return a `dolfinx.graph.AdjacencyList` instead of a raw nanobind `AdjacencyList_int32` (previously required at 6 call sites, one carrying a `TODO: can we improve on this interface?`). Opaque C++ partitioner handles (e.g. `graph.partitioner_scotch()`) are passed through unchanged so the fast path isn't lost.
- `mesh.uniform_refine`: no longer raises when refining a mesh with no UFL domain (e.g. a mixed-topology mesh) — returns an undecorated `Mesh` instead.
- `mesh`: newly wrapped `cell_normals`; re-exported `is_simplex`, `cell_entity_type`, `cell_num_vertices`.
- `log`: re-exported `set_thread_name`.
- `fem.set_diagonal` / `fem.set_bc_diagonal`: public wrappers for the two `_cpp.fem.insert_diagonal` overloads (previously internal-only, used directly by tests). Named after the underlying `fem::set_diagonal` rather than `insert_diagonal`, since `la.SparsityPattern.insert_diagonal` already exists with a different (structural, not value) meaning.
- `fem.assemble_matrix_fn`: the insertion-callback form of `assemble_matrix`, kept as a separate function rather than a `functools.singledispatch` registration (a UFL form is itself callable, so registering on `Callable` would break dispatch).

**Test suite**: mechanical swaps onto the above plus already-existing pure-Python equivalents (`cell_dim`, `io.utils.cell_perm_vtk`, `la.BlockMode`, `mesh.build_dual_graph`, `MatrixCSR.index_map`, `fem.form_cpp_class`, ...). Also deleted three permanently-skipped tests (`test_cell_inradius`, `test_cell_circumradius`, `test_cell_radius_ratio` in `test_mesh.py`, `test_area` in `test_face.py`) — they exercised `inradius`/`circumradius`/`radius_ratio`/`volume_entities`, which are no longer bound in C++ at all, so there was nothing to port them to.

**Out of scope**: mixed-topology mesh construction (`create_topology`, list-of-elements `create_mesh`, mixed `create_geometry`, ...) has no pure-Python API yet — that's a larger, separate piece of work. The tests that need it, plus one demo, keep a documented `dolfinx.cpp` exemption:
- `test/unit/conftest.py`, `mesh/test_mixed_topology.py`, `mesh/test_create_mixed_mesh.py`, `fem/test_mixed_mesh_dofmap.py`, `io/test_vtkhdf.py`, `demo_mixed-topology.py`
- `test/unit/common/test_version.py` (deliberately compares the C++ and Python package versions)

A few files use only `._cpp_object` (not `dolfinx.cpp` imports) and are intentionally left alone since they test the binding/wrapper layer itself: `fem/test_forms.py`, `fem/test_interpolate_geometry.py`, `la/test_vector_scatter.py`, and one nanobind-kwarg-name check in `fem/test_interpolation.py:1052`.

**Enforcement**: added a ruff `TID251` banned-api rule for `dolfinx.cpp` in `python/pyproject.toml`, scoped away from `python/dolfinx/**` (the wrapping layer itself) via `per-file-ignores`, with the exemptions above each given a one-line reason. Verified empirically: `ruff check` passes clean across the whole `python/` tree, and temporarily removing one exemption reproduces the expected violation. Note `._cpp_object` access is a plain attribute access, not an import, so it isn't (and can't be) caught by this rule — `SLF001` was considered and rejected as far too noisy (100+ unrelated hits).

## Test plan

- [x] `ruff check python/` and `ruff format --check python/` — clean
- [x] All touched files pass `python3 -m py_compile`
- [ ] Full `pytest python/test/unit` (serial) — not run; see note below
- [ ] Parallel run (`mpirun -n 2`/`-n 3`) for the partitioner/dofmap/adios2/matrix_csr tests, since `_wrap_partitioner` changes what crosses the Python↔C++ partitioner boundary and partitioning is only exercised in parallel
- [ ] Spot-check that built-in partitioner handles (e.g. `graph.partitioner_scotch()`) still take the fast opaque-handle path through `create_mesh` after `_wrap_partitioner`, not a Python-closure path

I could not build/run the Python test suite locally — the installed `dolfinx.cpp` extension fails to load in this environment due to a pre-existing, unrelated `libadios2_cxx_mpi` version mismatch (confirmed present before this branch). I did confirm `pyrefly check` (informational, since it also can't resolve `dolfinx.cpp` here) reports *fewer* errors after this change (57 vs. 74 on unmodified `main`), consistent with removing unresolvable `dolfinx.cpp` imports and introducing no new ones in the touched files.

AI assistance: I used Claude Sonnet 5 and Opus 5 to identify the `dolfinx.cpp` usage, design and implement the pure-Python API additions, migrate the test suite, and add the ruff enforcement rule. I reviewed the diff, verified the enforcement rule empirically, and take responsibility for the final contribution.


jhale:
Going to merge this, useful for enforcing requirements.

PR #4511, opened 2026-09-15
Fix FiniteElement definitions for manifolds

### LINKED ISSUE #3619: Interpolation from RT space on manifold broken
### Summarize the issue

The value check for interpolating on manifolds are wrong.
Introduced in: https://github.com/FEniCS/dolfinx/pull/3500.

### How to reproduce the bug

The following example runs on v0.9.0, but not nightly

### Minimal Example (Python)

```Python
from mpi4py import MPI
import numpy as np
import basix.ufl
import dolfinx
import ufl

points = np.array([[0.0, 0.0, 0.0], [1.0, 0, 1.0], [1.0, 1.0, 0.0]])
cells = np.array([[0, 1, 2]], dtype=np.int32)
ud = ufl.Mesh(basix.ufl.element("Lagrange", "triangle", 1, shape=(3,)))
mesh = dolfinx.mesh.create_mesh(MPI.COMM_WORLD, cells, points, ud)


el = basix.ufl.element("RT", mesh.basix_cell(), 2)
w = dolfinx.fem.Function(V)


el_o = basix.ufl.element("DG", mesh.basix_cell(), 1, shape=(mesh.geometry.dim,))
Q = dolfinx.fem.functionspace(mesh, el_o)
q = dolfinx.fem.Function(Q)
q.interpolate(w)
```

### Output (Python)

```bash
File "/root/shared/examples/eikonal/debug_interpolate_rt_manifold.py", line 21, in <module>
    q.interpolate(w)
  File "/usr/local/dolfinx-real/lib/python3.12/dist-packages/dolfinx/fem/function.py", line 460, in interpolate
    _interpolate(u0)
  File "/usr/lib/python3.12/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/dolfinx-real/lib/python3.12/dist-packages/dolfinx/fem/function.py", line 446, in _
    self._cpp_object.interpolate(u0._cpp_object, cells0, cells1)  # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Interpolation: elements have different value dimensions
```

### Version

main branch

### DOLFINx git commit

d45b65f6fff295e575f774385b8172054f36d908

### Installation

Running DOLFINx nightly container

### Additional information

_No response_

jorgensd:
# Value shapes and sizes on `fem::FiniteElement`

Fixes #3619.

## The problem

A basis is tabulated on the **reference** cell and pushed forward to a
**physical** cell. For a Piola-mapped element the push-forward contracts value
axes with the Jacobian `J`, which has shape `(gdim, tdim)`. Those axes therefore
have extent `gdim` in physical space and `tdim` on the reference cell — equal
only when `gdim == tdim`.

#3500 removed `fem::compute_value_shape` and `FunctionSpace::value_shape()`, the
only `gdim`-aware path in C++, moved `value_shape()` onto `fem::FiniteElement`
and initialised it from `basix::FiniteElement::value_shape()` — which is the
*reference* shape. `fem::FiniteElement` never saw a mesh, so on a manifold it
reported `tdim` components where UFL, FFCx and the Python layer all reported
`gdim`. Every value-shape check then compared a reference shape against a
physical one.

## The six quantities

| accessor | meaning |
| --- | --- |
| `value_shape()` | shape of the field on a **physical** cell. What a user sees, what `ufl.FunctionSpace.value_shape` reports, and what two spaces must agree on to be interpolation-compatible. |
| `value_size()` | product of `value_shape()`. |
| `reference_value_shape()` | shape Basix tabulates in. The shape of the data returned by `tabulate()`. |
| `reference_value_size()` | product of `reference_value_shape()`. |
| `block_size()` | number of dofs collocated at a dof point. |
| `base_value_size()` | **new.** Number of physical components in *one block*: the size of the push-forward of a single non-blocked basis function. |

`value_shape()` and `reference_value_shape()` differ for three independent
reasons: blocking, quadrature elements, and the map (the last one only on a
manifold).

## Worked examples

All on a mesh of triangles, so `tdim == 2`. `RT` is Raviart-Thomas, `sym` marks
a symmetric element. `vsize` and `rvsize` are just the products of the
respective shapes, so they are not tabulated.

| element        | gdim | vshape | bs | rvshape | bvsize |
| -------------- | ---: | -----: | -: | ------: | -----: |
| P1             |    2 |     {} |  1 |      {} |      1 |
| P1             |    3 |     {} |  1 |      {} |      1 |
| P1, shape {2}  |    2 |    {2} |  2 |      {} |      1 |
| P1, shape {2}  |    3 |    {2} |  2 |      {} |      1 |
| P1, shape {5}  |    2 |    {5} |  5 |      {} |      1 |
| P1, sym {3, 3} |    2 | {3, 3} |  6 |      {} |      1 |
| RT 1           |    2 |    {2} |  1 |     {2} |      2 |
| RT 1           |    3 |    {3} |  1 |     {2} |      3 |
| Regge 0        |    2 | {2, 2} |  1 |  {2, 2} |      4 |
| Regge 0        |    3 | {3, 3} |  1 |  {2, 2} |      9 |
| mixed          |  any | throws |  1 |  throws | throws |

Reading the table:

- **`gdim` changes nothing for the identity-mapped rows.** `P1` and
  `P1, shape {2}` are identical at `gdim == 2` and `gdim == 3`. Contrast the
  `RT 1` and `Regge 0` pairs, where the Piola push-forward does introduce
  `gdim`. This is the whole of the bug: only Piola-mapped elements on a manifold
  were wrong.
- **A blocked element's value shape is chosen by the caller**, and has nothing
  to do with `gdim`. `P1, shape {5}` on a 2D mesh keeps shape `{5}`; `gdim` must
  never be substituted into it. Blocked elements are built from *scalar* base
  elements, so their base map is identity or L2 Piola and the question does not
  arise.
- **`P1, sym {3, 3}` is the one case where `block_size != value_size`.** The
  value size is 9, but only the six independent components are stored, one dof
  each; the other three are recovered by symmetry. Hence `base_value_size()` rather than
  `value_size() / block_size()`, which gives 9/6 = 1 by integer truncation and
  4/3 = 1 for the 2D case — right by accident, wrong in principle.
- **A mixed element has no value shape of its own.** Both shapes are
  `std::nullopt` and every shape/size accessor throws; `is_mixed()` is
  implemented as "has no reference value shape". Only `block_size()` is defined,
  and is 1.

## The rule

`fem::compute_value_shape(map_type, reference_value_shape, gdim)` replaces the
trailing axes of the reference shape with `gdim`. How many is fixed by the map
type alone:

| `basix::maps::type`        | axes set to `gdim` |
| -------------------------- | -----------------: |
| `identity`                 |                  0 |
| `L2Piola`                  |                  0 |
| `covariantPiola`           |                  1 |
| `contravariantPiola`       |                  1 |
| `doubleCovariantPiola`     |                  2 |
| `doubleContravariantPiola` |                  2 |

This mirrors `AbstractPullback.physical_value_shape` in UFL, which is what
`ufl.FunctionSpace.value_shape` is built from. The two must agree, and a test
asserts they do across a matrix of elements and meshes.

## Changes

- **`fem::FiniteElement`**: the Basix and `BasixElementData` constructors take a
  required `gdim`. `_value_shape` is now the physical shape, computed with
  `compute_value_shape` when no block shape is given and taken verbatim when one
  is. `_bs` and `_reference_value_shape` are unchanged. New `base_value_size()`.
  `operator==` also compares value shapes, since the same Basix element on
  meshes of different `gdim` is not the same element and callers use equality to
  decide dofs can be copied directly.
- **Buffer sizes.** Several buffers holding *physical* values were sized with
  `reference_value_size()`, so they truncated on a manifold even once the checks
  passed: `_vals` in `piola_mapped_evaluation`, `basis0` and `mapped_values0` in
  `interpolate_nonmatching_maps`, `basis_values` in `interpolation_matrix`, and
  the push-forward output in `Function::eval`. `mapped_values0` holds *reference*
  values of the target element and is now sized `reference_value_size * bs`;
  previously it used the target's `value_size()`, which coincides only when
  `gdim == tdim` or the target is identity mapped.
- **`interpolation_matrix`** gains the value-shape compatibility check that
  `interpolate` already had. Without it, it silently built a wrong operator.
- **Python**: `finiteelement()` takes `gdim`; `reference_value_shape`,
  `reference_value_size`, `value_size` and `base_value_size` are now bound.
- Every C++ construction site (demos, tests, `interpolate_geometry`) passes
  `mesh->geometry().dim()`.

Not affected: IO. `VTKFile`, `XDMFFile` and the ADIOS2 writers reject any
element that is not identity mapped, and for an identity map the two shapes
coincide.

## Tests

New `python/test/unit/fem/test_manifold.py` (104 tests) and
`cpp/test/fem/manifold.cpp`. Every Python test is parametrised over
`gdim in {2, 3}` on the same triangulation — the `gdim == 2` case is a control
that must pass before and after. On `main`, 42 of the 104 fail and all 42 are
`gdim == 3`.

Covered: value shapes against UFL for RT/BDM/N1curl/N2curl/Regge/HHJ; blocked
shapes unrelated to `gdim`; interpolation from a callable, between spaces with
the same and with different map types, from an `Expression`, and between
non-matching meshes; `Function.eval`; `interpolation_matrix` and
`discrete_gradient`.

Two of the bugs failed *silently* rather than raising, so they have tests that
do not depend on the throwing paths: `Function::eval` returned
`[a, b, 0.0]` for a manifold RT field (third component never written), and
`interpolation_matrix` built an operator with an L2 error of 0.115 instead of
~1e-15.

## API break

The `fem::FiniteElement` constructors take a new required `gdim` argument, and
`python/dolfinx/fem/element.py::finiteelement` takes a new `gdim` parameter.

Almost nothing changes for the Python user, while the C++ user have to specify the gdim in the finite element constructor.

## AI disclosure
This PR has been generated with the assistance of CLAUDE (OPUS 5).


jhale:
While we're here, do we want to consider prepending these methods with physical or reference? Would make it easier to reason/remember which one to use.

jhale:
The LLMs are all a bit eager to write C++ tests. The default position is that tests should be written in Python, unless they are testing very subtle issues which require the tight control of C++.

I will add something to agents.md to push things in the right direction.

jorgensd:
> The LLMs are all a bit eager to write C++ tests. The default position is that tests should be written in Python, unless they are testing very subtle issues which require the tight control of C++.
> 
> I will add something to agents.md to push things in the right direction.

I'm ok with removing the C++ test.

jorgensd:
> While we're here, do we want to consider prepending these methods with physical or reference? Would make it easier to reason/remember which one to use.

I can look into that.

jorgensd:
> > While we're here, do we want to consider prepending these methods with physical or reference? Would make it easier to reason/remember which one to use.
> 
> I can look into that.

We should then rename it throughout the whole FEniCS framework (UFL, basix etc.).
Let's do that in another PR.

jhale:
Could you explain what happens in nonsensical or unsupported cases in terms of raising errors?

jorgensd:
> Could you explain what happens in nonsensical or unsupported cases in terms of raising errors?

```python
from mpi4py import MPI
import dolfinx
import basix.ufl
import numpy as np


x = np.array([[0.0, 0.0, 0.0], [1.0, 0.0, 0.0], [1.0, 1.0, 0.2]])
cells = np.array([[0, 1, 2]], dtype=np.int64)
c_el = basix.ufl.element("P", "triangle", 1, shape=(x.shape[1],))
mesh = dolfinx.mesh.create_mesh(MPI.COMM_WORLD, cells=cells, x=x, e=c_el)

el = basix.ufl.element("N1curl", mesh.basix_cell(), 1)
m_el = basix.ufl.mixed_element([el, el])

fs_m = dolfinx.fem.functionspace(mesh, m_el)
F = dolfinx.fem.Function(fs_m)
F.sub(0).interpolate(lambda x: np.array([x[0], x[1], x[2]]))

el = fs_m.element.physical_base_value_size
```
raises
```bash
  File "/root/shared/dolfinx/python/test/mwe_failing.py", line 19, in <module>
    el = fs_m.element.physical_base_value_size
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/shared/dolfinx/python/dolfinx/fem/element.py", line 354, in physical_base_value_size
    return self._cpp_object.physical_base_value_size
```
while a mixed_element doesn't get past basix construction
```python
b_el = basix.ufl.blocked_element(el, shape=(2,))
```
returns
```bash
  File "/root/shared/dolfinx/python/test/mwe_failing.py", line 22, in <module>
    b_el = basix.ufl.blocked_element(el, shape=(2,))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/shared/basix/python/basix/ufl.py", line 1940, in blocked_element
    raise ValueError("Cannot create a blocked element containing a non-scalar element.")
ValueError: Cannot create a blocked element containing a non-scalar element.
```

jhale:
Looks good to me.

jhale (approved):

PR #4515, opened 2026-09-15
mesh: validate rank-uniformly before MPI collectives in mesh creation

jhale:
Several O(1), communication-free validation checks sat behind an
MPI collective they should precede, or ran on only a subset of ranks,
so bad input could throw on some ranks while others entered the
collective and hung: create_rectangle/create_box degeneracy checks
(previously rank-0-only / subcomm-only), build_local_dual_graph's
celltypes/cells size check (after its empty-mesh early return),
create_mesh's cell-array-size check (between mpi_check and
graph::build::distribute), and a Developer-build sanity check in
create_topology (now MPI_LOR-reduced before throwing, matching
create_subtopology's existing pattern). Also fixes build_prism using
comm instead of subcomm for its cell range, which silently dropped
cells under a strict-subset subcomm, and reorders a locate_entities_boundary
check that sat between two collectives.

---


python/test/unit/mesh/test_mesh.py
```diff
@@ -357,6 +322,41 @@ def test_create_box_prism():
     assert mesh.topology.index_map(3).size_global == 48
 
 
+def test_create_rectangle_degenerate_raises_on_every_rank():
```
jhale, line 325: Can remove this test.

python/test/unit/mesh/test_mesh.py
```diff
@@ -357,6 +322,41 @@ def test_create_box_prism():
...
+        create_rectangle(
+            MPI.COMM_WORLD,
+            [[0.0, 0.0], [1e-8, 1.0]],
+            [4, 4],
+            CellType.triangle,
+            dtype=np.float32,
+        )
+
+
+def test_create_box_degenerate_raises_on_every_rank():
```
jhale, line 343: Can remove this test.

jhale:
Re-classification of exceptions is done in #4519.

chrisrichardson (approved):

PR #4516, opened 2026-09-15
mesh: convert O(1) argument-precondition asserts to unconditional throws

jhale:
Several O(1), communication-free checks on caller-supplied arguments
were guarded only by assert(), which is compiled out in Release builds.
Converts them to unconditional throws per AGENTS.md, including the
critical compute_connectivity precondition that lets compute_from_map
silently mismatch reference-edge tables (or read out of bounds) when
called with an unsupported dimension pair. Also adds two missing
checks (graphbuild's cell-array-size and cell-type-dimension
consistency) that had no guard at all.

---


cpp/dolfinx/mesh/permutationcomputation.cpp
```diff
@@ -251,9 +251,12 @@ compute_edge_reflections(const mesh::Topology& topology, int num_threads)
   auto c_to_v = topology.connectivity(tdim, 0);
   assert(c_to_v);
   auto c_to_e = topology.connectivity(tdim, 1);
-  assert(c_to_e);
+  if (!c_to_e)
+    throw std::runtime_error("Edges have not been computed.");
   auto e_to_v = topology.connectivity(1, 0);
-  assert(e_to_v);
+  if (!e_to_v)
```
chrisrichardson, line 257: style requires parentheses around multi-line statements {}.

cpp/dolfinx/mesh/utils.h
garth-wells: This is bad inside the loop. It needs to be hoisted.

garth-wells, line 843: Do we know the final size of `entity_xdofs`? Would be good to use `reserve` if we can.

garth-wells (dismissed):

cpp/dolfinx/mesh/utils.h
garth-wells: Fixed in b8da948651: hoisted `closure_dofs` out of the loop and reuse it via `assign()` instead of constructing a new vector every iteration.

garth-wells: Already handled — `entity_xdofs.reserve(entities.size() * num_entity_dofs)` at line 763 is above this branch and applies here too, so no change needed for this one.

garth-wells: Follow-up in 8c9603c271: the closure-DOF count for a given `dim` is actually invariant across local entity indices (the only exception, prism/pyramid facets, is already rejected earlier in the function), so `closure_dofs` can be sized once outside the loop and copied into directly, rather than `assign()`-ed from a variable-length range each iteration.

garth-wells (approved):

PR #4517, opened 2026-09-15
mesh: validate dimensions, null pointers and indices at the public boundary

jhale:
Several public entry points took dimension/index arguments, or
shared_ptr<const Topology> arguments, with no validation at all, so bad
input from Python was undefined behaviour in Release rather than an
exception: Topology::index_map(s) indexed _entity_types with unchecked
operator[] (the root cause several other constructors route through),
cell_types.cpp's entity/sub-entity lookups indexed basix tables directly,
EntityMap's and MeshTags' constructors dereferenced their Topology
arguments before checking them, create_geometry never validated its
gdim argument or cross-checked x/nodes/xdofs sizes, compute_incident_entities
passed caller entity indices straight into an unchecked AdjacencyList
lookup, reorder_cells never validated a user reorder_fn's returned
permutation, and transfer_meshtags_to_submesh never checked that its
EntityMap arguments actually related the topologies it was given -
swapping them (the natural mistake, since that's create_submesh's own
return order) was an unchecked OOB read and write.

---


python/test/unit/mesh/test_mesh.py
```diff
@@ -902,6 +902,44 @@ def test_mesh_single_process_distribution(partitioner):
...
+    tdim = msh.topology.dim
+    msh.topology.create_connectivity(0, tdim)
+    num_vertices = msh.topology.index_map(0).size_local + msh.topology.index_map(0).num_ghosts
+    with pytest.raises(IndexError):
+        dolfinx.mesh.compute_incident_entities(
+            msh.topology, np.array([num_vertices], dtype=np.int32), 0, tdim
+        )
+
+
+def test_transfer_meshtags_to_submesh_swapped_maps_raises():
```
jhale, line 922: Note that the next PR in the stack https://github.com/FEniCS/dolfinx/pull/4518 removes this discrepancy in argument ordering and therefore these tests.

python/test/unit/mesh/test_mesh.py
```diff
@@ -902,6 +902,44 @@ def test_mesh_single_process_distribution(partitioner):
             assert adj.links(i).size == 2
 
 
+def test_compute_incident_entities_out_of_range_index():
```
jhale, line 905: Can be removed, seems trivial?

cpp/dolfinx/mesh/EntityMap.h
```diff
@@ -43,6 +43,14 @@ class EntityMap
         _sub_topology_to_topology(std::forward<U>(sub_topology_to_topology)),
         _sub_topology(sub_topology)
   {
+    if (!topology)
+      throw std::invalid_argument("topology must not be null.");
+    if (!sub_topology)
+      throw std::invalid_argument("sub_topology must not be null.");
+    if (dim < 0 or dim > topology->dim() or dim > sub_topology->dim())
+      throw std::invalid_argument(
```
chrisrichardson, line 52: should be in parentheses {}

cpp/dolfinx/mesh/utils.cpp
```diff
@@ -93,6 +93,27 @@ std::vector<std::int64_t> mesh::impl::reorder_cells(
...
+    // Validate that the reordering function returned a permutation of
+    // [0, num_owned_cells), since remap is used as one below (both as a
+    // direct index and via reorder_list).
+    if (remap.size() != num_owned_cells)
+    {
+      throw std::invalid_argument(
+          "Cell reordering function returned the wrong number of cells.");
+    }
+    {
+      std::vector<bool> seen(num_owned_cells, false);
```
chrisrichardson, line 105: Is this sufficiently cheap to remain in `Release` mode?

chrisrichardson: OK, I see it has NDEBUG guard later in stack.

cpp/dolfinx/mesh/utils.h
```diff
@@ -1793,6 +1796,35 @@ MeshTags<T> transfer_meshtags_to_submesh(
...
+  // this function assumes. Passing the two maps in create_submesh's own
+  // return order (entity_map, vertex_map) rather than this function's
+  // (vertex_map, cell_map) is the natural mistake to make here.
+  if (cell_map.dim() != static_cast<std::size_t>(submesh_tdim))
+  {
+    throw std::invalid_argument(
+        "cell_map dimension must equal the submesh topology dimension.");
+  }
+  if (cell_map.topology() != topology)
+    throw std::invalid_argument(
```
garth-wells, line 1819: Add curly braces

cpp/dolfinx/mesh/utils.h
```diff
@@ -1793,6 +1796,35 @@ MeshTags<T> transfer_meshtags_to_submesh(
...
+        "cell_map topology must match tags.topology().");
+  if (cell_map.sub_topology() != submesh_topology)
+    throw std::invalid_argument(
```
garth-wells, line 1824: Curly braces

garth-wells (dismissed):

garth-wells (approved):

PR #4518, opened 2026-09-15
mesh: reorder transfer_meshtags_to_submesh args to match create_submesh

jhale:
create_submesh returns (submesh, entity_map, vertex_map, geom_map), but
transfer_meshtags_to_submesh took (tags, submesh, vertex_map, cell_map)
- vertex/cell swapped relative to how callers just received them, which
was the natural mistake the F12-1 validation now catches. Reorders to
(tags, submesh, cell_map, vertex_map) so create_submesh's result can be
unpacked and passed straight through. Breaking change, no deprecation
path.

---


jorgensd:
I'm a bit worried about this one, as it will be very hard for downstream users to catch (even if it is in the release notes).

chrisrichardson (approved):

jhale:
> I'm a bit worried about this one, as it will be very hard for downstream users to catch (even if it is in the release notes).

There is a 'tiny struct trick' in C++ which we can use for one or two rounds of releases to force users to make the switch through the typing system.

garth-wells (approved):

jhale:
@jorgensd I have another PR which types this, will submit separately.

PR #4519, opened 2026-09-15
mesh: reclassify exception types per AGENTS.md convention

jhale:
Mirrors the earlier common/ pass (f97670b64f). 46 throw sites describing
a violated argument precondition were std::runtime_error and are now
std::invalid_argument; 2 lookup-key failures in Topology::index_map are
now std::out_of_range. Purely mechanical - no behaviour change beyond
the Python-visible RuntimeError -> ValueError, and nothing in-tree
catches these by type. Updates the two WP1 regression tests whose
assertions depended on the old type.

---


chrisrichardson (approved):

cpp/dolfinx/mesh/utils.h
```diff
@@ -1341,7 +1341,8 @@ Mesh<typename std::remove_reference_t<typename U::value_type>> create_mesh(
   using T = typename std::remove_reference_t<typename U::value_type>;
 
   if (cells.size() != elements.size())
-    throw std::runtime_error("Number of cell arrays and elements must match.");
+    throw std::invalid_argument(
```
garth-wells, line 1352: Add braces.

garth-wells (dismissed):

Issue #4522, opened 2026-09-15
cpp/test/mesh/generation.cpp: build_tet cell-vertex order doesn't match hand-written expected tables


jhale:
## Summary

`cpp/test/mesh/generation.cpp` has a `CHECK_adjacency_list_equal` test helper whose per-row structural comparison is commented out:

```cpp
template <typename T>
void CHECK_adjacency_list_equal(
    const dolfinx::graph::AdjacencyList<T>& adj_list,
    const std::vector<std::vector<T>>& expected_list)
{
  REQUIRE(static_cast<std::size_t>(adj_list.num_nodes())
          == expected_list.size());

  // for (T i = 0; i < adj_list.num_nodes(); i++)
  // {
  //   CHECK_THAT(adj_list.links(i),
  //              Catch::Matchers::RangeEquals(expected_list[i]));
  // }
}
```

So every one of its ~9 call sites across the file (interval, triangle `left`/`right`/`crossed` diagonals, hexahedron, tetrahedron) currently only checks `num_nodes()` — the number of rows — never the actual vertex lists. The hand-written expected tables (e.g. the tet Kuhn-decomposition cell-to-vertex list in the `"Box tetrahedron mesh"` case) are dead text and give no structural coverage.

## Reproduction

Re-enable the loop (with `std::int32_t i` instead of `T i`, which doesn't compile as written) and rebuild:

```cpp
  for (std::int32_t i = 0; i < adj_list.num_nodes(); i++)
  {
    CHECK_THAT(adj_list.links(i),
               Catch::Matchers::RangeEquals(expected_list[i]));
  }
```

The `"Box tetrahedron mesh"` case then fails on the cell-to-vertex (`connectivity(3, 0)`) check, e.g.:

```
CHECK_THAT( adj_list.links(i), Catch::Matchers::RangeEquals(expected_list[i]) )
with expansion:
  { 0, 4, 1, 3 } elements are { 0, 1, 3, 4 }

CHECK_THAT( adj_list.links(i), Catch::Matchers::RangeEquals(expected_list[i]) )
with expansion:
  { 0, 5, 3, 2 } elements are { 0, 2, 5, 3 }

CHECK_THAT( adj_list.links(i), Catch::Matchers::RangeEquals(expected_list[i]) )
with expansion:
  { 0, 6, 4, 3 } elements are { 0, 4, 3, 6 }

[...]

PR #4524, opened 2026-09-15
mesh: fix transfer_meshtags_to_submesh sentinel collision, close test gaps

jhale:
Replaces the numeric_limits<T>::max() "unmapped" sentinel in
transfer_meshtags_to_submesh with an explicit found-flag vector: a
legitimate tag value equal to that sentinel was being silently dropped
from the transferred result.

Adds regression tests for three previously-untested paths: a
cells/celltypes length mismatch across ranks in build_dual_graph (the
same MPI-ordering defect class as WP1, now verified not to hang),
create_submesh with zero selected entities on some ranks (previously
only traced by hand, never executed), and the sentinel-collision fix
above.

Left cpp/test/mesh/generation.cpp's commented-out per-row adjacency
checks disabled: re-enabling them exposes a real discrepancy between
build_tet's cell-vertex order and the hand-written expected tables,
but Topology construction reorders cell vertices during distribution
and no ordering contract is documented for connectivity(tdim, 0), so
it's unclear whether this is a build_tet bug or an outdated fixture.
Left as a follow-up rather than guessing.

---


cpp/test/mesh/generation.cpp
```diff
@@ -32,7 +32,15 @@ void CHECK_adjacency_list_equal(
...
-  // for (T i = 0; i < adj_list.num_nodes(); i++)
+  // NOTE: re-enabling this per-row check (see git history) fails on the
+  // tetrahedron cell-to-vertex case: the actual row order differs from
+  // the hand-written expected tables by a permutation, even though
+  // Topology construction re-numbers/re-orders cell vertices during
+  // distribution and no ordering contract is documented for
+  // connectivity(tdim, 0). Left disabled pending a dedicated
+  // investigation into whether that's a real build_tet bug or the
+  // tables simply predate a later, legitimate reordering.
+  // for (std::int32_t i = 0; i < adj_list.num_nodes(); i++)
```
jorgensd: This belongs to another PR no?

python/test/unit/mesh/test_dual_graph.py
```diff
@@ -31,3 +32,25 @@ def test_dgrsph_1d():
...
+    # Every rank passes 2 cell types but only 1 cell array -- the mismatch
+    # itself, not the cell content, is what must be rejected uniformly.
+    if comm.rank == 0:
+        cells = [np.array([0, 1, 2], dtype=np.int64)]
+    else:
+        cells = [np.array([], dtype=np.int64)]
+    with pytest.raises(ValueError):
+        mesh.build_dual_graph(
+            comm, [mesh.CellType.triangle, mesh.CellType.quadrilateral], cells, None, 1
+        )
```
jorgensd: Also seems like it belongs in another PR

python/test/unit/mesh/test_mesh.py
```diff
@@ -902,6 +902,24 @@ def test_mesh_single_process_distribution(partitioner):
...
+    mesh = create_unit_square(MPI.COMM_WORLD, 8, 8)
+    tdim = mesh.topology.dim
+    if MPI.COMM_WORLD.rank == 0:
+        num_local = mesh.topology.index_map(tdim).size_local
+        entities = np.arange(num_local, dtype=np.int32)
+    else:
+        entities = np.empty(0, dtype=np.int32)
+    submesh, _entity_map, _vertex_map, _node_map = create_submesh(mesh, tdim, entities)
+    if MPI.COMM_WORLD.rank != 0:
+        assert submesh.topology.index_map(tdim).size_local == 0
```
jorgensd, line 910: Fine with keepng this test as it checks something concrete on submesh.

python/test/unit/mesh/test_mesh.py
```diff
@@ -919,6 +937,28 @@ def test_compute_incident_entities_out_of_range_index():
...
+    numeric_limits<T>::max() internally as an "unmapped" sentinel, which
+    collided with a legitimate tag value of exactly that value and
+    silently dropped it from the result.
+    """
+    mesh = create_unit_square(MPI.COMM_WORLD, 4, 4)
+    tdim = mesh.topology.dim
+    submesh, entity_map, vertex_map, _node_map = create_submesh(
+        mesh, tdim, np.arange(2, dtype=np.int32)
+    )
+    max_val = np.iinfo(np.int32).max
```
jorgensd: Fair enough test.

cpp/test/mesh/generation.cpp
jhale: Fair point — dropped. This was just a leftover investigation comment (no behavior change); the actual investigation is tracked in https://github.com/FEniCS/dolfinx/issues/4522, so it doesn't need to live here too.

python/test/unit/mesh/test_dual_graph.py
jhale: Agreed — moved to #4515 (WP1), since it's regression-testing WP1's own build_dual_graph check-ordering fix, not anything in this PR.

chrisrichardson (approved):

garth-wells (approved):

PR #4525, opened 2026-09-15
mesh: compute 1D inter-process facets from the vertex IndexMap

jhale:
Topology's tdim==1 constructor hack tried to derive inter-process
facets (vertices, for a 1D mesh) via mesh::compute_entities(dim=0),
but that function has always returned immediately for dim==0 since
vertices are set up directly rather than computed on demand. The
pushed (always-empty) vector satisfied interprocess_facets()'s "not
computed" guard, so it silently returned an empty list for every 1D
mesh instead of the true inter-process vertices, misclassifying
partition-boundary vertices as exterior facets under GhostMode::none.

Replace the hack with compute_interprocess_vertices, which reads the
answer directly from the vertex IndexMap's own ownership/ghosting
data instead of recomputing entity sharing: every vertex needed by a
locally owned cell is already resolved there (owned or ghost),
independently of GhostMode, so no extra communication round is
needed. Drops the now-unused num_threads parameter from Topology's
constructor.

---


cpp/dolfinx/mesh/Topology.h
```diff
@@ -65,15 +65,13 @@ class Topology
...
   /// `cells`.
-  /// @param[in] num_threads Number of threads to use. Must be >= 1.
   Topology(
       std::vector<CellType> cell_types,
       std::shared_ptr<const common::IndexMap> vertex_map,
       std::vector<std::shared_ptr<const common::IndexMap>> cell_maps,
       std::vector<std::shared_ptr<graph::AdjacencyList<std::int32_t>>> cells,
       const std::optional<std::vector<std::vector<std::int64_t>>>&
-          original_cell_index = std::nullopt,
-      int num_threads = 1);
```
jhale, line 76: This isn't used as far as I can see, but will check again.

jorgensd:
@jhale can you check that this is correct for t joints or even larger joints? Ie sometimes where a facet is shared between 3 to N cells?

cpp/dolfinx/mesh/Topology.h
jorgensd: The original cell index or the num_threads?

jorgensd:
The `compute_interprocess_vertices` discussion made me look at `index_to_dest_ranks`, where the observation that the IndexMap already carries the communication pattern is not currently exploited.

Therefore I asked CLAUDE (OPUS 5) to see if we could exploit this rewrite in `index_to_dest_ranks`. The following is a summary from my discussion with CLAUDE:
## Follow-up idea (not blocking): `IndexMap::index_to_dest_ranks` could be rewritten along the same lines**


I instrumented a copy of the current implementation (verified to return identical `data`/`offsets`) and timed it on the vertex map of a 32³ tetrahedral cube, `GhostMode::shared_facet`, 8 ranks, max over ranks, 20 reps:

```
index_to_dest_ranks (full)                   : 0.211 ms
  consensus NBX (dest discovery)             : 0.076 ms
  part 1: ghosts -> owner exchange           : 0.099 ms
  part 2: build send buffer                  : 0.013 ms
  part 2: comm (create comm + alltoallv)     : 0.053 ms
  part 2: unpack (2 sorts + binary searches) : 0.068 ms

Scatterer construction (same exchange as part 1)   : 0.055 ms
owner-side rows rebuilt from Scatterer data (local) : 0.003 ms
```

Three changes, in order of payoff:

1. **Drop the consensus round.** The function calls `compute_graph_edges_nbx` to discover `dest`, but the constructor already computed exactly that into `_dest` via `build_src_dest`. I checked on 4 and 8 ranks that the rediscovered list equals `dest()` on every rank. A caller-supplied `dest` is only verified under `#ifndef NDEBUG`, but a superset merely contributes empty blocks to the neighbourhood exchange, and a subset is a map that is already broken for `Scatterer`, `shared_indices` and every scatter operation, since they all trust `_dest`. So this is pure subtraction, and it removes a global `MPI_Ibarrier`-based synchronisation point that will cost more than 0.076 ms at high rank counts. It also makes the `tag` parameter dead, so the parameter and its `@note` in `IndexMap.h` can go (public API change).

2. **Part 1 is the `Scatterer` constructor.** Both send ghost global indices to their owners over the same `src`/`dest` neighbourhood, and `Scatterer::local_indices_block()` together with `_sizes_local` is exactly the `(owned local index, ghosting rank)` list that part 1 rebuilds with an `int64` pair sort. Given that data the owner-side rows are a counting sort: 0.003 ms against 0.099 ms, and one `MPI_Dist_graph_create_adjacent` disappears. This needs `Scatterer` to expose its per-neighbour sizes/displacements, or the regrouping to move inside `Scatterer`/`IndexMap`.

3. **Part 2: send in forward-scatter order.** Today each owned index `i` with `n_i` ghosting ranks sends `(global index, rank)` `int64` pairs to each of them, `2·Σn_i²` `int64` in total, and the receiver sorts its ghosts, binary-searches every received index, then sorts again. In scatter order the buffer position already identifies the ghost through `remote_indices_block()`, so the global index need not travel at all: `int32` counts plus `int32` rank lists, and the unpack becomes one linear pass. Payload on the mesh above: 146 KiB → 67 KiB, approaching 4× as `n_i` grows (corner vertices, high rank counts).

Extrapolating from the section timings, and without having prototyped it, that is roughly 0.21 → 0.07–0.09 ms, so 2–3×.

Independently of the above, `get_local_indexing` calls `index_to_dest_ranks` once per entity dimension on the same vertex map, so a 3D mesh that creates both edges and facets computes the identical result twice (more with mixed topology). Hoisting the call, or caching the result on the map, halves the cost before anything else changes. The other callers are `refinement/interval.h` and `refinement/plaza.h`.

Caveats on the numbers: shared-memory MPI on a 10-core machine at ≤ 8 ranks, so the balance between local work and communication will look different at scale; the per-section maxima land on different ranks and therefore do not sum to the total.

python/test/unit/mesh/test_branching_manifold.py
```diff
@@ -159,3 +186,124 @@ def on_boundary(x):
...
+    cell on the ranks owning its dual-graph neighbours.
+    """
+
+    def partitioner(comm, nparts, dual_graph, *args):
+        offset = comm.exscan(dual_graph.num_nodes) or 0
+        dests, offsets = [], [0]
+        for i in range(dual_graph.num_nodes):
+            owner = (offset + i) % nparts
+            d = [owner]
+            if ghost:
```
jorgensd, line 202: This is a very cool way of doing ghosting, which I believe could be used to extend the number of layers one would search outwards in the dual graph! @IgorBaratta

python/test/unit/mesh/test_branching_manifold.py
```diff
@@ -159,3 +186,124 @@ def on_boundary(x):
...
+
+def _interprocess_vertices_reference(topology):
+    """Global indices of the vertices attached to cells owned by two or
+    more ranks, as seen by this rank.
+    """
+    comm = topology.comm
+    c_to_v = topology.connectivity(topology.dim, 0)
+    v_map = topology.index_map(0)
+    vertices = {v for c in range(topology.index_map(1).size_local) for v in c_to_v.links(c)}
+    attached = set(v_map.local_to_global(np.fromiter(vertices, dtype=np.int32)))
```
jorgensd:
```suggestion
    assert topology.dim == 1
    num_local_edges = topology.index_map(topology.dim).size_local
    end_idx = c_to_v.offsets[num_local_edges]
    vertices = np.unique(c_to_v.array[:end_idx]).astype(np.int32)
    attached = v_map.local_to_global(vertices)
```
I think this is more readable and faster.

python/test/unit/mesh/test_branching_manifold.py
```diff
@@ -159,3 +186,124 @@ def on_boundary(x):
...
+    shared = comm.allgather(attached)
+    return {v for v in attached if sum(v in s for s in shared) > 1}
```
jorgensd:
```suggestion
    shared = comm.allgather(attached)
    all_vertices = np.concatenate(shared)
    uniques, counts = np.unique(all_vertices, return_counts=True)
    shared_globally = uniques[counts > 1]
    local_shared = np.intersect1d(attached, shared_globally)
    return set(local_shared)
```
More readable and faster (hopefully)

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -821,6 +822,92 @@ build_entity_types(const std::vector<CellType>& cell_types)
...
+  std::vector<std::int8_t> attached(num_local + num_ghosts, 0);
+  std::vector<std::shared_ptr<const common::IndexMap>> cell_maps
+      = topology.index_maps(tdim);
+  for (std::size_t i = 0; i < cell_maps.size(); ++i)
+  {
+    auto c_to_v = topology.connectivity({tdim, static_cast<int>(i)}, {0, 0});
+    assert(c_to_v);
+    for (std::int32_t c = 0; c < cell_maps[i]->size_local(); ++c)
+      for (std::int32_t v : c_to_v->links(c))
+        attached[v] = 1;
```
jorgensd:
```suggestion
    std::int32_t num_local_cells = cell_maps[i]->size_local();
    std::int32_t end_idx = c_to_v->offsets()[num_local_cells];
    std::span<const std::int32_t> all_links = c_to_v->array();
    std::span<const std::int32_t> local_links = all_links.first(end_idx);
    std::ranges::for_each(local_links, [&attached](std::int32_t v) {
        attached[v] = 1;
});
```
As we don't care about cell index here this should be better.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -821,6 +822,92 @@ build_entity_types(const std::vector<CellType>& cell_types)
...
+                            request);
+  scatter.scatter_rev_end(request);
+
+  std::vector<std::int8_t> interprocess(num_local + num_ghosts, 0);
+  {
+    std::vector<std::int32_t> count(num_local, 0);
+    for (std::int32_t v = 0; v < num_local; ++v)
+      count[v] = attached[v];
+    for (std::size_t i = 0; i < local_ind.size(); ++i)
+      count[local_ind[i]] += buffer_local[i];
```
jorgensd, line 889:
```suggestion
    auto indices = std::views::iota(0u, local_ind.size());
    std::ranges::for_each(indices, [&](std::size_t i) {
        count[local_ind[i]] += buffer_local[i];
});
```

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -821,6 +822,92 @@ build_entity_types(const std::vector<CellType>& cell_types)
...
+  MPI_Request request;
+  scatter.scatter_rev_begin(buffer_ghost.data(), buffer_local.data(), 1,
+                            request);
+  scatter.scatter_rev_end(request);
+
+  std::vector<std::int8_t> interprocess(num_local + num_ghosts, 0);
+  {
+    std::vector<std::int32_t> count(num_local, 0);
+    for (std::int32_t v = 0; v < num_local; ++v)
+      count[v] = attached[v];
```
jorgensd:
```suggestion
    std::vector<std::int32_t> count(attached.begin(),
                                std::next(attached.begin(), num_local));
```

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -821,6 +822,92 @@ build_entity_types(const std::vector<CellType>& cell_types)
...
+    for (std::size_t i = 0; i < local_ind.size(); ++i)
+      count[local_ind[i]] += buffer_local[i];
+    for (std::int32_t v = 0; v < num_local; ++v)
+      interprocess[v] = count[v] > 1;
```
jorgensd:
```suggestion
    std::ranges::transform(count, interprocess.begin(),
                           [](std::int32_t c) -> std::int8_t { return c > 1; });
```

cpp/dolfinx/mesh/Topology.h
jhale: num_threads

cpp/dolfinx/mesh/Topology.cpp
jhale: Will add a note to use std::ranges::zip when we move to C++23.

cpp/dolfinx/mesh/Topology.cpp
jhale: Nothing wrong with the loop. Performant `std::views` implementations are 'tricky', we should do some more investigations before using them in DOLFINx.

jorgensd: I was able to remove the views, as count is num_local big, so now it is a transform instead.

cpp/dolfinx/mesh/Topology.h
jorgensd: Agreed.

schnellerhase:
Which use cases does this fix? For the here tested branching manifolds, there is no general concept of interior or exterior facet.

jhale:
> Which use cases does this fix? For the here tested branching manifolds, there is no general concept of interior or exterior facet.

The PR fixes the computation interior and exterior facets with `GhostMode.None` for one-dimensional *non-branching* meshes (see e.g. the super simple included test with just a `UnitInterval`, nothing fancy). Don't get overly distracted by non-manifold meshes, this is just test harnessing to check the definition (more below!) is actually applied in more complex cases.

Before the fix, `interprocess_facets()` always returned empty, because the code path it relied on `compute_entities(dim=0)` which returns empty. With `GhostMode::none`, a vertex at a partition boundary has only one cell attached on its own rank, so it "looks" exterior to itself unless `interprocess_facets()` says something otherwise. This all mysteriously worked with `GhostMode::SharedFacet` because of the automatic duplication of ghost geometry data on interior facets at partition boundaries.

The definition of interior and exterior is:

Let $f$ be a facet (a $(d-1)$-dimensional entity of a mesh of topological dimension $d$), and let $c(f)$ denote the number of cells owned globally that have $f$ as a boundary entity.

Exterior facet: $c(f) = 1$.
Interior facet: $c(f) \geq 2$.

This definition already exists, from my understanding of the existing code.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -821,6 +822,93 @@ build_entity_types(const std::vector<CellType>& cell_types)
...
+/// @param[in] topology Topology with tdim == 1, whose vertex index map
+/// and cell-vertex connectivity have already been set.
+/// @return Sorted local indices of the inter-process vertices.
+///
+/// @note Collective.
+std::vector<std::int32_t>
+compute_interprocess_vertices(const Topology& topology)
+{
+  assert(topology.dim() == 1);
+  constexpr int tdim = 1;
```
schnellerhase:
```suggestion
    constexpr int tdim = 1;
    assert(topology.dim() == tdim);
```

schnellerhase:
Thanks, should this not be managed in the `create_entities` special case handling, rather than in the constructor?
1. Computing interprocess facets data in the constructor forces computational cost which is not always required,
2. We expect any code path to rely on facet information to call a `mesh.topology.create_entities(tdim-1)` at some point anyhow, so this should be consistent for `tdim=1` (for `tdim=1` we would have the special case that vertices are always available but interprocess facet data might not be available yet)

One other point, that I need to think more on: if the shared facet case worked before without this additional work path than there must be a no-op short cut to this computation at least in this case, which we should retain.

jhale:
> Thanks, should this not be managed in the `create_entities` special case handling, rather than in the constructor?

Yes.

> 1. Computing interprocess facets data in the constructor forces computational cost which is not always required,

Indeed, an easy fix, I will implement it through lazy computation.

> 2. We expect any code path to rely on facet information to call a `mesh.topology.create_entities(tdim-1)` at some point anyhow, so this should be consistent for `tdim=1` (for `tdim=1` we would have the special case that vertices are always available but interprocess facet data might not be available yet)

Agreed.

> One other point, that I need to think more on: if the shared facet case worked before without this additional work path than there must be a no-op short cut to this computation at least in this case, which we should retain.

It can be gated on the same `max_cell_facet_links` parameter that you introduced, and I guess `GhostMode` too which is lucky because @garth-wells fixed that very recently! But as it requires a bit of routing through, so I will do it on a follow up branch.

Edit: The old `GhostMode::shared_facet` route was a bit mysterious, but perfectly legitimate, and this is just a fix for `GhostMode::none`.

cpp/dolfinx/mesh/Topology.cpp
garth-wells: The for loop is clearer.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -821,6 +822,93 @@ build_entity_types(const std::vector<CellType>& cell_types)
   }
   return entity_types;
 }
+
+/// @brief Compute inter-process vertices for a topology with tdim == 1,
```
garth-wells, line 826: It's not clear what an 'inter-process' vertex is.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -821,6 +822,93 @@ build_entity_types(const std::vector<CellType>& cell_types)
...
+    std::vector<std::int32_t> count(attached.begin(),
+                                    std::next(attached.begin(), num_local));
+    for (std::size_t i = 0; i < local_ind.size(); ++i)
+      count[local_ind[i]] += buffer_local[i];
+    std::ranges::transform(count, interprocess.begin(),
+                           [](std::int32_t c) -> std::int8_t { return c > 1; });
+  }
+
+  // Send the verdict for owned vertices back to the ghosting ranks
+  std::ranges::transform(local_ind, buffer_local.begin(),
```
garth-wells, line 898: Using transforms in cases like this is harder to read than a plan loop.

cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -821,6 +822,93 @@ build_entity_types(const std::vector<CellType>& cell_types)
...
+  scatter.scatter_fwd_begin(buffer_local.data(), buffer_ghost.data(), 1,
+                            request);
+  scatter.scatter_fwd_end(request);
+  for (std::size_t i = 0; i < ghost_ind.size(); ++i)
+    interprocess[num_local + ghost_ind[i]] = buffer_ghost[i];
+
+  // Keep only the vertices attached to a cell owned by this rank
+  std::vector<std::int32_t> interprocess_vertices;
+  for (std::int32_t v = 0; v < num_local + num_ghosts; ++v)
+    if (interprocess[v] and attached[v])
```
garth-wells, line 907: Add brace for readability.

garth-wells (changes requested):
Comments inline.

No chance that we can get tdim==1 to follow the same code path as tdim>1?

cpp/dolfinx/mesh/Topology.cpp
jhale: It's described in the docstring:

```
/// Matches the definition used by mesh::compute_entities for dim >= 1:
/// an entity is inter-process if cells owned by two or more ranks are
/// attached to it.
```

And is consistent with the interprocess_facets which is in the public API.

jhale:
Keeping the `GhostMode::shared_facet` optimised path probably requires passing `GhostMode` to `Topology`. If that is an acceptable API change, then I can go ahead and do it.

The more straightforward alternative is to simply document that topologically 1-dimensional meshes with `GhostMode::None` should not be used, in perhaps e.g. `create_mesh`.

Issue #4521, opened 2026-09-15
Expose num_threads when reading in meshes from XDMFFile and VTKHDF


jorgensd:
Both are hard-coded to use 1 thread. As these are the main entry points for large meshes this option should be exposed, ref:
https://github.com/FEniCS/dolfinx/blob/cc5c6a26cbb95033b31f7264adbd7b7d89fea4bb/cpp/dolfinx/io/XDMFFile.cpp#L202
https://github.com/FEniCS/dolfinx/blob/cc5c6a26cbb95033b31f7264adbd7b7d89fea4bb/cpp/dolfinx/io/VTKHDF.h#L472-L475

PR #4527, opened 2026-09-15
Expose num_threads when reading in meshes from XDMFFile and VTKHDF

MariusCausemann:
Addresses #4521.

## Other relevant fixes
- `create_mesh` uses num_threads for entity creation
- `XDMFFile` uses `None` as partitioner in serial rather than the C++ default.


python/test/unit/io/test_xdmf_mesh.py
```diff
@@ -106,6 +108,34 @@ def test_save_and_load_3d_mesh(tempdir, encoding, cell_type):
     )
 
 
+@pytest.mark.parametrize("num_threads", [1, 4])
+def test_read_write_num_threads(tempdir, num_threads):
+    filename = Path(tempdir, "mesh_num_threads.xdmf")
+    mesh = create_unit_cube(MPI.COMM_WORLD, 8, 8, 8)
```
jorgensd:
```suggestion
    mesh = create_unit_cube(MPI.COMM_WORLD, 4, 4, 4)
```

python/test/unit/io/test_vtkhdf.py
```diff
@@ -33,6 +33,40 @@ def test_read_write_vtkhdf_mesh3d():
     assert mesh.topology.index_map(3).size_global == mesh2.topology.index_map(3).size_global
 
 
+@pytest.mark.parametrize("num_threads", [1, 4])
+def test_read_write_vtkhdf_num_threads(num_threads):
+    filename = "example_num_threads.vtkhdf"
+    mesh = create_unit_cube(MPI.COMM_WORLD, 8, 8, 8)
```
jorgensd:
```suggestion
    mesh = create_unit_cube(MPI.COMM_WORLD, 4, 3, 5)
```

jorgensd:
@MariusCausemann can you add some of the notes with respect to your findings of nanobind lookup?

MariusCausemann:
Sure! This is what claude said about it, shortened by me: 

"`_cpp.mesh.create_mesh` has two overloads, and the mixed-topology one is registered first. nanobind tries overloads in the order they were registered. For each one it converts the arguments from left to right and stops at the first argument that doesn't fit. The old read_mesh passed (comm, cells, cmap, x, ...) positionally, with cells as a 2D array of shape (662,243, 4): 
* nanobind tried the 1st overload.
* comm converted.
* cells also converted. The vector converter accepts any sequence, and a 2D NumPy array is a sequence of rows. Each row is a contiguous 1D int64 array, so it passes the per-element check. nanobind built 662,243 row objects.
* cmap failed, because a single CoordinateElement is not a sequence. All that work was thrown away.
* nanobind then tried the 2nd overload. That conversion is cheap, and the call went into C++.

nanobind matches keyword names before it converts anything. The 1st overload has no parameter called element (its elements), so nanobind rejects it straight away."

 On my example mesh, using keywords reduced `read_mesh` timing from 0.781s to 0.507s. 
I found this quite surprising, not sure if there is a general fix. It might be a good idea to use less overloading, and give each function (here the mixed dim read_mesh) its own name.

edit: On a larger mesh, the nanobind issue alone halves the `read_mesh` time, the wasted work is proportional to the number of cells.

schnellerhase:
Yikes, nice find! Does this mean this is now faster because of differing `elements` vs. `element` arg name 🫣?

If so I don't think this is the right fix. We should remove either one of the overloads to unify to a single `create_mesh` or give unique names. Secondly which will probably remove most of the runtime difference we should mark the input arrays with `noconvert()` to disallow conversion to even try.

python/test/unit/io/test_vtkhdf.py
```diff
@@ -33,6 +33,40 @@ def test_read_write_vtkhdf_mesh3d():
...
+            == mesh.topology.index_map(mesh.topology.dim).size_global
+        )
+        assert m.topology.index_map(0).size_global == mesh.topology.index_map(0).size_global
+
+    vol_1 = mesh_1.comm.allreduce(
+        dolfinx.fem.assemble_scalar(dolfinx.fem.form(1 * ufl.dx(domain=mesh_1))), op=MPI.SUM
+    )
+    vol_n = mesh_n.comm.allreduce(
+        dolfinx.fem.assemble_scalar(dolfinx.fem.form(1 * ufl.dx(domain=mesh_n))), op=MPI.SUM
+    )
```
jorgensd, line 63: Add dtype to fem.form.

python/test/unit/io/test_xdmf_mesh.py
```diff
@@ -106,6 +108,34 @@ def test_save_and_load_3d_mesh(tempdir, encoding, cell_type):
...
+            == mesh.topology.index_map(mesh.topology.dim).size_global
+        )
+        assert m.topology.index_map(0).size_global == mesh.topology.index_map(0).size_global
+
+    vol_1 = mesh_1.comm.allreduce(
+        dolfinx.fem.assemble_scalar(dolfinx.fem.form(1 * ufl.dx(domain=mesh_1))), op=MPI.SUM
+    )
+    vol_n = mesh_n.comm.allreduce(
+        dolfinx.fem.assemble_scalar(dolfinx.fem.form(1 * ufl.dx(domain=mesh_n))), op=MPI.SUM
+    )
```
jorgensd, line 141: Add dtype

jorgensd:
> Yikes, nice find! Does this mean this is now faster because of differing `elements` vs. `element` arg name 🫣?
> 
> If so I don't think this is the right fix. We should remove either one of the overloads to unify to a single `create_mesh` or give unique names. Secondly which will probably remove most of the runtime difference we should mark the input arrays with `noconvert()` to disallow conversion to even try.

I think using keyword arguments as done in this PR is fine. It of course doesn't fix the issue on a deeper level.

python/test/unit/io/test_vtkhdf.py
MariusCausemann: done

MariusCausemann:
I added a fix on the nanobind level now, allowing only list or tuple for the cells argument in the mixed dim create_mesh call. This fixes the issue, what do you think @jorgensd  and @schnellerhase ?
I tried `noconvert()`, but it nanobind still turns the 2D array into a tuple of rows, so it did not help. 

I also had claude look through the whole dolfinx repo for similar nanobind issues, but it claims everything else is fine.

schnellerhase:
> I tried noconvert(), but it nanobind still turns the 2D array into a tuple of rows, so it did not help.

You are right, that does not solve the overload standalone - the array also is a sequence so it is no conversion initially.

The new code path does not look preferable to me: a lot of special case handling and explicit casting. Revert [c4dc8c6](https://github.com/FEniCS/dolfinx/pull/4527/commits/c4dc8c6cd472906db194b30d9a4ab845ad85940b), file an issue and can be fixed else where as @jorgensd suggested?

python/dolfinx/wrappers/dolfinx_wrappers/mesh.h
```diff
@@ -377,10 +382,28 @@ void declare_mesh(nb::module_& m, std::string_view type)
...
+          if (!nb::try_cast(c, a))
+          {
+            throw std::invalid_argument(
+                std::format("cells[{}] is not convertible to a 1D int64 array.",
+                            cells_nb.size()));
+          }
+          cells_nb.push_back(std::move(a));
+        }
+
         std::vector<std::span<const std::int64_t>> cells;
```
schnellerhase, line 380: Simpler with `std::vector<std::span<const std::int64_t>> cells = vec_of_spans(cells_nb);`.

MariusCausemann:
I have reverted the last commit, and opened a new issue for the nanobind overload #4542

cpp/dolfinx/mesh/utils.h
```diff
@@ -1520,11 +1520,11 @@ Mesh<typename std::remove_reference_t<typename U::value_type>> create_mesh(
 
       spdlog::debug("Counting entity dofs, dim={}: {}", dim, dim_sum);
       if (dim_sum > 0)
-        topology.create_entities(dim);
+        topology.create_entities(dim, num_threads);
```
jorgensd, line 1523: Interesting that we haveb't picked up on this before.

cpp/dolfinx/mesh/utils.h
```diff
@@ -1520,11 +1520,11 @@ Mesh<typename std::remove_reference_t<typename U::value_type>> create_mesh(
...
     }
 
     if (elements[i].needs_dof_permutations())
-      topology.create_entity_permutations();
+      topology.create_entity_permutations(num_threads);
```
jorgensd, line 1527: Good to have for higher order grids (P3>=)

jorgensd:
Another nice side-effect of this PR is that by using `_get_mesh_partitioner` in XDMFFile.read_mesh we now get the `None` partitioner when in serial, which should speed up mesh creation for those instances by avoiding going through the dual graph route.

jorgensd (approved):

python/test/unit/io/test_vtkhdf.py
```diff
@@ -33,6 +33,46 @@ def test_read_write_vtkhdf_mesh3d():
...
+    assert np.isclose(vol_1, vol_n)
+
+
+def test_read_vtkhdf_num_threads_invalid():
+    filename = "example_num_threads_invalid.vtkhdf"
+    mesh = create_unit_square(MPI.COMM_WORLD, 4, 4)
+    write_mesh(filename, mesh)
+
+    with pytest.raises(RuntimeError):
```
jorgensd:
```suggestion
    with pytest.raises(ValueError):
```

PR #4528, opened 2026-09-15
Bundle libdolfinx with Python wheels

shimwell:
Would it be okay to add this opt-in CMake setting?

Downstream compiled wheels such as [scifem](https://github.com/scientificcomputing/scifem/blob/b415a808d2314c4b9165cc206f95a8546799540d/CMakeLists.txt#L74-L78) link to the stable DOLFINx SONAME.

Auditwheel normally renames `libdolfinx`, making those downstream extensions unable to resolve it.

Installing the imported runtime artifact through CMake preserves both the fully versioned library and its SONAME symlink.

Tested downstream in [shimwell/dolfinx#2](https://github.com/shimwell/dolfinx/pull/2), which builds the FEniCSx wheel stack and scifem. The downstream wheel workflow enables this option and excludes the already bundled `libdolfinx` from auditwheel rewriting.


jhale:
Yes but please remove the modifications to `build-wheels.yml` - you can do what you wish downstream of course.

jhale:
Have you seen this tool? Looks like a more proper solution to this problem.

https://pypi.org/project/consolidatewheels/

shimwell:
> Yes but please remove the modifications to `build-wheels.yml` - you can do what you wish downstream of course.

I have removed the modifications to `build-wheels.yml` as suggested, thanks for take a look

jhale:
Could you investigate the tool I mentioned? In the first instance perhaps it could be applied to the Basix and DOLFINx wheel builds where there is probably a `libbasix-mangled.so` around? If the tool works, I don't really see the need for this type of hack, which just grows exponentially more difficult.

shimwell:
> Could you investigate the tool I mentioned? In the first instance perhaps it could be applied to the Basix and DOLFINx wheel builds where there is probably a `libbasix-mangled.so` around? If the tool works, I don't really see the need for this type of hack, which just grows exponentially more difficult.

@jhale I took a peak at consolidate wheels and put the investigation here https://github.com/shimwell/dolfinx/pull/11

We can PR to consolidate wheels as a longer term solution and I raised an issue over there that we can track https://github.com/amol-/consolidatewheels/issues/4

How about I put in a TODO in this PR code reminding us to remove this if / when a cleaner solution is available?

shimwell:
Chatting on slack just now and best if we find another less hacky way to do this

Issue #4529, opened 2026-09-15
la::MatrixCSR: moved-from matrix retains its MPI_Request


jhale:
`la::MatrixCSR`'s move constructor is `= default` and carries a long-standing todo:

```cpp
/// Move constructor
/// @todo Check handling of MPI_Request
MatrixCSR(MatrixCSR&& A) = default;
```

`_request` is live only between `scatter_rev_begin()`, which starts an
`MPI_Ineighbor_alltoallv` into it, and `scatter_rev_end()`, which calls `MPI_Wait` on
it. `MPI_Request` is a plain handle, so a defaulted move copies the value and leaves
it set in the moved-from matrix: both objects then name the same request.

As far as I can tell this is currently harmless:

- a moved-from object must not be used, so nothing should call `scatter_rev_end()` on
  it, and
- `MatrixCSR` declares no destructor, so nothing acts on the stale handle when the
  moved-from matrix goes out of scope.

The receive buffer also survives a move: `_ghost_value_data_in`'s heap block transfers
to the target, so an in-flight write still lands in storage the target owns.

So this is a latent weakness rather than an active bug. What makes it worth recording
is that the correctness argument rests entirely on the moved-from convention rather
than on the class's own state, and the same shape is not always so forgiving.
`io::XDMFFile` has a defaulted move over a raw `hid_t` where the destructor *does* act
on the handle, and there a move produces a genuine double close — which terminates,
since `io::hdf5::close_file` throws and `~XDMFFile` is implicitly `noexcept`. I am
addressing that case separately.

Making `MatrixCSR` airtight is conceptually one line:

```cpp
_request(std::exchange(A._request, MPI_REQUEST_NULL))
```

but it means replacing `= default` with a hand-written move constructor spelling out
~14 members, which is presumably why the todo is still open. The trade-off is that
maintenance cost against removing the reliance on convention — worth a decision either
way, even if the decision is to keep `= default` and document it.

---

AI assistance: I used Claude Opus 5 via Claude Code to draft parts of this issue. I
reviewed, edited, and take responsibility for the final contribution.

PR #4532, opened 2026-09-15
Fix and complete special member functions across the library

jhale:
Every problem fixed here has the same cause: a class declared **some** of the five
special member functions, and the compiler silently suppressed the rest. A
user-declared destructor suppresses the move operations; a user-declared move
constructor suppresses both assignment operators. The result ranges from a missing
operator to a process abort.

Prompted by a batch of SonarCloud findings, but most of what follows was not among
them — it surfaced from the `static_assert` sweep added in this PR.

### Bugs

**`io::XDMFFile` terminated the process when moved.** The move constructor was
`= default` and `hid_t` is a plain integer handle, so a move left both objects
holding it. Both destructors then called `close()`, and `io::hdf5::close_file`
throws when `H5Fclose` fails — escaping the implicitly `noexcept` `~XDMFFile`.
Reproduced (stack shows `XDMFFile::close` → `std::terminate`), fixed by
transferring the handle, and covered by a regression test that was verified to
fail without the fix.

**`common::Timer` double-counted timings.** Its declared destructor suppressed the
move operations, so `is_move_constructible` was satisfied only by the *implicit
copy constructor* — "moving" a `Timer` copied it, and both copies registered the
same task on destruction. Copies are now deleted. The moves are hand-written, not
defaulted: a moved-from `std::optional` stays engaged, so a defaulted move would
leave the source registering an empty task name.

**`io::VTKFile` was neither movable nor copyable**, contradicting the rule in
`AGENTS.md` that no class in the library deletes a move operation.

### API and ABI changes

Two changes here are deliberate breaks and worth reviewing on their own merits:

- **`la::MatrixCSR` copy operations are deleted** (API). This resolves half of its
  standing `@todo Check handling of MPI_Request`: `_request` is live between
  `scatter_rev_begin()` and `scatter_rev_end()`, and a copy taken in that window
  would wait on data delivered into the original's buffer. Copying also
  deep-copies the matrix and duplicates the communicator, which is collective.
  Nothing in the library, tests or demos copies a same-type `MatrixCSR`; the
  copy-convert constructor is a separate template and is unaffected. The
  remaining move-path concern is filed as #4529.
- **Five `virtual` destructors removed** (ABI) — `Form`, `Expression`,
  `FunctionSpace`, `CoordinateElement`, `la::petsc::Vector`. No class in the
  library derives from any of them, the library declares no virtual member
  functions at all, and the nanobind bindings are plain `nb::class_` with no
  trampolines. These vtables existed solely for the destructor. Note this does not
  by itself restore the implicit moves — a user-declared destructor suppresses
  them whether or not it is virtual.

### Completing the five

`Form`, `Expression`, `MatrixCSR`, `EntityMap` and `MPI::Comm` gain their missing
or undocumented operators. Two needed justification rather than a default:

- `Form` caches `std::span`s in `_edata`/`_cdata` aliasing entity vectors owned by
  `_integrals`, which is why its copy constructor is deleted. Move assignment is
  safe for the same reason as move construction — move-assigning a `std::map`
  transfers its nodes, so the aliased addresses are stable. Verified
  experimentally, not just assumed.
- `MPI::Comm`'s deleted copy assignment now records *why*: `MPI_Comm_dup` is
  collective, so assigning into a live `Comm` would hide a collective duplication
  behind `=`.

### Test

`cpp/test/common/move_semantics.cpp` asserts non-throwing move construction (and
assignment where provided) for 29 classes across every scalar, geometry and mesh
tag type the library instantiates. The property is implicit — every move is
`= default` — so adding a member with a throwing move would otherwise silently
degrade moves to copies with no diagnostic.

It earned its place immediately: it caught a mistake in this PR's own first draft,
where `VTKFile`'s out-of-line defaulted moves were left without explicit
`noexcept` (implicit computation applies only to functions defaulted on their
*first* declaration).

`fem::Expression`'s move assignment is deliberately not asserted non-throwing: it
holds a `std::function` directly, and the standard does not require that type's
move assignment to be `noexcept`, even though libc++ and libstdc++ both make it so.

### Verification

C++ suite 30/30. Python suite 3567 passed, 107 skipped, 21 xfailed.

24 tests in `test_assemble_submesh.py` (`test_codim_1_gradient*`,
`test_codim_1_spatial_coordinate`) fail, but they fail identically on `main`
without this branch — I checked out the base and reproduced the same test IDs.
They were added by aa02d374d6 (#4477), whose CI was still pending at the time of
writing. Unrelated to this PR, but `main` appears to be red.

---

AI assistance: I used Claude Opus 5 via Claude Code to draft parts of this PR. I


cpp/dolfinx/la/MatrixCSR.h
```diff
@@ -199,13 +199,26 @@ class MatrixCSR
...
 
+  // Copy constructor (deleted). Copying deep-copies the matrix data and
+  // duplicates the communicator, which is collective. Both matrices
+  // would also share ::_request, so a copy taken while a scatter is in
+  // flight would wait on data delivered into the original's buffer.
+  MatrixCSR(const MatrixCSR& A) = delete;
+
   /// Move constructor
-  /// @todo Check handling of MPI_Request
+  /// @note ::_request is left set in the moved-from matrix. That is
```
jhale, line 209: It is possible to implement this properly but it requires manually dealing with all the members...

cpp/test/common/move_semantics.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+namespace
+{
+/// @brief Non-throwing move construction, required of every type.
+template <typename T>
+constexpr bool nothrow_move_c = std::is_nothrow_move_constructible_v<T>;
+
+/// @brief Non-throwing move assignment. Not asserted for io::XDMFFile,
+/// which does not provide a move assignment operator, nor for
+/// fem::Expression, whose std::function member is not required by the
```
jhale, line 57: Not really true in practice, will fix.

cpp/dolfinx/io/XDMFFile.h
```diff
@@ -71,7 +71,11 @@ class XDMFFile
            std::string_view file_mode, Encoding encoding = Encoding::HDF5);
 
   /// Move constructor
-  XDMFFile(XDMFFile&&) = default;
+  /// @note Defined in the source file, where pugi::xml_document is
```
jhale: Convert to comment.

cpp/test/CMakeLists.txt
```diff
@@ -39,9 +39,11 @@ add_executable(
   petsc.cpp
   slepc.cpp
   io.cpp
+  io_xdmf.cpp
```
jhale: Put into io.xdmf

cpp/test/io_xdmf.cpp
```diff
@@ -0,0 +1,27 @@
+// Copyright (C) 2026 Jack S. Hale
+//
+// This file is part of DOLFINx (https://www.fenicsproject.org)
+//
+// SPDX-License-Identifier:    LGPL-3.0-or-later
+
+#include <catch2/catch_test_macros.hpp>
+#include <dolfinx/io/XDMFFile.h>
+#include <filesystem>
```
jhale: Remove.

garth-wells (approved):

garth-wells (approved):

PR #4533, opened 2026-09-15
Select OpenMPI's `basic` topology component in the test environment

jorgensd:
## What

One environment variable in the `dev-env` stage of
`docker/Dockerfile.test-env`:

```dockerfile
ENV OMPI_MCA_topo=basic
```

## Why

Ubuntu 26.04 ships OpenMPI 5.0.10:

```
Open MPI v5.0.10, package: Debian OpenMPI, ident: 5.0.10, repo rev: v5.0.10, Feb 23, 2026
```

In that version `MPI_Dist_graph_create` deadlocks at random under the
`treematch` topology component — [open-mpi/ompi#13918][issue], confirmed
there by an Open MPI maintainer and reported as affecting *all* versions at
the time. From the fix, [open-mpi/ompi#13931][fix]:

> `MPI_Dist_graph_create` exchanges distributed-graph edge metadata with PML
> sends and receives on the parent communicator before the new communicator
> is fully created. The topo base code used hard-coded internal tags -50 and
> -51. Those tags are reachable by the dynamic nonblocking collective tag
> allocator. In repeated communicator creation, `comm->c_nbc_tag` can count
> down into -50/-51. […] The resulting tag collision can leave the CID
> allocation request waiting forever.

`OMPI_MCA_topo=basic` is the workaround recommended on that issue by the
maintainer who diagnosed it. MPICH ignores `OMPI_MCA_*`, so setting it is
unconditionally safe.

## How it was found

Debugging [jorgensd/adios4dolfinx#218][a4d]. Its 4-rank `pytest` job hung
until GitHub's 6-hour job limit. With per-rank stack dumps, every rank was
inside `MPI_Dist_graph_create` having passed valid, in-range neighbour
lists, after a number of tests that varied wildly between runs — 16, 28,
47, 77, 502, 686 — while open file descriptors and RSS stayed flat.

A ~70-line loop using only the standard library and `mpi4py` reproduces it
inside this image, and passes with the component selected. That matches the
upstream mechanism: a counter counting down into a reserved tag, rather than
any resource being exhausted.

## Does this affect DOLFINx itself?

Not directly. DOLFINx uses only `MPI_Dist_graph_create_adjacent`, which does
not perform the edge-metadata exchange the bug lives in; the non-adjacent
`MPI_Dist_graph_create` does not appear anywhere in `cpp/` or `python/`.

The change is therefore for the benefit of downstream users of the image —
packages and user code that do call the non-adjacent form, adios4dolfinx
among them. Reviewers may reasonably take the view that such projects should
set this themselves; the argument for setting it here is that the image is
the environment those projects are told to build against, and the variable
costs nothing.

## The real resolution

Upgrade OpenMPI. The fix is [#13931][fix] on `main` (merged 2026-06-03) and
[#13933][backport] on `v5.0.x` (merged 2026-06-24), first carried by the
**v5.0.11** tag (2026-08-25). Ubuntu 26.04 currently ships 5.0.10, which
predates it.

**Removal condition:** drop this `ENV` as soon as the base image provides
OpenMPI >= 5.0.11. `ompi_info --version` in the container says when.

## Testing

- `mpirun -n 4` over a repeated `MPI_Dist_graph_create` loop in this image:
  deadlocks without the variable, completes with it.
- The adios4dolfinx parallel suite that motivated this: 6/6 runs deadlocked
  before, green after.

## AI assistance

AI assistance: I used Claude Code to draft parts of this PR. I reviewed,

[issue]: https://github.com/open-mpi/ompi/issues/13918
[fix]: https://github.com/open-mpi/ompi/pull/13931
[backport]: https://github.com/open-mpi/ompi/pull/13933
[a4d]: https://github.com/jorgensd/adios4dolfinx/pull/218


jhale (approved):

drew-parsons (approved):
Looks safe.

Future releases of Debian (and ubuntu) should provide the fixed openmpi 5.0.11, but for the existing releases this environment variable will stabilise tests.

jhale:
@drew-parsons Is there any chance of backporting? The bug fix patch is trivial.

garth-wells (approved):

drew-parsons:
@jhale do you mean backporting the openmpi patch?  It can be done, but the process is a bit of a pain.
For general users affected by the problem I'd recommend the workaround of running with OMPI_MCA_topo=basic

With Ubuntu it's not so easily to backport fixes after release. For instance they never fixed a critical bug in mpich in Ubuntu 24.04 (the debian package got fixed but Ubuntu didn't manage to pull in the update).

PR #4539, opened 2026-09-17
Avoid heap allocations in hot loops (geometry, mesh, graph partitioning)

### LINKED ISSUE #4538: Heap allocations inside hot loops (geometry, mesh, graph partitioning)
A review of `cpp/dolfinx/` for heap allocations inside loops whose trip count scales with mesh/problem size (cells, facets, points) turned up a handful of cases outside the otherwise well-optimized assembly/sparsity-pattern/topology-computation paths, which already use caller-provided pre-sized buffers.

## Findings

**1. `geometry/utils.h:265-275`, `impl::_compute_closest_entity`**

For the `tdim > 0` branch, the squared-distance-to-entity computation goes through `squared_distance()` -> `shortest_vector()`, which heap-allocates a `std::vector<T>` (twice, including an inner `d` vector), per call. This executes once per BVH leaf visited during nearest-point descent, per query point, so it can fire many times across millions of query points (mesh-to-mesh interpolation, contact/collision detection, point evaluation). The `tdim == 0` branch immediately above already avoids this using a stack `std::array<T, 6>` — the `tdim > 0` branch should get the same stack-based treatment instead of routing through the vector-returning batch API built for lists of entities.

**2. `geometry/utils.h:314`, `impl::_compute_collisions_point`**

A fresh `std::deque<std::int32_t> stack` is default-constructed on every call, and this function runs once per query point from `compute_collisions()` — can be millions of points. `std::deque`'s default constructor always allocates its first internal chunk. The output `entities` vector is already passed in by reference and reused by the caller; `stack` should be handled the same way (passed in and cleared, or replaced with a reused `std::vector` used as a stack).

**3. `mesh/Topology.cpp:1670-1704`, `mesh::entities_to_index`**

Builds `std::map<std::vector<std::int32_t>, std::int32_t> entity_key_to_index`, inserting once per mesh entity of the given dimension — can be millions of edges/facets on a large mesh. Each `insert` heap-allocates both a copy of the key vector (the map's stored key) and a red-black-tree node, i.e. two allocations per entity. This is inconsistent with neighboring code (e.g. `mesh/topologycomputation.cpp`), which uses flat, sort-based or `boost::unordered_flat_map`/`_set` techniques specifically to avoid this pattern. A fixed-size key (`std::array<std::int32_t, 4>`, since the vertex count per entity is known and small) with `boost::unordered_flat_map` would remove both per-entity allocations.

**4. `graph/partitioners.cpp:414-427`, SCOTCH ghosting in the distributed partitioner**

`std::map<std::int32_t, std::set<std::int32_t>> local_node_to_dests` is populated in a loop over every local graph node (e.g. every cell being partitioned — can be millions) times its edges. Each boundary edge triggers a `std::map` node allocation (on first access to a new key, default-constructing a `std::set`) plus a `std::set` node allocation. This stands out because `compute_destination_ranks` earlier in the same file explicitly avoids this cost via `boost::unordered_flat_set`/sorted-vector techniques, with comments explaining why. Replacing with `boost::unordered_flat_map<std::int32_t, boost::container::flat_set<std::int32_t>>`, or building a flat `(node0, node1_rank)` pair list and sorting/deduping (the pattern already used elsewhere in this file), would remove essentially all per-edge allocations.

**5. `geometry/BoundingBoxTree.h:84-200`, `_build_from_leaf` / `_build_from_point`**

Output vectors (`bboxes`, `bbox_coordinates`) are grown via `push_back`/`back_inserter` through the recursive tree build, without a `reserve()`, even though the final size is known up front: a binary tree built from `N` leaves by bisection has exactly `2N - 1` nodes, so `bboxes.size() == 2*(2N-1)` and `bbox_coordinates.size() == 6*(2N-1)` are computable before recursing. This runs once per `BoundingBoxTree` construction (once per mesh, or per rebuild for a moving mesh) — a cheap `reserve()` at the entry points would avoid O(log N) reallocation/copy passes.

**Minor:** `mesh/utils.h`, `locate_entities`/`locate_entities_boundary` (~lines 593-609, 697-712) grow their output via `push_back` without an upper-bound `reserve()`. Low severity since growth is only geometric (amortized), but a cheap `reserve(e_to_v->num_nodes())` would remove the reallocation passes entirely.

## Suggested priority

Items 1-4 are in genuinely hot, size-scaling loops and are worth fixing. Item 5 and the minor note are cheap, low-risk cleanups.

garth-wells:
Closes #4538.

Replaces per-iteration heap allocations in loops whose trip count scales with mesh/problem size with caller-reused scratch buffers, `reserve()`, or flat/fixed-key containers. Started from the four flagged findings in #4538 and extended to further hot loops found in a follow-up pass over `geometry`, `mesh`, `graph`, `io`, `refinement` and `common`.

## Geometry (point queries)

- **`geometry::impl::_compute_closest_entity`**: the `tdim > 0` leaf case routed through the vector-returning `squared_distance()`/`shortest_vector()` batch API for a single entity, allocating on every candidate leaf visited per query point. Added `squared_distance_entity()`, taking a `ClosestEntityScratch` (a geometry-dof scratch buffer plus the coordinate element's `ElementDofLayout`) threaded through the recursion and reused across query points by the caller -- this also avoids re-deriving the dof layout on every leaf visit.
- **`geometry::impl::_compute_collisions_point`**: replaced the per-point `std::deque` (whose default constructor always heap-allocates its first chunk) with a `std::vector`-backed stack reused across query points -- the usage was already LIFO-only (`push_back`/`back`/`pop_back`), so the swap is behavior-preserving.
- **`geometry::BoundingBoxTree`**: `reserve()` the output vectors using the known final size (a binary tree built by bisection from `N` leaves has exactly `2N - 1` nodes) instead of growing them via `push_back`/`back_inserter` through the recursion.

## Mesh and graph partitioning

- **`mesh::entities_to_index`**: replaced `std::map<std::vector<int32_t>, int32_t>` (two heap allocations per entity: the map node and the vector key copy) with a `boost::unordered_flat_map` keyed on a fixed-size, `-1`-padded `std::array<int32_t, 8>` (8 being the largest vertex count of any supported cell type, hexahedron).
- **`mesh::topologycomputation::get_ownership`**: index directly into the input span instead of copying it into a fresh vector on every call -- this function is documented as running up to millions of times for a large, highly-ghosted mesh.
- **`graph::scotch::partitioner`**: replaced `std::map<int32_t, std::set<int32_t>>` for the ghosting destination-rank lookup (allocating a map node and a set node per boundary edge) with a flat vector of `(node0, dest_rank)` pairs, deduplicated via `boost::unordered_flat_set` and sorted -- matching the approach already used in `compute_destination_ranks` in the same file, which now also `reserve()`s its send buffer.

## `common::IndexMap`

- `compute_submap_indices`, `stack_index_maps`, `index_to_dest_ranks`: `reserve()` output and send buffers where an exact size or tight upper bound is already in hand at that point -- the received-index count, the ghost count, or the per-destination sizes just computed into `send_sizes`.

No counting pass is added. An earlier revision of this branch also sized each per-rank bucket in `communicate_ghosts_to_owners`, `stack_index_maps` and `index_to_dest_ranks` with a count-then-reserve pre-pass; that has been dropped. Each pre-pass duplicated the fill loop's index computation -- a second `lower_bound` binary search per ghost in the first two, and a restatement of the nested per-rank loop in the third -- which costs more than the geometric reallocations of trivially-copyable data it avoided, and left the duplicated predicates (notably `include_ghost`) free to drift from the fill loops without any test failing.

Right-sizing the buckets would in any case leave the vector-of-vectors structure in place. Converting these packs to the counts/offsets/flat-buffer layout used elsewhere in this file would remove the per-rank allocations outright, and is left to a separate change.

## IO and refinement

- **`io::distribute_entity_data`**: replaced a `std::map<int64_t, int32_t>` (one tree-node allocation per (cell, vertex) pair) with a `boost::unordered_flat_map`; `reserve()` the output vectors.
- **`io::VTKHDF::read_mesh`**: `reserve()` each per-cell-type buffer using an exact pre-pass over per-cell node counts, instead of growing via `push_back` with no reserve.
- **`refinement::plaza` (`compute_refinement`)**: `reserve()` `cell_topology`/`parent_cell`/`parent_facet` using the known maximum children-per-parent-cell bound (`2^tdim`, full/red refinement); hoist the per-cell `longest_edge` buffer out of the cell loop and reuse it via `clear()`.
- **`refinement::uniform_refine`**: hoist the per-cell `entities` buffer out of the cell loop and reuse it via `clear()`, instead of a fresh allocation per cell.

## Benchmarking

Built matched baseline (`main` at this PR's merge-base) and branch stacks and measured the changes that are reachable from Python, isolating each from unrelated noise (single-rank micro-benchmarks for the local/non-collective geometry and mesh operations; 1/8/16/32-MPI-rank runs of the full Poisson mesh-construction + assembly pipeline for the collective ones). Correctness was checked throughout: matrix nnz/Frobenius-norm and `compute_closest_entity`/`compute_collisions_points` outputs are bit-identical between baseline and branch.

| Change | Benchmark | Result |
|---|---|---|
| `mesh::entities_to_index` | `meshtags_from_entities` over 12.06M facets (unit cube N=100, 1 rank) | **~18-20% faster** (5.6-5.9s -> 4.66-4.76s median) |
| `geometry::BoundingBoxTree` construction | Tree build, 384k-cell cube (N=40), 1 rank | **~5-14% faster** (~0.33-0.35s -> ~0.29-0.31s) |
| `geometry::_compute_closest_entity` | 50k-200k random query points against the same mesh | ~2-3% faster -- dominated by the GJK distance evaluation and tree descent, not the allocation removed |
| `geometry::_compute_collisions_point` | `compute_collisions_points`, same query sets | No measurable difference at this scale |
| `common::IndexMap` / `io` / `refinement` / graph ghosting-map `reserve()`s | Full Poisson mesh-construction + matrix/vector assembly, 1/8/16/32 MPI ranks, up to 6M cells | No measurable end-to-end effect (within run-to-run noise); these reduce allocator/memory-churn pressure rather than wall-clock time at this scale |

## Testing

- `ninja` in the C++ build (`-Werror`) and the nanobind Python bindings build both compile cleanly.
- `clang-format --dry-run --Werror` passes on all touched files.
- C++ unit test suite: 90 passed, 3 skipped (parallel-only), 29858 assertions passed.
- Python test suites: `geometry`, `mesh` (including `test_meshtags.py`, `test_manifold_point_search.py`), `graph`, and `fem/test_interpolation.py` all pass (777 tests).
- 3-rank MPI sanity check of the SCOTCH ghosting path (`graph.partitioner_scotch()` with `GhostMode.shared_facet` on a unit cube): correct total cell count and valid ghost ownership across ranks.
- Doxygen (`WARN_IF_UNDOCUMENTED` + `WARN_AS_ERROR`) builds clean.


cpp/dolfinx/mesh/Topology.cpp
```diff
@@ -1665,15 +1665,29 @@ mesh::entities_to_index(const Topology& topology, int dim,
...
+  // Key is the sorted local vertex indices of an entity, padded with -1
+  // up to the largest vertex count of any supported cell type
+  // (hexahedron, 8 vertices). A fixed-size array key lets
+  // boost::unordered_flat_map store it inline, avoiding a per-entity
+  // heap allocation that a std::vector key (or a node-based std::map)
+  // would incur -- this loop runs once per mesh entity, which can
+  // number in the millions.
+  constexpr int max_vertices_per_entity = 8;
+  assert(num_vertices_per_entity <= max_vertices_per_entity);
+  using Key = std::array<std::int32_t, max_vertices_per_entity>;
```
jhale, line 1671: Wouldn't bother, type fully?

garth-wells: I think as is is clearer for the reader.

cpp/dolfinx/geometry/utils.h
```diff
@@ -238,13 +238,95 @@ constexpr bool bbox_in_bbox(std::span<const T, 6> a, std::span<const T, 6> b)
...
+  const mesh::Geometry<T>& geometry = mesh.geometry();
+  std::span<const T> geom_dofs = geometry.x();
+  auto x_dofmap = geometry.dofmaps().front();
+
+  std::array<T, 3> d;
+  if (dim == tdim)
+  {
+    assert(entity >= 0);
+    auto dofs = md::submdspan(x_dofmap, entity, md::full_extent);
+    scratch.nodes.resize(3 * dofs.size());
```
jorgensd, line 275: @garth-wells why do we have this resize? Couldn't we have this vector being the longest of all x_dof_nodes in a mixed topology, and pass a span based on the geometry cmap num nodes to GJK?

cpp/dolfinx/geometry/utils.h
```diff
@@ -238,13 +238,95 @@ constexpr bool bbox_in_bbox(std::span<const T, 6> a, std::span<const T, 6> b)
...
+    auto it0 = std::find(cell_entities.begin(), cell_entities.end(), entity);
+    assert(it0 != cell_entities.end());
+    const int local_cell_entity
+        = std::ranges::distance(cell_entities.begin(), it0);
+
+    // Tabulate geometry dofs for the entity
+    auto dofs = md::submdspan(x_dofmap, c, md::full_extent);
+    const std::vector<int>& entity_dofs
+        = scratch.cmap_dof_layout.entity_closure_dofs(dim, local_cell_entity);
+    scratch.nodes.resize(3 * entity_dofs.size());
```
jorgensd, line 309: Same comment here. As the number of closure dofs < max_geom_nodes_per_cell_type we could use a span here and avoid resizing.

PR #4544, opened 2026-09-18
Discontinuous CoordinateElement

jorgensd:
In the current version of the code, it is possible to pass a discontinuous element to the `CoordinateElement` constructor.
However, this is not respected in `interpolate_geometry`.
In `dolfinx.mesh.create_mesh` the `CoordinateElement` cannot be discontinuous, as it is used to construct the the topology through it's element dof layout (https://github.com/feniCS/dolfinx/) which is called in `create_mesh`. One could argue for a bigger refactor there, but that is not the goal of this PR.

The goal of this PR is to accept discontinuous CoordinateElement in `interpolate_geometry`. This would simplify curving periodic meshes, i.e. make them periodic on a linear grid, then curve it if needed.

This contribution was created with the help of CLAUDE (Opus 5).


cpp/dolfinx/fem/CoordinateElement.cpp
```diff
@@ -240,6 +240,13 @@ std::uint64_t CoordinateElement<T>::hash() const
 }
 //-----------------------------------------------------------------------------
 template <std::floating_point T>
+bool CoordinateElement<T>::is_discontinuous() const
+{
+  assert(_element);
+  return _element->discontinuous();
```
jhale, line 246: is_affine returns _is_affine, harmonize

python/test/unit/fem/test_dofmap.py
```diff
@@ -499,6 +507,44 @@ def warp(x):
...
+def test_discontinuous_coordinate_element(dtype, degree):
+    """A discontinuous coordinate element maps a single cell as usual."""
+    cmap = coordinate_element(
+        create_element(
+            ElementFamily.P,
+            BasixCellType.triangle,
+            degree,
+            LagrangeVariant.gll_isaac,
+            DPCVariant.unset,
+            True,
```
jhale, line 521: Name the argument, easier to read

python/test/unit/fem/test_interpolate_geometry.py
```diff
@@ -157,3 +159,89 @@ def test_curve_mesh(degree, dtype, R, cell_type, lagrange_variant):
...
+def _discontinuous_cmap(cell_type, degree, dtype, variant=LagrangeVariant.gll_isaac):
+    """Discontinuous Lagrange coordinate element."""
+    return coordinate_element(
+        basix.create_element(
+            basix.ElementFamily.P,
+            cell_type,
+            degree,
+            variant,
+            basix.DPCVariant.unset,
+            True,
```
jhale, line 173: Name argument

jhale (commented):
How much testing do we have for the whole pipeline with a DG mesh? Might be worth hooking up a Poisson solve.

cpp/dolfinx/fem/CoordinateElement.cpp
jorgensd: Do we really want to break API compatibility by changing `is_affine` to `affine`?

I don't really want to change the basix API (as `_element` is a basix element) to be `is_discontinuous` rather than `discontinuous`.

jhale (approved):

jorgensd:
> How much testing do we have for the whole pipeline with a DG mesh? Might be worth hooking up a Poisson solve.

Added an assembly (native matrix/vector) check to verify it, as we don't need linear algebra verification.

PR #4545, opened 2026-09-18
Harden `create_mesh` input checks

jorgensd:
1. Global index cell node index out of range (> max global node index -> throw error
2. Nodes not used in cells are silently dropped. Add warning under debug mode only.

Merged existing allreduce test with 1 to avoid multiple allreduce calls.

## AI disclosure statement

This code was created with the help of CLAUDE.
- Issue 1 was flagged by CLAUDE when I was looking into issue 2, which has been observed by many users, but usually when reading in entity tags at a later stage, resulting in a segfault. This would at least diagnose some of these issues early on.


cpp/dolfinx/common/MPI.h
```diff
@@ -825,23 +826,34 @@ distribute_data(MPI_Comm comm0, std::span<const std::int64_t> indices,
...
-    if (invalid)
-      throw std::invalid_argument("Non-empty data on null MPI communicator");
-  }
-
   std::int64_t shape0 = 0;
   int err
       = MPI_Allreduce(&shape0_local, &shape0, 1, MPI_INT64_T, MPI_SUM, comm0);
   dolfinx::MPI::check_error(comm0, err);
 
+  // Reduce the rank-local checks so that all ranks throw, or none: an
```
jhale, line 834: No need for comment, well understood.

cpp/dolfinx/common/MPI.h
```diff
@@ -825,23 +826,34 @@ distribute_data(MPI_Comm comm0, std::span<const std::int64_t> indices,
...
+  // early throw would block peers in the collectives below. An
+  // out-of-range index otherwise maps to a non-existent owner rank.
+  {
+    std::array<int, 2> invalid_local
+        = {comm1 == MPI_COMM_NULL and !x.empty(),
+           !std::ranges::all_of(indices, [shape0](std::int64_t i)
+                                { return i >= 0 and i < shape0; })};
+    std::array<int, 2> invalid{0, 0};
+    err = MPI_Allreduce(invalid_local.data(), invalid.data(), 2, MPI_INT,
```
jhale, line 843: My opinion is that this double packing is overoptimisation outside of hot loops. Makes code harder to follow as well. LLM seems to love doing it.

cpp/dolfinx/mesh/utils.h
```diff
@@ -1566,6 +1566,24 @@ Mesh<typename std::remove_reference_t<typename U::value_type>> create_mesh(
...
 
+#ifndef NDEBUG
+  // Nodes in `x` that no cell references are dropped silently.
+  {
+    std::int64_t num_nodes_local = xshape[0];
+    std::int64_t num_nodes = 0;
+    int err = MPI_Allreduce(&num_nodes_local, &num_nodes, 1,
+                            dolfinx::MPI::mpi_t<std::int64_t>, MPI_SUM, comm);
+    dolfinx::MPI::check_error(comm, err);
+    if (std::int64_t num_used = geometry.index_map()->size_global();
```
jhale, line 1577: Assignment in if statement really needed?

cpp/dolfinx/common/MPI.h
```diff
@@ -825,23 +826,34 @@ distribute_data(MPI_Comm comm0, std::span<const std::int64_t> indices,
...
-
   std::int64_t shape0 = 0;
   int err
       = MPI_Allreduce(&shape0_local, &shape0, 1, MPI_INT64_T, MPI_SUM, comm0);
   dolfinx::MPI::check_error(comm0, err);
 
+  // Reduce the rank-local checks so that all ranks throw, or none: an
+  // early throw would block peers in the collectives below. An
+  // out-of-range index otherwise maps to a non-existent owner rank.
+  {
```
garth-wells, line 847: Policy is that we perform checks with cost greater than O(1) only in debug. Same for collective operations that are just for error checking.

cpp/dolfinx/mesh/utils.h
jorgensd: Makes formatting less verbose as one is reusing it.

cpp/dolfinx/common/MPI.h
jorgensd: As Garth wants to move this to debug mode only, the optimzation of the reduction wouldn't be needed.

cpp/dolfinx/common/MPI.h
jorgensd: Fixed.

PR #4548, opened 2026-09-19
cpp: split fem utility headers

garth-wells:
Follow-up to #4547. This splits the broad fem/utils.h into focused public headers for integration domains, sparsity patterns, dofmap/function-space factories, form factories, expression factories, and geometry interpolation.

fem/utils.h remains as a small compatibility umbrella, while all in-tree consumers now include only the focused header they use. This removes its role as an internal dependency hub without breaking existing direct includes of fem/utils.h.

Testing:
- clang-format --dry-run --Werror
- Standalone compilation of all six new public headers and the compatibility header
- Developer-mode C++ library and unittests build
- ctest -R unittests: 3/3 passed (serial, 2 ranks, 3 ranks)

AI assistance: I used Codex (GPT-5) to draft parts of this PR. I reviewed, edited, tested, and take responsibility for the final contribution.


jhale:
Thanks, working on this file was getting a bit tough.

PR #4553, opened 2026-09-20
demo: add FFCx-generated kernel variant to custom_kernel demo

garth-wells:
## Summary

The `custom_kernel` demo compares three ways to assemble the same P1 mass
matrix and unit-source load vector:

1. hand-written kernels wrapped in `fem::Form` and passed to the public
   assembly functions;
2. the same hand-written kernels passed as concrete lambda types to the
   low-level cell assembly functions; and
3. kernels generated from `mass.py` by FFCx and passed through the same
   low-level assembly path.

For the third route, CMake runs FFCx and `main.cpp` includes the generated C
source. The generated function definitions and the templated DOLFINx cell
loops are therefore visible in the same translation unit, allowing direct
calls to be inlined. The demo currently calls the generated integral functions
by their hash-derived symbols.

The low-level examples also propagate problem-specific compile-time
information using static `mdspan` extents and `std::integral_constant` block
sizes. The documentation explains the reference-cell tensors, UFCx kernel ABI,
data normally organised by `fem::Form`, and the trade-off involved in using
internal assembly interfaces.

`custom_kernel` is excluded from `generate-cmakefiles.py` because its
`CMakeLists.txt` is intentionally hand-maintained to generate and include the
FFCx source.

## Why this matters

The demo shows that the low-level, concrete-callable assembly path is not
limited to hand-written kernels. Compiler-generated kernels can use the same
path without type erasure or routing assembly through `fem::Form`.

This is an advanced, performance-oriented example. Normal applications should
continue to express forms in UFL and use the public `fem::Form` assembly
interface.

## Simplification enabled by FFCx #885

[FEniCS/ffcx#885](https://github.com/FEniCS/ffcx/pull/885) adds stable,
user-controlled names for generated integral kernels. Once that change is
available, this demo can replace calls such as

```cpp
tabulate_tensor_integral_0f31b12dbe4babd413e1ac21cdf56eb73b71c30d_triangle(...);
```

with predictable names declared by the generated header. The names can be
derived automatically from the Python form variables or selected in
`mass.py` using `ffcx_kernel_name` integral metadata.

That follow-up will simplify this demo by removing:

- hard-coded SHA-1-derived function names;
- coupling to changes in the UFL signature, FFCx version or UFCx signature;
- the need to inspect generated source whenever a hash changes; and
- any temptation to recover the kernel indirectly through a `ufcx_form`
  function pointer, which would obscure the direct inlineable call.

The generated source can remain included in `main.cpp`, so the stable direct
call preserves the inlining demonstrated by this PR.

## Testing

- Built the demo in DOLFINx `Developer` mode.
- Ran the demo for `float` and `double`; all form-based, hand-written and
  FFCx-generated assembly norm comparisons pass.
- Rendered the Jupytext/MyST page and completed a full Sphinx parse.
- `clang-format --dry-run --Werror` and `gersemi --check` pass.

## AI assistance disclosure

This change was developed with assistance from Claude Code (Anthropic) and
OpenAI Codex, following [CONTRIBUTING.md](CONTRIBUTING.md). All changes were
reviewed and validated with formatting, build, runtime and documentation
checks.


garth-wells:
Pushed a follow-up commit addressing a real gap in the first version: calling the FFCx-generated kernel through a function pointer into a separately-linked `mass.c` meant it could not actually be inlined into the assembler loop, unlike the hand-written kernel.

`mass.c` is now `#include`d directly into `main.cpp` (with `restrict` redefined locally and `-Wunused-parameter` disabled around just the include, since the generated code relies on the bare C keyword and routinely leaves some fixed parameters unused) instead of being compiled and linked as its own translation unit. `CMakeLists.txt` no longer compiles `mass.c` separately; an explicit `OBJECT_DEPENDS` keeps build ordering correct.

Verified by disassembling the built demo binary: zero `call` instructions to the generated `tabulate_tensor` functions anywhere in the binary (the standalone copies remain only because their address is taken for `ufcx_integral`'s function-pointer field) -- every actual invocation is inlined into the assembler loop, matching the hand-written kernel. Re-checked numerically (matrix/vector norms unchanged) and re-ran `ctest` at 1/2/3 ranks -- all pass.

cpp/demo/custom_kernel/main.cpp
```diff
@@ -1,73 +1,186 @@
...
+// to the hand-written kernels below.
+//
+// ````{admonition} UFL forms implemented in Python
+// :class: dropdown
+// ![ufl-code]
+// ````
+//
+// ## C++ program
+//
+// FFCx normally compiles generated C source separately. Here it is included
```
jhale, line 84: Normal DOLFINx solvers compile the FFCx generated C source within a separate translation unit.

cpp/demo/custom_kernel/main.cpp
```diff
@@ -1,73 +1,186 @@
...
 #include <tuple>
 #include <type_traits>
 #include <utility>
 #include <vector>
 
+// ## Compile-time shapes
+//
+// The low-level assembly functions are templates over their `mdspan`
+// arguments. Encoding dimensions that are fixed by the problem in the
+// `mdspan` type makes those values available while the cell loop is compiled.
```
jhale, line 137: ...which allows the compiler extra scope for optimisation.

cpp/demo/custom_kernel/main.cpp
```diff
@@ -226,103 +387,159 @@ double assemble_vector1(const mesh::Geometry<T>& g, const fem::DofMap& dofmap,
...
+  if constexpr (std::is_same_v<T, double>)
+  {
+    // The lambdas call the generated functions directly. Because mass.c was
+    // included above, the optimiser can see and inline their definitions.
+    auto kernel_a_ffcx
+        = [](T* A, const T* w, const T* c, const T* coordinate_dofs,
+             const int* entity_local_index,
+             const uint8_t* quadrature_permutation, void* d)
+    {
+      tabulate_tensor_integral_0f31b12dbe4babd413e1ac21cdf56eb73b71c30d_triangle(
```
jhale: Is there an alias without the giant hash?

cpp/demo/custom_kernel/main.cpp
```diff
@@ -226,103 +387,159 @@ double assemble_vector1(const mesh::Geometry<T>& g, const fem::DofMap& dofmap,
...
+          A, w, c, coordinate_dofs, entity_local_index, quadrature_permutation,
+          d);
+    };
+    auto kernel_L_ffcx
+        = [](T* b, const T* w, const T* c, const T* coordinate_dofs,
+             const int* entity_local_index,
+             const uint8_t* quadrature_permutation, void* d)
+    {
+      tabulate_tensor_integral_0372d68a862ab7a471408bc8a4adef5abb12013d_triangle(
```
jhale: Same comment on hash

cpp/demo/custom_kernel/main.cpp
```diff
@@ -226,103 +387,159 @@ double assemble_vector1(const mesh::Geometry<T>& g, const fem::DofMap& dofmap,
...
+  const double norm_A1
+      = assemble_matrix1<T>(mesh->geometry(), *V->dofmap(), kernel_a, cells);
+  const double norm_b1
+      = assemble_vector1<T>(mesh->geometry(), *V->dofmap(), kernel_L, cells);
+  check_norm<T>(norm_A1, norm_A0);
+  check_norm<T>(norm_b1, norm_b0);
+
+  // Route 3: repeat the direct assembly using the kernels generated from
+  // mass.py. This FFCx invocation generates only float64 kernels.
+  if constexpr (std::is_same_v<T, double>)
```
jhale, line 509: What happens if T is not double?

cpp/demo/custom_kernel/main.cpp
```diff
@@ -226,103 +387,159 @@ double assemble_vector1(const mesh::Geometry<T>& g, const fem::DofMap& dofmap,
...
     T scale = detJ(mdspan2_t<const T, 3, 3>(x));
-    for (std::size_t i = 0; i < 3; ++i)
+    for (std::size_t i = 0; i < p1_triangle_dofs_per_cell; ++i)
       b[i] = scale * b_hat[i];
   };
 
-  // Assemble matrix and vector using std::function kernel
-  assemble_matrix0<T>(V, kernel_a, cells);
-  assemble_vector0<T>(V, kernel_L, cells);
+  // Route 1: wrap each hand-written kernel in a Form and use public assembly.
```
jhale, line 494: Remind reader about why this leads to an overhead.

cpp/demo/custom_kernel/main.cpp
garth-wells: No, but function names can now be specified (see https://github.com/FEniCS/ffcx/pull/885).

cpp/demo/custom_kernel/main.cpp
garth-wells: Reworded in 260b73f26b: the intro now says normal solvers compile the generated C in a separate TU, so the assembler only reaches kernels via function pointers, and that including it here allows inlining.

cpp/demo/custom_kernel/main.cpp
garth-wells: Added in 260b73f26b, with fully unrolling loops over cell dofs as an example.

cpp/demo/custom_kernel/main.cpp
garth-wells: Added in 260b73f26b: Form stores kernels as type-erased `std::function`, so each cell makes an indirect, non-inlinable call, which is significant for small kernels.

cpp/demo/custom_kernel/main.cpp
garth-wells: Route 3 is compiled only for `T = double` (`if constexpr`), since FFCx generates float64 kernels by default and they take `double*`. The comment now says this explicitly and notes that float32 would need a second FFCx run with `--scalar_type float32` and different kernel names (260b73f26b).

jhale (approved):

PR #4558, opened 2026-09-21
Fix `create_mesh` nanobind overload

### LINKED ISSUE #4542: nanobind overload slows down create_mesh
From #4521 

> "_cpp.mesh.create_mesh has two overloads, and the mixed-topology one is registered first. nanobind tries overloads in the order they were registered. For each one it converts the arguments from left to right and stops at the first argument that doesn't fit. The old read_mesh passed (comm, cells, cmap, x, ...) positionally, with cells as a 2D array of shape (662,243, 4):
>  * nanobind tried the 1st overload.
>  * comm converted.
 > *cells also converted. The vector converter accepts any sequence, and a 2D NumPy array is a sequence of rows. Each row is a contiguous 1D int64 array, so it passes the per-element check. nanobind built 662,243 row objects.
>  * cmap failed, because a single CoordinateElement is not a sequence. All that work was thrown away.
 > * nanobind then tried the 2nd overload. That conversion is cheap, and the call went into C++.
 > nanobind matches keyword names before it converts anything. The 1st overload has no parameter called element (its elements), so nanobind rejects it straight away."
 > On my example mesh, using keywords reduced read_mesh timing from 0.781s to 0.507s.
 > I found this quite surprising, not sure if there is a general fix. It might be a good idea to use less overloading, and give each function (here the mixed dim read_mesh) its own name.
 > On a larger mesh, the nanobind issue alone halves the read_mesh time, the wasted work is proportional to the number of cells.

This still needs a solution.

schnellerhase:
- Fixes https://github.com/FEniCS/dolfinx/pull/4527#discussion_r4034204268
- Fixes https://github.com/FEniCS/dolfinx/issues/4542 by renaming mixed mesh creation to `creat_mixed_mesh` - currently not Python wrapped anyhow and not ready for combination with the `create_mesh` standard path.


garth-wells:
We need to be careful here - it's very unlikely we want a `create_mixed_mesh` when the interface is sorted. Should be possible to route via `create_mesh`. Don't want to get trapped with an interface before the interface is designed.

schnellerhase:
The mixed mesh interface is not yet ready to be included into the public API completely; that's why it has also not been wrapped into a/the Python wrapper of `create_mesh` before. Since we need anyhow another interface merger once this is possible I don't see this counter acting the effort of combining the interface at a later point. As this fixes a
 performance issue that effects all `create_mesh` calls currently, I'd argue this has precedence.

I don't see another elegant fix which does not rely on unclear trickery that will be missed once the next person reads the code. https://github.com/FEniCS/dolfinx/issues/4542#issuecomment-5758101498 does not work, as the order only changes per exported scalar type, so we still interleave mixed and non-mixed `create_mesh` look ups.

jhale:
The new function should be named `_create_mixed_mesh` to clearly denote that it's a private/experimental API.

Edit: Also import it into the demos with the underscore.

PR #4560, opened 2026-09-21
Share neighbourhood communicators via `common::NeighbourhoodComms`

jhale:
`common::Scatterer` created two MPI dist-graph communicators per instance, so every `la::Vector` (and hence every `fem::Function`) built the same pair for the same `IndexMap`, and copying a `Scatterer` was a collective `MPI_Comm_dup`. #4489 made the `Scatterer` shareable between `Vector`s but left the default path creating communicators per object.

This PR adds `common::NeighbourhoodComms`, a move-only class owning the owner→ghost and ghost→owner communicators of an `IndexMap`, constructed collectively from the map. `Scatterer` now takes `(shared_ptr<const IndexMap>, shared_ptr<const NeighbourhoodComms>)` and creates no communicators; copy, cast-copy, move and destruction are no longer collective. `common::create_scatterer(map)` builds the pair for the one-per-layout case. `IndexMap` is unchanged.

`la::Vector(map, bs)` is removed: a `Vector` is constructed as `Vector(scatterer, bs)` and takes its `IndexMap` from the `Scatterer`, so a map can only be scattered where a `Scatterer` (and its communicators) is held, there is no redundant `(map, scatterer)` pair to keep consistent, and the sharing is enforced by the interface rather than by convention. `fem::DofMap` likewise is constructed from a `Scatterer` and owns it next to `index_map`, shared by every `Vector` with that layout; sub-dofmaps share the parent's, `collapse` builds a new one. `fem::Function` construction is therefore no longer collective. Communicator count per function space drops from 2 per `Function` to 2 per `DofMap`. `la::MatrixCSR` still holds its own ghost→owner communicator; that can be switched to the row map's `NeighbourhoodComms` in a follow-up.

Python: `dolfinx.common.NeighbourhoodComms`/`neighbourhood_comms(index_map)`, `scatterer(index_map, comms=None)`, `DofMap.scatterer`; `la.vector(scatterer, bs=1)` replaces `la.vector(map, bs)`, and `fem.create_vector(V)` uses the dofmap's scatterer.

**API changes:** `Scatterer(map, comms)` (was `Scatterer(const IndexMap&)`), `DofMap(element, scatterer, index_map_bs, dofmap, bs)` (the `index_map` argument is replaced by `scatterer`), `la::Vector(scatterer, bs)` (was `Vector(map, bs)`), Python `la.vector(scatterer, bs=1)` (was `la.vector(map, bs=1)`), `_cpp.fem.DofMap.__init__` takes `scatterer` instead of `index_map`.

**Alternatives considered.** (1) `IndexMap` owns the two communicators, created in its constructors: simplest, but every `IndexMap` then holds three communicators and pays two collective `MPI_Dist_graph_create_adjacent` calls whether or not it is ever scattered (mesh entity maps, sub-index maps, temporaries in the dofmap builder), and `IndexMap` becomes a resource owner rather than an index layout. It moves the cost from "many `Function`s" to "many `IndexMap`s" instead of removing it. (2) Lazy creation inside `IndexMap` on first use: avoids the cost for unscattered maps, but hides a collective operation behind a `const` accessor (`mutable` members, `once_flag` for thread safety), which is contrary to the rule that collective calls are explicit. (3) Using a dist-graph communicator as the `IndexMap`'s own `comm()`, or a single symmetric graph for both directions: rejected because a topology-attached communicator returned from `comm()` is a user-visible change in semantics. The explicit resource object keeps `IndexMap` pure, makes creation visibly collective, and puts the communicators with the objects that scatter.

**Tests:** `NeighbourhoodComms` graph topology and move; 2048 `Scatterer`s on one `NeighbourhoodComms` plus copy/cast-copy/move/destroy on a single rank (would exhaust MPICH's communicators or deadlock with per-`Scatterer` communicators); `Function`s on one space share the dofmap `Scatterer`, sub-spaces share it, collapsed spaces do not; 2048 `Function`s on one space, then destruction on a single rank and a forward scatter; Python equivalents.

Design and implementation were developed with Claude Code (Opus 5); the human contributor reviewed and is responsible for the changes.


cpp/dolfinx/common/NeighbourhoodComms.cpp
```diff
@@ -0,0 +1,49 @@
...
+//
+// SPDX-License-Identifier:    LGPL-3.0-or-later
+
+#include "NeighbourhoodComms.h"
+#include "IndexMap.h"
+#include "MPI.h"
+#include <mpi.h>
+#include <span>
+
+using namespace dolfinx;
```
jhale, line 13: Remove

python/dolfinx/common.py
```diff
@@ -463,17 +498,42 @@ def create_sub_index_map(
...
+    Args:
+        index_map: Index map that describes the communication pattern.
+
+    Returns:
+        Neighbourhood communicators for ``index_map``.
+    """
+    return NeighbourhoodComms(_cpp.common.NeighbourhoodComms(index_map._cpp_object))
+
+
+def scatterer(index_map: IndexMap, comms: NeighbourhoodComms | None = None) -> Scatterer:
```
jhale, line 517: Best to avoid easy path?

python/dolfinx/fem/dofmap.py
```diff
@@ -85,6 +85,19 @@ def index_map(self) -> IndexMap:
         """
         return IndexMap(self._cpp_object.index_map)
 
+    @functools.cached_property
+    def scatterer(self) -> Scatterer:
+        """Scatterer for :attr:`index_map`.
+
+        Shared by all vectors with this layout, e.g. those created by
```
jhale, line 92: Implies the sharing is automatic when it isn't

python/dolfinx/la/__init__.py
```diff
@@ -47,7 +47,15 @@
 
 
 class Vector(Generic[_T]):
-    """Distributed vector object."""
+    """Distributed vector object.
+
+    Vectors with the same layout share a :class:`dolfinx.common.Scatterer`
```
jhale, line 52: Implies sharing is automatic when it isn't

python/dolfinx/la/__init__.py
```diff
@@ -530,20 +538,26 @@ def matrix_csr(
...
     bs: int = 1,
-    scatterer: Scatterer | None = None,
     *,
     dtype: npt.DTypeLike = np.float64,
 ) -> Vector:
     """Create a distributed vector.
 
+    The scatterer holds the index map that describes the size and
+    distribution of the vector, the communication pattern and the MPI
+    communicators, and is shared by all vectors with the same layout.
```
jhale, line 550: Implies sharing is automatic when it isn't

schnellerhase:
I think `NeighbourhoodComms` and thus the comm should be rather attached to the `IndexMap`:
- purely depends on IndexMap's `map.comm()`, `map.src()` and  `map.dest()`.
- `IndexMap::shared_indices`, `IndexMap::weights_dest` rebuild the same comm, showing the need for the comm to exist there not just on derived types.

garth-wells:
I'm not sold on this change. It brings complexity. Also, minimising objects being attached to other objects is a good thing.

The present design already supports the advanced case of a user supplying a scatterer, for the very few cases where this is necessary. For advanced usage we document requirements; we don't need to guard absolutely against a user shooting themselves in the foot, i.e. the user can take responsibility for passing compatible index maps and scatterers.

jhale:
1. @schnellerhase If we're happy to have a design where some comms are made that are not used (an indexmap doesn't have to be scattered) then having the comms on the indexmap is probably the smoothest design.

2. @garth-wells We're still missing the Function(V) path. A user could pass their own Vector to Function to wrap, based on a common scatterer. There are also our own library calls to the Function constructor.

Indeed, these two possible fixes are ones where a user doesn't need to think about this issue again. My view is that we should have a design where we're not unnecessarily duplicating comms. The current design seems inelegant and produces unnecessary footguns for users.

jhale:
P.S. I will still try both of the alternatives out and see how they look.

garth-wells:
> 2. @garth-wells We're still missing the Function(V) path. A user could pass their own Vector to Function to wrap, based on a common scatterer. There are also our own library calls to the Function constructor.
> 
> Indeed, these two possible fixes are ones where a user doesn't need to think about this issue again. My view is that we should have a design where we're not unnecessarily duplicating comms. The current design seems inelegant and produces unnecessary footguns for users.

For duplicating Vectors, we can simply add a `clone/duplicate` method.

I think this proposal is less elegant than the present design. More code, greater complexity, and `common::NeighbourhoodComms` is isn't specific to neighbourhood communicators. It's basically an array of length 2 holding two MPI communicators.

I don't mind looking at this again, but I think we need to raise the design level at which we look at it. LLMs drop quickly into the weeds.

